PR #14869, opened 2025-10-03
Improve performances

charlesBochet:
## Improvements

- Add logs to all gql operations and rest calls to help debug CPU issues on the backend. These are temporary and should be removed
- Remove nested relations from workflowVersions load (used to add manual triggers in the side bar). On some workspaces this call result in a response of 4MB which is heavy on CPU
- Investigated Redis Usage ==> made a few improvements, we are should still migrate to the new cache service once available
- investigated db calls in messaging / calendar fetch list + workflow enqueue run cron jobs. Everything seems to be properly batched


packages/twenty-server/src/engine/core-modules/cache-storage/services/cache-storage.service.ts
```diff
@@ -15,7 +15,9 @@ export class CacheStorageService {
   ) {}
 
   async get<T>(key: string): Promise<T | undefined> {
-    return this.cache.get(this.getKey(key));
+    const value = await this.cache.get(this.getKey(key));
```
Weiko:
```typescript
    const value = await this.cache.get<T>(this.getKey(key));

    return value;
 ```
but I'm pretty sure it can infer better if you return directly

File: packages/twenty-server/src/engine/workspace-cache-storage/services/get-data-from-cache-with-recompute.service.ts [hunk 2]
```diff
@@ -0,0 +1,90 @@
...
+      }
+    }
+
+    this.cache.set(cacheKey, {
+      version: cachedVersion,
+      data: cachedData,
+    });
+
+    return {
+      version: cachedVersion as T,
```
Weiko: Not needed I think?

Weiko (approved):
LGTM

prastoin:
👀

prastoin:
👀

packages/twenty-server/src/engine/metadata-modules/workspace-permissions-cache/workspace-permissions-cache.service.ts
```diff
@@ -109,44 +122,48 @@ export class WorkspacePermissionsCacheService {
...
-      getCacheVersion: () =>
-        this.workspacePermissionsCacheStorageService.getRolesPermissionsVersion(
-          workspaceId,
-        ),
-      recomputeCache: (params) => this.recomputeRolesPermissionsCache(params),
-      cachedEntityName: ROLES_PERMISSIONS,
-      exceptionCode: TwentyORMExceptionCode.ROLES_PERMISSIONS_VERSION_NOT_FOUND,
-      logger: this.logger,
-    });
+    return this.getRolesPermissionsFromCacheWithRecomputeService.getFromCacheWithRecompute(
```
prastoin, line 125: ***Remark:*** We could consume the new `WorkspaceManyOrAllFlatEntityMapsCacheService` instead

packages/twenty-server/src/modules/workflow/workflow-trigger/automated-trigger/listeners/workflow-database-event-trigger.listener.ts
```diff
@@ -134,70 +136,91 @@ export class WorkflowDatabaseEventTriggerListener {
     payload: WorkspaceEventBatch<ObjectRecordCreateEvent>,
   ) {
     const workspaceId = payload.workspaceId;
+    const { objectMetadataMaps, objectMetadataItemWithFieldsMaps } =
+      await this.workflowCommonWorkspaceService.getObjectMetadataItemWithFieldsMaps(
+        payload.events[0].objectMetadata.nameSingular,
```
prastoin, line 141: Could be undefined

PR #14871, opened 2025-10-03
Publish twenty-cli npm package

martmull:
as title


packages/twenty-cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "twenty-cli",
-  "version": "0.1.0",
+  "version": "1.1.1",
```
Weiko: Why this big bump? I think CLI is still considered "alpha" or something close to that at that stage 😄

packages/twenty-cli/project.json
```diff
@@ -1,9 +1,18 @@
 {
   "name": "twenty-cli",
   "$schema": "../../node_modules/nx/schemas/project-schema.json",
-  "projectType": "application",
+  "projectType": "library",
```
Weiko: Why this change? This is still an application that you run. Library would be for twenty-shared or twenty-sdk for example imho

packages/twenty-cli/package.json
martmull: ok, what version should I choose then?

packages/twenty-cli/project.json
martmull: ok my bad

packages/twenty-cli/project.json
```diff
@@ -4,6 +4,15 @@
   "projectType": "application",
   "tags": ["scope:cli"],
   "targets": {
+    "deploy": {
+      "executor": "nx:run-commands",
+      "cache": true,
+      "options": {
+        "cwd": "packages/twenty-cli",
+        "commands": ["npm publish --access public"]
```
Weiko: I don't think this should be there, should be a CD responsibility / github action

packages/twenty-apps/martin-app/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "license": "MIT",
+  "engines": {
+    "node": "^24.5.0",
+    "npm": "please-use-yarn",
+    "yarn": ">=4.0.2"
+  },
+  "packageManager": "yarn@4.9.2",
+  "$schema": "https://raw.githubusercontent.com/twentyhq/twenty/main/packages/twenty-cli/schemas/appManifest.schema.json",
+  "universalIdentifier": "4cc2fdf3-7e0a-4bec-bcc8-6724ad82e250",
+  "name": "Martin application",
```
Weiko: Can we call it "Hello world" or something? This is more standard 😛

martmull: this is an error, i will remove

packages/twenty-cli/src/services/api.service.ts
```diff
@@ -143,4 +143,53 @@ export class ApiService {
...
+        '/metadata',
+        {
+          query: mutation,
+          variables,
+        },
+        {
+          headers: {
+            'Content-Type': 'application/json',
+            Accept: '*/*',
+            'x-schema-version': '6',
```
Weiko: Not needed, this is for metadata api

Weiko (approved):

PR #14873, opened 2025-10-03
Cache flush after database reset

prastoin:
When starting the app on a fresh database reset the cache would be filled with empty flat field metadata maps
Because the reset command hack through the repository directly in order to create views and stuff
implemented an iso flush as the one existing initially
added it to a workspace deletion tambien
```
{
   byId: {],
   universalIdById: {}
}
```

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/application/application-sync.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/application/application-sync.service.ts
@@ -190,7 +190,7 @@ export class ApplicationSyncService {
       await this.flatEntityMapsCacheService.getOrRecomputeManyOrAllFlatEntityMaps(
         {
           workspaceId,
-          flatEntities: ['flatObjectMetadataMaps'],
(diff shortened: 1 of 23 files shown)


prastoin:
@greptileai trigger

Weiko (commented):
Can't we use flushByPattern instead? @prastoin This would simplify a lot and I don't think executeActionForManyOrAllFlatEntity is needed otherwise

prastoin:
That's a very smart point :P 
@Weiko we could indeed have a fallback for flushing all you're right
About to add this to the global service

prastoin:
Thinking about it it's not enough isn't it ? We still need to flush the localCache too @Weiko -> I'm a fool flushing remote will result in recomputing cache anw

Weiko:
> Thinking about it it's not enough isn't it ? We still need to flush the localCache too @Weiko

Indeed, wouldn't work in this case

Weiko (approved):
LGTM

PR #14884, opened 2025-10-03
Handle empty values for boolean form field

Devessier:
This PR uses a select component instead of `BooleanInput` for boolean form fields. This makes it clear that the default value is `undefined`.

## Before

https://github.com/user-attachments/assets/84061638-0ba0-4f79-965e-90517d45aa3e

## After

https://github.com/user-attachments/assets/cdbcf935-5c5b-46d1-a1cc-110670b460ee

Fixes https://github.com/twentyhq/twenty/issues/14851


martmull (approved):

Devessier:
@greptileai

PR #14886, opened 2025-10-03
esBuild serverless

### LINKED ISSUE #1599: [Extensibility] Allow multiple files serverlessFunction
as title, today, we only use src/index.ts file, we would like to be able to run whole project serverless functions

martmull:
Use [esbuild](https://esbuild.github.io/) to Build serverless entire folder instead of just the src/index.ts file. Application serverless folders can contain code organized in folders

## Small example

This now works:

image

### DIFF
--- a/packages/twenty-cli/src/services/api.service.ts
+++ b/packages/twenty-cli/src/services/api.service.ts
@@ -55,10 +55,9 @@ export class ApiService {
   async validateAuth(): Promise<boolean> {
     try {
       const query = `
-        query FindManyAgents {
-          findManyAgents {
+        query GetCurrentUser {
+          currentUser {
             id
-            name
           }
         }
       `;
--- a/packages/twenty-cli/src/services/api.service.ts
+++ b/packages/twenty-cli/src/services/api.service.ts
@@ -55,10 +55,9 @@ export class ApiService {
   async validateAuth(): Promise<boolean> {
     try {
       const query = `
-        query GetCurrentUser {
-          currentUser {
-            id
-          }
+        query CurrentWorkspace {
+          currentWorkspace {
+          id
         }
       `;
 
--- a/packages/twenty-server/src/engine/core-modules/serverless/drivers/lambda.driver.ts
+++ b/packages/twenty-server/src/engine/core-modules/serverless/drivers/lambda.driver.ts
@@ -46,6 +46,7 @@ import {
   ServerlessFunctionExceptionCode,
 } from 'src/engine/metadata-modules/serverless-function/serverless-function.exception';
 import { buildServerlessFunctionInMemory } from 'src/engine/core-modules/serverless/drivers/utils/build-serverless-function-in-memory';
+import { formatBuildError } from 'src/engine/core-modules/serverless/drivers/utils/format-build-error';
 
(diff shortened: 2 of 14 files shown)


packages/twenty-server/src/engine/core-modules/serverless/drivers/lambda.driver.ts
```diff
@@ -318,68 +316,92 @@ export class LambdaDriver implements ServerlessDriver {
...
+          await buildServerlessFunctionInMemory(sourceTemporaryDir);
+      } catch (error) {
+        return {
+          data: null,
+          logs: '',
+          duration: Date.now() - startTime,
+          error: {
+            errorType: 'BuildError',
+            errorMessage:
+              // @ts-expect-error legacy noImplicitAny
```
FelixMalfait: We can do better here I guess (at least unknown always better than any)

packages/twenty-server/src/engine/core-modules/serverless/drivers/local.driver.ts
```diff
@@ -106,111 +104,119 @@ export class LocalDriver implements ServerlessDriver {
...
+          duration: Date.now() - startTime,
+          error: {
+            errorType: 'UnhandledError',
+            errorMessage: error.message || 'Unknown error',
+            stackTrace: error.stack ? error.stack.split('\n') : [],
+          },
+          status: ServerlessFunctionExecutionStatus.ERROR,
+        };
+      } finally {
+        // Restoring originalConsole
```
FelixMalfait: Remove comment?

packages/twenty-cli/src/services/api.service.ts
```diff
@@ -55,10 +55,9 @@ export class ApiService {
   async validateAuth(): Promise<boolean> {
     try {
       const query = `
-        query FindManyAgents {
-          findManyAgents {
+        query GetCurrentUser {
```
Weiko: I don't think this endpoint works with 'api' token.

packages/twenty-server/src/engine/core-modules/serverless/drivers/local.driver.ts
```diff
@@ -106,111 +104,119 @@ export class LocalDriver implements ServerlessDriver {
...
+          await buildServerlessFunctionInMemory(sourceTemporaryDir);
+      } catch (error) {
+        return {
+          data: null,
+          logs: '',
+          duration: Date.now() - startTime,
+          error: {
+            errorType: 'BuildError',
+            errorMessage:
+              // @ts-expect-error legacy noImplicitAny
```
FelixMalfait: same

packages/twenty-cli/src/services/api.service.ts
martmull: ok, I need to find something else then

FelixMalfait: Something related to metadata since the api will need to have metadata access (but does not need to have data access)

martmull: CurrentWorkspace

FelixMalfait (approved):

PR #14887, opened 2025-10-04
Fix Align all field-level permissions #14695 issue

mattedelman:


FelixMalfait:
Thanks! closing is code is dirty (commented bits, no PR description, etc.)

PR #14775, opened 2025-09-30
Fix: missing settings tab due to YAML parsing

Jmarin123:
The [settings page](https://twenty.com/user-guide/section/settings/settings) doesn't exist.

The settings.mdx had an info section with the value `Learn how to manage your workspace: permissions...`.

The string contained a colon which caused YAML to misinterpret the value. Added quotes to fix.
image

PR #14891, opened 2025-10-04
Update documentation

BOHEUS:
Fix small things and update documentation with more troubleshooting information


BOHEUS:
Forgot to write yesterday but could someone fix URL for implementation services (https://twenty.com/implementaion-services) before merging this PR? If not done, this specific page will become unaccessible by normal means

packages/twenty-website/src/content/developers/self-hosting/troubleshooting.mdx
```diff
@@ -120,6 +120,37 @@ That's expected as user is unauthorized when logged out since its identity is no
 </div>
 - If a **POST request** is received, your worker is running successfully. Otherwise, you need to troubleshoot your worker.
 
+#### Front-end fails to start and returns error TS5042: Option 'project' cannot be mixed with source files on a command line
```
charlesBochet, line 123: @BOHEUS does this happen for self hosting users or for contributors? this seems to be self-hosting section

packages/twenty-website/src/content/developers/self-hosting/troubleshooting.mdx
```diff
@@ -120,6 +120,37 @@ That's expected as user is unauthorized when logged out since its identity is no
...
+        babelOptions: {
+          presets: ['@babel/preset-typescript', '@babel/preset-react'],
+        },
+      }),
+    ],
+```
+
+#### Admin panel not accessible
+
+Run `UPDATE core."user" SET "canAccessFullAdminPanel" = TRUE WHERE email = 'you@yourdomain.com';` in database container to get access to admin panel.
```
charlesBochet, line 150: same here, do self-hosters havec access to adminPanel by default?

charlesBochet: For self hosters we should fix the issues "in-product", for contributors we can add a troubleshooting section to the local setup guide (note that we tend not to over invest on this section as it gets quickly outdated)

packages/twenty-website/src/content/user-guide/constants/UserGuideIndex.ts
```diff
@@ -34,7 +34,6 @@ export const USER_GUIDE_INDEX = {
       { fileName: 'import-export-data' },
     ],
     Settings: [
-      { fileName: 'settings' },
```
charlesBochet, line 37: isn't this an issue, looks like all other section have a fileName with same value as the section

packages/twenty-website/src/content/user-guide/getting-started/configure-workspace-in-three-steps.mdx
```diff
@@ -85,8 +85,8 @@ Here is how to proceed:
 -  **Save** your view and rename it
 
 -  **Choose the layout** 
-    - You can switch to a **Kanban layout** or a **Group By** layout -- as long as the object has a “Stage” or similar select-type field.  
--  **Add it to your Favorties** 
+    - You can switch to a **Kanban layout** or a **Group By** layout - as long as the object has a “Stage” or similar select-type field.
+-  **Add it to your Favourites**
```
charlesBochet: Favorites

charlesBochet (commented):
Thank you @BOHEUS, this is very helpful.

I've left a few comments!

(make sure to run the website when you make non trivial modifications:  npx nx run twenty-website:dev )

packages/twenty-website/src/content/developers/self-hosting/troubleshooting.mdx
BOHEUS: I assume it happens for contributors, it never happened to me but I saw one community member asking for this tips for others https://discord.com/channels/1130383047699738754/1130383048173682821/1421249429377257482

packages/twenty-website/src/content/user-guide/getting-started/configure-workspace-in-three-steps.mdx
BOHEUS: So :us: English :+1:

packages/twenty-website/src/content/user-guide/constants/UserGuideIndex.ts
BOHEUS: This was done to remove non-existing link in left menu, should I revert this then? 
image

packages/twenty-website/src/content/developers/self-hosting/troubleshooting.mdx
BOHEUS: Self-hosters don't have an access to adminPanel by default as far as I can tell, also, it was added based on frequent questions regarding access to admin panel (as admin role doesn't give access to admin panel)

packages/twenty-website/src/content/user-guide/constants/UserGuideIndex.ts
charlesBochet: Yes, the reason why it's appearing as blank is likely different as other sections do have it. I had already investigated similar issue in the past but I forgot what was the cause of this blank link

packages/twenty-website/src/content/developers/self-hosting/troubleshooting.mdx
charlesBochet: @FelixMalfait what do you think, should we make the first user admin if there is no other user?

packages/twenty-website/src/content/user-guide/constants/UserGuideIndex.ts
charlesBochet: I think it was this PR: https://github.com/twentyhq/twenty/pull/13972/files

charlesBochet: I think it's actually already fixed on main by this PR: https://github.com/twentyhq/twenty/pull/14775/files

BOHEUS: Reverted :+1:

charlesBochet (approved):

packages/twenty-website/src/content/developers/self-hosting/troubleshooting.mdx
charlesBochet: in this case, it should go to the local-setup section?

charlesBochet:
One last comment about "#### Front-end fails to start and returns error TS5042: Option 'project' cannot be mixed with source files on a command line" => if this is for contributors, it should be part of local setup and not self-hosting

BOHEUS:
@charlesBochet currently Troubleshooting page contains solutions for both self-hosting users and contributors, should this be separated then?

charlesBochet:
> @charlesBochet currently Troubleshooting page contains solutions for both self-hosting users and contributors, should this be separated then?

I am not sure to be honest, I think that historically we had a section under local setup but it got outdated so we removed it. Re-introducing it will likely ends up in the same result. Ideal solution would be to resolve this "in-product", so for contributors, would be great to catch the error message and to propose the solution in the command line output for instance

To me it dont make sense to add this kind of troubleshooting notes to the self-hosting that will never face it. 

Btw, I actually don't understand this fix and removing the checkers (so typescript) does not seems to be a good idea. Isn't it the root cause that people have spaces in their folder path?

charlesBochet:
Yes, it seems to be a space issue in folder path:
image


So, here is what I think we should do:
- for self-hosters, we keep maintaining the troubleshooting section for now
- for contributors, we don't. Instead: at the end of the local setup section, we recommend joining the Discord server and search the error on the server (I just add to type the error code to find the post see screenshot above), and then ask on #help channel

BOHEUS:
>I am not sure to be honest, I think that historically we had a section under local setup but it got outdated so we removed it.

Yes, it was there but it was changed with this PR #8312 

Personally, I don't think forcing users to join Discord to look up the answer for their problem is a good solution if it can be in documentation, after all, if there's a problem, most people search through the internet to find a way to resolve said problem. I'd highly suggest creating separate section for contributors in Troubleshooting or separate page but if those ideas are not good enough, I'll just move them to thread in #development channel and pin it.

charlesBochet:
Let's merge it!

PR #14894, opened 2025-10-04
Fix exception handling on messaging + optimistic on object delete

charlesBochet:
## What

In this PR, we are fixing two issues:
- while deleting an object, the views are not properly refreshed leading to FE bug. This is due to the fact that we were refetching **views associated to the object** which has been deleted. As views are properly deleted in the BE, the FE gets no views from this call and cannot optimistically react. In this case, I'm triggering a complete view refetch
- messaging error handling had a hole

### DIFF
--- a/packages/twenty-front/src/modules/views/hooks/useRefreshAllCoreViews.ts
+++ b/packages/twenty-front/src/modules/views/hooks/useRefreshAllCoreViews.ts
@@ -1,10 +1,10 @@
+import { FIND_ALL_CORE_VIEWS } from '@/views/graphql/queries/findAllCoreViews';
 import { coreViewsState } from '@/views/states/coreViewState';
 import { type FetchPolicy, useApolloClient } from '@apollo/client';
 import { useRecoilCallback } from 'recoil';
 import { isDefined } from 'twenty-shared/utils';
-import { type FindManyCoreViewsQuery } from '~/generated/graphql';
+import { type FindAllCoreViewsQuery } from '~/generated/graphql';
 import { isDeeplyEqual } from '~/utils/isDeeplyEqual';
-import { FIND_MANY_CORE_VIEWS } from '../graphql/queries/findManyCoreViews';
 
 export const useRefreshAllCoreViews = (
   fetchPolicy: FetchPolicy = 'network-only',
@@ -14,8 +14,8 @@ export const useRefreshAllCoreViews = (
   const refreshAllCoreViews = useRecoilCallback(
     ({ snapshot, set }) =>
       async () => {
-        const result = await client.query<FindManyCoreViewsQuery>({
-          query: FIND_MANY_CORE_VIEWS,
+        const result = await client.query<FindAllCoreViewsQuery>({
+          query: FIND_ALL_CORE_VIEWS,
           variables: {},
           fetchPolicy,
         });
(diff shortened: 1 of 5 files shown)


packages/twenty-server/src/modules/messaging/message-import-manager/services/messaging-import-exception-handler.service.ts
```diff
@@ -230,8 +229,15 @@ export class MessageImportExceptionHandlerService {
     messageChannel: Pick<MessageChannelWorkspaceEntity, 'id'>,
     workspaceId: string,
   ): Promise<void> {
-    if (syncStep === MessageImportSyncStep.FULL_MESSAGE_LIST_FETCH) {
-      return;
+    if (syncStep === MessageImportSyncStep.MESSAGE_LIST_FETCH) {
+      await this.handleUnknownException(
```
charlesBochet, line 233: this should never happen (I think we should actually refactor this part to stop trying to parse message-import and message-list-fetch exception in one place, they are two differents). Right now the logic is in common and NotFound should never happen in case of message list fetch

packages/twenty-server/src/modules/messaging/message-import-manager/services/messaging-import-exception-handler.service.ts
```diff
@@ -25,7 +25,6 @@ import {
 } from 'src/modules/messaging/message-import-manager/exceptions/message-import.exception';
 
 export enum MessageImportSyncStep {
-  FULL_MESSAGE_LIST_FETCH = 'FULL_MESSAGE_LIST_FETCH', // TODO: deprecate to only use MESSAGE_LIST_FETCH
```
charlesBochet, line 28: let's remove it as we were not handling both MESSAGE_LIST_FETCH and FULL_MESSAGE_LIST_FETCH

packages/twenty-server/src/modules/messaging/message-import-manager/services/messaging-import-exception-handler.service.ts
```diff
@@ -230,8 +229,15 @@ export class MessageImportExceptionHandlerService {
     messageChannel: Pick<MessageChannelWorkspaceEntity, 'id'>,
     workspaceId: string,
   ): Promise<void> {
-    if (syncStep === MessageImportSyncStep.FULL_MESSAGE_LIST_FETCH) {
-      return;
+    if (syncStep === MessageImportSyncStep.MESSAGE_LIST_FETCH) {
```
charlesBochet, line 232: for instance this condition was not checked in case of FULL_MESSAGE_LIST_FETCH

packages/twenty-front/src/modules/views/hooks/useRefreshAllCoreViews.ts
```diff
@@ -14,8 +14,8 @@ export const useRefreshAllCoreViews = (
   const refreshAllCoreViews = useRecoilCallback(
     ({ snapshot, set }) =>
       async () => {
-        const result = await client.query<FindManyCoreViewsQuery>({
-          query: FIND_MANY_CORE_VIEWS,
+        const result = await client.query<FindAllCoreViewsQuery>({
+          query: FIND_ALL_CORE_VIEWS,
```
charlesBochet, line 18: I'm surprise this is breaking now

PR #14895, opened 2025-10-04
Fix Admin control to disable workspace creation for non-admin users

### LINKED ISSUE #13460: Admin control to disable workspace creation for non-admin users
### Scope & Context
This feature request is about providing administrators with control over who can create new workspaces. In certain deployments, it's necessary to centralize instance management by preventing non-admin users from creating their own workspaces.

The goal is to introduce a configuration setting that restricts the ability to create workspaces to admin-level users only.

### Current behavior
Currently, any authenticated user, regardless of their role, has the ability to create a new workspace. The option is available within the user's workspace management menu.

### Expected behavior
A new configuration setting should be introduced to allow administrators to disable workspace creation for non-admin users. This could be implemented either as an environment variable for the instance or as an option within the admin setup GUI.

When this setting is enabled:
* Only users with the 'admin' role will see the UI elements for creating a new workspace.
* For all other non-admin users, these UI elements will be hidden.

ketanMehtaa:
fix #13460


packages/twenty-server/src/engine/core-modules/auth/services/sign-in-up.service.ts
```diff
@@ -355,6 +355,24 @@ export class SignInUpService {
     return { canImpersonate: false, canAccessFullAdminPanel: false };
   }
 
+  async isWorkspaceCreationAllowed(currentUser: User) {
+    const isWorkspaceCreationLimited = this.twentyConfigService.get(
+      'IS_WORKSPACE_CREATION_LIMITED_TO_WORKSPACE_ADMINS',
+    );
+
+    if (isWorkspaceCreationLimited) {
```
FelixMalfait: Instead of nest ifs, prefer early returns

packages/twenty-server/src/engine/core-modules/auth/auth.resolver.ts
```diff
@@ -490,6 +490,8 @@ export class AuthResolver {
     @AuthUser() currentUser: User,
     @AuthProvider() authProvider: AuthProviderEnum,
   ): Promise<SignUpOutput> {
+    await this.signInUpService.isWorkspaceCreationAllowed(currentUser);
```
FelixMalfait: Naming is a bit strange because we expect it to return a boolean but it's more like checkWorkspaceCreationIsAllowedOrThrow

packages/twenty-server/src/engine/core-modules/twenty-config/config-variables.ts
```diff
@@ -358,6 +358,15 @@ export class ConfigVariables {
...
   EMAIL_SMTP_PASSWORD: string;
 
+  @ConfigVariablesMetadata({
+    group: ConfigVariablesGroup.Other,
+    description:
+      'When enabled, only workspace admins can create new workspaces. Ignored during initial setup when no workspace exists.',
+    type: ConfigVariableType.BOOLEAN,
+  })
+  @IsOptional()
+  IS_WORKSPACE_CREATION_LIMITED_TO_WORKSPACE_ADMINS = false;
```
FelixMalfait: I was wrong in naming it this way, it's not Workspace Admins but Server Admins, sorry

FelixMalfait (approved):
Thank you!

PR #14914, opened 2025-10-06
chore(dependencies): remove unused dependencies from package.json and…

AMoreaux:
… yarn.lock


packages/twenty-server/package.json
```diff
@@ -116,15 +118,13 @@
...
     "imapflow": "1.0.187",
     "ioredis": "5.6.0",
     "jsdom": "^26.1.0",
     "json-schema": "0.4.0",
     "jsonc-eslint-parser": "2.4.0",
     "jsonwebtoken": "9.0.2",
-    "jwt-decode": "^4.0.0",
     "libphonenumber-js": "1.11.5",
     "lodash.camelcase": "4.3.0",
     "lodash.chunk": "4.2.0",
```
FelixMalfait, line 130: Still used in twenty-front I think? 

Image

https://github.com/auth0/jwt-decode/blob/main/lib/index.ts not a big lib though

packages/twenty-server/package.json
```diff
@@ -168,10 +170,8 @@
     "rxjs": "7.8.1",
     "semver": "7.6.3",
     "sharp": "0.32.6",
-    "slash": "5.1.0",
     "stripe": "17.3.1",
     "transliteration": "2.3.5",
-    "ts-morph": "^24.0.0",
     "tsconfig-paths": "^4.2.0",
     "tsdav": "^2.1.5",
```
FelixMalfait, line 176: Let's move it away from dependencies of root package json and move it to the dev dep of /twenty-shared only?

packages/twenty-server/package.json
AMoreaux: So the lib should appear in the package.json of twenty-front. I add it

FelixMalfait (approved):

packages/twenty-server/package.json
AMoreaux: ts-morph seams unused.

PR #14915, opened 2025-10-06
feat: workflow delay action (Pause - Wait/Sleep/Delay)

harshit078:
## Description

- This PR focuses on issue https://github.com/orgs/twentyhq/projects/1/views/33?pane=issue&itemId=93150683&issue=twentyhq%7Ccore-team-issues%7C20
- added Workflow delay as a Flow action
- for V1 added Type 1: Resume at a specific date or time

## Visual Appearance
Screenshot 2025-10-09 at 5 46 18 PM

Screenshot 2025-10-09 at 5 46 35 PM


FelixMalfait:
I feel like the `scheduledDateTime` use case might be harder to achieve based on your implementation (and you could argue it could be a trigger). Let's focus on `duration` for v1. We can always add 2 types later

FelixMalfait:
Sorry I'll close to keep the backlog clean because we have way to many PRs open but let me know when you need more feedback we can re-open any time!

packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/delay-actions/components/WorkflowEditActionDelay.tsx
```diff
@@ -0,0 +1,122 @@
...
+        ...action.settings,
+        input: {
+          scheduledDateTime: formData.scheduledDateTime,
+        },
+      },
+    });
+  }, 1_000);
+
+  useEffect(() => {
+    saveAction(formData);
```
FelixMalfait: Is this what is done in other components? There is a useEffect for saveAction.flush but not for this? Let's try to stick to existing patterns

packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/delay-actions/components/WorkflowEditActionDelay.tsx
```diff
@@ -0,0 +1,122 @@
...
+          dropdownId="workflow-edit-action-delay-duration"
+          label="Resume"
+          options={delayOptions}
+          value="specific_date_time"
+          onChange={() => {}}
+        />
+        <HorizontalSeparator noMargin />
+
+        <FormDateTimeFieldInput
+          label="Delay Until"
```
FelixMalfait: Use "t" from lingui for all translatable items

packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/delay-actions/components/WorkflowEditActionDelay.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+          actionOptions.onActionUpdate({
+            ...action,
+            name: newTitle,
+          });
+        }}
+      />
+
+      <WorkflowStepBody>
+        <Select
```
FelixMalfait, line 163: Let's drop this? Also it's not at a specific date it's the opposite!

packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/utils/getActionIconColorOrThrow.ts
```diff
@@ -22,8 +22,10 @@ export const getActionIconColorOrThrow = ({
     case 'FORM':
       return theme.color.orange;
     case 'ITERATOR':
-    case 'FILTER':
     case 'EMPTY':
+    case 'DELAY':
+      return theme.color.green60;
```
FelixMalfait, line 28: Image This looks odd? Please refer to Figma or check with @Bonapara

packages/twenty-server/src/modules/workflow/workflow-executor/workflow-actions/delay/delay.workflow-action.ts
```diff
@@ -0,0 +1,77 @@
...
+    });
+
+    if (!isWorkflowDelayAction(step)) {
+      throw new WorkflowStepExecutorException(
+        'Step is not a delay action',
+        WorkflowStepExecutorExceptionCode.INVALID_STEP_TYPE,
+      );
+    }
+
+    const settings = step.settings.input as WorkflowDelayActionInput;
```
FelixMalfait: Why do you need "as"? Looks like a code smell / something not properly types before if you needed to add this

packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/delay-actions/components/WorkflowEditActionDelay.tsx
FelixMalfait: I told you to do duration only and I think you did fixed data only 😅. Since we have the wrong one let's go with both then!

packages/twenty-server/src/modules/workflow/workflow-executor/workflow-actions/delay/jobs/delayed-resume.job.ts
```diff
@@ -0,0 +1,72 @@
...
+@Processor({ queueName: MessageQueue.workflowQueue, scope: Scope.REQUEST })
+export class DelayedResumeJob {
+  constructor(
+    @InjectMessageQueue(MessageQueue.workflowQueue)
+    private readonly messageQueueService: MessageQueueService,
+    private readonly workflowRunWorkspaceService: WorkflowRunWorkspaceService,
+    private readonly workflowRunQueueWorkspaceService: WorkflowRunQueueWorkspaceService,
+  ) {}
+
+  @Process(DelayedResumeJob.name)
```
FelixMalfait: I find the naming DelayedResumeJob strange? What are you trying to do here? ResumeDelayedJob maybe?

packages/twenty-server/src/modules/workflow/workflow-executor/workflow-actions/delay/delay.workflow-action.ts
```diff
@@ -0,0 +1,77 @@
...
+        WorkflowStepExecutorExceptionCode.INVALID_STEP_TYPE,
+      );
+    }
+
+    const scheduledDate = new Date(settings.scheduledDateTime);
+    const now = new Date();
+    const delayInMs = scheduledDate.getTime() - now.getTime();
+
+    await this.messageQueueService.add<DelayedResumeJobData>(
+      'DelayedResumeJob',
```
FelixMalfait: If the queue is full then the delay might not be respected so we might want to do a higher priority queue for those jobs (MessageQueue enum: add workflowDelayedJobsQueue). We don't use priorities right now but a good opportunity to start introducing them. Let's also put the emailQueue in priority / do something clean. lower numbers = higher priority in BullMQ.

packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/delay-actions/components/WorkflowEditActionDelay.tsx
harshit078: Yep, I'll use the shared T module

packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/utils/getActionIconColorOrThrow.ts
harshit078: I couldnt find a color for filter on Figma. Maybe I might have missed it looking there. I'll push to make sure it uses the same green color.

packages/twenty-server/src/modules/workflow/workflow-executor/workflow-actions/delay/delay.workflow-action.ts
harshit078: yep, I was removing this as type as while implementing I might have output it as I went along. I'll remove this in the next commit.

packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/utils/getActionIconColorOrThrow.ts
Bonapara: let's use the green tag color like other nodes from the Flow section

FelixMalfait: Should the filter also become green @Bonapara? (also in the flow section)

thomtrp (commented):
@harshit078 thank you for this PR! Taking over that work if you are ok 👍

thomtrp (commented):
Feature seems working well! @martmull @charlesBochet I would like your opinion on having a separated queue for delayed jobs. Also, if we set priorities, I guess we should re-order? Messaging could be first for example

PR #14916, opened 2025-10-06
Prevent querying too much on worklow index page

charlesBochet:
## Performance short term fix

Done in this PR:
workflowVersions and workflowRuns are heavy object. I'm preventing loading their data when they are loaded as relations. This will reduce the work on backend

## Long term fix

Todo:
We should make sure workflow data is ligher


martmull (approved):

packages/twenty-front/src/modules/object-record/record-field/hooks/useRecordFieldGqlFields.ts
```diff
@@ -37,38 +38,50 @@ export const useRecordFieldGqlFields = ({
...
+            recordField.fieldMetadataItemId
+          ];
 
-      if (!isDefined(fieldMetadataItem)) {
-        throw new Error(
-          `Field ${recordField.fieldMetadataItemId} is missing, please refresh the page. If the problem persists, please contact support.`,
-        );
-      }
+        // TODO: remove this once we have made the workflowVersion lighter
+        if (
```
prastoin, line 56: ***Bug*** Should be placed below invariant check

prastoin (approved):
From my understanding optimistic cache might be returning partial workflow when we expect complete one ?
Left a comment !

PR #14919, opened 2025-10-06
chore(workflows): update permissions across GitHub Actions workflows …

AMoreaux:
…for consistency


FelixMalfait (approved):
Can you check by changing the root package.json or something? (e.g. dotenv-cli appears in both dev and non-dev dependencies)

PR #14924, opened 2025-10-06
Potential fix for code scanning alert no. 173: Uncontrolled data used in path expression

AMoreaux:
Potential fix for [https://github.com/twentyhq/twenty/security/code-scanning/173](https://github.com/twentyhq/twenty/security/code-scanning/173)

To fix this vulnerability, ensure that any file path constructed from untrusted user input is strictly confined to the intended storage directory. This is best done through path normalization and validation. Before reading the file, the following steps should be taken:
1. **Resolve the file path**: Use `path.resolve` to normalize the resulting path, removing any ".." segments or symbolic links.
2. **Check the parent directory**: Ensure that the resolved path starts with the intended storage directory (`this.options.storagePath`).
3. **Throw an exception**: If the resolved path does not start with the storage root, throw an exception or otherwise deny the operation.

This check should be placed in all methods that build a file path from user input in `local.driver.ts`, with at least the `read` method addressed. The index for this fix will be on lines that construct and use `filePath`, specifically in the `read` method.

You will need to:
- Import `realpathSync` from `fs` (standard library).
- Add safe path resolution and containment check around file access.

---

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/file-storage/drivers/local.driver.ts
+++ b/packages/twenty-server/src/engine/core-modules/file-storage/drivers/local.driver.ts
@@ -83,17 +83,20 @@ export class LocalDriver implements StorageDriver {
       params.folderPath,
       params.filename,
     );
+    let filePath: string;
 
-    if (!existsSync(joinedPath)) {
+    try {
+      filePath = realpathSync(path.resolve(joinedPath));
+    } catch {
       throw new FileStorageException(
         'File not found',
         FileStorageExceptionCode.FILE_NOT_FOUND,
       );
     }
-
     const storageRoot = realpathSync(path.resolve(this.options.storagePath));
 
-    if (!joinedPath.startsWith(storageRoot + path.sep)) {
+    if (!filePath.startsWith(storageRoot + path.sep)) {
+      // Prevent directory traversal
       throw new FileStorageException(
         'Access denied',
         FileStorageExceptionCode.FILE_NOT_FOUND,
@@ -101,7 +104,7 @@ export class LocalDriver implements StorageDriver {
     }
 
     try {
-      return createReadStream(joinedPath);
+      return createReadStream(filePath);
     } catch (error) {
       if (error.code === 'ENOENT') {
         throw new FileStorageException(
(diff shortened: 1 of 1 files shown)


packages/twenty-server/src/engine/core-modules/file-storage/drivers/local.driver.ts
```diff
@@ -78,11 +78,28 @@ export class LocalDriver implements StorageDriver {
...
+    const storageRoot = realpathSync(path.resolve(this.options.storagePath));
+    if (!filePath.startsWith(storageRoot + path.sep)) {
+      // Prevent directory traversal
+      throw new FileStorageException(
+        'Access denied',
+        FileStorageExceptionCode.FILE_NOT_FOUND,
+      );
+    }
 
     if (!existsSync(filePath)) {
```
FelixMalfait: Duplicate?

packages/twenty-server/src/engine/core-modules/file-storage/drivers/local.driver.ts
```diff
@@ -78,11 +78,28 @@ export class LocalDriver implements StorageDriver {
...
+    try {
+      filePath = realpathSync(path.resolve(joinedPath));
+    } catch (err) {
+      throw new FileStorageException(
+        'File not found',
+        FileStorageExceptionCode.FILE_NOT_FOUND,
+      );
+    }
+    const storageRoot = realpathSync(path.resolve(this.options.storagePath));
+    if (!filePath.startsWith(storageRoot + path.sep)) {
```
FelixMalfait, line 98: Even folderpath shouldn't be allowed no?

FelixMalfait (approved):
It's never constructed from user input but it doesn't hurt. Also Local Driver isn't used in prod (S3 driver). Please address other comments before merging :). Thanks

packages/twenty-server/src/engine/core-modules/file-storage/drivers/local.driver.ts
```diff
@@ -78,21 +78,30 @@ export class LocalDriver implements StorageDriver {
...
   }): Promise<Readable> {
-    const filePath = join(
+    const joinedPath = join(
       `${this.options.storagePath}/`,
       params.folderPath,
       params.filename,
     );
 
-    if (!existsSync(filePath)) {
+    if (!existsSync(joinedPath)) {
```
FelixMalfait: Why keep using existsSync and not keep what you had done with realpathSync check?

AMoreaux: I rolled back this part

Issue #14900, opened 2025-10-06
Make Entire Table Header Cell Clickable for Field Menu


Bonapara:
# Current behavior

On tables, you have to click the header text to open the field menu:

https://github.com/user-attachments/assets/7cf5b069-5688-4c31-b635-12356b5c07e0

# Desired Behavior

The whole header cell should be clickable (only clicking the plus icon should not open this menu but create a new record)

Bonapara:
@constantinidan @willydouhard

vansh2408:
Please assign

runningpoem30:
Hi @Bonapara 
I believe the issue is related to how the header’s div or button element is being handled — the click event might be bound too narrowly, causing the field menu to open only when clicking directly on the text.

I’ve previously been assigned an issue by @charlesBochet and have been actively working on it. Over the past 2–3 days, I’ve become quite familiar with the Twenty repo and its tech stack, and I’m confident I can fix this issue.

Please assign it to me. Thanks!

Bonapara:
Hi @runningpoem30, great to hear! A PR has already been opened for this issue. Thanks a lot for contributing to Twenty.

msboffl:
Hi @Bonapara Is this issue still open?

FelixMalfait:
It's been completed. Thanks!

PR #14925, opened 2025-10-06
Make entire first cell clickable

pvrnn:
Solution: 
I removed the specific layout of the first cell that was reducing the clickable area to make room for the “+” button.
Instead, I added a layout to the button, so it’s position absolute, allowing the clickable area to cover 100% of the cell while keeping the button above it.
Another possible solution would be to modify `RecordTableColumnHeadWithDropdown` to include the button inside...

- Manual test: [Loom](https://www.loom.com/share/4ceb4daa07e5457a99a07a2489db2960?sid=15959502-e272-44bd-923a-5764d077642d)
- Issue: https://github.com/twentyhq/twenty/issues/14900

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-table/record-table-header/components/RecordTableHeaderFirstCell.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-table/record-table-header/components/RecordTableHeaderFirstCell.tsx
@@ -7,6 +7,7 @@ import { RecordTableHeaderResizeHandler } from '@/object-record/record-table/rec
 import { RecordTableHeaderCellContainer } from '@/object-record/record-table/record-table-header/components/RecordTableHeaderCellContainer';
 
 import { hasRecordGroupsComponentSelector } from '@/object-record/record-group/states/selectors/hasRecordGroupsComponentSelector';
+import { TABLE_Z_INDEX } from '@/object-record/record-table/constants/TableZIndex';
 import { RecordTableHeaderLabelIdentifierCellPlusButton } from '@/object-record/record-table/record-table-header/components/RecordTableHeaderLabelIdentifierCellPlusButton';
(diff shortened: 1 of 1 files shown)


packages/twenty-front/src/modules/object-record/record-table/record-table-header/components/RecordTableHeaderFirstCell.tsx
```diff
@@ -21,13 +21,11 @@ import { cx } from '@linaria/core';
...
-  display: flex;
-  flex-direction: row;
-  justify-content: space-between;
-
-  overflow: hidden;
+const StyledPlusButtonWrapper = styled.div`
+  position: absolute;
+  top: 0;
+  right: 0;
+  z-index: 2;
```
FelixMalfait: Can you please use TABLE_Z_INDEX.headerColumnswithoutGroups.headerColumnsSticky and check if it still works? Thanks!

FelixMalfait: Especially when scrolling on small screen / try edge cases

pvrnn:
Ipad dimension: 
image

Iphone SE dimension:
image

packages/twenty-front/src/modules/object-record/record-table/record-table-header/components/RecordTableHeaderFirstCell.tsx
pvrnn: Hey, thanks for the quick feedback, I added 2 screenshots for Ipad and Iphone dimension.

I just noticed that on phone dimension **(with or without my changes)**, when it’s in “mini size” (just company logo, no name) and I click on the header first cell, it doesn’t show the dropdown or the “+” button, but I guess that’s WAD?

FelixMalfait (approved):
Thank you, great work!

PR #14934, opened 2025-10-07
Fix auth + cli -V not picking up package json

Weiko:
- Auth was failing, GQL query was not well formatted
- -V was not working as expected and was reading hardcoded version instead of package.json
- CommanderError due to early exit (with -V for example) were thrown and logged, now not logging those


martmull (approved):
My bad thank you

PR #14943, opened 2025-10-07
Updates on chart selection

bosiraphael:
- Updated the disabled design
- Disabled the 3 charts which won't be released this sprint
- Updated the icon of the number chart
- Added labels

Before:
CleanShot 2025-10-07 at 15 40 31@2x

After:
CleanShot 2025-10-07 at 15 46 21@2x


ehconitin (approved):
LGTM

PR #14944, opened 2025-10-07
Fix app deletion missing cache invalidation

Weiko:
App deletion currently deletes all related entities through postgres cascade delete meaning we never run actual migrations ourself and never trigger cache invalidation (part of migration engine)


martmull (approved):
thanks

prastoin:
Not really the solution here, we should define a devxp in order to handle cache invalidation discussing with @Weiko

PR #14953, opened 2025-10-07
feat(auth): enhance error handling for sign-up and existing user checks

AMoreaux:
Refactored error handling in `useSignUpInNewWorkspace` and `useSignInUp` hooks to provide user-friendly messages. Added validation for existing users during sign-up, updating server-side logic to throw specific exceptions (`USER_ALREADY_EXIST`).

Fix https://twenty-v7.sentry.io/issues/6686753138/events/latest/?environment=prod&project=4507072499810304&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D%20QueryFailedError&referrer=latest-event&sort=date


packages/twenty-server/src/engine/core-modules/auth/services/sign-in-up.service.ts
```diff
@@ -465,6 +466,20 @@ export class SignInUpService {
     newUserParams: SignInUpNewUserPayload,
     authParams: AuthProviderWithPasswordType['authParams'],
   ) {
+    const existingUser = await this.userRepository.findOne({
```
FelixMalfait: Duplicating the check already done in signInUpWithSocialSSO? Can we do something cleaner to avoid checking the same thing twice?

FelixMalfait (approved):
Thanks! Approving but probably we can improve the server side code to avoid a double check + need to fix tests. Do you think this was the cause of users creating multiple workspaces? Would be nice if it was!

PR #14964, opened 2025-10-07
fixing storybook

guillim:
Storybook had  issues. This PR intends to solve them

Main things to care :
I had to change the way the provider was given to the Calendar view. Not sure how it may impact the final component, so if someone with some knowledge about it could take a look that would be great

TODO:
Only one remaining test to fix : prefetch loading. Couldn't find out how to solve this issue. When you open the story you see that the companies are not fetched.


packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/input/components/__stories__/RelationOneToManyFieldInput.stories.tsx
Devessier, line 134: When testing, I can't see any displayed records.

Furthermore, the standard is always to define stories with a `play` function that asserts elements are correctly rendered. Chromatic doesn't monitor stories without interactions.

Image

Devessier (approved):
💪 

Thanks for unblocking the stories. I left a follow-up comment.

PR #14700, opened 2025-09-24
fix: updated dropdown's bg for better readability over bright bg

meezumi:
# Fix dropdown readability in dark theme over bright backgrounds

## Problem

In dark theme, dropdowns wrapped by `OverlayContainer` had poor readability when displayed over bright/white backgrounds. The issue was caused by `theme.background.transparent.primary` having only 50% opacity, which didn't provide sufficient contrast against bright backgrounds even with the blur effect applied.

## Solution

Changed the background property in `OverlayContainer` from `theme.background.transparent.primary` to `theme.background.overlayPrimary`.

### Before
```typescript
background: ${({ theme }) => theme.background.transparent.primary};
```

### After
```typescript
background: ${({ theme }) => theme.background.overlayPrimary};
```

## Testing
Screenshot 2025-09-24 213507

Screenshot 2025-09-24 213649

## Technical Details

The `overlayPrimary` background provides better contrast across both themes:

- **Dark theme**: `RGBA(gray100, 0.8)` - 80% opacity white overlay for better contrast over bright backgrounds
- **Light theme**: `RGBA(gray80, 0.8)` - 80% opacity dark overlay for consistent contrast

## Files Changed

- `packages/twenty-front/src/modules/ui/layout/overlay/components/OverlayContainer.tsx`

The change ensures that dropdown content remains clearly readable regardless of what's behind it, while maintaining the aesthetic blur effect. Test by:

1. Switch to dark theme
2. Open a dropdown over a bright/white background image or element
3. Verify that dropdown content is clearly readable
4. Test in light theme to ensure no regression

Fixes #11916

PR #14965, opened 2025-10-07
Darker Dropdown for Improved Visibility

DavidSebastian05:
This is just a simple fix. Since text readability was low due to the background blur effect, I manually made the background of the dropdown darker, which should help readability across both light and dark mode.

Fixes #14700


charlesBochet:
@DavidSebastian05 thanks for taking a look! Unfortunately this issue has not been created by the core team and @Bonapara  don't think it's a good idea.

I'm therefore going to close this PR, sorry about that :(

PR #14975, opened 2025-10-08
Fix RelationToOneSection assignment bug

charlesBochet:
While refactoring SingleRecordPicker to support morph use case, we forgot to QA RelationDetailSection for ManyToOne.


packages/twenty-front/src/modules/object-record/record-field-list/record-detail-section/relation/components/RecordDetailRelationSectionDropdownToOne.tsx
```diff
@@ -84,7 +84,7 @@ export const RecordDetailRelationSectionDropdownToOne = () => {
     if (!selectedMorphItem?.recordId || !relationFieldMetadataItem?.name)
       return;
 
-    onSubmit?.({ newValue: selectedMorphItem.recordId });
+    onSubmit?.({ newValue: { id: selectedMorphItem.recordId } });
```
Weiko, line 87: Is it not possible to have it strongly typed?

Weiko (approved):
LGTM

PR #14989, opened 2025-10-08
Return data or raise error in serverless controller

### LINKED ISSUE #1588: [Serverless] Follow ups
- Route should be able to return html (it looks to do it already but check again) (add a return format column and use it in the execute service)
- change serverlessFunction.serverlessFunctionLayer to not nullable -> Need to handle common layer, and use a layer record

martmull:
as title
TODO validate we do not need to add a new column

### DIFF
--- a/packages/twenty-server/src/engine/metadata-modules/route-trigger/route-trigger.controller.ts
+++ b/packages/twenty-server/src/engine/metadata-modules/route-trigger/route-trigger.controller.ts
@@ -6,20 +6,17 @@ import {
   Post,
   Put,
   Req,
-  UseFilters,
   UseGuards,
 } from '@nestjs/common';
 
 import { Request } from 'express';
 
-import { RestApiExceptionFilter } from 'src/engine/api/rest/rest-api-exception.filter';
 import { PublicEndpointGuard } from 'src/engine/guards/public-endpoint.guard';
 import { HTTPMethod } from 'src/engine/metadata-modules/route-trigger/route-trigger.entity';
 import { RouteTriggerService } from 'src/engine/metadata-modules/route-trigger/route-trigger.service';
 
 @Controller('s')
 @UseGuards(PublicEndpointGuard)
-@UseFilters(RestApiExceptionFilter)
 export class RouteTriggerController {
   constructor(private readonly routeTriggerService: RouteTriggerService) {}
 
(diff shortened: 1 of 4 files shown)


packages/twenty-server/src/engine/metadata-modules/route-trigger/route-trigger.service.ts
```diff
@@ -133,11 +135,18 @@ export class RouteTriggerService {
...
+    const result =
+      await this.serverlessFunctionService.executeOneServerlessFunction(
+        routeTriggerWithPathParams.routeTrigger.serverlessFunction.id,
+        routeTriggerWithPathParams.routeTrigger.workspaceId,
+        executionParams,
+        'draft',
+      );
+
+    if (result.error) {
+      throw new HttpException(result.error, HttpStatus.INTERNAL_SERVER_ERROR);
```
Weiko: Services should be agnostic of the kind of API (if one day we want to call this with a resolver, we don't want http exception but gql error) imho we should move this to the controller instead

packages/twenty-server/src/engine/metadata-modules/route-trigger/route-trigger.controller.ts
```diff
@@ -6,20 +6,17 @@ import {
...
 import { Request } from 'express';
 
-import { RestApiExceptionFilter } from 'src/engine/api/rest/rest-api-exception.filter';
 import { PublicEndpointGuard } from 'src/engine/guards/public-endpoint.guard';
 import { HTTPMethod } from 'src/engine/metadata-modules/route-trigger/route-trigger.entity';
 import { RouteTriggerService } from 'src/engine/metadata-modules/route-trigger/route-trigger.service';
 
 @Controller('s')
 @UseGuards(PublicEndpointGuard)
-@UseFilters(RestApiExceptionFilter)
```
Weiko, line 22: Why are we removing this?

packages/twenty-server/src/engine/metadata-modules/route-trigger/route-trigger.service.ts
martmull: ok, well seen

packages/twenty-server/src/engine/metadata-modules/route-trigger/route-trigger.controller.ts
martmull: because this is catching HttpException and format serverless error message

packages/twenty-server/src/engine/metadata-modules/route-trigger/utils/format-response.ts
```diff
@@ -0,0 +1,20 @@
+import { HttpException, HttpStatus } from '@nestjs/common';
+
+import { isDefined } from 'twenty-shared/utils';
+
+import { type ServerlessExecuteResult } from 'src/engine/core-modules/serverless/drivers/interfaces/serverless-driver.interface';
+
+export const formatResponse = (result?: ServerlessExecuteResult | null) => {
```
Weiko: Can we use a less generic name for that

Weiko (approved):

PR #14990, opened 2025-10-08
Move `view` in `metadata-modules/` and create atomic folder + module for each view entity

### LINKED ISSUE #1703: move view in metadata modules and refactor file tree structure to be atomic to each entity

prastoin:
# Introduction
Preparing view-filter and view-group introduction in v2 core engine
Moving view from `core-modules` to `metadata-modules`

### Created dedicated modules for each view entity:
- ViewFieldModule
- ViewFilterModule
- ViewFilterGroupModule
- ViewGroupModule
- ViewSortModule

### Each module is now completely independent with its own:
- Controller
- Resolver
- Service
- Entity

### Created dedicated abstraction metadata module folder for:
- flat-view-field
- flat-view

### Dependencies 
- Eleminated circular dep on ViewModule to all others ones
- Granular import not importing the whole viewModule anymore everywhere

close https://github.com/twentyhq/core-team-issues/issues/1703


Weiko (commented):
Can we have a dedicated "view-field" module inside "view" instead? I know we mix services and types and such in the view module but I don't think this will scale well with all the things we need to introduce for migration v2 (also this will allow us to import only viewField when necessary without the whole view module). Ideally we should have viewField at the same level as view but I'm ok having them as subfolders for now @prastoin

prastoin:
Agreed

Weiko (dismissed):
LGTM! @prastoin

prastoin:
@Weiko thanks for the review but in the end went down the tunnel, please find updated PR desc

prastoin:
@greptileai trigger

prastoin:
We should exclude locales mutation

prastoin:
@greptileai trigger please search for anything duplicated or deleted and not moved

prastoin:
All good here

packages/twenty-server/src/engine/metadata-modules/view-field/services/tests/view-field.service.spec.ts
```diff
@@ -50,10 +49,15 @@ describe('ViewFieldService', () => {
           },
         },
         {
-          provide: ViewService,
+          provide: getRepositoryToken(ViewEntity),
```
charlesBochet, line 52: **thought**: didn't now about this getRepositoryToken!

packages/twenty-server/src/modules/view/standard-view.module.ts
```diff
@@ -0,0 +1,30 @@
...
+      ViewSortEntity,
+    ]),
+    TypeOrmModule.forFeature([ObjectMetadataEntity]),
+    FeatureFlagModule,
+  ],
+
+  providers: [],
+  exports: [],
+})
+export class StandardViewModule {}
```
charlesBochet, line 30: **question**: not sure about this standardView naming, is this module actually useful? (does not provides nor exports)

charlesBochet (approved):
LGTM, I've left only one comment!

packages/twenty-server/src/modules/view/standard-view.module.ts
prastoin: I think that's just a abstraction pattern, the naming wasn't accurate due to the nature of the below business logic
It was colliding with mine which was more legitimate IMO
No strong opinion here

PR #14996, opened 2025-10-09
feat: mutualize CRUD tools between workflows and AI

### LINKED ISSUE #1662: Extract crud tools from workflow to mutualize with AI
Like we have done with Send Email, we need to mutualize the tools between workflows and AI and not duplicate code.

We should mutualize all the CRUD tools.

abdulrahmancodes:
Closes [#1662](https://github.com/twentyhq/core-team-issues/issues/1662)


packages/twenty-server/src/modules/workflow/workflow-executor/workflow-actions/record-crud/types/workflow-record-crud-action-input.type.ts
```diff
@@ -4,7 +4,7 @@ import {
 } from 'src/engine/api/graphql/workspace-query-builder/interfaces/object-record.interface';
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-type ObjectRecord = Record<string, any>;
+export type ObjectRecord = Record<string, any>;
```
FelixMalfait: There is workspace-query-builder/interfaces/object-record.interface already, I think it will cause conflicts/unclarity.

packages/twenty-server/src/engine/metadata-modules/field-metadata/composite-types/actor.composite-type.ts
```diff
@@ -12,6 +12,7 @@ export enum FieldActorSource {
   EMAIL = 'EMAIL',
   CALENDAR = 'CALENDAR',
   WORKFLOW = 'WORKFLOW',
+  AGENT = 'AGENT',
```
FelixMalfait, line 15: It will be interesting to handle this. I think for workflow and for agent there should be 2 modes in the future, one that acts on behalf of users and respects user's permission. And one that bypasses user permissions and act as an independent workflow/agent directly like you did here.

packages/twenty-server/src/engine/core-modules/record-crud/services/find-records.service.ts
```diff
@@ -0,0 +1,183 @@
+import { Injectable, Logger } from '@nestjs/common';
+
+import { type Entity } from '@microsoft/microsoft-graph-types';
```
FelixMalfait: Error

packages/twenty-server/src/engine/core-modules/record-crud/services/find-records.service.ts
```diff
@@ -0,0 +1,183 @@
...
+    const withFilterQueryBuilder = graphqlQueryParser.applyFilterToBuilder(
+      queryBuilder,
+      objectName,
+      filter ?? ({} as ObjectRecordFilter),
+    );
+
+    const orderByWithIdCondition = [
+      ...(orderBy ?? []),
+      { id: OrderByDirection.AscNullsFirst },
+    ] as ObjectRecordOrderBy;
```
FelixMalfait: There are still quite a lot of "as xxx" in this PR, could you please check if they can be replaced by proper typing `const orderByWithIdCondition: ObjectRecordOrderBy` instead? (I know some of this might be old code that's ported but let's not lose opportunities to improve!). Thank you!

FelixMalfait (approved):
A few minor comment. Great work!

PR #14998, opened 2025-10-09
1588 serverless follow ups 2

### LINKED ISSUE #1588: [Serverless] Follow ups

martmull:
- set `serverlessFunction.serverlessFunctionLayer` not nullable
- add proper migration file with null `serverlessFunctionLayer` record management
- remove useless `latestVersionInputSchema` and `layerVersion` columns from `serverlessFunctionEntity`


packages/twenty-server/src/engine/metadata-modules/serverless-function/serverless-function.entity.ts
```diff
@@ -79,16 +72,16 @@ export class ServerlessFunctionEntity
...
-  @Column({ nullable: true, type: 'uuid' })
-  serverlessFunctionLayerId: string | null;
+  @Column({ nullable: false, type: 'uuid' })
+  serverlessFunctionLayerId: string;
 
   @ManyToOne(
     () => ServerlessFunctionLayerEntity,
     (serverlessFunctionLayer) => serverlessFunctionLayer.serverlessFunctions,
-    { nullable: true },
+    { nullable: false },
```
Weiko, line 81: Should we set on delete cascade?

packages/twenty-server/src/engine/core-modules/serverless/drivers/lambda.driver.ts
```diff
@@ -134,11 +133,7 @@ export class LambdaDriver implements ServerlessDriver {
   }
 
   private getLayerName(serverlessFunction: ServerlessFunctionEntity) {
```
Weiko, line 135: Seems the function should be renamed. Although I'm not really sure this one is still really bringing value?

packages/twenty-server/src/engine/metadata-modules/serverless-function/serverless-function.service.ts
```diff
@@ -314,20 +314,30 @@ export class ServerlessFunctionService {
   }
 
   async createOneServerlessFunction(
-    serverlessFunctionInput: CreateServerlessFunctionInput,
+    serverlessFunctionInput: CreateServerlessFunctionInput & {
+      serverlessFunctionLayerId?: string;
```
Weiko, line 318: Can't we add this to the input object?

packages/twenty-server/src/engine/metadata-modules/serverless-function/services/serverless-function-v2.service.ts
```diff
@@ -28,10 +30,10 @@ export class ServerlessFunctionV2Service {
...
   ) {}
 
-  async createOne(
-    serverlessFunctionInput: CreateServerlessFunctionInput,
-    workspaceId: string,
-  ) {
+  async createOne({
+    createServerlessFunctionInput,
+    workspaceId,
+  }: FromCreateServerlessFunctionInputToFlatServerlessFunctionArgs) {
```
Weiko, line 36: Nit: This type is a bit confusing to use here I think

Weiko (approved):
Comments to address otherwise LGTM

packages/twenty-server/src/engine/metadata-modules/serverless-function/serverless-function.entity.ts
martmull: I would say no, restrict if we want to specify something

packages/twenty-server/src/engine/core-modules/serverless/drivers/lambda.driver.ts
martmull: It makes clear layerName eq checksum

packages/twenty-server/src/engine/metadata-modules/serverless-function/serverless-function.entity.ts
martmull: @OneToMany(
    () => ServerlessFunctionEntity,
    (serverlessFunction) => serverlessFunction.serverlessFunctionLayer,
    {
      onDelete: 'RESTRICT',
    },
  )
  serverlessFunctions: Relation<ServerlessFunctionEntity[]>;
  Already restrict

PR #15011, opened 2025-10-09
Add middleware boundary padding as a prop in dropdown

bosiraphael:
Add middleware boundary padding as a prop in dropdown + fix bottom end placement in graph editor


lucasbordeau (approved):
LGTM

PR #15012, opened 2025-10-09
Add timeoutMs to webhooks calls

### LINKED ISSUE #1656: [Post mortem 07/10] Add a timeout on call webhook job
Solves worker down issue root cause A

See https://twenty-eng.twenty.com/object/note/e6577f89-d6d2-4790-af95-b7c1683c116b

We might look at bullmq groups also to make sure that a single workspaceId does not take all the workers
https://docs.bullmq.io/bullmq-pro/groups

martmull:
5 second timeout
tested with an application serverlessFunction that delays

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/webhook/jobs/call-webhook.job.ts
+++ b/packages/twenty-server/src/engine/core-modules/webhook/jobs/call-webhook.job.ts
@@ -74,7 +74,10 @@ export class CallWebhookJob {
       const response = await this.httpService.axiosRef.post(
         getAbsoluteUrl(data.targetUrl),
         payloadWithoutSecret,
-        { headers },
+        {
+          headers,
+          timeout: 5_000,
+        },
       );
 
       const success = response.status >= 200 && response.status < 300;


packages/twenty-server/src/engine/core-modules/webhook/jobs/call-webhook.job.ts
```diff
@@ -74,7 +74,10 @@ export class CallWebhookJob {
       const response = await this.httpService.axiosRef.post(
```
charlesBochet, line 74: **question:** have you chekced that in case of timeout, we go in the catch section where we log the issue?

charlesBochet (approved):
LGTM, left a question

charlesBochet:
Tests are read though!

packages/twenty-server/src/engine/core-modules/webhook/jobs/call-webhook.job.ts
martmull: yes

PR #15013, opened 2025-10-09
feat: display specific action types in workflow side panel tooltips

### LINKED ISSUE #14878: Workflow - Display type of action on side panel icon hover
# Desired Behavior

Add a tooltip with the node name on node icon hover in the workflow side panel

Image

## Original issue

once a user edits the title of an action, it is no longer possible to know which action was selected - which is confusing for ex for update vs create. It prevents other users from collaborating on this workflow and will be painful for the workflow creator when going back at it in a few weeks / months.

**Example:** 

Image
Image

## Expected behavior

Idea: Replace "Action" by the true action "Update Record", "Create Record"

cc @martmull @thomtrp

DipakHalkude:
## What I've Implemented

I've added tooltips that show the original action type when you hover over workflow step icons in the side panel. Now even if you rename an action to something custom, you can still see what type of action it actually is by hovering over the icon.

## The Problem This Solves

Before this change, once you renamed a workflow action (like changing "Create Record" to "Add New Customer"), there was no way to tell what the original action type was. This made it really confusing when collaborating with others or when coming back to your own workflows after some time - you couldn't tell if an action was a "Create Record" or "Update Record" or something else.

## What Changed

Updated action type labels: Instead of showing just "Action" for all record operations, the system now shows specific types like "Create Record", "Update Record", "Delete Record", etc.

Added hover tooltips: When you hover over the action icons in the workflow side panel, a tooltip appears showing the original action type, even if you've renamed the step title.

## Before vs After

Before: All actions showed "Action" in the side panel, making it impossible to distinguish between different types after renaming.

After: Each action shows its specific type in tooltips, so you always know what you're working with.

This should make workflow management much clearer, especially when multiple people are collaborating on the same workflow!

Resolves #14878

## Demo Video

See it in action: https://www.loom.com/share/c0e0ec24e4524d0685b841e9ceb011d0?t=65&sid=dc05e58a-8869-4969-aa67-9772242fe697

### DIFF
--- a/packages/twenty-front/src/modules/command-menu/components/SidePanelHeader.tsx
+++ b/packages/twenty-front/src/modules/command-menu/components/SidePanelHeader.tsx
@@ -3,7 +3,7 @@ import { TitleInput } from '@/ui/input/components/TitleInput';
 import { useTheme } from '@emotion/react';
 import styled from '@emotion/styled';
 import { useState } from 'react';
-import { type IconComponent } from 'twenty-ui/display';
+import { AppTooltip, type IconComponent } from 'twenty-ui/display';
 
 const StyledHeader = styled.div`
   background-color: ${({ theme }) => theme.background.secondary};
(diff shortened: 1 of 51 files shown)


DipakHalkude:
@Bonapara  Please have a look :)

thomtrp:
Hi @DipakHalkude, thank you for contributing. I need to validate the feature we want in that discussion first https://github.com/twentyhq/twenty/issues/14878#issuecomment-3370881095

thomtrp (approved):
Thank you for your work @DipakHalkude!

I added a refacto to use constants for default label and use it as tooltip 👍

DipakHalkude:
Thank you so much for the review and the improvements! 🙏
I really appreciate your feedback and guidance. Glad the implementation works well — happy to make any further tweaks if needed. Wish you a Happy Diwali :)

PR #15017, opened 2025-10-09
Enable filters in iterators

thomtrp:
Filters should not cut the whole workflow. These should only stop the branch. This PR:
- adds a new skipped status 
- when a filter stops, it still goes to the next step
- the next step will execute if there is at least a successful step
- if only skipped step, it will be skipped as well

It allows to use filters in iterators.

https://github.com/user-attachments/assets/1cfca052-55c0-4ce5-9eb8-63736618d082


martmull (approved):
Very nice

PR #15024, opened 2025-10-10
Inter-groups orderBy on groupByDimension values

### LINKED ISSUE #1563: Implement inter-groups orderBy on groupByDimension values
[In this PR](https://github.com/twentyhq/twenty/pull/14690) we allowed orderBy values on groupBy queries to be aggregated values only. This makes sense to sort groups between themselves, for instance I could want to see the average number of employees of companies grouped by their city, and sort the groups by the average number of employees. 

We could also want to sort the group by their city in alphabetical order instead of by average number of employees, or as secondary sorting criteria if we keep the average number of employees criteria but that two groups or more have the exact same average number of employees.

Let's implement that!

Later when we will deal with connections (= loading of records belonging to a group thanks to a separate findMany query), we will add the possibility to add orderBy criteria on mere fields, and that criteria will be used to sort records within a group. this is planned here https://github.com/twentyhq/core-team-issues/issues/1557.

ijreilly:
Closes https://github.com/twentyhq/core-team-issues/issues/1563.

We now have two inter-group orderBy criteria: one on the aggregate values, the other on the dimension values. 
Ex: I am grouping companies by city and querying the average number of employees for each group. I can either order the groups by the city name, or by the average number of employees, or by one then the other. 
I could actually also order groups by an aggregated value that I did not ask for (ex: average ARR), but I cannot order groups by a field value I did not group records by (ex: country).
[See discord discussion](https://discord.com/channels/1130383047699738754/1425438213555753050/1425438223097921659)

An example of query variables: 
```
{
  "groupBy": [
    {
      "createdAt": {
        "granularity": "QUARTER_OF_THE_YEAR"
      }
    },
    {"city": true}
  ],
  "orderBy": [
    { 
      "aggregate": {
        "avgEmployees": "DescNullsLast"
      }
    },
    { 
      "aggregate": {
        "percentyEmptyEmployees": "DescNullsLast"
      }
    },
    {
      "city": "AscNullsLast"
    },
    {
      "createdAt": {
        "orderBy": "AscNullsLast",
        "granularity": "QUARTER_OF_THE_YEAR",
      }
    }
  ]
}
```

The aggregate orderBy criteria had already been implemented, but I updated the implementation to add an "aggregate" key to prefix them, in order to avoid confusion between the two + for the schema generation not to break (otherwise we would have an issue when generating OrderByWithGroupByInput if a user has created a field that has the same name as an aggregate field, such as avgEmployees).


packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-order/graphql-query-order.parser.ts
```diff
@@ -147,87 +203,299 @@ export class GraphqlQueryOrderFieldParser {
...
+  };
+
+  parseObjectRecordOrderByForScalarField = ({
+    groupByFields,
+    orderByArg,
+    objectMetadataItemWithFieldMaps,
+    fieldMetadata,
+  }: {
+    groupByFields: GroupByField[];
+    orderByArg: ObjectRecordOrderByForScalarField;
```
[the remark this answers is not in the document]
ijreilly: you can't order by a date field without indicating a granularity. GraphQL validation will break if you try to because for date fields, a specific input type is created, which expects something like `{ orderBy: direction; granularity: DateGranularityEnum }`. So the query wont be sent in that case

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-order/utils/parse-composite-field-for-order.util.ts
```diff
@@ -0,0 +1,60 @@
...
+import { OrderByDirection } from 'src/engine/api/graphql/workspace-query-builder/interfaces/object-record.interface';
+
+import { type OrderByCondition } from 'src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-order/graphql-query-order.parser';
+import { convertOrderByToFindOptionsOrder } from 'src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-order/utils/convert-order-by-to-find-options-order';
+import { compositeTypeDefinitions } from 'src/engine/metadata-modules/field-metadata/composite-types';
+import { type FieldMetadataEntity } from 'src/engine/metadata-modules/field-metadata/field-metadata.entity';
+import { type CompositeFieldMetadataType } from 'src/engine/metadata-modules/workspace-migration/factories/composite-column-action.factory';
+
+export const parseCompositeFieldForOrder = (
+  fieldMetadata: FieldMetadataEntity,
```
etiennejouan, line 12: I wonder if TS could enable a better typing, kind of FieldMetadataEntity<Composite> that would let us skip the compositeType check + 'as CompositeFieldMetadataType' casting.

It comes often in the codebase, i don't think it has to be tackle here.

Weiko (approved):
LGTM!

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query.parser.ts
```diff
@@ -97,22 +100,28 @@ export class GraphqlQueryParser {
   public applyOrderToBuilder(
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     queryBuilder: WorkspaceSelectQueryBuilder<any>,
-    orderBy: ObjectRecordOrderBy,
+    orderBy: ObjectRecordOrderBy | OrderByWithGroupBy,
     objectNameSingular: string,
     isForwardPagination = true,
     isGroupBy = false,
+    groupByFields?: GroupByField[],
```
etiennejouan: creating a `applyGroupByOrderToBuilder` specific method could lighten the `applyOrderToBuilder` one ?

It's out of this PR scope but I don't know what to think about the .parser. service. W discuss it with Weiko, order-by parser file is getting larger and larger, i wonder how we could improve the code structure.

etiennejouan (approved):
Tested & Approved ! 👏

PR #15028, opened 2025-10-10
Design adjustments on Bar chart

bosiraphael:
- Make the ticks dynamic when we resize the graph
- Fix formatting
- Fix gradient color not working when the index have a space in it
- Fix the maximum number of groups for a grouped by graph
- Update the tooltip design and display the group

Video:

https://github.com/user-attachments/assets/9f304b6c-3dec-4ce2-9127-41d27f393d90


packages/twenty-front/src/modules/page-layout/widgets/graph/utils/formatDimensionValue.ts
```diff
@@ -85,6 +89,25 @@ export const formatDimensionValue = ({
...
     }
 
+    case FieldMetadataType.NUMBER:
+    case FieldMetadataType.CURRENCY: {
+      if (
+        fieldMetadata.type === FieldMetadataType.CURRENCY &&
+        subFieldName === 'currencyCode'
+      ) {
+        if (!isNonEmptyString(value)) {
+          return 'Not Set';
```
ijreilly: use translation

packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetBarChart/utils/computeBarChartCategoryTickValues.ts
```diff
@@ -0,0 +1,32 @@
+import { BAR_CHART_MARGINS } from '@/page-layout/widgets/graph/graphWidgetBarChart/constants/BarChartMargins';
+import { type BarChartDataItem } from '@/page-layout/widgets/graph/graphWidgetBarChart/types/BarChartDataItem';
+
+const MINIMUM_WIDTH_PER_TICK = 100;
```
ijreilly, line 4: I think this may be too wide, or not dynamic enough. 
Capture d’écran 2025-10-13 à 11 00 16

But I tried re-working on it and I found the result to make the very wide charts with lots of bars such as the one in the description to be rendered in a less good version (lots of truncated labels). 
I guess ideally we would want something dynamic depending on the labels length but this may cause a lot re renders when re sizing, idk.
Not a straightforward fix then, so leaving this as is, but we may want to re-work this.

ijreilly (approved):
I think maybe there is still room for improvement for the labels

PR #15033, opened 2025-10-10
Improve and fix source selection in charts

bosiraphael:
- Improvements: Hide advanced objects in a subMenu and order objects alphabetically

CleanShot 2025-10-10 at 17 32 59@2x

CleanShot 2025-10-10 at 17 33 24@2x

- Fix: reset all fields when changing object


Weiko (approved):
LGTM!

PR #15034, opened 2025-10-10
Remove edit widget button and add on click edition

bosiraphael:
Remove edit widget button and add on click edition when in edit mode

Before:

https://github.com/user-attachments/assets/1579aa35-f67e-4ba5-9500-5999c7b2164b

After:

https://github.com/user-attachments/assets/3c7677ab-57f7-485d-b619-116ea9ec6c06

### DIFF
--- a/packages/twenty-front/src/modules/page-layout/widgets/components/WidgetRenderer.tsx
+++ b/packages/twenty-front/src/modules/page-layout/widgets/components/WidgetRenderer.tsx
@@ -8,6 +8,7 @@ import { WidgetHeader } from '@/page-layout/widgets/components/WidgetHeader';
 import { useWidgetPermissions } from '@/page-layout/widgets/hooks/useWidgetPermissions';
 import { useRecoilComponentValue } from '@/ui/utilities/state/component-state/hooks/useRecoilComponentValue';
 import styled from '@emotion/styled';
+import { type MouseEvent } from 'react';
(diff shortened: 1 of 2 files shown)


packages/twenty-front/src/modules/page-layout/widgets/components/WidgetRenderer.tsx
```diff
@@ -32,18 +33,26 @@ export const WidgetRenderer = ({ widget }: WidgetRendererProps) => {
 
   const { hasAccess, restriction } = useWidgetPermissions(widget);
 
+  const handleClick = isPageLayoutInEditMode
```
lucasbordeau: Could we put the if inside instead ? More readable

lucasbordeau (approved):
LGTM small comment

packages/twenty-front/src/modules/page-layout/widgets/components/WidgetRenderer.tsx
bosiraphael: If I put it inside handleClick will never be undefined so a cursor pointer will appear even when we are not in edit mode.
What I can do is
```
 const handleClick = () => {
    handleEditWidget({
      widgetId: widget.id,
      widgetType: widget.type,
    });
  };
```

and `onClick={isPageLayoutInEditMode ? handleClick : undefined}` instead

PR #15035, opened 2025-10-10
Improvement AI chat error handling

abdulrahmancodes:
- Fixed AI chat failures when tool calls were present in conversation history
- Improved error handling and user feedback for AI streaming errors

https://github.com/user-attachments/assets/ca85820f-32c0-4f42-86ee-98f4543e6038


packages/twenty-front/src/modules/ai/components/AIChatMessage.tsx
```diff
@@ -174,6 +184,7 @@ export const AIChatMessage = ({
             <AIChatAssistantMessageRenderer
               isLastMessageStreaming={isLastMessageStreaming}
               messageParts={message.parts}
+              hasError={Boolean(showError)}
```
charlesBochet: we don't use Boolean() in the code base, showError should be typed as a boolean

packages/twenty-front/src/modules/ai/contexts/AgentChatContext.ts
```diff
@@ -19,6 +19,7 @@ export type AgentChatContextValue = {
 
   scrollWrapperId: string;
   context: Array<AIChatObjectMetadataAndRecordContext>;
+  retry: () => void;
```
charlesBochet: onRetry or retry?

packages/twenty-front/src/modules/ai/components/AIChatMessage.tsx
```diff
@@ -139,13 +140,22 @@ const StyledFilesContainer = styled.div`
 export const AIChatMessage = ({
   message,
   isLastMessageStreaming,
+  error,
+  onRetry,
+  isRetrying,
 }: {
   message: UIMessageWithMetadata;
   isLastMessageStreaming: boolean;
+  error?: Error | null;
```
charlesBochet, line 151: we usually avoid passing callbacks to Components:
- this often lead to drilling down the props through multiple components and slow down refactoring as all components props need to be modified. 
- We could then think about adding a Context to resolve the drilldown issue, but then we face a more complex devXP where this is hard to find what the actual callback is

What we recommend:
- if you are dealing with low level UI component, passing callback is fine as we don't drilldown much + there is no other way around, a UI component can be used in various scenarios
- if you are dealing with a business oriented component (which I think is the case here). Avoid passing a callback and direclty call the function through the child component. (you could create a hook useAIChatRetry for instance that would expose a retry method and you can call it from the underlying component)
- in this second case you might actually need to have different retry functions, meaning that you are actually supporting different business (let's imagine you want a different retry behavior for Agent and Chat, in this case we would recommend to actually introduce two different components as business logic inside will be different and then you can still call the method direcltly)

Sometimes we are a bit in a grey zone between UI components (where we would pass a callback) and business logic component (where we would prefer calling precise hook directly)

charlesBochet: In a nutshell avoid adding callbacks unless you are in a low level UI component (ex: <Select onChange /> or <Button onClick />) as this will deterior the dev XP

charlesBochet:
Thank you this is better but we can still go one step further:

1) We should avoid "umbrella hooks" like `useAgentChat` which is returning everything that's needed for the chat. While this is a legit pattern, especially on backend where we tend to encapsulate things to simplify/group things in our cerebral context, this does not cope well with React. React is "reactive" and will re-render if anythings within useAgentChat triggers a re-render (props change, state change...) 
=> let's say I only need the `handleRetry` function from `useAgentChat`, I'll import `useAgentChat` and then my component will start re-rendering for a lot of reason while I only needed the `handleRetry` function which is made to be called synchronously.
=> we prefer introducing dedicated hooks: `useAgentChatRetry()` for instance

2) You are passing the `handleRetry` function in the context. Here there is no need to do that, remove it from the context, and call directly your `const { handleRetry } = useAgentChatRetry()` hook where you need it

abdulrahmancodes:
@charlesBochet Thanks for the feedback! I've addressed the issue and updated the code accordingly

FelixMalfait (approved):

PR #15038, opened 2025-10-10
FIX(#15036): arrows in webhooks

shantanugupta2004:
**Fixes #15036**
Changes:
Centered Webhook URL: The webhook URL is now properly centered within its table cell.
Conditional Scroll Arrows: The small arrows for horizontal scrolling now only appear when the webhook URL is longer than the width of the cell. This prevents the arrows from being displayed for shorter URLs.

Files changed:
1. `packages/twenty-front/src/modules/settings/developers/components/SettingsWebhooksTable.tsx`
The grid-template-columns of the StyledTableRow component was changed from a fixed-width layout (444px 68px) to a flexible layout (1fr auto).
This allows the first column, which contains the webhook URL, to grow and shrink dynamically based on the available space, which is crucial for proper centering and handling of different URL lengths.
2. `packages/twenty-front/src/modules/settings/developers/components/SettingsDevelopersWebhookTableRow.tsx`
The overflow property of the StyledUrlTableCell component was changed from hidden to auto. This ensures that the horizontal scrollbar (the "small arrows") is only displayed when the content of the cell (the webhook URL) overflows the cell's width.The redundant StyledApisFieldTableRow component was removed to simplify the code, as its styles were being overridden by the parent component.

How to Test
1. Log in to the application.
2. Go to Settings > API & Webhooks.
4. Add a new webhook with a short URL (e.g., http://localhost:3000).
5. Expected Behavior: The URL should be centered within the table cell, and no scroll arrows should be visible.
6. Add another webhook with a long URL.
7. Expected Behavior: The URL should be centered, and horizontal scroll arrows should be visible, allowing you
to scroll to see the entire URL.


charlesBochet:
Putting on hold since I cannot reproduce (see https://github.com/twentyhq/twenty/issues/15036)

shantanugupta2004:
I have added the screenshot of the issue reproduced on chrome. I am running the local setup on WSL. Please see #15036 and review.

packages/twenty-front/src/modules/settings/developers/components/SettingsWebhooksTable.tsx
```diff
@@ -16,7 +16,7 @@ const StyledTableBody = styled(TableBody)`
 `;
 
 const StyledTableRow = styled(TableRow)`
-  grid-template-columns: 444px 68px;
+  grid-template-columns: 1fr auto;
```
charlesBochet: this change is unrelated, it's tight to the header but why not :)

charlesBochet:
This is what I see when I checkout the PR:
image

packages/twenty-front/src/modules/settings/developers/components/SettingsDevelopersWebhookTableRow.tsx
```diff
@@ -7,10 +7,6 @@ import { getUrlHostnameOrThrow, isValidUrl } from 'twenty-shared/utils';
 import { IconChevronRight } from 'twenty-ui/display';
 import { type Webhook } from '~/generated-metadata/graphql';
 
-export const StyledApisFieldTableRow = styled(TableRow)`
-  grid-template-columns: 1fr 28px;
```
charlesBochet, line 11: removing this feels wrong, grid layout will then force two equal columns

packages/twenty-front/src/modules/settings/developers/components/SettingsDevelopersWebhookTableRow.tsx
```diff
@@ -19,7 +15,8 @@ const StyledIconTableCell = styled(TableCell)`
 
 const StyledUrlTableCell = styled(TableCell)`
   color: ${({ theme }) => theme.font.color.primary};
-  overflow-x: scroll;
+  overflow: auto;
+  text-overflow: ellipsis;
```
charlesBochet: this is changing the behavior to add an ellispis instead of scrolling. Which could make sense, in this case I think we want to use OverflowingTextWithTooltip

packages/twenty-front/src/modules/settings/developers/components/SettingsDevelopersWebhookTableRow.tsx
charlesBochet: could you revert this?

packages/twenty-front/src/modules/settings/developers/components/SettingsWebhooksTable.tsx
charlesBochet: let's revet this, seems unrelated

charlesBochet:
I think behavior is different on Windows and MacOS. Assuming that, I've reviewed the PR and left a few comments

packages/twenty-front/src/modules/settings/developers/components/SettingsDevelopersWebhookTableRow.tsx
shantanugupta2004: Done.

packages/twenty-front/src/modules/settings/developers/components/SettingsWebhooksTable.tsx
shantanugupta2004: Done.

packages/twenty-front/src/modules/settings/developers/components/SettingsDevelopersWebhookTableRow.tsx
shantanugupta2004: Yes it seems to be the case. I have reverted this and used OverflowingTextWithTooltip.

shantanugupta2004:
I have updated the necessary changes and reverted the redundant parts. Ready for review @charlesBochet

charlesBochet:
Does not seem to be right on my side: 
image

shantanugupta2004:
I think you were right about the behaviour being different on Windows and MacOS. It is showing on my system as I am using WSL on Windows 11:

shantanugupta2004:
Hi @charlesBochet, I have tried to modify the `OverflowingTextWithTooltip.tsx` by adding overflow-x: auto rather than 
overflow: hidden so that the slider is visible on your side as well as I have tested it and it is visible on my system. Please review and check it and suggest any changes if required.

charlesBochet:
@shantanugupta2004 I've updated your PR with I think is the right fix, we should not touch the OverflowingTextWithTooltip behavior unless we have a very good reason to (this component is used in a lot of places, it's working as expected)

We were just missing a overflow: hidden on the cell itself

charlesBochet:
Can you confirm this is working on Windows too and the we can merge it :)

charlesBochet (approved):

shantanugupta2004:
Yes it's working perfectly fine on Windows.

charlesBochet:
Wonderful, thanks for the help ;)

PR #15040, opened 2025-10-11
New resize handle on layout grid

ehconitin:
before - 
Screenshot 2025-10-12 at 02 14 17

after - 

Screenshot 2025-10-14 at 18 51 01


packages/twenty-front/src/modules/page-layout/components/PageLayoutGridResizeHandle.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+const StyledIcon = styled(IconBorderCornerSquare)`
+  bottom: ${({ theme }) => theme.spacing(1)};
+  rotate: 180deg;
+  cursor: nwse-resize;
+  position: absolute;
+  right: ${({ theme }) => theme.spacing(1)};
+`;
+
+export const PageLayoutGridResizeHandle = forwardRef<
```
charlesBochet, line 27: **note:** not a fan of forwardRef but I guess we can't avoid it here because this is Handle is passed to react-grid-layout and that's the way it work, right?

packages/twenty-front/src/modules/page-layout/components/PageLayoutGridResizeHandle.tsx
```diff
@@ -0,0 +1,57 @@
...
+      <StyledIcon
+        size={theme.spacing(4)}
+        stroke={theme.spacing(1)}
+        color={theme.font.color.extraLight}
+      />
+    </div>
+  );
+});
+
+PageLayoutGridResizeHandle.displayName = 'PageLayoutGridResizeHandle';
```
charlesBochet: **question:** non standard too, is this needed?

charlesBochet: or is because of the forwardRef too? (react debug is lost without it?)

charlesBochet (commented):
I have left one question and two minor comments!

packages/twenty-front/src/modules/page-layout/components/PageLayoutGridResizeHandle.tsx
ehconitin: yes! thats the way it works indeed!

packages/twenty-front/src/modules/page-layout/components/PageLayoutGridResizeHandle.tsx
ehconitin: Yes! but I did it just because Claude said so -- was not aware that we do not keep displayName for debugging :)

doesn't add value tbh -- removing it

ehconitin:
We would need @Bonapara 's sign-off on this one!

Bonapara:
Gave @ehconitin the figma links by message

lucasbordeau (approved):
LGTM

PR #15042, opened 2025-10-12
feat: CRON trigger design improvements

msboffl:
issue - https://github.com/twentyhq/twenty/issues/14723

**After**
image

### DIFF
--- a/packages/twenty-front/src/modules/workflow/workflow-trigger/components/WorkflowEditTriggerCronForm.tsx
+++ b/packages/twenty-front/src/modules/workflow/workflow-trigger/components/WorkflowEditTriggerCronForm.tsx
@@ -15,6 +15,7 @@ import { getTriggerIconColor } from '@/workflow/workflow-trigger/utils/getTrigge
 import { useTheme } from '@emotion/react';
 import { t } from '@lingui/core/macro';
 import { isNumber } from '@sniptt/guards';
+import cronstrue from 'cronstrue';
 import { useState } from 'react';
 import { isDefined } from 'twenty-shared/utils';
 import { useIcons } from 'twenty-ui/display';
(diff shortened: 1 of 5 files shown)


packages/twenty-front/src/modules/workflow/workflow-trigger/components/WorkflowEditTriggerCronForm.tsx
```diff
@@ -15,6 +15,7 @@ import { getTriggerIconColor } from '@/workflow/workflow-trigger/utils/getTrigge
 import { useTheme } from '@emotion/react';
 import { t } from '@lingui/core/macro';
 import { isNumber } from '@sniptt/guards';
+import cronstrue from 'cronstrue';
```
charlesBochet, line 18: @FelixMalfait we don't want to add another package to handle cron, right?

charlesBochet (commented):
Hi @msboffl, thanks for opening a PR, before I review that, could you add a PR description and link the original issue? I'm not sure what product behavior change we are trying to achieve here

msboffl:
> Hi @msboffl, thanks for opening a PR, before I review that, could you add a PR description and link the original issue? I'm not sure what product behavior change we are trying to achieve here

@charlesBochet here is the issue - https://github.com/twentyhq/twenty/issues/14723

packages/twenty-front/src/modules/workflow/workflow-trigger/components/WorkflowEditTriggerCronForm.tsx
```diff
@@ -126,10 +133,21 @@ export const WorkflowEditTriggerCronForm = ({
 
                 try {
                   CronExpressionParser.parse(newPattern);
+
+                  const humanReadable = cronstrue.toString(newPattern, {
```
charlesBochet, line 137: I'm a bit surprised/lost, on the issue it looks like we are just moving a text from one section to the other, why do we need to introduce a new library?

charlesBochet:
@msboffl Thank you :) Left a high level question which seems important before we can dive into details

PR #15054, opened 2025-10-13
feat(auth): integrate captcha validation and component for sign-in/up…

AMoreaux:
… process

Added captcha token validation to the sign-in/up flow. Introduced `StyledLoaderContainer` for better UI feedback during loading. Enhanced error handling within `useSignInUp` to handle GraphQL errors effectively.


packages/twenty-front/src/pages/auth/SignInUp.tsx
```diff
@@ -121,7 +135,13 @@ export const SignInUp = () => {
   ]);
 
   const signInUpForm = useMemo(() => {
-    if (loading) return null;
+    if (loading || isUndefinedOrNull(captcha)) {
```
FelixMalfait: Is this for Captcha or CaptchaToken? 

From a UX perspective for Captcha is works fine. But if it was a mistake and was meant to be CaptchaToken then I think we have a UX issue and should rather handle the conflict upon click (maybe an ugly snackbar or something since it should happen so rarely that they click before the captcha is fully loaded ; we don't want to degrade the experience and show a loading state for everyone else during this period just for this small edge case). Let me know if I misunderstood!

AMoreaux: It's for Captcha. It’s a loader that waits for the captcha configuration from the client-config call.
This ensures that users without captcha are unaffected, since the value comes from the same client-config call executed by all users.

FelixMalfait (approved):

FelixMalfait:
As discussed let's not merge yet and add an additional check

FelixMalfait (approved):

PR #15068, opened 2025-10-13
Enqueue a new job every 20 step executions

thomtrp:
To avoid huge workflows to block the worker, we will enqueue a new job every 20 steps.
This could be more than 20 if there are branches but I think this is fine, the goal is only to have a limit set.

Also cleaning a bit the code to mark running steps as failed when workflow fails.

I tested it on a huge workflow:

https://github.com/user-attachments/assets/d7b8e345-d1a1-4467-96fd-92117b500120

### DIFF
--- a/packages/twenty-server/src/modules/workflow/workflow-runner/jobs/run-workflow.job.ts
+++ b/packages/twenty-server/src/modules/workflow/workflow-runner/jobs/run-workflow.job.ts
@@ -12,19 +12,15 @@ import { TwentyConfigService } from 'src/engine/core-modules/twenty-config/twent
 import { WorkflowRunStatus } from 'src/modules/workflow/common/standard-objects/workflow-run.workspace-entity';
 import { WorkflowCommonWorkspaceService } from 'src/modules/workflow/common/workspace-services/workflow-common.workspace-service';
 import { WorkflowExecutorWorkspaceService } from 'src/modules/workflow/workflow-executor/workspace-services/workflow-executor.workspace-service';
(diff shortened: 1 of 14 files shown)


Devessier (approved):

packages/twenty-server/src/modules/workflow/workflow-runner/jobs/run-workflow.job.ts
```diff
@@ -38,7 +34,7 @@ export class RunWorkflowJob {
     private readonly workflowRunQueueWorkspaceService: WorkflowRunQueueWorkspaceService,
   ) {}
 
-  @Process(RunWorkflowJob.name)
+  @Process(RUN_WORKFLOW_JOB_NAME)
```
martmull, line 37: why? we do everywhere else `@Process(<JobClass>.name)`

martmull: Circular import?

martmull (commented):
I am thinking about the issue, but in fact, I don't see the point of stopping a workflow that runs fast even if it has more than 20 steps. In the contrary, a 3 steps workflow can lasts a lot of time and black the workers.

I feel that a **workflow execution timeout** would provide a better control on worker ressource allocation 

What do you think?

thomtrp:
> I am thinking about the issue, but in fact, I don't see the point of stopping a workflow that runs fast even if it has more than 20 steps. In the contrary, a 3 steps workflow can lasts a lot of time and black the workers.
> 
> I feel that a **workflow execution timeout** would provide a better control on worker ressource allocation
> 
> What do you think?

Checked with you directly

PR #15072, opened 2025-10-13
feat(lastViewAt): add recently view on company

AMoreaux:


packages/twenty-server/src/modules/company/standard-objects/company.workspace-entity.ts
```diff
@@ -291,4 +292,20 @@ export class CompanyWorkspaceEntity extends BaseWorkspaceEntity {
...
+    standardId: COMPANY_STANDARD_FIELD_IDS.lastViewedAt,
+    type: FieldMetadataType.DATE_TIME,
+    label: msg`Last view at`,
+    description: msg`Last view at`,
+    icon: 'IconCalendar',
+    defaultValue: null,
+    settings: {
+      displayFormat: DateDisplayFormat.RELATIVE,
+    },
+    storage: 'redis',
```
FelixMalfait, line 306: I think I would have used `isVirtualField: boolean` and then put the storage within the `settings` JSONB
Because as discussed we will have virtual fields on Postgres storage

packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-runner-v2/utils/generate-column-definitions.util.ts
```diff
@@ -197,6 +203,14 @@ export const generateColumnDefinitions = ({
     return relationColumn ? [relationColumn] : [];
   }
 
+  // Skip non-persisted redis-backed fields: company.lastViewedAt
+  if (
+    flatObjectMetadata.standardId === STANDARD_OBJECT_IDS.company &&
+    flatFieldMetadata.standardId === COMPANY_STANDARD_FIELD_IDS.lastViewedAt
```
FelixMalfait, line 213: I guess this is temporary code and is already obvious for you, but this would need to be refactored into something more generic

packages/twenty-front/src/modules/object-record/record-show/components/RecordShowEffect.tsx
```diff
@@ -53,5 +57,15 @@ export const RecordShowEffect = ({
     }
   }, [record, setRecordStore, loading]);
 
+  useEffect(() => {
+    eventTracker(AnalyticsType.TRACK, {
```
FelixMalfait, line 61: We already track pageviews, can't we infer it from that or do something smarter rather than send 2 extremely similar track events? 
We could either stop tracking unstructured pageviews and only store events like that ; or we could eventually let the endpoint support an array and send 2 events at once...? I think I prefer solution 1

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/utils/wrapper-with-double-quote-when-uppercase.ts
```diff
@@ -0,0 +1,5 @@
+export const wrapperWithDoubleQuoteWhenUpperCase = (value: string) => {
```
FelixMalfait, line 1: Why don't we always add quotes? I know we don't add much comments in the codebase but maybe it's worth adding one here because it's not clear to me which we cannot always add quotes and get rid of this function entirely

Weiko:
Let's reopen when you are back! 🙂

PR #15083, opened 2025-10-14
Common Api - createOne/Many

### LINKED ISSUE #1578: Common - createMany
- Include field validation (In ORM ?) 
https://github.com/twentyhq/twenty/issues/14351#issuecomment-3266169877
https://discord.com/channels/1130383047699738754/1421058666705977436/1421066418102865930

etiennejouan:
closes https://github.com/twentyhq/core-team-issues/issues/1578


packages/twenty-server/src/engine/api/graphql/workspace-resolver-builder/factories/create-many-resolver.factory.ts
```diff
@@ -27,6 +38,51 @@ export class CreateManyResolverFactory
     const internalContext = context;
 
     return async (_source, args, _context, info) => {
+      const isCommonApiEnabled = await this.featureFlagService.isFeatureEnabled(
```
Weiko: Note: I wouldn't put all this stuff in the factory but in the service instead. A factory shouldn't have too many business logic imho, it's there to dynamically "create" objects and return them.

packages/twenty-server/src/engine/api/graphql/workspace-resolver-builder/factories/create-one-resolver.factory.ts
```diff
@@ -27,6 +37,49 @@ export class CreateOneResolverFactory
     const internalContext = context;
 
     return async (_source, args, _context, info) => {
+      const isCommonApiEnabled = await this.featureFlagService.isFeatureEnabled(
```
Weiko, line 40: Also this means here we are querying the cache on every request (which is better than calling the DB but could be avoided).
If you follow my comment above, you should have access to `executionArgs.workspaceDataSource.featureFlagMap` which will be cached locally (on the pod) instead

packages/twenty-server/src/engine/api/rest/core/handlers/rest-api-find-many.handler.ts
```diff
@@ -123,7 +123,7 @@ export class RestApiFindManyHandler extends RestApiBaseHandler {
         pageInfo,
       );
     } catch (error) {
-      workspaceQueryRunnerRestApiExceptionHandler(error);
+      return workspaceQueryRunnerRestApiExceptionHandler(error);
```
Weiko: I don't think you need to return here, this method should return void and only throw when needed

packages/twenty-server/src/engine/api/rest/core/services/rest-api-core.service.ts
```diff
@@ -39,11 +39,23 @@ export class RestApiCoreService {
   }
 
   async createOne(request: AuthenticatedRequest) {
-    return await this.restApiCreateOneHandler.handle(request);
+    const isCommonApiEnabled = await this.isCommonApiEnabled(request);
+
+    if (isCommonApiEnabled) {
+      return await this.restApiCreateOneHandler.commonHandle(request);
+    } else {
```
Weiko, line 46: nit: else is unnecessary since you return in your if, up to you though!

Weiko (commented):
Left some comments 🙏

Weiko (approved):
LGTM

packages/twenty-server/src/engine/api/graphql/workspace-resolver-builder/factories/create-many-resolver.factory.ts
etiennejouan: As discussed, we could use the base resolver service and then redirect if featureFlag (fetch from dataSource) to common. Not enough gain to do it.

PR #15086, opened 2025-10-14
Fix: Correct inverted logic in signUpWithoutWorkspace causing 'User already exists' error

FelixMalfait:
## Problem

When trying to signup on localhost:3001 with a new random email, users were getting an error message saying 'User already exists', even though they were new users.

## Root Cause

The `signUpWithoutWorkspace` method in `sign-in-up.service.ts` had inverted logic when checking if a user exists. It was using `findUserByEmailOrThrow` which:
- **Returns the user** if found
- **Throws the provided error** if NOT found

This caused the opposite behavior:
- ❌ **New user (doesn't exist)**: Threw 'User already exists' error
- ❌ **Existing user**: Continued to create duplicate user

## Solution

Changed to use `findUserByEmail` and explicitly check if the user exists before throwing the appropriate error:

```typescript
const existingUser = await this.userService.findUserByEmail(newUserParams.email);

if (existingUser) {
  throw new AuthException(
    'User already exist',
    AuthExceptionCode.USER_ALREADY_EXIST,
    { userFriendlyMessage: msg`User already exists` },
  );
}
```

This matches the correct pattern already used in `signUpInWorkspace` (line 431-441 in auth.resolver.ts).

## Changes
- Fixed inverted logic in `signUpWithoutWorkspace` method
- Now correctly validates that user does NOT exist before creating new user
- Matches the pattern used in `signUpInWorkspace`

## Testing
The fix corrects the logic so that:
- ✅ New users can sign up successfully
- ✅ Existing users get the correct 'User already exists' error


packages/twenty-server/src/engine/core-modules/auth/services/sign-in-up.service.ts
```diff
@@ -465,14 +465,17 @@ export class SignInUpService {
...
   ) {
-    await this.userService.findUserByEmailOrThrow(
+    const existingUser = await this.userService.findUserByEmail(
       newUserParams.email,
-      new AuthException(
+    );
+
+    if (existingUser) {
+      throw new AuthException(
         'User already exist',
```
FelixMalfait: Please rename to existS – also rename exception code

FelixMalfait:
✅ Addressed feedback:
- Renamed `existingUser` to `userExists`
- Renamed `AuthExceptionCode.USER_ALREADY_EXIST` to `USER_ALREADY_EXISTS` (plural)
- Updated error message to 'User already exists' (plural)
- Updated all references in auth exception handlers

Pushed in commit e454adb69f

charlesBochet (approved):

PR #15087, opened 2025-10-14
Simple design for empty node

thomtrp:
I did an acceptable design for empty node and iterators for release:
- use array field for iterator node. Added an util to stringify arrays for backward compatibily 
- remove icon for empty node
- allow to select a node on empty node selection

https://github.com/user-attachments/assets/b00037a8-aa1d-4784-b973-05973649b46e

### DIFF
--- a/packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/components/WorkflowEditActionEmpty.tsx
+++ b/packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/components/WorkflowEditActionEmpty.tsx
@@ -1,12 +1,8 @@
-import { SidePanelHeader } from '@/command-menu/components/SidePanelHeader';
+import { CommandMenuWorkflowSelectAction } from '@/command-menu/pages/workflow/action/components/CommandMenuWorkflowSelectAction';
+import { CommandMenuWorkflowEditStepType } from '@/command-menu/pages/workflow/step/edit/components/CommandMenuWorkflowEditStepType';
 import { type WorkflowEmptyAction } from '@/workflow/types/Workflow';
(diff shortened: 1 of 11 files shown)


packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/iterator-action/WorkflowEditActionIterator.tsx
```diff
@@ -89,15 +120,13 @@ export const WorkflowEditActionIterator = ({
         disabled={actionOptions.readonly}
       />
       <WorkflowStepBody>
-        <FormTextFieldInput
+        <FormArrayFieldInput
```
Devessier, line 123: Hell yeah!

packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/components/WorkflowEditActionEmpty.tsx
```diff
@@ -18,37 +14,11 @@ type WorkflowEditActionEmptyProps = {
...
   actionOptions,
 }: WorkflowEditActionEmptyProps) => {
-  const { getIcon } = useIcons();
-  const { headerTitle, headerIcon, headerIconColor, headerType } =
-    useWorkflowActionHeader({
-      action,
-      defaultTitle: 'Empty Node',
-    });
+  if (actionOptions.readonly === true) {
+    return <CommandMenuWorkflowSelectAction onActionSelected={() => {}} />;
```
Devessier, line 20: As mentioned by messages, I think we would like to display nothing when the user clicks on an empty action node. I don't think it's really meaningful to display the list of available actions in read-only mode.

Devessier (approved):
Great job!

PR #15044, opened 2025-10-12
feat: Attachement for Send Email workflow  node

harshit078:
## Description

-  This PR addresses the one issue out of https://github.com/twentyhq/core-team-issues/issues/1685
-  Added backend support for workflow node to support attachement
- updated send email schema, core utility 
- added workflowattachmentRow and workflowsendEmailAttachment file to handle file attachment in email workflow
- updated Google and Microsoft to use MailComposer which unifies with SMTP provider and improves mail structure

## Visual Appearance

https://github.com/user-attachments/assets/16478569-0a83-417e-a85e-70e41fe83343

PR #15089, opened 2025-10-14
Reuse OAuth access tokens

neo773:


packages/twenty-server/src/modules/connected-account/oauth2-client-manager/drivers/google/google-oauth2-client-manager.service.ts
```diff
@@ -25,7 +25,7 @@ export class GoogleOAuth2ClientManagerService {
       );
 
       oAuth2Client.setCredentials({
-        refresh_token: refreshToken,
+        access_token: accessToken,
       });
```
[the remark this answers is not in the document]
neo773: handled in `ConnectedAccountRefreshTokensService`

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
```diff
@@ -167,4 +168,24 @@ export class JwtWrapperService {
       return ExtractJwt.fromUrlQueryParameter('token')(request);
     };
   }
+
+  isTokenExpired(token: string, expirationBufferInSeconds = 5 * 60): boolean {
```
charlesBochet: let's extract this to an util, its nice to have this tooling but we shouldn't have to instantiate a service and to import the module when we want to use it (also easier to add unit test).

isTokenExpiredOrInvalid would be a more proper naming given the behavior with exceptions

charlesBochet (approved):

packages/twenty-server/src/modules/connected-account/refresh-tokens-manager/services/connected-account-refresh-tokens.service.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+  let microsoftAPIRefreshAccessTokenService: MicrosoftAPIRefreshAccessTokenService;
+  let twentyORMManager: TwentyORMManager;
+
+  const mockWorkspaceId = 'workspace-123';
+  const mockConnectedAccountId = 'account-456';
+  const mockAccessToken = 'valid-access-token';
+  const mockRefreshToken = 'valid-refresh-token';
+  const mockNewAccessToken = 'new-access-token';
+
+  beforeEach(async () => {
```
charlesBochet, line 31: to be honest, we should not invest on unit testing these services but rather on integration testing them. We should mock the low level microsoft/google clients so they return realistic API response and check that result is right in DB then

charlesBochet: (but that's fine for this PR, just a general note)

charlesBochet (approved):
It all makes sense to me, nice work!

packages/twenty-server/src/modules/messaging/message-import-manager/services/messaging-send-message.service.ts
```diff
@@ -41,11 +37,14 @@ export class MessagingSendMessageService {
...
-          await this.oAuth2ClientProvider.getOAuth2Client(connectedAccount);
+          await this.oAuth2ClientManagerService.getGoogleOAuth2Client(
+            connectedAccount,
+          );
+
+        const gmailClient = oAuth2Client.gmail({
+          version: 'v1',
+        });
 
         const { data } = await oAuth2Client.userinfo.get();
```
martmull, line 49: This does not work anymore, fixing that here https://github.com/twentyhq/twenty/pull/15044/commits/872692ec6ae76b0266300a3fafb4940e165a0aaa

PR #15095, opened 2025-10-14
Added dashboard chart advanced filter components

lucasbordeau:
This PR adds the components for advanced filters on chart settings.

It also refactors what's in common between this new command menu page and the workflow advanced filters.


lucasbordeau:
@thomtrp For information, it was really easy to refactor.

lucasbordeau:
@charlesBochet Everything in common is refactored in this PR, there's not much tech debt except some checks in the tree to know if we are in workflows or not, but that's acceptable.

thomtrp:
> @thomtrp For information, it was really easy to refactor.

@lucasbordeau I am a bit surprised. I thought you were against having common filters with workflows since that would require a special treatment. Happy to discuss it.

lucasbordeau:
@thomtrp In that case it's more about creating a common command menu filter module, because that is exactly the same except for a few places with variable picker and some logic for filtering, which is different than the advanced filter module in the view bar.

I think that there is still indeed something to find to avoid having the context of the workflow in this module, but we'll surely find a way.

But I still don't think we should merge it with the view bar advanced filter module.

ijreilly (approved):

PR #15099, opened 2025-10-14
Log invalid uuids

RaphiTobi:
I had an issue with invalid UUIDs, and I think it would be easier to find the offending one if the UUID were included in the error message.

This way a database dump can be easily searched.


prastoin (commented):
Hey thanks for your contribution
This will require a snapshot update of exceptions integrations tests + anon of the uuid in them
We might rather send a value in the exception payload to ease uuid matching

RaphiTobi:
> Hey thanks for your contribution This will require a snapshot update of exceptions integrations tests + anon of the uuid in them We might rather send a value in the exception payload to ease uuid matching

Hi thanks for your feedback, I dont think i am going to get into it this deep to figure out the sanpshot tests. Sending the value not as part of the Exeption message seams the best solution, unfortunatly as far as my research goes the toString of the exeption does not include anything besides the message and stacktrace.

As this is nothing that important I will probably just close this PR.

prastoin:
Hey @RaphiTobi, of course no worries
I agree the uuid should at least be available in the exception payload maybe not directly in the message though
Will try to take this at the latest tomorrow evening

prastoin:
@greptileai trigger

martmull (approved):
lgtm thank you

PR #15106, opened 2025-10-15
Updated the visuals to match the format of the other ones

StephanieJoly4:
- Removed a few visuals that are not necessary in the end (and not used yet)
- Replaced visuals that were added yesterday with a new version, matching the expected format


prastoin:
@greptileai trigger

prastoin (approved):
Hey there unsure to find the pages where these images are all displayed
Please double check that filename renaming does not introduce any regression in their rendering ( cannot find any references to the old ones anw so I guess it's dynamic somehow )

StephanieJoly4:
@prastoin yes those files are not used yet (I added them yesterday EOD in order to use them in the new version of the user guide (which is still WIP on another branch) so the new names won't be an issue, thank you for the review!

PR #15109, opened 2025-10-15
first step in the show record page with a morph relation.

### LINKED ISSUE #1324: [frontend] Morph on Show page
- [x] as show page (full page) and in the in the side panel : missing the Helper field
   
Image

guillim:
Adds the field on the Show page and side panel for Morph relation fields. 

Updates `packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/input/hooks/useOpenMorphRelationOneToManyFieldInput.tsx` to use the correct `recordPickerInstanceId` when opening the picker so the helper consistently renders (fixes [core-team-issues #1324](https://github.com/twentyhq/core-team-issues/issues/1324)).


charlesBochet:
Functional review:
- Add New button in ManyToOne morph dropdown on Show page (+ width is not large)
image

packages/twenty-front/src/modules/object-record/record-field-list/components/RecordFieldList.tsx
```diff
@@ -212,37 +214,80 @@ export const RecordFieldList = ({
...
-              objectMetadataItem,
-            }),
-            useUpdateRecord: useUpdateOneObjectRecordMutation,
-            isDisplayModeFixHeight: true,
-            isRecordFieldReadOnly: isRecordFieldReadOnly({
-              isRecordReadOnly,
-              objectPermissions: getObjectPermissionsFromMapByObjectMetadataId({
-                objectPermissionsByObjectMetadataId,
-                objectMetadataId: objectMetadataItem.id,
+      {boxedRelationFieldMetadataItems
```
charlesBochet, line 217: **change:** this part looks duplicated, maybe we should do the switch between RecordDetailRelationSection and RecordDetailMorphRelationSection lower?

packages/twenty-front/src/modules/object-record/record-field-list/record-detail-section/relation/components/RecordDetailMorphRelationRecordsList.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { RecordDetailRelationRecordsListItemEffect } from '@/object-record/record-field-list/record-detail-section/relation/components/RecordDetailRelationRecordsListItemEffect';
+import { type ObjectRecord } from '@/object-record/types/ObjectRecord';
+
+type RecordDetailRelationRecordsListProps = {
+  objectNameSingular: string;
+  value: ObjectRecord;
+  fieldMetadataId: string;
+};
+
+export const RecordDetailMorphRelationRecordsList = ({
```
charlesBochet: **change:** looks like this could be put in common between Relation and MorphRelation?

packages/twenty-front/src/modules/object-record/record-field-list/record-detail-section/relation/components/RecordDetailMorphRelationSection.tsx
```diff
@@ -0,0 +1,96 @@
...
+  loading,
+}: RecordDetailMorphRelationSectionProps) => {
+  const { recordId, fieldDefinition } = useContext(FieldContext);
+
+  const metadata = fieldDefinition.metadata as FieldMorphRelationMetadata;
+  const { objectMetadataNameSingular } = metadata;
+
+  const isMobile = useIsMobile();
+
+  const recordsWithObjectNameSingular =
```
charlesBochet, line 36: **change:** it seems that this is actually the only difference between relation and morph relations in the list part (if we say that relation list should also take an [record, objectNameSingular] array as parameter, the record list looks very similar

Then the other changes in the dropdown itself

charlesBochet: Edit: I think that we should:
- keep separated: RecordDetailMorphRelationSection vs RecordDetailRelationSection
- keep separated for dropdown too
- merge RecordDetailRelationRecordsList and RecordDetailMorphRelationRecordsList (let's make them take [record + objectNameSingular] array as props even for relation case which is a subcase of morph (like we have done in settings)

packages/twenty-front/src/modules/object-record/record-field-list/record-detail-section/relation/components/RecordDetailMorphRelationSectionDropdown.tsx
```diff
@@ -0,0 +1,67 @@
...
+  loading: boolean;
+};
+
+export const RecordDetailMorphRelationSectionDropdown = ({
+  loading,
+}: RecordDetailMorphRelationSectionDropdownProps) => {
+  const { fieldDefinition, isRecordFieldReadOnly, recordId } =
+    useContext(FieldContext);
+  const { relationType, morphRelations } =
+    fieldDefinition.metadata as FieldMorphRelationMetadata;
```
charlesBochet: question: we don't have a type guard for that (instead of using as)

charlesBochet (commented):
@guillim Overall looks good to me! I think that the display of the recordlist or record in the section could be factorized between morph and relations. (the inputs, dropdown and parent section container cannot without too much complexity, that's the right balance)

Unrelated note for future: seing your PR made me think that we could actually deprecate relations when we are done with morph relations (relations are a subcase of morph)

packages/twenty-front/src/modules/object-record/record-field-list/record-detail-section/relation/components/RecordDetailMorphRelationRecordsList.tsx
charlesBochet: see comment below

packages/twenty-front/src/modules/object-record/record-field-list/record-detail-section/relation/components/RecordDetailMorphRelationSectionDropdown.tsx
guillim: we do you are correct. Fixing it

packages/twenty-front/src/modules/object-record/record-field-list/record-detail-section/relation/components/RecordDetailMorphRelationRecordsList.tsx
guillim: on it

packages/twenty-front/src/modules/object-record/record-field-list/record-detail-section/relation/components/RecordDetailMorphRelationSection.tsx
guillim: totally agreed.

packages/twenty-front/src/modules/object-record/record-field-list/components/RecordFieldList.tsx
guillim: you are correct. on it

guillim:
> Functional review:
> 
> * Add New button in ManyToOne morph dropdown on Show page (+ width is not large)
> 
> image

My guess: you are on a normal RELATION but on the other side it is a morph. Am I right ?
If yes,  the width is not a problem but the issue was the display of the ` . ObejctName` next to the record. this was fixed yesterday.
The Add new however should be fixed. I will fix it

packages/twenty-front/src/modules/object-record/record-field-list/record-detail-section/relation/components/RecordDetailRelationRecordsList.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { RecordDetailRelationRecordsListItemEffect } from '@/object-record/record-field-list/record-detail-section/relation/components/RecordDetailRelationRecordsListItemEffect';
+import { type ObjectRecord } from '@/object-record/types/ObjectRecord';
+
+type RecordDetailRelationRecordsListProps = {
+  objectNameSingular: string;
+  value: ObjectRecord;
+  fieldMetadataId: string;
+};
+
+export const RecordDetailRelationRecordsList = ({
```
charlesBochet, line 14: rename filename: RecordDetailRelationRecordsList

packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/input/hooks/useAddNewRecordAndOpenRightDrawer.ts
```diff
@@ -8,21 +8,26 @@ import { type FieldMetadataItem } from '@/object-metadata/types/FieldMetadataIte
...
+
+import { useUpdateOneRecordV2 } from '@/object-record/hooks/useUpdateOneRecordV2';
 import { viewableRecordIdState } from '@/object-record/record-right-drawer/states/viewableRecordIdState';
 import { viewableRecordNameSingularState } from '@/object-record/record-right-drawer/states/viewableRecordNameSingularState';
 import { getOperationName } from '@apollo/client/utilities';
-import { isDefined } from 'twenty-shared/utils';
+import { computeMorphRelationFieldName, isDefined } from 'twenty-shared/utils';
 import { FieldMetadataType, RelationType } from '~/generated-metadata/graphql';
 
 type RecordDetailRelationSectionProps = {
```
charlesBochet: rename props name: useAddNewRecordAndOpenRightDrawerProps

packages/twenty-front/src/modules/object-record/record-field/ui/hooks/usePersistField.ts
```diff
@@ -190,10 +190,28 @@ export const usePersistField = ({
...
+
+            upsertRecordsInStore([
+              getRecordFromRecordNode({
+                recordNode: newRecord,
+              }),
+            ]);
+            return;
+          }
+
+          if (fieldIsMorphRelationManyToOne) {
```
charlesBochet, line 213: I thought we said that this was not working?

charlesBochet:
Also check that we skip aggregateCall to count morph relation records on show page

charlesBochet:
+ On Relation side, when detaching a morph (for instance from company show page, this should update opportunity.oppOwnerCompanyId, it seems that it tries to update opportunity.oppOwnerId)

This is line 164 of `RecordDetailRelationRecordsListItem` file (we should take both case into account: opposite is morph or relation)

charlesBochet (approved):

packages/twenty-front/src/modules/object-record/record-field/ui/hooks/usePersistField.ts
guillim: well after reviewing the code, sounds like we still need it unless we adjust a bit our code.

The dependency comes from `usePersistFieldFromFieldInputContext` that is sometimes called within a MORPH fieldDefinition context. As we can see in the `RecordTableCellFieldInput` for instance.

If we really want to get rid of the morph part here, we could add an if else on fieldDefinition from `RecordTableCellFieldInput` to choose the appropriate persit morph when suitable.

I am open to this change if you believe it's worth it

PR #15111, opened 2025-10-15
Fix infinite loop on new table

charlesBochet:
https://github.com/user-attachments/assets/7002cd6c-08d9-4ad7-833f-21bc23d5eea2


ijreilly:
freeze seems fixes, but when I created a new view it created two of them, see first 10 seconds of video:

https://github.com/user-attachments/assets/93f40574-16b7-473c-8ca0-f28f191b64fc

ijreilly (approved):

PR #15117, opened 2025-10-15
chore(security): force happy-dom@20.0.0 via Yarn resolutions to patch RCE vulnerability.

mabdullahabaid:
Fixes [Dependabot Alert 290](https://github.com/twentyhq/twenty/security/dependabot/290) — Remote Code Execution vulnerability in happy-dom <20.0.0.

- Added Yarn "resolutions" entry for happy-dom@20.0.0 to override transitive dependency in @wyw-in-js/transform.
- Verified that all affected packages (@wyw-in-js/vite and @wyw-in-js/transform) now resolve to happy-dom@20.0.0.
- Ran full build, tests, and manual frontend checks - nothing breaks per my analysis.
- Confirmed happy-dom is used only during build-time DOM emulation and not in runtime bundles.

### DIFF
--- a/package.json
+++ b/package.json
@@ -224,7 +224,8 @@
     "typescript": "5.9.2",
     "graphql-redis-subscriptions/ioredis": "^5.6.0",
     "prosemirror-view": "1.40.0",
-    "prosemirror-transform": "1.10.4"
+    "prosemirror-transform": "1.10.4",
+    "happy-dom": "20.0.0"
   },
   "version": "0.2.1",
   "nx": {},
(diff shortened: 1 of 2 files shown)


mabdullahabaid:
Fixes [Dependabot Alert 292](https://github.com/twentyhq/twenty/security/dependabot/292) too - a follow up alert asking to upgrade happy-dom from 20.0.0 to 20.0.2 instead.

FelixMalfait:
Thanks! (Sent feedback on Discord, I don't think it's the right solution)

PR #15118, opened 2025-10-15
[Dashboards] - Min Max range on secondary axis bar charts

ehconitin:
video QA

https://github.com/user-attachments/assets/70c37188-2398-43de-bbf6-5882bb79940a


ehconitin:
@greptileai trigger

charlesBochet (commented):
I don't think we should add rangeMin, rangeMax or omitNull in the backend

If we do this, we should implement a proper "having" API (that would specify on which aggregate it applies). This is not a must + a bit expensive

Let's do these checks on FE side, there is no pagination issue at this point it should be fine for 99% of our cases

packages/twenty-front/src/modules/command-menu/components/CommandMenuItemNumberInput.tsx
```diff
@@ -0,0 +1,95 @@
...
+import styled from '@emotion/styled';
+import { useState } from 'react';
+import { Key } from 'ts-key-enum';
+import { isDefined } from 'twenty-shared/utils';
+import {
+  canBeCastAsNumberOrNull,
+  castAsNumberOrNull,
+} from '~/utils/cast-as-number-or-null';
+
+const ERROR_STATE_WITHOUT_MESSAGE = ' ';
```
charlesBochet: question: I'm a bit confused by this, why is it needed?

ehconitin: Its kind of a workaround to show the red error border but not the message!

charlesBochet: if we want to display the field in red with no error message we should pass empty string?

ehconitin: We could have the error written, but i chose not to

ehconitin: It is an empty string no?
Just wanted to be explicit that its empty :)

packages/twenty-front/src/modules/command-menu/pages/page-layout/components/ChartSettings.tsx
charlesBochet: not directly related to this PR but I feel this ChartSettings are starting to be too complex (it's hard to read).

We should either:
- introduce a ChartBarSettings, ChartXXXSettings...
- introduce an util isCommandMenuChartDisabled(chartType, itemId) 

I'm not a fan on this dependsOn logic which is hard to troubleshoot. The only reason to have such a complex logic is if this aims to be abstracted in the settings (if widget were dynamic, which I don't think is the plan, all widgets will be hard coded in the code and so can their settings be)

charlesBochet: all these if is a sign that compleixity is exploding. We should prefer cut the compleixty higher in the hiearchy (introduce a Setting per chart) or abstract it to an util **IF** the abstraction is consistent across charts (we don't want logic edge cases to be too complex)

packages/twenty-front/src/modules/command-menu/pages/page-layout/components/ChartSettings.tsx
```diff
@@ -148,6 +150,56 @@ export const ChartSettings = ({ widget }: { widget: PageLayoutWidget }) => {
               });
             };
 
+            const createRangeValidator = (itemId: string) => {
```
charlesBochet: change: this is too complex, why do we need a function that returns a function, just call the function with itemId.

packages/twenty-front/src/modules/command-menu/pages/page-layout/components/ChartSettings.tsx
```diff
@@ -148,6 +150,56 @@ export const ChartSettings = ({ widget }: { widget: PageLayoutWidget }) => {
...
+                        itemId as keyof typeof CHART_CONFIGURATION_SETTING_TO_CONFIG_KEY_MAP
+                      ]
+                    : itemId;
+
+                if (configKey === 'rangeMin') {
+                  if (
+                    'rangeMax' in configuration &&
+                    isDefined(configuration.rangeMax) &&
+                    isDefined(value) &&
+                    value > configuration.rangeMax
```
charlesBochet: question: I'm confused:
- value is the value typed by the user?
- what is configuration.value?

why are we comparing both?

packages/twenty-front/src/modules/command-menu/pages/page-layout/components/ChartSettings.tsx
```diff
@@ -166,6 +218,36 @@ export const ChartSettings = ({ widget }: { widget: PageLayoutWidget }) => {
               );
             }
 
+            if (item.isInput === true) {
```
charlesBochet: === true not needed

packages/twenty-front/src/modules/command-menu/pages/page-layout/components/ChartSettings.tsx
```diff
@@ -166,6 +218,36 @@ export const ChartSettings = ({ widget }: { widget: PageLayoutWidget }) => {
...
+              const settingValue = getChartSettingsValues(item.id);
+              const stringValue = isString(settingValue) ? settingValue : '';
+
+              const placeholder =
+                item.id === CHART_CONFIGURATION_SETTING_IDS.MIN_RANGE
+                  ? t`Min`
+                  : item.id === CHART_CONFIGURATION_SETTING_IDS.MAX_RANGE
+                    ? t`Max`
+                    : t`Auto`;
```
charlesBochet: not a fan, what is auto corresponding to?

We should either:
- make item "type" configurable: isInput: boolean => type: MinRange | MaxRange
- make placeholder configurable on itme: isInput + inputPlacholder

packages/twenty-front/src/modules/command-menu/pages/page-layout/constants/settings/RangeMaxSetting.ts
```diff
@@ -0,0 +1,12 @@
+import { CHART_CONFIGURATION_SETTING_LABELS } from '@/command-menu/pages/page-layout/constants/settings/ChartConfigurationSettingLabels';
+import { CHART_CONFIGURATION_SETTING_IDS } from '@/command-menu/pages/page-layout/types/ChartConfigurationSettingIds';
+import { type ChartSettingsItem } from '@/command-menu/pages/page-layout/types/ChartSettingsGroup';
+import { IconMathMax } from 'twenty-ui/display';
+
+export const RANGE_MAX_SETTING: ChartSettingsItem = {
+  isBoolean: false,
+  Icon: IconMathMax,
+  label: CHART_CONFIGURATION_SETTING_LABELS.MAX_RANGE,
```
charlesBochet, line 10: let's put the placeholder here for now

packages/twenty-front/src/modules/command-menu/components/CommandMenuItemNumberInput.tsx
```diff
@@ -0,0 +1,95 @@
...
+const ERROR_STATE_WITHOUT_MESSAGE = '';
+
+type CommandMenuItemNumberInputProps = {
+  value: string;
+  onChange: (value: number | null) => void;
+  onValidate?: (value: number | null) => boolean;
+  placeholder?: string;
+};
+const StyledRightAlignedTextInput = styled(TextInput)`
+  input {
```
charlesBochet, line 18: this should be reviewed with design to make the TextInput component evolve rather than overriding here

packages/twenty-front/src/modules/command-menu/components/CommandMenuItemNumberInput.tsx
```diff
@@ -0,0 +1,95 @@
...
+import styled from '@emotion/styled';
+import { useState } from 'react';
+import { Key } from 'ts-key-enum';
+import { isDefined } from 'twenty-shared/utils';
+import {
+  canBeCastAsNumberOrNull,
+  castAsNumberOrNull,
+} from '~/utils/cast-as-number-or-null';
+
+const ERROR_STATE_WITHOUT_MESSAGE = '';
```
charlesBochet: does not seems useful

packages/twenty-front/src/modules/command-menu/components/CommandMenuItemNumberInput.tsx
```diff
@@ -0,0 +1,95 @@
...
+    <StyledRightAlignedTextInput
+      value={draftValue}
+      sizeVariant="sm"
+      onChange={handleChange}
+      onFocus={handleFocus}
+      onBlur={handleBlur}
+      onKeyDown={handleKeyDown}
+      placeholder={placeholder}
+      error={hasError ? ERROR_STATE_WITHOUT_MESSAGE : undefined}
+      noErrorHelper
```
charlesBochet, line 89: if we add noErrorHelper, do we need the error message?

packages/twenty-front/src/modules/command-menu/pages/page-layout/utils/getConfigKeyFromSettingId.ts
```diff
@@ -0,0 +1,14 @@
+import {
+  type CHART_CONFIGURATION_SETTING_IDS,
+  CHART_CONFIGURATION_SETTING_TO_CONFIG_KEY_MAP,
+} from '@/command-menu/pages/page-layout/types/ChartConfigurationSettingIds';
+
+export const getConfigKeyFromSettingId = (
+  settingId: CHART_CONFIGURATION_SETTING_IDS,
+): string => {
+  return settingId in CHART_CONFIGURATION_SETTING_TO_CONFIG_KEY_MAP
```
charlesBochet, line 9: I'm not a fan of this "ConfigKey", this is hard to understand thing but can be reworked later
looks more that we need to map it to a GQLField (might only be naming)

packages/twenty-front/src/modules/command-menu/pages/page-layout/utils/isChartSettingDisabled.ts
```diff
@@ -0,0 +1,22 @@
...
+import { type ChartSettingsItem } from '@/command-menu/pages/page-layout/types/ChartSettingsGroup';
+import { isNonEmptyString } from '@sniptt/guards';
+
+export const isChartSettingDisabled = (
+  item: ChartSettingsItem,
+  objectMetadataId: string,
+  isGroupByEnabled: boolean,
+): boolean => {
+  const hasNoObjectMetadata = !isNonEmptyString(objectMetadataId);
+  const dependsOnSource = item?.dependsOn?.includes(
```
charlesBochet, line 11: this dependsOn might be a bit overengineered but if it works we can keep it.

I would have a callback by Setting that could check other settings porperty (a bit like we do for actions)

packages/twenty-front/src/modules/command-menu/pages/page-layout/utils/validateRangeValue.ts
```diff
@@ -0,0 +1,36 @@
...
+  configuration: ChartConfiguration,
+): boolean => {
+  const configKey = getConfigKeyFromSettingId(settingId);
+
+  if (configKey === 'rangeMin') {
+    if (
+      'rangeMax' in configuration &&
+      isDefined(configuration.rangeMax) &&
+      isDefined(newValue) &&
+      newValue > configuration.rangeMax
```
charlesBochet: I'm a bit confused, we want the minValue to be bigger than the max?

packages/twenty-front/src/modules/page-layout/widgets/graph/utils/filterGroupByResults.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    if (omitNullValues === true && !isDefined(aggregateValue)) {
+      return false;
+    }
+
+    if (omitNullValues === true && aggregateValue === 0) {
+      return false;
+    }
+
+    if (typeof aggregateValue === 'number') {
```
charlesBochet, line 57: we should not use typeOf but rely on aggregateField type instead and trust it

packages/twenty-front/src/modules/page-layout/widgets/graph/utils/transformGroupByDataToBarChartData.ts
```diff
@@ -105,6 +107,20 @@ export const transformGroupByDataToBarChartData = ({
...
+    rawResults,
+    filterOptions: {
+      rangeMin: configuration.rangeMin ?? undefined,
+      rangeMax: configuration.rangeMax ?? undefined,
+      omitNullValues: configuration.omitNullValues ?? false,
+    },
+    aggregateField,
+    aggregateOperation:
+      configuration.aggregateOperation as unknown as ExtendedAggregateOperations,
+    aggregateOperationFromRawResult: aggregateOperation,
```
charlesBochet, line 120: I'm a bit confused by this

charlesBochet (approved):

PR #15129, opened 2025-10-16
refactor: avoid uuid package usage in frontend application

AnastasiyaHladina:
This PR removes the uuid package usage from the frontend application and replaces it with standard randomUUID usage from [Web crypto](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID)


FelixMalfait:
Hello. Thanks for this! This would break for non-https (or  non - *.locahost) environments, I don't know if it's a tradeoff we're okay to accept. Removing dependencies is always nice so it's tempting.

AnastasiyaHladina:
Hello @FelixMalfait.
Not sure how it is related to non-https envronments, crypto is available in browser API.

charlesBochet (changes requested):
@AnastasiyaHladina thanks for opening this PR. I'm not against it but this will break on older browsers.

Could you introduce a function in twenty-shared instead and use it in twenty-front then:
```
export function generateV4Uuid() {
  if (typeof crypto !== 'undefined' && crypto.randomUUID) {
    return crypto.randomUUID();
  }
  return uuidv4();
}
```

FelixMalfait:
@AnastasiyaHladina no crypto is not available on non-https, except for a few domains (localhost). So if I create a custom domain locally like http://testing.dev then the application will not work with crypto. It's the same for the copy/paste API, we've had users complained it doesn't work and the reason was not HTTPS

charlesBochet:
I'm not sure it worth it TBH

charlesBochet:
> @AnastasiyaHladina thanks for opening this PR. I'm not against it but this will break on older browsers.
> 
> Could you introduce a function in twenty-shared instead and use it in twenty-front then:
> 
> ```
> export function generateV4Uuid() {
>   if (typeof crypto !== 'undefined' && crypto.randomUUID) {
>     return crypto.randomUUID();
>   }
>   return uuidv4();
> }
> ```

Forget this comment, this is actually what the uuid package is doing:
https://github.com/uuidjs/uuid/blob/main/src/v4.ts

FelixMalfait:
Thanks for the suggestion! Let's keep this dependency for now

charlesBochet:
@AnastasiyaHladina We have discussed it internally and we think it's not worth it:
- we would need to polyfill anyway (for non HTTPs context and also for older browsers)
- this polyfill is actually already in place in uuid package, so let's keep it as it is

PR #15132, opened 2025-10-16
Companies show page as record page

Devessier:
Example showing two widgets positioned in a vertical list:

https://github.com/user-attachments/assets/cf3e9793-7d80-4cab-b708-aad9f2bca7d8


charlesBochet:
@Devessier let me know when this is ready for review

Devessier:
It is, @charlesBochet!

packages/twenty-front/src/generated-metadata/graphql.ts
```diff
@@ -1250,6 +1250,7 @@ export enum FeatureFlagKey {
   IS_PAGE_LAYOUT_ENABLED = 'IS_PAGE_LAYOUT_ENABLED',
   IS_POSTGRESQL_INTEGRATION_ENABLED = 'IS_POSTGRESQL_INTEGRATION_ENABLED',
   IS_PUBLIC_DOMAIN_ENABLED = 'IS_PUBLIC_DOMAIN_ENABLED',
+  IS_RECORD_PAGE_ENABLED = 'IS_RECORD_PAGE_ENABLED',
```
charlesBochet: feature flag name is not super clear? IS_RECORD_PAGE_LAYOUT_ENABLED?

packages/twenty-front/src/modules/page-layout/types/PageLayoutTab.ts
```diff
@@ -5,4 +5,8 @@ import {
 
 export type PageLayoutTab = Omit<PageLayoutTabGenerated, 'widgets'> & {
   widgets: PageLayoutWidget[];
+  /**
+   * Only available behind IS_RECORD_PAGE_ENABLED for now.
+   */
+  layoutMode?: 'grid' | 'vertical-list';
```
charlesBochet, line 11: **question:** I'm not sure about this: 
- why not make the widget full width? (taking max columns)

packages/twenty-front/src/modules/page-layout/widgets/fields/components/FieldsWidget.tsx
```diff
@@ -0,0 +1,31 @@
...
+type FieldsWidgetProps = {
+  widget: PageLayoutWidget;
+};
+
+export const FieldsWidget = ({ widget }: FieldsWidgetProps) => {
+  const targetRecord = useTargetRecord();
+  const { isInRightDrawer } = useLayoutRenderingContext();
+
+  // Parse configuration from widget
+  const configuration = widget.configuration as FieldCardConfiguration | null;
```
charlesBochet: question: we don't have runtime typing on this part?

Let's avoid "as"

packages/twenty-front/src/modules/page-layout/widgets/fields/components/FieldsWidget.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  return (
+    <RightDrawerProvider value={{ isInRightDrawer }}>
+      <div style={{ display: 'flex', flexDirection: 'column' }}>
```
charlesBochet: non standard

packages/twenty-front/src/modules/page-layout/components/PageLayoutGridLayout.tsx
```diff
@@ -76,20 +88,38 @@ export const PageLayoutGridLayout = () => {
...
 
   const isLayoutEmpty =
     !isDefined(activeTabWidgets) || activeTabWidgets.length === 0;
 
   const layouts = isLayoutEmpty
     ? EMPTY_LAYOUT
-    : pageLayoutCurrentLayouts[activeTabId] || EMPTY_LAYOUT;
+    : (pageLayoutCurrentLayouts[activeTabId] ?? EMPTY_LAYOUT);
+
+  const Widgets = isLayoutEmpty ? (
```
charlesBochet, line 112: polish: I'm never a big fan of declaring components within components, I feel it hurts the devXP: we have to scroll-down and up the file: if it worth introducing another component I would put it in a separate file. Here I think we can just let it in the react JSX below

charlesBochet: oh got you, it's because you are re-using it twice, it's a bit wrong to me, if they are the same why not handling the empty case earlier:

if (empty) { return EmptyPlaceholder }
if (!empty) { case verticalLayout / case gridLayout }

That's nitpick, LGTM

packages/twenty-front/src/modules/page-layout/components/PageLayoutGridLayout.tsx
```diff
@@ -102,43 +132,42 @@ export const PageLayoutGridLayout = () => {
...
+            resizeHandle={
+              isPageLayoutInEditMode ? (
+                <PageLayoutGridResizeHandle />
+              ) : undefined
+            }
+            resizeHandles={['se']}
+            onLayoutChange={handleLayoutChange}
+            onBreakpointChange={(newBreakpoint) =>
+              setPageLayoutCurrentBreakpoint(
+                newBreakpoint as PageLayoutBreakpoint,
```
charlesBochet, line 164: let's avoid as

packages/twenty-front/src/modules/page-layout/components/PageLayoutInitializationQueryEffect.tsx
```diff
@@ -24,15 +26,20 @@ export const PageLayoutInitializationQueryEffect = ({
...
       id: pageLayoutId,
     },
+    skip: isDefaultLayout,
   });
 
-  const pageLayout: PageLayout | undefined = data?.getPageLayout
-    ? transformPageLayout(data.getPageLayout)
-    : undefined;
+  const pageLayout: PageLayout | undefined = isDefaultLayout
+    ? DEFAULT_PAGE_LAYOUT
```
charlesBochet, line 39: we should migrate to backend asap :)

packages/twenty-front/src/modules/page-layout/types/PageLayoutTab.ts
Devessier: It doesn't seem the library supports rendering items with variable heights. In this case, we want elements to take all the height they need in one column.

We plan to still use the library for editing mode, but the most straightforward solution we found is to ditch the library for a vertical flex container.

packages/twenty-front/src/modules/page-layout/components/PageLayoutGridLayout.tsx
Devessier: I'll find a better DX later in the process 👍

packages/twenty-front/src/modules/page-layout/widgets/fields/components/FieldsWidget.tsx
Devessier: I'm removing the line of code because I don't want to manage the new widgets' configuration now.

FelixMalfait (approved):

charlesBochet (approved):

PR #15139, opened 2025-10-16
Reset password - Stuck on password resetting modal after validating 'change password'

vinay-kumar-3574:
I updated pages/auth/PasswordReset.tsx to handle redirection more robustly after a successful password change and implicit sign-in:

### DIFF
--- a/packages/twenty-front/src/pages/auth/PasswordReset.tsx
+++ b/packages/twenty-front/src/pages/auth/PasswordReset.tsx
@@ -4,11 +4,15 @@ import { Title } from '@/auth/components/Title';
 import { useAuth } from '@/auth/hooks/useAuth';
 import { useIsLogged } from '@/auth/hooks/useIsLogged';
 import { workspacePublicDataState } from '@/auth/states/workspacePublicDataState';
+import { countAvailableWorkspaces, getFirstAvailableWorkspaces } from '@/auth/utils/availableWorkspacesUtils';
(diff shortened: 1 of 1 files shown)


packages/twenty-front/src/pages/auth/PasswordReset.tsx
```diff
@@ -158,6 +166,22 @@ export const PasswordReset = () => {
...
+          const targetWorkspace = getFirstAvailableWorkspaces(user.availableWorkspaces);
+          await redirectToWorkspaceDomain(
+            getWorkspaceUrl(targetWorkspace.workspaceUrls),
+            AppPath.Index,
+          );
+          return;
+        }
+      } catch (e) {
+        // Fallback to local navigation if loading current user fails
+      }
```
etiennejouan, line 184: - Currently, resetting password is only available from workspace domain sign in. So user is already located on his workspace we want to redirect him on.
I think you can simply use the `useRedirect` hook here.

- Please, run the `npx nx lint twenty-server --fix` command before pushing (you can see that ci has failed)

- We try as much as possible to avoid comments in the codebase.

etiennejouan (changes requested):
Thank you for your contribution, I left some comments.

etiennejouan:
Hey @vinay-kumar-3574, thank you for your contribution. I close because of the requested changes and inactivity on this PR. Please feel free to take other issues.

PR #15143, opened 2025-10-16
1635 extensibilitytwenty cli app vars

### LINKED ISSUE #1621: [Extensibility] [Serverless] Ability to define env variables for serverless functions
As title

Solutions:
- create a command to add a core entity KeyValuePair to store an env variable for an applicationId
- create a new core entity envVariable dedicated to store env variables for a serverlessFunction or an application
- inject them during serverless execution

### LINKED ISSUE #1635: [Extensibility][twenty-cli] app vars
Do not use keyValuePair and create a new dedicated table "ApplicationVars"

martmull:
- Update twenty-cli to support application env variable definition
- Update twenty-server to create a new `core.applicationVariable` entity to store env variables and provide env var when executing serverless function 
- Update twenty-front to support application environment variable value setting

image

image

image

image

image

image


packages/twenty-server/src/engine/core-modules/applicationVariable/dtos/application-variable.dto.ts
```diff
@@ -0,0 +1,22 @@
...
+  @Field()
+  value: string;
+
+  @IsString()
+  @Field()
+  description: string;
+
+  @IsBoolean()
+  @Field()
+  isSecret: boolean;
```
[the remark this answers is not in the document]
martmull: in another PR

packages/twenty-front/src/modules/settings/serverless-functions/components/SettingsServerlessFunctionTabEnvironmentVariablesSection.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { LinkChip } from 'twenty-ui/components';
+import { getSettingsPath } from 'twenty-shared/utils';
+import { useParams } from 'react-router-dom';
+
+export const SettingsServerlessFunctionTabEnvironmentVariablesSection = () => {
+  const { applicationId = '' } = useParams<{ applicationId: string }>();
+  return (
+    <Section>
+      <H2Title
+        title="Environment Variables"
```
Weiko: use translation

packages/twenty-server/src/engine/core-modules/applicationVariable/application-variable.entity.ts
```diff
@@ -0,0 +1,65 @@
...
+  id: string;
+
+  @Column({ nullable: false, type: 'text' })
+  key: string;
+
+  @Column({ nullable: false, type: 'text', default: '' })
+  value: string;
+
+  @Column({ nullable: true, type: 'text' })
+  description?: string;
```
Weiko: Ideally nothing should be undefined in the returned entity. Can be changed later once we implement FlatApplicationVariable though

packages/twenty-server/src/engine/core-modules/applicationVariable/application-variable.module.ts
```diff
@@ -0,0 +1,23 @@
...
+import { NestjsQueryTypeOrmModule } from '@ptc-org/nestjs-query-typeorm';
+
+import { ApplicationVariable } from 'src/engine/core-modules/applicationVariable/application-variable.entity';
+import { TypeORMModule } from 'src/database/typeorm/typeorm.module';
+import { ApplicationVariableService } from 'src/engine/core-modules/applicationVariable/application-variable.service';
+import { ApplicationVariableResolver } from 'src/engine/core-modules/applicationVariable/application-variable.resolver';
+
+@Module({
+  imports: [
+    NestjsQueryGraphQLModule.forFeature({
```
Weiko: no need for nestjsquery module here since you are creating your own resolver and service. You can simply import the tyeprom module to retrieve your entity repository

packages/twenty-server/src/engine/core-modules/applicationVariable/application-variable.service.ts
```diff
@@ -0,0 +1,84 @@
...
+  async update({
+    key,
+    value,
+    applicationId,
+  }: Pick<ApplicationVariable, 'key' | 'value'> & { applicationId: string }) {
+    const { id } = await this.applicationVariableRepository.findOneOrFail({
+      where: { key, applicationId },
+    });
+
+    await this.applicationVariableRepository.update(id, {
```
Weiko: update can directly take a FindOption OR criteria I believe (your where above) instead of an id, this would avoid the extra SELECT query. Unless you really want to check earlier if the application exists which is fine but you'll need to properly throw a APPLICATION_VARIABLE_NOT_FOUND otherwise findOneOrFail will throw a typeorm error that will be mapped to a 500

packages/twenty-server/src/engine/core-modules/applicationVariable/application-variable.service.ts
```diff
@@ -0,0 +1,84 @@
...
+      string,
+      {
+        value?: string;
+        description?: string;
+        isSecret: boolean;
+      }
+    >;
+    applicationId: string;
+  }) {
+    const existingVariables = await this.applicationVariableRepository.find({
```
Weiko: note: typeorm has an upsert method which should do a `INSERT ... ON CONFLICT DO UPDATE`, might be better to use if it fits your case?

packages/twenty-server/src/engine/core-modules/applicationVariable/dtos/application-variable.dto.ts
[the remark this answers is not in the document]
Weiko: Ok with that, we can hide the secret value in a future PR indeed 👍

Weiko (approved):
Great! LGTM (I've left some comments though @martmull)
Note: ApplicationVariable should be a syncable entity in the future, can be done later though

martmull:
> ApplicationVariable should be a syncable entity in the future, can be done later though

Are you sure we want that? I think maintaining a universalId in the manifest for each env is a pain
Do you think we can create a universalId from KEY and application.universalId ?
@Weiko

PR #15145, opened 2025-10-16
[command fix] Limit migration to workflow versions

ijreilly:
This migration updates the filter operand values of the workflowVersion and workflowRuns in order to capitalize them as they should be (the product works with both deprecated camel case and capitalized). But that may involve thousands of workflowRuns! 
Let's update the command to only update workflowVersion, and update the cleanWorkflowRuns job to remove workflow runs that are more than 14 days old. 

This way after the command is run, all new workflow runs will have the new value for the filter operand, and after fourteen days there will be no trace of the workflow runs with the deprecated filter operand.


thomtrp (approved):
LGTM

PR #15146, opened 2025-10-16
Iterators design updates

thomtrp:
A few fixes :
- add a counter of iteration for nodes that have been run
Capture d’écran 2025-10-16 à 17 50 07

- if a node has fail, we should end the workflow even if a step is still running
- fix position for add node button in iterator loop :
Before

https://github.com/user-attachments/assets/eafeda64-977f-415d-bab4-b7d0c44406b0

After

https://github.com/user-attachments/assets/6ff221a3-b287-45f6-977f-b127f9998692


Devessier (approved):
Excellent improvements! Thanks.

PR #15103, opened 2025-10-15
implement "acting on behalf of user" for workflows and agents

abdulrahmancodes:
## Summary
**Step 1 of 2:** Implements the "acting on behalf of user" concept for workflows and agents to prevent permission escalation and maintain proper audit trails.

## Problem
Previously, workflows and agents would bypass permissions regardless of who initiated them, allowing users to escalate their privileges by triggering workflows that performed actions they couldn't do directly.

### For Workflows
Introduced `WorkflowExecutionContext` service that determines execution mode:
- **Manual triggers/test button**: Uses user's roleId for permissions, user's identity for `createdBy`
- **Automated triggers** (cron, database events, webhooks): Bypasses permissions, uses workflow identity

### For Agents
**In Chat:**
- Always act on behalf of the user
- Use user's roleId for permission checks
- Use user's identity for `createdBy`

# Step 1 vs Step 2

### ✅ Step 1 (This PR): Acting on Behalf Concept
- Introduced `isActingOnBehalfOfUser` boolean concept
- Single roleId used for permission checks (user's OR system bypass)
- `createdBy` field properly attributes actions to initiator
- Prevents permission escalation in user-initiated flows

### 🔜 Step 2 (Future): Multi-Role Permission Support
- Support role intersection: `{ intersection: ['roleA', 'roleB'] }`
- Support role union: `{ union: ['roleA', 'roleB', 'roleC'] }`
- Enable user+agent collaboration scenarios
- Update `WorkspaceEntityManager` and `WorkspaceDatasource` to handle multiple roleIds

PR #15150, opened 2025-10-16
feat: multi role permission intersection

### LINKED ISSUE #1661: Add custom field from option menu
# Current behavior:

The "field" menu item from the option menu shows both displayed columns, and hidden columns

Screenshot 2023-09-19 at 12 12 43

# Desired behavior

The "field" menu item from the option menu should only display visible columns (and offer the opportunity to hide them with the hover menu). To display a new column, the user should click on "Add field at the bottom of the menu.

image

https://www.figma.com/file/xt8O9mFeLl46C5InWwoMrN/Twenty?type=design&node-id=6377-158599&mode=design&t=iF7gvO767aMu93SU-4

Clicking on this button should swap the menu with the same menu that is used by pressing the "+" button on the table title bar:

image

https://www.figma.com/file/xt8O9mFeLl46C5InWwoMrN/Twenty?type=design&node-id=11153-52210&mode=design&t=iF7gvO767aMu93SU-4

Clicking on "Add custom field" should open swap the menu with the "field type" menu:

image

https://www.figma.com/file/xt8O9mFeLl46C5InWwoMrN/Twenty?type=design&node-id=11153-52211&mode=design&t=iF7gvO767aMu93SU-4

Then, the new column should be created and the column menu be opened:

image

https://www.figma.com/file/xt8O9mFeLl46C5InWwoMrN/Twenty?type=design&node-id=11153-52212&mode=design&t=iF7gvO767aMu93SU-4

abdulrahmancodes:
Implements permission intersection (AND logic) to prevent permission escalation when agents act on behalf of users.

### Changes:
- **Permission Intersection**: Operations requiring both user AND agent permissions
- **RoleContext Type**: Unified type supporting single `roleId` or multiple `roleIds` for intersection
- **CRUD Services**: Updated to accept `roleContext` for granular permission control
- **Agent Integration**: Chat agents now use user + agent role intersection for all operations
- **ORM Layer**: Enhanced `getRepository` to support multi-role permission checks

### Related:
- Part 2 of  ["Acting on behalf of user" concept PR](https://github.com/twentyhq/twenty/pull/15103)

[Closes #1661](https://github.com/twentyhq/core-team-issues/issues/1661)


packages/twenty-server/src/engine/twenty-orm/entity-manager/workspace-entity-manager.ts
```diff
@@ -88,14 +89,32 @@ export class WorkspaceEntityManager extends EntityManager {
     permissionOptions?: {
       shouldBypassPermissionChecks?: boolean;
       roleId?: string;
```
FelixMalfait: It's a bit strange to had `roleId` and `roleIds` living in parallel, do you plan to deprecate and remove `roleId` in this PR?

packages/twenty-server/src/engine/twenty-orm/utils/build-permission-options.util.ts
```diff
@@ -0,0 +1,18 @@
+import { type RoleContext } from 'src/engine/metadata-modules/role/types/role-context.type';
+
+export const buildPermissionOptions = (
+  roleContext: RoleContext,
+):
+  | { roleId: string }
+  | { roleIds: { intersection: string[] } }
```
FelixMalfait: Took me a while to understand this api, before I realized I suggested it, so maybe it would be nice to add `union` and leaving it with a TODO/throwing an exception, that way the reader would understand what is the intent here? We can also create a dedicated type for this as it's redeclared multiple times

packages/twenty-server/src/engine/twenty-orm/utils/compute-permission-intersection.util.ts
```diff
@@ -0,0 +1,87 @@
+import {
+  type ObjectsPermissionsDeprecated,
+  type RestrictedFieldPermissions,
+} from 'twenty-shared/types';
+
+export const computePermissionIntersection = (
+  permissionsArray: ObjectsPermissionsDeprecated[],
```
FelixMalfait: Would it be costly to use the new ObjectPermissions instead of ObjectPermissionsDeprecated? It feels like we are creating quite a lot of debt here otherwise (canRead vs canReadObjectRecords etc)

packages/twenty-server/src/database/commands/upgrade-version-command/1-8/1-8-deduplicate-unique-fields.command.ts
```diff
@@ -187,7 +187,7 @@ export class DeduplicateUniqueFieldsCommand extends ActiveOrSuspendedWorkspacesM
   }) {
     const workspaceMemberRepository = dataSource.getRepository(
       'workspaceMember',
-      true,
+      { shouldBypassPermissionChecks: true },
```
FelixMalfait, line 190: more clear indeed!

packages/twenty-server/src/engine/twenty-orm/entity-manager/workspace-entity-manager.ts
abdulrahmancodes: Both will coexist: `roleId` for single-role cases, `roleIds` for permission intersection (agents). Could you clarify why you think `roleId` should be removed? Am I missing something here?

packages/twenty-server/src/engine/core-modules/record-crud/services/create-record.service.ts
```diff
@@ -43,7 +44,7 @@ export class CreateRecordService {
         await this.twentyORMGlobalManager.getRepositoryForWorkspace(
           workspaceId,
           objectName,
-          roleId ? { roleId } : { shouldBypassPermissionChecks: true },
+          buildPermissionOptions(roleContext || {}),
```
FelixMalfait: nit: maybe set {} as default parameter value in buildPermissionOptions?

FelixMalfait: (and you can replace the multiple `roleContext || {}`

packages/twenty-server/src/engine/twenty-orm/entity-manager/workspace-entity-manager.ts
FelixMalfait: When you design an API whether it's a function calling signature, a react component, etc. It's nice to check if it is pretty and if and outside developper would make sense of it, here that is not the case. Having 1 role id is just a sub-case of the multi-role mode.

It could be something like this:

```typescript
type RoleId = string;

type RoleContext = {
  unionOf?: RoleId[];  
  intersectionOf?: RoleId[];
  shouldBypassPermission?: boolean;
};
```

And if you want you can just use `unionOf` as a substitute for single roleId for now, and we will implement union later down the line (one user will have multiple roles)

packages/twenty-server/src/modules/workflow/workflow-executor/workflow-actions/record-crud/create-record.workflow-action.ts
```diff
@@ -64,7 +64,9 @@ export class CreateRecordWorkflowAction implements WorkflowAction {
       objectRecord: workflowActionInput.objectRecord,
       workspaceId,
       createdBy,
-      roleId: executionContext.roleId,
+      rolePermissionConfig: executionContext.roleId
```
FelixMalfait: This doesn't feel pretty, should your execution context return a RolePermissionConfig directly?

packages/twenty-shared/src/types/ObjectPermissionsDeprecated.ts
```diff
@@ -1,10 +0,0 @@
-import { type RestrictedFieldsPermissions } from './RestrictedFieldsPermissions';
-
-// TODO: DEPRECATE THIS
```
FelixMalfait, line 3: Great! Thank you!

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/resolvers/graphql-query-merge-many-resolver.service.ts
```diff
@@ -335,8 +335,11 @@ export class GraphqlQueryMergeManyResolverService extends GraphqlQueryBaseResolv
       try {
         const repository = executionArgs.workspaceDataSource.getRepository(
           relationField.objectMetadata.nameSingular,
-          executionArgs.isExecutedByApiKey,
-          executionArgs.roleId,
+          executionArgs.isExecutedByApiKey
```
ijreilly: by the way i thought api keys had roles now?

packages/twenty-server/src/engine/twenty-orm/utils/compute-permission-intersection.util.ts
```diff
@@ -0,0 +1,92 @@
+import {
+  type ObjectsPermissions,
+  type RestrictedFieldPermissions,
+} from 'twenty-shared/types';
+
+export const computePermissionIntersection = (
```
ijreilly, line 6: can you please add a test on this? i think it's important

packages/twenty-server/src/engine/twenty-orm/entity-manager/workspace-entity-manager.ts
```diff
@@ -83,43 +88,69 @@ export class WorkspaceEntityManager extends EntityManager {
...
 
     let objectPermissions = {};
+    let shouldBypassPermissionChecks = false;
+    const objectPermissionsByRoleId = dataSource.permissionsPerRoleId;
+
+    if (
+      rolePermissionConfig &&
+      'shouldBypassPermissionChecks' in rolePermissionConfig
+    ) {
+      shouldBypassPermissionChecks = true;
```
ijreilly:
```suggestion
      shouldBypassPermissionChecks = rolePermissionConfig.shouldBypassPermissionChecks;
```

seems less dangerous, even though for now rolePermissionConfig only accepts truthy value for shouldBypassPermissionChecks when present

ijreilly (commented):
I have not tested but looks good!
Please add some tests on intersected permissions computing!

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/resolvers/graphql-query-merge-many-resolver.service.ts
abdulrahmancodes: Oh yeah, right! It was like this previously so I kept it as-is, but I've updated it now to use the API key's role instead of bypassing permissions.

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/helpers/process-nested-relations-v2.helper.ts
```diff
@@ -147,8 +147,11 @@ export class ProcessNestedRelationsV2Helper {
 
     const targetObjectRepository = workspaceDataSource.getRepository(
       targetObjectMetadata.nameSingular,
-      shouldBypassPermissionChecks,
-      roleId,
+      shouldBypassPermissionChecks
```
FelixMalfait: Like in other places where you've addressed that feedback already, I would have tried to introduce rolePermissionConfig at the level above already to avoid that kind of nested ternaries

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/interfaces/base-resolver-service.ts
```diff
@@ -155,8 +155,11 @@ export abstract class GraphqlQueryBaseResolverService<
 
       const repository = workspaceDataSource.getRepository(
         objectMetadataItemWithFieldMaps.nameSingular,
-        shouldBypassPermissionChecks,
-        roleId,
+        shouldBypassPermissionChecks
```
FelixMalfait: Probably the same here too? (RolePermissionConfig built earlier possible?)

packages/twenty-server/src/engine/core-modules/ai/services/tool.service.ts
```diff
@@ -65,7 +89,7 @@ export class ToolService {
         return;
       }
 
-      if (objectPermission.canUpdate) {
+      if (objectPermission.canUpdateObjectRecords) {
```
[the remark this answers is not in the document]
abdulrahmancodes: this is intentional

packages/twenty-server/src/modules/workflow/workflow-executor/workflow-actions/ai-agent/services/ai-agent-executor.service.ts
```diff
@@ -45,21 +46,33 @@ export class AiAgentExecutorService {
...
     });
 
-    const effectiveRoleId = roleIdOverride || roleTarget?.roleId;
+    const agentRoleId = roleTarget?.roleId;
 
-    if (!effectiveRoleId) {
-      const actionTools = await this.toolAdapterService.getTools();
+    if (!rolePermissionConfig && !agentRoleId) {
+      return await this.toolAdapterService.getTools();
+    }
```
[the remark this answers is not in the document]
abdulrahmancodes: intentional

packages/twenty-server/src/engine/core-modules/ai/services/tool.service.ts
```diff
@@ -43,7 +45,29 @@ export class ToolService {
...
+    if ('unionOf' in rolePermissionConfig) {
+      if (rolePermissionConfig.unionOf.length === 1) {
+        objectPermissions = rolesPermissions[rolePermissionConfig.unionOf[0]];
+      } else {
+        // TODO: Implement union logic for multiple roles
+        throw new Error(
+          'Union permission logic for multiple roles not yet implemented',
+        );
+      }
+    } else if ('intersectionOf' in rolePermissionConfig) {
```
FelixMalfait, line 59: We don't need to address it now. In my mind in the future it's not really a if, but it's:
 Step 1 — make the union (which will be all roles the user has)
 Step 2 — intersect the union result with whatever role (if any) is provided in the intersection array

This works for now!

FelixMalfait (approved):
Great work!!!

PR #15158, opened 2025-10-17
FIX: error in devtools while opening domain settings

### LINKED ISSUE #15154: Error in devtools while opening Domains settings
**Scenario:**
1. Log in
2. Go to Settings > Domains

**Actual:** Error in devtools appears

Image

**Expected:** No error in devtools

shantanugupta2004:
Fixes #15154 
Addressed a React warning that occurred on the "Settings > Domains" page: `Warning: Cannot update a component (Batcher) while rendering a different component (SettingsPublicDomainsListCard)`

The warning was caused by a state update being called directly within the render function of the `SettingsPublicDomainsListCard` component.  This PR fixes the issue by moving the state update into a useEffect hook. This ensures that the state is updated only after the component has rendered, which is the correct way to handle side effects in React.

Changes:  Wrapped the setSelectedPublicDomain(undefined) call in a useEffect hook in `SettingsPublicDomainsListCard.tsx`.


FelixMalfait:
Thanks for this! But I think we want to solve it differently, introducing a useEffect here doesn't seem like a clean fix

PR #15161, opened 2025-10-17
Fix: Enable destroy/restore actions when viewing deleted records dire…

### LINKED ISSUE #15156: Can't destroy or restore a deleted record
**Scenario:**
1. Log in
2. Open any records table view
3. Delete one record
4. See deleted records via command menu
5. Try to destroy a record

**Actual:** Destroy and restore buttons are not available in table view or in command menu

https://github.com/user-attachments/assets/c91db101-2662-4171-a350-80fe2cfa739a

**Expected:** Destroy and restore buttons must be available in table view and in command menu

**Note:** Buttons are available once record details have been checked (only when said record is selected, doesn't matter if it's only selected record or there are more)

Om7035:
## Issue
When a user deletes a record and then tries to access it via the command menu, the "Destroy" and "Restore" buttons are not available in either the table view or the command menu.

## Root Cause
The action registration logic only considered filters to determine if we're viewing deleted records, but didn't account for the case where we're viewing a deleted record directly (e.g., through the command menu search).

## Solution
Modified the [useShouldActionBeRegisteredParams](cci:1://file:///f:/HackContri/twenty/packages/twenty-front/src/modules/action-menu/hooks/useShouldActionBeRegisteredParams.ts:19:0-127:2) hook to also consider a record as being in a deleted view when the selected record has `deletedAt` set. This ensures that both single record actions (like `DestroySingleRecordAction`) and multiple records actions (like `DestroyMultipleRecordsAction`) are properly registered when viewing deleted records directly.

## Testing
- Verified that destroy/restore actions appear when viewing deleted records through command menu
- Verified that existing functionality for filtered deleted record views still works

Fixes #15156

### DIFF
--- a/packages/twenty-front/src/modules/action-menu/hooks/useShouldActionBeRegisteredParams.ts
+++ b/packages/twenty-front/src/modules/action-menu/hooks/useShouldActionBeRegisteredParams.ts
@@ -52,9 +52,14 @@ export const useShouldActionBeRegisteredParams = ({
 
   const { isInRightDrawer } = useContext(ActionMenuContext);
 
-  const hasAnySoftDeleteFilterOnView = useRecoilComponentValue(
+  const hasAnySoftDeleteFilterOnViewFromFilters = useRecoilComponentValue(
     hasAnySoftDeleteFilterOnViewComponentSelector,
   );
+  
+  // When viewing a deleted record directly (e.g., through command menu), 
+  // treat it as if we're in a deleted records view
+  const hasAnySoftDeleteFilterOnView = 
+    hasAnySoftDeleteFilterOnViewFromFilters || isDefined(selectedRecord?.deletedAt);
 
   const isShowPage =
     useRecoilComponentValue(contextStoreCurrentViewTypeComponentState) ===


FelixMalfait:
Thanks! Can you please fix the CI? Also the diff is very strange, can you figure out why? Maybe linked to linting? That makes it very hard to review otherwise

charlesBochet:
Hi @Om7035, thanks for opening a PR. the help is really appreciated

I'm going to close it though:
- the diff displayed by Github is not correct (not sure what's wrong but we should only see the changes, might be the way you are using git)
- CIs are red
- this bug seems to be have been resolved due to another root cause see: https://github.com/twentyhq/twenty/pull/15185

Om7035:
> Hi @Om7035, thanks for opening a PR. the help is really appreciated
> 
> I'm going to close it though:
> 
> * the diff displayed by Github is not correct (not sure what's wrong but we should only see the changes, might be the way you are using git)
> * CIs are red
> * this bug seems to be have been resolved due to another root cause see: [Add timestamps to GQL fields in useRecordsFieldVisibleGqlFields hook #15185](https://github.com/twentyhq/twenty/pull/15185)

got it!!!

PR #15162, opened 2025-10-17
Fix: Match cell editor width to column width

### LINKED ISSUE #14599: Celleditors not matching column width
## Bug Description

When a cell is opened for editing, the pop out cell editors seems to have a prespecified with, not matching the width of the corresponding column that triggered the cell editor to emerge.

[screenshot]
Image

## Expected behavior

The expected behaviour would be for the width of the popout editor to = the width of the column that triggered it to emerge, the current behaviour seems strange from a UI perspective.

## Technical inputs

Nothing besides i would think this should be quite a simple css adjustment if this is indeed not intended behaviour

Thanks! :)

Om7035:
## Issue
When a cell is opened for editing, the popup cell editor has a pre-specified width that doesn't match the width of the corresponding column that triggered the cell editor to emerge.

## Root Cause
The `OverlayContainer` in [RecordTableCellEditMode](cci:1://file:///f:/HackContri/twenty/packages/twenty-front/src/modules/object-record/record-table/record-table-cell/components/RecordTableCellEditMode.tsx:45:0-122:2) was using its default width instead of matching the column width.

## Solution
Modified [RecordTableCellEditMode](cci:1://file:///f:/HackContri/twenty/packages/twenty-front/src/modules/object-record/record-table/record-table-cell/components/RecordTableCellEditMode.tsx:45:0-122:2) to:
1. Import [getRecordTableColumnFieldWidthCSSVariableName](cci:1://file:///f:/HackContri/twenty/packages/twenty-front/src/modules/object-record/record-table/utils/getRecordTableColumnFieldWidthCSSVariableName.ts:0:0-4:2) utility
2. Create `StyledOverlayContainer` that extends `OverlayContainer` with dynamic width
3. Use CSS variable `--record-table-column-field-{columnIndex}` to set width based on column index
4. Pass `cellPosition.column` as `columnIndex` prop

## Technical Details
- Uses existing CSS variable system that stores column widths
- Leverages `cellPosition.column` from `RecordTableCellContext`
- Maintains all existing functionality while fixing the width issue
- No breaking changes to existing API

## Testing
- Cell editors now properly match their column widths
- All existing functionality is preserved
- Follows existing CSS variable patterns in the codebase

## Expected Behavior
The cell editor width now dynamically matches the width of the column that triggered it, providing a consistent and intuitive user experience.

Fixes #14599

### DIFF
--- a/packages/twenty-front/src/modules/action-menu/hooks/useShouldActionBeRegisteredParams.ts
+++ b/packages/twenty-front/src/modules/action-menu/hooks/useShouldActionBeRegisteredParams.ts
@@ -16,6 +16,7 @@ import { recordStoreFamilyState } from '@/object-record/record-store/states/reco
 import { useRecoilComponentValue } from '@/ui/utilities/state/component-state/hooks/useRecoilComponentValue';
 import { useContext } from 'react';
 import { useRecoilCallback, useRecoilValue } from 'recoil';
+import { isDefined } from 'twenty-shared/utils';
 
 export const useShouldActionBeRegisteredParams = ({
   objectMetadataItem,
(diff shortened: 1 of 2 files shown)


charlesBochet:
Closing for same reasons as the other one

PR #15173, opened 2025-10-17
Fix settings admin workspace

ijreilly:
We have a bug in production where activeWorkspace isn't defined and call to getWorkspaceSchemaName() breaks. 
Let's fix it!


etiennejouan (approved):

packages/twenty-front/src/modules/settings/admin-panel/components/SettingsAdminWorkspaceContent.tsx
```diff
@@ -169,7 +169,9 @@ export const SettingsAdminWorkspaceContent = ({
     {
       Icon: IconId,
       label: t`Schema name`,
-      value: getWorkspaceSchemaName(activeWorkspace?.id!),
+      value: isDefined(activeWorkspace?.id)
```
charlesBochet, line 172: my bad on this one,  I reviewed it too fast and we actually have more to fix:
- getWorkspaceSchemaName util file name is not following FE convention

PR #15175, opened 2025-10-17
Introduce a command to regenerate search vectors for standard and custom objects to fix accent issue.

mabdullahabaid:
Legacy workspaces still hold the old stored expression, which omits public.unaccent_immutable, so their tsvectors remain accented and can’t match the new, unaccented queries. Metadata sync doesn’t touch asExpression, so only a targeted drop/recreate fixes the underlying column.

In simpler words, the search vector should contain `mader` instead of `mäder` for the search to work properly. Therefore, this command regenerates the search vector across every object that uses `SEARCH_FIELDS_FOR_*`.

Note that dashboard has a searchVector, but breaks the pattern of using `SEARCH_FIELDS_FOR_DASHBOARD`. If you look at packages/twenty-server/src/modules/dashboard/standard-objects/dashboard.workspace-entity.ts:116, the searchVector field is hard-coded as 

```
asExpression: `to_tsvector('english', title)`
```

Therefore, the following code snippet.

```
    const storedExpression = hasAsExpressionSetting(
      searchVectorFieldMetadata.settings,
    )
      ? searchVectorFieldMetadata.settings.asExpression
      : undefined;

    if (storedExpression) {
      return storedExpression;
    }

    return undefined;
```

It checks whether the searchVector field already carries its own asExpression value in metadata. If the settings object includes that string, it returns it so the upgrade can reuse the existing expression for objects that aren’t in our predefined lists. If not, it returns undefined, signaling there’s no stored expression to fall back on.


packages/twenty-server/src/database/commands/upgrade-version-command/1-10/1-10-regenerate-search-vectors.command.ts
```diff
@@ -0,0 +1,328 @@
...
+    let queryRunner: QueryRunner | undefined;
+
+    try {
+      queryRunner = this.coreDataSource.createQueryRunner();
+
+      await queryRunner.connect();
+      await queryRunner.startTransaction();
+
+      for (const objectMetadata of searchableObjects) {
+        if (objectMetadata.standardId === STANDARD_OBJECT_IDS.dashboard) {
```
FelixMalfait: Why not fix the root cause (i.e. align dashboard's behavior with others) instead of the symptom?

packages/twenty-server/src/database/commands/upgrade-version-command/1-10/1-10-regenerate-search-vectors.command.ts
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    await queryRunner.commitTransaction();
+  }
+
+  private async ensureUnaccentFunction(): Promise<void> {
+    const result = await this.coreDataSource.query(`
+      SELECT EXISTS (
+        SELECT 1 FROM pg_proc p
+        JOIN pg_namespace n ON p.pronamespace = n.oid
```
FelixMalfait, line 384: Are you sure this won't break for the same reason I told you accessing schema breaks? (not always a root user)

packages/twenty-server/src/database/commands/upgrade-version-command/1-10/1-10-regenerate-search-vectors.command.ts
```diff
@@ -0,0 +1,328 @@
...
+    const storedExpressionFromSettings = hasAsExpressionSetting(
+      searchVectorFieldMetadata.settings,
+    )
+      ? searchVectorFieldMetadata.settings.asExpression
+      : undefined;
+
+    if (storedExpressionFromSettings) {
+      return storedExpressionFromSettings;
+    }
+
```
FelixMalfait, line 230: I don't understand all those edge cases. This seem over-engineered. And if not we should split the code into readable/explicit sub-functions with good name, because it's very hard to understand what each case represent exactly? (it doesn't have a standard id but also is not custom? But is has a stored expression? And then we use Reflect?

packages/twenty-server/src/database/commands/upgrade-version-command/1-10/1-10-regenerate-search-vectors.command.ts
mabdullahabaid: We ran this with 1-8 and before pushing that command, I created a database with different roles and tried it out.

"Every normal PostgreSQL user can query these catalog tables, because they’re readable by all roles. They’re essentially metadata about the database structure, not its private data.

Your query just checks if a function called unaccent_immutable exists in the public schema — nothing more invasive than peeking into the database’s list of functions."

FelixMalfait: OK!

packages/twenty-server/src/database/commands/upgrade-version-command/1-10/1-10-regenerate-search-vectors.command.ts
mabdullahabaid: It is over-engineered to circumvent the issue with dashboard search vector - without fetching the current state of that search vector, the command overwrites it. 

The ideal solution is to fix dashboard first - I pushed this command without touching dashboard because I tried to fix it for the end-user who reported the issue.

However, if you prefer we always solve things from the root, I can do that from here on.

packages/twenty-server/src/database/commands/upgrade-version-command/1-10/1-10-regenerate-search-vectors.command.ts
mabdullahabaid: Left this for two reasons:

1. We had a conversation around wasted effort when we were moving to dynamic search fields anyway (it's a shortcut, but if the shortcuts should be avoided, that makes perfect sense).
2. Wrote this command with the perspective that we need to fix the issue faced by the end-user who reported it quickly.

FelixMalfait: A good long-term fix is never a wasted effort! For me the wasted effort was because updating the search fields would come later down the line with the "search field become configurable in the UI" workstream anyway, so this was investing a lot in temporary migrations commands that weren't useful on a long-term horizon

packages/twenty-server/src/database/commands/upgrade-version-command/1-10/1-10-regenerate-search-vectors.command.ts
FelixMalfait: "Fixing dashboard" didn't seem like a complex thing to me but I didn't dig into it, is it more complex than I thought? From what I thought changing the dashboard expression seemed easier than writing/understanding this complex logic you have here, but maybe I missed part of the complexity!

mabdullahabaid: No, it should not be hard. I think it is my style of solving problems that needs to change and match here.

I have been used to hot-fixes to get things working due to past experiences, so I need to adapt how I think.

packages/twenty-server/src/database/commands/upgrade-version-command/1-10/1-10-regenerate-search-vectors.command.ts
mabdullahabaid: Okay, I will re-work this.

mabdullahabaid:
@FelixMalfait ready for review.

FelixMalfait:
I'll push one more change: this wraps all workspace into one big transaction, we usually prefer smaller transactions than long-running transactions. For now I'll just remove transactions entirely (unless it's a dry-run) but what you didn't wasn't wrong at all.

FelixMalfait (approved):
Thank you, great work!

PR #14917, opened 2025-10-06
chore(twenty-website): upgrade next

AMoreaux:
Standardized `PageProps` usage across multiple pages. Upgraded `next`, `eslint-config-next`, and related dependencies to `^15.5.4` for compatibility and performance improvements.

PR #15181, opened 2025-10-17
Release - revert/downgrade twenty website next version

etiennejouan:
[PR #14917](https://github.com/twentyhq/twenty/pull/14917/files#diff-e37dead9533eef25d3a1ac323bb68e93ad2edbb932e972e48f4c756e3c2d5c0f) upgraded twenty-website to Next.js v15, which requires React 19. However, the twenty-ui package (imported by twenty-website) uses React 18.2, causing a React version mismatch error.
Solution :  Downgrade Next.js from ^15.5.4 to ^14.2.0 to maintain compatibility with React 18.2 used across the monorepo.

### DIFF
--- a/packages/twenty-website/package.json
+++ b/packages/twenty-website/package.json
@@ -23,7 +23,7 @@
     "drizzle-kit": "^0.20.14",
     "facepaint": "^1.2.1",
     "gray-matter": "^4.0.3",
-    "next": "^15.5.4",
+    "next": "^14.2.0",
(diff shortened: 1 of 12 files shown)


ijreilly (approved):

FelixMalfait:
If this isn't merged today, I'll try a partial React 19 upgrade today, it's too bad to go back in the past 😄. No problem if we need to merge this before though

PR #15184, opened 2025-10-17
removed  visible scrollbar  

### LINKED ISSUE #15152: Visible scrolls while adding new field to an object
**Scenario:**
1. Log in
2. Go to Settings > Data model > any existing object > Add field > Relation

**Actual:** There are visible scrolls in select dropdown

Image

**Expected:** There should be visible scrolls offered by Twenty in select dropdown

nissu99:
Fixes  #15152

Added CSS to override the scrollbar.

Before
image

After
image


charlesBochet:
Hi @nissu99 thanks for opening a PR, any help is much appreciated

I'm not sure about this fix, I cannot reproduce the issue as I think it's only happening in windows setup. That's what I see on MacOS:
image

1) you confirm it's happening on main? (cc @BOHEUS)
2) If yes, do we have the same behavior on regular Select? (the one used to select the timezone for instance)
3) If yes, why is this happening only on this `SettingsMorphRelationMultiSelect` component that seems to be completely similar to the regular `Select.tsx` we are usually using?

Once we understand why this is happening, we can discuss the fix itself which is not following codebase standards

nissu99:
@charlesBochet  ,I reproduced the issue on ubuntu ,tried different browsers still it persisted.

BOHEUS:
@charlesBochet 
1. yes, it's still on main
2. yes, it's the same for timezone as well 
image

nissu99:
@charlesBochet @BOHEUS  ,I think the maxHeight thing might be doing this when the number of option exceeds.

charlesBochet:
Ok, so that's a global fix that needs to be done on Dropdowns (just to confirm, do we have same issue on Filter / Sort button on table pages @BOHEUS )?

In this is the case, we should fix: DropdownMenuItemsContainer and take inspiration from ScrollWrapper which is used on Tables / Kanban main container. I'm also thinking that Dropdown could use ScrollWrapper internally what do you think @lucasbordeau?

charlesBochet:
@BOHEUS, is ScrollWrapper actually behaving well on Windows, do we see the scrollbars on the Table or the Kanban for instance?

lucasbordeau:
I'm closing this PR because there is a miscomprehension on this issue and the core team has to take it.

Thank you for contributing.

charlesBochet:
Discussed with @lucasbordeau indeed, we have merged the following fix for the CSS: https://github.com/twentyhq/twenty/pull/15199 and we will revisit this specific issue. It seems that this is a local problem that Lucas will be able to troubleshoot but it's far from being trivial

Thanks for the help @nissu99

nissu99:
Indeed,  at first , I thought this must be something fairly simple to solve but it was not that simple, anyway CSS hack is working . Thank You!!

PR #15185, opened 2025-10-18
Add timestamps to GQL fields in useRecordsFieldVisibleGqlFields hook

### LINKED ISSUE #15156: Can't destroy or restore a deleted record

ManikanthMartha:
This fixes #15156 

Issue: 
Restore and Destroy buttons not appearing in action menu for deleted records until the record detail view is opened.

Cause:
The record index/table view queries only fetched fields that were visible as table columns
The [deletedAt] field (along with [createdAt] and [updatedAt]) was not included in these queries since it's not a visible column
The action menu logic checks [selectedRecord?.deletedAt] to determine if a record is deleted and which actions to display
Without the [deletedAt] field in the record store, the action menu couldn't detect deleted records
Opening the detailed view would fetch all fields (including [deletedAt]), which is why the buttons would appear afterward

Solution
Modified [useRecordsFieldVisibleGqlFields] to always include the standard fields ([createdAt], [updatedAt], [deletedAt]) in record index queries, regardless of column visibility. This ensures the action menu can immediately detect deleted records


ManikanthMartha:
https://github.com/user-attachments/assets/3cc3da89-17e4-4e98-8e68-00868d2ee699

FelixMalfait (approved):
Thank you!

PR #15186, opened 2025-10-18
Fix env not optional + serverless logging

martmull:
Several fixes after discussing with @BOHEUS 
- set applicationManifest env key optional
- fix server local serverless function logging (introduces a new env variable `SERVERLESS_LOGS_ENABLED` defaulting to false)


packages/twenty-server/src/engine/metadata-modules/route-trigger/route-trigger.service.ts
```diff
@@ -133,11 +137,32 @@ export class RouteTriggerService {
...
+    return this.formatServerlessControllerResponse(result);
   }
+
+  formatServerlessControllerResponse = (result: ServerlessExecuteResult) => {
+    if (!isDefined(result)) {
+      return result;
+    }
+
+    if (result.error) {
+      throw new HttpException(
```
charlesBochet: **change:** Services should throw CustomExceptions (as they could be called through different entrypoints: rest, graphql, jobs...) and controllers should catch them and throw HttpExceptions (use nestjs filters for that)

packages/twenty-server/src/engine/metadata-modules/route-trigger/route-trigger.service.ts
```diff
@@ -133,11 +137,32 @@ export class RouteTriggerService {
...
+        {
+          errorType: result.error.errorType,
+          message: result.error.errorMessage,
+        },
+        HttpStatus.INTERNAL_SERVER_ERROR,
+      );
+    }
+
+    return result.data;
```
charlesBochet, line 160: **question:** are we sure that we want the service to only return data and not errors?

charlesBochet (changes requested):
Letf comments!

packages/twenty-server/src/engine/metadata-modules/route-trigger/route-trigger.service.ts
martmull: If there is an error it throws the error

charlesBochet: ok!

packages/twenty-server/src/engine/metadata-modules/route-trigger/route-trigger.service.ts
```diff
@@ -133,11 +137,32 @@ export class RouteTriggerService {
...
+        routeTriggerWithPathParams.routeTrigger.serverlessFunction.id,
+        routeTriggerWithPathParams.routeTrigger.workspaceId,
+        executionParams,
+        'draft',
+      );
+
+    return this.formatServerlessControllerResponse(result);
   }
+
+  formatServerlessControllerResponse = (result: ServerlessExecuteResult) => {
```
charlesBochet: I'm not sure about the naming: `formatServerlessControllerResponse`

1) why ControllerResponse as we are in a service?
2) let's add orThrow here as this is a big part of the role of this function

martmull: ok, i removed that and moved the code to handle

charlesBochet (approved):
LGTM!

PR #15189, opened 2025-10-18
Fix: Make CI .env manipulation robust against missing trailing newlines

FelixMalfait:
## Problem

CI  workflow started timing out on October 14, 2025 after commit `d750df7fff` removed the trailing newline from `.env.example`.

## Root Cause

When `.env.example` lacks a trailing newline:
```bash
# Last line without newline
# CLICKHOUSE_URL=...twenty
```

And CI runs:
```bash
echo "NODE_PORT=3002" >> .env
```

Result:
```bash
# CLICKHOUSE_URL=...twentyNODE_PORT=3002  ← Commented out!
```

Server starts on default port 3000 instead of 3002, health check fails.

## Fix

1. **Restore trailing newline** to `.env.example`
2. **Make all CI `.env` operations robust** by adding `echo "" >> .env` before appending
3. **Simplified `set_env_var`** function to always add newline first

Now works regardless of whether template files have trailing newlines.

## Files Changed

- 6 CI workflow files
- 1 .env.example file


FelixMalfait:
Unfortunately ESLINT does not apply on .env.example ; I'll still do a followup PR to add consistency to JS files but that's unrelated to the issue here

PR #15194, opened 2025-10-19
fix: Convert NaN to null in GraphQL aggregate responses to prevent serialization errors

### LINKED ISSUE #15192: Object Listing not working
## Bug Description

I added a custom object (Company Details) Data model which are connected to company with one to one relationship, then I imported 2 companies using API which populates company and related company details. Now listing for the custom object (company details) is not working. It shows me toast with message "An error occurred.". Checking console, I can see below error:

```
useFindManyRecords for "companyDetails" error : ApolloError: Float cannot represent non numeric value: NaN
Float cannot represent non numeric value: NaN
```

Companies and company details were imported successfully as I can their entries in database and  API also had returned 201 status.

Note: Company Details are not part of company, because these are a set of company data that came from a particular source, there can be multiple sources each having different data, sometimes same data with different values, so have separated company details from company.

Image

rawadhossain:
### Related issue(s)/PR(s)
- ### Fixes #15192

## Problem
When listing records with aggregated fields (especially AVG operations on CURRENCY fields like `avgAnnualRecurringRevenueAmountMicros`), the application was failing with the error:

```
ApolloError: Float cannot represent non numeric value: NaN
```

This error occurs when GraphQL's `Float` scalar type attempts to serialize a JavaScript `NaN` value, which is not allowed by the GraphQL specification.

## Solution

The fix adds NaN sanitization in two key locations where aggregate results are processed before being passed to GraphQL:

### 1. `ObjectRecordsToGraphqlConnectionHelper` (for findMany queries)

**File**: `packages/twenty-server/src/engine/api/graphql/graphql-query-runner/helpers/object-records-to-graphql-connection.helper.ts`

**Change**: Modified the `extractAggregatedFieldsValues` method to check for `NaN` values and convert them to `null`:

```typescript
// Guard against NaN values that would cause GraphQL Float serialization errors
// Convert NaN to null to ensure proper GraphQL type compatibility
const sanitizedValue =
  typeof aggregatedFieldValue === 'number' &&
  Number.isNaN(aggregatedFieldValue)
    ? null
    : aggregatedFieldValue;

return {
          ...acc,
          [aggregatedFieldName]: sanitizedValue,
        };
```

### 2. `formatResultWithGroupByDimensionValues` (for groupBy queries)

**File**: `packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/resolvers/utils/format-result-with-group-by-dimension-values.util.ts`

**Change**: Added sanitization of the entire group object before processing:

```typescript
    // Sanitize NaN values in the group object to prevent GraphQL Float serialization errors
    const sanitizedGroup = Object.entries(group).reduce<Record<string, unknown>>(
      (acc, [key, value]) => {
        acc[key] = typeof value === 'number' && Number.isNaN(value) ? null : value;
        return acc;
      },
      {},
    );

    const groupWithValueMappedToUnaliasedColumn = {
      ...sanitizedGroup,
      ...groupByColumnsWithQuotes.reduce<Record<string, unknown>>(
        (acc, groupByColumn) => {
          const value = sanitizedGroup[groupByColumn.alias];

          acc[removeQuotes(groupByColumn.columnNameWithQuotes)] =
            getTranslatedValueIfApplicable(
              value,
              groupByColumn.dateGranularity,
            );

          return acc;
        },
        {},
      ),
    };
```

Both now convert NaN values to null before GraphQL serialization.

## Changes
- ✅ Sanitize aggregate values in findMany queries
- ✅ Sanitize aggregate values in groupBy queries
- ✅ Add unit tests for both sanitization paths
- ✅ Preserve all other numeric values (0, negatives, Infinity)

## Why This Fix Works
1. **Type Safety**: The fix checks `typeof value === 'number'` before calling `Number.isNaN()` to ensure we only process numeric values.
2. **GraphQL Compatibility**: By converting `NaN` to `null`, we ensure all values passed to GraphQL are valid for the `Float` scalar type, which accepts finite numbers and `null`.
3. **Minimal Impact**: The fix only affects values that are already problematic (`NaN`). Normal numeric values (including `0`, negative numbers, and `Infinity`) pass through unchanged.
4. **Semantic Correctness**: Converting `NaN` to `null` is semantically appropriate for aggregate operations, as both represent "no valid value" or "undefined result".

## Testing
Added unit tests:
- **`object-records-to-graphql-connection.helper.spec.ts`**
- **`format-result-with-group-by-dimension-values.util.spec.ts`**
Unit tests were added for both modified functions:
1. **`object-records-to-graphql-connection.helper.spec.ts`**: Tests that `NaN` values in aggregated fields are converted to `null` while preserving valid numeric values.
2. **`format-result-with-group-by-dimension-values.util.spec.ts`**: Tests that `NaN` values in group-by results are converted to `null`.
- Unit tests added for both modified functions
- Verified NaN → null conversion
- Verified other numeric values pass through unchanged

## Impact
- Users can now list records with aggregate fields that have no/null values
- API returns `null` instead of throwing an error
- Minimal performance impact (simple type check)

Fixes #15192

### DIFF
--- a/packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/resolvers/utils/__tests__/format-result-with-group-by-dimension-values.util.spec.ts
+++ b/packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/resolvers/utils/__tests__/format-result-with-group-by-dimension-values.util.spec.ts
@@ -1,0 +1,110 @@
+import { formatResultWithGroupByDimensionValues } from '../format-result-with-group-by-dimension-values.util';
+
+describe('formatResultWithGroupByDimensionValues', () => {
+  it('should convert NaN values to null in group results', () => {
+    const result = [
+      {
+        city: 'Paris',
+        avgEmployees: NaN,
+        totalCount: 5,
+        maxRevenue: 100000,
+      },
+      {
+        city: 'London',
+        avgEmployees: 25,
+        totalCount: 3,
+        maxRevenue: NaN,
+      },
+    ];
+
+    const groupByColumnsWithQuotes = [
+      {
+        columnNameWithQuotes: '"city"',
+        alias: 'city',
+      },
+    ];
+
+    const formatted = formatResultWithGroupByDimensionValues(
+      result,
+      groupByColumnsWithQuotes,
+    );
+
+    expect(formatted).toHaveLength(2);
+    expect(formatted[0]).toMatchObject({
+      city: 'Paris',
+      avgEmployees: null, // NaN should be converted to null
+      totalCount: 5,
+      maxRevenue: 100000,
+    });
+    expect(formatted[1]).toMatchObject({
+      city: 'London',
+      avgEmployees: 25,
+      totalCount: 3,
+      maxRevenue: null, // NaN should be converted to null
+    });
+  });
+
+  it('should handle normal numeric values correctly', () => {
+    const result = [
+      {
+        category: 'A',
+        avgValue: 42.5,
+        minValue: 0,
+        maxValue: -10.5,
+        totalCount: 100,
+      },
+    ];
+
+    const groupByColumnsWithQuotes = [
+      {
+        columnNameWithQuotes: '"category"',
+        alias: 'category',
+      },
+    ];
+
+    const formatted = formatResultWithGroupByDimensionValues(
+      result,
+      groupByColumnsWithQuotes,
+    );
+
+    expect(formatted).toHaveLength(1);
+    expect(formatted[0]).toMatchObject({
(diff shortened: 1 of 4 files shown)


charlesBochet:
Hi @rawadhossain, thanks for opening a PR.

I don't fix this is a right fix: unless I'm mistaken, we should not end with NaN in postgres in the first place. This is actually showing that we are missing a proper validation layer while inserting data through graphql / rest.
We have an ongoing work that we call the "Common API" led by @etiennejouan  This will fix it as part of the effort.

I'm closing the PR but I'm assigning @etiennejouan to make sure I'm not missing something.

Thanks for the help anyway, this is much appreciated

PR #15203, opened 2025-10-20
Fix table virtualization data load < 120 records

### LINKED ISSUE #15196: Table infinite load / auto pagination is not working between 60 and 120 records
## Bug Description

It seems that the table pagination is not properly working if it contains between 60 and 120 records:
- below 60, all records are displayed
- between 60 and 120, only the first 60 are displayed
- above 120 records, all records are displayed.

Flagging as CRITICAL as this is impacting customers

lucasbordeau:
This PR fixes two bugs : 
- Data loading not working between 60 and 120 records on an object
- Table not refreshing when creating a record in an empty table.

Fixes https://github.com/twentyhq/twenty/issues/15196


packages/twenty-front/src/modules/object-record/record-table/virtualization/components/RecordTableRowVirtualizedDebugRowHelper.tsx
```diff
@@ -11,12 +11,27 @@ import styled from '@emotion/styled';
...
 import { isDefined } from 'twenty-shared/utils';
 
+const StyledDebugRow = styled.div`
+  position: absolute;
+  left: 250px;
+  top: 5px;
+  z-index: 20;
+  color: darkblue;
+  background-color: white;
+  border: 1px solid blue;
```
charlesBochet: we should use theme here even for Debug :p

StephanieJoly4:
@lucasbordeau @charlesBochet would that make sense to do a patch for this one? this is truly blocking for a few users.
what do you think?

PR #15213, opened 2025-10-20
Remove dependency on lodash.pick.

mabdullahabaid:
Fixes [Dependabot Alert 85](https://github.com/twentyhq/twenty/security/dependabot/85) - prototype pollution in lodash.

Added a shared pick helper (with unit tests) in twenty-shared and refactored front-end/server code to import { pick } from the shared barrel instead of lodash.pick.


charlesBochet:
@mabdullahabaid is this ready for review?

I'm also thinking, do we need pick at all? it seems that we could unstructure this way:

```
const myObject = { a: 'valueA', b: 'valueB', c: 'valueC' }
const { a, b } = myObject
```

Could we refator in this direction?

mabdullahabaid:
@charlesBochet yes, had this conversation with Felix on Discord. Working on it.

packages/twenty-server/src/engine/core-modules/serverless/drivers/layers/2/package.json
```diff
@@ -0,0 +1,45 @@
+{
```
FelixMalfait: You can just replace the previous layer, not worth creating a new layer for that imo

packages/twenty-server/src/engine/core-modules/serverless/drivers/layers/2/yarn.lock
```diff
@@ -0,0 +1,3185 @@
+# This file is generated by running "yarn install" inside your project.
```
FelixMalfait: Because it's a new layer it makes it impossible to review the diff on yarn.lock, double check it when you've updated it!

FelixMalfait (commented):
Almost there!

packages/twenty-server/src/engine/core-modules/serverless/drivers/layers/2/package.json
mabdullahabaid: Command `1-8:fill-null-serverless-function-layer-id` hard-codes dependencies against the hash of layer one. 

Per my understanding, it would cause newer workspace instances to also have lodash.pick and the solution appeared to be adding another layer.

If you feel it would not, let me know and I can fix right away.

packages/twenty-server/src/engine/core-modules/serverless/drivers/layers/2/yarn.lock
mabdullahabaid: I did. This yarn.lock has a different checksum + removed lodash.debounce and lodash.pick references (the former package had already been removed from package.json, but yarn.lock had not been regenerated).

packages/twenty-front/src/modules/activities/hooks/__tests__/useCreateActivityInDB.test.tsx
```diff
@@ -12,8 +11,13 @@ const mockedDate = '2024-03-15T12:00:00.000Z';
...
 global.Date.prototype.toISOString = toISOStringMock;
 
+const { id, title, bodyV2, status, dueAt } = mockedTasks[0];
 const mockedActivity = {
-  ...pick(mockedTasks[0], ['id', 'title', 'bodyV2', 'type', 'status', 'dueAt']),
+  id,
+  title,
+  bodyV2,
+  status,
+  dueAt,
```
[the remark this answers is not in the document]
mabdullahabaid: That is because there is no `type` property for tasks in the database. Even the mockedTasks array has no property called `type` - pick function was dropping type silently since mockedTasks[0] (source) had none.

packages/twenty-server/src/engine/core-modules/serverless/drivers/layers/2/package.json
FelixMalfait: Ok! cc @martmull for input

martmull: Indeed, you did well but we introduced some changes recently about serverlessFunctionLayers.

Now we have the core.serverlessFunctionLayer table that lists all the dependencies of the layer. We don't have the concept of `layerVersion` anymore in the app. So yes you can just update the base layer (ie: version 1) and this updated version will be used for newly created serverlessFunctions without affecting the existing onces.

I will create a ticket to update the `packages/twenty-server/src/engine/core-modules/serverless/drivers/layers` folder so it is more clear.

martmull: @mabdullahabaid here is the fix -> https://github.com/twentyhq/twenty/pull/15219

martmull (approved):
Nice thank you

packages/twenty-server/src/engine/core-modules/serverless/drivers/layers/2/package.json
mabdullahabaid: Oh okay great!!

Thank youu @martmull

PR #15214, opened 2025-10-20
Fixed date filter

lucasbordeau:
This PR fixes a bug with date filters that behave differently when the timezone of the server changes.

Fixes https://github.com/twentyhq/twenty/issues/14689


charlesBochet:
@lucasbordeau, I have tested and the behavior on DATE_TIME is correct!

Regarding DATE, it does not seems right:
1) I'm trying to make a filter on a date field and the gql query are see for IS and IS_TODAY are doing lte + gte which seems not intended by your code
2) For DATE, we should not take timezone into account at all, we are storing the intention of the user to be on a specific date whatever the timezone is (usually it's business related, if it's a rental start date, it will be in the rental location timezone for instance), so in this case we should always store / display in UTC (never applying a timezone at any level). I think we can use this PR to fix this behavior

PR #15217, opened 2025-10-20
fix: Add reserved subdomains constant and update validation on `generateSubdomain`

### LINKED ISSUE #15160: Reserved sub-domains
See comment here: https://github.com/twentyhq/twenty/pull/15159#discussion_r2438651706

This list should be enforced not just for workspace name update but also during creation. Even though we don't explicitly ask for the name during creation, there could be a conflict as the name is inferred from the email address.

naineel1209:
## Description

Fixes #15160 

- Moved the reserved subdomains to a separate shared constant file: `packages/twenty-server/src/engine/core-modules/workspace/constants/reserved-subdomains.constant.ts`
- Updated the validation while generating subdomain to check if the extracted subdomain (from email or display name) is reserved
- When a reserved subdomain is detected, the server will automatically fall back to a random subdomain


packages/twenty-server/src/engine/core-modules/workspace/dtos/update-workspace-input.ts
```diff
@@ -12,135 +12,15 @@ import {
...
 import { UUIDScalarType } from 'src/engine/api/graphql/workspace-schema-builder/graphql-types/scalars';
+import { RESERVED_SUBDOMAINS } from 'src/engine/core-modules/workspace/constants/reserved-subdomains.constant';
 
 @InputType()
 export class UpdateWorkspaceInput {
   @Field({ nullable: true })
   @IsString()
   @IsOptional()
   @Matches(/^(?!api-).*^[a-z0-9][a-z0-9-]{1,28}[a-z0-9]$/)
-  @IsNotIn([
```
FelixMalfait, line 22: The automatedAI review raised a good point on casing, I would guess it affects updates too and there is an issue. Maybe we should get rid of Matches/IsNotIn and perform those checks in `updateWorkspace` method instead so that we can always force-convert to lowercase before?

FelixMalfait (commented):
Thank you! One small comment

packages/twenty-server/src/engine/core-modules/workspace/services/workspace.service.ts
```diff
@@ -100,6 +101,20 @@ export class WorkspaceService extends TypeOrmQueryService<Workspace> {
   }
 
   private async validateSubdomainUpdate(newSubdomain: string) {
+    if (!subdomainPattern.test(newSubdomain)) {
```
charlesBochet: **change:** this part should be extracted to an util (as it's a useful function but we don't want to have to import/provide the service when we need it).

naming: `validateSubdomainOrThrow`

charlesBochet: **polish:** rename subdomainPattern to validSubdomainPattern

packages/twenty-server/src/engine/core-modules/domain-manager/services/domain-manager.service.ts
```diff
@@ -206,8 +207,14 @@ export class DomainManagerService {
   }
 
   async generateSubdomain(params?: { email?: string; displayName?: string }) {
-    const subdomain =
-      this.extractSubdomain(params) ?? generateRandomSubdomain();
+    let subdomain = this.extractSubdomain(params);
+
+    // If extracted subdomain is reserved or matching the reserved subdomain pattern, use random instead
+    if (subdomain && (!subdomainPattern.test(subdomain) || RESERVED_SUBDOMAINS.includes(subdomain))) {
```
charlesBochet: Not a great pattern to define a variable, put it to null, then re-assign it.

let's do something like:
const subdomainFromParams = this.extractSubdomain(params);
const isSubdomainFromParamsReserved = ...

charlesBochet:
@naineel1209 I took a look at the code and I think we should use this opportunity to improve the way we are validating domains.
The code is both in workspace.service.ts validating the domain should be moved in domain-manager.service.ts and be re-used. We are implmeenting twice the same logic here

I'm going to push on your PR

charlesBochet:
@naineel1209 I have pushed a refactor of domains module to setup a maintainable framework:
- `domains/domain-server-config`: low level service that returns the server domain configuration
- `domains/workspace-domains`: low level service that returns workspace domain configuration
- `domains/subdomain-manager`: high level service to handle / validate subdomains
- `domains/custom-domain-manager`: high level service to handle / validate custom domains

With this base, we can unify the logic to handle / validate subdomain only in subdomain-manager (before it was duplicated in `workspace` module and in `domain-manager` module

charlesBochet:
@naineel1209 See this commit for the changes discussed (and applied post folder refactor): https://github.com/twentyhq/twenty/pull/15217/commits/83cc3ef18c67a3e2fee76b7f4f749243211baf36

charlesBochet (approved):

PR #15230, opened 2025-10-21
1750 extensibility twenty sdk v2 use twenty sdk to define an object

### LINKED ISSUE #1750: [Extensibility] [twenty-sdk] [v2] Use twenty-sdk to define an object
Create First decorator if twenty-sdk `Object`

```typescript
import { Object } from `twenty/apps` // or find a relevant barrel here

@Object({
    universalIdentifier: <uuid>,
    nameSingular: "postCard",
    namePlural: "postCards",
    labelSingular: "Post card",
    labelPlural: "Post cards"
})
export class PostCard {}
```

- Use static text analysis it to generate the package.json sent to sync server endpoint

- Keep `object.manifest.json` support but set it as deprecated 
- Update hello-world so it uses the `Object` decorator

martmull:
We maintain jsonc object definition but will deprecate them pretty soon

## Before
image

## After
image


packages/twenty-sdk/src/decorators/object.decorator.ts
```diff
@@ -0,0 +1,13 @@
+type ObjectOptions = {
```
prastoin: ***Question:*** Would this makes sense for `twenty-sdk` to have a local dependency to `twenty-server` and extends flat entities to derivate its local types ? ( typeorm declaration -> flat entities -> twenty-sdk decorators )

prastoin: So we avoid any desync and enhance type safety across both twenty-server expectations and twenty-sdk

martmull: seen with @prastoin will do maybe later

packages/twenty-cli/src/commands/app-add.command.ts
```diff
@@ -56,6 +58,31 @@ export class AppAddCommand {
...
+
+        const objectFileName = `${camelcase(entityName)}.ts`;
+
+        const className = capitalize(camelcase(entityName));
+
+        const decoratorOptions = Object.entries(entityData)
+          .map(([key, value]) => `  ${key}: '${value}',`)
+          .join('\n');
+
+        const decoratedObject = `import { ObjectMetadata } from 'twenty-sdk';
```
Weiko: Can we move this to a constant file?

packages/twenty-cli/src/commands/app-add.command.ts
```diff
@@ -56,6 +58,31 @@ export class AppAddCommand {
 
       const entityData = await this.getEntityToCreateData(entity, entityName);
 
+      if (entity === SyncableEntity.OBJECT) {
+        delete entityData['standardId'];
```
Weiko, line 62: I believe there is already some kind of switch in getEntityToCreateData on the kind of entity. Maybe we can simply not include these in the first place

Weiko (approved):
LGTM

Weiko:
As discussed, many things to improve in terms of code readability/quality but we are ok to move forward with this for now! Thanks!

packages/twenty-cli/src/commands/app-add.command.ts
martmull: those will be removed when jsonc schema will be deprecated

Issue #15201, opened 2025-10-20
Merge function not selecting the correct Company


StephanieJoly4:
## Bug Description

When merging 2 people records, with first prio over second record, name and email from record 1 are shown in the preview but the Company from the preview (and the final one chosen) is the one from the second record.
This was reported by a customer and I'm able to reproduce.

<img width="493" height="207" alt="Image" src="https://github.com/user-attachments/assets/b3a5c4db-f6b7-4d9c-95bf-17e87a02773f" />
<img width="1474" height="791" alt="Image" src="https://github.com/user-attachments/assets/8f9abaae-4e3c-43b4-ae25-b839e9dadb93" />
<img width="1487" height="477" alt="Image" src="https://github.com/user-attachments/assets/fe42ada8-6ee6-4894-9073-f1256262af11" />

I have not checked yet but there is a need to check if this happens only for Company or all relations + if this happens when merging Companies, with other relations. 

Thanks!

harshit078:
@StephanieJoly4 Shall we close this issue as both the PR's are merged and issue is fixed ?

PR #15240, opened 2025-10-21
fix: record merging

anadi45:
Fixes - https://github.com/twentyhq/twenty/issues/15201

- Added support for a priority index in the merging process to determine which record takes precedence when merging many-to-one relationships.
- Updated `mergeManyToOneRelationship` to accept a priority index and return the corresponding record if it exists and is not null.
- Modified `mergeRecordRelationshipData` to pass the priority index to the merging utility.
- Enhanced tests to verify the correct behavior of the priority index in various scenarios, ensuring fallback to the first non-null value when necessary.
- Updated integration tests to validate the new functionality in the context of merging records.

image
image
image


etiennejouan:
Hi @anadi45, thank you so much for your contribution. 

I'm sorry to close the PR: it's not the best approach in my opinion. Ideally, the mergeMany mutation should return the properly previewed record with all the correct relations included. The frontend would then only need to display the received preview without implementing any additional logic.

This could be split into 2 PRs:
- Modify the resolver service file (packages/twenty-server/src/engine/api/graphql/graphql-query-runner/resolvers/graphql-query-merge-many-resolver.service.ts) to ensure that dryRun mode returns the previewed record with all the correct relations populated.
- In front-end, update the merge preview tab to directly display the previewed record returned from the backend, eliminating the need for additional client-side logic.

@FelixMalfait as you have lead the first implementation of mergeMany, you may have additional inputs ?

@anadi45 please, feel free to ping me on Discord if you have questions (and if you want to continue working on this issue)

PR #15243, opened 2025-10-22
chore: body-parser vulnerable to denial of service when url encoding is enabled

mabdullahabaid:
Fixes [Dependabot Alert 125](https://github.com/twentyhq/twenty/security/dependabot/125) - body-parser vulnerable to denial of service when url encoding is enabled.


charlesBochet (approved):
ty!

PR #15244, opened 2025-10-22
fix: axios requests vulnerable to possible SSRF and credential leakage via absolute URL 

mabdullahabaid:
Fixes [Dependabot Alert 211](https://github.com/twentyhq/twenty/security/dependabot/211) - axios requests vulnerable to possible SSRF and credential leakage via absolute URL.


mabdullahabaid:
Per my information, upgraded dependencies in the serverless layer would not impact the existing serverless functions - only the newer ones get impacted, so axios upgrade should not cause any issues.

FelixMalfait (approved):

PR #15248, opened 2025-10-22
Use aggregate operations in the widget configuration instead of extended aggregate operations

bosiraphael:
- Use aggregate operations in the widget configuration instead of extended aggregate operations
- Use aggregate operation from generated graphql in the frontend


ijreilly (approved):
great !

PR #15250, opened 2025-10-22
feat: self-signed certificate download for redis

EChaffraix:
I use Scaleway managed service to host redis, and they have a self signed certificate. Downloading and adding the NODE_EXTRA_CA_CERTS fix the following error : 
```
server-1  | Error: self-signed certificate; if the root CA is installed locally, try running Node.js with --use-system-ca
server-1  |     at TLSSocket.onConnectSecure (node:_tls_wrap:1631:34)
server-1  |     at TLSSocket.emit (node:events:508:28)
server-1  |     at TLSSocket._finishInit (node:_tls_wrap:1077:8)
server-1  |     at ssl.onhandshakedone (node:_tls_wrap:863:12) {
server-1  |   code: 'DEPTH_ZERO_SELF_SIGNED_CERT'
server-1  | }
```

### DIFF
--- a/packages/twenty-docker/docker-compose.yml
+++ b/packages/twenty-docker/docker-compose.yml
@@ -45,6 +45,8 @@ services:
       # EMAIL_SMTP_USER: ${EMAIL_SMTP_USER:-}
       # EMAIL_SMTP_PASSWORD: ${EMAIL_SMTP_PASSWORD:-}
 
+      # SELF_SIGNED_CERTIFICATE_URL: ${SELF_SIGNED_CERTIFICATE_URL}
+
     depends_on:
       db:
         condition: service_healthy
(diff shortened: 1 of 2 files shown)


charlesBochet:
Hi @EChaffraix, I'm not sure about this, it seems a bit heavy for this specific usecase and relying on an external url does not seems great

Would instead just adding an extra volume next to localstorage to store certs work for you?

charlesBochet:
@EChaffraix Actually, re-thinking about it, let's do like we have done for postgres so far:

Let's introduce a `REDIS_SSL_ALLOW_SELF_SIGNED` and implement the same mechanism as `PG_SSL_ALLOW_SELF_SIGNED`:
In the client instantiation:

```
process.env.PG_SSL_ALLOW_SELF_SIGNED === 'true'
      ? {
          rejectUnauthorized: false,
        }
      : undefined,
```

charlesBochet:
@EChaffraix I'm going to close this one (see my comment above), please feel free to open a new one with the strategy above!

Thanks for the help!

EChaffraix:
@charlesBochet Ok, thank you for the comment, I'll implement it with variable

ohuu:
Has this been resolved? I tried the REDIS_SSL_ALLOW_SELF_SIGNED but it doesn't seem that env var exists. Is there a work around here?

PR #15256, opened 2025-10-22
Use array fields in filter steps

### LINKED ISSUE #15247: [workflows] Filter action - display the fields of type array
## Scope & Context

**Example:** 
A customer wants to store in an array field the list of automated emails he already sent to a given person (to check before potentially sending it again). He would do this using a Filter, but the field of type array are not available in the filter nodes.

## Current behavior

Filter action do not display fields of type array

## Expected behavior
Filter action allow to filter on fields of type array

We found a workaround -- which is creating one boolean per email type - but filtering on an array could be helpful  to not have to create all those fields (this might get heavy, based on how many email templates they will need)

cc @thomtrp

thomtrp:
Fixes https://github.com/twentyhq/twenty/issues/15247

Before - arrays not available
Capture d’écran 2025-10-22 à 14 32 22

After
Capture d’écran 2025-10-22 à 14 32 03

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-field/ui/form-types/components/FormArrayFieldInput.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-field/ui/form-types/components/FormArrayFieldInput.tsx
@@ -25,6 +25,7 @@ import { isStandaloneVariableString } from '@/workflow/utils/isStandaloneVariabl
 import { useTheme } from '@emotion/react';
 import styled from '@emotion/styled';
 import { useLingui } from '@lingui/react/macro';
+import { isNonEmptyArray } from '@sniptt/guards';
 import { useId, useRef, useState } from 'react';
 import { isDefined } from 'twenty-shared/utils';
 import { IconPlus } from 'twenty-ui/display';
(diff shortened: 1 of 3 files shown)


Devessier (approved):

packages/twenty-front/src/modules/object-record/record-field/ui/form-types/components/FormArrayFieldInput.tsx
```diff
@@ -74,6 +74,24 @@ const parseSpacingValueAsNumber = (value: string) => {
...
+  if (!isDefined(defaultValue)) {
+    return [];
+  }
+
+  if (Array.isArray(defaultValue)) {
+    return defaultValue;
+  }
+
+  try {
+    return JSON.parse(defaultValue);
```
charlesBochet: this will throw so we should name functionOrThrow. If you don't want it to throw, use parseJson

packages/twenty-front/src/modules/object-record/record-field/ui/form-types/components/FormArrayFieldInput.tsx
```diff
@@ -74,6 +74,24 @@ const parseSpacingValueAsNumber = (value: string) => {
   return Number(value.replace('px', ''));
 };
 
+const getDefaultArrayValue = (
+  defaultValue: FieldArrayValue | string | undefined,
```
charlesBochet: let's cut the possibilities early

packages/twenty-front/src/modules/object-record/record-field/ui/form-types/components/FormArrayFieldInput.tsx
thomtrp: used parseJson

packages/twenty-front/src/modules/object-record/record-field/ui/form-types/components/FormArrayFieldInput.tsx
thomtrp: I did it in the Filter step. This was from where strings were sent. Filter values are stored as string even for arrays

PR #15270, opened 2025-10-22
Allow to stop running workflow

thomtrp:
https://github.com/user-attachments/assets/599154e4-8743-471b-b05a-721b635bcf4e

On stoppage:
- if no running steps, mark pending as failed and end the workflow
- if running steps, set as stopping and exit. Going to the next step, as the workflow is not running anymore, it will naturally stop

### DIFF
--- a/packages/twenty-front/src/modules/action-menu/actions/record-actions/constants/WorkflowRunsActionsConfig.tsx
+++ b/packages/twenty-front/src/modules/action-menu/actions/record-actions/constants/WorkflowRunsActionsConfig.tsx
@@ -33,6 +33,7 @@ export const WORKFLOW_RUNS_ACTIONS_CONFIG = inheritActionsFromDefaultConfig({
       availableOn: [
         ActionViewType.SHOW_PAGE,
         ActionViewType.INDEX_PAGE_SINGLE_RECORD_SELECTION,
+        ActionViewType.INDEX_PAGE_BULK_SELECTION,
       ],
       component: <StopWorkflowRunSingleRecordAction />,
     },
--- a/packages/twenty-front/src/modules/action-menu/actions/record-actions/constants/WorkflowRunsActionsConfig.tsx
+++ b/packages/twenty-front/src/modules/action-menu/actions/record-actions/constants/WorkflowRunsActionsConfig.tsx
@@ -3,21 +3,44 @@ import { NoSelectionRecordActionKeys } from '@/action-menu/actions/record-action
 import { SingleRecordActionKeys } from '@/action-menu/actions/record-actions/single-record/types/SingleRecordActionsKey';
 import { SeeVersionWorkflowRunSingleRecordAction } from '@/action-menu/actions/record-actions/single-record/workflow-run-actions/components/SeeVersionWorkflowRunSingleRecordAction';
 import { SeeWorkflowWorkflowRunSingleRecordAction } from '@/action-menu/actions/record-actions/single-record/workflow-run-actions/components/SeeWorkflowWorkflowRunSingleRecordAction';
+import { StopWorkflowRunSingleRecordAction } from '@/action-menu/actions/record-actions/single-record/workflow-run-actions/components/StopWorkflowRunSingleRecordAction';
 import { WorkflowRunSingleRecordActionKeys } from '@/action-menu/actions/record-actions/single-record/workflow-run-actions/types/WorkflowRunSingleRecordActionsKeys';
 import { inheritActionsFromDefaultConfig } from '@/action-menu/actions/record-actions/utils/inheritActionsFromDefaultConfig';
 import { ActionScope } from '@/action-menu/actions/types/ActionScope';
 import { ActionType } from '@/action-menu/actions/types/ActionType';
 import { ActionViewType } from '@/action-menu/actions/types/ActionViewType';
 import { msg } from '@lingui/core/macro';
-import { IconSettingsAutomation, IconVersions } from 'twenty-ui/display';
+import {
+  IconPlayerStop,
+  IconSettingsAutomation,
+  IconVersions,
+} from 'twenty-ui/display';
 
 export const WORKFLOW_RUNS_ACTIONS_CONFIG = inheritActionsFromDefaultConfig({
   config: {
+    [WorkflowRunSingleRecordActionKeys.STOP]: {
+      key: WorkflowRunSingleRecordActionKeys.STOP,
+      label: msg`Stop`,
+      shortLabel: msg`Stop`,
+      position: 0,
+      isPinned: true,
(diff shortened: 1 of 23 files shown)


martmull:
- Not sure about the `stopping` behavior. When clicking on stop, I just expect the workflow to stop, and not wait for a step to terminate
- I would love to see the stop button for multiple workflowRun selection in CMd+K or in quick actions 
image

image

packages/twenty-front/src/modules/action-menu/actions/record-actions/constants/WorkflowRunsActionsConfig.tsx
```diff
@@ -3,21 +3,44 @@ import { NoSelectionRecordActionKeys } from '@/action-menu/actions/record-action
...
-import { IconSettingsAutomation, IconVersions } from 'twenty-ui/display';
+import {
+  IconPlayerStop,
+  IconSettingsAutomation,
+  IconVersions,
+} from 'twenty-ui/display';
 
 export const WORKFLOW_RUNS_ACTIONS_CONFIG = inheritActionsFromDefaultConfig({
   config: {
+    [WorkflowRunSingleRecordActionKeys.STOP]: {
```
martmull, line 21: great, can we also add that action on view page when selecting multiple running workflow runs?

packages/twenty-front/src/modules/workflow/graphql/mutations/stopWorkflowRun.ts
```diff
@@ -0,0 +1,11 @@
+import { gql } from '@apollo/client';
+
+export const STOP_WORKFLOW_RUN = gql`
+  mutation StopWorkflowRun($workflowRunId: UUID!) {
+    stopWorkflowRun(workflowRunId: $workflowRunId) {
+      id
+      status
+      __typename
```
martmull, line 8: why __typename?

packages/twenty-server/src/database/commands/upgrade-version-command/1-10/1-10-add-workflow-run-stop-statuses.command.ts
```diff
@@ -0,0 +1,124 @@
...
+        color: 'orange',
+      });
+      workflowRunStatusFieldMetadataOptions?.push({
+        value: WorkflowRunStatus.STOPPED,
+        label: 'Stopped',
+        position: 6,
+        color: 'gray',
+      });
+
+      await this.fieldMetadataRepository.save(workflowRunStatusFieldMetadata);
```
martmull, line 90: is this the standard way to perform an enum update? I though this was taken care of by our schema migration system 🤔

packages/twenty-server/src/modules/workflow/common/standard-objects/workflow-run.workspace-entity.ts
```diff
@@ -40,8 +42,15 @@ export enum WorkflowRunStatus {
...
   FAILED = 'FAILED',
   ENQUEUED = 'ENQUEUED',
+  STOPPING = 'STOPPING',
+  STOPPED = 'STOPPED',
 }
 
+registerEnumType(WorkflowRunStatus, {
+  name: 'WorkflowRunStatusEnum',
+  description: 'Status of the workflow run',
+});
```
martmull, line 52: Why do we need to register an enumType now? I don't get why we didn't need it before but we need it now

packages/twenty-server/src/modules/workflow/workflow-runner/workspace-services/workflow-runner.workspace-service.ts
```diff
@@ -191,4 +196,52 @@ export class WorkflowRunnerWorkspaceService {
...
+      });
+      newStatus = WorkflowRunStatus.STOPPING;
+    } else {
+      await this.workflowRunWorkspaceService.endWorkflowRun({
+        workflowRunId,
+        workspaceId,
+        status: WorkflowRunStatus.STOPPED,
+      });
+      newStatus = WorkflowRunStatus.STOPPED;
+    }
```
martmull, line 240: I would call

```typescript
await this.workflowRunWorkspaceService.endWorkflowRun({
        workflowRunId,
        workspaceId,
        status: WorkflowRunStatus.STOPPED,
      });
```

in both cases

packages/twenty-front/src/modules/workflow/graphql/mutations/stopWorkflowRun.ts
thomtrp: so it updates apollo cache

packages/twenty-server/src/database/commands/upgrade-version-command/1-10/1-10-add-workflow-run-stop-statuses.command.ts
thomtrp: Nope, unfortunately we still need to perform the migration ourselves. Did a double check with @Weiko

packages/twenty-front/src/modules/action-menu/actions/record-actions/constants/WorkflowRunsActionsConfig.tsx
thomtrp: good point

packages/twenty-server/src/modules/workflow/workflow-runner/workspace-services/workflow-runner.workspace-service.ts
thomtrp: you need to cut the running states first. And in my case, I let these cut by themselves. We need to discuss the global strategy for this

packages/twenty-server/src/database/commands/upgrade-version-command/1-10/1-10-add-workflow-run-stop-statuses.command.ts
martmull: ok

packages/twenty-server/src/modules/workflow/workflow-runner/workspace-services/workflow-runner.workspace-service.ts
martmull: yes, if we keep stopping status we can resolve this comment

packages/twenty-server/src/modules/workflow/common/standard-objects/workflow-run.workspace-entity.ts
thomtrp: because we return it in DTO. Apollo needs to understand we are talking about workflow run status when it does the matching in cache

martmull (approved):
multiple runs stop would be nice but lgtm

PR #15276, opened 2025-10-23
Add Role tab to agent detail page and fix restricted fields permission issue

abdulrahmancodes:


packages/twenty-front/src/pages/settings/ai/components/SettingsAgentRoleTab.tsx
```diff
@@ -0,0 +1,118 @@
...
+  margin-bottom: ${({ theme }) => theme.spacing(8)};
+`;
+
+const StyledRoleSelector = styled.div`
+  flex: 1;
+`;
+
+type SettingsAgentRoleTabProps = {
+  formValues: SettingsAIAgentFormValues;
+  onFieldChange: (
```
charlesBochet, line 35: not a big deal but we usually avoid this. the field change can be handled from the child directly, that's fine as long as we don't drill down too much

packages/twenty-front/src/pages/settings/ai/components/SettingsAgentSettingsTab.tsx
```diff
@@ -0,0 +1,174 @@
...
+  color: ${({ theme }) => theme.color.red};
+  font-size: ${({ theme }) => theme.font.size.sm};
+  margin-top: ${({ theme }) => theme.spacing(1)};
+`;
+
+const DELETE_AGENT_MODAL_ID = 'delete-agent-modal';
+
+type SettingsAgentSettingsTabProps = {
+  formValues: SettingsAIAgentFormValues;
+  onFieldChange: (
```
charlesBochet, line 46: same here

packages/twenty-front/src/pages/settings/ai/components/SettingsAgentRoleTab.tsx
charlesBochet: I've read the whole code and I think this is acceptable here

packages/twenty-server/src/engine/api/utils/get-all-updatable-fields.util.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { type ObjectMetadataItemWithFieldMaps } from 'src/engine/metadata-modules/types/object-metadata-item-with-field-maps';
+import { getFieldMetadataIdToColumnNamesMap } from 'src/engine/twenty-orm/utils/get-field-metadata-id-to-column-names-map.util';
+
+type GetAllUpdatableFieldsArgs = {
+  restrictedFields: RestrictedFieldsPermissions;
+  objectMetadata: { objectMetadataMapItem: ObjectMetadataItemWithFieldMaps };
+};
+
+export const getAllUpdatableFields = ({
```
charlesBochet: fileName should be .util.ts (not .utils.ts)

charlesBochet: this file is probably duplicated from getAllSelectableFields, LGTM as it makes naming super clear but could be refactored together

packages/twenty-server/src/engine/core-modules/record-crud/services/create-record.service.ts
```diff
@@ -83,14 +86,61 @@ export class CreateRecordService {
           objectMetadataMapItem: objectMetadataItemWithFieldsMaps,
         });
 
-      const insertResult = await repository.insert({
+      let selectedColumns = undefined;
```
charlesBochet: **question:** is this code also used by workflow? If yes, have you QA workflow to make sure we don't introduce a regression?

packages/twenty-server/src/engine/core-modules/record-crud/services/create-record.service.ts
```diff
@@ -83,14 +86,61 @@ export class CreateRecordService {
...
         ...transformedObjectRecord,
         position,
         createdBy: params.createdBy ?? {
           source: FieldActorSource.WORKFLOW,
           name: 'Workflow',
         },
-      });
+      };
+
+      const restrictedFields =
```
charlesBochet, line 87: same here

packages/twenty-server/src/engine/core-modules/record-crud/services/find-records.service.ts
```diff
@@ -1,5 +1,6 @@
 import { Injectable, Logger } from '@nestjs/common';
 
+import isEmpty from 'lodash.isempty';
```
charlesBochet, line 3: @ijreilly could you take a look at the permission logic here to see if you are aligned with it?

packages/twenty-server/src/engine/core-modules/record-crud/services/create-record.service.ts
```diff
@@ -83,14 +86,61 @@ export class CreateRecordService {
...
+          },
+        });
+
+        const filteredData = Object.fromEntries(
+          Object.entries(transformedObjectRecord).filter(
+            ([key]) => updatableFields[key] !== false,
+          ),
+        );
+
+        dataToInsert = {
```
charlesBochet: @FelixMalfait @abdulrahmancodes discussed this @ijreilly, this looks fine but from a product perspective, I wonder if we should rather let the record creation fail if the agent / workflow node is trying to update a field it cannot update (the agent could retry with a smaller set of field but would be aware). Here it will think that it succeeds to update all fields but it will actually only update the subset it has permission to

packages/twenty-server/src/engine/core-modules/record-crud/services/find-records.service.ts
ijreilly: discussed live

charlesBochet (commented):
LGTM from a code perspective, I have left comment + question.

Leaving the stamp to @FelixMalfait

packages/twenty-server/src/engine/core-modules/record-crud/services/create-record.service.ts
abdulrahmancodes: Yes, it’s used by workflows as well, and I’ve tested it to make sure everything works fine there, too.

FelixMalfait:
Screenshot 2025-10-24 at 10 46 58
Screenshot 2025-10-24 at 10 45 09
Looks like it still needs to be fixed for update/create. Find now works well!

packages/twenty-server/src/engine/core-modules/record-crud/services/upsert-record.service.ts
```diff
@@ -135,13 +150,33 @@ export class UpsertRecordService {
         )
         .filter(isDefined);
 
-      const upsertResult = await repository.upsert(transformedObjectRecord, {
```
abdulrahmancodes, line 138: Not related to this PR, but this also fixes an issue where updating a record wasn’t working.

packages/twenty-server/src/engine/core-modules/record-crud/services/upsert-record.service.ts
```diff
@@ -73,16 +75,29 @@ export class UpsertRecordService {
...
-          if (fieldsToUpdateArray.includes(key)) {
-            return {
-              ...acc,
-              [key]: objectRecord[key],
-            };
+        (acc, fieldName) => {
+          if (!fieldsToUpdateArray.includes(fieldName)) {
+            return acc;
+          }
+
```
FelixMalfait, line 85: I feel like we should have solved the permission issue in generateFindToolInputSchema and not here, no? Here you kind of make the AI think that it's possible but then it's not. We should just not expose what is not possible?

FelixMalfait: I'm not sure it's worth keeping it here, is it a good idea to swallow errors?

FelixMalfait: I think you could argue in some case you want it to go through and in some case you want it to fail. 
But why do this in Workflow/AI Tools and not in REST/GraphQL for exemple? It feels like we should keep a similar behavior and fix the root cause which is the AI doing incorrect tool calls

packages/twenty-server/src/engine/core-modules/record-crud/zod-schemas/record-input.zod-schema.ts
```diff
@@ -5,19 +5,29 @@ import { type ObjectMetadataEntity } from 'src/engine/metadata-modules/object-me
 
 export const generateRecordInputSchema = (
   objectMetadata: ObjectMetadataEntity,
+  includeId = false,
```
FelixMalfait: If we go with proper tool definition we'll probably need to split this into two functions for update/create (but which will share common logic). The includeId felt a bit hacky already

packages/twenty-server/src/engine/api/utils/get-all-updatable-fields.util.ts
FelixMalfait: Where is this used?

FelixMalfait (approved):
Great, thank you!

FelixMalfait:
I'll create a followup issue for workflows because we probably need to be smarter in the action definition (here linkedin url is left empty so shouldn't be taken into account but yet the action fails). I guess this already happens on main because I only see changes that go in the right direction in your PR

Screenshot 2025-10-26 at 09 16 32

FelixMalfait:
Forget my previous message, I didn't restart the worker, your PR actually fixes this issue.

PR #15279, opened 2025-10-23
Common - Update one/many, restore one/many, findDuplicates & mergeMany

### LINKED ISSUE #1739: Common - Update, Restore, Duplicates, Merge

etiennejouan:
closes https://github.com/twentyhq/core-team-issues/issues/1739

### DIFF
--- a/packages/twenty-server/src/engine/api/common/common-query-runners/common-find-duplicates-query-runner.service.ts
+++ b/packages/twenty-server/src/engine/api/common/common-query-runners/common-find-duplicates-query-runner.service.ts
@@ -19,6 +19,7 @@ import { CommonFindDuplicatesOutputItem } from 'src/engine/api/common/types/comm
 import {
   CommonExtendedInput,
   CommonInput,
+  CommonQueryNames,
   FindDuplicatesQueryArgs,
 } from 'src/engine/api/common/types/common-query-args.type';
 import { getPageInfo } from 'src/engine/api/common/utils/get-page-info.util';
@@ -31,6 +32,8 @@ export class CommonFindDuplicatesQueryRunnerService extends CommonBaseQueryRunne
   FindDuplicatesQueryArgs,
   CommonFindDuplicatesOutputItem[]
 > {
+  protected readonly operationName = CommonQueryNames.FIND_DUPLICATES;
+
   async run(
     args: CommonExtendedInput<FindDuplicatesQueryArgs>,
     queryRunnerContext: CommonExtendedQueryRunnerContext,
--- a/packages/twenty-server/src/engine/api/common/common-query-runners/common-find-duplicates-query-runner.service.ts
+++ b/packages/twenty-server/src/engine/api/common/common-query-runners/common-find-duplicates-query-runner.service.ts
@@ -1,0 +1,205 @@
+import { Injectable } from '@nestjs/common';
+
+import isEmpty from 'lodash.isempty';
+import { QUERY_MAX_RECORDS } from 'twenty-shared/constants';
+import { ObjectRecord, OrderByDirection } from 'twenty-shared/types';
+import { isDefined } from 'twenty-shared/utils';
+import { In } from 'typeorm';
+
+import { WorkspaceAuthContext } from 'src/engine/api/common/interfaces/workspace-auth-context.interface';
+
+import { CommonBaseQueryRunnerService } from 'src/engine/api/common/common-query-runners/common-base-query-runner.service';
+import {
+  CommonQueryRunnerException,
+  CommonQueryRunnerExceptionCode,
+} from 'src/engine/api/common/common-query-runners/errors/common-query-runner.exception';
+import { CommonBaseQueryRunnerContext } from 'src/engine/api/common/types/common-base-query-runner-context.type';
+import { CommonExtendedQueryRunnerContext } from 'src/engine/api/common/types/common-extended-query-runner-context.type';
+import { CommonFindDuplicatesOutputItem } from 'src/engine/api/common/types/common-find-duplicates-output-item.type';
+import {
+  CommonExtendedInput,
+  CommonInput,
+  FindDuplicatesQueryArgs,
+} from 'src/engine/api/common/types/common-query-args.type';
+import { getPageInfo } from 'src/engine/api/common/utils/get-page-info.util';
+import { buildColumnsToSelect } from 'src/engine/api/graphql/graphql-query-runner/utils/build-columns-to-select';
+import { buildDuplicateConditions } from 'src/engine/api/utils/build-duplicate-conditions.utils';
+import { ObjectMetadataMaps } from 'src/engine/metadata-modules/types/object-metadata-maps';
+
+@Injectable()
+export class CommonFindDuplicatesQueryRunnerService extends CommonBaseQueryRunnerService<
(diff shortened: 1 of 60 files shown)


packages/twenty-server/src/engine/api/common/common-query-runners/common-find-duplicates-query-runner.service.ts
```diff
@@ -0,0 +1,205 @@
...
+
+          const duplicates = (await duplicateRecordsQueryBuilder
+            .setFindOptions({
+              select: columnsToSelect,
+            })
+            .take(QUERY_MAX_RECORDS)
+            .getMany()) as ObjectRecord[];
+
+          const aggregateQueryBuilder = duplicateRecordsQueryBuilder.clone();
+          const totalCount = await aggregateQueryBuilder.getCount();
```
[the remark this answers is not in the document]
etiennejouan: yes

packages/twenty-server/src/engine/api/common/common-query-runners/common-find-duplicates-query-runner.service.ts
```diff
@@ -0,0 +1,208 @@
...
+            duplicateRecordsQueryBuilder,
+            objectMetadataItemWithFieldMaps.nameSingular,
+            duplicateConditions,
+          );
+
+          const duplicates = (await duplicateRecordsQueryBuilder
+            .setFindOptions({
+              select: columnsToSelect,
+            })
+            .take(QUERY_MAX_RECORDS)
```
ijreilly, line 106: This is new right? out of curiosity did you add it because it should actually be on every queries?

packages/twenty-server/src/engine/api/common/common-query-runners/common-find-duplicates-query-runner.service.ts
```diff
@@ -0,0 +1,208 @@
...
+            .getMany()) as ObjectRecord[];
+
+          const aggregateQueryBuilder = duplicateRecordsQueryBuilder.clone();
```
ijreilly, line 109: I guess we do this to get the actual count of duplicates without being limited by the QUERY_MAX_RECORDS. But is this totalCount useful? Or are we adding it because it's expected in the response? 
it makes sense though

packages/twenty-server/src/engine/api/common/common-query-runners/common-merge-many-query-runner.service.ts
```diff
@@ -0,0 +1,455 @@
...
+    );
+
+    const queryBuilder = queryRunnerContext.repository.createQueryBuilder(
+      objectMetadataItemWithFieldMaps.nameSingular,
+    );
+
+    await queryBuilder
+      .softDelete()
+      .whereInIds(idsToDelete)
+      .returning('*')
```
ijreilly: Not on you this PR but this is incompatible with field-level permissions since we try to return everything:
Capture d’écran 2025-10-24 à 10 26 02

so this is a bug. if you can't tackle it now (there s no rush to), let's create a high bug, you can assign it to me.

packages/twenty-server/src/engine/api/common/common-query-runners/common-merge-many-query-runner.service.ts
```diff
@@ -0,0 +1,455 @@
...
+      .execute();
+
+    const updatedRecord = await this.updatePriorityRecord(
+      args,
+      queryRunnerContext,
+      priorityRecord.id,
+      mergedData,
+    );
+
+    await this.processNestedRelations({
```
ijreilly, line 119: I see in graphql merge many resolver that 5 days ago @abdulrahmancodes added made this execution conditional to 
```
if (executionArgs.rolePermissionConfig) {
```

Idk why though but worth digging

packages/twenty-server/src/engine/api/common/types/common-find-duplicates-output-item.type.ts
```diff
@@ -0,0 +1,10 @@
+import { type ObjectRecord } from 'twenty-shared/types';
+
+export type CommonFindDuplicatesOutputItem = {
+  records: ObjectRecord[];
+  totalCount: number;
```
ijreilly, line 5: as mentioned I think we need to clarify what is expected here - the total count of duplicates through all the codebase (which does make sense), or the total count of records returned in the result

packages/twenty-server/src/engine/api/rest/core/query-builder/utils/path-parsers/parse-core-path.utils.ts
```diff
@@ -32,10 +35,27 @@ export const parseCorePath = (
     return { object: queryAction[1] };
   }
 
-  if (queryAction[1] === 'duplicates' || queryAction[1] === 'group') {
+  if (
+    queryAction[1] === 'duplicates' ||
+    queryAction[1] === 'group' ||
```
ijreilly, line 40: groupBy

But im correcting that in my next pr anyway

ijreilly (approved):
Looks good ! 

as discussed to be able to remove FF next, we will need to delete all graphql resolver files, to make sure no one keeps editing them, and also to make sure we are up to date with the latests changes on all resolvers before doing so. As I mentioned in a comment, i spotted a recent edit (5 days ago) that was not included in your pr, but i dont know if its because you started it before the change was merged or if its because we dont need it anymore. But we need to be careful not to overlook any recent change

packages/twenty-server/src/engine/api/common/common-query-runners/common-find-duplicates-query-runner.service.ts
etiennejouan: Yes, it's new. I add it after discussing of it with Charles. In front we only show 5 duplicates but the resolver returns all duplicates without limit. it's not game changing as it's not a resolver widely used I think, but it was not very costly to set this limit.

packages/twenty-server/src/engine/api/common/common-query-runners/common-find-duplicates-query-runner.service.ts
etiennejouan: Yes, exactly. I add it to keep the same response structure.

packages/twenty-server/src/engine/api/common/types/common-find-duplicates-output-item.type.ts
etiennejouan: In findMany, totalCount returns totalCount within all the tables, not the returned/paginated ones. I think we should do the same.

packages/twenty-server/src/engine/api/common/common-query-runners/common-merge-many-query-runner.service.ts
etiennejouan: I'll contact him

etiennejouan: After reading the [PR](https://github.com/twentyhq/twenty/pull/15150/files#diff-2e14aeb6d5d3bfa60e7fffee9761963bcd4f534fbbcb8ab60f2645ce3b9bd0dfR98) introducing the condition, it seems to be a check for roleId definition. With the new typing introduced, roleId is always defined.

PR #15283, opened 2025-10-23
Encrypt/decrypt app variables

Weiko:
## Context
This PR introduces a new SecretEncryption service. This can be used to encrypt application variables before storing them in the DB and decrypt them later when executed in serverless functions. We are also adding a mask before returning them through the API

Screenshot 2025-10-23 at 15 30 22

### DIFF
--- a/packages/twenty-server/src/app.module.ts
+++ b/packages/twenty-server/src/app.module.ts
@@ -4,7 +4,6 @@ import {
   Module,
   RequestMethod,
 } from '@nestjs/common';
-import { ConfigModule } from '@nestjs/config';
 import { GraphQLModule } from '@nestjs/graphql';
 import { ServeStaticModule } from '@nestjs/serve-static';
 
@@ -47,10 +46,6 @@ const MIGRATED_REST_METHODS = [
 @Module({
   imports: [
     SentryModule.forRoot(),
-    ConfigModule.forRoot({
-      isGlobal: true,
-      envFilePath: process.env.NODE_ENV === 'test' ? '.env.test' : '.env',
-    }),
     GraphQLModule.forRootAsync<YogaDriverConfig>({
       driver: YogaDriver,
       imports: [GraphQLConfigModule, MetricsModule, DataloaderModule],
--- a/packages/twenty-server/src/engine/core-modules/application/application-sync.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/application/application-sync.service.ts
@@ -4,18 +4,18 @@ import { isDefined } from 'twenty-shared/utils';
 
 import { ApplicationEntity } from 'src/engine/core-modules/application/application.entity';
 import {
-  ApplicationException,
-  ApplicationExceptionCode,
+    ApplicationException,
+    ApplicationExceptionCode,
 } from 'src/engine/core-modules/application/application.exception';
 import { ApplicationService } from 'src/engine/core-modules/application/application.service';
 import { ApplicationInput } from 'src/engine/core-modules/application/dtos/application.input';
 import {
-  AgentManifest,
-  ObjectManifest,
-  ServerlessFunctionManifest,
-  ServerlessFunctionTriggerManifest,
+    AgentManifest,
+    ObjectManifest,
+    ServerlessFunctionManifest,
+    ServerlessFunctionTriggerManifest,
 } from 'src/engine/core-modules/application/types/application.types';
-import { ApplicationVariableEntityService } from 'src/engine/core-modules/applicationVariable/application-variable.service';
+import { ApplicationVariableService } from 'src/engine/core-modules/applicationVariable/application-variable.service';
 import { AgentService } from 'src/engine/metadata-modules/agent/agent.service';
 import { CronTriggerV2Service } from 'src/engine/metadata-modules/cron-trigger/services/cron-trigger-v2.service';
 import { FlatCronTrigger } from 'src/engine/metadata-modules/cron-trigger/types/flat-cron-trigger.type';
@@ -42,7 +42,7 @@ export class ApplicationSyncService {
 
   constructor(
     private readonly applicationService: ApplicationService,
-    private readonly applicationVariableService: ApplicationVariableEntityService,
+    private readonly applicationVariableService: ApplicationVariableService,
     private readonly serverlessFunctionLayerService: ServerlessFunctionLayerService,
     private readonly objectMetadataServiceV2: ObjectMetadataServiceV2,
     private readonly serverlessFunctionV2Service: ServerlessFunctionV2Service,
--- a/packages/twenty-server/src/engine/core-modules/application/application-sync.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/application/application-sync.service.ts
@@ -4,16 +4,16 @@ import { isDefined } from 'twenty-shared/utils';
 
 import { ApplicationEntity } from 'src/engine/core-modules/application/application.entity';
 import {
-    ApplicationException,
-    ApplicationExceptionCode,
+  ApplicationException,
+  ApplicationExceptionCode,
 } from 'src/engine/core-modules/application/application.exception';
 import { ApplicationService } from 'src/engine/core-modules/application/application.service';
 import { ApplicationInput } from 'src/engine/core-modules/application/dtos/application.input';
 import {
-    AgentManifest,
-    ObjectManifest,
-    ServerlessFunctionManifest,
-    ServerlessFunctionTriggerManifest,
+  AgentManifest,
+  ObjectManifest,
+  ServerlessFunctionManifest,
+  ServerlessFunctionTriggerManifest,
 } from 'src/engine/core-modules/application/types/application.types';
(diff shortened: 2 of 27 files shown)


packages/twenty-server/src/engine/core-modules/application/application.service.ts
```diff
@@ -32,6 +58,8 @@ export class ApplicationService {
         'applicationVariables',
       ],
     });
+
+    return this.decryptManyApplicationVariables(applications);
```
charlesBochet: I don't like the naming: decryptManyApplicationVariables should take variables as parameter. Here it's more decryptVariablesWithinApplication

packages/twenty-server/src/engine/core-modules/applicationVariable/application-variable.service.ts
```diff
@@ -1,28 +1,61 @@
...
 
+@Injectable()
 export class ApplicationVariableEntityService {
   constructor(
     @InjectRepository(ApplicationVariableEntity)
     private readonly applicationVariableRepository: Repository<ApplicationVariableEntity>,
+    private readonly secretEncryptionService: SecretEncryptionService,
   ) {}
 
+  private encryptSecretValue(value: string, isSecret: boolean): string {
```
charlesBochet, line 23: **change:** I feel we should have application variables always encrypted. So this service becomes a repository and we don't need udpate / find methods

this service should only handle appVaraibles encryption

packages/twenty-server/src/engine/core-modules/environment/environment.module.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import {
+    ConfigVariables,
+    validate,
+} from 'src/engine/core-modules/twenty-config/config-variables';
+import { CONFIG_VARIABLES_INSTANCE_TOKEN } from 'src/engine/core-modules/twenty-config/constants/config-variables-instance-tokens.constants';
+import { EnvironmentConfigDriver } from 'src/engine/core-modules/twenty-config/drivers/environment-config.driver';
+
+@Global()
+@Module({
```
charlesBochet, line 12: remove?

packages/twenty-server/src/engine/core-modules/secret-encryption/secret-encryption.service.ts
```diff
@@ -0,0 +1,64 @@
...
+    return this.environmentConfigDriver.get('APP_SECRET');
+  }
+
+  encrypt(value: string): string {
+    if (!value) {
+      return value;
+    }
+
+    try {
+      const appSecret = this.getAppSecret();
```
charlesBochet, line 29: append something here to avoid bruteforce attack against appKey

packages/twenty-server/src/engine/core-modules/environment/environment.module.ts
Weiko: EnvironmentConfigDriver was provided in packages/twenty-server/src/engine/core-modules/twenty-config/twenty-config.module.ts which causes a circular dependency

packages/twenty-server/src/app.module.ts
```diff
@@ -47,10 +46,6 @@ const MIGRATED_REST_METHODS = [
 @Module({
   imports: [
     SentryModule.forRoot(),
-    ConfigModule.forRoot({
```
Weiko, line 50: moved to global environment module

packages/twenty-server/src/engine/core-modules/application/application-sync.service.ts
```diff
@@ -42,7 +42,7 @@ export class ApplicationSyncService {
 
   constructor(
     private readonly applicationService: ApplicationService,
-    private readonly applicationVariableService: ApplicationVariableEntityService,
```
Weiko, line 45: renaming to be more aligned with twenty guidelines

packages/twenty-server/src/engine/core-modules/application/application.service.ts
```diff
@@ -23,7 +23,7 @@ export class ApplicationService {
   async findManyApplications(
     workspaceId: string,
   ): Promise<ApplicationEntity[]> {
-    return this.applicationRepository.find({
+    return await this.applicationRepository.find({
```
Weiko, line 26: not needed, forgot to remove, not super important though

packages/twenty-server/src/engine/core-modules/applicationVariable/application-variable.entity.ts
```diff
@@ -10,24 +8,17 @@ import {
...
 import { Relation } from 'src/engine/workspace-manager/workspace-sync-metadata/interfaces/relation.interface';
 
-import { UUIDScalarType } from 'src/engine/api/graphql/workspace-schema-builder/graphql-types/scalars';
 import { ApplicationEntity } from 'src/engine/core-modules/application/application.entity';
 
-@Entity({
-  name: 'applicationVariable',
-  schema: 'core',
-})
-@ObjectType('ApplicationVariable')
```
Weiko, line 24: not needed on an entity, we already have a dedicated dto

packages/twenty-server/src/engine/core-modules/applicationVariable/application-variable.entity.ts
```diff
@@ -10,24 +8,17 @@ import {
...
+@Entity('applicationVariable')
 @Unique('IDX_APPLICATION_VARIABLE_KEY_APPLICATION_ID_UNIQUE', [
   'key',
   'applicationId',
 ])
 export class ApplicationVariableEntity {
-  @IDField(() => UUIDScalarType)
```
Weiko, line 30: same

packages/twenty-server/src/engine/core-modules/applicationVariable/application-variable.resolver.ts
```diff
@@ -1,22 +1,40 @@
...
+export class ApplicationVariableResolver {
   constructor(
-    private readonly applicationVariableService: ApplicationVariableEntityService,
+    private readonly applicationVariableService: ApplicationVariableService,
   ) {}
 
+  @ResolveField(() => String)
+  value(@Parent() applicationVariable: ApplicationVariableEntity): string {
+    if (applicationVariable.isSecret) {
+      return SECRET_APPLICATION_VARIABLE_MASK;
```
Weiko, line 29: Var is already encrypted. This is just a mask for extra security and because it does not make sense to return the encrypted string anyway

packages/twenty-server/src/engine/core-modules/twenty-config/twenty-config.module.ts
```diff
@@ -18,14 +13,7 @@ export class TwentyConfigModule extends ConfigurableModuleClass {
...
 
-    const imports = [
-      ConfigModule.forRoot({
-        isGlobal: true,
-        expandVariables: true,
-        validate,
-        envFilePath: process.env.NODE_ENV === 'test' ? '.env.test' : '.env',
-      }),
-    ];
+    const imports = [];
```
Weiko, line 16: this was duplicated in multiple places, now available through global environment module

packages/twenty-server/src/engine/core-modules/applicationVariable/application-variable.entity.ts
```diff
@@ -10,24 +8,17 @@ import {
...
 
-import { UUIDScalarType } from 'src/engine/api/graphql/workspace-schema-builder/graphql-types/scalars';
 import { ApplicationEntity } from 'src/engine/core-modules/application/application.entity';
 
-@Entity({
-  name: 'applicationVariable',
-  schema: 'core',
-})
-@ObjectType('ApplicationVariable')
+@Entity('applicationVariable')
```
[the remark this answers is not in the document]
Weiko: not needed, the global datasource contains a unique schema (see object-metadata.entity as example)

packages/twenty-server/src/engine/core-modules/application/application.service.ts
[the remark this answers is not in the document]
Weiko: removed

packages/twenty-server/src/engine/core-modules/applicationVariable/application-variable.resolver.ts
martmull, line 29: I think we should return some caracteres of the value, like return something like

```typescript
const N = 5
value = secret.slice(0, Math.min(N, Math.floor(secret/10))) + '*'.repeat(N);
```

Image

Weiko: The value is returned encrypted so it won't be really interesting to return the last n chars

martmull: can't we apply this formatting on decrypt(secret)?

Weiko: Ok

Weiko:
Reopening tmr

PR #15285, opened 2025-10-23
`FindAllCoreViews` graphql cache operation invalidation in view related v2 action run

prastoin:
# Introduction
Adding a view field to a view in v2 would be optimistically rendered by the front but on refresh would not get persisted.
That's because we cache both:
```ts
      useCachedMetadata({
        cacheGetter: cacheStorageService.get.bind(cacheStorageService),
        cacheSetter: cacheStorageService.set.bind(cacheStorageService),
        operationsToCache: ['ObjectMetadataItems', 'FindAllCoreViews'],
      }),
```
With keys that look like:
```ts
    return `graphql:operations:${operationName}:${workspace.id}:${workspaceMetadataVersion}:${locale}:${queryHash}`;
```

It was functional in v1 as we would be incrementing metadata version often.

In v2 it gets incremented only if implies an interaction to metadata object or fields ( will be deprecated in the future though, until we finish the // run )

The fix was to check if an `view` or related has been processed in the workspace migration or if we incremented the metadata in order to manually flush the `findAllCoreViews` redis cache.

## Conclusion
Related to https://github.com/twentyhq/core-team-issues/issues/1753

### DIFF
--- a/packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-runner-v2/services/workspace-migration-runner-v2.service.ts
+++ b/packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-runner-v2/services/workspace-migration-runner-v2.service.ts
@@ -64,7 +64,7 @@ export class WorkspaceMigrationRunnerV2Service {
       );
     }
 
-    const shouldInvalidFindCoreViewsGraphqlChacheOperation = actions.some(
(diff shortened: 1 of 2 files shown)


prastoin:
> Score reflects that the logic is sound and follows established patterns, but there's potential concern around the hardcoded action type checks which could become maintenance burden as new view-related action types are added

Yeah you're right, this legacy cache invalidation will be removed after v2 official release

charlesBochet (approved):

PR #15286, opened 2025-10-23
Fix defaultValue with enum options update in migration v2

Weiko:
## Context
When updating both enum options and defaultValue, the old default might not be in the new options (or vice versa), causing PostgreSQL constraint violations regardless of update order.

## Solution
Sort updates to process defaultValue last; before updating options, temporarily set the new defaultValue in metadata so alterEnumValues creates the column with the correct default, then skip the redundant defaultValue update handler.

### DIFF
--- a/packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-runner-v2/action-handlers/field/services/update-field-action-handler.service.ts
+++ b/packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-runner-v2/action-handlers/field/services/update-field-action-handler.service.ts
@@ -170,6 +170,20 @@ export class UpdateFieldActionHandlerService extends WorkspaceMigrationRunnerAct
         });
         optimisticFlatFieldMetadata.name = update.to;
       }
+      if (isPropertyUpdate(update, 'defaultValue')) {
+        if (wasDefaultValueHandledByEnumUpdate) {
+          optimisticFlatFieldMetadata.defaultValue = update.to;
+        } else {
+          await this.handleFieldDefaultValueUpdate({
+            queryRunner,
+            schemaName,
+            tableName,
+            flatFieldMetadata: optimisticFlatFieldMetadata,
+            update,
+          });
+          optimisticFlatFieldMetadata.defaultValue = update.to;
+        }
+      }
       if (
         isPropertyUpdate(update, 'options') &&
         isEnumFlatFieldMetadata(optimisticFlatFieldMetadata)
@@ -191,20 +205,6 @@ export class UpdateFieldActionHandlerService extends WorkspaceMigrationRunnerAct
           update,
         });
         optimisticFlatFieldMetadata.options = update.to ?? [];
-      }
-      if (isPropertyUpdate(update, 'defaultValue')) {
-        if (wasDefaultValueHandledByEnumUpdate) {
-          optimisticFlatFieldMetadata.defaultValue = update.to;
-        } else {
-          await this.handleFieldDefaultValueUpdate({
-            queryRunner,
-            schemaName,
-            tableName,
-            flatFieldMetadata: optimisticFlatFieldMetadata,
-            update,
-          });
-          optimisticFlatFieldMetadata.defaultValue = update.to;
-        }
       }
       if (
         isMorphOrRelationFlatFieldMetadata(optimisticFlatFieldMetadata) &&
--- a/packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-runner-v2/action-handlers/field/services/update-field-action-handler.service.ts
+++ b/packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-runner-v2/action-handlers/field/services/update-field-action-handler.service.ts
@@ -27,6 +27,7 @@ import { fieldMetadataTypeToColumnType } from 'src/engine/metadata-modules/works
 import { WorkspaceSchemaManagerService } from 'src/engine/twenty-orm/workspace-schema-manager/workspace-schema-manager.service';
 import { isMorphOrRelationFieldMetadataType } from 'src/engine/utils/is-morph-or-relation-field-metadata-type.util';
 import { PropertyUpdate } from 'src/engine/workspace-manager/workspace-migration-v2/types/property-update.type';
+import { findFlatEntityPropertyUpdate } from 'src/engine/workspace-manager/workspace-migration-v2/utils/find-flat-entity-property-update.util';
 import { isPropertyUpdate } from 'src/engine/workspace-manager/workspace-migration-v2/utils/is-property-update.util';
 import { type UpdateFieldAction } from 'src/engine/workspace-manager/workspace-migration-v2/workspace-migration-builder-v2/builders/field/types/workspace-migration-field-action-v2';
 import { serializeDefaultValueV2 } from 'src/engine/workspace-manager/workspace-migration-v2/workspace-migration-builder-v2/utils/serialize-default-value-v2.util';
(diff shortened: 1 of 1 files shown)


packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-runner-v2/action-handlers/field/services/update-field-action-handler.service.ts
```diff
@@ -159,19 +171,31 @@ export class UpdateFieldActionHandlerService extends WorkspaceMigrationRunnerAct
...
-        await this.handleFieldDefaultValueUpdate({
-          queryRunner,
-          schemaName,
-          tableName,
-          flatFieldMetadata: optimisticFlatFieldMetadata,
-          update,
-        });
-        optimisticFlatFieldMetadata.defaultValue = update.to;
+        if (wasDefaultValueHandledByEnumUpdate) {
+          optimisticFlatFieldMetadata.defaultValue = update.to;
```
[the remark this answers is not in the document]
Weiko: It is done later in the enum update because we always re-create a new column including the columnDefinition based on the optimisticFlatFieldMetadata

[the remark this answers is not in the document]
Weiko: It is done later in the enum update because we always re-create a new column including the columnDefinition based on the optimisticFlatFieldMetadata

packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-runner-v2/action-handlers/field/services/update-field-action-handler.service.ts
```diff
@@ -147,7 +147,19 @@ export class UpdateFieldActionHandlerService extends WorkspaceMigrationRunnerAct
...
+      (u) => u.property === 'defaultValue',
+    );
+    const hasDefaultValueUpdate = isDefined(defaultValueUpdate);
+
+    let wasDefaultValueHandledByEnumUpdate = false;
+
+    const sortedUpdatesWithDefaultValuesUpdateLast = [...updates].sort(
+      (a, b) =>
+        +(a.property === 'defaultValue') - +(b.property === 'defaultValue'),
+    );
```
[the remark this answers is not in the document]
Weiko: No?

packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-runner-v2/action-handlers/field/services/update-field-action-handler.service.ts
```diff
@@ -147,7 +147,19 @@ export class UpdateFieldActionHandlerService extends WorkspaceMigrationRunnerAct
 
     let optimisticFlatFieldMetadata = structuredClone(currentFlatFieldMetadata);
 
-    for (const update of updates) {
+    const defaultValueUpdate = updates.find(
```
charlesBochet: let's add an integration test

prastoin (dismissed):
Hey there ! Thanks about that !

prastoin:
~~UPDATE: This should already be covered by~~
```ts
// packages/twenty-server/test/integration/metadata/suites/field-metadata/enum/successful-update-one-enum-field-metadata-v2.integration-spec.ts
    it('Should update default value to null even if it was set before', async () => {
```

~~I don't understand why this wasn't failing before~~
=> ~~Might be interesting to dig deeper~~

Nevermind this whole test suite does not cover updating the default value option value but the default value itself ( 2nd degree )

packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-runner-v2/action-handlers/field/services/update-field-action-handler.service.ts
```diff
@@ -147,7 +147,19 @@ export class UpdateFieldActionHandlerService extends WorkspaceMigrationRunnerAct
...
+      (u) => u.property === 'defaultValue',
```
prastoin: ***Polish:*** naming

packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-runner-v2/action-handlers/field/services/update-field-action-handler.service.ts
prastoin: ***Suggestion:*** Could use `findFlatEntityPropertyUpdate` for smooth typing

packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-runner-v2/action-handlers/field/services/update-field-action-handler.service.ts
```diff
@@ -147,7 +147,19 @@ export class UpdateFieldActionHandlerService extends WorkspaceMigrationRunnerAct
...
+    );
+    const hasDefaultValueUpdate = isDefined(defaultValueUpdate);
+
+    let wasDefaultValueHandledByEnumUpdate = false;
+
+    const sortedUpdatesWithDefaultValuesUpdateLast = [...updates].sort(
```
prastoin, line 159: ***Question:*** Idk what's the most optimal nor human readable here, either filtering the default value out and pushing it in the array as last element or sorting

prastoin (commented):
Technical review !

packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-runner-v2/action-handlers/field/services/update-field-action-handler.service.ts
martmull, line 162: the code is not very easy to understand, can you add comment to detail why you need to sort?

martmull (approved):

packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-runner-v2/action-handlers/field/services/update-field-action-handler.service.ts
Weiko: You'll need to extract it from the array first then, right? Not sure if this is that much better. [...updates] is probably not needed though (the copy is usually preferable since sort mutates the array but we don't use "updates" after so it could be removed)

prastoin, line 157: Here's what I meant
```ts
    const sortedUpdatesWithDefaultValuesUpdateLast = hasDefaultValueUpdate
      ? [
          ...updates.filter((update) => update.property !== 'defaultValue'),
          defaultValueUpdate,
        ]
      : updates;
```

prastoin: Here's what I meant, we already have the `defaultValueUpdate`
```ts
    const sortedUpdatesWithDefaultValuesUpdateLast = hasDefaultValueUpdate
      ? [
          ...updates.filter((update) => update.property !== 'defaultValue'),
          defaultValueUpdate,
        ]
      : updates;
```

prastoin:
About to create the PR to handle the integration test and few perf and readability nitpicks

PR #15287, opened 2025-10-23
fix: authorization bypass in next.js middleware 

mabdullahabaid:
Fixes [Dependabot Alert 216](https://github.com/twentyhq/twenty/security/dependabot/216) - authorization bypass in next.js middleware.

Updated `react-email` version from `4.0.3` to `4.0.4`. This bumps up Next.js to a safer version for the mentioned critical alert. However, even the latest `react-email` package has not upgraded to Next.js 15.4.7 - the recommended version by dependabot.

Since `react-email` is a devDependency used to preview email templates during development, it never gets inserted into the production build. Therefore, I marking the following alerts as `vulnerable code is never used` with a comment that it never makes it to the production build.

<p align="center">
image
</p>

The only other place where we have next imported is twenty-website, which uses the safe version `14.2.33`.

<p align="center">
image
</p>

### DIFF
--- a/packages/twenty-emails/package.json
+++ b/packages/twenty-emails/package.json
@@ -24,7 +24,7 @@
     "@tiptap/core": "^3.4.2",
     "@types/react": "^19",
     "@types/react-dom": "^19",
-    "react-email": "4.0.3"
+    "react-email": "4.0.4"
   },
   "exports": {
     ".": {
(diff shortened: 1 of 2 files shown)


charlesBochet (approved):
Ty @mabdullahabaid 
I think we can merge this one but react-email is used in send email tool if I'm not mistaken

mabdullahabaid:
@charlesBochet thank you.

The base package `react-email` is not used anywhere - no imports. Send email imports render and toPlainText from `react-email/render`. All the JSX templates use `react-email/components`.

PR #15307, opened 2025-10-23
[groupBy] Load record within groups (without relations)

ijreilly:
First step of https://github.com/twentyhq/core-team-issues/issues/1726. I will handle relations in another PR.
Another ticket is planned to allow for sorting among the records.

When querying records we have set the number of maximum groups to 50, and of maximum records par group to 10.

image


packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
```diff
@@ -0,0 +1,188 @@
...
+      .limit(GROUPS_LIMIT)
+      .getRawMany();
+
+    if (groupsResult.length === 0) {
+      return [];
+    }
+
+    const columnsToSelect = buildColumnsToSelect({
+      select: selectedFieldsResult.select,
+      relations: selectedFieldsResult.relations, // TODO - not handled for now
```
ijreilly, line 48: I will handle in the next PR to be able to already send this for review

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
```diff
@@ -0,0 +1,188 @@
...
+    queryBuilderWithGroupBy: WorkspaceSelectQueryBuilder<ObjectLiteral>;
+    queryBuilderWithFiltersAndWithoutGroupBy: WorkspaceSelectQueryBuilder<ObjectLiteral>;
+    groupByDefinitions: GroupByDefinition[];
+    objectMetadataItemWithFieldMaps: ObjectMetadataItemWithFieldMaps;
+    selectedFieldsResult: GraphqlQuerySelectedFieldsResult;
+    objectMetadataMaps: ObjectMetadataMaps;
+    repository: WorkspaceRepository<ObjectLiteral>;
+  }): Promise<CommonGroupByOutputItem[]> {
+    const groupsResult = await queryBuilderWithGroupBy
+      .limit(GROUPS_LIMIT)
```
ijreilly, line 39: for now we have a hardcoded number of groups and number of records per groups limit. They may become configurable parameters later

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    const queryBuilderWithPartitionBy = this.addPartitionByToQueryBuilder({
+      queryBuilderForSubQuery: queryBuilderWithFiltersAndWithoutGroupBy,
+      columnsToSelect,
+      groupsResult,
+      groupByDefinitions,
+      repository,
+    });
+
+    const recordsResult = await queryBuilderWithPartitionBy.getRawMany();
```
ijreilly, line 61: getMany() is unfortunately not an option with our subquery

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    const subQuery = queryBuilderForSubQuery
+      .select(recordSelectWithAlias)
+      .addSelect(groupBySelectWithAlias)
+      .addSelect(`ROW_NUMBER() OVER (PARTITION BY ${groupByExpressions})`, 'rn')
+      .andWhere(groupConditions);
+
+    let mainQueryQueryBuilder = repository.createQueryBuilder();
+
+    const mainQuery = mainQueryQueryBuilder
```
ijreilly, line 116: some unavoidable (i think) raw sql is there but i did my best to still use query builder to be able to run permissions

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
```diff
@@ -0,0 +1,188 @@
...
+      .from(`(${subQuery.getQuery()})`, 'ranked_records')
+      .setParameters(queryBuilderForSubQuery.expressionMap.parameters)
+      .select(groupByAliases)
+      .addSelect(
+        `JSON_AGG(
+        CASE WHEN rn <= ${RECORDS_PER_GROUP_LIMIT} THEN
+          JSON_BUILD_OBJECT(
+            ${[
+              ...Object.keys(columnsToSelect).map(
+                (col) => `'${col}', "${SUB_QUERY_PREFIX}${col}"`,
```
ijreilly, line 126: this prefix is used to avoid sql confusion if a column is used for both grouping and as a requested field

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
```diff
@@ -0,0 +1,188 @@
...
+            ].join(',\n              ')}
+          )
+        END
+      ) FILTER (WHERE rn <= ${RECORDS_PER_GROUP_LIMIT})`,
+        'records',
+      )
+      .groupBy(groupByAliases);
+
+    // Remove initial "from" condition (typeOrm limitation)
+    mainQuery.expressionMap.aliases = mainQuery.expressionMap.aliases.filter(
```
ijreilly, line 140: when using repository.createQueryBuilder, the expressionMap is loaded with the alias of the repository's entity, even if we use a different from condition afterwards - as we do in the mainQuery where our "from" condition is expressed with the postgre CTE built in our subQuery. 
Without this operation, the "from"s are concatenated (looking like `FROM "workspace_1wgvd1injqtife6y4rvfbu3h5"."opportunity" "opportunity", (SELECT "opportunity"."name"...` -> thats our CTE), which is interpreted as a kind of CROSS JOIN - not what we want!.

packages/twenty-server/src/engine/twenty-orm/repository/permissions.utils.ts
```diff
@@ -17,7 +17,34 @@ import {
...
+        | 'update'
+        | 'delete'
+        | 'restore'
+        | 'soft-delete'
+        | 'relation';
+    } => {
+  const isSubQuery = expressionMap.aliases[0].subQuery;
+
+  if (isSubQuery) {
+    return {
```
ijreilly, line 43: we can early return because the subQuery permissions are evaluated when it is being executed by the query builder. 
This is not ideal though because it is taylored to our implementation here: if there is a subQuery, it is only because we are in the case of a partitioned query for groupBy; and if it's ok to only check the permissions of the subQuery, it's because in the case of a groupBy with our implementation, the query fo aggregated values are handled in a different query that will be permission-checked on its side; and on the partitioned query, it is the subQuery who does the query to the entity's fields. So all due checks are performed. But if tomorrow we have another implementation that uses a subQuery, we will need to make sure all permissions are still checked...

packages/twenty-server/src/engine/api/common/common-query-runners/common-group-by-query-runner.service.ts
```diff
@@ -11,59 +11,98 @@ import {
...
-  ): Promise<CommonGroupByOutputItem[]> {
-    const {
-      repository,
-      commonQueryParser,
+  async run({
+    args,
+    authContext,
+    objectMetadataMaps,
+    objectMetadataItemWithFieldMaps,
+    shouldIncludeRecords,
```
etiennejouan: I wonder if we can deduce it from the selected fields.

packages/twenty-server/src/engine/api/graphql/workspace-resolver-builder/factories/group-by-resolver.factory.ts
```diff
@@ -36,38 +31,64 @@ export class GroupByResolverFactory
...
+              objectName:
+                internalContext.objectMetadataItemWithFieldMaps.nameSingular,
+              objectRecordsAggregatedValues: {},
+              selectedAggregatedFields: {},
+              take: group.records?.length || 0,
+              totalCount: group.records?.length || 0,
+              order: undefined,
+              depth: 0,
+            }),
+            cursor: '',
```
etiennejouan: using `typeORMObjectRecordsParser.createConnection` may be more appropriate to compute cursor ?

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
```diff
@@ -0,0 +1,188 @@
...
+      .select(recordSelectWithAlias)
+      .addSelect(groupBySelectWithAlias)
+      .addSelect(`ROW_NUMBER() OVER (PARTITION BY ${groupByExpressions})`, 'rn')
+      .andWhere(groupConditions);
+
+    let mainQueryQueryBuilder = repository.createQueryBuilder();
+
+    const mainQuery = mainQueryQueryBuilder
+      .from(`(${subQuery.getQuery()})`, 'ranked_records')
+      .setParameters(queryBuilderForSubQuery.expressionMap.parameters)
```
etiennejouan, line 118: Why do we need this ?

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
```diff
@@ -0,0 +1,188 @@
...
+      .select(groupByAliases)
+      .addSelect(
+        `JSON_AGG(
+        CASE WHEN rn <= ${RECORDS_PER_GROUP_LIMIT} THEN
```
etiennejouan, line 122: Really nice to aggregate here all the records per group ! 
I wonder though how to fetch the relations if the records are grouped by groupBy. You can't use easily the .processNestedRelations for all records

etiennejouan (approved):
Tough one 👏 
I only left few comments or questions for my understanding.

I'm wondering how we will deal with nested relations if records are returned (from orm) per group.

packages/twenty-server/src/engine/api/common/common-query-runners/common-group-by-query-runner.service.ts
ijreilly: yes that's what we do on the graphql side:
```
      const shouldIncludeRecords =
        isDefined(selectedFields.edges?.node) &&
        Object.keys(selectedFields.edges?.node).length > 0;
```

On rest side, we need to read from a query argument, which makes us determine if we should compute selectedFields. I find it more straightforward to still tie shouldIncludeRecords to the query argument, and i think its more clear like that, but that's questionnable indeed! i dont think it matters much though

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
ijreilly: subQuery.getQuery() does not include the parameters, which we need for the filters

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
ijreilly: let's discuss this next - i think we will anyway need to do a separate query to get the relations, as we currently do for findMany if im not mistaken. We could do a separate query then re-build the groups. I will handle this next!

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
```diff
@@ -0,0 +1,188 @@
...
+    if (groupsResult.length === 0) {
+      return '';
+    }
+
+    const groupConditions = groupsResult.map((group) => {
+      const conditions = groupByDefinitions
+        .map((def) => {
+          const value = group[def.alias];
+
+          return `${def.expression} = ${this.formatValueForSql(value)}`;
```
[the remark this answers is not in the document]
ijreilly: groupByDefinition is safe to use in queries because it is built from parsed group by args which throw if any of the group by args does not match an existing fieldMetadata

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
```diff
@@ -0,0 +1,188 @@
...
+    }
+    if (typeof value === 'boolean') {
+      return value === true ? 'TRUE' : 'FALSE';
+    }
+    if (value instanceof Date) {
+      return `'${value.toISOString()}'`;
+    }
+
+    return `'${String(value).replace(/'/g, "''")}'`;
+  }
```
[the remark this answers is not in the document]
ijreilly: I agree, I will use parameters instead

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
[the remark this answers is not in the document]
Weiko: The parsing is done outside of the service though, right? nitpicking here ofc, I don't see this service being used outside of what you've built!

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
```diff
@@ -0,0 +1,188 @@
...
+        })
+        .join(' AND ');
+
+      return `(${conditions})`;
+    });
+
+    return `(${groupConditions.join(' OR ')})`;
+  }
+
+  private formatValueForSql(value: unknown): string {
```
Weiko: would it make sense to be a util in twenty orm module?

packages/twenty-server/src/engine/twenty-orm/repository/permissions.utils.ts
Weiko: LGTM, maybe add a comment that explains that? Probably one of the rare cases where a comment can be meaningful :)

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
```diff
@@ -0,0 +1,188 @@
...
+      .join(', ');
+
+    const groupBySelectWithAlias = groupByDefinitions
+      .map((def) => `${def.expression} as "${def.alias}"`)
+      .join(', ');
+
+    const subQuery = queryBuilderForSubQuery
+      .select(recordSelectWithAlias)
+      .addSelect(groupBySelectWithAlias)
+      .addSelect(`ROW_NUMBER() OVER (PARTITION BY ${groupByExpressions})`, 'rn')
```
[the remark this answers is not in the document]
ijreilly: groupByExpressions is safe to use

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
ijreilly: Actually i removed it, it was unnecessary

PR #15310, opened 2025-10-23
Stop parsing the PG_DATABASE_URL, use it directly

EliRibble:
Addresses https://github.com/twentyhq/twenty/issues/15274

There's no need to parse the URL, since psql is happy to use it directly. If you are going to parse the URL, you should parse it correctly - the logic removed here make a number of assumptions about the URL that are inaccurate and reject many types of valid URIs.

This does drop the ability to create the database which may be handy for people that don't do database administration. For people that do, this is an anti-feature.

### DIFF
--- a/packages/twenty-docker/docker-compose.yml
+++ b/packages/twenty-docker/docker-compose.yml
@@ -109,8 +109,9 @@ services:
     volumes:
       - db-data:/var/lib/postgresql/data
     environment:
-      POSTGRES_USER: ${PG_DATABASE_USER:-postgres}
+      POSTGRES_DB: ${PG_DATABASE_NAME:-default}
       POSTGRES_PASSWORD: ${PG_DATABASE_PASSWORD:-postgres}
+      POSTGRES_USER: ${PG_DATABASE_USER:-postgres}
     healthcheck:
       test: pg_isready -U ${PG_DATABASE_USER:-postgres} -h localhost -d postgres
       interval: 5s
--- a/packages/twenty-docker/twenty/entrypoint.sh
+++ b/packages/twenty-docker/twenty/entrypoint.sh
@@ -8,27 +8,15 @@ setup_and_migrate_db() {
     fi
 
     echo "Running database setup and migrations..."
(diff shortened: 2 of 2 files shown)


FelixMalfait:
I agree, creating the DB was probably a wrong idea. Sometimes we try to make things easy for end-users, but this is probably not adding clarity. Would you mind checking the failing docker-composer test and creating the DB there then?

Thank you!

EliRibble:
I'm open to the idea of fixing the tests, but it looks like they use github workflows exclusively, and do so by using YAML to define manipulations of more YAML that defines containers. That's a pretty precarious stack, and I'm not aware of any instructions on how to run these tests locally. This could take _a lot_ of trial-and-error to get right.

FelixMalfait:
Yes, it's a pain to work with locally! We can takeover this PR, no worries. Thanks for this already!

EliRibble:
Okay, looks like the tests are manipulating [packages/twenty-docker/docker-compose.yml](https://github.com/twentyhq/twenty/blob/main/packages/twenty-docker/docker-compose.yml) which is specifying the canonical Postgres container image. That means [it'll default to the name of the POSTGRES_USER](https://www.docker.com/blog/how-to-use-the-postgres-docker-official-image/) since `POSTGRES_DB` isn't defined. [The test failure](https://github.com/twentyhq/twenty/actions/runs/18759177423/job/53518946189?pr=15310) indicates that it's using `default` in the client code to connect:

```
server-1  | Running database setup and migrations...
server-1  | psql: error: connection to server at "db" (172.18.0.3), port 5432 failed: FATAL:  database "default" does not exist
```

I'm assuming this is because of the way [the drizzle connection is made](https://github.com/twentyhq/twenty/issues/15274#issuecomment-3433924393).

I'll try specifying the database name and see what we get...

FelixMalfait:
I was thinking of adding "create default DB" as a step in the docker-compose CI ; I might be missing a deeper root cause which you have in mind. I'll have a look tomorrow (I'm on CET).

Drizzle is not really related to this as it's only used in twenty-website which is the old code of our marketing website which we'll probably throw away soon. The app's code is in twenty-server.

FelixMalfait (approved):
Thanks!

PR #15321, opened 2025-10-24
fix(billing): adapt current phase to current period

### LINKED ISSUE #1748: Billing Issue
When I switch from yearly to monthly, the change takes effect at the end of the month instead of the end of the year.

AMoreaux:
Fix https://github.com/twentyhq/core-team-issues/issues/1748


packages/twenty-front/src/modules/billing/components/SettingsBillingSubscriptionInfo.tsx
```diff
@@ -339,6 +339,8 @@ export const SettingsBillingSubscriptionInfo = ({
     }
   };
 
+  console.log('>>>>>>>>>>>>>>', nextBillingPhase);
```
charlesBochet: we should remove that :p

charlesBochet (approved):
LGTM, once CI are green!

PR #15323, opened 2025-10-24
Remove unused dependencies from root package.json.

mabdullahabaid:
These removed dependencies are not being imported anywhere in the code base. 

Both `twenty-server` and `twenty-front` build properly, ensuring the dependent packages like `twenty-emails`, `twenty-ui`, `twenty-shared` etc are building properly.


package.json
```diff
@@ -44,8 +42,6 @@
     "lodash.snakecase": "^4.1.1",
     "lodash.upperfirst": "^4.3.1",
     "microdiff": "^1.3.2",
-    "moize": "^6.1.6",
-    "patch-package": "^8.0.0",
```
charlesBochet, line 48: I'm not sure we can remove this one, it's what we use to create yarn patches, it should be in dev though

charlesBochet (approved):

PR #15340, opened 2025-10-24
Add sort section in find records action

thomtrp:
https://github.com/user-attachments/assets/020f3e97-316b-41db-b95d-7b938a7f82cf

### DIFF
--- a/packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/find-records-action/components/WorkflowFindRecordsSorts.tsx
+++ b/packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/find-records-action/components/WorkflowFindRecordsSorts.tsx
@@ -67,7 +67,7 @@ export const WorkflowFindRecordsSorts = ({
 
     const newSort: RecordSort = {
       id: uuidv4(),
-      fieldMetadataId: '',
+      fieldMetadataId: sortableFields[0]?.value ?? '',
       direction: ViewSortDirection.ASC,
     };
 
--- a/packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/find-records-action/components/WorkflowFindRecordsSorts.tsx
+++ b/packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/find-records-action/components/WorkflowFindRecordsSorts.tsx
@@ -1,0 +1,161 @@
+import { type ObjectMetadataItem } from '@/object-metadata/types/ObjectMetadataItem';
+import { filterSortableFieldMetadataItems } from '@/object-metadata/utils/filterSortableFieldMetadataItems';
+import { type RecordSort } from '@/object-record/record-sort/types/RecordSort';
+import { Select } from '@/ui/input/components/Select';
+import styled from '@emotion/styled';
+import { useLingui } from '@lingui/react/macro';
+import { IconArrowsSort, useIcons } from 'twenty-ui/display';
+import { Button, type SelectOption } from 'twenty-ui/input';
+import { v4 as uuidv4 } from 'uuid';
+import { ViewSortDirection } from '~/generated/graphql';
+
+const StyledContainer = styled.div`
+  display: flex;
+  flex-direction: column;
+  gap: ${({ theme }) => theme.spacing(2)};
+`;
+
+const StyledSortItemContainer = styled.div`
+  align-items: center;
+  display: flex;
+  gap: ${({ theme }) => theme.spacing(2)};
+`;
+
+const StyledAddButtonContainer = styled.div`
+  margin-top: ${({ theme }) => theme.spacing(1)};
+`;
+
+type WorkflowFindRecordsSortsProps = {
+  sorts: RecordSort[];
+  onChange: (sorts: RecordSort[]) => void;
+  objectMetadataItem: ObjectMetadataItem;
+  readonly: boolean;
+};
+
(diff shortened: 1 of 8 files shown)


packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/find-records-action/components/WorkflowFindRecordsSorts.tsx
```diff
@@ -0,0 +1,161 @@
...
+              options={directionOptions}
+              onChange={(value) =>
+                handleDirectionChange(sort.id, value as ViewSortDirection)
+              }
+            />
+
+            {!readonly && (
+              <Button
+                onClick={() => handleRemoveSort(sort.id)}
+                Icon={getIcon('IconTrash')}
```
Devessier: nitpick: Could we directly import the IconTrash component instead?

packages/twenty-server/src/engine/core-modules/record-crud/types/record-crud-input.type.ts
```diff
@@ -32,7 +32,11 @@ export type FindRecordsInput = {
     recordFilters?: any;
     gqlOperationFilter?: Partial<ObjectRecordFilter>[];
   };
-  orderBy?: Partial<ObjectRecordOrderBy>;
+  sort?: {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    sorts?: any;
```
Devessier: nitpick: Can we use `unknown` instead of `any`?

packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/find-records-action/components/WorkflowFindRecordsSorts.tsx
```diff
@@ -0,0 +1,161 @@
...
+  const directionOptions: Array<SelectOption<ViewSortDirection>> = [
+    {
+      label: t`Ascending`,
+      value: ViewSortDirection.ASC,
+    },
+    {
+      label: t`Descending`,
+      value: ViewSortDirection.DESC,
+    },
+  ];
```
Devessier, line 61: As far as I know, `Ascending` means `AscNullsFirst` when sorting views.

Do you think we should expose this low-level behavior here? I feel advanced users might want to decide what to do with null values.

Devessier (approved):
Great work 💪 I left a few non-blocking nitpicks.

packages/twenty-server/src/engine/core-modules/record-crud/types/record-crud-input.type.ts
```diff
@@ -32,7 +32,11 @@ export type FindRecordsInput = {
     recordFilters?: any;
     gqlOperationFilter?: Partial<ObjectRecordFilter>[];
   };
-  orderBy?: Partial<ObjectRecordOrderBy>;
+  sort?: {
```
Weiko: I don't fully understand why we have this pattern

Weiko: Why not simply OrderByInput[] like the API?

Weiko: Also maybe check with @etiennejouan if this is aligned with the CommonAPI if you intend to use it in the future 🤔

Weiko (changes requested):
Let's discuss this before merging please 🙏

packages/twenty-server/src/engine/core-modules/record-crud/types/record-crud-input.type.ts
FelixMalfait: I agree with Coco the API is not clear to me as an external user, what is the difference between sort and orderBy? To me sort and order were ~ synomnyms so we should even batch-rename in the codebase to only keep OrderBy (?)

thomtrp: True that this is unclear. The main reason is that I re-used filer/sort frontend logic. It carries a lot of utils to turn the filter objects into gql fragments. I did not want to move all of these to twenty shared, so I decided that before storing the filter object, I would turn it into a fragment `gqlOperationFilter` and store it as well. 

This way, the filter can be easily modified and stored. But on execution, this is the gql that is used. 

I did the same for sorting but poorly named. `Sorts` are the frontend objects. `OrderBy` are the gql fragments. I can either rename sorts to orderBy and orderBy to gqlOperationOrderBy. Or I move all objects and utils I need to twenty-shared but that may be heavy and if I do it for sort I should do it for filters.

packages/twenty-server/src/engine/core-modules/record-crud/types/record-crud-input.type.ts
thomtrp: makes it way harder to use afterwards 😓

packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/find-records-action/components/WorkflowFindRecordsSorts.tsx
thomtrp: I guess we could give the choice between both indeed. But for now let's keep consistent behaviors with other parts of the app 👍

FelixMalfait (approved):

PR #15344, opened 2025-10-24
CleanOrphanedKanbanAggregateOperationFieldMetadataIdCommand

Weiko:
## Context

We want to introduce a FK between view and fieldMetadata through KanbanAggregateOperationFieldMetadataId so we need to clean up orphan ones


packages/twenty-server/src/database/commands/upgrade-version-command/1-10/1-10-clean-orphaned-kanban-aggregate-operation-field-metadata-id.command.ts
```diff
@@ -0,0 +1,90 @@
...
+        'No orphaned kanbanAggregateOperationFieldMetadataId references found',
+      );
+
+      return;
+    }
+
+    const orphanedViewIds = orphanedViews.map((view) => view.id);
+
+    if (!options.dryRun) {
+      await this.viewRepository.update(
```
prastoin: We could dyamically check if the layout is kanban and then delete the view, if it's a table nullify as you do !

prastoin (approved):
Image

PR #15347, opened 2025-10-24
1751 extensibility twenty sdk v2 use twenty sdk to define a serverless function trigger

### LINKED ISSUE #1751: [Extensibility] [twenty-sdk] [v2] Use twenty-sdk to define a serverless function trigger
As title

- RouteTrigger
- DatabaseEventTrigger
- CronTrigger

We need to add 2 handlerPath and handlerName to serverlessFunction entity so we can locate and trigger serverless contained in an application project.

In another ticket we will add a new abstraction serverlessFunctionProject that can contains multiple serverlessFunctions. It will permit to store only one version of the serverlessFunctionProject in the localStorage and not one for each serverlessFunction. 

We might conclude in the end that a serverlessFunctionProject should belong to an application

### LINKED ISSUE #1752: [Extensibility] [twenty-sdk] [v2] Use twenty-sdk to define a serverlessFunction
As title, 

define `ServerlessFunction` decorator in twenty-sdk

Support serverlessFunction.manifest.jsonc but set it as deprecated

Update hello-world so it uses `ServerlessFunction` decorator

martmull:
This PR adds 2 columns handlerPath and handlerName in serverlessFunction to locate the entrypoint of a serverless in a codebase

It adds the following decorators in twenty-sdk:
- ServerlessFunction
- DatabaseEventTrigger
- RouteTrigger
- CronTrigger
- ApplicationVariable

It still supports deprecated entity.manifest.jsonc 

Overall code needs to be cleaned a little bit, but it should work properly so you can try to test if the DEVX fits your needs

See updates in hello-world application

```typescript
import axios from 'axios';
import {
  DatabaseEventTrigger,
  ServerlessFunction,
  RouteTrigger,
  CronTrigger,
  ApplicationVariable,
} from 'twenty-sdk';

@ApplicationVariable({
  universalIdentifier: 'dedc53eb-9c12-4fe2-ba86-4a2add19d305',
  key: 'TWENTY_API_KEY',
  description: 'Twenty API Key',
  isSecret: true,
})
@DatabaseEventTrigger({
  universalIdentifier: '203f1df3-4a82-4d06-a001-b8cf22a31156',
  eventName: 'person.created',
})
@RouteTrigger({
  universalIdentifier: 'c9f84c8d-b26d-40d1-95dd-4f834ae5a2c6',
  path: '/post-card/create',
  httpMethod: 'GET',
  isAuthRequired: false,
})
@CronTrigger({
  universalIdentifier: 'dd802808-0695-49e1-98c9-d5c9e2704ce2',
  pattern: '0 0 1 1 *', // Every year 1st of January
})
@ServerlessFunction({
  universalIdentifier: 'e56d363b-0bdc-4d8a-a393-6f0d1c75bdcf',
})
class CreateNewPostCard {
  main = async (params: { recipient: string }): Promise<string> => {
    const { recipient } = params;

    const options = {
      method: 'POST',
      url: 'http://localhost:3000/rest/postCards',
      headers: {
        'Content-Type': 'application/json',
        Authorization: `Bearer ${process.env.TWENTY_API_KEY}`,
      },
      data: { name: recipient ?? 'Unknown' },
    };

    try {
      const { data } = await axios.request(options);

      console.log(`New post card to "${recipient}" created`);

      return data;
    } catch (error) {
      console.error(error);
      throw error;
    }
  };
}

export const createNewPostCardHandler = new CreateNewPostCard().main;

```

### [edit] V2 

After the v1 proposal, I see that using a class method to define the serverless function handler is pretty confusing. Lets leave serverlessFunction configuration decorators on the class, but move the handler like before. Here is the v2 hello-world serverless function:

```typescript
import axios from 'axios';
import {
  DatabaseEventTrigger,
  ServerlessFunction,
  RouteTrigger,
  CronTrigger,
  ApplicationVariable,
} from 'twenty-sdk';

@ApplicationVariable({
  universalIdentifier: 'dedc53eb-9c12-4fe2-ba86-4a2add19d305',
  key: 'TWENTY_API_KEY',
  description: 'Twenty API Key',
  isSecret: true,
})
@DatabaseEventTrigger({
  universalIdentifier: '203f1df3-4a82-4d06-a001-b8cf22a31156',
  eventName: 'person.created',
})
@RouteTrigger({
  universalIdentifier: 'c9f84c8d-b26d-40d1-95dd-4f834ae5a2c6',
  path: '/post-card/create',
  httpMethod: 'GET',
  isAuthRequired: false,
})
@CronTrigger({
  universalIdentifier: 'dd802808-0695-49e1-98c9-d5c9e2704ce2',
  pattern: '0 0 1 1 *', // Every year 1st of January
})
@ServerlessFunction({
  universalIdentifier: 'e56d363b-0bdc-4d8a-a393-6f0d1c75bdcf',
})
export class ServerlessFunctionDefinition {}

export const main = async (params: { recipient: string }): Promise<string> => {
  const { recipient } = params;

  const options = {
    method: 'POST',
    url: 'http://localhost:3000/rest/postCards',
    headers: {
      'Content-Type': 'application/json',
      Authorization: `Bearer ${process.env.TWENTY_API_KEY}`,
    },
    data: { name: recipient ?? 'Unknown' },
  };

  try {
    const { data } = await axios.request(options);

    console.log(`New post card to "${recipient}" created`);

    return data;
  } catch (error) {
    console.error(error);
    throw error;
  }
};

```

### [edit] V3

After the v2 proposal, we don't really like decorators on empty classes. We decided to go with a Vercel approach with a config constant

```typescript
import axios from 'axios';
import { ServerlessFunctionConfig } from 'twenty-sdk';

export const main = async (params: { recipient: string }): Promise<string> => {
  const { recipient } = params;

  const options = {
    method: 'POST',
    url: 'http://localhost:3000/rest/postCards',
    headers: {
      'Content-Type': 'application/json',
      Authorization: `Bearer ${process.env.TWENTY_API_KEY}`,
    },
    data: { name: recipient ?? 'Unknown' },
  };

  try {
    const { data } = await axios.request(options);

    console.log(`New post card to "${recipient}" created`);

    return data;
  } catch (error) {
    console.error(error);
    throw error;
  }
};

export const config: ServerlessFunctionConfig = {
  universalIdentifier: 'e56d363b-0bdc-4d8a-a393-6f0d1c75bdcf',
  routeTriggers: [
  {
    universalIdentifier: 'c9f84c8d-b26d-40d1-95dd-4f834ae5a2c6',
    path: '/post-card/create',
    httpMethod: 'GET',
    isAuthRequired: false,
  }
  ],
  cronTriggers: [
    {
      universalIdentifier: 'dd802808-0695-49e1-98c9-d5c9e2704ce2',
      pattern: '0 0 1 1 *', // Every year 1st of January
    }
  ],
  databaseEventTriggers: [
  {
    universalIdentifier: '203f1df3-4a82-4d06-a001-b8cf22a31156',
    eventName: 'person.created',
   }
  ]
}

```


packages/twenty-apps/hello-world/src/actions/create-new-post-card.ts
```diff
@@ -0,0 +1,60 @@
+import axios from 'axios';
+import {
+  DatabaseEventTrigger,
+  ServerlessFunction,
+  RouteTrigger,
+  CronTrigger,
+  ApplicationVariable,
+} from 'twenty-sdk';
+
+@ApplicationVariable({
```
Weiko: I feel like ApplicationVariable should be handled differently, maybe in a config file, to reflect the fact that it is registered globally under the application 🤔

martmull: yes i agree, in fact we may define application with a decorated class also ^^

Weiko: Mmmm not sure this is super useful, I also feel like it should be a config. I don't see a strong benefit between
```
@Application({ ... })
class MyApplication {}
```
and a json file here.
This is useful when your class is not empty like for objects with its fields or serverless with its methods but probably not always the case 🤔

Weiko:
Can we avoid having to instantiate the class for it to work? (maybe using static) @martmull

martmull:
I will try to instanciate in the driver.execute

packages/twenty-cli/src/commands/app-add.command.ts
```diff
@@ -38,7 +39,7 @@ export const isSyncableEntity = (value: string): value is SyncableEntity => {
 export class AppAddCommand {
   async execute(entityType?: SyncableEntity): Promise<void> {
     try {
-      const appPath = CURRENT_EXECUTION_DIRECTORY;
+      const appPath = CURRENT_EXECUTION_DIRECTORY + '/' + ROOT_FOLDER;
```
Devessier: I think we could use `path.join` here

packages/twenty-cli/src/utils/app-manifest-loader.ts
```diff
@@ -121,6 +122,44 @@ export const loadManifest = async (
...
+
+  const objectFromManifests = await loadCoreEntity(
+    path.join(appPath, 'objects'),
+    (manifest, path) => validateSchema('object', manifest, path),
+  );
+
+  const serverlessFunctionsFromManifest = await loadCoreEntity(
+    path.join(appPath, 'serverlessFunctions'),
+    (manifest, path) => validateSchema('serverlessFunction', manifest, path),
+  );
```
Devessier: Would it make sense to load these three core entities concurrently with a `Promise.all`?

packages/twenty-server/src/engine/core-modules/serverless/drivers/constants/executor/index.mjs
```diff
@@ -6,20 +6,23 @@ export const handler = async (event) => {
...
   try {
-    const { code, params, env } = event;
+    const { code, params, env, handlerName } = event;
 
     await fs.writeFile(mainPath, code, 'utf8');
 
+    // eslint-disable-next-line no-undef
     process.env = { ...process.env, ...(env ?? {}) };
 
     const mainFile = await import(mainPath);
```
Devessier, line 20: Thought: One thing to know about dynamic imports is that they can cause memory leaks. An imported module is never garbage collected. If a server keeps dynamically importing modules, the free memory will continue decreasing. Do you think it could cause issues here? I'm unsure where this file is used, actually.

See https://jakearchibald.com/2025/importing-vs-fetching-json/#caching-and-garbage-collection

Devessier (commented):
This feature looks great. Congrats!

I left a few small comments in my review.

I also wrote a few thoughts on Discord about the shape of the code users will have to write to create serverless functions.

packages/twenty-server/src/engine/core-modules/applicationVariable/application-variable.service.ts
```diff
@@ -27,33 +27,52 @@ export class ApplicationVariableEntityService {
...
+    applicationVariables?: AppManifest['application']['applicationVariables'];
     applicationId: string;
   }) {
-    if (!isDefined(env)) {
+    if (!isDefined(applicationVariables)) {
       return;
     }
 
-    for (const [key, { value, description, isSecret }] of Object.entries(env)) {
-      await this.applicationVariableRepository.upsert(
```
Weiko, line 42: why not using upsert anymore?

martmull: because I don't want to save the value in update mode, but i want to save it in create mode

Weiko (approved):

PR #15354, opened 2025-10-25
Reset limit to 1 when object is changed

### LINKED ISSUE #15345: Limit field isn't reset when selecting another object in Find Records action
## Bug Description

https://github.com/user-attachments/assets/514fdb94-132f-4917-b05a-8dbd23d166a3

## Expected behavior

Selecting another object should reset the input to 1.

BKM14:
Closes #15345 

The issue was that when the `onChange` was called on the Select component in `WorkflowEditActionFindRecords.tsx`, the limit was being reset to 1, but it wasn't rendering immediately. The sidebar had to be closed and reopened. 

I have added a `useEffect` as a hack to re-render the `FormNumberFieldInput` when it's `defaultValue` is changed. I understand that using `useEffect` is not a good choice. Please suggest a better fix if any and I will implement it.


charlesBochet (commented):
@BKM14 Thanks for opening a PR, this is too hacky, we should avoid useEffect and instead reset the field from the outside when changing the object.

I'm investigated how this could be properly fixed

charlesBochet:
@BKM14 I took a deeper look. Here is what I think:

1) ideally we should have a recoilState (actually a recoilComponentState but that's a detail) instead of the useState. This would allow to keep the state behavior (not changing the draftValue at each re-render) and to allow the parent to change the value from outside
2) However, given the number of components impacted, I think this refacto is heavy and does not worth it right now
3) So, going with a simpler solution which is what you proposed: useState + useEffect. This is actually equivalent (there are some nuances) to just deriving the draftValue from the props value at each re-render
4) Note that this can cause the following issue: input can feel laggy if the render loop is too slow which does not seem to be the case here but will likely appear in the future as developments continue

So, I'm going with simpler solution as it does not worth the investment but we should do 1) at some point

charlesBochet (approved):

Devessier:
@charlesBochet, we put a `useState` in the form fields to decouple the browser input's state from the backend's state.

If we don't use a separate state for all the `FormNumberFieldInput`, we easily end up with a buggy UX, even when testing on localhost, where the backend should respond quickly. See:

https://github.com/user-attachments/assets/d999168e-6a84-4b4d-8f95-aa2bbdf2fa3b

I believe this PR introduced a regression, and we should seek a more suitable solution. I'm creating an issue to follow-up: https://github.com/twentyhq/twenty/issues/15709.

PR #15363, opened 2025-10-26
Add listkit to tiptap extension in workflow node

harshit078:
## Description

- This PR address issue - https://github.com/twentyhq/core-team-issues/issues/1768
- Added listkit bundle from tiptap which includes BulletList, orderedList, ListItem and ListKeymap in one single import
- This bundle also includes keyboards shortcuts - `Cmd + Shift + 7` and `Cmd + Shift + 8` for ordered and bullet list

## Visual Appearance

https://github.com/user-attachments/assets/7eff1233-8503-4854-bad2-2521898bc568

## Why this Approach

- our current version of tiptap is 3.4.2 while the latest is 3.8.0 hence installing these versions manually would install the latest version of 3.8.0. The issue when downgrading to 3.4.2 was that Version 3.8.0 of `@tiptap/extension-list` requires `renderNestedMarkdownContent` from @tiptap/core
but our `@tiptap/core` version 3.4.2 doesn't export this function.


thomtrp:
Hi @harshit078, thank you for your contribution. I tested sending an email and the lists are not displayed.
Capture d’écran 2025-10-28 à 18 10 31
Capture d’écran 2025-10-28 à 18 10 07

FYI the fact that email content is centered is coming from another issue.

harshit078:
@thomtrp Let me check into the issue and revert with a fix. Thanks for the feedback !

thomtrp (approved):
Code LGTM, thank you @harshit078!

PR #15366, opened 2025-10-26
fix: login redirection to active objects

anadi45:
Fixes - https://github.com/twentyhq/twenty/issues/15364

- Removed `activeNonSystemObjectMetadataItems` from the hook as it was not necessary. Hook use `alphaSortedActiveNonSystemObjectMetadataItems` now.
- Correctly check for read permissions.


packages/twenty-front/src/modules/navigation/hooks/useDefaultHomePagePath.ts
```diff
@@ -76,37 +74,25 @@ export const useDefaultHomePagePath = () => {
...
-          getActiveObjectMetadataItemMatchingId(
-            lastVisitedObjectMetadataItemId,
-          );
+        const lastVisitedObjectMetadataItem = isDefined(
+          lastVisitedObjectMetadataItemId,
+        )
+          ? getActiveObjectMetadataItemMatchingId(
+              lastVisitedObjectMetadataItemId,
+            )
+          : null;
```
prastoin: ***Polish:*** Would return undefined instead of null here to avoid introducing a new potential type in the union

packages/twenty-front/src/modules/navigation/hooks/useDefaultHomePagePath.ts
```diff
@@ -76,37 +74,25 @@ export const useDefaultHomePagePath = () => {
...
 
         if (isDefined(lastVisitedObjectMetadataItem)) {
           return {
-            view: getFirstView(lastVisitedObjectMetadataItemId),
+            view: getFirstView(lastVisitedObjectMetadataItemId!),
```
prastoin: ***Request:*** Risky and unless I'm mistaken not required non null operator

prastoin (commented):
Hey there, thanks !
Cool fix and refactor !

prastoin:
Unless I'm mistaken that does not fix the initial issue either:
Reproduction:
- Disable companies
- click Exit settings CTA -> still redirect to companies
- sign out
- sign in -> still redirect to companies

packages/twenty-front/src/modules/navigation/hooks/useDefaultHomePagePath.ts
anadi45: done

packages/twenty-front/src/modules/navigation/hooks/useDefaultHomePagePath.ts
anadi45: done

anadi45:
@prastoin thanks for the review. I tested with the same steps and it was not redirecting to companies. Let me test again and will revert back

prastoin:
Ah maybe I've been testing on the wrong branch 🤔 
Will double check this during the afternoon

prastoin:
Unless I'm mistaken still occurring, or I might be misunderstanding the initial issue

https://github.com/user-attachments/assets/0154ead7-28d5-4a78-93a9-6f636198cb4b

anadi45:
https://github.com/user-attachments/assets/b544c091-1522-4d44-a0d0-02c90cbf6008

Hi @prastoin, after a lot of debugging I found that the `useObjectPermissions` hook was causing race conditions — `useDefaultHomePagePath` was picking up objects before they were fully refreshed.

As a fix, I’m now using `setIsAppWaitingForFreshObjectMetadata` during workspace initialization to prevent these race conditions. Let me know if you’re aligned with this approach.

anadi45:
Hey @prastoin , could you please re-review this when you get a chance? I’m participating in Hacktoberfest, so I’d really appreciate a quick review and merge if everything looks good. Thank you!

charlesBochet (approved):

PR #15370, opened 2025-10-27
morph bug fix

guillim:
Two bug fixes here: 

# MorphRelationOneToManyFieldDisplay
we expect an empty array, not undefined (run time error otherwise)

Fixes `Cannot read properties of undefined (reading 'length'): Cannot read properties of undefined (reading 'length')` in **MorphRelationOneToManyFieldDisplay**

```js
morphValuesWithObjectNameSingular.every(
      (morphValueWithObjectNameSingular) =>
        morphValueWithObjectNameSingular.value.length === 0,
    );
```

# create record in table mode
if the record has a relation there is currenlty a bug in production when it is created. A cache issue.

`Missing field 'companyFoundedId' while writing result `

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-store/states/selectors/recordStoreFieldValueSelector.ts
+++ b/packages/twenty-front/src/modules/object-record/record-store/states/selectors/recordStoreFieldValueSelector.ts
@@ -58,7 +58,10 @@ export const recordStoreFieldValueSelector = selectorFamily({
       const relationType = morphRelations[0].type;
 
       if (relationType === RelationType.ONE_TO_MANY) {
-        return morphValuesWithObjectName as {
+        return morphValuesWithObjectName.map((morphValue) => ({
+          ...morphValue,
+          value: morphValue.value ? morphValue.value : [],
+        })) as {
           objectNameSingular: string;
           value: ObjectRecord[];
         }[];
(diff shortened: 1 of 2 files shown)


packages/twenty-front/src/modules/object-record/record-store/states/selectors/recordStoreFieldValueSelector.ts
```diff
@@ -58,7 +58,10 @@ export const recordStoreFieldValueSelector = selectorFamily({
       const relationType = morphRelations[0].type;
 
       if (relationType === RelationType.ONE_TO_MANY) {
-        return morphValuesWithObjectName as {
+        return morphValuesWithObjectName.map((morphValue) => ({
+          ...morphValue,
+          value: morphValue.value ? morphValue.value : [],
```
prastoin, line 63: ***Polish:*** `morphValue.value ?? []`

packages/twenty-front/src/modules/object-record/record-store/states/selectors/recordStoreFieldValueSelector.ts
```diff
@@ -58,7 +58,10 @@ export const recordStoreFieldValueSelector = selectorFamily({
...
+        })) as {
```
prastoin, line 64: ***Polish:*** no need to cast but passing a generic to map should be enough
was the root cause of the current issue btw

prastoin (approved):
Nice job !

packages/twenty-front/src/modules/object-record/record-store/states/selectors/recordStoreFieldValueSelector.ts
guillim: true !

packages/twenty-front/src/modules/object-record/record-store/states/selectors/recordStoreFieldValueSelector.ts
guillim: as discussed, ObjectRecord does not make it easy

PR #15371, opened 2025-10-27
Common - Remove feature flag

etiennejouan:
To be done in an other PR, move graphql-query-runner handlers in common-api-query-runner folder (+ renaming + typing improvments)

Fixes : 
- totalCount on findMany (Rest)
- getAllSelectedFields did not handle composite field (Rest)
- issue with endCursor (Rest)
- args processing for updateOne and createOne (Rest + Gql)
- fix findDuplicates (Gql)


packages/twenty-server/src/engine/api/common/utils/get-page-info.util.ts
```diff
@@ -14,12 +14,16 @@ export const getPageInfo = (
...
   isForwardPagination: boolean,
 ): CommonPageInfo => {
-  const { hasNextPage, hasPreviousPage } = getPaginationInfo(
+  const { hasNextPage, hasPreviousPage, hasMoreRecords } = getPaginationInfo(
     records,
     limit,
     isForwardPagination,
   );
 
+  if (hasMoreRecords) {
```
ijreilly, line 23: why is that?

packages/twenty-server/src/engine/api/common/common-query-runners/common-create-one-query-runner.service.ts
```diff
@@ -49,9 +49,20 @@ export class CommonCreateOneQueryRunnerService extends CommonBaseQueryRunnerServ
...
     args: CommonInput<CreateOneQueryArgs>,
-    _queryRunnerContext: CommonBaseQueryRunnerContext,
+    queryRunnerContext: CommonBaseQueryRunnerContext,
   ): Promise<CommonInput<CreateOneQueryArgs>> {
-    return args;
+    const { authContext, objectMetadataItemWithFieldMaps } = queryRunnerContext;
+
+    return {
+      ...args,
+      data: (
```
ijreilly, line 58: what is this for?

ijreilly (approved):
left a few comments to help me understand :)

packages/twenty-server/src/engine/api/common/common-query-runners/common-create-one-query-runner.service.ts
etiennejouan: Args computing enables to transform data, for example primaryLinkUrl or phones field (checking calling code, country code, ...)
As createOne and updateOne use createMany, updateMany I thought it was ok. But args processing is done during the .execute method of base query runner.

createOne.execute > base.execute (createOne.validate, createOne.compute, createOne.run,..) > createMany.run

It never went through createMany.compute.

packages/twenty-server/src/engine/api/common/utils/get-page-info.util.ts
etiennejouan: To know if you have more records, we fetch total of requested records + 1. 
But when computing endCursor, we want the last cursor of the requested records, not the +1. So we need to remove it before.

Issue #15312, opened 2025-10-24
redis connection failure, coolify


panquequelol:
## bug description

total noob here, but

basically "server" was trying to connect to "redis" before redis was ready

i just added "redis" to the depends_on of "server" and "healthcheck" to "redis"

```
    depends_on:
      db:
        condition: service_healthy
      redis:
        condition: service_healthy
```

now it runs perfectly, i dunno if someone else has this issue

## context

deploying twenty on google cloud's computer engine vm with e2-medium (2 vCPUs, 4 GB Memory)

prastoin:
Hey @panquequelol, indeed there's might be a race condition
Please feel free to create a pull-request applying the changes you're suggesting on `packages/twenty-docker/docker-compose.yml` directly

PR #15375, opened 2025-10-27
`[DOCKER_COMPOSE]` server depends on redis

anadi45:
…or Redis service

Fixes - https://github.com/twentyhq/twenty/issues/15312

### DIFF
--- a/packages/twenty-docker/docker-compose.yml
+++ b/packages/twenty-docker/docker-compose.yml
@@ -48,6 +48,8 @@ services:
     depends_on:
       db:
         condition: service_healthy
+      redis:
+        condition: service_healthy
     healthcheck:
       test: curl --fail http://localhost:3000/healthz
       interval: 5s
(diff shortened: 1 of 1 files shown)


packages/twenty-docker/docker-compose.yml
```diff
@@ -100,6 +102,8 @@ services:
     depends_on:
       db:
         condition: service_healthy
+      redis:
```
prastoin: Redundant as already depends on server

prastoin (commented):

prastoin:
@greptileai trigger

packages/twenty-docker/docker-compose.yml
anadi45: makes sense, removed it

prastoin (approved):
Thanks !

packages/twenty-docker/docker-compose.yml
prastoin: Same for the twenty-db in the end, but that's nitpicking

PR #15377, opened 2025-10-27
[Breaking Change] Implement reliable date picker utils to handle all timezone combinations

### LINKED ISSUE #1781: Date picker and filters fixes
This issue lists the quick fixes for date picker on filters and fields, following-up the refactor of timezone management for DATE and DATE_TIME.

- [x] Datepicker DATE opens with an empty input and strange values for year picker

Image

- [x] IS today not working for DATE
- [x]  Is in past et Is in future have the same problem (seems related to operator switch)
- [x]  The IS operator for DATE_TIME does not seem relevant and does not reliably work :

Image

- [x] The DATE_TIME filter chip shouldn't show seconds

Image

- [x] When creating a DATE_TIME filter, initialize seconds and microseconds to 0 

Image

- [x] Difference in date between input and calendar picker, seems to happen only when switching user settings timezone (verify)

Image

- [x]  IS relative THIS DAY does not work for DATE_TIME
- [x]  Past 1 day does not work for DATE_TIME (oddly enough past 2 days works)

- [x] DATE IS should not use time in picker

Image

- [x] Add timezone code when user settings is not system

Image

Image

- [x] Filter chip should use the same format as everything else 

Image

- [x] The current logic for week computation is one day odd, maybe date-fns that uses a wrong start of week day.
- [x] DATE IS does not work
- [x] Fix date range picker for both DATE and DATE_TIME
- [x] Advanced filter relative date calendar picker is broken (related to highlighted date range)

Image

- [x] Remove displayValue props from RecordFilter and compute it dynamically, at least for DATE_TIME for now

Image

- [x]  Filter chip and date picker do not use the same timezone (at the beginning), should be fixed by dynamic computation of displayValue

- [x] Advanced filter date time picker crashes

Image

lucasbordeau:
This PR implements the necessary tools to have `react-datepicker` calendar and our date picker components work reliably no matter the timezone difference between the user execution environment and the user application timezone.

Fixes https://github.com/twentyhq/core-team-issues/issues/1781

This PR won't cover everything needed to have Twenty handle timezone properly, here is the follow-up issue : https://github.com/twentyhq/core-team-issues/issues/1807

# Features in this PR

This PR brings a lot of features that have to be merged together.

- DATE field type is now handled as string only, because it shouldn't involve timezone nor the JS Date object at all, since it is a day like a birthday date, and not an absolute point in time.
- DATE_TIME field wasn't properly handled when the user settings timezone was different from the system one
- A timezone abbreviation suffix has been added to most DATE_TIME display component, only when the timezone is different from the system one in the settings.
- A lot of bugs, small features and improvements have been made here : https://github.com/twentyhq/core-team-issues/issues/1781

# Handling of timezones

## Essential concepts

This topic is so complex and easy to misunderstand that it is necessary to define the precise terms and concepts first. It resembles character encoding and should be treated with the same care.

- Wall-clock time : the time expressed in the timezone of a user, it is distinct from the absolute point in time it points to, much like a pointer being a different value than the value that it points to.
- Absolute time : a point in time, regardless of the timezone, it is an objective point in time, of course it has to be expressed in a given timezone, because we have to talk about when it is located in time between humans, but it is in fact distinct from any wall clock time, it exists in itself without any clock running on earth. However, by convention the low-level way to store an absolute point in time is in UTC, which is a timezone, because there is no way to store an absolute point in time without a referential, much like a point in space cannot be stored without a referential.
- DST : Daylight Save Time, makes the timezone shift in a specific period every year in a given timezone, to make better use of longer days for various reasons, not all timezones have DST. DST can be 1 hour or 30 min, 45 min, which makes computation difficult.
- UTC : It is NOT an “absolute timezone”, it is the wall-clock time at 0° longitude without DST, which is an arbitrary and shared human convention. UTC is often used as the standard reference wall-clock time for talking about absolute point in time without having to do timezone and DST arithmetic. PostgreSQL stores everything in UTC by convention, but outputs everything in the server’s SESSION TIMEZONE.

## How should an absolute point in time be stored ?

Since an absolute point in time is essentially distinct from its timezone it could be stored in an absolute way, but in practice it is impossible to store an absolute point in time without a referential. We have to say that a rocket launched at X given time, in UTC, EST, CET, etc. And of course, someone in China will say that it launched at 10:30, while in San Francisco it will have launched at 19:30, but it is THE SAME absolute point in time. 

Let’s take a related example in computer science with character encoding. If a text is stored without the associated encoding table, the correct meaning associated to the bits stored in memory can be lost forever. It can become impossible for a program to guess what encoding table should be used for a given text stored as bits, thus the glitches that appeared a lot back in the early days of internet and document processing.

The same can happen with date time storing, if we don’t have the timezone associated with the absolute point in time, the information of when it absolutely happened is lost.

It is NOT necessary to store an absolute point in time in UTC, it is more of a standard and practical wall-clock time to be associated with an absolute point in time. But an absolute point in time MUST be store with a timezone, with its time referential, otherwise the information of when it absolutely happened is lost. 

For example, it is easier to pass around a date as a string in UTC, like `2024-01-02T00:00:00Z` because it allows front-end and back-end code to “talk” in the same standard and DST-free wall-clock time, BUT it is not necessary. Because we have date libraries that operate on the standard ISO timezone tables, we can talk in different timezone and let the libraries handle the conversion internally.

It is false to say that UTC is an absolute timezone or an absolute point in time, it is just the standard, conventional time referential, because one can perfectly store every absolute points in time in UTC+10 with a complex DST table and have the exactly correct absolute points in time, without any loss of information, without having any UTC+0 dates involved.

Thus storing an absolute point in time without a timezone associated, for example with `timestamp` PostgreSQL data type, is equivalent to storing a wall-clock time and then throwing away voluntarily the information that allows to know when it absolutely happened, which is a voluntary data-loss if the code that stores and retrieves those wall-clock points in time don’t store the associated timezone somewhere.

This is why we use `timestamptz` type in PostgreSQL, so that we make sure that the correct absolute point in time is stored at the exact time we send it to PostgreSQL server, no matter the front-end, back-end and SQL server's timezone differences.

## The JavaScript Date object

The native JavaScript Date object is now officially considered legacy ([source](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date)), the Date object stores an absolute point in time BUT it forces the storage to use its execution environment timezone, and one CANNOT modify this timezone, this is a legacy behavior.

To obtain the desired result and store an absolute point in time with an arbitrary timezone there are several options : 

- The new Temporal API that is the successor of the legacy Date object.
- Moment / Luxon / @date-fns/tz that expose objects that allow to use any timezone to store an absolute point in time.

## How PostgreSQL stores absolute point in times

PostgreSQL stores absolute points in time internally in UTC ([source](https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-DATETIME-INPUT-TIME-STAMPS)), but the output date is expressed in the server’s session timezone ([source](https://www.postgresql.org/docs/current/sql-set.html)) which can be different from UTC.

Example with the object companies in Twenty seed database, on a local instance, with a new “datetime” custom column : 

image

image

## Why can’t I just use the JavaScript native Date object with some manual logic ?

Because the JavaScript Date object does not allow to change its internal timezone, the libraries that are based on it will behave on the execution environment timezone, thus leading to bugs that appear only on the computers of users in a timezone but not for other in another timezone.

In our case the `react-datepicker` library forces to use the `Date` object, thus forcing the calendar to behave in the execution environment system timezone, which causes a lot of problems when we decide to display the Twenty application DATE_TIME values in another timezone than the user system one, the bugs that appear will be of the off-by-one date class, for example clicking on 23 will select 24, thus creating an unreliable feature for some system / application timezone combinations.

A solution could be to manually compute the difference of minutes between the application user and the system timezones, but that’s not reliable because of DST which makes this computation unreliable when DST are applied at different period of the year for the two timezones.

## Why can’t I compute the timezone difference manually ?

Because of DST, the work to compute the timezone difference reliably, not just for the usual happy path, is equivalent to developing the internal mechanism of a date timezone library, which is equivalent to use a library that handles timezones.

## Using `@date-fns/tz` to solve this problem

We could have used `luxon` but it has a heavy bundle size, so instead we rely here on `@date-fns/tz` (~1kB) which gives us a `TZDate` object that allows to use any given timezone to store an absolute point-in-time.

The solution here is to trick `react-datepicker` by shifting a Date object by the difference of timezone between the user application timezone and the system timezone.

Let’s take a concerte example.

System timezone : Midway,  ⇒ UTC-11:00, has no DST.

User application timezone : Auckland, NZ ⇒ UTC+13:00, has a DST.

We’ll take the NZ daylight time, so that will make a timezone difference of 24 hours !

Let’s take an error-prone date : `2025-01-01T00:00:00` . This date is usually a good test-case because it can generate three classes of bugs : off-by-one day bugs, off-by-one month bugs and off-by-one year bugs, at the same time.

Here is the absolute point in time we take expressed in the different wall-clock time points we manipulate

Case | In system timezone ⇒ UTC-11 | In UTC | In user application timezone ⇒ UTC+13
-- | -- | -- | --
Original date | `2024-12-31T00:00:00-11:00` | `2024-12-31T11:00:00Z` | `2025-01-01T00:00:00+13:00`
Date shifted for react-datepicker | `2025-01-01T00:00:00-11:00` | `2025-01-01T11:00:00Z` | `2025-01-02T00:00:00+13:00`

We can see with this table that we have the number part of the date that is the same (`2025-01-01T00:00:00`) but with a different timezone to “trick” `react-datepicker` and have it display the correct day in its calendar.

You can find the code in the hooks `useTurnPointInTimeIntoReactDatePickerShiftedDate` and `useTurnReactDatePickerShiftedDateBackIntoPointInTime` that contain the logic that produces the above table internally.

## Miscellaneous

Removed FormDateFieldInput and FormDateTimeFieldInput stories as they do not behave the same depending of the execution environment and it would be easier to put them back after having refactored FormDateFieldInput and FormDateTimeFieldInput


prastoin:
Damn, please make an article from this

packages/twenty-front/src/modules/localization/utils/detection/detectCalendarStartDay.ts
```diff
@@ -1,10 +1,16 @@
 import { type CalendarStartDay } from '@/localization/constants/CalendarStartDay';
+import { type ExcludeLiteral } from '~/types/ExcludeLiteral';
 
 const MONDAY_KEY: keyof typeof CalendarStartDay = 'MONDAY';
 const SATURDAY_KEY: keyof typeof CalendarStartDay = 'SATURDAY';
 const SUNDAY_KEY: keyof typeof CalendarStartDay = 'SUNDAY';
 
-export const detectCalendarStartDay = (): keyof typeof CalendarStartDay => {
+export type NonSystemCalendarStartDay = ExcludeLiteral<
```
charlesBochet, line 8: **polish**: type is exported so should be in it's own file
**polish**: folder structure is not respected (utils has a subfolder)

packages/twenty-front/src/modules/localization/utils/formatDateISOStringToDateTime.ts
```diff
@@ -15,6 +15,8 @@ export const formatDateISOStringToDateTime = ({
   timeFormat: TimeFormat;
   localeCatalog: Locale;
 }) => {
+  // TODO: replace this with shiftPointInTimeToFromTimezoneDifference to remove date-fns-tz, which formatInTimeZone is doig under the hood :
```
charlesBochet, line 18: **question:** I thought we need to keep date-fns-tz?

packages/twenty-front/src/modules/object-record/object-filter-dropdown/components/ObjectFilterDropdownDateInput.tsx
```diff
@@ -1,73 +1,88 @@
...
+        : userDefinedCalendarStartDay
+    ) as NonSystemCalendarStartDay;
+
     const newFilterValue = relativeDate
-      ? computeVariableDateViewFilterValue(
-          relativeDate.direction,
-          relativeDate.amount,
-          relativeDate.unit,
-        )
+      ? stringifyRelativeDateFilter({
```
charlesBochet, line 80: **question:** git is a bit lost but this is not a breaking change right? it's just a function rename, we are not changing the way we store the viewFilter?

packages/twenty-front/src/modules/object-record/object-filter-dropdown/hooks/useGetInitialFilterValue.ts
```diff
@@ -0,0 +1,115 @@
...
+import { DATE_TYPE_FORMAT } from 'twenty-shared/constants';
+import { type FilterableAndTSVectorFieldType } from 'twenty-shared/types';
+
+const activeDatePickerOperands = [
+  RecordFilterOperand.IS_BEFORE,
+  RecordFilterOperand.IS,
+  RecordFilterOperand.IS_AFTER,
+];
+
+export const useGetInitialFilterValue = () => {
```
charlesBochet, line 20: **question**: is this only for date and datetime? do we want to introduce a global hook or should we just put it in DateTime and Date local implems

charlesBochet: Edit: found the old file below, LGTM

packages/twenty-front/src/modules/object-record/object-filter-dropdown/hooks/useGetDateTimeFilterDisplayValue.ts
```diff
@@ -0,0 +1,49 @@
...
+export const useGetDateTimeFilterDisplayValue = () => {
+  const {
+    userTimezone,
+    isSystemTimezone,
+    getTimezoneAbbreviationForPointInTime,
+  } = useUserTimezone();
+  const { userDateFormat } = useUserDateFormat();
+  const { userTimeFormat } = useUserTimeFormat();
+
+  const getDateTimeFilterDisplayValue = (correctPointInTime: Date) => {
```
charlesBochet, line 18: **polish**: I don't get the naming "correct" in the parameter name

packages/twenty-front/src/modules/object-record/record-field/ui/form-types/components/__stories__/FormDateFieldInput.stories.tsx
```diff
@@ -1,409 +0,0 @@
...
-  decorators: [I18nFrontDecorator, WorkflowStepDecorator],
-};
-
-export default meta;
-
-type Story = StoryObj<typeof FormDateFieldInput>;
-
-const currentYear = new Date().getFullYear();
-
-export const Default: Story = {
```
charlesBochet, line 32: change: **let's put back the stories**

packages/twenty-front/src/modules/ui/input/components/internal/date/components/TimeZoneAbbreviation.tsx
```diff
@@ -0,0 +1,32 @@
+import { useUserTimezone } from '@/ui/input/components/internal/date/hooks/useUserTimezone';
+import styled from '@emotion/styled';
+
+const StyledTimezoneAbbreviation = styled.span<{ hasError?: boolean }>`
```
charlesBochet, line 4: **polish**: feels that this should not be part of input folder

packages/twenty-front/src/modules/ui/input/components/internal/date/hooks/useParseDateInputStringToJSDate.ts
```diff
@@ -0,0 +1,23 @@
+import { useDateTimeFormat } from '@/localization/hooks/useDateTimeFormat';
+import { isValid, parse } from 'date-fns';
+import { getDateFormatStringForDatePickerInputMask } from '~/utils/date-utils';
+
+export const useParseDateInputStringToJSDate = () => {
```
charlesBochet, line 5: same for all these, would be nice to have a date util / hook api that can be re-used everywhere (out of the input)

packages/twenty-front/src/modules/ui/input/components/internal/date/hooks/useUserDateFormat.ts
```diff
@@ -0,0 +1,14 @@
+import { currentWorkspaceMemberState } from '@/auth/states/currentWorkspaceMemberState';
+import { useRecoilValue } from 'recoil';
+import { WorkspaceMemberDateFormatEnum } from '~/generated/graphql';
+
+export const useUserDateFormat = () => {
```
charlesBochet, line 5: **change:** the 5 utils above seem to be really local to react date picker "hack" but useUserDateFormat is not and should be add to the toolsuite we have to handle dates and times

packages/twenty-shared/src/utils/filter/dates/utils/getPlainDateFromDate.ts
```diff
@@ -0,0 +1,6 @@
+import { DATE_TYPE_FORMAT } from '@/constants';
+import { format } from 'date-fns';
+
+export const getPlainDateFromDate = (date: Date) => {
```
charlesBochet, line 4: polish: we should add test for all these utils

packages/twenty-front/src/modules/views/view-filter-value/utils/stringifyRelativeDateFilter.ts
```diff
@@ -0,0 +1,37 @@
...
+    !isDefined(relativeDateFilter.amount) ||
+    relativeDateFilter.amount <= 0
+  ) {
+    throw new Error(
+      'Amount must be defined and greater than 0 for relative date filters',
+    );
+  }
+
+  if (isNonEmptyString(relativeDateFilter.timezone)) {
+    relativeDateFilterStringified = `${relativeDateFilterStringified};;${relativeDateFilter.timezone};;`;
```
charlesBochet, line 22: @lucasbordeau I am not fan of this format, we should prefer a json stringified format but I understand this avoids breaking changes

charlesBochet: Also, I'm a bit surprised that we need that much information:
tested: PAST_2_DAY;;Europe/Paris;;2025-11-01;;MONDAY;;

why do we also need 2025-11-01 ?

charlesBochet:
@lucasbordeau we also need to fix that: 
image

PR #15383, opened 2025-10-27
fix invalid condition for google refresh token

neo773:


charlesBochet (approved):

prastoin (commented):
Hey there, as greptile says it means we will refresh google tokens on every call
Is that what's intented ?

prastoin:
> Hey there, as greptile says it means we will refresh google tokens on every call Is that what's intented ?

Discussed irl it was intended

PR #15384, opened 2025-10-27
Refactor `flatEntityMaps` to be `universalIdentifier` oriented

prastoin:


prastoin:
As discussed irl I prefer prioritizing universalId and applicationId non nullable refactor introducing the custom default app and twenty-standard app first

PR #15385, opened 2025-10-27
Hide page layout tab bar when there are less than two tabs to display

Devessier:
- DIsplay the tab bar only if more than 1 tabs can be displayed
- Remove the selfDisplayMode feature: the first tab is now considered to be the pinned tab on record pages

https://github.com/user-attachments/assets/3f61931d-fde4-4065-b646-5b3a7bdebcc0


ehconitin (commented):
LGTM code-wise!
I have a few questions regarding the behavior, such as link behavior on record pages.

DMing you the same

ehconitin (approved):
https://github.com/twentyhq/core-team-issues/issues/1784.

PR #15388, opened 2025-10-28
fix: made standard object labels editable

### LINKED ISSUE #15302: [Data Model] - Standard object labels can no longer be edited
## Bug Description

Standard object labels are now gray - on all workspaces - and can no longer be edited in the Data model.
We used to let users edit those to give them the flexibility to have the labels match their business.

ShashwatPS:
Fixes #15302 

## Video ( Fixed ) :

[Uploading IssueFixed.mov…](https://drive.google.com/file/d/1XNk3i3ae9Wafelqgbj5GujaPAJRQRcQh/view?usp=sharing)


prastoin (approved):
Hey there thanks for your contribution.
It does not seem to be bug initially but rather a decision, discussed irl prevented behavior should not be stab

prastoin:
We should test in v1 before merging, doing

prastoin (dismissed):
We should only allow label edition on standard objects

charlesBochet (approved):

PR #15392, opened 2025-10-28
Enhance role-check system with stricter checks

FelixMalfait:
## Overview

This PR strengthens our permission system by introducing more granular role-based access control across the platform.

### New Permissions Added
- **Applications** - Control who can install and manage applications
- **Layouts** - Control who can customize page layouts and UI structure
- **AI** - Control access to AI features and agents
- **Upload File** - Separate permission for file uploads
- **Download File** - Separate permission for file downloads (frontend visibility)

### Security Enhancements
- Implemented whitelist-based validation for workspace field updates
- Added explicit permission guards to core entity resolvers
- Enhanced ESLint rule to enforce permission checks on all mutations
- Created `CustomPermissionGuard` and `NoPermissionGuard` for better code documentation

### Affected Components
- Core entity resolvers: webhooks, files, domains, applications, layouts, postgres credentials
- Workspace update mutations now use whitelist validation
- Settings UI updated with new permission controls

### Developer Experience
- ESLint now catches missing permission guards during development
- Explicit guard markers make permission requirements clear in code review
- Comprehensive test coverage for new permission logic

## Testing
- ✅ All TypeScript type checks pass
- ✅ ESLint validation passes
- ✅ New permission guards properly enforced
- ✅ Frontend UI displays new permissions correctly

## Migration Notes
Existing workspaces will need to assign the new permissions to roles as needed. By default, all new permissions are set to `false` for non-admin roles.


packages/twenty-front/src/modules/settings/roles/role-permissions/permission-flags/components/SettingsRolePermissionsSettingsSection.tsx
```diff
@@ -96,6 +101,36 @@ export const SettingsRolePermissionsSettingsSection = ({
         description: t`Impersonate workspace users`,
         Icon: IconSpy,
       },
+      {
+        key: PermissionFlagType.APPLICATIONS,
```
ijreilly, line 121: Not very important but this should only show if IS_APPLICATION_ENABLED is true i think?

packages/twenty-front/src/modules/settings/roles/role-permissions/permission-flags/components/SettingsRolePermissionsSettingsSection.tsx
```diff
@@ -96,6 +101,36 @@ export const SettingsRolePermissionsSettingsSection = ({
...
+        Icon: IconApps,
+      },
+      {
+        key: PermissionFlagType.LAYOUTS,
+        name: t`Layouts`,
+        description: t`Customize page layouts and UI structure`,
+        Icon: IconLayoutSidebarRightCollapse,
+      },
+      {
+        key: PermissionFlagType.MANAGE_VIEWS,
```
ijreilly: I think if we want to do that one right we need a comprehensive work in the FE. 

For instance at the moment when a user choose to hide or show a field on a view, it updates the view. 
I think a user that does not have the MANAGE_VIEW permission should still be able to display a field that is being hidden, but that it should not update the view for everyone. 

Overall I think the introduction of this permission would be less painful for users who won't have it if we introduce private / common views, otherwise users without the permissions will be restricted in their attempts to display data as they want to (if they need to hide / show a field each time they visit a view for instance)

packages/twenty-server/src/engine/core-modules/billing/billing.resolver.ts
```diff
@@ -59,7 +60,7 @@ export class BillingResolver {
   @Query(() => BillingSessionOutput)
   @UseGuards(
     WorkspaceAuthGuard,
-    SettingsPermissionsGuard(PermissionFlagType.WORKSPACE),
+    SettingsPermissionsGuard(PermissionFlagType.BILLING),
```
ijreilly, line 63: i think that involves a migration if we want to make sure that all roles that enable WORKSPACE permissions now enable BILLING

packages/twenty-server/src/engine/metadata-modules/database-event-trigger/resolvers/database-event-trigger.resolver.ts
```diff
@@ -16,8 +17,12 @@ import { UpdateDatabaseEventTriggerInput } from 'src/engine/metadata-modules/dat
 import { DatabaseEventTriggerEntity } from 'src/engine/metadata-modules/database-event-trigger/entities/database-event-trigger.entity';
 import { DatabaseEventTriggerV2Service } from 'src/engine/metadata-modules/database-event-trigger/services/database-event-trigger-v2.service';
 import { databaseEventTriggerGraphQLApiExceptionHandler } from 'src/engine/metadata-modules/database-event-trigger/utils/database-event-trigger-graphql-api-exception-handler.utils';
+import { PermissionFlagType } from 'src/engine/metadata-modules/permissions/constants/permission-flag-type.constants';
 
-@UseGuards(WorkspaceAuthGuard)
+@UseGuards(
+  WorkspaceAuthGuard,
+  SettingsPermissionsGuard(PermissionFlagType.WORKFLOWS),
```
ijreilly, line 24: why is this linked to workflows?

FelixMalfait: @thomtrp could you please double check? I assumed it was but if you can double check I'm not missing anything, thanks!

packages/twenty-server/src/engine/core-modules/billing/billing.resolver.ts
FelixMalfait: Ideally we should yes but I think we can discard that, people who need it today will most likely have "all settings"

packages/twenty-front/src/modules/settings/roles/role-permissions/permission-flags/components/SettingsRolePermissionsSettingsSection.tsx
FelixMalfait: Good point! Done. Same for AI

packages/twenty-front/src/modules/settings/roles/role-permissions/permission-flags/components/SettingsRolePermissionsSettingsSection.tsx
FelixMalfait: I've made two changes that should help:
- Moved Views from Settings to Tools which means it's now provisioned by default for Member roles (including existing)
- Made changes on the frontend so that changes are not persisted when the permission is missing, that way the user can still resize or move columns but it's only a local change

I agree private views would be nice as a followup.

packages/twenty-server/src/engine/metadata-modules/database-event-trigger/resolvers/database-event-trigger.resolver.ts
thomtrp: We are planning to migrate workflows on that entity but for now I think this is only used by applications. Could you confirm @martmull?

FelixMalfait:
TODO: Don't load agent threads/ask ai if there's no AI permission

charlesBochet:
@FelixMalfait, I'm closing it, I know that you'll work on it soon, let's just re-open it once you are!

packages/twenty-server/src/database/typeorm/core/migrations/common/1762351626807-viewVisibility.ts
```diff
@@ -0,0 +1,37 @@
...
+      `CREATE TYPE "core"."view_visibility_enum" AS ENUM('WORKSPACE', 'UNLISTED')`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "core"."view" ADD "visibility" "core"."view_visibility_enum" NOT NULL DEFAULT 'WORKSPACE'`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "core"."view" ADD "createdByUserWorkspaceId" uuid`,
+    );
+    await queryRunner.query(
+      `CREATE INDEX "IDX_VIEW_VISIBILITY" ON "core"."view" ("visibility") `,
```
charlesBochet, line 17: @FelixMalfait I wonder if this is the right index, I expect the queries to be made on workspaceId + visibility which would require a composite index. Can you double check the queries that will be made?

packages/twenty-server/src/engine/metadata-modules/database-event-trigger/resolvers/database-event-trigger.resolver.ts
martmull: It is currently not linked to workflows. It is only used to create database event trigger for serverless functions

image

PR #15393, opened 2025-10-28
Remove deprecated `view` and related workspace entities

prastoin:
# Introduction
A while ago we migrated view from workspace to metadata
Their standard objects workspace entities declaration remained we can now remove them

## Deprecating commands before 1.5
The view migration command from workspace to metadata was introduced in `1.5.0`. Removing the `baseWorkspaceEntity` make this command obsolete.
If tomorrow twenty handles auto upgrade in latest and a user having an instance in `1.3.0` starts auto-upgrading he won't be able to migrate his views ( that's why we should not support upgrade before 1.5 anymore here )
We will have the same use case with FavoritesFolders


prastoin:
@greptileai trigger, please let me know what you think removing the commands before 1.5

prastoin:
> What happens if someone tries to upgrade from 1.3 → latest?

Please not you 😆

prastoin:
Fixing tests

martmull (approved):
lgtm

PR #15414, opened 2025-10-28
Complete color refactoring

### LINKED ISSUE #1779: Use Radix colors
Replace all colors in the theme by radix colors.

https://www.radix-ui.com/colors

bosiraphael:
# Complete color refactoring

Closes https://github.com/twentyhq/core-team-issues/issues/1779

- Updated all colors to use Radix colors with P3 color space allowing for brighter colors
- Created our own gray scale interpolated on Radix's scale to have the same values for grays as the old ones in the app
- Introduced dark and light colors as well as there transparent versions
- Added many new colors from radix that can be used in the tags or in the graphs
- Updated multiple color utilities to match new behaviors
- Changed the computation of Avatar colors to return only colors from the theme (before it was random hsl)

These changes allow the user to use new colors in tags or charts, the colors are brighter and with better contrast. We have a full range of color variations from 1 to 12 where before we only had 4 adaptative colors.
All these changes will allow us to develop custom themes for the user soon, where users can choose their accent colors, background colors and there contrast.


bosiraphael:
@greptileai

packages/twenty-ui/src/utilities/color/utils/stringToThemeColorP3String.ts
```diff
@@ -0,0 +1,31 @@
...
+  for (let i = 0; i < string.length; i++) {
+    hash = string.charCodeAt(i) + ((hash << 5) - hash);
+  }
+
+  const colorIndex = Math.abs(hash) % MAIN_COLOR_NAMES.length;
+  return MAIN_COLOR_NAMES[colorIndex];
+};
+
+export const stringToThemeColorValue = ({
+  string,
```
lucasbordeau, line 14: Could we use a better naming ? Is it a hexColorString ? Or maybe an identifierString ? I think we could provide more context here.

lucasbordeau (approved):
Ok

packages/twenty-ui/src/utilities/color/utils/stringToThemeColorP3String.ts
bosiraphael: It is a p3 scale string from the theme

bosiraphael: I'm renaming the util

PR #15415, opened 2025-10-28
Fix infinite loop on FE

### LINKED ISSUE #15369: Tab crashes on custom object with no record view creation
Reproduction:
- create object ( here bug )
- create field text ( here www )
- exit settings
- navigate to bug view
- create a view ( here table prastoin )
- tab crashes. ~~`findManyBugs` infinite loops~~ server seems to send back a response might be an infinite loop front side

Video [here](https://github.com/twentyhq/private-issues/issues/334)

charlesBochet:
Fixes https://github.com/twentyhq/twenty/issues/15369

It's the first time we are hitting this weird behavior on apollo onCompleted / onError. Theoretically if the reference of the onCompleted (resp onError) callback is changing, this will trigger a re-run of onCompleted. But also theretically the reference here is a recoilCallback so should never change.

As this is synchronous I have move this logic to a sync way

### DIFF
--- a/packages/twenty-front/src/modules/object-record/hooks/useLazyFindManyRecords.ts
+++ b/packages/twenty-front/src/modules/object-record/hooks/useLazyFindManyRecords.ts
@@ -14,6 +14,7 @@ import { cursorFamilyState } from '@/object-record/states/cursorFamilyState';
 import { hasNextPageFamilyState } from '@/object-record/states/hasNextPageFamilyState';
 import { type ObjectRecord } from '@/object-record/types/ObjectRecord';
 import { getQueryIdentifier } from '@/object-record/utils/getQueryIdentifier';
+import { useEffect } from 'react';
 import { QUERY_DEFAULT_LIMIT_RECORDS } from 'twenty-shared/constants';
 import { isDefined } from 'twenty-shared/utils';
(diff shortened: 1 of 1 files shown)


packages/twenty-front/src/modules/object-record/hooks/useLazyFindManyRecords.ts
```diff
@@ -126,6 +124,13 @@ export const useLazyFindManyRecords = <T extends ObjectRecord = ObjectRecord>({
         }
 
         const result = await findManyRecords();
+        if (isDefined(result?.error)) {
```
prastoin, line 127: ***Question:*** Couldn't both be defined ? Also what if none are

prastoin (commented):
Nice catch ! thanks for investigating that
Could this be applicable somewhere else in the app ?

packages/twenty-front/src/modules/object-record/hooks/useLazyFindManyRecords.ts
charlesBochet: I haven't actually tested the behavior with error, I can revisit that!

prastoin (approved):

PR #15417, opened 2025-10-28
Introduce SSO bypass permission.

### LINKED ISSUE #1772: Introduce SSO Bypass permission
I think we need the concept of an "SSO Bypass Permission" with different allowed connection methods associated to it.

It would be a small link "SSO Bypass" when you're on a workspace subdomain. Maybe to replace the terms/privacy text.  

And then in settings we should be allow to customize login methods for users who have SSO bypas permissions.

That way the root user can still login with password (or Google or MS) if SSO is broken.

Image

Let's only display it if there is 1 just 1 login method and it's SSO

mabdullahabaid:
Closes [Core Issue #1772](https://github.com/twentyhq/core-team-issues/issues/1772). 

---

> [!NOTE]
> Introduces SSO bypass with a new permission flag and workspace-level provider toggles, enabling permitted users to log in via Google/Microsoft/Password when SSO-only, with backend enforcement and frontend UI/hooks/queries.
> 
> - **Backend**:
>   - **Permission & Enforcement**: Add `PermissionFlagType.SSO_BYPASS`; update `AuthService` to allow login via non-SSO providers when workspace bypass is enabled and user has `SSO_BYPASS`.
>   - **Workspace Model**: Add `isGoogleAuthBypassEnabled`, `isMicrosoftAuthBypassEnabled`, `isPasswordAuthBypassEnabled` (migration, entity, update input, service validation).
>   - **Public API**: Extend `PublicWorkspaceDataOutput` with `authBypassProviders`; resolver computes it; permissions defaults include `SSO_BYPASS`.
> - **Frontend**:
>   - **GraphQL/State**: Generate new types/fields; add `authBypassProviders` to `GetPublicWorkspaceDataByDomain`; new states `workspaceAuthBypassProvidersState`, `workspaceBypassModeState`.
>   - **Auth UI/Logic**: Add `useWorkspaceBypass`; update sign-in form and footer to offer "Bypass SSO" and use merged providers when enabled; remove auto-redirect when single SSO.
>   - **Settings**: Add Security section to toggle bypass methods per provider; conditionally show Change Password via `useCanChangePassword`.
>   - **Tests/Mocks**: Update mocks and tests to include bypass flags/providers.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8c393b2bad387fb6e8b8f40027f8637dd6e85723. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/twenty-front/src/modules/settings/security/components/SettingsSecurityAuthBypassOptionsList.tsx
```diff
@@ -0,0 +1,206 @@
...
+      });
+    } catch (err) {
+      setCurrentWorkspace(previousWorkspace);
+      enqueueErrorSnackBar({
+        apolloError: err instanceof ApolloError ? err : undefined,
+      });
+    }
+  };
+
+  const visibleToggles = useMemo(() => {
```
FelixMalfait: Let's not use useMemo unless there is a real performance issue

packages/twenty-front/src/modules/settings/security/components/SettingsSecurityAuthBypassOptionsList.tsx
```diff
@@ -0,0 +1,206 @@
...
+        Icon: IconPassword,
+        providerKey: 'password',
+        title: t`Password`,
+        description: t`Allow email and password login for users with SSO bypass permissions.`,
+      },
+    ],
+    [t],
+  );
+
+  useEffect(() => {
```
FelixMalfait: Let's not use `useEffect` unless we are strictly forced to, use `onChange` or other trigger-based methods instead

packages/twenty-front/src/modules/auth/sign-in-up/hooks/useWorkspaceBypass.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    return google || microsoft || password;
+  }, [workspaceAuthBypassProviders]);
+
+  const shouldOfferBypass =
+    isWorkspaceContext &&
+    hasOnlySSOProvidersEnabled &&
+    hasBypassProvidersAvailable;
+
+  useEffect(() => {
```
FelixMalfait: We only allow effects in dedicated Effect components or in a very specific cases ; in most case case you shouldn't even need the effect component

https://react.dev/learn/you-might-not-need-an-effect

packages/twenty-front/src/modules/auth/sign-in-up/hooks/useWorkspaceBypass.ts
```diff
@@ -0,0 +1,78 @@
...
+    if (!shouldOfferBypass && workspaceBypassMode) {
+      setWorkspaceBypassMode(false);
+    }
+  }, [workspaceBypassMode, setWorkspaceBypassMode, shouldOfferBypass]);
+
+  const enableBypass = useCallback(() => {
```
FelixMalfait: Why useCallback?

packages/twenty-front/src/modules/auth/sign-in-up/hooks/useWorkspaceBypass.ts
```diff
@@ -0,0 +1,78 @@
...
+    workspaceBypassModeState,
+  );
+
+  const { isOnAWorkspace } = useIsCurrentLocationOnAWorkspace();
+  const isMultiWorkspaceEnabled = useRecoilValue(isMultiWorkspaceEnabledState);
+
+  const isWorkspaceContext =
+    isOnAWorkspace || isMultiWorkspaceEnabled === false;
+
+  const hasOnlySSOProvidersEnabled = useMemo(() => {
```
FelixMalfait: Why useMemo?

packages/twenty-front/src/modules/settings/security/components/SettingsSecurityAuthBypassOptionsList.tsx
```diff
@@ -0,0 +1,119 @@
...
+  };
+
+  const toggleAuthBypassMethod = async (
+    authProvider: keyof Omit<AuthProviders, '__typename' | 'magicLink' | 'sso'>,
+  ) => {
+    if (!currentWorkspace?.id) {
+      throw new Error(t`User is not logged in`);
+    }
+
+    const key = `is${capitalize(authProvider)}AuthBypassEnabled`;
```
FelixMalfait, line 49: I would try to avoid this kind of optimization, you prefer to keep something typed and explicit. It might look "smart" and repeating a few things might look dumb but here you're making it much more likely to create bugs if we refactor this in the future (capitalize make it very unstable)

packages/twenty-front/src/modules/users/hooks/useLoadCurrentUser.ts
```diff
@@ -103,6 +109,22 @@ export const useLoadCurrentUser = () => {
 
     setCurrentWorkspace(workspace);
 
+    if (!isDefined(workspace) || !authProviders) {
+      setWorkspaceAuthBypassProviders(null);
+    } else {
+      setWorkspaceAuthBypassProviders({
+        google: Boolean(
```
FelixMalfait: Boolean conversion like that is not a standard codebase patterns. Rather than forcing casting you should use the right utils (isDefined, isUndefinedOrNull, etc.)

packages/twenty-front/src/modules/users/hooks/useLoadCurrentUser.ts
```diff
@@ -103,6 +109,22 @@ export const useLoadCurrentUser = () => {
 
     setCurrentWorkspace(workspace);
 
+    if (!isDefined(workspace) || !authProviders) {
```
FelixMalfait: Why do we need this? This if/else structure looks odd, why don't we need it for other vars?

Generally prefer early returns to if/else (but sometimes if/else are acceptable)

packages/twenty-front/src/pages/settings/security/SettingsSecurity.tsx
```diff
@@ -82,6 +87,22 @@ export const SettingsSecurity = () => {
     saveWorkspace(value);
   };
 
+  const hasSsoIdentityProviders = SSOIdentitiesProviders.length > 0;
+  const hasDirectAuthEnabled = Boolean(
```
FelixMalfait: Like in the other comment, don't force cast to boolean

packages/twenty-front/src/modules/settings/security/components/SettingsSecurityAuthBypassOptionsList.tsx
mabdullahabaid: I did this because SettingsSecurityAuthProvidersOptionsList does the same thing - I tried to match SettingsSecurityAuthBypassOptionsList to SettingsSecurityAuthProvidersOptionsList as much as possible so refactor and reading code is consistent later.

Will change if you'd prefer to change.

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
```diff
@@ -131,17 +134,53 @@ export class AuthService {
...
     targetWorkspace?: WorkspaceEntity,
   ) {
-    if (targetWorkspace && !targetWorkspace.isPasswordAuthEnabled) {
-      throw new AuthException(
-        'Email/Password auth is not enabled for this workspace',
-        AuthExceptionCode.FORBIDDEN_EXCEPTION,
-      );
-    }
-
     const user = await this.userRepository.findOne({
```
FelixMalfait, line 141: By moving the condition check below you're probably allowing an attacker to know if a user exists or not on an SSO-only workspace which can be considered a security issue

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
```diff
@@ -157,6 +196,21 @@ export class AuthService {
     }
 
     if (targetWorkspace) {
+      const canBypass =
+        !targetWorkspace.isPasswordAuthEnabled &&
```
FelixMalfait: Why add ` !targetWorkspace.isPasswordAuthEnabled` here  and not just canBypass = await...

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
```diff
@@ -131,17 +134,53 @@ export class AuthService {
     );
   }
 
+  private async canUserBypassAuthProvider({
```
FelixMalfait: I don't understand, is this used for Google/Microsoft? I only see it in Password

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
```diff
@@ -131,17 +134,53 @@ export class AuthService {
...
+        default:
+          return false;
+      }
+    })();
+
+    if (!isBypassEnabledForProvider) {
+      return false;
+    }
+
+    const userWorkspace =
```
FelixMalfait: Can't we access it from user directly instead of doing one more db request?

FelixMalfait: Also this is poorly named (not you) if this is called checkXXXExists but is what should be use for the find method!

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
```diff
@@ -230,7 +284,28 @@ export class AuthService {
     }
 
     if (isDefined(workspace)) {
-      workspaceValidator.isAuthEnabledOrThrow(authParams.provider, workspace);
+      const isProviderEnabled = workspaceValidator.isAuthEnabled(
+        authParams.provider,
+        workspace,
+      );
+
+      if (!isProviderEnabled) {
```
FelixMalfait: You should instead do an early return before if provider is enabled? Here you'll do one db query for each login while it's not needed in 99% of cases because they didn't go through the bypass form

packages/twenty-front/src/modules/auth/sign-in-up/components/SignInUpWorkspaceScopeForm.tsx
```diff
@@ -33,26 +39,33 @@ export const SignInUpWorkspaceScopeForm = () => {
     return null;
   }
 
+  const providers =
+    shouldOfferBypass && shouldUseBypass
+      ? {
+          ...workspaceAuthProviders,
+          ...workspaceAuthBypassProviders,
```
FelixMalfait, line 45: This seems right/to be working but intuitively I would have expected  `...workspaceAuthBypassProviders`to be enough (no need for `...workspaceAuthProviders`). What am I missing?

packages/twenty-front/src/modules/settings/security/components/SettingsSecurityAuthBypassOptionsList.tsx
FelixMalfait: Ah that's a good reason to do it indeed!
Yes if you can update both file to do something less engineered that'd be great

FelixMalfait: Screenshot 2025-10-31 at 16 54 12
(e.g. by Cursor when asking it to do it)

mabdullahabaid: Sure, thank you!

packages/twenty-front/src/modules/auth/sign-in-up/components/SignInUpWorkspaceScopeForm.tsx
mabdullahabaid: You're right. Expanding `workspaceAuthProviders` here is redundant and maybe even introduces a minor risk of displaying providers we do not want to display (chances of a mistake during refactor).

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
mabdullahabaid: As discussed on Discord, the attacker can also find this information using checkUserExists query.

Secondly, allowing bypass without making the code more messy is kind of hard - discussed on Discord on how we need a flag because we have no centralized context and everything is tightly coupled.

packages/twenty-front/src/modules/users/hooks/useLoadCurrentUser.ts
mabdullahabaid: Removed this and replaced with the following.

```
    if (isDefined(workspace)) {
      setWorkspaceAuthBypassProviders({
        google: authProviders.google && workspace.isGoogleAuthBypassEnabled,
        microsoft:
          authProviders.microsoft && workspace.isMicrosoftAuthBypassEnabled,
        password:
          authProviders.password && workspace.isPasswordAuthBypassEnabled,
      });
    }
```

We need to check for workspace definition since if we look two lines above, workspace can be null.

```
const workspace = user.currentWorkspace ?? null;
```

authProviders check was redundant - added it to wait for authProviders from the config, but they're preloaded and removing the check does not impact anything.

packages/twenty-front/src/modules/auth/sign-in-up/components/SignInUpWorkspaceScopeForm.tsx
mabdullahabaid: Fixed.

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
mabdullahabaid: Removed the nested if statement and also fixed this.

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
mabdullahabaid: Done. Had a separate fetch since Google and Microsoft did not fetch userWorkspace relation. 

However, updated those to fetch workspace when fetching user by email and then replaced the DB query with find.

mabdullahabaid: In terms of naming, since we are not checking in the DB, I maintained the name of userWorkspace.

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
mabdullahabaid: Added the early return.

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
mabdullahabaid: `signInUpWithSocialSSO` method calls `signInUp` which in turn calls `isAuthProviderEnabledOrThrow` which calls `canUserBypassAuthProvider`.

Password flow was not dependent on `isAuthProviderEnabledOrThrow`, so I had to add `canUserBypassAuthProvider` separately in `validateLoginWithPassword`.

It's a tightly-coupled code problem, auth needs extremely careful rework to fix and not introduce regressions.

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
```diff
@@ -230,10 +241,74 @@ export class AuthService {
...
+        (await this.canUserBypassAuthProvider({
+          user: existingUser,
+          workspace,
+          provider: authParams.provider,
+        }))
+      ) {
+        return;
+      }
+
       workspaceValidator.isAuthEnabledOrThrow(authParams.provider, workspace);
```
mabdullahabaid, line 267: This is redundant to avoid introducing multiple sources of error for the same thing (isAuthEnabled used earlier in the function).

I tried to avoid, but then an explicit throw had to be used and the error messages would need to be maintained in multiple places.

packages/twenty-front/src/modules/auth/sign-in-up/hooks/useWorkspaceBypass.ts
```diff
@@ -0,0 +1,55 @@
+import { useRecoilState, useRecoilValue } from 'recoil';
+
+import { useIsCurrentLocationOnAWorkspace } from '@/domain-manager/hooks/useIsCurrentLocationOnAWorkspace';
+import { workspaceAuthBypassProvidersState } from '@/workspace/states/workspaceAuthBypassProvidersState';
+import { workspaceAuthProvidersState } from '@/workspace/states/workspaceAuthProvidersState';
+import { workspaceBypassModeState } from '@/workspace/states/workspaceBypassModeState';
+
+export const useWorkspaceBypass = () => {
```
FelixMalfait, line 8: See this comment from Charles: https://github.com/twentyhq/twenty/pull/15035#issuecomment-3394142471 -> avoid umbrella hooks.

It's acceptable in that case in my opinion because it's very tightly coupled, but be careful with this pattern and think about re-renders :)

FelixMalfait (approved):
great work!

PR #15424, opened 2025-10-29
Move http node to backend

### LINKED ISSUE #14491: HTTP Requests should always be sent from the server when testing.
## Bug Description

When testing an HTTP Request to an API in Workflows, the request is sent by the browser, causing an "origin" header to be present, which some APIs do not like. When testing the entire workflow though, the request is sent by the server without an "origin" header, leading to a successful API Request.

## Expected behavior

The API request should always be sent by the backend to have a consistent and accurate experience when testing workflows.

## Technical inputs

-

thomtrp:
Fixes https://github.com/twentyhq/twenty/issues/14491

Also adding an url validator.

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/tool/tool.module.ts
+++ b/packages/twenty-server/src/engine/core-modules/tool/tool.module.ts
@@ -1,0 +1,21 @@
+import { Module } from '@nestjs/common';
+import { TypeOrmModule } from '@nestjs/typeorm';
+
+import { FileEntity } from 'src/engine/core-modules/file/entities/file.entity';
+import { FileModule } from 'src/engine/core-modules/file/file.module';
+import { ToolRegistryService } from 'src/engine/core-modules/tool/services/tool-registry.service';
+import { HttpTool } from 'src/engine/core-modules/tool/tools/http-tool/http-tool';
+import { SearchArticlesTool } from 'src/engine/core-modules/tool/tools/search-articles-tool/search-articles-tool';
+import { SendEmailTool } from 'src/engine/core-modules/tool/tools/send-email-tool/send-email-tool';
+import { MessagingImportManagerModule } from 'src/modules/messaging/message-import-manager/messaging-import-manager.module';
+
+@Module({
+  imports: [
+    MessagingImportManagerModule,
+    TypeOrmModule.forFeature([FileEntity]),
+    FileModule,
+  ],
+  providers: [HttpTool, SendEmailTool, SearchArticlesTool, ToolRegistryService],
+  exports: [ToolRegistryService],
+})
+export class ToolModule {}
--- a/packages/twenty-server/src/engine/core-modules/tool/tools/http-tool/http-tool.ts
+++ b/packages/twenty-server/src/engine/core-modules/tool/tools/http-tool/http-tool.ts
@@ -1,5 +1,8 @@
 import { Injectable } from '@nestjs/common';
 
+import http from 'http';
+import https from 'https';
+
 import axios, { type AxiosRequestConfig } from 'axios';
 import { isDefined } from 'twenty-shared/utils';
 import { parseDataFromContentType } from 'twenty-shared/workflow';
@@ -9,7 +12,7 @@ import { type HttpRequestInput } from 'src/engine/core-modules/tool/tools/http-t
 import { type ToolInput } from 'src/engine/core-modules/tool/types/tool-input.type';
 import { type ToolOutput } from 'src/engine/core-modules/tool/types/tool-output.type';
 import { type Tool } from 'src/engine/core-modules/tool/types/tool.type';
-import { isSafeURL } from 'src/engine/core-modules/tool/utils/is-safe-url.util';
+import { getSafeUrlIP } from 'src/engine/core-modules/tool/utils/get-safe-url-ip.util';
 
 @Injectable()
 export class HttpTool implements Tool {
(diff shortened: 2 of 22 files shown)


thomtrp:
@greptileai trigger

packages/twenty-server/src/engine/core-modules/tool/tools/http-tool/http-tool.ts
```diff
@@ -21,12 +26,50 @@ export class HttpTool implements Tool {
...
+
+          if (options.all) {
+            callback(null, [
+              { address: safeUrlIP, family: options.family || 4 },
+            ]);
+          } else {
+            callback(null, safeUrlIP, options.family || 4);
+          }
+        },
+      });
```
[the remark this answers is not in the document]
thomtrp: this is wrong. The url hostname is not used, since lookup function has been overridden.

thomtrp:
@greptileai trigger

packages/twenty-server/src/engine/core-modules/tool/tool.module.ts
```diff
@@ -0,0 +1,20 @@
...
+@Module({
+  imports: [
+    MessagingImportManagerModule,
+    TypeOrmModule.forFeature([FileEntity]),
+    FileModule,
+  ],
+  providers: [HttpTool, SendEmailTool, ToolRegistryService],
+  exports: [ToolRegistryService],
+})
+export class ToolModule {}
```
Weiko, line 21: As discussed, would prefer a better name but not a blocker for now!

packages/twenty-server/src/engine/core-modules/tool/tools/http-tool/http-tool.ts
```diff
@@ -21,12 +26,53 @@ export class HttpTool implements Tool {
     const isMethodForBody = ['POST', 'PUT', 'PATCH'].includes(method);
 
     try {
+      const parsedUrl = new URL(url);
+
+      if (!['http:', 'https:'].includes(parsedUrl.protocol)) {
+        throw new Error('Invalid URL protocol');
```
charlesBochet: CustomException with friendlyMessage

packages/twenty-server/src/engine/core-modules/tool/tools/http-tool/http-tool.ts
```diff
@@ -21,12 +26,53 @@ export class HttpTool implements Tool {
...
+      }
+
       const axiosConfig: AxiosRequestConfig = {
         url,
         method: method,
         headers: headersCopy,
       };
 
+      if (!process.env.DISABLE_SSRF_PROTECTION) {
```
charlesBochet: HTTP_TOOL_SAFE_MODE_ENABLED

packages/twenty-server/src/engine/core-modules/tool/utils/get-safe-url-ip.util.ts
```diff
@@ -0,0 +1,150 @@
...
+    /^0x7f\./i.test(addr) ||
+    /^fe80::1$/i.test(addr) ||
+    /^::1$/.test(addr) ||
+    /^::$/.test(addr)
+  );
+};
+
+const normalizeToLong = (addr: string) => {
+  const parts = addr.split('.').map((part) => {
+    // Handle hexadecimal format
```
charlesBochet: remove comments

packages/twenty-server/src/engine/core-modules/twenty-config/config-variables.ts
```diff
@@ -66,6 +66,15 @@ export class ConfigVariables {
...
   IS_EMAIL_VERIFICATION_REQUIRED = false;
 
+  @ConfigVariablesMetadata({
+    group: ConfigVariablesGroup.OTHER,
+    description:
+      'Disable SSRF protection checks for HTTP requests (allows private IPs and other security risks)',
+    type: ConfigVariableType.BOOLEAN,
+  })
+  @IsOptional()
+  DISABLE_SSRF_PROTECTION = false;
```
charlesBochet: default true?

packages/twenty-server/src/engine/core-modules/workflow/resolvers/workflow-version-step.resolver.ts
```diff
@@ -142,4 +147,37 @@ export class WorkflowVersionStepResolver {
...
+        success: result.success,
+        message: result.message,
+        result: result.result,
+        error: result.error,
+        status: result.status,
+        statusText: result.statusText,
+        headers: result.headers,
+      };
+    } catch (error) {
+      return {
```
charlesBochet: shouldn't this be in an exception filter (and catch CustomException)

packages/twenty-server/src/engine/core-modules/tool/tools/http-tool/http-tool.ts
thomtrp: this error is catch bellow. Only updating the message then

packages/twenty-server/src/engine/core-modules/twenty-config/config-variables.ts
thomtrp: inverted so yes :D

Weiko: Reading "POSITIVE" form is more straightforward imho, I'd prefer ENABLE(D) (this also makes condition check simpler in the codebase) @thomtrp @charlesBochet wdyt

packages/twenty-server/src/engine/core-modules/tool/utils/get-secure-axios-adapter.util.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    if (!['http:', 'https:'].includes(url.protocol)) {
+      throw new Error('URL should use http/https protocol');
+    }
+
+    const { hostname } = url;
+
+    const { address: resolvedIp } = await dns.lookup(hostname);
+
+    const isSafeModeEnabled = process.env.HTTP_TOOL_SAFE_MODE_ENABLED ?? true;
```
charlesBochet: we should use config variable module here. The issue is that we are in an util i guess can't we pass the variable from the service to this util as props? might be too heavy?

packages/twenty-server/src/engine/core-modules/tool/tools/http-tool/http-tool.ts
```diff
@@ -36,12 +38,19 @@ export class HttpTool implements Tool {
         }
       }
 
-      const response = await axios(axiosConfig);
+      const safeAxiosClient = axios.create({
```
charlesBochet: we could pass the config variable here (see comment above)

charlesBochet:
Tested LGTM
image

charlesBochet:
https://github.com/user-attachments/assets/eac81fa1-9538-4aa2-a66c-5d53f2f0344a

charlesBochet:
image

charlesBochet (approved):

PR #15433, opened 2025-10-29
groupBy fix + typeMapper fix

ijreilly:
groupBy fix: when a group's dimension value is NULL, we need to adapt the raw sql (stage: NULL -> stage IS NULL)

typeMapper fix: a graphql type should be made non-nullable if was indicated so + does not have a default value. our check on not having a default value was limited to having a null defaultValue instead of having a null or undefined defaultValue. This is a breaking change, but all the queries that were providing a null value for these args were not functioning anyway, and luckily in the FE we declared all queries adding a `!` already.

### DIFF
--- a/packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
+++ b/packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
@@ -221,10 +221,15 @@ export class GroupByWithRecordsService {
       const conditions = groupByDefinitions
         .map((def, defIndex) => {
           const paramName = `groupValue_${groupIndex}_${defIndex}`;
+          const paramValue = group[def.alias];
 
-          queryBuilder.setParameter(paramName, group[def.alias]);
+          if (paramValue === null || paramValue === undefined) {
+            return `${def.expression} IS NULL`;
+          } else {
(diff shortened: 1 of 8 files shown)


packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
```diff
@@ -221,10 +221,15 @@ export class GroupByWithRecordsService {
       const conditions = groupByDefinitions
         .map((def, defIndex) => {
           const paramName = `groupValue_${groupIndex}_${defIndex}`;
+          const paramValue = group[def.alias];
 
-          queryBuilder.setParameter(paramName, group[def.alias]);
+          if (paramValue === null || paramValue === undefined) {
```
Weiko: nit: use isDefined

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
```diff
@@ -221,10 +221,15 @@ export class GroupByWithRecordsService {
...
+            return `${def.expression} IS NULL`;
+          } else {
```
Weiko: nit: else can be omit since you return above, up to you!

Weiko (approved):
LGTM!

PR #15442, opened 2025-10-29
Make upsert action body common with create record instead of update record

thomtrp:
We do not want the fields to update multiselect for upsert record action. We want all available fields displayed by default.
This makes upsert record action closer to create record than update record.

This PR: 
- deletes WorkflowUpdateRecordBody that was common between update and upsert and put back content into update
- creates WorkflowCreateRecordBody that is now common between create and upsert
- simplifies shouldDisplayFormField

Before - using fields to update as update record action
Capture d’écran 2025-10-30 à 10 00 24

After - displaying all fields as create record action
Capture d’écran 2025-10-30 à 10 00 04


Devessier (approved):
👍

PR #15443, opened 2025-10-29
feat: Migrate documentation to Mintlify and implement Helper Agent with search functionality

abdulrahmancodes:


packages/twenty-server/src/engine/metadata-modules/agent/agent-tool-generator.service.ts
```diff
@@ -34,6 +40,14 @@ export class AgentToolGeneratorService {
     let tools: ToolSet = {};
 
     try {
+      const agent = await this.agentRepository.findOne({
+        where: { id: agentId },
+      });
+
+      if (agent?.standardId === HELPER_AGENT.standardId) {
```
FelixMalfait, line 47: I feel like this should have been a Tool Permission ; same for HTTP which we could also migrate to a Tool.

It works well from the backend perspective I think, it's more elegant.
I don't think it works as well from a frontend/ux perspective (a bit strange to have a permission to Send HTTP Request and to Search Help Center for traditional users but why not).

These changes will likely conflict with one of my ongoing PR where I touch this Permission page a bit

packages/twenty-server/src/engine/workspace-manager/workspace-sync-metadata/standard-agents/agents/helper-agent.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export const HELPER_AGENT: StandardAgentDefinition = {
+  standardId: '20202020-0002-0001-0001-000000000004',
+  name: 'helper',
+  label: 'Helper',
+  description:
+    'AI agent specialized in helping users learn how to use Twenty CRM',
+  icon: 'IconHelp',
+  applicationId: null,
+  createHandoffFromDefaultAgent: true,
```
FelixMalfait, line 11: Is createHandoffFromDefaultAgent still relevant or should we remove it?

packages/twenty-server/src/engine/core-modules/tool/tools/search-articles-tool/search-articles-tool.ts
```diff
@@ -0,0 +1,78 @@
...
+      if (!MINTLIFY_API_KEY) {
+        return {
+          success: false,
+          message: 'Mintlify API key not configured',
+          error:
+            'The documentation search service is not properly configured. Please contact your administrator.',
+        };
+      }
+      const response = await axios.post(
+        `https://api-dsc.mintlify.com/v1/search/${MINTLIFY_SUBDOMAIN}`,
```
FelixMalfait: I think we should create a proxy on twenty-help.com so that self-hosters can also have access to this without passing an API key ; so let's fallback to this address if no API key is provided. We can then setup a Cloudflare worker that proxies Mintlify

packages/twenty-server/src/engine/workspace-manager/workspace-sync-metadata/standard-agents/agents/helper-agent.ts
abdulrahmancodes: Hmm, not entirely sure. The agent handoff tools are still in use, and removing this would prevent handoffs from being created. So we’d either need to remove all the agent handoff-related code or keep it as is

packages/twenty-server/src/engine/metadata-modules/agent/agent-tool-generator.service.ts
abdulrahmancodes: I think tool permissions are currently a bit messy because they mix two quite different things:  
1. User actions (import/export CSV)  
2. Agent capabilities (search articles, HTTP)  

It doesn’t really make sense to expose agent tools in the permissions settings page — users shouldn’t need to care about a “search help articles” permission since that’s more of an internal agent concern.  

Also, our current agent-tool assignment feels inconsistent. Some are hardcoded, some are permission-based, and it’s not clear to users why a particular agent has access to specific tools. There’s also no visibility in the UI around agent-tool relationships.  

I think we should ideally split this:  
- keep tool permissions for user-facing actions only (import/export, send email)  
- create a separate agent tools system configured at the agent level, not in role permissions  
- expose this in the UI so users can see and configure which tools each agent has access to  

That said, this would be a larger architectural change. For this PR, should we just stick with the current approach for now and plan the refactor separately? Or would you prefer to handle it within the tool permissions for now?

packages/twenty-server/src/engine/workspace-manager/workspace-sync-metadata/standard-agents/agents/helper-agent.ts
FelixMalfait: the "default agent" is what confused me here ; for me there was no default agent anymore?

packages/twenty-server/src/engine/metadata-modules/agent/agent-tool-generator.service.ts
FelixMalfait: Agree with you but this is the part where I have doubts:

> create a separate agent tools system configured at the agent level, not in role permissions

We still haven't nailed the UX for agent <> role creation ; I'm okay to merge like this an continue the discussion

FelixMalfait (approved):

Issue #14995, opened 2025-10-08
Settings AI agent detail crashes when no AI models are configured (Select options empty leads to .Icon access on undefined)


s-yhc:
**Summary:**

Navigating to Settings > AI > Agent Detail (/settings/ai/agents/:agentId) crashes when no AI models are configured for the workspace.
Root cause: The shared Select component assumes at least one option or an emptyOption; when both are absent, it computes an undefined selectedOption, then SelectControl tries to render selectedOption.Icon.

**Impacted areas:**
packages/twenty-front/src/pages/settings/ai/SettingsAgentForm.tsx
packages/twenty-front/src/pages/settings/ai/forms/components/SettingsAIAgentForm.tsx
packages/twenty-front/src/modules/ui/input/components/Select.tsx
packages/twenty-front/src/modules/ui/input/components/SelectControl.tsx
 (reads .Icon from selectedOption)

**Steps to Reproduce:**

Ensure workspace has no AI models configured (so aiModelsState is empty).
Open /settings/ai/agents/<some-id>.
Observe crash.

Expected:
Page renders; model select shows a placeholder when no models are available; no crash.

Actual:
Console shows:
TypeError: Cannot read properties of undefined (reading 'Icon')
Stack originates from 
SelectControl
 rendering with undefined selectedOption.

Root Cause:

In Select.tsx, selectedOption was computed with: options.find(...) || emptyOption || options[0]
If options is empty and emptyOption is not provided by caller, result is undefined. SelectControl.tsx
 uses selectedOption.Icon and selectedOption.label, causing the exception when selectedOption is undefined.
The AI Model select in SettingsAIAgentForm.tsx didn’t pass an emptyOption, and with empty aiModelsState the crash occurs.

Proposed Fix: (couldn't make a PR)

Add emptyOption to the AI Model select so a placeholder is always present when no models exist.

File: 
packages/twenty-front/src/pages/settings/ai/forms/components/SettingsAIAgentForm.tsx

Change: emptyOption={{ label: t\No models available`, value: '' }}`

Harden 
Select.tsx to provide a safe fallback when both options and emptyOption are absent:
Use fallbackOption = { label: '', value: '' as Value } and include it in selection pipeline: ... || fallbackOption.

Why this is safe:
The fallback avoids runtime exceptions in shared UI.
The form already shows a helper message: “No models available...” and disables the control.

**Environment:**
Frontend: twenty-front package
[...]

PR #15455, opened 2025-10-30
fix: ai settings page crash

anadi45:
Fixes - https://github.com/twentyhq/twenty/issues/14995


anadi45:
@charlesBochet please check this as well

packages/twenty-front/src/modules/ui/input/components/SelectControl.tsx
```diff
@@ -80,19 +80,19 @@ export const SelectControl = ({
...
-      {isDefined(selectedOption.Icon) ? (
+      {isDefined(selectedOption?.Icon) ? (
         <selectedOption.Icon
           color={isDisabled ? theme.font.color.light : theme.font.color.primary}
           size={theme.icon.size.md}
           stroke={theme.icon.stroke.sm}
         />
       ) : null}
-      <OverflowingTextWithTooltip text={selectedOption.label} />
+      <OverflowingTextWithTooltip text={selectedOption?.label ?? ''} />
```
prastoin: ***Remark:*** Either redundant or would need to update type for this to be optional, about to double check

prastoin (commented):
Hello there thanks for your contribution, left a comment

prastoin (commented):
Indeed the issue comes from the label being undefined, this fix is not enough

packages/twenty-front/src/modules/ui/input/components/Select.tsx
```diff
@@ -93,11 +93,29 @@ export const Select = <Value extends SelectValue>({
   const selectContainerRef = useRef<HTMLDivElement>(null);
 
   const [searchInputValue, setSearchInputValue] = useState('');
+  const getSelectedOption = () => {
+    const fromMatchingOption = options.find(
+      ({ value: optionValue }) => optionValue === value,
+    );
 
-  const selectedOption =
-    options.find(({ value: key }) => key === value) ||
```
prastoin, line 98: ***Remark:*** this did not handle `''` or `0` atomic values correctly before same for `null`

prastoin (commented):
Hey @lucasbordeau could you please have a look to this when you have some free time ?

packages/twenty-front/src/modules/ui/input/components/Select.tsx
```diff
@@ -93,11 +93,29 @@ export const Select = <Value extends SelectValue>({
...
-    emptyOption ||
-    options[0];
+    if (fromMatchingOption !== undefined) {
```
lucasbordeau: use `!isDefined()`

packages/twenty-front/src/modules/ui/input/components/Select.tsx
```diff
@@ -93,11 +93,29 @@ export const Select = <Value extends SelectValue>({
...
+      return fromMatchingOption;
+    }
+
+    if (emptyOption !== undefined) {
```
lucasbordeau: Same

packages/twenty-front/src/modules/ui/input/components/Select.tsx
prastoin: Would involve a regression,`fromMatchingOption` is nullable

packages/twenty-front/src/modules/ui/input/components/Select.tsx
prastoin: Same answer

packages/twenty-front/src/modules/ui/input/components/Select.tsx
prastoin:
```ts

export type SelectOption<
  Value extends string | number | boolean | null = string,
> = {
  Icon?: IconComponent | null;
  label: string;
  value: Value;
  disabled?: boolean;
  color?: ThemeColor | 'transparent';
};

```

lucasbordeau: Why ? We treat null and undefined the same in the application, we shouldn't make a difference, that's why we have isDefined

lucasbordeau: Creating different code paths for null and undefined is not a good idea as we have already discussed a few months back, the default way should be to use isDefined everywhere and in some very rare edge cases to differentiate between null and undefined, not the other way around.

prastoin: What if your empty option has a `null` value, tbh the current implementation does not handle correctly not being passed any options at all
In my opinion in such low level and generic implementation it's not a problem
Especially when we need the grain between both `undefined` and `null`

lucasbordeau: I think we're mixing concerns here, `null | undefined` is a `Nullable<T>` this should be handled with `isDefined` as we do in 99% of the application, there is no need here to differentiate undefined and null, as in 99% of use cases.

lucasbordeau: Whether it's just `null` or just `undefined` doesn't call for avoiding `isDefined`, on the contrary, using `isDefined` prevents introducing future bugs if someone changes the type to `undefined | null` (for example because a caller of Select wants to pass null and that it is easier to change the API of Select. 

This can happen a lot, handling just undefined or just null leaves the door open for bugs, it's an open-ended bug trap, while using isDefined deterministically closes the door to any bug that would involve a `null` value not being caught by `=== undefined` or the opposite, thus avoiding altogether the need to maintain null or undefined specific code paths, which in 99.999999% of the cases brings no benefit being treated separately.

lucasbordeau: I modified the file and applied the correct typeguards that should be used everywhere so you can see what we expect in the app.

lucasbordeau: Also I hope that we agree on that very soon because that's how we handle nullable values since 2 years, so better not introducing and favoring another way to work with nullable now.

lucasbordeau: To go in your direction, we could separate here indeed the option.value treatment between null and undefined, but this is not required here and the code that was checking for undefined wasn't checking option.value but option object, which shouldn't be treated as undefined only but nullable with `isDefined`

lucasbordeau: For example in `filteredOptions` we could add a `option.value !== null` but using a `isDefined(option.value)` would also work great, I don't see a precise use case that would require handling null value separately here, using isDefined would work also. 

That's the thing with undefined and null, from experience they should not be separated because very often they can indeed be treated the same and they should be from a maintenance and cognitive load perspective.

lucasbordeau: And you're right for empty option array, but that's again another use case than null | undefined. 

See this issue : https://github.com/twentyhq/core-team-issues/issues/1826

lucasbordeau (approved):
LGTM

PR #15456, opened 2025-10-30
Fix Advanced-filter Opportunity Page Crash

### LINKED ISSUE #15440: in Opportunities, filter -> advanced filter crashes the page
## Bug Description

When you are in the opportunities view. And you select filter. And subsequently advanced filter. The moment you click on name inside the advanced filter the page crashes. 

## Expected behavior

to be able to set variables for an advanced filter.

PriyanshuPandey92:
Fixes #15440

Issue:
The FULL_NAME field is a composite type with filterable subfields (firstName, lastName), but wasn't configured to require subfield selection like ACTOR and CURRENCY. This caused the UI to crash when clicking on the name field in advanced filters. (Opportunity Page)

Fix:
Added FULL_NAME to the list of composite types that require explicit subfield selection and set the default subfield to 'firstName' to ensure the filter always has a valid state.


lucasbordeau:
Hi @PriyanshuPandey92 the bug was due to a problem with the fieldMetadataItem not being defined in the field list selection dropdown, which has already been fixed in a previous PR and will very soon be deployed into production.

And to give you more explanation, here we want to actually keep FULL_NAME filterable as any because this way, adding a filter on a person's name field will create by default a field that allows to filter on first and last name, which is the expect product behavior. If the user wants to filter only on first name or only on last name, or apply a different filter for each one, the user can use advanced filters.

I'm closing this PR, thank you for contributing.

PR #15471, opened 2025-10-30
Support workflows in record page layouts

Devessier:
https://github.com/user-attachments/assets/275a02a3-7054-45d1-9423-75bb5223a8ba


packages/twenty-front/src/modules/page-layout/types/PageLayoutTab.ts
```diff
@@ -12,5 +12,5 @@ export type PageLayoutTab = Omit<PageLayoutTabGenerated, 'widgets'> & {
   /**
    * Only available behind IS_RECORD_PAGE_LAYOUT_ENABLED for now.
    */
-  layoutMode?: 'grid' | 'vertical-list';
+  layoutMode?: 'grid' | 'vertical-list' | 'canvas';
```
Devessier: `canvas = 1 fullscreen widget`

packages/twenty-front/src/modules/page-layout/constants/DefaultWorkflowPageLayout.ts
```diff
@@ -0,0 +1,53 @@
+import { DEFAULT_WORKFLOW_PAGE_LAYOUT_ID } from '@/page-layout/constants/DefaultWorkflowPageLayoutId';
+import { type PageLayout } from '@/page-layout/types/PageLayout';
+import { PageLayoutType, WidgetType } from '~/generated/graphql';
+
+/**
+ * Default Workflow PageLayout.
```
FelixMalfait, line 6: Docblock-style comments aren't really used in the codebase. I don't mind them personally but maybe we should remove them for consistency

FelixMalfait (approved):
As discussed in DM let's maybe avoid introducing  a new concept for widget card context, or if we do then name it more explicitly to clearly differentiate what it is and why it's needed

PR #15478, opened 2025-10-30
switch migrationActionsWithParent to migrationActionsWithParentTmp temporary

guillim:
The removal of views from the workspaces schema implies the deletion of 
```
view
viewFilter
viewFilterGroup
viewGroup
viewSort
```
Before it is created again in the core schema.

However, the syncmetadata that executes the pending migration fails because it will try to 
`query failed: DROP TABLE "workspace_xxxx"."view"`
before the removal of the other view related tables that contains a view foreign key with this kind of message
> constraint FK_c5ab40cd4debb51d588752a4857 on table "viewField" depends on table view

### DIFF
--- a/packages/twenty-server/src/engine/workspace-manager/workspace-migration-runner/workspace-migration-runner.service.ts
+++ b/packages/twenty-server/src/engine/workspace-manager/workspace-migration-runner/workspace-migration-runner.service.ts
@@ -67,10 +67,20 @@ export class WorkspaceMigrationRunnerService {
       `SET LOCAL search_path TO ${schemaName}`,
     );
 
+    // temporary fix to skip view migrations issue during upgrade 1.8 -> 1.10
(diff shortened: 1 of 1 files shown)


Weiko (approved):

Weiko (changes requested):
Actually approved too quickly, I don't see how it fixes the issue, workspace-migration-runner is for workspace entities and view is a core entity, right? 🤔

guillim:
> workspace

yes. but since we remove the view related tables from the workspaces entities, we need it

Weiko:
> > workspace
> 
> yes. but since we remove the view related tables from the workspaces entities, we need it

Ok, read too quickly, I thought you were trying to solve another problem. (The FK on core.view we discussed the other day).

guillim:
No worries. let me know if there are anything else you notice

Weiko (approved):

PR #15479, opened 2025-10-30
Some fixes for colors

Devessier:
Please note the background colors of the tags and of the nodes

## Before

CleanShot 2025-10-30 at 16 03 21@2x

## After

CleanShot 2025-10-30 at 17 03 25@2x


bosiraphael (approved):
LGTM

PR #15480, opened 2025-10-30
Fix orderBy, groupBy, and orderByForRecords on foreignKey

ijreilly:
For the variables orderBy (for findMany and groupBy), groupBy (for groupBy) and orderByForRecords (for groupBy), we were wrongfully adding the foreign key field (eg: pointOfContact) by its joinColumnName (eg: pointOfContactId) as the variable key (eg. `orderBy: { pointOfContactId: "AscNullsFirst" } }`. That broke because then this key is used to identify the field in the parsers. 

This went unnoticed because this order / group option is not very interesting as it is limited to the id for now, but it s still better to have it work than crash!

before (on findMany)
flawn_order_by_pocId

after (on findMany)
image


packages/twenty-server/src/engine/twenty-orm/utils/format-column-name-for-relation-field.util.ts
```diff
@@ -0,0 +1,19 @@
+import { RelationType } from 'twenty-shared/types';
+import { isDefined } from 'twenty-shared/utils';
+
+import { type FieldMetadataRelationSettings } from 'src/engine/metadata-modules/field-metadata/interfaces/field-metadata-settings.interface';
+
+export const formatColumnNameForRelationField = (
```
etiennejouan: What do you think of throwing for oneToMany relation here ? because there is no associated column name

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-order/graphql-query-order.parser.ts
```diff
@@ -54,14 +56,15 @@ export class GraphqlQueryOrderFieldParser {
   ): Record<string, OrderByCondition> {
     return orderBy.reduce(
       (acc, item) => {
-        Object.entries(item).forEach(([key, value]) => {
-          const fieldMetadataId = this.objectMetadataMapItem.fieldIdByName[key];
+        Object.entries(item).forEach(([fieldName, value]) => {
+          const fieldMetadataId =
+            this.objectMetadataMapItem.fieldIdByName[fieldName];
```
etiennejouan: After reviewing, I think it would have been easier to fix initial issue by fetching fieldMetadata here doing : 

```
const fieldMetadataId = this.objectMetadataMapItem.fieldIdByName[fieldName] || this.objectMetadataMapItem.fieldIdByJoinColumnName[fieldName];
```

etiennejouan (approved):
LGTM. Left comments

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-order/graphql-query-order.parser.ts
ijreilly: much better indeed, did not know about fieldIdByJoinColumnName !

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-order/graphql-query-order.parser.ts
```diff
@@ -79,9 +83,18 @@ export class GraphqlQueryOrderFieldParser {
...
+            const columnName = isFieldMetadataRelationOrMorphRelation(
+              fieldMetadata,
+            )
+              ? formatColumnNameForRelationField(
+                  fieldMetadata.name,
+                  fieldMetadata.settings,
+                )
+              : fieldName;
+
+            acc[`"${objectNameSingular}"."${columnName}"${orderByCasting}`] =
```
etiennejouan, line 95: I thought columnName and fieldName where the same here, then no need to call formatColumnNameForRelationField @ijreilly

PR #15484, opened 2025-10-30
fix: updated dryrun to return correct populated record

harshit078:
## Description

- This PR approaches to solve https://github.com/twentyhq/twenty/issues/15201
- updated `createDryRunResponse` to return  fully populated merged record
- this way the frontend can render the populated data it as-is without recomputing relations
- Dry-run now uses the same nested-relations population path as the non-dryRun flow

PR #15486, opened 2025-10-31
fix: removeuseMergeRecordRelatationship and simplify dry run response

harshit078:
## Description

- This PR addresses issue https://github.com/twentyhq/twenty/issues/15201
- Removed [useMergeRecordRelationships.ts](https://github.com/twentyhq/twenty/compare/main...harshit078:fix-merge-frontend?expand=1#diff-5d0366f814ff0c0e8961f40120690206a7acd3f7b66135cfb8c4ad8d3f6bc681)
- now backend returns populated result and frontend just shows it without refetching relations.
- to be merged after PR- https://github.com/twentyhq/twenty/pull/15484

## Visual Appearance

 

https://github.com/user-attachments/assets/09d7548a-74a9-4742-af49-e98b4174f68c


etiennejouan (changes requested):
Preview on oneToMany relations does not work

here opportunities are not displayed but available in mergeMany payload response
Image

harshit078:
I realized the issue was in `useMergePreview.ts` and in `common-merge-many-query-runner.service.ts`

- Now relations are populated on `recordsToMerge` before `performDeepMerge` runs as a result so `performDeepMerge` can merge them. 
- Frontend shows complete record coming from backend and makes no extra redudant call

https://github.com/user-attachments/assets/bfae24b0-4113-4e32-b3de-553eaba640a6

etiennejouan (approved):
Hi @harshit078 its works well. Well done and thank you for your contribution !

etiennejouan:
Even though this PR fixes the initial issue Stephanie reported, the merge feature could benefit from some other improvements (not a priority):

- The preview merge record is a fake one—we should not insert it in the record store.
My approach would be to use a specific preview record state, fetched by the preview record component (this would require modifications to how the FieldCard component fetches records—maybe too costly?).
@harshit078 has a different and interesting approach:

> We create a context that provides the preview record so that fake records never reach the store—they stay in the context. As a result, recordStoreFamilySelector will check the context first, and if a preview record exists in context with the requested ID, we return data from the context we created. I'm not sure if this would be best, but I think this could keep the fake records separate and prevent them from being inserted in the record store.

- On the preview tab, you can edit relations—you should not be able to.
- On the preview tab, the relation total is not displayed.


@charlesBochet have you any hint to share ?

charlesBochet:
1) I think that most components / hooks will expect the recordStore to contain all related records to the merge preview record and it looks a bit too heavy to change this logic to make them cope with both recordStoreState and mergedRecordState
2) It leaves us with two alternatives: a) modify recordStoreFamilySelector as suggested by @harshit078 but I'm not 100% sure that you can read a context from an atom. We could also introduce another state: virtualRecordStoreState and make the selector read from both. I still think it's a bit overkill. b) accepts that we can have fake records in the recordStore (we could even attach a type to each record in the record store but I'm not sure it's worth here). Why is it an issue?

etiennejouan:
I was worried that the fake record might interfere with other features. 

This fake record is related (relation) to other real records. If there's ever a fetch performed in the store based on these relationships, it would come up. Definitely unlikely?

We could also say that we clean up this record when exiting the preview.

Issue #15368, opened 2025-10-27
Import Function - Clarify the field (domain URL) expected to create relationships


StephanieJoly4:
## FR Description
This is what is shown to users when trying to make a relation from People and Companies (uploading a csv file for People), via the import function.
Can we explicit the field needed is the domainName URL and not Label, as some users got confused? thanks!

<img width="511" height="92" alt="Image" src="https://github.com/user-attachments/assets/48a6080b-4ef0-4d50-be6c-97de5fbae84b" />

cc @etiennejouan

Ayush975637:
Hi! 👋 I’d love to work on this issue if it’s still available.  
I’ve gone through the codebase and understood the affected module.  
Please assign it to me so I can start working and open a PR soon. Thanks!

etiennejouan:
Hi @Ayush975637 Thank you for your motivation. Please go ahead.

We would like to display "Company / Domain Name / Primary Link Url" instead of what we see on the screenshot. In the case of the company domain name, "primary link url" is the subfield which hold the unicity. Note that it only affects the composite field.

Do not hesitate to contact me on Discord if you have any question.

PR #15489, opened 2025-10-31
fix: composite field label retrieval in spreadsheet import utility

anadi45:
Fixes - https://github.com/twentyhq/twenty/issues/15368

Seems like the sub field was not being used for the actual label lookup part 

image


anadi45:
@etiennejouan looked like the label lookup for subfield was incorrect. Please review and let me know if I hunch is correct or not

etiennejouan (approved):
Hi @anadi45 it works well ! Thank you for your contribution 🙏

PR #15491, opened 2025-10-31
Common - Field validation

### LINKED ISSUE #1622: Common - Add field value validation
https://github.com/twentyhq/twenty/pull/15194

https://github.com/twentyhq/twenty/issues/15298

etiennejouan:
Closes : https://github.com/twentyhq/core-team-issues/issues/1622

To do in other PR : 
- Add migration command for non nullable text, raw_json & array fields
- Add null transformation


packages/twenty-server/src/engine/api/common/common-args-handlers/data-arg-handler/coercer-utils/coerce-actor-field-or-throw.util.ts
```diff
@@ -0,0 +1,60 @@
...
+  value: unknown,
+  fieldName?: string,
+  isNullEquivalenceEnabled: boolean = false,
+) => {
+  if (isNull(value)) return null;
+
+  try {
+    const parsedValue = JSON.parse(JSON.stringify(value));
+
+    if (!isObject(parsedValue)) throw Error;
```
Weiko: Can we throw a proper error?

Weiko: Ah I see, you swallow and rethrow in the same util. I guess it's fine then, error.message will be empty though 🤔

Weiko: although I know Error is a constructor function and is equivalent but we prefer to do "throw **new** Error..." to be more consistent 🙏

packages/twenty-server/src/engine/api/common/common-args-handlers/data-arg-handler/coercer-utils/coerce-address-field-or-throw.util.ts
```diff
@@ -0,0 +1,63 @@
...
+import { coerceNumberFieldOrThrow } from 'src/engine/api/common/common-args-handlers/data-arg-handler/coercer-utils/coerce-number-field-or-throw.util';
+import { coerceTextFieldOrThrow } from 'src/engine/api/common/common-args-handlers/data-arg-handler/coercer-utils/coerce-text-field-or-throw.util';
+import {
+  CommonDataCoercerException,
+  CommonDataCoercerExceptionCode,
+} from 'src/engine/api/common/common-args-handlers/data-arg-handler/errors/common-data-coercer.exception';
+
+export const coerceAddressFieldOrThrow = (
+  value: unknown,
+  fieldName?: string,
```
Weiko: why is fieldName optional? Won't you have issue when you'll concatenate
```typescript
         coerceTextFieldOrThrow(
            parsedValue[subField],
            `${fieldName}.${subField}`,
          );
```
below?

packages/twenty-server/src/engine/api/common/common-args-handlers/data-arg-handler/coercer-utils/coerce-actor-field-or-throw.util.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export const coerceActorFieldOrThrow = (
+  value: unknown,
+  fieldName?: string,
+  isNullEquivalenceEnabled: boolean = false,
+) => {
+  if (isNull(value)) return null;
+
+  try {
+    const parsedValue = JSON.parse(JSON.stringify(value));
```
Weiko: Is it to make a copy? it seems you are not mutating the object so I'm not sure why? If you need a copy, prefer structuredClone as it is much faster and robuste (with json.stringify/parse you'll lose types such as Date etc)

packages/twenty-server/src/engine/api/common/common-args-handlers/data-arg-handler/coercer-utils/coerce-array-field-or-throw.util.ts
```diff
@@ -0,0 +1,43 @@
...
+      throw new CommonDataCoercerException(
+        `Invalid value ${inspect(value)} for field "${fieldName}"`,
+        CommonDataCoercerExceptionCode.INVALID_ARRAY,
+      );
+    }
+
+    if (parsedValue.length === 0 && isNullEquivalenceEnabled) return null;
+
+    return value as unknown[];
+  } catch {
```
Weiko: To be consistent with your other utils, you missed error binding here

packages/twenty-server/src/engine/api/common/common-args-handlers/data-arg-handler/coercer-utils/coerce-array-field-or-throw.util.ts
```diff
@@ -0,0 +1,43 @@
...
+    if (typeof parsedValue === 'string') {
+      return [parsedValue];
+    }
+
+    if (
+      !Array.isArray(parsedValue) ||
+      parsedValue.some((item) => typeof item !== 'string')
+    ) {
+      throw new CommonDataCoercerException(
+        `Invalid value ${inspect(value)} for field "${fieldName}"`,
```
Weiko: Nit: maybe specify "array values need to be string" or something like that 🙂

packages/twenty-server/src/engine/api/common/common-args-handlers/data-arg-handler/coercer-utils/coerce-rich-text-v2-field-or-throw.util.ts
```diff
@@ -0,0 +1,44 @@
...
+      .get(FieldMetadataType.RICH_TEXT_V2)
+      ?.properties.filter(
+        (prop) => prop.hidden !== true && prop.hidden !== 'input',
+      )
+      .map((prop) => prop.name);
+
+    if (!subfields.every((subfield) => richTextV2Subfields?.includes(subfield)))
+      throw Error;
+
+    return await transformRichTextV2Value(value);
```
Weiko: why is this util async?

packages/twenty-server/src/engine/api/common/common-args-handlers/data-arg-handler/coercer-utils/coerce-actor-field-or-throw.util.ts
```diff
@@ -0,0 +1,60 @@
...
+        case 'source':
+          coerceRatingAndSelectFieldOrThrow(
+            parsedValue[subField],
+            Object.keys(FieldActorSource),
+            `${fieldName}.${subField}`,
+            isNullEquivalenceEnabled,
+          );
+          break;
+        case 'context':
+          coerceRawJsonFieldOrThrow(
```
Weiko: note: if this throws, you'll probably have a message like `"Invalid value ... for actor field "field" - Invalid value ... - Invalid value ..."` (with coerceRawJsonFieldOrThrow having already nested messages potentially) 🙂

packages/twenty-server/src/engine/api/common/common-args-handlers/data-arg-handler/coercer-utils/coerce-text-field-or-throw.util.ts
```diff
@@ -0,0 +1,24 @@
+import { inspect } from 'util';
+
+import { isNull } from '@sniptt/guards';
+
+import {
+  CommonDataCoercerException,
+  CommonDataCoercerExceptionCode,
+} from 'src/engine/api/common/common-args-handlers/data-arg-handler/errors/common-data-coercer.exception';
+
+export const coerceTextFieldOrThrow = (
```
Weiko: I feel like those utils are mostly "validators", they don't really transform the response right? (for example sometimes you don't even use the returned value in nested calls) Is it still really Coercion in this case?
This is not a big blocker for me, just wanted to share this

Weiko: Well, in this case it does transform the response by returning null indeed 👍

packages/twenty-server/src/engine/api/common/common-args-handlers/data-arg-handler/coercer-utils/coerce-actor-field-or-throw.util.ts
```diff
@@ -0,0 +1,60 @@
...
+  CommonDataCoercerException,
+  CommonDataCoercerExceptionCode,
+} from 'src/engine/api/common/common-args-handlers/data-arg-handler/errors/common-data-coercer.exception';
+import { FieldActorSource } from 'src/engine/metadata-modules/field-metadata/composite-types/actor.composite-type';
+
+export const coerceActorFieldOrThrow = (
+  value: unknown,
+  fieldName?: string,
+  isNullEquivalenceEnabled: boolean = false,
+) => {
```
Weiko: nit: missing return type

packages/twenty-server/src/engine/api/common/common-args-handlers/data-arg-handler/coercer-utils/coerce-numeric-field-or-throw.util.ts
```diff
@@ -0,0 +1,18 @@
...
+
+//Need to handle stringified numbers because of BigFloatScalarType custom gql type
+
+export const coerceNumericFieldOrThrow = (
+  value: unknown,
+  fieldName?: string,
+): number | null => {
+  if (value === '' || isNull(value)) return null;
+
+  const numberValue = Number(String(value));
```
Weiko: nit: Why convert to String first? I can see the case where Number(null) returns 0 and we don't want that but you check nullable above

packages/twenty-server/src/engine/api/common/common-args-handlers/data-arg-handler/coercer-utils/coerce-multi-select-field-or-throw.util.ts
```diff
@@ -0,0 +1,31 @@
...
+) => {
+  const parsedValue = coerceArrayFieldOrThrow(
+    value,
+    fieldName,
+    isNullEquivalenceEnabled,
+  );
+
+  if (parsedValue === null) return null;
+
+  if (!parsedValue.every((item) => options?.includes(item as string))) {
```
Weiko: Seems options can be undefined. Should we throw a different error first if that's the case?

packages/twenty-server/src/engine/core-modules/record-transformer/services/record-input-transformer.service.ts
```diff
@@ -72,85 +64,18 @@ export class RecordInputTransformerService {
           'Rich text is not supported, please use RICH_TEXT_V2 instead',
         );
       case FieldMetadataType.RICH_TEXT_V2:
-        return this.transformRichTextV2Value(value);
+        return await transformRichTextV2Value(value);
       case FieldMetadataType.LINKS:
         return transformLinksValue(value);
       case FieldMetadataType.EMAILS:
-        return this.transformEmailsValue(value);
+        return transformEmailsValue({ input: value });
```
Weiko: is it expected to pass an object with input? It seems it's not used that way in the util and you have another usage where we pass the value directly? Might be ok but it's hard to follow with all those any/unknown 😅

Weiko (commented):
Left some comments! Mostly good to me but I feel like most of the coercion done here is actually validation so name is a bit misleading imho. Also there is a commit "update test + add feature flag" but I don't see any 🤔
Well done!

packages/twenty-server/src/engine/api/common/common-args-handlers/data-arg-handler/coercer-utils/coerce-address-field-or-throw.util.ts
etiennejouan: It's a mistake. At first I had defined scalar coerce functions also used by composite fields logic. When changing strat, I forget to remove these undefined

packages/twenty-server/src/engine/api/common/common-args-handlers/data-arg-handler/coercer-utils/coerce-actor-field-or-throw.util.ts
etiennejouan: I've seen this check in some GQL custom field validation but it seems way more complex than a simple type of  === object.
The only advantage is that it can throw error with circular object... not that useful

packages/twenty-server/src/engine/api/common/common-args-handlers/data-arg-handler/coercer-utils/coerce-rich-text-v2-field-or-throw.util.ts
etiennejouan: It uses blocksToMarkdownLossy which is async

packages/twenty-server/src/engine/api/common/common-args-handlers/data-arg-handler/coercer-utils/coerce-text-field-or-throw.util.ts
etiennejouan: As discussed, we now have validation and transformation in a same processor service

packages/twenty-server/src/engine/api/common/common-args-handlers/data-arg-handler/coercer-utils/coerce-numeric-field-or-throw.util.ts
etiennejouan: Not sure to remember. Removed.

Important thing is that numeric type field accepts stringified number, number type field not.

packages/twenty-server/src/engine/core-modules/record-transformer/services/record-input-transformer.service.ts
etiennejouan: good catch, it should not. I have not take time to improve typing on these composite transformation util, ideally it will be done, it's noted

etiennejouan:
As discussed with @Weiko, data arg `processor` service handles field `validation` and `transformation` (if needed).

In some transformation utils, you can see beginning ( isNullEquivalenceEnabled optional input with default value) of null equivalence feature, coming in next PR.

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/transformer-utils/transform-address-field.util.ts
```diff
@@ -0,0 +1,56 @@
+import { isNull, isUndefined } from '@sniptt/guards';
+
+import { transformNumericField } from 'src/engine/api/common/common-args-processors/data-arg-processor/transformer-utils/transform-numeric-field.util';
+import { transformTextField } from 'src/engine/api/common/common-args-processors/data-arg-processor/transformer-utils/transform-text-field.util';
+
+export const transformAddressField = (
+  value: {
+    addressStreet1?: string | null;
```
Weiko, line 8: I'm surprised we don't have TS types for composite types (I've checked!), probably something to improve for later!

Weiko (approved):
LGTM!

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/transformer-utils/transform-address-field.util.ts
martmull: we have them, but not with optional values (see packages/twenty-shared/src/types/composite-types/address.composite-type.ts)

image

PR #15493, opened 2025-10-31
fix: Orderby secondary dimension sum

ehconitin:
before:

https://github.com/user-attachments/assets/6919c45b-1435-408f-87e3-0452dd70fbdb

after:

https://github.com/user-attachments/assets/efa60d28-ac04-4219-831a-92aada13fa9c


bosiraphael (approved):
LGTM

PR #15496, opened 2025-10-31
Instant widget placeholder placement on drag end

ehconitin:
closes - https://discord.com/channels/1130383047699738754/1430596561137700957

fixed a bug - where on chart type change in side panel -- wont update the widgets minimum size


packages/twenty-front/src/modules/page-layout/components/PageLayoutGridLayout.tsx
```diff
@@ -86,6 +92,16 @@ export const PageLayoutGridLayout = ({ tabId }: PageLayoutGridLayoutProps) => {
 
   const { handleLayoutChange } = usePageLayoutHandleLayoutChange();
 
+  const handleLayoutChangeWithoutPendingPlaceholder = useCallback(
```
bosiraphael: No need for useCallback here

packages/twenty-front/src/modules/command-menu/pages/page-layout/components/ChartSettings.tsx
```diff
@@ -75,51 +98,89 @@ export const ChartSettings = ({ widget }: { widget: PageLayoutWidget }) => {
   const [hasWidgetTooManyGroups, setHasWidgetTooManyGroups] =
     useRecoilComponentState(hasWidgetTooManyGroupsComponentState);
 
-  const handleGraphTypeChange = (graphType: GraphType) => {
-    const configToUpdate: Record<string, any> = {
-      __typename: GRAPH_TYPE_TO_CONFIG_TYPENAME[graphType],
-      graphType,
-    };
+  const handleGraphTypeChange = useRecoilCallback(
```
bosiraphael: Extract `handleGraphTypeChange` into a hook file. The component is becoming way too big.

Good reflex to convert it to a recoil callback, but you shouldn't use

```
 const activeTabId = useRecoilComponentValue(
    activeTabIdComponentState,
    tabListInstanceId,
  );
  const currentlyEditingWidgetId = useRecoilComponentValue(
    pageLayoutEditingWidgetIdComponentState,
    pageLayoutId,
  );
```

This creates a dependency on these states in the component.
Rather extract the states with `useRecoilComponentCallbackState` and read the values from the snapshot

packages/twenty-front/src/modules/page-layout/components/PageLayoutGridLayout.tsx
```diff
@@ -109,21 +129,40 @@ export const PageLayoutGridLayout = ({ tabId }: PageLayoutGridLayoutProps) => {
...
-      <div key={widget.id} data-select-disable="true">
-        <WidgetRenderer widget={widget} />
-      </div>
-    ))
-  );
+  const layouts = hasPendingPlaceholder
+    ? addPendingPlaceholderToLayouts(baseLayouts, pageLayoutDraggedArea)
+    : baseLayouts;
+
+  const widgetsArray = isLayoutEmpty
```
bosiraphael: This isn't very readable, improve style

packages/twenty-front/src/modules/command-menu/pages/page-layout/components/ChartSettings.tsx
```diff
@@ -75,51 +98,89 @@ export const ChartSettings = ({ widget }: { widget: PageLayoutWidget }) => {
...
+              currentlyEditingWidgetId,
+              { minW: minimumSize.w, minH: minimumSize.h },
+            );
+
+            set(pageLayoutCurrentLayoutsState, {
+              ...currentLayouts,
+              [activeTabId]: updatedTabLayouts,
+            });
+          }
+        }
```
bosiraphael: I'm not sure I understand why we need that

ehconitin: refactored into `updateWidgetMinimumSizeForGraphType` utility.

this ensures widgets meet the minimum size requirements for the new graph type - prevents charts from being lost for minimun size when switching types.

packages/twenty-front/src/modules/command-menu/pages/page-layout/components/ChartSettings.tsx
```diff
@@ -69,41 +62,20 @@ export const ChartSettings = ({ widget }: { widget: PageLayoutWidget }) => {
...
-          isDefined(aggregateField) &&
-          isFieldMetadataDateKind(aggregateField.type) &&
-          (configuration.aggregateOperation === AggregateOperations.MIN ||
-            configuration.aggregateOperation === AggregateOperations.MAX)
-        ) {
-          configToUpdate.aggregateOperation = AggregateOperations.COUNT;
-        }
-      }
-    }
+    const configToUpdate = updateGraphTypeConfig(graphType);
```
lucasbordeau, line 80: +1

packages/twenty-front/src/modules/page-layout/components/PageLayoutGridLayout.tsx
```diff
@@ -97,10 +117,50 @@ export const PageLayoutGridLayout = ({ tabId }: PageLayoutGridLayoutProps) => {
...
     : (pageLayoutCurrentLayouts[tabId] ?? EMPTY_LAYOUT);
 
+  const layouts = hasPendingPlaceholder
+    ? addPendingPlaceholderToLayouts(baseLayouts, pageLayoutDraggedArea)
+    : baseLayouts;
+
+  const widgets = [];
+
+  if (isLayoutEmpty) {
+    widgets.push(
```
lucasbordeau: No that's a pattern to avoid, doing mutation at the root of a react component, in the react render loop is a bad idea, use a useMemo or an external hook but never this pattern.

lucasbordeau: Moreover, let's avoid putting components into variable, that's a pattern that can ALWAYS be replaced by a simple .map on an array that will have been carefully crafted. 

Another reason is that we already keep all JSX-related code into the JSX return statement and that it is a convention that helps the reader avoid surprise JSX code into variables somewhere else in the code. 

JSX goes in the JSX part, array manipulation goes into the array manipulation logic, .map connects both, every time.

lucasbordeau (commented):
Nice keep up the good work and improving on clean code

lucasbordeau (approved):
LGTM

bosiraphael:
Inconsistent behavior:

https://github.com/user-attachments/assets/aa649029-c696-416b-8dcb-508d5c326b07

PR #15499, opened 2025-10-31
Bar chart: Prevent left tick labels to overlap on label

bosiraphael:
Prevent left tick labels to overlap on label by truncating the tick label


bosiraphael:
@greptileai

ehconitin (approved):
lgtm! grep is worried about numbers, which will already be shortened!
I dont think its an issue!

PR #15501, opened 2025-10-31
1630 extensibility twenty cli ability to create edit and delete fields

### LINKED ISSUE #1630: [Extensibility] [twenty-cli] ability to create, edit and delete fields
As title

Done for object, lets continue for fieldMetadata

The idea is to converge with workspace entities

Image

martmull:
As title

- adds decorators in twenty-sdk
- update twenty-cli load-manifest to it gets @FieldMetadata infos + testing
- update twenty-server so it CRUD fields properly, using universalIdentifier
- Fix UI so we can update managed objects records
- move FieldMetadata items from twenty-server to twenty-shared


packages/twenty-sdk/src/application/base-object-metadata.ts
```diff
@@ -0,0 +1,23 @@
+export abstract class BaseObjectMetadata {
+  id: string;
+
+  name: string;
```
Weiko, line 4: I think we should only have id, createdAt, updatedAt, deletedAt here. I don't think we always want the rest 🤔

Weiko: FYI @charlesBochet

packages/twenty-server/src/engine/core-modules/application/application-sync.service.ts
```diff
@@ -155,6 +161,131 @@ export class ApplicationSyncService {
     return application;
   }
 
+  private async syncFields({
```
Weiko, line 164: Few things probably won't work until we implement a proper transpilation + single call to the builder. For example if you remove an object on your app I suspect this will both send object deletion + n field deletion but the object deletion will happen before and field deletion will fail later. There is a lot of logic in the builder due to order of things for that and we don't want to reimplement it there so it's fine for now

packages/twenty-shared/src/types/FieldMetadataDefaultValue.ts
```diff
@@ -0,0 +1,311 @@
...
+  IsString,
+  IsUUID,
+  Matches,
+  ValidateIf,
+  type ValidationArguments,
+  type ValidationOptions,
+  registerDecorator,
+} from 'class-validator';
+
+const IsQuotedString = (validationOptions?: ValidationOptions) => {
```
Weiko, line 22: should be in a dedicated file

packages/twenty-shared/src/types/FieldMetadataDefaultValue.ts
```diff
@@ -0,0 +1,311 @@
...
+
+export const fieldMetadataDefaultValueFunctionName = {
+  UUID: 'uuid',
+  NOW: 'now',
+} as const;
+
+export type FieldMetadataDefaultValueFunctionNames =
+  (typeof fieldMetadataDefaultValueFunctionName)[keyof typeof fieldMetadataDefaultValueFunctionName];
+
+export class FieldMetadataDefaultValueString {
```
Weiko, line 49: Why are we moving all this stuff, is it for the SDK?

martmull: seen together

packages/twenty-shared/src/types/FieldMetadataDefaultValue.ts
martmull: Even for non exported const?

Weiko (approved):
LGTM

PR #15509, opened 2025-11-01
FIX: export view not exporting all records

### LINKED ISSUE #15508: Export view doesn't export all records
**Scenario:**
1. Log in
2. Go to any view (e.g. default People view) and export it
3. Check how many records are exported
4. Import it

**Actual:** Not all records in view are exported, out of 1200 records in dev seed, only 560 are exported

https://github.com/user-attachments/assets/bec498f3-56d4-409a-9f26-800781397488

**Expected:** All records should be exported

shantanugupta2004:
Fixes #15508 
Addresses the issue where the "Export View" functionality was not exporting all records from a view. Users reported that only a subset of records were exported, even when the view contained more. This led to incomplete data exports.

The core of the problem was found within the `useLazyFetchAllRecords` hook. The `fetchMoreRecordsLazy` function, which is responsible for fetching subsequent pages of data during the export process, was inadvertently using a default page limit (`DEFAULT_SEARCH_REQUEST_LIMIT` of 60 records) instead of the intended `pageSize` (200 records). This discrepancy caused the export process to prematurely stop fetching records.

The PR modifies `packages/twenty-front/src/modules/object-record/hooks/useLazyFetchAllRecords.ts` to explicitly pass the `limit` parameter (which correctly reflects the `pageSize`) to the `fetchMoreRecordsLazy` function. This ensures that all subsequent data fetches during an export operation adhere to the configured page size, allowing for the complete retrieval of all records.

### DIFF
--- a/packages/twenty-front/src/modules/object-record/hooks/useLazyFetchAllRecords.ts
+++ b/packages/twenty-front/src/modules/object-record/hooks/useLazyFetchAllRecords.ts
@@ -87,7 +87,7 @@ export const useLazyFetchAllRecords = <T>({
         await sleep(delayMs);
       }
 
-      const rawResult = await fetchMoreRecordsLazy();
+      const rawResult = await fetchMoreRecordsLazy(limit);
 
       const fetchMoreResult = rawResult?.data;
 


charlesBochet (approved):
Thank you @shantanugupta2004 :)

PR #15511, opened 2025-11-01
Last email interaction extension

BOHEUS:
Challenge 4 from "Call for projects" list


packages/twenty-apps/last_email_interaction/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "description": "Updates Last interaction and Interaction status fields based on last received email",
+  "env": {
+    "TWENTY_API_KEY": {
+      "isSecret": true,
+      "value": "",
+      "description": "Required to send requests to Twenty"
+    },
+    "TWENTY_API_URL": {
+      "isSecret": false,
+      "value": "",
```
martmull, line 22: Please, add default value to "value"

packages/twenty-apps/last_email_interaction/serverlessFunctions/test/src/index.ts
```diff
@@ -0,0 +1,275 @@
...
+      await setTimeout(DELAY);
+    }
+    if (person_interaction_status === undefined) {
+      const response2 = await axios.request(
+        create_interaction_status(person_object.id),
+      );
+      if (response2.status === 201) {
+        console.log('Successfully created person interaction status field');
+      }
+      await setTimeout(DELAY);
```
martmull, line 176: Why do you need `await setTimeout(DELAY);` everywhere?

martmull (commented):
Hey, thank you for your submission!
Code looks nice, did not test it yet
Some comments if you want to take a look
Cheers

packages/twenty-apps/last_email_interaction/serverlessFunctions/test/src/index.ts
BOHEUS: Original idea was to just add a little bit of delay between requests to prevent accidentally DDoSing workspace but I think something else should be done if it's necessary

packages/twenty-apps/last_email_interaction/package.json
BOHEUS: By default you mean `"default":"{value}"` or `"value":"{value}"`? Also, is it necessary when in first lines of code I'm checking if TWENTY_API_URL is defined and if not, it defaults to cloud API URL

BOHEUS:
@martmull thanks for review, left questions to comments as I'm not entirely sure about them

martmull (approved):

PR #15526, opened 2025-11-02
Child folders followup

neo773:


packages/twenty-front/src/modules/settings/accounts/components/message-folders/utils/organizeFoldersHierarchy.util.ts
```diff
@@ -0,0 +1,57 @@
+import { type MessageFolder } from '@/accounts/types/MessageFolder';
+import { isDefined } from 'twenty-shared/utils';
+
+export type HierarchicalFolder = {
+  folder: MessageFolder;
+  children: HierarchicalFolder[];
+  hasChildren: boolean;
+};
+
+export const organizeFoldersHierarchy = (
```
charlesBochet: in FE (react convention), we don't name the file .util.ts, this is nest convention

charlesBochet: computeHierarchicalFolders

packages/twenty-front/src/modules/settings/accounts/components/message-folders/SettingsMessageFoldersTreeItem.tsx
```diff
@@ -0,0 +1,166 @@
...
+  text-overflow: ellipsis;
+  white-space: nowrap;
+`;
+
+const StyledCheckboxWrapper = styled.div`
+  align-items: center;
+  display: flex;
+`;
+
+export const SettingsMessageFoldersTreeItem = ({
```
charlesBochet, line 100: we need to uniformize the naming:
Sometime it's parent > children, sometime it's hiearchical and sometime it's tree

Let's keep:
- only the parentId on MessageFolder
- keep MessageFolderTree and MessageFolderTreeItem (and get rid of hiearchical)

packages/twenty-front/src/modules/settings/accounts/components/message-folders/utils/organizeFoldersHierarchy.util.ts
charlesBochet: computeMessageFolderTree actually

packages/twenty-server/src/modules/messaging/message-folder-manager/drivers/gmail/gmail-get-all-folders.service.ts
```diff
@@ -82,4 +98,24 @@ export class GmailGetAllFoldersService implements MessageFolderDriver {
       throw error;
     }
   }
+
+  private getParentFolderId(
+    labelName: string,
+    labelNameToIdMap: Map<string, string>,
+  ): string | null {
+    if (!labelName.includes('/')) {
```
charlesBochet: question: is parent encoded like this in Gmail or is it your convention?

charlesBochet: also feel this should be extracted to util and unit tested as this is Gmail business logic

packages/twenty-server/src/modules/messaging/message-folder-manager/drivers/microsoft/microsoft-get-all-folders.service.ts
```diff
@@ -105,4 +117,57 @@ export class MicrosoftGetAllFoldersService implements MessageFolderDriver {
...
   }
+
+  private getStandardFolderFromWellKnownName(
+    wellKnownName?: string,
+  ): StandardFolder | null {
+    if (!isDefined(wellKnownName)) {
+      return null;
+    }
+
+    switch (wellKnownName.toLowerCase()) {
```
charlesBochet, line 125: I'm confused, why do we still need this logic here? don't we completely get rid of StandardFolders as we let users define their own folders?

charlesBochet: I feel we should always import all folders into account and the use regex / known names to try to identify:
- messageFolder.isSynced default (we should sync them all except the one we identify as draft or junk)
- messageFolder.isSentFolder default (same)

If the user changes it, we should respect it.

so I would introduce two functions instead of getStandardFolderFromWellKnownName and remove the notion of StandardFolder
- shouldBeSyncedByDefault
- shouldBeSentFolderByDefault

Also thinking about that we should actually prevent the user to sync draft emails in gmail as this will lead to bugs (message are disappearing when the are sent)

charlesBochet (commented):
Left comments! Happy to discuss them lives if needed

charlesBochet (approved):

PR #15528, opened 2025-11-03
Improve workflow performances

thomtrp:
- on draft creation, do not fetch the full version. Avoid the fetch of steps and trigger
- on activation, we were performing 8 queries/mutations synchronously + 2 additional for automated triggers. I refacto the call it it gets reduced to 4 queries/mutations + 2 additional for automated triggers

### DIFF
--- a/packages/twenty-front/src/generated/graphql.ts
+++ b/packages/twenty-front/src/generated/graphql.ts
@@ -1854,6 +1854,7 @@ export type Mutation = {
 
 
 export type MutationActivateWorkflowVersionArgs = {
+  workflowId: Scalars['UUID'];
   workflowVersionId: Scalars['UUID'];
 };
 
--- a/packages/twenty-front/src/generated/graphql.ts
+++ b/packages/twenty-front/src/generated/graphql.ts
@@ -1854,7 +1854,6 @@ export type Mutation = {
 
 
 export type MutationActivateWorkflowVersionArgs = {
-  workflowId: Scalars['UUID'];
   workflowVersionId: Scalars['UUID'];
 };
 
--- a/packages/twenty-server/src/modules/workflow/workflow-trigger/workspace-services/workflow-trigger.workspace-service.ts
+++ b/packages/twenty-server/src/modules/workflow/workflow-trigger/workspace-services/workflow-trigger.workspace-service.ts
@@ -1,6 +1,7 @@
 import { Injectable } from '@nestjs/common';
 
 import { msg } from '@lingui/core/macro';
+import { Not } from 'typeorm';
 
 import { type ActorMetadata } from 'src/engine/metadata-modules/field-metadata/composite-types/actor.composite-type';
 import { ScopedWorkspaceContextFactory } from 'src/engine/twenty-orm/factories/scoped-workspace-context.factory';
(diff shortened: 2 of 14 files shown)


packages/twenty-front/src/generated/graphql.ts
```diff
@@ -1854,6 +1854,7 @@ export type Mutation = {
 
 
 export type MutationActivateWorkflowVersionArgs = {
+  workflowId: Scalars['UUID'];
```
charlesBochet: why is it necessary?

packages/twenty-server/src/modules/workflow/workflow-trigger/workspace-services/workflow-trigger.workspace-service.ts
```diff
@@ -78,50 +79,63 @@ export class WorkflowTriggerWorkspaceService {
...
     const workspaceId = this.getWorkspaceId();
     const workflowVersionRepository =
       await this.twentyORMGlobalManager.getRepositoryForWorkspace<WorkflowVersionWorkspaceEntity>(
-        this.getWorkspaceId(),
+        workspaceId,
         'workflowVersion',
         { shouldBypassPermissionChecks: true }, // settings permissions are checked at resolver-level
       );
 
-    const workflowVersionNullable = await workflowVersionRepository.findOne({
```
charlesBochet, line 90: this does not look heavy? maybe because we are returning the steps?

packages/twenty-server/src/modules/workflow/workflow-trigger/workspace-services/workflow-trigger.workspace-service.ts
```diff
@@ -78,50 +79,63 @@ export class WorkflowTriggerWorkspaceService {
...
     const workflowRepository =
       await this.twentyORMGlobalManager.getRepositoryForWorkspace<WorkflowWorkspaceEntity>(
         workspaceId,
         'workflow',
         { shouldBypassPermissionChecks: true }, // settings permissions are checked at resolver-level
       );
 
-    const workflow = await workflowRepository.findOne({
-      where: { id: workflowVersion.workflowId },
+    const allWorkflowVersions = await workflowVersionRepository.find({
```
charlesBochet: we should do a select here to avoid loading heavy fields

packages/twenty-front/src/generated/graphql.ts
thomtrp: previous flow :
- load the workflowVersion and retrieve workflowId in there
- fetching potential published version from workflowId

new flow :
- fetch workflowVersion to activate + potential published version from workflowId in props

=> saving a query

packages/twenty-server/src/modules/workflow/workflow-trigger/workspace-services/workflow-trigger.workspace-service.ts
thomtrp: still need to load trigger in case I have to disable it. But I can avoid loading steps

packages/twenty-server/src/modules/workflow/workflow-trigger/workspace-services/workflow-trigger.workspace-service.ts
thomtrp: yes but we need to ensure the version can be activated

packages/twenty-server/src/modules/workflow/workflow-trigger/workspace-services/workflow-trigger.workspace-service.ts
```diff
@@ -78,50 +79,80 @@ export class WorkflowTriggerWorkspaceService {
...
+            type: true,
+          },
+          trigger: {
+            type: true,
+          },
+        },
+      });
+
+    const workflowVersionToActivate =
+      await this.workflowCommonWorkspaceService.getValidWorkflowVersionOrFail(
```
charlesBochet: use assertVersionIsValid

charlesBochet (approved):

charlesBochet (approved):

PR #15538, opened 2025-11-03
add documentation i18n workflows for Crowdin

abdulrahmancodes:


.github/workflows/docs-i18n-pull.yaml
```diff
@@ -0,0 +1,106 @@
...
+          base_url: 'https://twenty.api.crowdin.com'
+          skip_untranslated_files: true
+          push_translations: false
+          create_pull_request: false
+          skip_ref_checkout: true
+          dryrun_action: false
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+          CROWDIN_PROJECT_ID: '1'
+          CROWDIN_PERSONAL_TOKEN: ${{ secrets.CROWDIN_PERSONAL_TOKEN }}
```
charlesBochet, line 76: this CI should not be part of twenty repository (public) as this is using production secrets

charlesBochet (changes requested):
this CI should not be part of twenty repository (public) as this is using production secrets

FelixMalfait:
@charlesBochet let's tackle that in another PR since it was already the case for Crowdin

charlesBochet:
> @charlesBochet let's tackle that in another PR since it was already the case for Crowdin

Ok, I thought crowdin was already in the other side, LGTM not worse :p

PR #15452, opened 2025-10-29
fix: tmp allows arbitrary temporary file / directory write via symbolic link `dir` parameter

mabdullahabaid:
Resolves [Dependabot Alert 255](https://github.com/twentyhq/twenty/security/dependabot/255) - fix: tmp allows arbitrary temporary file / directory write via symbolic link `dir` parameter.

Updated the dev-dependency `zapier-platform-cli` for it to depend on tmp 0.2.4 and also ran `yarn up tmp --recursive` to update the version of tmp elsewhere.

Not expecting any breaking changes to twenty-zapier since `zapier-platform-cli` is marked as a development dependency.

PR #15539, opened 2025-11-03
Revert "fix: tmp allows arbitrary temporary file / directory write via symbolic link `dir` parameter"

charlesBochet:
Reverts twentyhq/twenty#15452


charlesBochet:
@mabdullahabaid @martmull this is breaking main, I'm reverting it.

@mabdullahabaid could you:
- investigate why it doesn't work (I'm a bit surprised by the size of changes in yarn.lock) and fix it
- add yarn.lock to the file watched by CI server and CI front (CI should have failed here but was skipped

PR #15543, opened 2025-11-03
Add date granularity and timezone and first day of the week to graphs

bosiraphael:
- Allow users to choose the date granularity of the x axis and the group by of the y axis on a bar chart
- Display those options conditionally
- Store timezones in graphs: each graph has its own timezone, defaults to the user timezone. There will be a picker in the v2 to choose the timezone. For now the timezone is not used by the backend, but it will be used in filters and in group by queries.
- Store first day of the week

https://github.com/user-attachments/assets/66a5d156-dd93-4ebe-8c8f-d172f93e25be


packages/twenty-front/src/modules/command-menu/pages/page-layout/utils/shouldHideChartSetting.ts
```diff
@@ -1,11 +1,35 @@
...
+): boolean => {
+  if (!isDefined(fieldMetadataId)) {
+    return true;
+  }
+
+  const field = objectMetadataItem.fields.find(
+    (field) => field.id === fieldMetadataId,
+  );
+
+  if (!isDefined(field)) {
```
ijreilly, line 20: when is that expected?

ijreilly (approved):

packages/twenty-front/src/modules/command-menu/pages/page-layout/utils/shouldHideChartSetting.ts
bosiraphael: When we switch between sources

bosiraphael:
@greptileai

PR #15545, opened 2025-11-03
Add command to make sure v1.8 workspaces are not using FULL or PARTIAL sync stages (that should be already deprecated)

charlesBochet:
In v1.8, we have already run a command to deprecate FULL or PARTIAL sync stages.

However the code was fully deprecated in v1.10 and some workspaces might still have this status used. This is to double check


guillim (approved):
Juss wondering why we use coreDatasource directly instead of using the twentyORMGlobalManager.getRepositoryForWorkspace to be more "protective"

PR #15547, opened 2025-11-03
[demo] Allow workspace to work with a non-system workspaceMember object

ijreilly:
For demo purposes, we want to work with workspaceMember has a non-system object, allowing it to be displayed on the product, to be added custom fields etc. 
It is something we may want to do later anyway. 
This PR adds a bypassPermissionChecks on workspaceMember repository calls. This does not provide more permissions than before for other workspaces: workspaceMember being a system object, permissions are already bypassed for this one. (Note that actions such inviting a new workspaceMember, updating a workspaceMember are protected by a system permission checked in the relevant places).

This work does **not** allow any workspace to switch their workspaceMember object to non-system, the switch is still manual.


Weiko (approved):
LGTM

PR #15557, opened 2025-11-03
`[REQUIRED_FOR_1_10]` Fix kanban foreign key migration

prastoin:
# Introduction
We introduced a foreign key addition that will fail in production due to orphan views targetting non existing fields

## Migration
The migration will be run for any new workspace successfully or any twenty instance without corrupted data

## Upgrade command
The upgrade command will at some point allow the migration to be run manually after removing any corrupted data

## Release note
We should remove the migration we've manually set as being run in production

### DIFF
--- a/packages/twenty-server/src/database/commands/upgrade-version-command/1-10/1-10-clean-orphaned-kanban-aggregate-operation-field-metadata-id.command.ts
+++ b/packages/twenty-server/src/database/commands/upgrade-version-command/1-10/1-10-clean-orphaned-kanban-aggregate-operation-field-metadata-id.command.ts
@@ -110,5 +110,7 @@ export class CleanOrphanedKanbanAggregateOperationFieldMetadataIdCommand extends
     this.logger.log(
       `${options.dryRun ? 'DRY RUN - Would have' : ''}Deleted ${deletedCount} KANBAN views and updated ${updatedCount} other views with orphaned kanbanAggregateOperationFieldMetadataId references`,
     );
+
+    // We need to duplicated what's done in the migration command here and try catch it globally becasue it will be run several time for all workspaces
   }
 }
--- a/packages/twenty-server/src/database/commands/upgrade-version-command/1-10/1-10-clean-orphaned-kanban-aggregate-operation-field-metadata-id.command.ts
+++ b/packages/twenty-server/src/database/commands/upgrade-version-command/1-10/1-10-clean-orphaned-kanban-aggregate-operation-field-metadata-id.command.ts
@@ -111,6 +111,21 @@ export class CleanOrphanedKanbanAggregateOperationFieldMetadataIdCommand extends
       `${options.dryRun ? 'DRY RUN - Would have' : ''}Deleted ${deletedCount} KANBAN views and updated ${updatedCount} other views with orphaned kanbanAggregateOperationFieldMetadataId references`,
     );
 
-    // We need to duplicated what's done in the migration command here and try catch it globally becasue it will be run several time for all workspaces
+    // Apply the migration manually after cleanup 1760965667836-kanbanFieldMetadataIdentifierView.ts
+    // This is workspace-agnostic and will only succeed once
+    if (!options.dryRun) {
+      try {
+        await this.viewRepository.query(
+          `ALTER TABLE "core"."view" ADD CONSTRAINT "FK_b3cc95732479f7a1337350c398f" FOREIGN KEY ("kanbanAggregateOperationFieldMetadataId") REFERENCES "core"."fieldMetadata"("id") ON DELETE CASCADE ON UPDATE NO ACTION`,
+        );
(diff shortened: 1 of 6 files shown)


packages/twenty-server/src/database/commands/upgrade-version-command/1-10/1-10-clean-orphaned-kanban-aggregate-operation-field-metadata-id.command.ts
```diff
@@ -110,5 +110,22 @@ export class CleanOrphanedKanbanAggregateOperationFieldMetadataIdCommand extends
     this.logger.log(
       `${options.dryRun ? 'DRY RUN - Would have' : ''}Deleted ${deletedCount} KANBAN views and updated ${updatedCount} other views with orphaned kanbanAggregateOperationFieldMetadataId references`,
     );
+
+    // Apply the migration manually after cleanup 1760965667836-kanbanFieldMetadataIdentifierView.ts
+    // This is workspace-agnostic and will only succeed once
+    if (!options.dryRun) {
+      try {
+        await this.viewRepository.query(
```
prastoin: Should not be view repo here

prastoin:
> This PR has a critical architectural issue with race conditions in multi-workspace scenarios

Upgrade commands are run synchronously ( unfortunately )

packages/twenty-server/src/database/commands/upgrade-version-command/1-10/1-10-clean-orphaned-kanban-aggregate-operation-field-metadata-id.command.ts
```diff
@@ -110,5 +112,22 @@ export class CleanOrphanedKanbanAggregateOperationFieldMetadataIdCommand extends
...
+        await this.coreDataSource.query(
+          `ALTER TABLE "core"."view" ADD CONSTRAINT "FK_b3cc95732479f7a1337350c398f" FOREIGN KEY ("kanbanAggregateOperationFieldMetadataId") REFERENCES "core"."fieldMetadata"("id") ON DELETE CASCADE ON UPDATE NO ACTION`,
+        );
+        this.logger.log(
+          'Successfully added foreign key constraint for kanbanAggregateOperationFieldMetadataId',
+        );
+      } catch {
+        this.logger.log(
+          'Foreign key constraint already exists or could not be created (this is expected for subsequent workspaces)',
+        );
```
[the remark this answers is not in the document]
prastoin: Agreed it's hacky but that's the only we have right now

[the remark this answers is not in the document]
prastoin: There's no race condition possible as every upgrade are run sequentially

packages/twenty-server/src/database/commands/upgrade-version-command/1-10/1-10-clean-orphaned-kanban-aggregate-operation-field-metadata-id.command.ts
```diff
@@ -110,5 +112,22 @@ export class CleanOrphanedKanbanAggregateOperationFieldMetadataIdCommand extends
     this.logger.log(
       `${options.dryRun ? 'DRY RUN - Would have' : ''}Deleted ${deletedCount} KANBAN views and updated ${updatedCount} other views with orphaned kanbanAggregateOperationFieldMetadataId references`,
     );
+
+    // Apply the migration manually after cleanup 1760965667836-kanbanFieldMetadataIdentifierView.ts
+    // This is workspace-agnostic and will only succeed once
+    if (!options.dryRun) {
+      try {
+        await this.coreDataSource.query(
```
Weiko: Could we implement a runGlobally/runOnce in ActiveOrSuspendedWorkspacesMigrationCommandRunner for things we want to run only once @prastoin @charlesBochet this would avoid this query to be executed on each workspace

prastoin: Agreed, but here is not really applicable
As several workspace could contain corrupted data

prastoin: We could think of an upcoming command that will only run once using a static value though

prastoin: Done ! nice catch

prastoin:
@greptileai trigger

Weiko (approved):
LGTM

PR #15559, opened 2025-11-03
Graph margins vary depending on presence of labels

ijreilly:
https://github.com/user-attachments/assets/3c27fa7b-78a3-49eb-9795-53571662cf12


bosiraphael (approved):
LGTM

PR #15568, opened 2025-11-03
Anchored tooltip

ehconitin:
Issue - 
- tooltip gets too long when groupby has too much data points
- we need max height on tooltips
- we need to be able to scroll inside the tooltip
- not possible if the tooltip is following the cursor (which library enforces)
- its not easy to customize Nivo's own tooltip to make it work how we want it

what we want - 
- let the tooltip anchor to the bar in such a way -- that the top of the bar aligns with the vertical middle of the tooltip
- add max height to the tooltip

what I did - 
- use floating portal from floating-ui/react
- get the hover datum (ie hovered bars) dimensions on mouse enter to render the tooltip
- anchor the tooltip at the calculated position
- floating-ui handles the basic things like flipping/offset/shift
- clear the states as required


lucasbordeau:
Could you provide a description in your PR and also link an existing issue ?

lucasbordeau:
@Bonapara I suggest that we put in bold the line in the tooltip that corresponds to the hovered group, example : 

Capture d’écran 2025-11-04 à 11 34 41

Bonapara:
Great idea @lucasbordeau - medium + darker colors (secondary for the label & tertiary for the value)

ehconitin:
@lucasbordeau mb, updated the description!

and will push the new requirements shortly after

ehconitin:
@greptileai

lucasbordeau:
I think that you shouldn't introduce new conventions to the project, stick to the existing, we don't use `$var`, so let's not use it for Styled components because that is not used in the app. Let's avoid introducing new conventions, look at how it is done in other components and just copy, because the app is very consistent already and you don't need to reinvent the wheel for conventions.

lucasbordeau:
Let's add a story for tooltip vertical scroll with too many items to display

packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetBarChart/components/CustomBarItem.tsx
```diff
@@ -87,10 +88,20 @@ export const CustomBarItem = <D extends BarDatum>({
...
+      const barDatumWithGeometry = {
+        ...barData,
+        barElement: event.currentTarget,
+        barX: bar.x,
+        barY: bar.y,
+        barWidth: bar.width,
+        barHeight: bar.height,
+        barAbsoluteX: bar.absX,
+        barAbsoluteY: bar.absY,
+      } as BarDatumWithGeometry;
```
lucasbordeau: prefere either satisfy or barDatumWithGeometry: BarDatumWithGeometry because as is less safe

packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetBarChart/components/CustomBarItem.tsx
```diff
@@ -87,10 +88,20 @@ export const CustomBarItem = <D extends BarDatum>({
...
+      onMouseEnter?.(barDatumWithGeometry as typeof barData, event);
```
lucasbordeau: Not sure that using as here is a good idea, couldn't change the type of the argument of onMouseEnter ?

packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetBarChart/hooks/useBarChartFloatingTooltip.ts
```diff
@@ -0,0 +1,123 @@
...
+        width: datum.barWidth!,
+        height: 1,
+        x: svgContainerRectangle.left + datum.barAbsoluteX!,
+        y: svgContainerRectangle.top + datum.barAbsoluteY!,
+      };
+    },
+  };
+};
+
+export const useBarChartFloatingTooltip = () => {
```
lucasbordeau: Please break down this hook, one hook per returned reactive function.

lucasbordeau: The problem with barrel hooks like this one is that it makes every caller subscribe to ALL reactive functions, even if a caller does not need to subscribe, which we should avoid.

And as a clean code rule of thumb, it is generally better to have small functions, small hooks, small components, instead of big component, big hooks, big functions that do everything in one place, this is bug-prone, less readable and less maintainable in general.

packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetBarChart/components/GraphWidgetBarChart.tsx
```diff
@@ -265,21 +294,28 @@ export const GraphWidgetBarChart = ({
...
-            }
-          }}
-          onMouseLeave={() => setHoveredBar(null)}
+          onMouseEnter={handleBarMouseEnter}
+          onMouseLeave={handleBarMouseLeave}
           theme={chartTheme}
           borderRadius={parseInt(theme.border.radius.sm)}
         />
       </GraphWidgetChartContainer>
+      {isDefined(hoveredBarDatum) && (
```
lucasbordeau: Could you extract this in another component ?

packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetBarChart/hooks/useBarChartFloatingTooltip.ts
```diff
@@ -0,0 +1,123 @@
...
+
+export type BarDatumWithGeometry = ComputedDatum<BarDatum> & {
+  barElement?: SVGElement;
+  barAbsoluteX?: number;
+  barAbsoluteY?: number;
+  barWidth?: number;
+  barHeight?: number;
+};
+
+const createFloatingAnchorFromBarGeometry = (
```
lucasbordeau: Extract this in a file

packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetBarChart/components/GraphWidgetBarChart.tsx
```diff
@@ -99,11 +101,36 @@ export const GraphWidgetBarChart = ({
...
+    scheduleTooltipHide,
+    cancelTooltipHide,
+    hideTooltip,
+  } = useBarChartFloatingTooltip();
+
+  const handleBarMouseEnter = (datum: BarDatumWithGeometry) => {
+    showTooltipForBar(datum);
+  };
+
+  const handleBarMouseLeave = () => {
```
lucasbordeau: The handlers are good, let's now break down the functions in small standalone hooks

lucasbordeau:
@Bonapara Should the lines in the tooltip be ordered in the same order as the bars on the graph ? When using it it feels odd that it is not the case, because we already have name + color for our eyes to identify quickly which group it is.

The current behavior is that the tooltip applies the same ordering for any bar.

@ehconitin This can be made in another PR, let's create an issue if we want to improve in this direction.

image

lucasbordeau:
Problem with the bars ? 

image

lucasbordeau:
The tooltip doesn't get closed when the mouse leaves, even if we click outside or press escape, that's a problem, the tooltip should be closed reliably.


https://github.com/user-attachments/assets/f00700bd-cb21-457f-ad57-c18b9bae65ef

packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetBarChart/components/CustomBarItem.tsx
```diff
@@ -87,17 +96,34 @@ export const CustomBarItem = <D extends BarDatum>({
...
+      bar,
+      onShowTooltip,
+      onMouseEnter,
+      showTooltipFromEvent,
+      renderTooltip,
+    ],
   );
   const handleMouseLeave = useCallback(
     (event: MouseEvent<SVGRectElement>) => {
+      onScheduleHideTooltip?.();
```
lucasbordeau: This is mixing responsability and concerns, a custom bar item shouldn't be aware of the tooltip, so there shouldn't be events for tooltip here.

The BarItem just tells callers that it is hovered or not, so onBarHover, onBarLeave, then GraphWidgetBarChat just knows what bar is hovered with its state, and it can tell if it wants to show a tooltip or not. 

Then the tooltip is responsible for its floating ui logic, and it just takes what it needs to compute the virtual element inside itself. 

Let's refactor this, this is a good exercise for you to really nail the separation of concerns at this level, that's important.

packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetBarChart/components/GraphWidgetBarChartTooltip.tsx
```diff
@@ -0,0 +1,86 @@
...
+  data: BarChartDataItem[];
+  indexBy: string;
+  formatOptions: GraphValueFormatOptions;
+  enableGroupTooltip?: boolean;
+  onCancelHide: () => void;
+  onRequestHide: () => void;
+};
+
+export const GraphWidgetBarChartTooltip = ({
+  floatingRef,
```
lucasbordeau: Floating refs shouldn't be passed as props but computed inside this component.

packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetBarChart/utils/createFloatingAnchorFromBarGeometry.ts
```diff
@@ -0,0 +1,47 @@
+import { type VirtualElement } from '@floating-ui/react';
+import { isDefined } from 'twenty-shared/utils';
+
+export type BarGeometry = {
```
lucasbordeau: Let's export this in a standalone file

ehconitin:
> The tooltip doesn't get closed when the mouse leaves, even if we click outside or press escape, that's a problem, the tooltip should be closed reliably.
> 
>  Enregistrement.de.l.ecran.2025-11-06.a.11.22.36.mov

weird -- this never happens in app -- only in the story books
same with bars -- checking!

ehconitin:
WIP - I am trying to refactor the tooltip in a way that it would be reused by all the charts elegantly!
will first try and see if what I am thinking would work, and then discuss with @lucasbordeau if this abstraction is up to par with what this codebase should have

am heavily trying to do what Nivo does with their tooltip -- but with our own requirements!

ehconitin:
Code up to this commit violates AHA principles - fell into over-abstraction!
first thoughts - was trying to code a provider that would serve all the charts
for eg a graph could just do 
```
   showTooltip(
        event.currentTarget,
        tooltipData,
      );     
```

currentTarget being the one where the tooltip is supposed to 'anchor'

but then, on applying the same principle to the line chart -- realized each chart would have its distinct interaction
for example - 
in the bar chart, the tooltip anchors at the top position of the hovered bar
but in the line chart -- it should anchor on to the slice where the crosshair is and at the mouse Y position

complete diff requirements - 
Doubling down, I switched to two separate providers! -- which was again dumb since they are tightly coupled now, serving one provider per char

Now I will go back to simple, contained hooks and states

But it works nicely now -- has to set the floating boundary to that of the chart and not the root! -- open to discuss if product wants root!

@lucasbordeau fyi

ehconitin:
https://github.com/user-attachments/assets/28dce879-4e1c-4756-a073-8926f2811d2f

ehconitin:
@greptileai

packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetBarChart/components/CustomBarItem.tsx
```diff
@@ -218,8 +202,8 @@ export const CustomBarItem = <D extends BarDatum>({
               <animated.rect
                 x={clipPathX}
                 y={clipPathY}
-                rx={clipRx}
-                ry={clipRy}
+                rx={clipBorderRadiusX}
```
lucasbordeau, line 188: Nice one, indeed, when a library doesn't offer clean naming, we're free to use well-named variables anyway

packages/twenty-front/src/modules/page-layout/widgets/graph/hooks/useTooltipFloating.ts
```diff
@@ -0,0 +1,48 @@
...
+  flip,
+  offset,
+  shift,
+  useFloating,
+  type VirtualElement,
+} from '@floating-ui/react';
+import { useMemo } from 'react';
+import { isDefined } from 'twenty-shared/utils';
+
+export const useTooltipFloating = (
```
lucasbordeau: I would say useGraphWidgetTooltip to stay coherent

packages/twenty-front/src/modules/page-layout/widgets/graph/components/GraphWidgetFloatingTooltip.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+type GraphWidgetFloatingTooltipProps = {
+  tooltipData: GraphWidgetTooltipData;
+  onScheduleHide?: () => void;
+  onCancelScheduledHide?: () => void;
+};
+
+export const GraphWidgetFloatingTooltip = ({
+  tooltipData,
+  onScheduleHide,
```
lucasbordeau: That's still not at the right semantic level, onMouseEnter and onMouseLeave here too, or onTooltipFocus, onTooltipBlur if you want to be more generic, the tooltip shouldn't know how it gets hidden by its parent, because that's not his responsability to show himself, this component just handles placement and floating mechanism.

packages/twenty-front/src/modules/page-layout/widgets/graph/components/GraphWidgetFloatingTooltip.tsx
```diff
@@ -0,0 +1,91 @@
...
+  onCancelScheduledHide,
+}: GraphWidgetFloatingTooltipProps) => {
+  const theme = useTheme();
+
+  const { reference, boundary } = useMemo((): {
```
lucasbordeau: Could be cleaned a bit but at least it's at the right abstraction level

packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetBarChart/hooks/useBarChartHandlers.ts
```diff
@@ -12,13 +11,10 @@ export const useBarChartHandlers = ({
   data,
   indexBy,
 }: UseBarChartHandlersProps) => {
-  const [hoveredBar, setHoveredBar] = useState<{
-    key: string;
-    indexValue: string | number;
-  } | null>(null);
-
   const handleBarClick = (datum: ComputedDatum<BarDatum>) => {
```
lucasbordeau: Can't we put those directly into the bar chart component ? We try to avoid creating external handlers as it is less readable.

The idea is that a component have the hanlders and the JSX part always in the component at the same place, so it's easier to read any component in the codebase. 

External logic is in hooks, but handlers always stay inside the component and they call external hooks.

packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetBarChart/hooks/useBarChartTooltip.ts
```diff
@@ -8,42 +8,50 @@ import { type BarDatum, type ComputedDatum } from '@nivo/bar';
...
-  hoveredBar: { key: string; indexValue: string | number } | null;
   enrichedKeys: BarChartEnrichedKey[];
   data: BarChartDataItem[];
   indexBy: string;
   formatOptions: GraphValueFormatOptions;
   enableGroupTooltip?: boolean;
+  layout?: 'vertical' | 'horizontal';
 };
 
 export const useBarChartTooltip = ({
```
lucasbordeau: Can we create a util instead of a hook here ? Because there isn't any reactive dependency.

packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetBarChart/components/GraphWidgetBarChart.tsx
```diff
@@ -264,22 +295,25 @@ export const GraphWidgetBarChart = ({
...
-          onMouseLeave={() => setHoveredBar(null)}
+          onMouseEnter={handleBarEnter}
+          onMouseLeave={handleBarLeave}
           theme={chartTheme}
           borderRadius={parseInt(theme.border.radius.sm)}
         />
       </GraphWidgetChartContainer>
+
+      {activeTooltipData && (
+        <GraphWidgetFloatingTooltip
```
lucasbordeau: Here you could create a GraphBarChartTooltip component, that would call getBarChartTooltipData, instead of doing `tooltipData.anchor.type === 'bar-element-anchor'`.

Here the important point is that it is better to create one GraphWidgetTooltip per type of chart instead of creating a generic tooltip component, since the logic is different.

lucasbordeau: Think of those types of components as thin wrappers that allow creating different code paths for each use case, instead of creating a giant object with everything and passing it to a generic code path that will force to create switch everywhere in the generic tree.

It's always easier to maintain to have a bit of duplication and parallel code path with thin wrappers that have a lot of semantic in their naming and serve not only as code but also to carry meaning and identify code paths more clearly than to rely on a completely dynamic generic code path that respond to data.

packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetBarChart/components/GraphWidgetBarChart.tsx
```diff
@@ -264,22 +295,25 @@ export const GraphWidgetBarChart = ({
...
-          }}
-          onMouseLeave={() => setHoveredBar(null)}
+          onMouseEnter={handleBarEnter}
+          onMouseLeave={handleBarLeave}
           theme={chartTheme}
           borderRadius={parseInt(theme.border.radius.sm)}
         />
       </GraphWidgetChartContainer>
+
+      {activeTooltipData && (
```
lucasbordeau: Let's use a boolean variable `shouldShowTooltip` instead ?

lucasbordeau (commented):
Overall that's better yes, we could refine the behavior of the tooltips with Bonapara later, but for now let's focus on the code, I think that you can push it a little bit further.

packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetLineChart/constants/LineChartMarginTop.ts
```diff
@@ -0,0 +1 @@
+export const LINE_CHART_MARGIN_TOP = 20;
```
charlesBochet, line 1: polish: we could group these into one file:
export const LINE_CHART_MARGINS = { right: / top: }

charlesBochet (approved):

PR #15573, opened 2025-11-03
Fix chart filter settings back button behavior

bosiraphael:
Before:

https://github.com/user-attachments/assets/23881978-3234-4a0b-95ec-15470d26026d

After:

https://github.com/user-attachments/assets/80d222f6-585a-4fff-8476-dae327e34699


ehconitin (approved):
lgtm!

PR #15577, opened 2025-11-03
add inner padding on bar chart

ehconitin:
before:

CleanShot 2025-11-03 at 22 06 45

after:

CleanShot 2025-11-03 at 22 06 08


bosiraphael (approved):
LGTM

Weiko (approved):

PR #15578, opened 2025-11-03
Remove relations from manual trigger output schema

thomtrp:
Fixes https://github.com/twentyhq/twenty/issues/15319
Fixes https://github.com/twentyhq/twenty/issues/15255

On index page, we load record with only a few relation fields. So record selected on manual trigger do not contain all the fields. To match that behavior, let's remove relations from output schema. The user will do a search on relation id if needed.


Weiko (approved):
LGTM!

martmull (approved):

PR #15580, opened 2025-11-03
fix inconsistent widget placeholder placement

ehconitin:
fixes - https://github.com/twentyhq/twenty/pull/15496#issuecomment-3481636780

and https://discord.com/channels/1130383047699738754/1434971632539009055

- make the pending placeholder to be non-static -- so that it behaves similarly to compact behavior like other widgets do
- make sure users can't set dashboards to be opened in the  side panel -- similar to workflows

before- 

https://github.com/user-attachments/assets/d991db56-1388-4e4b-b743-a63f56b6187a

after- 

https://github.com/user-attachments/assets/9031f2c6-1e3b-4604-a5a0-3124d14484ce


lucasbordeau (approved):
LGTM

PR #15582, opened 2025-11-03
(Breaking change) Switch between set password and change password on the settings page.

mabdullahabaid:
Here is what the PR does:

- Surface password state in validatePasswordResetToken, returning hasPassword so the client can tell whether a user is setting or changing their password.
- Consume that flag throughout the front end (mock data, stories, GraphQL types) and update the Reset/Set Password modal to swap the heading/button label and success toast accordingly.
- After a successful password set/reset, immediately update the logged-in user’s hasPassword flag so the Settings screen reflects the new state without a reload.

Modal has two states now - reset password modal uses change password state since it made intuitive sense.

<p align="center">
image
</p>

<p align="center">
image
</p>


packages/twenty-front/src/pages/auth/PasswordReset.tsx
```diff
@@ -147,9 +154,18 @@ export const PasswordReset = () => {
         return;
       }
 
+      const successMessage =
+        isTargetUserPasswordSet === false
+          ? t`Password has been set`
+          : t`Password has been updated`;
+
+      setCurrentUser((currentUser) =>
```
mabdullahabaid, line 162: This causes a re-render, but we need to update the currentUserState for settings to display "Change Password" once the user without password has actually set a password, otherwise the state does not update until manual reload.

mabdullahabaid:
@FelixMalfait the storybook test is failing since I replaced `Trans` with `t` in `PasswordReset.tsx`. 

I have spent a few hours trying to figure out how to mock `t`, but I am unable to figure out how to fix this test. Can you point me in the direction?

FelixMalfait:
@mabdullahabaid did you try adding `I18nFrontDecorator` already?

FelixMalfait (approved):
Great work!

PR #15584, opened 2025-11-03
refactor reconnect account logic

neo773:


packages/twenty-server/src/modules/calendar/calendar-event-import-manager/jobs/calendar-relaunch-failed-calendar-channel.job.ts
```diff
@@ -0,0 +1,69 @@
...
+    ) {
+      return;
+    }
+
+    await calendarChannelRepository.update(calendarChannelId, {
+      syncStage: CalendarChannelSyncStage.CALENDAR_EVENT_LIST_FETCH_PENDING,
+      syncStatus: CalendarChannelSyncStatus.ACTIVE,
+    });
+
+    await this.accountsToReconnectService.removeAccountToReconnect(
```
charlesBochet: @neo773 as we are handling FAILED status and we said that this banner should only be there for missing permissions, should we have this call here?

packages/twenty-server/src/modules/calendar/calendar-event-import-manager/jobs/calendar-relaunch-failed-calendar-channel.job.ts
```diff
@@ -0,0 +1,69 @@
...
+export type CalendarRelaunchFailedCalendarChannelJobData = {
+  workspaceId: string;
+  calendarChannelId: string;
+};
+
+@Processor({
+  queueName: MessageQueue.calendarQueue,
+  scope: Scope.REQUEST,
+})
+export class CalendarRelaunchFailedCalendarChannelJob {
```
charlesBochet, line 22: I was wondering if we should also add a metric but I think this is fine as we have the FAILED syncStage change metric in the first place

LGTM

packages/twenty-server/src/modules/connected-account/refresh-tokens-manager/drivers/google/services/google-api-refresh-tokens.service.ts
```diff
@@ -23,19 +23,25 @@ export class GoogleAPIRefreshAccessTokenService {
...
-    if (!isDefined(token)) {
-      throw new ConnectedAccountRefreshAccessTokenException(
-        'Failed to refresh google access token',
-        ConnectedAccountRefreshAccessTokenExceptionCode.REFRESH_ACCESS_TOKEN_FAILED,
-      );
+    try {
+      const { token } = await oAuth2Client.getAccessToken();
+
+      return {
+        accessToken: token as string,
```
charlesBochet: I prefer the previous implementation with runtime typing (isDefined + throw)

charlesBochet (approved):
LGTM, left a minor comment

packages/twenty-server/src/modules/calendar/calendar-event-import-manager/jobs/calendar-relaunch-failed-calendar-channel.job.ts
neo773: Actually yeah, I overlooked it, resolved.

PR #15585, opened 2025-11-04
Fix groupMode toggle

ehconitin:
groupMode should be undefined if no groupby is set on the secondary axis!
This default value is also the reason -- all the logic -- where groupMode was checked for conditional rendering -- for eg
, negative data labels when there is no groupby -- the labels should appear below wasn't happening -- since there was always a default to groupMode!
changes - 
- getting rid of the default value on the bar chart DTO for groupMode
- on front -- make sure groupMode is only set when the secondary axis gets groupBy id and proper cleanup


ehconitin:
@greptileai

lucasbordeau (approved):
LGTM

PR #15587, opened 2025-11-04
Remove default feature flag feature

Weiko:
## Context
Having rollout feature for new workspaces creates bad experience for new users and doesn't bring as much value as existing ones anyway. Removing migration v2 from default feature flag and the whole concept of default feature flag


bosiraphael (approved):
LGTM

PR #15588, opened 2025-11-04
Filter out null values in tooltip

bosiraphael:
Before:
CleanShot 2025-11-04 at 11 44 14@2x

After:
CleanShot 2025-11-04 at 11 43 49@2x


Weiko (approved):
LGTM!

PR #15591, opened 2025-11-04
Create many view groups

prastoin:
# Introduction
Same as https://github.com/twentyhq/twenty/pull/15576 but for view groups creation

When creating a kanban view with v2 flag activated in production result in race condition due to request being slow and //.
That's why we're introducing a batch create on view group here
closing https://github.com/twentyhq/core-team-issues/issues/1847

## In v2
- batch create view group endpoint is available
- frontend will target the new endpoint

## In v1
- batch create view group endpoint is not available
- frontend will stick to old fake batch view group creation loop

### v2
image

### v1
image


Weiko (approved):
LGTM!

PR #15601, opened 2025-11-04
Create a feature flag for dashboards v2

bosiraphael:
Prevent users from creating v2 chart types via the api.
Only created unit tests and not integration tests (since it's not that important, and the v2 will be released soon), but tested via the api playground.


Weiko (approved):
LGTM

PR #15606, opened 2025-11-04
Set default run limit in cache

thomtrp:
Will be used when there are too many workflow runs to enqueue in workspace


packages/twenty-server/src/modules/workflow/workflow-runner/workflow-run-queue/workspace-services/workflow-run-queue.workspace-service.ts
```diff
@@ -74,8 +75,10 @@ export class WorkflowRunQueueWorkspaceService {
   ): Promise<number> {
     const currentCount =
       await this.getCurrentWorkflowRunQueuedCount(workspaceId);
+    const throttleLimit =
+      await this.getWorkflowRunQueueThrottleLimit(workspaceId);
```
charlesBochet, line 79: question: so you plan to create this key in redis manually when needed?

packages/twenty-server/src/modules/workflow/workflow-runner/workflow-run-queue/utils/get-cache-workflow-run-queue-throttle-limit-key.util.ts
```diff
@@ -0,0 +1,3 @@
+export const getWorkflowRunQueueThrottleLimitKey = (
+  workspaceId: string,
+): string => `workflow-run-queue-throttle-limit:${workspaceId}`;
```
martmull, line 3: return { ... }

martmull (approved):

packages/twenty-server/src/modules/workflow/workflow-runner/workflow-run-queue/workspace-services/workflow-run-queue.workspace-service.ts
thomtrp: yes

PR #15608, opened 2025-11-04
Move browser-extension to proper folder

martmull:
As title
- move all applications and browser extension into a `packages/twenty-apps/hacktoberfest-2025` folder


yarn.lock
```diff
@@ -17,16 +17,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@1natsu/wait-element@npm:^4.1.2":
```
charlesBochet, line 20: @martmull are we sure about all these changes? does twenty server / front still work?

charlesBochet (approved):
LGTM, left a question to double check

yarn.lock
martmull: checking

Weiko (approved):
LGTM

yarn.lock
martmull: good

PR #15614, opened 2025-11-04
Fix user deletion flows

ijreilly:
**Before**
- any user with workpace_members permission was able to remove a user from their workspace. This triggered the deletion of workspaceMember + of userWorkspace, but did not delete the user (even if they had no workspace left) nor the roleTarget (acts as junction between role and userWorkspace) which was left with a userWorkspaceId pointing to nothing. This is because roleTarget points to userWorkspaceId but the foreign key constraint was not implemented
- any user could delete their own account. This triggered the deletion of all their workspaceMembers, but not of their userWorkspace nor their user nor the roleTarget --> we have orphaned userWorkspace, not technically but product wise - a userWorkspace without a workspaceMember does not make sense

So the problems are
- we have some roleTargets pointing to non-existing userWorkspaceId (which caused https://github.com/twentyhq/twenty/issues/14608 )
- we have userWorkspaces that should not exist and that have no workspaceMember counterpart
- it is not possible for a user to leave a workspace by themselves, they can only leave all workspaces at once, except if they are being removed from the workspace by another user

**Now**
- if a user has multiple workspaces, they are given the possibility to leave one workspace while remaining in the others (we show two buttons: Leave workspace and Delete account buttons). if a user has just one workspace, they only see Delete account
- when a user leaves a workspace, we delete their workspaceMember, userWorkspace and roleTarget. If they don't belong to any other workspace we also soft-delete their user
- soft-deleted users get hard deleted after 30 days thanks to a cron
- we have two commands to clean the orphans roleTarget and userWorkspace (TODO: query db to see how many must be run)

**Next**
- once the commands have been run, we can implement and introduce the foreign key constraint on roleTarget

Fixes https://github.com/twentyhq/twenty/issues/14608


packages/twenty-server/src/engine/twenty-orm/utils/format-column-name-for-relation-field.util.ts
```diff
@@ -7,6 +7,10 @@ export const formatColumnNameForRelationField = (
   fieldName: string,
   fieldMetadataSettings: FieldMetadataRelationSettings,
 ): string => {
+  if (fieldMetadataSettings.relationType === RelationType.ONE_TO_MANY) {
```
ijreilly, line 11: forgotten from another pr

packages/twenty-server/src/database/commands/upgrade-version-command/1-11/1-11-clean-orphaned-role-targets.command.ts
```diff
@@ -0,0 +1,91 @@
...
+    protected readonly twentyORMGlobalManager: TwentyORMGlobalManager,
+  ) {
+    super(workspaceRepository, twentyORMGlobalManager);
+  }
+
+  override async runOnWorkspace({
+    options,
+  }: RunOnWorkspaceArgs): Promise<void> {
+    if (this.hasRunOnce) {
+      this.logger.log('Skipping kanban field metadata id foreign key creation');
```
Weiko: I guess you didn't mean kanban field here? 🙂

packages/twenty-server/src/database/commands/upgrade-version-command/1-11/1-11-clean-orphaned-role-targets.command.ts
```diff
@@ -0,0 +1,91 @@
...
+      return;
+    }
+    const isDryRun = options.dryRun || false;
+
+    if (isDryRun) {
+      this.logger.log('Dry run mode: No changes will be applied');
+    }
+
+    const orphanedRoleTargets = await this.roleTargetsRepository
+      .createQueryBuilder('roleTarget')
```
Weiko, line 47: nit: did we need to use builder instead of a repo here? (not super important, just curious)

packages/twenty-server/src/engine/core-modules/user/services/user.service.ts
```diff
@@ -99,78 +104,119 @@ export class UserService extends TypeOrmQueryService<UserEntity> {
...
+        id: userId,
+      },
+      withDeleted: true,
+    });
+  }
 
-        const workspaceMember = workspaceMembers.find(
-          (member: WorkspaceMemberWorkspaceEntity) => member.userId === userId,
-        );
+  async deleteUserWorkspaceAndPotentiallyDeleteUser({
```
Weiko, line 121: This method is running multiple mutations, we could have orphan data if any step fails midway. Maybe we can run this in a transaction to avoid that

Weiko: Actually not that easy though since you are running operations with different datasources (on multiple schemas) so different connections

packages/twenty-front/src/modules/object-record/hooks/useFindManyRecords.ts
```diff
@@ -85,7 +85,7 @@ export const useFindManyRecords = <T extends ObjectRecord = ObjectRecord>({
 
   const hasReadPermission = objectPermissions.canReadObjectRecords;
 
-  const { data, loading, error, fetchMore } =
+  const { data, loading, error, fetchMore, refetch } =
```
Weiko, line 88: I'm not sure about exposing the refetch for useFindMany for this new use case only 🤔 Could we use optimistic update / apollo cache mutation instead?

Weiko (commented):
Left some comments otherwise LGTM

packages/twenty-front/src/modules/object-record/hooks/useFindManyRecords.ts
ijreilly: why not? by fear it would be use in the wrong context?

ijreilly:
@Bonapara @StephanieJoly4 @FelixMalfait 
Do you agree with soft-deleted users being hard-deleted after 30 days (unlike today - they are never hard deleted), or should we keep them forever?

@etiennejouan @charlesBochet fyi

packages/twenty-server/src/database/commands/upgrade-version-command/1-11/1-11-clean-orphaned-role-targets.command.ts
ijreilly: oops

packages/twenty-server/src/database/commands/upgrade-version-command/1-11/1-11-clean-orphaned-role-targets.command.ts
ijreilly: the join + null filtering pattern requires createQueryBuilder

packages/twenty-server/src/engine/core-modules/user/services/user.service.ts
ijreilly: right - what's your recommendation on this then? I also thought we were not to do any transaction

packages/twenty-front/src/modules/object-record/hooks/useFindManyRecords.ts
ijreilly: @lucasbordeau wdyt?

packages/twenty-server/src/engine/core-modules/user/services/user.service.ts
Weiko: Right, I think it's too complex with the tooling we have now. Let's just keep it in mind. LGTM for now!

packages/twenty-front/src/modules/object-record/hooks/useFindManyRecords.ts
lucasbordeau: There's dedicated tooling for fetchMore and doing a refetch like that has not been QAed, so I would not expose this method.

Weiko (approved):
LGTM

PR #15624, opened 2025-11-05
Fix missing creation attributes

martmull:
Fixes serverlessFucntion are created with default handlerPath and handlerName values


Weiko (approved):
You probably want some kind of validation otherwise LGTM

PR #15648, opened 2025-11-05
Edit workflow and serverless throttling

thomtrp:
Serverless:  1000 / min.

Workflows: 100 / min. This is a security in case of infinite loops.

### DIFF
--- a/packages/twenty-server/src/modules/workflow/workflow-runner/jobs/run-workflow.job.ts
+++ b/packages/twenty-server/src/modules/workflow/workflow-runner/jobs/run-workflow.job.ts
@@ -172,8 +172,9 @@ export class RunWorkflowJob {
 
   private async throttleExecution(workflowId: string) {
     try {
-      await this.throttlerService.throttle(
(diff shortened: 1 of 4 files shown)


thomtrp:
@greptileai

Weiko (approved):
LGTM, need to update env var right before the deploy then

packages/twenty-server/src/modules/workflow/workflow-runner/jobs/run-workflow.job.ts
```diff
@@ -172,8 +172,9 @@ export class RunWorkflowJob {
 
   private async throttleExecution(workflowId: string) {
     try {
-      await this.throttlerService.throttle(
+      await this.throttlerService.tokenBucketThrottleOrThrow(
```
charlesBochet, line 175: this soft TTL is perfect. We should also add a hard TTL to prevent running the job if it's above this limit

charlesBochet (approved):

PR #15655, opened 2025-11-05
Fix scroll to start when resize or move around columns

ijreilly:
Fixes https://github.com/twentyhq/private-issues/issues/361

There's room for more improvement here - triggerInitialRecordTableDataLoad does a lot of things, should not be triggered so much. actually even RecordTableVirtualizedInitialDataLoadEffect should not be triggered when there's just a metadata field change (if we trust the name initialDataLoad)

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-table/virtualization/components/RecordTableVirtualizedInitialDataLoadEffect.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-table/virtualization/components/RecordTableVirtualizedInitialDataLoadEffect.tsx
@@ -79,15 +79,8 @@ export const RecordTableVirtualizedInitialDataLoadEffect = () => {
         const lastFields = lastContextStoreVisibleRecordFields || [];
         const currentFields = visibleRecordFields || [];
 
-        const shouldRefetchData = lastFields.some((lastField, index) => {
-          const currentField = currentFields[index];
-          return (
-            !currentField ||
-            lastField.fieldMetadataItemId !==
-              currentField.fieldMetadataItemId ||
-            lastField.isVisible !== currentField.isVisible
-          );
-        });
+        const shouldRefetchData = currentFields.length > lastFields.length;
+
         if (shouldRefetchData) {
           await triggerInitialRecordTableDataLoad({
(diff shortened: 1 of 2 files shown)


packages/twenty-front/src/modules/object-record/record-table/virtualization/components/RecordTableVirtualizedInitialDataLoadEffect.tsx
```diff
@@ -75,7 +76,16 @@ export const RecordTableVirtualizedInitialDataLoadEffect = () => {
...
 
-        await triggerInitialRecordTableDataLoad();
+        const lastFields = lastContextStoreVisibleRecordFields || [];
+        const currentFields = visibleRecordFields || [];
+
+        const shouldRefetchData = currentFields.length > lastFields.length;
+
+        if (shouldRefetchData) {
+          await triggerInitialRecordTableDataLoad({
+            shouldScrollToStart: isEmpty(lastFields),
```
ijreilly, line 86: isEmpty(lastFields) helps determining if it is an initial load or not

packages/twenty-front/src/modules/object-record/record-table/virtualization/components/RecordTableVirtualizedInitialDataLoadEffect.tsx
```diff
@@ -75,7 +76,16 @@ export const RecordTableVirtualizedInitialDataLoadEffect = () => {
       ) {
         setLastContextStoreVisibleRecordFields(visibleRecordFields);
 
-        await triggerInitialRecordTableDataLoad();
+        const lastFields = lastContextStoreVisibleRecordFields || [];
+        const currentFields = visibleRecordFields || [];
+
+        const shouldRefetchData = currentFields.length > lastFields.length;
```
lucasbordeau, line 82: What if we hide a field ?

ijreilly: Then we don't need to fetch the data again because it is hidden

lucasbordeau (approved):
LGTM

PR #15664, opened 2025-11-06
Add isUiReadOnly and isUnique to field decorators

martmull:
Add missing fields properties


Weiko (commented):
I'm not sure we want to expose isUIReadOnly, this shouldn't be used in place of permissions and could be confused. This is mostly for computed fields in the future and agents managed fields. I'm not fully against it I just feel like it could be confused by users.
As for isUnique imho we should deprecate this since we have composite indexes. Unicity constraint should be on the index metadata level (and in your case, its decorator)

PR #15667, opened 2025-11-06
Increase main chunk size to 6.1MB

thomtrp:
As title


Weiko (approved):

Weiko:
LGTM. We need to catch this before merging new prs. As suggested by Felix we could introduce this build to the ci

PR #15686, opened 2025-11-06
Fix AI chat scroll behavior

abdulrahmancodes:


packages/twenty-front/src/modules/ai/components/AIChatTab.tsx
```diff
@@ -80,6 +81,31 @@ export const AIChatTab = () => {
   const { createChatThread } = useCreateNewAIChatThread();
   const { navigateCommandMenu } = useCommandMenu();
 
+  const handleScrollToBottom = (element: HTMLDivElement | null) => {
+    if (!isDefined(element)) {
+      return;
+    }
+
+    const scrollElement = document.getElementById(
```
charlesBochet: please use useScrollWrapperHTMLElement

charlesBochet: we also have states: scrollWrapperScrollBottomComponentState...

charlesBochet (approved):
LGTM codewise

charlesBochet (changes requested):
Actually I missed something :p leaving a comment

packages/twenty-front/src/modules/ai/components/AIChatTab.tsx
```diff
@@ -109,6 +131,7 @@ export const AIChatTab = () => {
                   />
                 );
               })}
+              <div ref={handleScrollToBottom} />
```
charlesBochet: I'm very confused by this ref taking a callback as value, what's the reasoning behind it?

abdulrahmancodes: The reason we’re using a callback ref here is to trigger the “scroll to bottom” logic at the exact moment the sentinel element is mounted, without going through another useEffect. If you think it would read clearer with a different pattern, happy to switch.

charlesBochet: @abdulrahmancodes yes, please extract handleScrollToBottom in a hook and call this hook synchronously when we open AIChat :)

abdulrahmancodes: @charlesBochet Done, moved the scroll logic into a hook

packages/twenty-front/src/modules/ai/hooks/useAgentChatScrollToBottom.ts
```diff
@@ -0,0 +1,30 @@
...
+}: {
+  scrollWrapperId: string;
+  messages: ExtendedUIMessage[];
+}) => {
+  const { scrollWrapperHTMLElement } =
+    useScrollWrapperHTMLElement(scrollWrapperId);
+
+  const currentAIChatThread = useRecoilValue(currentAIChatThreadState);
+
+  useLayoutEffect(() => {
```
charlesBochet: @abdulrahmancodes this is quite hacky.

1) useEffect or useLayoutEffect lead to race condition between effects, and impredictable behaviors. If there is a way to execute the logic synchronously (on browser event), that's better. Here I think we want to auto-scroll to the bottom when: the user opens the chat, the chat content is updated. This hook should expose a callback and the callback should be simply called when needed

2) here what you are doing is very hacky, registering a layoutEffect depending on currentAIChatThread which is actually not used within the LayoutEffect

abdulrahmancodes: @charlesBochet 
There are three cases where we need to scroll to the bottom:
1. When the chat is opened or the thread changes  
2. When a new message is sent  
3. When a new streaming chunk is received  

The first two can definitely be handled through the callback approach, but for the third one, the library currently doesn’t expose any event or callback that triggers during streaming updates. I explored a few possible solutions but couldn’t find a clean way to achieve this yet.  

For now, the only reliable approaches that work consistently across all three cases are either using a ref callback or a `useEffect`, even though I agree both feel a bit hacky

abdulrahmancodes: What do you think? How would you prefer we handle this?

charlesBochet: Okay so 1, 2 => let's call the scrollToBottom function exposed by the hook

For 3, let's add a ComponentEffect in AgentChatProvider that will contain a useEffect, check chatState.messages against a local messages state and if it detect a changes (deepEqual), triggers the scrollToBottom function
Let's also add a comment saying why we are force adding this useEffect here

charlesBochet (approved):
That's nice work, thanks for taking all the feedback

PR #15691, opened 2025-11-07
Do not revert isSystem update for workspaceMember

ijreilly:
Until this is done: [Make workspaceMembers non-system](https://github.com/twentyhq/core-team-issues/issues/2328) 

Let's make that update to allow us to have workspaces with workspaceMember not being system object, to allow user to customize their data model.


Weiko (approved):
👌

PR #15699, opened 2025-11-07
Fix typing error

martmull:
Fix typing error

### DIFF
--- a/packages/twenty-server/src/modules/workflow/workflow-executor/workspace-services/workflow-executor.workspace-service.ts
+++ b/packages/twenty-server/src/modules/workflow/workflow-executor/workspace-services/workflow-executor.workspace-service.ts
@@ -195,10 +195,11 @@ export class WorkflowExecutorWorkspaceService {
     const isIteratorStep = isWorkflowIteratorAction(executedStep);
 
     if (isIteratorStep) {
-      const iteratorStepResult =
-        executedStepResult.result as WorkflowIteratorResult;
+      const iteratorStepResult = executedStepResult.result as
+        | WorkflowIteratorResult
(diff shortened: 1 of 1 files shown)


packages/twenty-server/src/modules/workflow/workflow-executor/workspace-services/workflow-executor.workspace-service.ts
```diff
@@ -195,10 +195,11 @@ export class WorkflowExecutorWorkspaceService {
...
 
     if (isIteratorStep) {
-      const iteratorStepResult =
-        executedStepResult.result as WorkflowIteratorResult;
+      const iteratorStepResult = executedStepResult.result as
+        | WorkflowIteratorResult
+        | undefined;
 
-      if (!iteratorStepResult.hasProcessedAllItems) {
+      if (!iteratorStepResult?.hasProcessedAllItems) {
```
Weiko, line 202: is it expected for isIteratorStep to be true and iteratorStepResult to be undefined? Should we check or throw in that case?

Weiko (approved):
Left a comment otherwise LGTM!

packages/twenty-server/src/modules/workflow/workflow-executor/workspace-services/workflow-executor.workspace-service.ts
Weiko: @martmull

martmull: yes it can be iterator results and step type are not the same things

PR #15701, opened 2025-11-07
Fix workflow run updates real time

martmull:
as title

## after

https://github.com/user-attachments/assets/34239b6e-16d6-4afd-9a54-cb4767179bea


Devessier (approved):
Great job!

PR #15702, opened 2025-11-07
Fix horizontal bar chart spacing between ticks

bosiraphael:
This PR fixes a bug where the spacing between ticks on the Y axis was adjusted according to the width of the widget where it should have been adjusted according to the height. Same for the X axis.

**Before**:
Works for the vertical layout but not for the horizontal layout

https://github.com/user-attachments/assets/f0b94103-1ba6-4dbe-bbc6-c47541c1b5fa

**After**:
Works for both

https://github.com/user-attachments/assets/165faada-2943-4e05-92d6-a31def569500


packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetBarChart/utils/computeBarChartValueTickCount.ts
```diff
@@ -1,14 +1,21 @@
-const MIN_TICK_SPACING_HEIGHT_RATIO = 2.5;
+import { BAR_CHART_MINIMUM_WIDTH_PER_TICK } from '@/page-layout/widgets/graph/graphWidgetBarChart/constants/BarChartMinimumWidthPerTick';
+import { computeMinHeightPerTick } from '@/page-layout/widgets/graph/graphWidgetBarChart/utils/computeMinHeightPerTick';
 
 type ComputeBarChartValueTickCountProps = {
-  height: number;
+  axisSize: number;
   axisFontSize: number;
+  layout: 'vertical' | 'horizontal';
```
lucasbordeau: Can we create a re-use an existing type here ?

lucasbordeau (approved):
LGTM

packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetBarChart/utils/computeBarChartValueTickCount.ts
bosiraphael: Indeed

PR #15708, opened 2025-11-07
Remove intensive operation

martmull:
as title


charlesBochet:
@martmull let's implement a proper solution, production should be fine and it's important to keep the real time updates on workflowRuns

PR #15714, opened 2025-11-07
Improved Kanban performance

lucasbordeau:
This PR improves Kanban performance to attain the stock library speed.

There are still some performance issues at load time and drop, but they are harder to address.

Before, everything is re-rendering at every card move, full re-render at drop : 

https://github.com/user-attachments/assets/7a1cf419-c8d8-4d56-a1a3-4269dce7b5a9

After, only the columns and card outer containers get re-rendered for d&d, only two columns re-render at drop : 

https://github.com/user-attachments/assets/d2dea914-5cc3-4693-9c2d-c7c789ae3452

We can still improve performance and re-render but that would represent a global effort on D&D on general, table has the same problem currently, but since we implemented virtualization it is bearable for now.


charlesBochet:
That's better! I've just noted following performance profile
image

It seems that not visible cards take longer to render (almost 15ms)

charlesBochet (approved):

PR #15715, opened 2025-11-07
handle invalid/expired syncCursor

neo773:


packages/twenty-server/src/modules/messaging/message-import-manager/services/messaging-clear-cursors.service.ts
```diff
@@ -0,0 +1,48 @@
+import { Injectable, Logger } from '@nestjs/common';
+
+import { type WorkspaceEntityManager } from 'src/engine/twenty-orm/entity-manager/workspace-entity-manager';
+import { TwentyORMManager } from 'src/engine/twenty-orm/twenty-orm.manager';
+import { MessageChannelWorkspaceEntity } from 'src/modules/messaging/common/standard-objects/message-channel.workspace-entity';
+import { MessageFolderWorkspaceEntity } from 'src/modules/messaging/common/standard-objects/message-folder.workspace-entity';
+
+@Injectable()
+export class MessagingClearCursorsService {
```
charlesBochet, line 9: LGTM but introducing a new module just for that seems a bit overkill to me, is there a particular reason why?

charlesBochet (approved):
left a comment / question, LGTM

PR #15736, opened 2025-11-09
(breaking change) Allow users with a single workspace to update their email.

mabdullahabaid:
- Users with a single workspace are allowed to update their email across `core.user` and `workspace_xyz.workspaceMember`. 
- The latter happens asynchronously (built it like this for non-blocking with multiple workspaces), but since we restrict the email update functionality to a single user, we can also update the email in workspaceMember synchronously - I left asynchronous there to receive feedback on whether we should move to synchronous or not.
- Merged main and resolved conflicts to ensure we use the `SettingsPermissionGuard` and the updated `workspace.service.ts` code.

One edge-case that I was trying to communicate on Discord: 

Say that an admin is a member of multiple workspaces. Therefore, they can allow roles with PROFILE_INFORMATION permission to update their email.

<p align="center">
image
</p>

However, since the admin is part of multiple workspaces, he/she cannot even update own email - the field stays disabled, leading to some confusion.

<p align="center">
image
</p>

However, the workspace can have another member with admin role or some other role that has PROFILE_INFORMATION permission flag. That user will be and should be allowed to update email, so we cannot hide `email` from dropdown options.

<p align="center">
image
</p>

The behavior is fine imo, just a little confusing for members with more than one workspace.

I have also tested the flow by signing up to YC workspace with my org google account (twenty.com), then changing email to my personal address.
- After changing, I need to login using Google with my personal account to access YC workspace again.
- If I login using Google with org google account (twenty.com), a new user account is created.

This behavior is consistent with Notion and Linear. 

Finally, as for the verification of email, the user is asked to verify email while they're logged in, but just in case they logout without verifying, the next login would force them to verify their email in the email/password flow.

However, for Social/SSO, they must verify before they logout or else they'd have to contact support for assistance. I have not looked into how to show verification screen while logging in via Social/SSO yet, but if that's something critical for completeness here, I shall revisit it.


packages/twenty-front/src/modules/settings/profile/components/EmailField.tsx
```diff
@@ -1,18 +1,85 @@
...
+  const { updateEmail } = useUpdateEmail();
+
+  if (!currentUser) {
+    return null;
+  }
+
+  const [email, setEmail] = useState(currentUser.email);
+
+  useEffect(() => {
+    setEmail(currentUser.email);
```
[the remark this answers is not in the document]
mabdullahabaid: Removed the use of hook, it was redundant.

packages/twenty-server/src/database/typeorm/core/migrations/common/1762488656872-editableProfileFields.ts
```diff
@@ -0,0 +1,17 @@
+import { type MigrationInterface, type QueryRunner } from 'typeorm';
+
+export class EditableProfileFields1762488656872 implements MigrationInterface {
+  name = 'EditableProfileFields1762488656872';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "core"."workspace" ADD "editableProfileFields" text DEFAULT 'email'`,
```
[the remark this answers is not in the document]
mabdullahabaid: TypeORM parses this correctly as long as entity definition mentions simple-array as type.

packages/twenty-emails/src/emails/send-email-verification-link.email.tsx
```diff
@@ -9,22 +9,33 @@ import { type APP_LOCALES } from 'twenty-shared/translations';
...
   locale,
+  isEmailUpdate = false,
 }: SendEmailVerificationLinkEmailProps) => {
   const i18n = createI18nInstance(locale);
+  const title = isEmailUpdate
+    ? i18n._('Confirm your new email address')
+    : i18n._('Confirm your email address');
+  const bodyId = isEmailUpdate
+    ? 'We received a request to change the email address associated with your Twenty account. Click below to confirm this change.'
+    : 'Thanks for registering for an account on Twenty! Before we get started, we just need to confirm that this is you. Click below to verify your email address.';
```
FelixMalfait, line 26: I'm not sure if this will work, does it? The string need to be detect by Lingui during extract. I don't think it would here.

packages/twenty-server/src/database/typeorm/core/migrations/common/1762488656872-editableProfileFields.ts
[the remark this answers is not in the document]
FelixMalfait: Why use simple-array and not an array? I'm not familiar with this. Any downside with array which seems more common?

packages/twenty-server/src/engine/core-modules/email-verification/services/email-verification.service.ts
```diff
@@ -116,7 +130,14 @@ export class EmailVerificationService {
       );
     }
 
-    const user = await this.userService.findUserByEmailOrThrow(email);
+    // TODO: Remove the dependency on querying user altogether when the endpoint is authenticated.
```
FelixMalfait, line 134: Not sure I understand this comment sorry

packages/twenty-server/src/engine/core-modules/email-verification/services/email-verification.service.ts
FelixMalfait, line 145: I'm not sure I understand the logic, I will need to read more to understand.
But for me a user that updates his email is Verified so this would fail here.
What we do when the user asks for an email update is not update and unverify immediatly but leverage AppToken table when we store the temporary new email and only once it's confirmed then the user is verified again but with the new email. So we never go through a state where the user isn't verified?

packages/twenty-server/src/engine/core-modules/user/jobs/update-workspace-member-email.job.ts
```diff
@@ -0,0 +1,69 @@
...
+  async handle({
+    userId,
+    email,
+    workspaceIds,
+  }: UpdateWorkspaceMemberEmailJobData): Promise<void> {
+    if (!workspaceIds.length) {
+      return;
+    }
+
+    const dataSources = await this.dataSourceService.getManyDataSourceMetadata({
```
FelixMalfait: Why do you loop over workspaceMembers / data sources if it's limited to users with a single workspace? We rarely do cross-workspace operations (the only place I've seen it is in migration scripts) because it can be very dangerous and rarely make sense from a security perspective (one workspace shouldn't be able to act on another)

packages/twenty-server/src/engine/core-modules/user/services/user.service.ts
```diff
@@ -282,4 +299,90 @@ export class UserService extends TypeOrmQueryService<UserEntity> {
...
+
+    if (existingUser && existingUser.id !== user.id) {
+      throw new UserInputError('Email already in use', {
+        subCode: UserExceptionCode.EMAIL_ALREADY_IN_USE,
+        userFriendlyMessage: msg`Email already in use`,
+      });
+    }
+
+    user.email = normalizedEmail;
+    user.isEmailVerified = false;
```
FelixMalfait: Wouldn't it be better to wait for them to confirm before switching their email? Otherwise they would get locked out if they did a typo and can never access their account again?

packages/twenty-server/src/database/typeorm/core/migrations/common/1762488656872-editableProfileFields.ts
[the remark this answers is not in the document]
FelixMalfait: No we don't support multiple DB type. I feel we should go with array

packages/twenty-emails/src/emails/send-email-verification-link.email.tsx
mabdullahabaid: Searched up the strings from `send-email-verification-link.email.tsx` in the extract files `(.po extension)` and none of the strings are there. 

In fact, no text in any email template is translated even if it has `<Trans />` or `i18n` tag.

The text you pointed out is wrapped in `<Trans />`, but it's a deeper issue altogether with translations.

packages/twenty-server/src/engine/core-modules/email-verification/services/email-verification.service.ts
mabdullahabaid: Yep, I removed a circular dependency here: userService was injected in emailVerificationService and I needed to inject emailVerificationService in userService. 

Saw an existing comment in email-verification.resolver.ts file that said we had to make resendEmail resolver authenticated (probably for security). Therefore, when the resolver is authenticated, we can grab user details from the token and remove the dependency on userRepository or userService altogether. 

Left a comment as a reminder.

packages/twenty-server/src/engine/core-modules/user/services/user.service.ts
mabdullahabaid: Ahh, that's a good point. 

Both solutions exist - e.g. Twitter and Instagram do it this way, but moving from verified state to verified state could be a better idea to avoid spam email usage + users locking themselves out.

packages/twenty-server/src/engine/core-modules/user/jobs/update-workspace-member-email.job.ts
mabdullahabaid: Yeah, mentioned this in the PR description that this was over-engineered for multi-workspace update before we leaned towards single workspace update only.

Wanted feedback, will make the change.

packages/twenty-server/src/engine/core-modules/email-verification/services/email-verification.service.ts
mabdullahabaid: I set the verification flag to false to reuse the existing flow for verification, but if we want to move from verified state to verified state, things will be different, yes.

packages/twenty-server/src/engine/core-modules/email-verification/services/email-verification.service.ts
```diff
@@ -23,32 +23,42 @@ import {
...
+    locale,
+    verifyEmailRedirectPath,
+    context = EmailVerificationContext.SIGN_UP,
+  }: {
+    userId: string;
+    email: string;
+    workspace: WorkspaceDomainConfig | undefined;
+    locale: keyof typeof APP_LOCALES;
+    verifyEmailRedirectPath?: string;
+    context?: EmailVerificationContext;
```
FelixMalfait: Nit: I don't love context, I feel like it's too generic so it puts weight on the developer. Prefer something more explicit (you had isEmailUpdate which adds a second concept later on, you could have kept that. Or I understand why you would want to leave room for a third option but then it could be named something like verificationPurpose or something more generic that user can search for in the codebase and not "context")

packages/twenty-server/src/engine/core-modules/email-verification/email-verification.constants.ts
```diff
@@ -0,0 +1,4 @@
+export enum EmailVerificationContext {
+  SIGN_UP = 'sign_up',
```
FelixMalfait: I think most enum are all caps (some exceptions when they are tied to external resources)

packages/twenty-server/src/engine/core-modules/auth/auth.resolver.ts
```diff
@@ -252,19 +253,25 @@ export class AuthResolver {
         getAuthTokenFromEmailVerificationTokenInput,
       );
```
FelixMalfait, line 254: I don't really like what you're doing here. The endpoint is called `getLoginTokenFromEmailVerificationToken` but you do some mutation in there, it feels off / like an architecture issue

FelixMalfait: It was already the case before though. Maybe we should rename it then?

packages/twenty-server/src/engine/core-modules/email-verification/services/email-verification.service.ts
FelixMalfait: OK thanks!

packages/twenty-front/src/modules/settings/profile/components/EmailField.tsx
```diff
@@ -1,18 +1,77 @@
...
+        />
+        <Button
+          variant="secondary"
+          title={t`Update email`}
+          onClick={handleUpdate}
+          disabled={!canEdit || !isEmailChanged}
+        />
+      </StyledFieldRow>
+      {canEdit && (
+        <StyledHelperText>
```
FelixMalfait: We are always very careful avoiding introducing new ui patterns, that's why maybe the modal was maybe a better component if you needed to display text? 
We prefer an average design that's very consistent across the app that a good local design that makes exceptions

packages/twenty-server/src/database/typeorm/core/migrations/common/1762488656872-editableProfileFields.ts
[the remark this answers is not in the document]
FelixMalfait: Ping @mabdullahabaid why not array if it's an array?

packages/twenty-server/src/engine/core-modules/user/jobs/update-workspace-member-email.job.ts
```diff
@@ -0,0 +1,50 @@
...
+      const workspaceMemberRepository =
+        await this.twentyORMGlobalManager.getRepositoryForWorkspace<WorkspaceMemberWorkspaceEntity>(
+          workspace.id,
+          'workspaceMember',
+          { shouldBypassPermissionChecks: true },
+        );
+
+      await workspaceMemberRepository.update({ userId }, { userEmail: email });
+    } catch (error) {
+      this.logger.error(
```
FelixMalfait: We probably don't need this tryc/catch. Logger.error is mostly for debuging. Either it's a real programming error and we don't want to catch it (it will be sent to sentry). Either it's something expected and then probably shouldn't be logged

packages/twenty-server/src/engine/core-modules/user/services/user.service.ts
```diff
@@ -284,4 +301,94 @@ export class UserService extends TypeOrmQueryService<UserEntity> {
...
+    newEmail,
+    verifyEmailRedirectPath,
+  }: {
+    user: UserEntity;
+    workspace: WorkspaceEntity;
+    newEmail: string;
+    verifyEmailRedirectPath?: string;
+  }): Promise<void> {
+    const normalizedEmail = newEmail.trim().toLowerCase();
+
```
FelixMalfait, line 332: Is there any impact on billing? Maybe it's not correlated and it's best like that I don't know / didn't look

packages/twenty-front/src/modules/settings/profile/components/NameFields.tsx
```diff
@@ -21,20 +22,16 @@ const StyledComboInputContainer = styled.div`
...
-  onLastNameUpdate,
-}: NameFieldsProps) => {
+export const NameFields = ({ autoSave = true }: NameFieldsProps) => {
   const { t } = useLingui();
   const currentUser = useRecoilValue(currentUserState);
   const [currentWorkspaceMember, setCurrentWorkspaceMember] = useRecoilState(
     currentWorkspaceMemberState,
   );
+  const { canEdit: canEditFirstName } = useCanEditProfileField('firstName');
+  const { canEdit: canEditLastName } = useCanEditProfileField('lastName');
```
FelixMalfait, line 34: Should we add the profile picture also as one of the elements with firstName/lastName/email?

packages/twenty-front/src/modules/settings/profile/components/EmailField.tsx
mabdullahabaid: I feel we can remove the text and rely on the snackBar to inform user that we have sent an email as we already do.

Modal will add another pattern to profile screen imo - firstName and lastName are editable fields, changing email opens a modal for user to type an email, set password sends a URL which later opens up a modal when user visits the URL, so the screen becomes a little inconsistent.

I might be wrong, but we have no other setting opening up a modal either (just checked, most do inline edits redirect to a sub-path e.g. `settings/api-webhooks/apis/new`)

If you want to keep the text and go with the modal, I can do that, but changePassword vs changeEmail are very different imo. For password, we need to know if the user updating actually owns the underlying email account before changing the password, but for email, we only really confirm the new email after it is changed.

We can make it better to compare with Notion where you must verify the existing email before you set a new email, but it might be unnecessary friction. Therefore, I followed Linear's pattern instead.

packages/twenty-server/src/engine/core-modules/user/services/user.service.ts
mabdullahabaid: Per my analysis, Stripe still captures the user email only when a customer is first created. There is no follow-up path that updates the Stripe customer email after a user changes it, so invoices and receipts will continue to go to the original address. Because of that, the new feature doesn’t break billing flows, but it can create a mismatch between the user’s current email inside Twenty and the email stored in Stripe.

packages/twenty-front/src/modules/settings/profile/components/NameFields.tsx
mabdullahabaid: Umm if you think we should, we can. I have not used enough B2B products to know if others do the same, but no harm imo since admin can still always allow via the dropdown.

If you want to wait for user-feedback before making this decision, I am good with that too.

packages/twenty-server/src/database/typeorm/core/migrations/common/1762488656872-editableProfileFields.ts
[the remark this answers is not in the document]
mabdullahabaid: Done, replaced.

packages/twenty-server/src/engine/core-modules/email-verification/email-verification.constants.ts
mabdullahabaid: Changed to all caps.

packages/twenty-server/src/engine/core-modules/auth/auth.resolver.ts
mabdullahabaid: Changed the name to `verifyEmailAndGetLoginToken`. Also changed the name of the other mutation to `verifyEmailAndGetWorkspaceAgnosticToken`. 

Let me know if that works or if we want to name it differently.

packages/twenty-server/src/engine/core-modules/user/jobs/update-workspace-member-email.job.ts
mabdullahabaid: Done.

packages/twenty-front/src/modules/settings/profile/components/NameFields.tsx
FelixMalfait: I think we should yes! It's the same idea, if an admin wants to autoprovision this with SCIM in the future and lock it, fine, but they can also give freedom (default)

packages/twenty-server/src/engine/core-modules/user/services/user.service.ts
FelixMalfait: Yes... I guess we can leave it like that for now. Might make it harder to reconciliate, we'll see

packages/twenty-server/src/engine/core-modules/email-verification/services/email-verification.service.ts
mabdullahabaid: Renamed to `verificationTrigger` - captures which UX flow is triggering the email send.

packages/twenty-front/src/modules/settings/profile/components/NameFields.tsx
mabdullahabaid: Done.

packages/twenty-front/src/modules/settings/profile/components/EmailField.tsx
mabdullahabaid: Switched to using config-variable type edit field with IconPencil, IconCheck and IconX.

FelixMalfait (approved):

PR #15738, opened 2025-11-10
Increase QUERY_MAX_RECORDS to 10,000 and update related components to…

### LINKED ISSUE #15704: Workflow - SEARCH node returns a maximum of 200 records
## Bug Description

The search actions return a maximum of 200 records.` export const QUERY_MAX_RECORDS = 200;`
However, there is nothing to inform the user of this
-> we either need to wave this condition or inform users that this is happening. 
Ideas: hard code a limit to 200 and prevent users from increasing this number (or any better options). 
We discussed it live and @martmull  mentioned this limit could be raised to 10k. 

Other item to check while we're in this topic. I believe @thomtrp told me there is a limit in the # of records that can be selected with the Global / Bulk / Single manual trigger. But this information is not given to users which could lead to the same issues.

KhademOHAli1:
# Introduction
fixes https://github.com/twentyhq/twenty/issues/15704

## Content
This pull request increases the maximum number of records that can be processed in bulk workflow actions from 200 to 10,000 and ensures that this new limit is consistently enforced and communicated throughout the UI. It updates logic, validation, and user feedback to handle scenarios where users exceed the maximum allowed records, and improves form handling for related workflow configuration screens.

**Bulk Workflow Record Limit Increase and Enforcement**

* Increased the value of `QUERY_MAX_RECORDS` from 200 to 10,000 in `twenty-shared/constants/QueryMaxRecords.ts`, raising the cap for bulk workflow operations.
* Updated the logic in `useRunWorkflowRecordActions` to enforce the 10,000 record limit: if a user selects more than the limit, only the first 10,000 are processed and a warning snackbar is shown to inform the user. [[1]](diffhunk://#diff-e97b5d7997b4b2202af3ef61bce1bbab33c9ed42581603d85083f8fa04961992R56-R81) [[2]](diffhunk://#diff-e97b5d7997b4b2202af3ef61bce1bbab33c9ed42581603d85083f8fa04961992L71-R95) [[3]](diffhunk://#diff-e97b5d7997b4b2202af3ef61bce1bbab33c9ed42581603d85083f8fa04961992R29) [[4]](diffhunk://#diff-e97b5d7997b4b2202af3ef61bce1bbab33c9ed42581603d85083f8fa04961992R8-R17) [[5]](diffhunk://#diff-e97b5d7997b4b2202af3ef61bce1bbab33c9ed42581603d85083f8fa04961992L88-R112)

**User Interface and Validation Improvements**

* Enhanced the `WorkflowEditActionFindRecords` form to cap the limit field at 10,000, display an error if the user enters a value above the limit, and show a hint about the maximum. Also, if the action is loaded with a value above the new limit, it is automatically adjusted down. [[1]](diffhunk://#diff-5f3eec6a6ba653b2e9b14e832a6fcbce6a09185b229e0af0660a929c05267eafL80-R92) [[2]](diffhunk://#diff-5f3eec6a6ba653b2e9b14e832a6fcbce6a09185b229e0af0660a929c05267eafR156-R181) [[3]](diffhunk://#diff-5f3eec6a6ba653b2e9b14e832a6fcbce6a09185b229e0af0660a929c05267eafL280-R341) [[4]](diffhunk://#diff-5f3eec6a6ba653b2e9b14e832a6fcbce6a09185b229e0af0660a929c05267eafR27) [[5]](diffhunk://#diff-5f3eec6a6ba653b2e9b14e832a6fcbce6a09185b229e0af0660a929c05267eafR70)
* Updated the manual trigger configuration UI to clarify that bulk triggers will only pass up to 10,000 records to the workflow, ensuring users are aware of the new cap. [[1]](diffhunk://#diff-76fd74936daddd3a4c6e96c00e1a7e177072dad014d8453e1f9c90810a8a7679R20) [[2]](diffhunk://#diff-76fd74936daddd3a4c6e96c00e1a7e177072dad014d8453e1f9c90810a8a7679R65) [[3]](diffhunk://#diff-76fd74936daddd3a4c6e96c00e1a7e177072dad014d8453e1f9c90810a8a7679L85-R87)

### DIFF
--- a/packages/twenty-shared/src/constants/QueryMaxRecords.ts
+++ b/packages/twenty-shared/src/constants/QueryMaxRecords.ts
@@ -1,1 +1,1 @@
-export const QUERY_MAX_RECORDS = 200;
+export const QUERY_MAX_RECORDS = 10_000;
--- a/packages/twenty-front/src/modules/action-menu/actions/record-actions/run-workflow-actions/hooks/useRunWorkflowRecordActions.tsx
+++ b/packages/twenty-front/src/modules/action-menu/actions/record-actions/run-workflow-actions/hooks/useRunWorkflowRecordActions.tsx
@@ -5,6 +5,7 @@ import { ActionType } from '@/action-menu/actions/types/ActionType';
 import { contextStoreTargetedRecordsRuleComponentState } from '@/context-store/states/contextStoreTargetedRecordsRuleComponentState';
 import { type ObjectMetadataItem } from '@/object-metadata/types/ObjectMetadataItem';
 import { recordStoreFamilyState } from '@/object-record/record-store/states/recordStoreFamilyState';
+import { useSnackBar } from '@/ui/feedback/snack-bar-manager/hooks/useSnackBar';
 import { useRecoilComponentValue } from '@/ui/utilities/state/component-state/hooks/useRecoilComponentValue';
 import { useActiveWorkflowVersionsWithManualTrigger } from '@/workflow/hooks/useActiveWorkflowVersionsWithManualTrigger';
 import { useRunWorkflowVersion } from '@/workflow/hooks/useRunWorkflowVersion';
@@ -12,6 +13,8 @@ import { useRunWorkflowVersion } from '@/workflow/hooks/useRunWorkflowVersion';
 import { type WorkflowVersion } from '@/workflow/types/Workflow';
 import { COMMAND_MENU_DEFAULT_ICON } from '@/workflow/workflow-trigger/constants/CommandMenuDefaultIcon';
 import { useRecoilCallback } from 'recoil';
+import { t } from '@lingui/core/macro';
+import { QUERY_MAX_RECORDS } from 'twenty-shared/constants';
 import { capitalize, isDefined } from 'twenty-shared/utils';
 import { useIcons } from 'twenty-ui/display';
 
@@ -23,6 +26,7 @@ export const useRunWorkflowRecordActions = ({
   skip?: boolean;
 }) => {
   const { getIcon } = useIcons();
+  const { enqueueWarningSnackBar } = useSnackBar();
   const contextStoreTargetedRecordsRule = useRecoilComponentValue(
     contextStoreTargetedRecordsRuleComponentState,
   );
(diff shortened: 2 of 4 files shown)


prastoin (changes requested):
Hello there, thanks for you contribution. Please feel free to share your use case
The most self hosted oriented solution would be IMO making `QUERY_MAX_RECORDS` an env variable in order to avoid selfhosters having to fork and build their custom twenty image. As I assume you're a selfhoster that needs to edit this limit.

This PR modify the default value of `QUERY_MAX_RECORDS` which is way to invasive for your own use case

KhademOHAli1:
> their

hi thank you for the feedback. Goal was to address this issue https://github.com/twentyhq/twenty/issues/15704

prastoin:
Ok thanks for linking the issue reopening this, I still don't think that's a good idea to set `10_000` limit IMO but rather make a batch loop in the frontend if the selected count is more than the limit 🤔

packages/twenty-shared/src/constants/QueryMaxRecords.ts
```diff
@@ -1 +1 @@
-export const QUERY_MAX_RECORDS = 200;
+export const QUERY_MAX_RECORDS = 10_000;
```
charlesBochet, line 1: I'm against this change:
- regarding bulk edition, it's not reasonable to trigger a workflow on that many record, let's keep the limit at 200 and surface that we can't do more.
- regarding Search records (which also covered in the original issue), I think search record node should keep returning 200 records. If the user needs more he can either: make the search node part of a loop and use offseting, or create a custom code function, this will be more efficient. Making actions on 10k records is not realistic right now (and not sure it will be one day)

charlesBochet:
Hi @KhademOHAli1, since we haven't heard back from you, I wanted to take over this PR. Unfortunately I cannot help you as you have opened it from your repository main branch and your main branch is locked.

I'll open another PR to take over,
Thanks a lot for opening this one

KhademOHAli1:
Hi, No worries. Wanted to help out. Honestly I am confused what happened here. If there are parts you can use feel free, if not than it's resolved on my end.

charlesBochet:
ty :) The issue is that you opened a PR from [KhademOHAli1:main](https://github.com/KhademOHAli1/twenty/tree/main), and your main branch is protected so I cannot help (that's github default setting)

Usually if you create another branch and open the PR from there, I can also contribute (no github default setting to protect it)

PR #15740, opened 2025-11-10
Fix concurrent page layout edition bug

bosiraphael:
This PR fixes a bug which could happen if two people were editing a page layout at the same time.
If one person deletes a widget or a tab and saves first, and the second person saves after but doesn't delete this widget or tab, an error is raised.

This is because, in the backend, a diff is computed to know which tab or widget to create, update or delete. But the logic was maid without considering soft deletion. So, when the second person saves, the diff tries to create the widget or tab which had been soft deleted by the first use. Since they have the same id, a duplicate primary key error is raised.

Now we always consider that the last user to save has the truth. So we restore the tab/widget and update it.


packages/twenty-server/src/engine/core-modules/page-layout/services/page-layout-tab.service.ts
```diff
@@ -39,17 +39,19 @@ export class PageLayoutTabService {
...
+    withDeleted = false,
   ): Promise<PageLayoutTabEntity[]> {
     const repository = this.getPageLayoutTabRepository(transactionManager);
 
     return repository.find({
       where: {
         pageLayoutId,
         pageLayout: { workspaceId },
-        deletedAt: IsNull(),
+        ...(withDeleted ? {} : { deletedAt: IsNull() }),
```
Weiko: I think both lines are redundant, prefer withDeleted: true or am I missing something?

packages/twenty-server/src/engine/core-modules/page-layout/services/page-layout-widget.service.ts
```diff
@@ -46,16 +46,18 @@ export class PageLayoutWidgetService {
...
 
     return repository.find({
       where: {
         pageLayoutTabId,
         workspaceId,
-        deletedAt: IsNull(),
+        ...(withDeleted ? {} : { deletedAt: IsNull() }),
       },
       order: { createdAt: 'ASC' },
+      ...(withDeleted ? { withDeleted: true } : {}),
```
Weiko: same

Weiko (approved):
LGTM

PR #15741, opened 2025-11-10
Move composite types to twenty-shared

martmull:
Type postcard fullName and address fields properly (before typed to string)
- Moves composites to twenty-shared
- Exports composites types from twenty-sdk
- Update hello-world


FelixMalfait (approved):

martmull:
@FelixMalfait i will rename exported types in a dedicated PR

PR #15625, opened 2025-11-05
Twenty standard and workspace custom applications 1/3

prastoin:
# Introduction
related to https://github.com/twentyhq/core-team-issues/issues/1833
In this PR we're starting the sync-metadata and standardIds deprecation by introducing `twenty-standard` application that will regroup every standard object such as company and opportunities. But also the `custom-workspace-application` which is an app created at the same time as a workspace and that will regroup everything configure within the workspace ( custom objects fields etc )

## What's done
On both new workspace and seeded workspace creation:
- Creating a custom workspace app
- Creating a twenty standard app
- Refactored the seed core schema and workspace creation to be run within a transaction in order to handle circular dependency foreignkey requirements ( which is deferred for app toward workspace )
- Updated workspace entity to have a custom workspace relation ( nullable for the moment until we implem an upgrade command to handle retro comp )
- Integration testing on user, workspace creation deletion and expected default apps creation
- ~~Soft deleted user on `deleteUser`~~ Done by marie and rebased on it

## What's next
- Update seeder to propagate the `twenty-standard` workspace `applicationId` to every standard synchronized entities ( cheap and fast iteration through the about to be deprecated sync-metadata as an easy way to synchronize standards metadata entities ).
- Update seeder to propagate the `custom-workspace-application` workspace `applicationId` to anything custom ( `pets` and `rockets` )
- Prepend `custom-workspace-application` `applicationId` to every metadata API operations ( create a specific cache etc )
- Upgrade command on all existing workspace to create a custom app and associate its applicationId to any existing custom entities
- Make `universalIdentifier` and `applicationId` required for any syncable entity

PR #15745, opened 2025-11-10
Twenty standard and workspace custom applications 2/3

prastoin:
# Introduction
In this PR we're migration the syncable entity to have required `applicationId` and `universalIdenfitier`.
It has several impacts:
- Refactor the sync metadata to be passing the twenty standard application id of the current workspace
- Upgrade commands in order to migrate existing workspace's entity for the migration to be run successfully

## API Metadata
Any operation going through the api metadata nows automatically scope the related entity to the workspace custom application instance. This breaks the existing "hack" implementation of the synchronize application service that calls all the services atomically.

We need to either ignore the tests or remove the cli status check from the blocking status badges for a PR to be merged

## New workspace
Already handled in previous https://github.com/twentyhq/twenty/pull/15625, when a workspace is created it gets created a twenty standard and custom workspace instance

## Existing workspace
- Fails the migration ( swallowed )
- Passes the upgrade commands that: creates standard app, associate standard entities to it and set their universalIdentifier to their standardId ( not all though ). Creates a custom workspace applications and associate all non app associated yet entities to the workspace custom app

## FlatApplication
Introduced the flatApplication and cache. It's a bit hacky right now but will migrate existing `MetadataName` to be `SyncableMetadataName` in a following PR

## V1 Deprecation
Official v1 deprecation as it hasn't been refactored to handle new required appid and universalId
Will remove the code after this PR

### Sync-metadata governance
The sync metadata has been modified to still allow workspace creation after we apply the required applicationId and universalIdentifier migration
It handles the synchronization of those fields for the following entities:
- Objects
- Fields
- Roles
- Agents
- 
- Indexes (**only create and delete**, update will be handled by an upgrade command as it's not supported by the sync metadata )


prastoin:
I need to remove v1 code source before this

prastoin:
Blocked by https://github.com/twentyhq/twenty/pull/15823

prastoin:
I need to refactor migrations command to be handling view universal id and view universal id but also handle the field metadata are some standardId seems ot be invalid by definition but has never been consumed before

prastoin:
I don't like what's happening here
Even though it works at some point it's not the good strategy in the end
We should not hack in the sync metadata at all
If it's not only being thankful for his loyal services so far
I will now tend to describe a json twenty standard configuration as const server side and generate and maintain a workspace migration from it and programmatically run them on a workspace creation

prastoin:
It's not only lost but will restart from scratch

PR #15751, opened 2025-11-10
fix: Replace angle bracket placeholders with curly braces in docs to fix crowdin's tags mismatch errors

abdulrahmancodes:


FelixMalfait (approved):
(not merging since workflow files need to be reverted but feel free to merge)

Also could we add a lint rule or something to prevent this from happening again?

FelixMalfait (approved):

PR #15756, opened 2025-11-11
fix(docs): Use content instead of props for Card titles to enable translation

abdulrahmancodes:


FelixMalfait (commented):
This feels very hacky. I think we can do without a bash script and just reference the key in the component for example? There must be a better way. Bash should only be a solution of last-resort

FelixMalfait (approved):

PR #15757, opened 2025-11-11
message channel change #3

neo773:


packages/twenty-server/src/engine/core-modules/auth/services/create-message-channel.service.ts
```diff
@@ -75,8 +75,10 @@ export class CreateMessageChannelService {
...
         workspaceId,
-        messageChannelId: newMessageChannel.id,
-        connectedAccount,
+        messageChannel: {
+          ...newMessageChannel,
+          connectedAccount,
+        },
         manager,
       });
     }
```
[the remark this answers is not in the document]
neo773: Bot seems to be hallucinating, the fix it suggested already exists

packages/twenty-server/src/modules/messaging/message-import-manager/crons/jobs/messaging-process-folder-actions.cron.job.ts
```diff
@@ -0,0 +1,76 @@
...
+import { MessageQueueService } from 'src/engine/core-modules/message-queue/services/message-queue.service';
+import { WorkspaceEntity } from 'src/engine/core-modules/workspace/workspace.entity';
+import { getWorkspaceSchemaName } from 'src/engine/workspace-datasource/utils/get-workspace-schema-name.util';
+import { MessageFolderPendingSyncAction } from 'src/modules/messaging/common/standard-objects/message-folder.workspace-entity';
+import {
+  MessagingProcessFolderActionsJob,
+  type MessagingProcessFolderActionsJobData,
+} from 'src/modules/messaging/message-import-manager/jobs/messaging-process-folder-actions.job';
+
+export const MESSAGING_PROCESS_FOLDER_ACTIONS_CRON_PATTERN = '*/15 * * * *';
```
charlesBochet, line 21: shouldn't we trigger this automatically on listener + integrate it to the message-list-fetch process?

charlesBochet: I feel we don't need to add another cron

packages/twenty-server/src/modules/messaging/message-import-manager/crons/jobs/messaging-process-group-email-actions.cron.job.ts
```diff
@@ -0,0 +1,73 @@
...
+import { MessageChannelPendingGroupEmailsAction } from 'src/modules/messaging/common/standard-objects/message-channel.workspace-entity';
+import {
+  MessagingProcessGroupEmailActionsJob,
+  type MessagingProcessGroupEmailActionsJobData,
+} from 'src/modules/messaging/message-import-manager/jobs/messaging-process-group-email-actions.job';
+
+export const MESSAGING_PROCESS_GROUP_EMAIL_ACTIONS_CRON_PATTERN = '0 */2 * * *';
+
+@Processor(MessageQueue.cronQueue)
+export class MessagingProcessGroupEmailActionsCronJob {
```
charlesBochet, line 24: reading this action work, I wonder if we should actually push the concept further to unify:
FOLDER_ACTION: CREATE / SYNC / DELETE and leverage everything as part as messaging list fetch. 
Message list fetch would:
- perform deletions
- then creation / sync

Just thinking, doesn't mean we can't try what you have done

packages/twenty-server/src/modules/messaging/message-import-manager/services/messaging-process-group-email-actions.service.ts
```diff
@@ -0,0 +1,157 @@
...
+    pendingGroupEmailsAction: MessageChannelPendingGroupEmailsAction,
+  ): Promise<void> {
+    const messageChannelRepository =
+      await this.twentyORMManager.getRepository<MessageChannelWorkspaceEntity>(
+        'messageChannel',
+      );
+
+    await messageChannelRepository.update(
+      { id: messageChannel.id },
+      { pendingGroupEmailsAction },
```
charlesBochet, line 40: should we derive this from messageFolders instead?

charlesBochet (commented):
Left comments but I'm eager to try as it is

charlesBochet:
Overall code looks very good, thanks for batching everything!

charlesBochet:
Functional review:
- setting up a channel on main and checking out your PR should not break existing behavior: 
- setting up a channel on your PR without folder feature flag should not break behavior
- setting up a channel on your PR with feature flag should also work (it's fine if degraded behavior for now)

charlesBochet:
Functional note:
- biggest thing is I think that folder modification should trigger an immediate update => I would generalize the concept of folder action / messageChannel action and make the message list fetch handle it. Then when we change something (we also need to add a save button), trigger the message list fetch process that should be idem potent
- add a save button on messageFolder edit (and also calendarChannel to be consistent)
- we should block messageChannel update as long as there is an ongoing message channel sync ongoing

charlesBochet (approved):
Merging as it does not break existing behavior, let's keep iterating

PR #15762, opened 2025-11-11
fix: use value from filter if recordUUID is unavailable

BKM14:
Closes #15067

When filtering by a record's `id` using the "Is" operator in a workflow's `Search Records` action, the filter was ignored. The action returned all records for that object type instead of the single record matching the `id`.

## Root cause
Inside `computeRecordGqlOperationFilter.ts`, all filters are iterated over and passed on to `turnRecordFilterIntoRecordGqlOperationFilter`. In `turnRecordFilterIntoRecordGqlOperationFilter`, in the case of `uuid`, `recordIdsForUuid` is required. But since it is undefined, it is early returned, causing the filter to be dropped.

```javascript
case 'UUID': {
      const recordIds = recordIdsForUuid;
      if (!isDefined(recordIds) || recordIds.length === 0) return;
```

## Solution
In the case where `recordIdsForUuid`, there is an additional check:
```javascript
const uuidValue =
        typeof recordFilter.value === 'string' ? recordFilter.value : undefined;
if (isDefined(uuidValue) && uuidValue.trim() !== '')
```

If `uuidValue` is valid, then `eq` is used with `uuidValue` and returned.

Another potential solution could be to pass `recordIdsForUuid` as a singular array for each iteration in `computeRecordGqlOperationFilter`.
```javascript
recordFilters
      .filter((filter) => !isDefined(filter.recordFilterGroupId))
      .map((regularFilter) => {
        return turnRecordFilterIntoRecordGqlOperationFilter({
          recordFilter: regularFilter,
          fieldMetadataItems: fields,
          filterValueDependencies,
          recordIdsForUuid: [regularFilter.value]
        });
```

## Screenshot
image
image

## Note
- New workflows: Will correctly filter by UUID when using the "Is" operator
- Existing workflows: Workflows saved with the bug will need to be re-saved to regenerate the `gqlOperationFilter`

### DIFF
--- a/packages/twenty-shared/src/utils/filter/turnRecordFilterIntoGqlOperationFilter.ts
+++ b/packages/twenty-shared/src/utils/filter/turnRecordFilterIntoGqlOperationFilter.ts
@@ -45,9 +45,9 @@ import { z } from 'zod';
 
 import { type DateTimeFilter } from '@/types/RecordGqlOperationFilter';
 import {
+  CustomError,
   checkIfShouldComputeEmptinessFilter,
   checkIfShouldSkipFiltering,
-  CustomError,
   getFilterTypeFromFieldType,
   getPlainDateFromDate,
   isDefined,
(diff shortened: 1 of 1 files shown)


lucasbordeau:
Duplicate of https://github.com/twentyhq/twenty/pull/15817

PR #15764, opened 2025-11-12
fix: hotkey visibility in buttons with blue accent

BKM14:
Closes #15728

## Problem
The `color` of `ButtonHotKeys` was set to `theme.border.color.blue` when the button accent was blue. This led to poor visibility of the hotkey.

## Solution
The `color` is now set to `GRAY_SCALE_LIGHT.gray7`

### Before
image
image

### After
image
image

### DIFF
--- a/packages/twenty-ui/src/input/button/components/Button/internal/ButtonHotKeys.tsx
+++ b/packages/twenty-ui/src/input/button/components/Button/internal/ButtonHotKeys.tsx
@@ -14,7 +14,7 @@ const StyledSeparator = styled.div<{
   background: ${({ theme, accent }) => {
     switch (accent) {
       case 'blue':
-        return theme.border.color.blue;
+        return GRAY_SCALE_LIGHT.gray7;
       case 'danger':
         return theme.border.color.danger;
       default:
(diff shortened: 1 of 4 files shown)


lucasbordeau:
@Bonapara I think that we should keep the `theme` here but maybe the color is not right since color refactor ?

Bonapara:
> @Bonapara I think that we should keep the `theme` here but maybe the color is not right since color refactor ?

sure we should use the accent color and not gray.

lucasbordeau:
@Bonapara 

The accent color for dark theme gives this : 

image

Which is this color on the palette : 

Capture d’écran 2025-11-18 à 09 57 19

Is it ok with you ?

```
export const ACCENT_DARK = {
  primary: COLOR_DARK.blue5,
  secondary: COLOR_DARK.blue5,
  tertiary: COLOR_DARK.blue3,
  quaternary: COLOR_DARK.blue2,
  accent3570: COLOR_DARK.blue8,
  accent4060: COLOR_DARK.blue8,
};
```

```
blue5: RadixColors.indigoDarkP3.indigo5,
```

Bonapara:
It would be better in white indeed, you're right @lucasbordeau thanks for checking! 

Let me call you

lucasbordeau:
New version : 

Main : 
image

image

This PR : 

image

image

lucasbordeau (approved):

PR #15765, opened 2025-11-12
Twenty-shared tests parses decorator

prastoin:
## Introduction
Since we've moved some class validator instances from twenty-server to twenty-shared tests are red because they do not know how to parse decorators declarations

We've fixed this by explicitly installing `class-validator` in `twenty-shared` and configuring jest swc accordingly 

## Twenty-server class validator patch
I don't even know if that's something we need anymore
Seems to be a patch either fixing or introducing credit card and phone number validation
Would prefer discussing the need or not to either before merging this as it could introduce regression at runtime:
- Centralize the patch to be consumed in both `twenty-server` and `twenty-shared`
- Remove the patch

We should also document every patch motivations we do as it's quite though to iterate over a such huge one

### DIFF
--- a/packages/twenty-server/package.json
+++ b/packages/twenty-server/package.json
@@ -86,7 +86,7 @@
     "cache-manager-redis-yet": "^4.1.2",
     "chalk": "4.1.2",
     "class-transformer": "0.5.1",
-    "class-validator": "patch:class-validator@0.14.0#./patches/class-validator+0.14.0.patch",
+    "class-validator": "0.14.0",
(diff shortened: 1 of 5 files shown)


martmull (approved):

prastoin:
@greptileai could you please read the PR description again and let me know your though about the class validator patch applied in twenty-server ?

Weiko (commented):
As discussed, let's remove the patch, this is probably not used anymore and if catch something the fix should be to use a custom validator and not patch class-validator

PR #15766, opened 2025-11-12
Rename sdk utils

martmull:
As title


packages/twenty-apps/hello-world/src/objects/postCard.ts
```diff
@@ -14,7 +14,7 @@ enum PostCardStatus {
   RETURNED = 'RETURNED',
 }
 
-@ObjectMetadata({
+@Object({
```
FelixMalfait, line 17: I thought this wasn't possible? Nice!

FelixMalfait (approved):

packages/twenty-apps/hello-world/src/objects/postCard.ts
martmull: There is an alias in the import above

martmull: I can try remove the alias

Issue #15746, opened 2025-11-10
Workflow: Filter by ID does not work for search record


rdelassus:
## Bug Description

When given a filter "ID is [CompanyID]", the returned company does not have [CompanyID] as ID.

**Example:** 

<img width="978" height="1051" alt="Image" src="https://github.com/user-attachments/assets/fd436f61-470a-4f6d-9b09-a2fdaeec84b3" />

<img width="959" height="756" alt="Image" src="https://github.com/user-attachments/assets/3d723bf2-0889-494d-a7eb-3a69b265b668" />

<img width="503" height="937" alt="Image" src="https://github.com/user-attachments/assets/fec8872c-783c-4ec3-9c37-d3041eac95c4" />

The output ID is not the input companyID at all, despite the search configuration.

## Expected behavior

When filtered by "ID is [CompanyID", result ID should be [CompanyID]

## Technical inputs

self hosted

BOHEUS:
Duplicate of #15067

PR #15772, opened 2025-11-12
Fix workflow “Search Records” with UUID filter

rdelassus:
## Summary
- add a regression test that reproduces the workflow “Search Records → ID is <UUID>” failure (issue #15067 / #15746)
- adjust `turnRecordFilterIntoRecordGqlOperationFilter` so UUID filters fall back to the literal value when no `recordIdsForUuid` context is provided, always emitting an `in` clause

## Testing
- npx nx test twenty-shared -- --testPathPattern=computeRecordGqlOperationFilter.test.ts --coverage=false


charlesBochet:
@rdelassus twentyhq:fix-twenty-shared-test PR has been merged which resulted in closing your PR.
Could you open a new one pointing to (based on) our main branch?

Thanks?

PR #15774, opened 2025-11-12
Fix Snackbar crashing the whole app

### LINKED ISSUE #1869: Verify that all GraphQLErrors are correctly translated before being sent back to the frontend
We recently add a bug that involved Snackbar trying to render a MessageDescriptor object : https://github.com/twentyhq/twenty/issues/15685

This means that somehow some errors don't get translated in `useGraphQLErrorHandlerHook` before being sent to the front end.

Though it could also be in an endpoint, REST or GQL, that processes tokens and login, because the error appears when a token is invalid somehow (hard to know precisely what causes this bug).

For the moment a temporary fix has been implemented in the frontend in `getErrorMessageFromApolloError`.

In the frontend, we previously forced the casting of `userFriendlyMessage` to string, but apparently it can receive `MessageDescriptor` as well so for now the frontend will handle both to avoid crashing the app.

### LINKED ISSUE #15685: Error Object are not valid as React child while loading app
Image

lucasbordeau:
This PR fixes a crash of the app when a Snackbar tries to render an object.

image

What happened : the backend sent a MessageDescriptor in `extensions.userFriendlyMessage` while it should have sent a translated string.

But it is a problem that the Snackbar can end up rendering objects and crashing the whole app because it is a the highest level in the tree.

So this PR hardens many points to avoid future bugs : 
- Sanitize what is rendered by the Snackbar to avoid any crash
- Translate any MessageDescriptor object that could end up in the frontend
- Fix the places in the backend where a MessageDescriptor wasn't translated and sent directly to the frontend in `use-graphql-error-handler.hook.ts`

Fixes https://github.com/twentyhq/twenty/issues/15685
Fixes https://github.com/twentyhq/core-team-issues/issues/1869

### DIFF
--- a/packages/twenty-front/src/modules/ui/feedback/snack-bar-manager/components/SnackBar.tsx
+++ b/packages/twenty-front/src/modules/ui/feedback/snack-bar-manager/components/SnackBar.tsx
@@ -1,3 +1,4 @@
+import { sanitizeMessageToRenderInSnackbar } from '@/ui/feedback/snack-bar-manager/utils/sanitizeMessageToRenderInSnackbar';
 import { useTheme } from '@emotion/react';
 import styled from '@emotion/styled';
 import { useLingui } from '@lingui/react/macro';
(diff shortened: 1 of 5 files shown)


bosiraphael (approved):
LGTM

PR #15775, opened 2025-11-12
fix(ui): layout-shift-on-tasks

### LINKED ISSUE #12090: Previous rows are slighted move up and down when moving on focus on a grouped record table
## Bug Description

https://github.com/user-attachments/assets/ef8ad101-3bca-422c-9dba-606c35d6ccf9

## Expected behavior

I expect the layout to be stable and not to move when the focus changes.

codiearyan:
Fixes #12090 

### Summary
Fixes a UI layout shift issue in table rows in Tasks page, where bottom borders appeared inconsistently across rows.
The problem occurred because `shouldDisplayBorderBottom` conditionally applied the bottom border, which caused slight height differences between rows and visual "jumping" when rendering focused styling. 

### Changes Made
- Ensured consistent `border-bottom` rendering
- Removed conditional rendering prop to always display bottom border

### Steps to Reproduce (Before Fix)
1. Navigate to - /objects/tasks?viewId=xxxx
2. Interact with row elements
3. Notice a slight layout shift when first row is active/focused

### Before 
https://github.com/user-attachments/assets/ab95db5d-0922-4460-a086-a03f58375824

### After

https://github.com/user-attachments/assets/a6b4d50c-b6f0-456b-90d4-fc3f0cde42bc


lucasbordeau (approved):
Ok LGTM

PR #15778, opened 2025-11-12
Add profile to twenty-cli authentication

### LINKED ISSUE #1827: [Extensibility] Twenty cli, provide multiple env authentication
as title. Developer voice:

> - twenty-cli would benefit multi-env authentication (local + remote), switching often was little irritating

See https://github.com/twentyhq/twenty/pull/15527#issuecomment-3480985275

martmull:
As title
Eg: 
image

Not breaking


FelixMalfait:
I feel like profile isn't a known concept. Why not Workspace? Should it be something with the workspace domain?

martmull:
yes indeed, we can replace "profile" with "workspace"

martmull:
@FelixMalfait done

packages/twenty-cli/src/services/config.service.ts
```diff
@@ -3,49 +3,121 @@ import * as os from 'os';
...
 export class ConfigService {
   private readonly configPath: string;
+  private static activeWorkspace = DEFAULT_WORKSPACE_NAME;
 
   constructor() {
     this.configPath = path.join(os.homedir(), '.twenty', 'config.json');
   }
 
+  static setActiveWorkspace(name?: string) {
+    this.activeWorkspace = name ?? DEFAULT_WORKSPACE_NAME;
```
FelixMalfait, line 21: Can we get the workspaceName from the API key? Workspace Name being the workspace.subdomain probably?

martmull: Not sure we should do that because as we use multiple environnement we can have duplicated subdomains

FelixMalfait: I feel like we could do a more graceful default behavior at least. But if you're not convinced we can skip!

martmull: yes, current behavior solves the original issue, is simple and not too magic. I would give that a try and improve if necessary

FelixMalfait (approved):

PR #15807, opened 2025-11-13
Improve infer deletion from entities

Weiko:
## Context
inferDeletionFromEntities only accepts a set of keys for each entities that needs deletion. This could be error-prone if tmr we want to add a new side effect and forget to add the entity when in practice you want to delete all entities that are in the fromToAllFlatEntityMaps (this is the case for applications for example)

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/application/application-sync.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/application/application-sync.service.ts
@@ -2,7 +2,6 @@ import { Injectable, Logger } from '@nestjs/common';
 
 import { parse } from 'path';
 
-import { ALL_METADATA_NAME, AllMetadataName } from 'twenty-shared/metadata';
 import { isDefined } from 'twenty-shared/utils';
 
 import { ApplicationEntity } from 'src/engine/core-modules/application/application.entity';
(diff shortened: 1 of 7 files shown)


packages/twenty-server/src/engine/core-modules/application/application-sync.service.ts
prastoin, line 998: ***Nitpick:*** Could use `computeFromToFlatEntityMaps`

packages/twenty-server/src/engine/core-modules/application/application-sync.service.ts
```diff
@@ -1012,15 +1011,7 @@ export class ApplicationSyncService {
...
-          inferDeletionFromMissingEntities: {
-            ...Object.values(ALL_METADATA_NAME).reduce(
-              (acc, metadataName) => ({
-                ...acc,
-                [metadataName]: true,
-              }),
-              {} as Partial<Record<AllMetadataName, boolean>>,
-            ),
-          },
+          inferDeletionFromMissingEntities: true,
```
prastoin, line 1014: Cool

packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-builder-v2/types/workspace-migration-builder-options.type.ts
```diff
@@ -1,6 +1,8 @@
 import { type AllMetadataName } from 'twenty-shared/metadata';
 
 export type WorkspaceMigrationBuilderOptions = {
-  inferDeletionFromMissingEntities?: Partial<Record<AllMetadataName, boolean>>;
+  inferDeletionFromMissingEntities?:
+    | true
+    | Partial<Record<AllMetadataName, boolean>>;
```
prastoin, line 6: ***Nitpick:*** boolean could be true

prastoin (approved):
Neat !

PR #15808, opened 2025-11-13
Fixed hovering of kanban cards

lucasbordeau:
This PR fixes the record inline cells that disappeared on board card due to recent refactors.


packages/twenty-front/src/modules/object-record/record-inline-cell/components/RecordInlineCellDisplayMode.tsx
```diff
@@ -102,10 +100,12 @@ export const RecordInlineCellDisplayMode = ({
...
       >
         <StyledRecordInlineCellNormalModeInnerContainer>
-          {shouldShowEmptyPlaceholder ? (
+          {shouldShowValue ? (
+            children
+          ) : shouldShowEmptyPlaceholder ? (
             <StyledEmptyField>{emptyPlaceHolder}</StyledEmptyField>
           ) : (
-            children
+            <></>
```
bosiraphael: nit: return null instead of empty fragment

bosiraphael (approved):
LGTM

PR #15810, opened 2025-11-13
Fix authContext missing in ALS WorkspaceContext for global datasource

### LINKED ISSUE #15809: Workspace member is always undefined in timeline activity
## Bug Description

Workspace member is always undefined in the timeline activity and it falls back to the workspace name.

Image

## Expected behavior

Should be defined

Weiko:
Fixes https://github.com/twentyhq/twenty/issues/15809

## Context
We recently introduced a global datasource now consuming workspace context from ALS store however the authContext was missing (only the workspaceId was there) which "broke" event emission, now missing the workspaceMemberId

This PR adds the missing authContext so we can access from anywhere in the datasource. We are still passing it as a parameters on repository level for legacy but in theory we should be able to remove it from everywhere and consume the context

Screenshot 2025-11-13 at 18 58 16


bosiraphael (approved):
LGTM

PR #15815, opened 2025-11-14
seperate v1 and v2 seeds, make groupMode to add default in case its not provided through api but primary groupBy is present

ehconitin:
Context - 
- refactoring v1 and v2 seeds to test v2 dashboard flag both on and off
Right now, the db reset command fails if the feature flag is off

- whenever there is a groupBy field -- we should add a default groupMode - we already do that on the front -- but implementing the same on server so that api users get the same response

This is also the reason why the second seed on dashboards lags a lot -- because  on that particular dashboard's widget -- we set groupBy but no groupMode (in seeding util) -- and the effect of limiting the bars to fifty runs depending on the groupMode


packages/twenty-server/src/engine/core-modules/page-layout/utils/validate-and-transform-widget-configuration.util.ts
```diff
@@ -58,6 +60,13 @@ const validateGraphConfiguration = ({
         throw errors;
       }
 
+      if (
+        instance.secondaryAxisGroupByFieldMetadataId &&
```
lucasbordeau: `isDefined(instance.secondaryAxisGroupByFieldMetadataId)` ?

packages/twenty-server/src/engine/core-modules/page-layout/utils/validate-and-transform-widget-configuration.util.ts
```diff
@@ -75,6 +84,13 @@ const validateGraphConfiguration = ({
...
```
lucasbordeau: Same ?

lucasbordeau (approved):

PR #15820, opened 2025-11-14
Dual-write for users and userWorkspaces.

mabdullahabaid:
Summary:
- Replicated all the fields from users to userWorkspaces entity.
- Introduced some helper functions to map the fields from userEntity that we want to write to userWorkspaceEntity.
- Mapping all fields in every write so that if a user updates password, but nothing else, the entire user entity is mapped to user workspace table along with the updated password.


packages/twenty-front/src/generated-metadata/graphql.ts
```diff
@@ -4553,12 +4553,20 @@ export type UserMappingOptionsUser = {
 
 export type UserWorkspace = {
   __typename?: 'UserWorkspace';
+  canAccessFullAdminPanel: Scalars['Boolean'];
+  canImpersonate: Scalars['Boolean'];
   createdAt: Scalars['DateTime'];
   deletedAt?: Maybe<Scalars['DateTime']>;
+  disabled?: Maybe<Scalars['Boolean']>;
```
FelixMalfait, line 4560: I don't thin disabled is used, let's get rid of it!

charlesBochet:
As discussed with Felix, it seems to early to change this behavior, let's close this one :)

PR #15826, opened 2025-11-14
Introduce primary actions

bosiraphael:
- These actions are displayed in the accent color
- Added new context store state `contextStoreIsPageInEditModeComponentState`
- Reverted the order in which the actions are displayed (the first action should appear to the right of the top bar action menu)

https://github.com/user-attachments/assets/3a0769cf-61ed-4619-8c4d-e3a01765b63c


packages/twenty-front/src/modules/action-menu/components/PageHeaderActionMenuButtons.tsx
```diff
@@ -15,7 +15,7 @@ export const PageHeaderActionMenuButtons = () => {
   const { actions } = useContext(ActionMenuContext);
   const theme = useTheme();
 
-  const pinnedActions = actions.filter((entry) => entry.isPinned);
+  const pinnedActions = actions.filter((entry) => entry.isPinned).reverse();
```
lucasbordeau: toReversed() ? Otherwise it's in place, maybe not what you want

packages/twenty-front/src/modules/app/effect-components/PageChangeEffect.tsx
```diff
@@ -107,10 +121,16 @@ export const PageChangeEffect = () => {
     if (!previousLocation || previousLocation !== location.pathname) {
       setPreviousLocation(location.pathname);
       executeTasksOnAnyLocationChange();
+      resetIsPageInEditMode();
```
lucasbordeau: Could we put this into `executeTasksOnAnyLocationChange`, that would be a good use case here isn't it ?

lucasbordeau:
image

packages/twenty-front/src/modules/app/effect-components/PageChangeEffect.tsx
bosiraphael: Indeed

packages/twenty-front/src/modules/action-menu/components/PageHeaderActionMenuButtons.tsx
bosiraphael: Good catch

lucasbordeau (approved):

PR #15827, opened 2025-11-14
feat: use relative and absolute dates in email

BKM14:
Closes #15322

Used `formatToHumanReadableDate` to show absolute dates, rather than only relative dates. When the date is clicked, it flips between relative and absolute. The click on the date, doesn't show or hide the email body. Attaching a recording of the same below.

https://github.com/user-attachments/assets/77adcd3e-8efe-4af9-a57e-6317c93bcf22

https://github.com/user-attachments/assets/0731de02-60bb-42b1-a28c-55a94cff24b8


charlesBochet:
@Bonapara what do you think?

Bonapara:
thanks @BKM14!We should rather implement a tooltip

BKM14:
Hey guys, I have implemented the tooltip. I chose to use the `sentAt` as the tooltipId. I think this may cause some confusion if two emails have the same `sentAt`. Another alternative is to use the `useId` hook and use that as the tooltipId. Let me know if that needs to be changed. Thank you.

https://github.com/user-attachments/assets/23bd5449-8b84-449a-8fba-2617ded7b627


https://github.com/user-attachments/assets/5b0c531e-b6e8-4fbe-8724-510b2da27e2f

Bonapara:
We don't want a background color on the date hover, just a tooltip when hovering with the mouse.

BKM14:
Cool, I have removed the background color that appears on hover.

lucasbordeau (approved):

PR #15828, opened 2025-11-14
Improve record group aggregate query performance

lucasbordeau:
This PR is a first step for improving the performance on boards and table with groups.

It is related to : https://github.com/twentyhq/core-team-issues/issues/1870

Here we implement only a groupBy query for aggregate values in the group section.

This also allows to improve the DX of aggregate computing and group by query creation and parsing.

## Demo 

Main : 

https://github.com/user-attachments/assets/5d2a8077-5322-4928-a551-f03583bcfb87

This PR : 

https://github.com/user-attachments/assets/d0e82b28-72c3-40f0-b5cb-045f1a736ffb

## Aggregate update bug fix

This PR also solves a bug with aggregate update that was already present on main.

The bug is linked to core views not being updated properly during a modification of the aggregate operation on a view. 

We should probably improve the view lifecycle and state management because it is a bit too complex right now.

Main : 

https://github.com/user-attachments/assets/10dbfb8b-dfa0-4f21-8698-d222871a43e7

This PR : 

https://github.com/user-attachments/assets/bac41890-5191-4e4c-b82b-19b1039e9ab5

## Miscellaneous 

- Fixed optimistic rendering of group by queries, when adding a new record, the aggregate recomputes well. 

## TODO 

- We might want to improve the optimistic for group by queries that don't have records nor more than one dimension.


packages/twenty-front/src/modules/apollo/optimistic-effect/group-by/utils/doesRecordBelongToGroup.ts
```diff
@@ -1,48 +1,67 @@
 import { normalizeGroupByDimensionValue } from '@/apollo/optimistic-effect/group-by/utils/normalizeGroupByDimensionValue';
 import { type RecordGqlNode } from '@/object-record/graphql/types/RecordGqlNode';
+import { isArray } from '@apollo/client/utilities';
 import { isDefined } from 'twenty-shared/utils';
 
 export const doesRecordBelongToGroup = (
   record: RecordGqlNode,
   groupByDimensionValues: readonly string[],
-  groupByConfig?: Array<Record<string, boolean | Record<string, string>>>,
+  groupByConfig?:
```
charlesBochet, line 9: @lucasbordeau why do we need to support both? I'm a bit lost here! isn't the Record<string, boolean | Record<string, string>> case a sub case of Array with only one item in the array?

charlesBochet: Actually taking it back groupByConfig is never an array to me

charlesBochet: Image

packages/twenty-front/src/modules/object-metadata/hooks/useDeleteOneFieldMetadataItem.ts
```diff
@@ -23,19 +26,25 @@ export const useDeleteOneFieldMetadataItem = () => {
...
+  const setRecordIndexGroupAggregateOperation = useSetRecoilComponentState(
+    recordIndexGroupAggregateOperationComponentState,
+    MAIN_CONTEXT_STORE_INSTANCE_ID,
+  );
+
   const [
-    recordIndexKanbanAggregateOperation,
-    setRecordIndexKanbanAggregateOperation,
-  ] = useRecoilState(recordIndexKanbanAggregateOperationState);
+    recordIndexGroupAggregateFieldMetadataItem,
```
charlesBochet, line 35: ok but I think it is still named this way in db

packages/twenty-front/src/modules/object-record/record-index/hooks/useLoadRecordIndexStates.ts
```diff
@@ -180,36 +194,58 @@ export const useLoadRecordIndexStates = () => {
...
+          );
+
+        if (isDefined(view.kanbanAggregateOperation)) {
+          const convertedAggregateOperation =
+            convertAggregateOperationToExtendedAggregateOperation(
+              view.kanbanAggregateOperation,
+              recordIndexGroupAggregateFieldMetadataItem?.type,
+            );
+
+          setRecordIndexGroupAggregateOperation(convertedAggregateOperation);
```
charlesBochet, line 229: ok

packages/twenty-front/src/modules/object-record/record-index/hooks/useSetRecordIndexAggregateDisplayLabel.ts
```diff
@@ -0,0 +1,34 @@
...
+        recordIndexGroupAggregateOperation: ExtendedAggregateOperations,
+        recordIndexGroupAggregateFieldMetadataItem: FieldMetadataItem,
+      ) => {
+        const { aggregateLabel } = getRecordAggregateDisplayLabel({
+          aggregateOperation: recordIndexGroupAggregateOperation,
+          aggregateFieldMetadataItem:
+            recordIndexGroupAggregateFieldMetadataItem,
+        });
+
+        set(recordIndexAggregateValueByGroupValueCallbackState, aggregateLabel);
```
charlesBochet: do we really need a state for that? I feel this should be derived

packages/twenty-front/src/modules/object-record/record-index/types/RecordAggregateValueByRecordGroupValue.ts
```diff
@@ -0,0 +1,6 @@
+import { type Nullable } from 'twenty-shared/types';
+
+export type RecordAggregateValueByRecordGroupValue = {
```
charlesBochet, line 3: RecordGroupValue => RecordGroup ?

packages/twenty-front/src/modules/views/hooks/useUpdateViewAggregate.ts
```diff
@@ -1,77 +1,91 @@
...
+        if (updatedViewResult.status === 'successful') {
+          const updatedCoreView = updatedViewResult.response.data
+            ?.updateCoreView as CoreView;
+
+          if (!isDefined(updatedCoreView)) {
+            return;
+          }
+
+          set(coreViewsState, (currentCoreViews) =>
+            produce(currentCoreViews, (draftCoreViews) => {
```
charlesBochet: a bit exotic?

packages/twenty-front/src/modules/object-metadata/hooks/useDeleteOneFieldMetadataItem.ts
lucasbordeau: Let's consider DB deprecated naming then ? Because otherwise it means we'll have a hard time maintaining the wrong semantic in our codebase.

lucasbordeau: We could even transform this into our transform utils between core views and views but it risks introducing bugs

packages/twenty-front/src/modules/object-record/record-index/hooks/useSetRecordIndexAggregateDisplayLabel.ts
lucasbordeau: No it should have been removed already.

packages/twenty-front/src/modules/views/hooks/useUpdateViewAggregate.ts
lucasbordeau: Ok right, I'll create an array util for this use case which is quite common.

packages/twenty-front/src/modules/object-record/record-index/types/RecordAggregateValueByRecordGroupValue.ts
lucasbordeau: I would stick to this because RecordGroup alone can be misleading, it does not hurt here to be very clear.

packages/twenty-front/src/modules/object-record/record-index/hooks/useSetRecordIndexAggregateDisplayLabel.ts
lucasbordeau: Yes we need a state to avoid weird flashes when switching between views, it avoids having complex loading management with queries, page switching, other states being set before the query finishes.

charlesBochet (approved):

PR #15823, opened 2025-11-14
`[BREAKING_CHANGE/GRAPHQL/OBJECT_METADATA_CREATE_ONE]` Remove object/fields/view-fields v1 implementation

prastoin:
# Introduction
Remove the v2 feature flag for view-field field-metadata and object-metadata metadata entities

## Some details
- Disabled nestjs-query for object metadata creation and explicitly calling it
- removed all v1 integration tests files

## Remarks
Not remove v2 referencing in both filenaming right now will handle that globally later

## Breaking change
Due to object metadata resolver createOne standardization had to rename the input from `CreateObjectInput` to `CreateOneObjectInput`

related https://github.com/twentyhq/core-team-issues/issues/1911

PR #15830, opened 2025-11-14
Runner v2 on cascade `delete_field/object` + refactor workspace deletion side effect

prastoin:
# Introduction
While removing the v1 https://github.com/twentyhq/twenty/pull/15823 I've encountered the method `objectMetadataService.deleteObjectsMetadata` that I didn't wanted to migrate as it is and if it's not challenging its existence legitimacy.

## Motivations
In a nutshell on a workspace deletion object metadata and field metadata cascading deletion is correclty handled
But that's not the case for all of a workspaces entities ( roles, workspaceMigrations ) I suspect that we did not defined the foreignKey explicitly through `typeorm`

## Battle testing v2
I still decided to give a try to a complex operation in the v2 such as a workspace all object metadata deletion.
Spoiler it failed due to object being interdependent between them and not being topologically sorted ( morph relation can introduce circular dep anw ).
Note: Even after removing the delete field on delete object aggregator we end up with an equivalent circular dep error which an object and its field metadata identifier connection.

## Elegant `DEFERRED` and `DEFERRABLE` foreign keys
The most safe, low level solution would be to make all field relations deferrable and start the runner transaction as deferred. But this requires a quite invasive migration of existing FK

## On cascade solution
I've opted for the quick fix, on object or field deletion spread cacasde.
It's pretty safe as the builder priorly validates the deletion and and its related entities integrity 
Only for both field and object deletion action types in v2 runner

## Integration coverage
Added a test that will scan a new workspace database core schema tables and expect now result after workspace deletion through its only user deletion

### DIFF
--- a/packages/twenty-server/src/engine/metadata-modules/object-metadata/object-metadata-v2.service.ts
+++ b/packages/twenty-server/src/engine/metadata-modules/object-metadata/object-metadata-v2.service.ts
@@ -180,6 +180,7 @@ export class ObjectMetadataServiceV2 {
     }
 
     const [deletedObjectMetadataDto] = deletedObjectMetadataDtos;
+
     return deletedObjectMetadataDto;
   }
 
--- a/packages/twenty-server/src/engine/metadata-modules/object-metadata/object-metadata-v2.service.ts
+++ b/packages/twenty-server/src/engine/metadata-modules/object-metadata/object-metadata-v2.service.ts
@@ -218,57 +218,54 @@ export class ObjectMetadataServiceV2 {
       flatObjectMetadatasToDeleteById: {},
     };
 
-    const flatEntityToProcessById = deleteObjectInputs.reduce(
-      (accumulator, deleteObjectInput) => {
-        const {
-          flatFieldMetadatasToDelete,
-          flatObjectMetadataToDelete,
-          flatIndexToDelete,
-        } = fromDeleteObjectInputToFlatFieldMetadatasToDelete({
-          flatObjectMetadataMaps,
-          flatIndexMaps,
-          flatFieldMetadataMaps,
-          deleteObjectInput,
-        });
+    const {
+      flatFieldMetadatasToDeleteById,
+      flatIndexToDeleteById,
+      flatObjectMetadatasToDeleteById,
+    } = deleteObjectInputs.reduce((accumulator, deleteObjectInput) => {
+      const {
+        flatFieldMetadatasToDelete,
+        flatObjectMetadataToDelete,
+        flatIndexToDelete,
+      } = fromDeleteObjectInputToFlatFieldMetadatasToDelete({
+        flatObjectMetadataMaps,
+        flatIndexMaps,
+        flatFieldMetadataMaps,
+        deleteObjectInput,
+      });
 
-        return {
-          flatFieldMetadatasToDeleteById: {
-            ...accumulator.flatFieldMetadatasToDeleteById,
-            ...fromArrayToUniqueKeyRecord({
-              array: flatFieldMetadatasToDelete,
-              uniqueKey: 'id',
-            }),
-          },
-          flatIndexToDeleteById: {
-            ...accumulator.flatIndexToDeleteById,
-            ...fromArrayToUniqueKeyRecord({
-              array: flatIndexToDelete,
-              uniqueKey: 'id',
-            }),
-          },
-          flatObjectMetadatasToDeleteById: {
-            ...accumulator.flatObjectMetadatasToDeleteById,
-            [flatObjectMetadataToDelete.id]: flatObjectMetadataToDelete,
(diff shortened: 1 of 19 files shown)


packages/twenty-server/src/engine/metadata-modules/object-metadata/object-metadata.entity.ts
```diff
@@ -107,6 +107,7 @@ export class ObjectMetadataEntity
   @Column({ default: false })
   isLabelSyncedWithName: boolean;
 
+  // TODO create a relation to workspace with cascade delete will also create foreignKey
```
prastoin, line 110: @charlesBochet what do you think about this ?
Would it be a goal to be able to delete a workspace from one delete operation ?

prastoin:
@greptileai trigger

packages/twenty-server/src/engine/metadata-modules/object-metadata/object-metadata-v2.service.ts
```diff
@@ -165,50 +166,127 @@ export class ObjectMetadataServiceV2 {
     workspaceId: string;
     isSystemBuild?: boolean;
   }): Promise<ObjectMetadataDTO> {
+    const deletedObjectMetadataDtos = await this.deleteManyObjectMetadatas({
```
FelixMalfait, line 169: nit: for me DTO was referring to the transport layer not the data itself

packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-runner-v2/interfaces/workspace-migration-runner-action-handler-service.interface.ts
```diff
@@ -73,21 +73,26 @@ export abstract class BaseWorkspaceMigrationRunnerActionHandlerService<
...
+          method: async () => this.executeForWorkspaceSchema(context),
+        }),
+      ]);
+
+      return this.optimisticallyApplyActionOnAllFlatEntityMaps({
+        action: context.action,
+        allFlatEntityMaps: context.allFlatEntityMaps,
+      });
+    } catch (error) {
+      this.logger.error(`${this.actionType} execution failed`, error);
```
FelixMalfait, line 93: Feels a bit off that we have to wrap in try catch to log an re-throw the error? If it's re-thrown then shouldn't it be logged by whatever catches it at the end?

packages/twenty-server/test/integration/graphql/suites/workspace/successful-user-and-workspace-creation.integration-spec.ts
```diff
@@ -118,4 +119,128 @@ describe('Successful user and workspace creation', () => {
...
+      'fieldMetadata',
+      'indexMetadata',
+      'searchFieldMetadata',
+      'workspaceMigration',
+      'role',
+      'roleTargets',
+      'objectPermission',
+      'fieldPermission',
+      'permissionFlag',
+      'serverlessFunction',
```
FelixMalfait, line 192: Those are deleted via foreign key cascading from workspace directly?

FelixMalfait (approved):

packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-runner-v2/interfaces/workspace-migration-runner-action-handler-service.interface.ts
prastoin: I understand your concern, the plus value of an intermediary log here is logging the current action type that failed being executed
Indeed this exception will be catch by the builder v2 globally

packages/twenty-server/test/integration/graphql/suites/workspace/successful-user-and-workspace-creation.integration-spec.ts
prastoin: Yes I think so, we should standardize explicitly declaring forgein key relation between entities and their workspaces so we can realy on on delete cascading for all of them
Or at least document the ones that do not comply to this standard and the motivations why they don't

packages/twenty-server/src/engine/metadata-modules/object-metadata/object-metadata-v2.service.ts
prastoin: Indeed we could add an abstraction layer here to make the `deleteManyObjectMetadatas` return `FlatObjectMetadata` instead of dtos

prastoin: done in https://github.com/twentyhq/twenty/pull/15823

PR #15834, opened 2025-11-15
feat: resolve preview merge record uses a fake one

harshit078:
## Description

- This PR fixes the first issue out of #15486
- created a  Recoil atom mergePreviewRecordFamilyState that is used  to store the preview record.
- Preview items are now stored in `mergePreviewRecordFamilyState` which avoiding pollution of the `recordStoreFamilyState`


packages/twenty-front/src/modules/object-record/record-merge/hooks/useMergePreview.ts
```diff
@@ -33,59 +35,106 @@ export const useMergePreview = ({
...
+          setMergePreviewRecord(null);
+          setPreviewRecordById(null, lastPreviewId);
+          setLastPreviewId('');
+        }
       } finally {
-        setIsGeneratingPreview(false);
-        setIsInitialized(true);
+        if (runId === currentRunId + 1) {
+          setIsGeneratingPreview(false);
+        }
```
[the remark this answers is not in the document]
harshit078: fixing this !

packages/twenty-front/src/modules/object-record/record-merge/hooks/useMergePreview.ts
```diff
@@ -17,10 +17,13 @@ type UseMergePreviewProps = {
...
 }: UseMergePreviewProps) => {
+  const [isGeneratingPreview, setIsGeneratingPreview] = useState(false);
   const [mergePreviewRecord, setMergePreviewRecord] =
     useState<ObjectRecord | null>(null);
-  const [isGeneratingPreview, setIsGeneratingPreview] = useState(false);
-  const [isInitialized, setIsInitialized] = useState(false);
+  const [lastPreviewId, setLastPreviewId] = useState<string>('');
+  const [lastPreviewSignature, setLastPreviewSignature] = useState<string>('');
+  // eslint-disable-next-line @nx/workspace-no-state-useref
+  const currentRunIdRef = useRef(0);
```
charlesBochet, line 26: @harshit078 we don't use useRef in the project

packages/twenty-front/src/modules/object-record/record-merge/hooks/useMergePreview.ts
```diff
@@ -33,59 +36,105 @@ export const useMergePreview = ({
     commandMenuNavigationMorphItemsByPageState,
   );
 
-  const selectedRecordIds =
-    commandMenuNavigationMorphItemsByPage
-      .get(CommandMenuPages.MergeRecords)
-      ?.map((morphItem) => morphItem.recordId) ?? [];
+  const selectedRecordIds = useMemo(
```
charlesBochet, line 39: useMemo does not seems necessary here?

packages/twenty-front/src/modules/object-record/record-merge/hooks/useMergePreview.ts
```diff
@@ -33,59 +36,105 @@ export const useMergePreview = ({
...
       recordIds: selectedRecordIds,
     }),
   );
 
-  const { upsertRecordsInStore } = useUpsertRecordsInStore();
+  const setPreviewRecordById = useRecoilCallback(
+    ({ set }) =>
+      (record: ObjectRecord | null, previousId: string) => {
+        if (previousId !== '') {
+          set(mergePreviewRecordFamilyState(previousId), null);
```
charlesBochet, line 56: I'm not sure about this, this does not seem robust, why do we need to clean the state?

packages/twenty-front/src/modules/object-record/record-merge/hooks/useMergePreview.ts
```diff
@@ -33,59 +36,105 @@ export const useMergePreview = ({
...
+        }
+        if (record !== null) {
```
charlesBochet, line 58: use isDefined, isNonEmptyString helpers

packages/twenty-front/src/modules/object-record/record-merge/hooks/useMergePreview.ts
```diff
@@ -33,59 +36,105 @@ export const useMergePreview = ({
...
+  const previewSignature = useMemo(
+    () =>
+      `${selectedRecordIds.join(',')}|${mergeSettings.conflictPriorityIndex}`,
+    [selectedRecordIds, mergeSettings.conflictPriorityIndex],
+  );
 
   useEffect(() => {
     const fetchPreview = async () => {
-      if (selectedRecords.length < 2 || isMergeInProgress || isInitialized)
+      if (selectedRecordIds.length < 2 || isMergeInProgress) {
```
charlesBochet, line 73: this code is not robust IMO, isInitialized seemed right. What issue are we trying to solve?

charlesBochet (changes requested):
@harshit078 ty

Feedbacks:
- let's not use useRef. What we should aim for: hooks that are called synchronously. Here, I think we should add records to the mergePreviewFamilyState in useOpenMergeRecordsPageInCommandMenu for instance. this would remove the need for useRef, useEffect, useMemo and all other hacks.

I think we should not try to clean the mergePreviewFamilyState, it does not worth it.

Also, I'm not convinced by the fact that recordStore is now loading data from two different sources but why not, it does not harm too much

harshit078:
@charlesBochet Hi charles, thanks for the comments. I am super aware that Twenty ooesnt use useRef but I couldnt find another approach round it which would pass lint test and was ideal. I'll address the comments and refactor the approach thanks !

charlesBochet:
@harshit078 I'm going to close it for now, let's re-open when ready

PR #15837, opened 2025-11-16
add case for calendar revoked refresh token

neo773:


Weiko (approved):
LGTM

packages/twenty-server/src/modules/calendar/calendar-event-import-manager/services/calendar-account-authentication.service.ts
```diff
@@ -86,6 +86,7 @@ export class CalendarAccountAuthenticationService {
           );
         case ConnectedAccountRefreshAccessTokenExceptionCode.REFRESH_ACCESS_TOKEN_FAILED:
         case ConnectedAccountRefreshAccessTokenExceptionCode.REFRESH_TOKEN_NOT_FOUND:
+        case ConnectedAccountRefreshAccessTokenExceptionCode.INVALID_REFRESH_TOKEN:
           throw new CalendarEventImportDriverException(
```
Weiko, line 90: Just realised we don't track those errors like we do in MessagingAccountAuthenticationService, we might want to do that as well (cc @charlesBochet)

Weiko: FYI @neo773

PR #15840, opened 2025-11-16
fix: babel vulnerable to arbitrary code execution when compiling specifically crafted malicious code

mabdullahabaid:
Resolves [Dependabot Alert 95](https://github.com/twentyhq/twenty/security/dependabot/95) - babel vulnerable to arbitrary code execution when compiling specifically crafted malicious code.

These were the few options we had for a direct drop-in replacement.
- [x-var](https://www.npmjs.com/package/x-var?activeTab=readme)
- [cross-let](https://www.npmjs.com/package/cross-let)
- [cross-var-no-babel](https://www.npmjs.com/package/cross-var-no-babel)

x-var has the most weekly downloads among the three and it is also the most actively maintained fork of the original cross-var package that introduced the vulnerability. There is no syntax difference per the documentation, but I do not have a windows machine to test.

`cross-var-no-babel` offers the most minimal changes, but is also abandoned without a public-facing repo.


FelixMalfait:
As discussed in DM let's try to remove cross-var instead

FelixMalfait (approved):

PR #15866, opened 2025-11-17
expose imap username field

neo773:
Had an edge with a customer where they had a custom username.


packages/twenty-server/src/modules/messaging/message-import-manager/drivers/imap/providers/imap-client.provider.ts
```diff
@@ -88,7 +88,7 @@ export class ImapClientProvider {
         port: connectionParameters.IMAP?.port || 993,
         secure: connectionParameters.IMAP?.secure,
         auth: {
-          user: connectedAccount.handle,
+          user: connectionParameters.IMAP?.username ?? connectedAccount.handle,
```
charlesBochet: @neo773 let's use isNonEmptyString indeed

charlesBochet (approved):
Left a comment, approving otherwise

charlesBochet (approved):

PR #15872, opened 2025-11-17
Refactor feature flag migration v2

Weiko:


packages/twenty-server/src/engine/core-modules/feature-flag/services/feature-flag.service.ts
```diff
@@ -36,100 +33,149 @@ export class FeatureFlagService {
...
-  }): Promise<FeatureFlagEntity> {
-    featureFlagValidator.assertIsFeatureFlagKey(
-      featureFlag,
-      new FeatureFlagException(
-        'Invalid feature flag key',
-        FeatureFlagExceptionCode.INVALID_FEATURE_FLAG_KEY,
-      ),
-    );
-
-    if (shouldBePublic) {
```
Weiko, line 99: TODO: Check how to move that logic to v2 validator, not that easy since shouldBePublic is not expected in the flat entity

Weiko:
TODO: discuss if it's actually relevant to have feature flag as syncable, we might simply create a flat entity and cache it with the rest without being syncable

Weiko:
Should only implement in cache and not make it syncable. Closing for now

PR #15873, opened 2025-11-17
preventing the last item to be unselected

### LINKED ISSUE #15867: [MORPH RELATIONS] - Cannot select destination objects anymore once all were uselected
## Bug Description
If you unselect all the possible destination objects, you can no longer access the UI to select one. 

Image

Image

Image

guillim:
Fix https://github.com/twentyhq/twenty/issues/15867


packages/twenty-front/src/modules/settings/components/SettingsMorphRelationMultiSelect.tsx
```diff
@@ -147,6 +147,9 @@ export const SettingsMorphRelationMultiSelect = ({
   const addOrRemoveFromArray = (array: string[], item: string) => {
     let newArray = new Set(array);
```
Devessier, line 148: nitpick: the variable is called `newArray` but it's not an array

Devessier (approved):
LGTM!

PR #15877, opened 2025-11-17
Enhance StyledInnerContainer for responsiveness (Resizable Left Panel)

submgr:
This PR addresses the issue where field values in the Left Panel were getting truncated. Thus, even small to medium values, like 30 characters long, would get truncated, —  which is a bad user experience, as the user needs to hover over it to see its full value.

I have implemented resizable sidebar logic that allows users to drag the right edge of the column to adjust its width according to their needs.

**Future Proposal:** while my solution is pretty basic, in the future it would be great to implement saving the last width set by the user (maybe as a preference in localStorage of user's browser). This would remove the need to expand the left panel each time in cases when, for example, this object always has medium-sized and longer field values... 

Before:
Screenshot 2025-11-17 at 23 06 48

After:
Screenshot 2025-11-17 at 23 07 38

This is my first contribution, hope it helps!

### DIFF
--- a/packages/twenty-front/src/modules/ui/layout/show-page/components/ShowPageLeftContainer.tsx
+++ b/packages/twenty-front/src/modules/ui/layout/show-page/components/ShowPageLeftContainer.tsx
@@ -20,7 +20,19 @@ const StyledOuterContainer = styled.div<{ isMobile: boolean }>`
 const StyledInnerContainer = styled.div<{ isMobile: boolean }>`
   display: flex;
   flex-direction: column;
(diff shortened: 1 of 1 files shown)


submgr:
> 1. **syntax:** String value `'auto'` should not be quoted - it becomes a literal string `'auto'` instead of the CSS keyword `auto`.

It's not part of my edit

charlesBochet:
@submgr thanks for opening a PR, unfortunately, I don't think we want to go in this direction from a UI perspective. Our long term vision is to let user customize the size of the leftPanel, but this comes as a broader feature called "Custom Layout"

I'm going to closing it but appreciate the energy

PR #15884, opened 2025-11-18
fix: update glob version to 11.1.0

mabdullahabaid:
Some Glob CLI related alerts were generated last night. I believe they're safe to dismiss since I do not expect us to use the Glob CLI in production environment, and those alerts do not impact the API, but still updating the dependency version just in case.

Not sure if this would resolve all/any of the alerts since glob is a dependency for many other dependencies, so a good number of dependency variants are pulled in.

The alert that confirms it's just a CLI related vulnerability: [Dependabot Alert 307](https://github.com/twentyhq/twenty/security/dependabot/307)


FelixMalfait (approved):

packages/twenty-sdk/package.json
```diff
@@ -20,6 +20,7 @@
     "vite-tsconfig-paths": "^4.2.1"
   },
   "dependencies": {
+    "glob": "^11.1.0",
```
FelixMalfait, line 23: It's used in twenty-ui and twenty-shared too?

mabdullahabaid: Yep. `migrateFromSingleToMultiBarrelImport` script uses it.

FelixMalfait: But you only update it in one place? Also is it a dependency or a dev dependency?

mabdullahabaid: No, I actually updated it all three. There are two commits, you might be looking at the second one only.

There are a total of 6 files changed. It is listed as a dependency (so I did not try moving to devDependency), although I do believe we can specify it as a devDependency since it's used in scripts and those scripts are probably only invoked in local, not on prod.

There are a total of 5 scripts dependent on it.
- `generateBarrels.ts` in twenty-sdk, twenty-ui, and twenty-shared.
- `migrateFromSingleToMultiBarrelImport.ts` in twenty-ui and twenty-shared.

The second one is probably not needed in prod, but I am unsure about the generateBarrels.ts script.

Can you confirm?

yarn.lock
```diff
@@ -36724,6 +36708,38 @@ __metadata:
   languageName: node
   linkType: hard
 
+"glob@npm:^11.0.0":
```
charlesBochet: @mabdullahabaid looks like we still have 11.0 version?

charlesBochet (approved):

mabdullahabaid:
@charlesBochet fixed it. 

@lingui/cli was using 11.0.1, but allowed the semver upgrade (^11.0.0). So, running `yarn up glob --recursive` ensured it uses 11.1.0 instead.

yarn.lock still has an entry saying ^11.0.0, but the actual version installed is 11.1.0 as confirmed by the second line in the screenshot.

<p align="center">
image
</p>

We cannot unfortunately fix the entire issue because `@nest/cli` uses glob fixed at 11.0.3 and `@nest/cli` is already the latest version.

One interesting thing: 
Dependabot alerts for Glob have completely disappeared from the security tab - they're not open, they're not closed. Instead, visiting the URL says that they have been withdrawn. So, we can decide to merge/work on this PR or disregard it since it no longer serves a purpose.

<p align="center">
image
</p>

charlesBochet:
Let's merge it if it doesn't harm

PR #15885, opened 2025-11-18
Add optional limit variable to groupBy queries

### LINKED ISSUE #1600: Add limit to groupBy api

ijreilly:
Closes https://github.com/twentyhq/core-team-issues/issues/1600.

Two remarks 
- This `limit` variable does not reduce postgre's work at it still needs to scan the whole table. It did not seem possible to me to optimize this as we cannot foresee which dimensions will be used by the user, and an optimization could only result from an index on the dimension(s) (e.g.: group companies by addressCity limit 50 can be optimized if we have an index on companies.addressCity + we had a default orderBy on adressCity). But this will still optimize the FE which at the moment receives all groups and truncates the result. 
- I have not done the work on the FE as the addition of limit is a breaking change, and will break until the workspaces' schema is rebuilt, so we need to flush the cache. I think this could be acceptable as the feature is in the lab but I preferred not doing it yet as it would have no impact since in the BE I added a default limit to 50 groups, and I expect more FE work will be done to allow the user to choose their own limit

### DIFF
--- a/packages/twenty-front/src/modules/page-layout/widgets/graph/hooks/useGraphWidgetGroupByQuery.ts
+++ b/packages/twenty-front/src/modules/page-layout/widgets/graph/hooks/useGraphWidgetGroupByQuery.ts
@@ -6,6 +6,7 @@ import { generateGroupByQuery } from '@/page-layout/widgets/graph/utils/generate
 import { generateGroupByQueryVariablesFromChartConfiguration } from '@/page-layout/widgets/graph/utils/generateGroupByQueryVariablesFromChartConfiguration';
 import { useQuery } from '@apollo/client';
 import { useMemo } from 'react';
+import { DEFAULT_NUMBER_OF_GROUPS_LIMIT } from 'twenty-shared/constants';
 import { isDefined } from 'twenty-shared/utils';
 
 export const useGraphWidgetGroupByQuery = ({
@@ -52,6 +53,7 @@ export const useGraphWidgetGroupByQuery = ({
   const variables = {
     ...groupByQueryVariables,
     filter: gqlOperationFilter,
+    limit: DEFAULT_NUMBER_OF_GROUPS_LIMIT,
   };
 
   const query = generateGroupByQuery({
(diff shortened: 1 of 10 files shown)


packages/twenty-front/src/modules/page-layout/widgets/graph/hooks/useGraphWidgetGroupByQuery.ts
```diff
@@ -52,6 +53,7 @@ export const useGraphWidgetGroupByQuery = ({
   const variables = {
     ...groupByQueryVariables,
     filter: gqlOperationFilter,
+    limit: DEFAULT_NUMBER_OF_GROUPS_LIMIT,
```
ijreilly, line 56: has no effect until we update generateGroupByQuery, but i did not want to do it yet to avoid a breaking change

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/utils/get-group-limit.util.ts
```diff
@@ -0,0 +1,14 @@
...
+  if (
+    typeof limit === 'number' &&
+    Number.isFinite(limit) &&
+    limit > 0 &&
+    Number.isInteger(limit)
+  ) {
+    return limit;
+  }
+
+  return DEFAULT_NUMBER_OF_GROUPS_LIMIT;
```
etiennejouan, line 13: Should we not add a limit max ? (and not just default - can be in addition, default and max)
For example, I can group by id with records and so getting all records

etiennejouan (approved):
LGTM 👍

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/utils/get-group-limit.util.ts
ijreilly: thats a good idea, but since we have not introduced pagination here it could be an unavoidable limitation, with some groups of records you could never actually. So Id be in favour of not doing it now, but we can re-think about it if we see abuse.

PR #15892, opened 2025-11-18
Set TTL >> max await + increase max await from 1s to 2s

martmull:
increase max await from 1s to 5s

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/cache-lock/cache-lock.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/cache-lock/cache-lock.service.ts
@@ -26,7 +26,7 @@ export class CacheLockService {
     key: string,
     options?: CacheLockOptions,
   ): Promise<T> {
-    const { ms = 50, maxRetries = 20, ttl = 500 } = options || {};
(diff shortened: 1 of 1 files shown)


packages/twenty-server/src/engine/core-modules/cache-lock/cache-lock.service.ts
```diff
@@ -26,7 +26,7 @@ export class CacheLockService {
     key: string,
     options?: CacheLockOptions,
   ): Promise<T> {
-    const { ms = 50, maxRetries = 20, ttl = 500 } = options || {};
+    const { ms = 50, maxRetries = 40, ttl = 10_000 } = options || {};
```
charlesBochet, line 29: @martmull shouldn't we put the ttl to be approximatively 50*40 = 2000

martmull: chatgpt tells me to keep a generous margin

charlesBochet: 5_000 + ms 50, maxRetries 100

charlesBochet (approved):

PR #15895, opened 2025-11-18
Fix test

etiennejouan:


charlesBochet (commented):
It's because twenty-companies.com was likely down, we should mock the httpService

charlesBochet (approved):

Weiko (commented):
LGTM

PR #15898, opened 2025-11-18
fix-15838

### LINKED ISSUE #15838: Relations not visible anymore from tasks - v1.11.0 Self Hosted
## Bug Description

Hey team!

Since the last update this weekend, no relations are visible in tasks. They are still created correctly and the relation is saved properly.
During task creation, the relation is visible, but after a refresh, it disappears.

Relations (opportunity, person, company, etc.) are still accessible from their respective views, so the issue seems to be mainly in how they are displayed.

Thanks for your help!

Image
Image
Image

## Technical inputs

Nothing much on the logs, but to be noted: the database is not impacted, the relations are indeed well saved.

guillim:
I know you did this change in order to make the frontend faster and less resource eating. However it broke this optimistic rendering part @lucasbordeau. 

I suggest this quick fix, but I am open to a sharper one as well

Fixes https://github.com/twentyhq/twenty/issues/15838


charlesBochet (changes requested):
I'm in favor of fixing the root cause, it should be straight forward :)

Why is the issue happening? getting labelIdentifiers should be enough to display the picker!

charlesBochet:
@guillim I tried to reproduce on main but could not?

lucasbordeau:
This is probably because the optimization does not work in the opposite side of a task target, in this case we should query only the label identifier of each possible task target record

charlesBochet:
Yes exactly, I think: what we want is
image

So the changes: 
-` depth: 1` agree
- `shouldOnlyLoadActivityIdentifiers: false` (i'm not sure why this boolean is useful, I don't think we ever want to only load only task or note @lucasbordeau 
- internally to `generateActivityTargetGqlFields`, let's add `shouldOnlyLoadRelationIdentifiers = true`

packages/twenty-front/src/modules/object-record/graphql/record-gql-fields/utils/generateActivityTargetGqlFields.ts
charlesBochet, line 7: would be nice to add unit tests on this one, looks accurate here

charlesBochet: we should Pick here

charlesBochet:
Tests are red and would be nice to add a unit test on the util

packages/twenty-front/src/modules/object-record/graphql/record-gql-fields/utils/generateActivityTargetGqlFields.ts
```diff
@@ -45,15 +45,33 @@ export const generateActivityTargetGqlFields = ({
     const activityLabelIdentifierFieldMetadataItem =
       getLabelIdentifierFieldMetadataItem(activityObjectMetadataItem);
 
-    return {
-      id: true,
-      [activityObjectNameSingular]: {
+    if (depth === 0) {
```
charlesBochet: to me it doesn't make sense:
depth 0, we should not load the taskTarget.task

We should have an util that:
- loads either `task | note` OR `relations` OR `both` (shouldLoadRelations: 'activity' | 'relations' | 'both')
- no need for depth: 0 (why would we load only the target without anything inside it?
- always apply  shouldOnlyLoadRelationIdentifiers: true (which is the case now)

charlesBochet (approved):

guillim:
We should do a patch in prod once this is pushed. @StephanieJoly4 suggested it.

PR #15899, opened 2025-11-18
Fix merge button

ijreilly:
Fixes https://github.com/twentyhq/private-issues/issues/362

**How to reproduce**
Link a person that has a duplicate to an opportunity. (you can create a duplicate by giving two people the same linkedin link).
Open the opportunity record from opportunity table. 
Click on the related person (in "Point of contact").
In front of "Duplicates", click on the merge button (two arrows becoming one).
You should here see an empty "Merge preview" and an error when clicking "First" tab.

**Issue**
The issue is that CommandMenuMergeRecordPage is getting the referenced objectMetadataItem from useContextStoreObjectMetadataItemOrThrow without an instance id. contextStoreObjectMetadataItem is still "opportunity" as it should be, being on an opportunity view. 
So further down it attempts to display the record page according the label identifier field from opportunity, which is a text field, "name", and it breaks because the record is actually a person for who the "name" field is not a text but a full_name type. 

**Fix**
I suggested a fix that offers the possibility to find the referenced objectMetadataItem from the MergeRecords instanceId. But I still gave flexibility to avoid having to set that state everytime we open the merge tab, by falling back to the default contextStoreObjectMetadataItem. (this is used when we merge records from ticking two records from a view and open the command menu).
Im not sure this is the best option. Open to suggestions !


ijreilly:
probably attempts fixing the same error as here https://github.com/twentyhq/twenty/pull/15834/files , see which approach is best

packages/twenty-front/src/modules/command-menu/hooks/useCommandMenu.ts
```diff
@@ -36,6 +37,12 @@ export const useCommandMenu = () => {
           removeFocusItemFromFocusStackById({
             focusId: SIDE_PANEL_FOCUS_ID,
           });
+          set(
+            contextStoreCurrentObjectMetadataItemIdComponentState.atomFamily({
+              instanceId: CommandMenuPages.MergeRecords,
```
charlesBochet: I believe we already have what we need in contextStoreCurrentObjectMetadataItemIdComponentState with instanceId: COMMAND_MENU_COMPONENT_INSTANCE_ID

checking

charlesBochet: Discussed with Raphael, we think it's actually the right approach

packages/twenty-front/src/modules/command-menu/pages/record-page/components/CommandMenuMergeRecordPage.tsx
```diff
@@ -20,12 +20,17 @@ const StyledRightDrawerRecord = styled.div<{
...
 
-  const { objectMetadataItem } = useContextStoreObjectMetadataItemOrThrow();
-
   const commandMenuPageInstanceId = useComponentInstanceStateContext(
     CommandMenuPageComponentInstanceContext,
   )?.instanceId;
 
+  const canDefaultToDefaultObjectMetadataItem = true;
+
+  const objectMetadataItem = useContextStoreObjectMetadataItemOrThrow(
```
charlesBochet: I think this is not useful, let's always have it available for mergePreview

packages/twenty-front/src/modules/command-menu/hooks/useOpenMergeRecordsPageInCommandMenu.tsx
```diff
@@ -35,22 +39,43 @@ export const useOpenMergeRecordsPageInCommandMenu = ({
     },
   });
 
-  const openMergeRecordsPageInCommandMenu = async () => {
-    await updateCommandMenuNavigationMorphItemsByPage({
-      pageId: CommandMenuPages.MergeRecords,
-      objectMetadataId: objectMetadataItem.id,
+  const openMergeRecordsPageInCommandMenu = useRecoilCallback(
```
charlesBochet, line 42: Change to the original PR:
- use a pageId = v4() instead of CommandMenu.MergePage to avoid conflicts between merge pages
- set the contextStore in this specific hook instead of generic openInCommandMenu (we only need to set it when we open merge page)

packages/twenty-front/src/modules/command-menu/pages/record-page/components/CommandMenuMergeRecordPage.tsx
```diff
@@ -34,15 +37,21 @@ export const CommandMenuMergeRecordPage = () => {
     <RecordComponentInstanceContextsWrapper
       componentInstanceId={`record-merge-${commandMenuPageInstanceId}`}
     >
-      <ActionMenuComponentInstanceContext.Provider
-        value={{ instanceId: commandMenuPageInstanceId }}
+      <ContextStoreComponentInstanceContext.Provider
```
charlesBochet, line 40: Change 2:
- adding a ContextStoreProvider here so we child component can get the information easily

packages/twenty-front/src/modules/object-record/record-merge/components/MergePreviewTab.tsx
```diff
@@ -14,7 +14,7 @@ type MergePreviewTabProps = {
 export const MergePreviewTab = ({
   objectNameSingular,
 }: MergePreviewTabProps) => {
-  const { mergePreviewRecord, isGeneratingPreview } = useMergePreview({
+  const { mergePreviewRecord, isGeneratingPreview } = usePerformMergePreview({
```
charlesBochet, line 17: Not related to the original PR, I'm splitting useMergePreview in two hooks:
- usePerformMergePreview: used to make the call to the backend to generate merge (dry run), this hooks actually contains a useEffect
- useMergeRecordsSelectedRecords: to get selectedRecords

Before it was mixed and when we wanted to get the selectedRecords, we were sending a call to the backend 🤦 to generate the merge preview => 4 calls instead of 1

packages/twenty-server/src/engine/api/common/common-query-runners/common-merge-many-query-runner.service.ts
```diff
@@ -95,7 +95,7 @@ export class CommonMergeManyQueryRunnerService extends CommonBaseQueryRunnerServ
     });
 
     await queryBuilder
-      .softDelete()
+      .delete()
```
charlesBochet, line 98: Not related to the original bug but discovered it while testing:

Issue was with the following case
(1) Mark Young, email empty
(2) Mark Young, mark@example.com

When we merge (prioirty on 1), we:
- soft delete (2), mark@example.com
- update (1) mark@example.com
 
So we end up with (softDelete or not) two records with same emails.

Discussed with Thomas, let's destroy the old one

charlesBochet (approved):

PR #15901, opened 2025-11-18
Update lock parameters

martmull:
as title

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/cache-lock/cache-lock.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/cache-lock/cache-lock.service.ts
@@ -26,7 +26,7 @@ export class CacheLockService {
     key: string,
     options?: CacheLockOptions,
   ): Promise<T> {
-    const { ms = 50, maxRetries = 20, ttl = 500 } = options || {};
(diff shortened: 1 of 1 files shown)


charlesBochet (approved):

packages/twenty-server/src/engine/core-modules/cache-lock/cache-lock.service.ts
```diff
@@ -26,7 +26,7 @@ export class CacheLockService {
     key: string,
     options?: CacheLockOptions,
   ): Promise<T> {
-    const { ms = 50, maxRetries = 40, ttl = 10_000 } = options || {};
+    const { ms = 100, maxRetries = 50, ttl = 5_500 } = options || {};
```
Weiko, line 29: Your issue seems related to workflow, shouldn't we specifically update the withLock decorator option in this specific case?

Weiko: @martmull

martmull: Today it is only used in workflow and it will be mostly use to solve concurrency db updates so I guess updating default values makes sense

PR #15902, opened 2025-11-18
Removed the unnecessary imports of console in codebase

araj00:
As console is a built in funtion in javascript that it was not at all required to import the console in codebase. That's why, did some cleaning on it.


charlesBochet (approved):

charlesBochet:
@araj00 thanks for opening a PR, makes sense but in this case the linter is complaining about console.log now, we should add a comment to disable linter on these specific lines

charlesBochet:
Closing for now, let's re-open with feedbacks taken into account :)

PR #15911, opened 2025-11-18
Metadata api create entity in workspace custom app

prastoin:
# Introduction
Cleaner and fewer scope version of https://github.com/twentyhq/twenty/pull/15745 ( removed sync-metadata hack through, too ambitious migration and upgrade )

Please note that this PR won't have any interaction with the existing sync-metadata
Which mean that the sync metadata does not update the standard entities applicationId and universalIdentifier, and it won't we will deprecate it on favor of a workspace migration aka twenty-standard app installation

## API Metadata
Any operation going through the api metadata nows automatically scope the related entity to the workspace custom application instance. ( optionally passing an applicationId to allow current hacky implem of app sync service )

We need to either ignore the tests or remove the cli status check from the blocking status badges for a PR to be merged

## New workspace
Already handled in previous https://github.com/twentyhq/twenty/pull/15625, when a workspace is created it gets created a twenty standard and custom workspace instance

All his views and permissions will be prefilled to the its twenty standard app instance with a specific universalIdentifier

## New universalIdentifier
At the contrary as before with standardIds, universalIdentifier are unique for a given workspace
This means that createdAt field of both object company and opportunity will have a unique universalIdentifier whereas they share the same standardId

## FlatApplication
Introduced the flatApplication and cache. Will migrate existing `MetadataName` to be `SyncableMetadataName` in a following PR

## What's next
Next we will describe a twenty standard app configuration as json that will be used to generate a workspace migration that will be run instead of the sync metadata, in a nutshell we aim to deprecated the sync metadata
So we can standardize any entity to have a non nullable applicationId and universalIdentifier

## Upgrade command
Introduced an upgrade command that will create a custom workspace instance for any workspace that do not have one in order to align with the new behavior when creating a new workspace


packages/twenty-server/src/engine/core-modules/application/application.service.ts
```diff
@@ -118,7 +220,13 @@ export class ApplicationService {
...
     }
 
-    return this.applicationRepository.save(application);
+    const savedApplication = this.applicationRepository.save(application);
+
+    await this.workspaceFlatApplicationMapCacheService.invalidateCache({
+      workspaceId: data.workspaceId,
+    });
+
+    return savedApplication;
```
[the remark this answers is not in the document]
prastoin: Nice catch

prastoin:
There's seems to be an issue with the build deps order of cli e2e tests. Nothing critical
Might give this a look in this PR but seems unrelated to it

prastoin:
@greptileai trigger

packages/twenty-server/src/database/commands/upgrade-version-command/1-10/1-10-seed-dashboard-view.command.ts
```diff
@@ -59,7 +61,17 @@ export class SeedDashboardViewCommand extends ActiveOrSuspendedWorkspacesMigrati
       );
     }
 
-    const views = [dashboardsAllView([dashboardObjectMetadata], true)];
+    const { twentyStandardFlatApplication } =
+      await this.applicationService.findWorkspaceTwentyStandardAndCustomApplicationOrThrow(
```
charlesBochet, line 65: **polish:** honestly, as we don't support multi version upgrade yet, I would just remove 1-10/1-11 commands (I won't QA them on my side)

packages/twenty-server/src/database/commands/upgrade-version-command/1-12/1-12-set-standard-application-not-uninstallable.command.ts
charlesBochet, line 29: **question:** isn't it easier to only have one 1-12 command as it's not released yet?

packages/twenty-server/src/engine/core-modules/application/constants/standard-object.constant.ts
```diff
@@ -0,0 +1,1919 @@
...
+  WORKFLOW_STANDARD_FIELD_IDS,
+  WORKFLOW_VERSION_STANDARD_FIELD_IDS,
+  WORKSPACE_MEMBER_STANDARD_FIELD_IDS,
+} from 'src/engine/workspace-manager/workspace-sync-metadata/constants/standard-field-ids';
+
+export const STANDARD_OBJECTS = {
+  attachment: {
+    universalIdentifier: STANDARD_OBJECT_IDS.attachment,
+    fields: {
+      id: { universalIdentifier: '20202020-a01a-4001-8a01-1d5f8e3c7b2a' },
```
charlesBochet, line 45: **comment**: universalIdentifiers, the start of an era!

**polish:** this file is too big, we should cut it in files in a subfolder

packages/twenty-server/src/engine/core-modules/application/services/workspace-flat-application-map-cache.service.ts
```diff
@@ -0,0 +1,60 @@
...
+import { CacheStorageNamespace } from 'src/engine/core-modules/cache-storage/types/cache-storage-namespace.enum';
+import { MetadataToFlatEntityMapsKey } from 'src/engine/metadata-modules/flat-entity/types/metadata-to-flat-entity-maps-key';
+import { WorkspaceFlatMapCache } from 'src/engine/workspace-flat-map-cache/decorators/workspace-flat-map-cache.decorator';
+import { WorkspaceFlatMapCacheService } from 'src/engine/workspace-flat-map-cache/services/workspace-flat-map-cache.service';
+
+@Injectable()
+@WorkspaceFlatMapCache(
+  'flatApplicationMaps' as MetadataToFlatEntityMapsKey<AllMetadataName>,
+) // TODO prastoin introduce SyncableMetadata notion
+export class WorkspaceFlatApplicationMapCacheService extends WorkspaceFlatMapCacheService<FlatApplicationCacheMaps> {
```
charlesBochet, line 21: discussed with Coco, we were not sure application themselves were syncable or would need Flat tooling but why not TBH

packages/twenty-server/src/database/commands/upgrade-version-command/1-12/1-12-set-standard-application-not-uninstallable.command.ts
prastoin: It's for new workspaces unless I'm mistaken

packages/twenty-server/src/engine/metadata-modules/cron-trigger/services/cron-trigger-v2.service.ts
```diff
@@ -43,6 +57,8 @@ export class CronTriggerV2Service {
       {
         createCronTriggerInput: cronTriggerInput,
         workspaceId,
+        workspaceCustomApplicationId:
```
charlesBochet, line 60: **question/change:** this feels wrong: we don't always pass the customApplicationId to customApplicationId

packages/twenty-server/src/engine/metadata-modules/cron-trigger/utils/from-create-cron-trigger-input-to-flat-cron-trigger.util.ts
```diff
@@ -6,18 +6,21 @@ import { type FlatCronTrigger } from 'src/engine/metadata-modules/cron-trigger/t
...
+  workspaceCustomApplicationId: string;
 }): FlatCronTrigger => {
   const now = new Date();
+  const id = v4();
 
   return {
-    id: v4(),
-    universalIdentifier: createCronTriggerInput.universalIdentifier ?? v4(),
+    id,
+    universalIdentifier: createCronTriggerInput.universalIdentifier ?? id,
```
charlesBochet, line 20: **question/change:** id and universalIdentifier should not necessarily match: looks wrong too? or is it our hypothesis for workspaceCustomApplicationId? actually it would be a nice idea, not sure I'm getting all the impacts of this decision

packages/twenty-server/src/engine/metadata-modules/database-event-trigger/services/database-event-trigger-v2.service.ts
```diff
@@ -47,6 +61,8 @@ export class DatabaseEventTriggerV2Service {
       fromCreateDatabaseEventTriggerInputToFlatDatabaseEventTrigger({
         createDatabaseEventTriggerInput: databaseEventTriggerInput,
         workspaceId,
+        workspaceCustomApplicationId:
```
charlesBochet, line 64: same here

packages/twenty-server/src/engine/metadata-modules/database-event-trigger/utils/from-create-database-event-trigger-input-to-flat-database-event-trigger.util.ts
```diff
@@ -6,21 +6,24 @@ import { type FlatDatabaseEventTrigger } from 'src/engine/metadata-modules/datab
...
   createDatabaseEventTriggerInput,
   workspaceId,
+  workspaceCustomApplicationId,
 }: {
   createDatabaseEventTriggerInput: CreateDatabaseEventTriggerInput;
   workspaceId: string;
+  workspaceCustomApplicationId: string;
 }): FlatDatabaseEventTrigger => {
   const now = new Date();
+  const id = uuidV4();
```
charlesBochet, line 16: same + use v4()

packages/twenty-server/src/database/commands/upgrade-version-command/1-12/1-12-set-standard-application-not-uninstallable.command.ts
charlesBochet: I don't get it, 1-12 is not released yet!

packages/twenty-server/src/engine/metadata-modules/cron-trigger/services/cron-trigger-v2.service.ts
charlesBochet: got it, it's to be deprecated, right?

packages/twenty-server/src/engine/metadata-modules/cron-trigger/utils/from-create-cron-trigger-input-to-flat-cron-trigger.util.ts
charlesBochet: why not!

packages/twenty-server/src/database/commands/upgrade-version-command/1-12/1-12-set-standard-application-not-uninstallable.command.ts
prastoin: There's already existing workspace with both standard and custom applications ( new workspace in 1.11 have both )

charlesBochet (approved):
LGTM!

packages/twenty-server/src/engine/metadata-modules/cron-trigger/utils/from-create-cron-trigger-input-to-flat-cron-trigger.util.ts
prastoin: I'm not sure either, in the first place I though might have been interesting to keep this consistency but I'm not sure of the benefits versus the risk it brings to the table

packages/twenty-server/src/engine/metadata-modules/database-event-trigger/services/database-event-trigger-v2.service.ts
prastoin: I though I've added jsdocs deprecated to this though

PR #15914, opened 2025-11-18
Agent role edition

FelixMalfait:
### 1. Role Editing Not Working for New Agents
When creating a new agent and then creating a role for it, the role permissions appeared as non-editable. This was caused by:
- In create mode, `agentId = ''` (empty string from `useParams`)
- Empty string is **falsy** but `isDefined('')` returns **true**
- This caused incorrect behavior in role exclusivity checks and API calls

### 2. Missing Role Data Loading
The agent form wasn't loading role data needed for permission editing because it was missing the `SettingsRolesQueryEffect` component.

### 3. Navigation Conflicts
The `useSaveDraftRoleToDB` hook contained navigation logic that caused conflicts when used from different contexts (role detail page vs agent form).

### 4. Linting Errors
Unused `useIcons` import in `SettingsAgentRoleTab.tsx`.

---

### Agent Role Tab (`SettingsAgentRoleTab.tsx`)
- ✅ Use `isNonEmptyString(agentId)` to validate agentId (follows codebase patterns)
- ✅ Improved role exclusivity logic to handle both create and edit modes:
  - **Edit mode**: Role must be assigned exclusively to this agent
  - **Create mode**: Role must not be assigned to anyone yet
- ✅ Only call `assignRoleToAgent` when a valid agentId exists
- ✅ Pass `undefined` instead of empty string for `fromAgentId` prop

### Role Hook (`useSaveDraftRoleToDB.ts`)
- ✅ Removed navigation logic from the hook (better separation of concerns)
- ✅ Removed `useNavigateSettings` and `SettingsPath` dependencies
- ✅ Hook now only handles data persistence, not navigation

### Role Component (`SettingsRole.tsx`)
- ✅ Added navigation after successful role creation in create mode
- ✅ Navigation now handled by the component using the hook

### Agent Form (`SettingsAgentForm.tsx`)
- ✅ Added `SettingsRolesQueryEffect` to ensure role data is loaded
- ✅ Agent form handles its own navigation after save

---

## ✅ Testing Scenarios

All these scenarios now work correctly:
1. ✅ Create new agent → Create role → Edit permissions → Save agent
2. ✅ Edit existing agent → Create role → Edit permissions → Save
3. ✅ Edit existing agent → Try to edit shared role (shows warning message)
4. ✅ Navigate to object-level permissions from agent form with proper breadcrumbs

---

### Root Cause Analysis
The main issue was that empty string was being treated differently in different checks:
- `agentId &&` → evaluates to `false` (empty string is falsy)
- `isDefined(agentId)` → returns `true` (empty string is defined)

This inconsistency caused the role to appear as non-editable and attempted invalid API calls.

### Solution
Used `isNonEmptyString()` from `@sniptt/guards` which properly checks both that the value is defined AND not empty, following codebase conventions.

---

## 🎯 Impact

- Fixes critical bug preventing role permission configuration for new agents
- Improves code quality with better separation of concerns
- Makes navigation logic more predictable and maintainable
- Follows codebase patterns and guidelines

### DIFF
--- a/packages/twenty-front/src/modules/settings/roles/components/SettingsRolesList.tsx
+++ b/packages/twenty-front/src/modules/settings/roles/components/SettingsRolesList.tsx
@@ -1,19 +1,27 @@
+import { Dropdown } from '@/ui/layout/dropdown/components/Dropdown';
+import { DropdownContent } from '@/ui/layout/dropdown/components/DropdownContent';
+import { DropdownMenuItemsContainer } from '@/ui/layout/dropdown/components/DropdownMenuItemsContainer';
 import { Table } from '@/ui/layout/table/components/Table';
 import styled from '@emotion/styled';
 import { t } from '@lingui/core/macro';
 
 import { SettingsRolesTableHeader } from '@/settings/roles/components/SettingsRolesTableHeader';
 import { SettingsRolesTableRow } from '@/settings/roles/components/SettingsRolesTableRow';
-import { ROLES_LIST_TABS } from '@/settings/roles/constants/RolesListTabs';
 import { settingsAllRolesSelector } from '@/settings/roles/states/settingsAllRolesSelector';
-import { activeTabIdComponentState } from '@/ui/layout/tab-list/states/activeTabIdComponentState';
 import { TableCell } from '@/ui/layout/table/components/TableCell';
-import { useRecoilComponentValue } from '@/ui/utilities/state/component-state/hooks/useRecoilComponentValue';
 import { useRecoilValue } from 'recoil';
 import { SettingsPath } from 'twenty-shared/types';
-import { H2Title, IconPlus, IconSearch } from 'twenty-ui/display';
+import {
+  H2Title,
+  IconPlus,
+  IconSearch,
+  IconFilter,
+  IconRobot,
+  IconKey,
+} from 'twenty-ui/display';
 import { Button } from 'twenty-ui/input';
 import { Section } from 'twenty-ui/layout';
+import { MenuItemToggle } from 'twenty-ui/navigation';
 import { useNavigateSettings } from '~/hooks/useNavigateSettings';
 import { sortByAscString } from '~/utils/array/sortByAscString';
 import { SettingsTextInput } from '@/ui/input/components/SettingsTextInput';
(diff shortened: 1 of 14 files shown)


packages/twenty-front/src/modules/settings/roles/components/SettingsRolesList.tsx
```diff
@@ -57,52 +68,71 @@ export const SettingsRolesList = () => {
...
-        matchesTab = role.canBeAssignedToApiKeys;
-        break;
-      default:
-        matchesTab = role.canBeAssignedToUsers;
-    }
-
-    return (
-      matchesTab && role.label?.toLowerCase().includes(searchTerm.toLowerCase())
-    );
+    // Filter by role type (always show user roles)
```
Weiko: nit: those comments are probably not needed

FelixMalfait:
Spotted issue on main: https://github.com/twentyhq/twenty/issues/15918 ; couldn't resolve it quickly

packages/twenty-front/src/pages/settings/roles/SettingsRoleAddObjectLevel.tsx
```diff
@@ -4,27 +4,54 @@ import { SettingsRolePermissionsObjectLevelObjectPicker } from '@/settings/roles
...
 export const SettingsRoleAddObjectLevel = () => {
   const { roleId } = useParams();
+  const [searchParams] = useSearchParams();
+  const fromAgentId = searchParams.get('fromAgent');
+
   const settingsDraftRole = useRecoilValue(
     settingsDraftRoleFamilyState(roleId ?? ''),
   );
 
+  const { data: agentData } = useFindOneAgentQuery({
```
Weiko, line 22: Do we use a recoil state already for that?

Weiko (approved):
Left some nit comments, otherwise LGTM!

packages/twenty-front/src/modules/settings/roles/components/SettingsRolesList.tsx
FelixMalfait: My bad I forgot to push the last commit

packages/twenty-front/src/pages/settings/roles/SettingsRoleAddObjectLevel.tsx
FelixMalfait: I think maybe in a refactor we could push the breadcrumb title or something like that yes. Otherwise here we would need to fetch api key name and application name also in the future

PR #15927, opened 2025-11-19
Migrate role to migration v2

Weiko:


Weiko:
To reopen later, after discussing application role scope

PR #15931, opened 2025-11-19
Fix raw json validation

etiennejouan:
Before validation in common layer, stringified json were accepted in raw json field. Fix regression

related to https://discord.com/channels/1130383047699738754/1130383048173682821/1440723288090218636


Weiko (approved):
LGTM!

Weiko (approved):
After discussion, formatData already does the transform so we only want to update the validate logic to accept string as input. LGTM! 👏

PR #15935, opened 2025-11-19
Access to system object in Workflows Search Node

Bonapara:
CleanShot 2025-11-19 at 19 24 01


charlesBochet (changes requested):
@Bonapara 

Feedbacks before reviewing further:
- duplicated code, the dropdown should be extracted to its own component
- non respecting code standards (non usage of isNonEmptyString
- non justified usages of useMemo (legit usages are quite rare IMO)
- listening to states (useRecoilValue) to use them in sync callbacks (in this case we should use useRecoilCallback to avoid re-rendering for no reason)
- variable naming can be improved: "regularObject"?

PR #15937, opened 2025-11-19
Updated by extension

BOHEUS:
Working updated by extension which shows the workspace member behind the newest change


charlesBochet (changes requested):
@BOHEUS I don't think this scales well:

For each record updated:
- we fetchAllObjects which is already heavy
- we findObjectName which will trigger n queries

I feel you should:
- have two different serverless function: 1 for setup that should be triggered once and explain in your doc that this serverless function has to be run and another to set the updatedBy value

Also, let's make clear when we deal with objectMetadata / fieldMetadata and when we deal with objectRecord / fieldValue: in your code there is a confusion between updateUpdatedByField (record world) and createUpdatedByField (metadata world)

BOHEUS:
@charlesBochet what should be the correct flow in case user adds another object after running setup function? Trigger it again? About the trigger itself, I think I'll add a route trigger

I'm aware that fetchAllObjects and findObjectName are heavy, I can also wait until @martmull exposes the object name in properties so those 2 functions can be removed but don't think that's a good idea given that more and more people are asking about the extension

martmull:
Hey @BOHEUS thank you for your PR. I have tested it and works nice for UI updates. Not taking APIKEY updates into account but that is a nice workaround waiting the core UPDATE_BY Actor field implementation

Lets fix @charlesBochet comments:

Now you should have the objectMetadata in the updatedBy event sent to the serverlessFunction. Look for `params.objectMetadata.nameSingular`

BOHEUS:
Thanks a lot @martmull! Now the app can be even more optimized, I'm starting the work on it, today I should be able to push fixed version

charlesBochet:
@BOHEUS other idea:
you try to make the updatedBy value update and if it fails then you add the missing field

(optimistic behavior)

BOHEUS:
@charlesBochet done

packages/twenty-apps/updated-by/serverlessFunctions/updated/src/index.ts
```diff
@@ -0,0 +1,237 @@
...
+    },
+    url: `${TWENTY_API_URL}/rest/metadata/objects`,
+  };
+  const response = await axios.request(options);
+  return response.data.data.objects.find(
+    (object: any) => object.namePlural === 'workspaceMembers',
+  ).id;
+};
+
+const createUpdatedByField = async (
```
charlesBochet, line 53: createUpdateByFieldMetadata

packages/twenty-apps/updated-by/serverlessFunctions/updated/src/index.ts
```diff
@@ -0,0 +1,237 @@
...
+    const response = await axios.request(options);
+    return response.status === 200;
+  } catch (error) {
+    if (axios.isAxiosError(error)) {
+      throw error;
+    }
+  }
+};
+
+const updateUpdatedByField = async (
```
charlesBochet, line 127: updateUpdatedByFieldValue

packages/twenty-apps/updated-by/serverlessFunctions/updated/src/index.ts
```diff
@@ -0,0 +1,237 @@
...
+  } catch (error) {
+    if (axios.isAxiosError(error)) {
+      throw error;
+    }
+  }
+};
+
+export const main = async (params: {
+  properties: Record<string, any>;
+  objectMetadata: Record<string, any>;
```
charlesBochet, line 155: oh so you actually have the objectMetadata, that's even better

charlesBochet (approved):
left polish comments, LGTM

PR #15938, opened 2025-11-19
Handle Microsoft calendar sync cursor error

neo773:
Resolves https://twenty-v7.sentry.io/issues/6567295627


charlesBochet (approved):
LGTM but I'm still not happy with the error handling between calendar-list-fetch / calendar-import to be mixed

PR #15949, opened 2025-11-20
Fix wrong empty string formatting

### LINKED ISSUE #15912: [Workflows] - invalid input syntax for type numeric: ""
## Bug Description

If you try to propage the values of numeric fields from one field to another numeric field, the workflow cannot run (seeing the error `invalid input syntax for type numeric: ""` when the values of the first field is null. 

**Example:** 
Image
Image
Image
Image

## Expected behavior

Workflow manages the null values. If not, that means that users must create code nodes in between to test if the value is not null (which cannot be our desired experience)

I've not checked the other type but I'm assuming a similar issue might arise.
not sure if this a bug or feature, cc @martmull  in case this is something you can unblock 
thank you!

martmull:
As title, solves this kind of issues https://twentyfortwenty.twenty.com/object/workflowRun/d6aca50e-68ba-4715-8835-ea22bd44fb88

Solves null currencyDisplay when null currencyCode and null amountMicros

### Before
image

 ### After
 
image

### DIFF
--- a/packages/twenty-front/src/modules/ui/field/display/components/CurrencyDisplay.tsx
+++ b/packages/twenty-front/src/modules/ui/field/display/components/CurrencyDisplay.tsx
@@ -23,8 +23,6 @@ export const CurrencyDisplay = ({
 }: CurrencyDisplayProps) => {
   const theme = useTheme();
 
-  const shouldDisplayCurrency = isDefined(currencyValue?.currencyCode);
-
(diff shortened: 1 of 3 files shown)


FelixMalfait (approved):
Approved but if currencyCode was null it should have not displayed anything already no?

martmull:
> Approved but if currencyCode was null it should have not displayed anything already no?

I don't know why, but no, if currencyCode was null or undefined, it displayed "0"
image

PR #15951, opened 2025-11-20
[groupBy] groupBy relation fields

ijreilly:
Example query
Here person has 
- a N - 1 relationship with company
- a N - 1 morph relationship with pet or company

image
image

truncated response
image

We don't allow grouping by relations of relations.

Left to do
- rest api
- tests on permissions


packages/twenty-server/src/engine/api/graphql/workspace-schema-builder/graphql-type-generators/input-types/relation-field-metadata-gql-type.generator.ts
```diff
@@ -143,32 +149,89 @@ export class RelationFieldMetadataGqlInputTypeGenerator {
...
+
+    if (
+      isDefined(fieldMetadata.relationTargetObjectMetadataId) &&
+      isDefined(objectMetadataMaps)
+    ) {
+      const targetObjectMetadata =
+        objectMetadataMaps.byId[fieldMetadata.relationTargetObjectMetadataId];
+
+      if (isDefined(targetObjectMetadata)) {
+        const targetGroupByInputTypeKey = computeObjectMetadataInputTypeKey(
```
ijreilly, line 213: I need to double check that I don't have issues with this: I thought it could happen that `targetGroupByInputType` has not been created yet, but I did not run into the issue. `fields` of the parent type are defined as a callback function  (of which this code belongs) so maybe they are always defined after all groupByInputTypes have been declared.

etiennejouan: I need to review more carefully but you can give a look to packages/twenty-server/src/engine/api/graphql/workspace-schema-builder/graphql-type-generators/object-types/object-metadata-with-relations-gql-object-type.generator.ts which has the same constraint.

ijreilly: Thanks @etiennejouan 
So indeed GraphQLInputObjectType is created with a lazy fields function (as i clumsily suggested in my first comment). so first all types are stored in gqlTypesStorage, but generateFields() is not called until printSchema() is called.

etiennejouan: Ok but I still not understand how you can prevent from fetching a type not generated yet. Or, the object with relations object generating step is also useless ?

packages/twenty-server/src/engine/api/common/common-query-runners/common-group-by-query-runner.service.ts
```diff
@@ -104,26 +108,45 @@ export class CommonGroupByQueryRunnerService extends CommonBaseQueryRunnerServic
...
+
       const alias =
-        removeQuotes(columnNameWithQuotes) +
+        formatColumnNameAsAlias(columnNameWithQuotes) +
         (isGroupByDateField(groupByField)
           ? `_${groupByField.dateGranularity}`
-          : '');
+          : isGroupByRelationField(groupByField) && groupByField.dateGranularity
+            ? `_${groupByField.dateGranularity}`
+            : '');
```
etiennejouan: nit : can be simplify with just one ternary ? or done like this one purpose to highlight this case ?

packages/twenty-server/src/engine/api/common/common-query-runners/common-group-by-query-runner.service.ts
```diff
@@ -132,9 +155,11 @@ export class CommonGroupByQueryRunnerService extends CommonBaseQueryRunnerServic
...
         }),
         alias,
-        dateGranularity: isGroupByDateField(groupByField)
-          ? groupByField.dateGranularity
-          : undefined,
+        dateGranularity:
+          isGroupByDateField(groupByField) ||
+          isGroupByRelationField(groupByField)
+            ? groupByField.dateGranularity
+            : undefined,
```
etiennejouan: nit : and here, ternary useful ? it seems if it's not a groupByDate (Relation or not), it would return undefined. Or could have the same condition as in code commented above.

etiennejouan: or maybe the GroupByDateField type could be merge with the regular one (as we don't have the GroupByDateRelationField ?)

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/resolvers/utils/is-group-by-date-field.util.ts
```diff
@@ -1,10 +1,20 @@
...
   groupByField: GroupByField,
 ): groupByField is GroupByDateField => {
-  return 'dateGranularity' in groupByField;
+  return (
+    'dateGranularity' in groupByField &&
+    !('nestedFieldMetadata' in groupByField)
+  );
+};
+
+export const isGroupByRelationField = (
```
etiennejouan: nit : I would create a new file (sorry for this greptile like nit)

etiennejouan (approved):
Nice ! Very powerful 👏 

Not sure to understand why the schema generation works but I'll dig for my knowledge

charlesBochet:
@etiennejouan  could you take this review? :)

packages/twenty-server/src/engine/api/common/common-query-runners/common-group-by-query-runner.service.ts
```diff
@@ -160,6 +188,12 @@ export class CommonGroupByQueryRunnerService extends CommonBaseQueryRunnerServic
       const queryBuilderWithFiltersAndWithoutGroupBy =
```
Weiko: Should we instead clone the builder before applying groupBy to the common query builder? This would avoid having to parse things twice I think

packages/twenty-server/src/engine/api/common/common-query-runners/common-group-by-query-runner.service.ts
```diff
@@ -104,26 +109,47 @@ export class CommonGroupByQueryRunnerService extends CommonBaseQueryRunnerServic
...
 
+    const objectAlias = getObjectAlias(objectMetadataItemWithFieldMaps);
+
+    this.addJoinForGroupByOnRelationFields({
+      queryBuilder,
+      groupByFields,
+      objectAlias,
+    });
+
     const groupByDefinitions = groupByFields.map((groupByField) => {
```
Weiko: Should we extract this whole work to a sub method

packages/twenty-server/src/engine/api/common/common-query-runners/common-group-by-query-runner.service.ts
```diff
@@ -343,6 +377,47 @@ export class CommonGroupByQueryRunnerService extends CommonBaseQueryRunnerServic
...
+
+    for (const groupByField of groupByFields) {
+      if (isGroupByRelationField(groupByField)) {
+        const joinAlias = groupByField.fieldMetadata.name;
+
+        if (
+          !groupByField.fieldMetadata.settings ||
+          !isFieldMetadataRelationOrMorphRelation(groupByField.fieldMetadata)
+        ) {
+          throw new Error(
```
Weiko: Should we throw a common api exception instead?

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/resolvers/utils/parse-group-by-args.util.ts
```diff
@@ -69,6 +84,125 @@ export const parseGroupByArgs = (
...
+        const targetObjectMetadata = getTargetObjectMetadataOrThrow(
+          fieldMetadata,
+          objectMetadataMaps,
+        );
+
+        const nestedFieldNames = Object.keys(nestedFieldGroupByDefinitions);
+
+        if (nestedFieldNames.length > 1) {
+          throw new GraphqlQueryRunnerException(
+            'You cannot provide multiple nested fields in one relation GroupByInput, split them into multiple GroupByInput',
```
Weiko: I can see this one being raised at multiple places, maybe we want a util for that? (validateSingleKey for example)

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/resolvers/utils/parse-group-by-args.util.ts
```diff
@@ -69,6 +84,125 @@ export const parseGroupByArgs = (
         throw new Error(`Unidentified field in groupBy: ${fieldName}`);
       }
 
+      const isGroupByRelationField =
```
Weiko, line 41: this util is deeply nested (seems to reach 6 depth at some point) which makes it a bit hard to read, can we extract the logic in some sub methods?

packages/twenty-server/src/engine/api/graphql/workspace-schema-builder/graphql-type-generators/input-types/relation-field-metadata-gql-type.generator.ts
```diff
@@ -143,32 +149,89 @@ export class RelationFieldMetadataGqlInputTypeGenerator {
...
+    const fields: GraphQLInputFieldConfigMap = {};
+
+    if (
+      isDefined(fieldMetadata.relationTargetObjectMetadataId) &&
+      isDefined(objectMetadataMaps)
+    ) {
+      const targetObjectMetadata =
+        objectMetadataMaps.byId[fieldMetadata.relationTargetObjectMetadataId];
+
+      if (isDefined(targetObjectMetadata)) {
```
etiennejouan, line 212: With this implementation, we enable unlimited depth in schema definition.

Opportunity > pointOfContact (person) > company. Not a big deal if we add validation in common (needed when rest side will be done) but misleading for gql users

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
```diff
@@ -156,8 +157,10 @@ export class GroupByWithRecordsService {
...
     );
 
+    const objectAlias = getObjectAlias(objectMetadataItemWithFieldMaps);
+
     const recordSelectWithAlias = Object.keys(columnsToSelect)
-      .map((col) => `"${col}" as "${SUB_QUERY_PREFIX}${col}"`)
+      .map((col) => `"${objectAlias}"."${col}" as "${SUB_QUERY_PREFIX}${col}"`)
       .join(', ');
 
     const groupBySelectWithAlias = groupByDefinitions
```
Weiko, line 166:
```
`ROW_NUMBER() OVER (PARTITION BY ${groupByExpressions})`, 'rn'
```
I'm pretty sure this will be applied to all matching records then once it's in sql memory we will apply your filter logic with JSON_AGG to filter out once RECORDS_PER_GROUP_LIMIT is reached which will be "too late". I think we should leverage SQL for that part and see if we can improve performances even more

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
```diff
@@ -156,8 +157,10 @@ export class GroupByWithRecordsService {
       queryBuilderForSubQuery,
     );
```
Weiko, line 159: L94 `if (isDefined(selectedFieldsResult.relations))` are we sure this is not always defined? Should we check if it's empty as well to avoid unnecessary work?

etiennejouan (commented):
Left one comment on schema generation with a potential issue (it depends if we tolerate it) on nested group-by input.
Permissions update looks good

packages/twenty-server/src/engine/twenty-orm/repository/permissions.utils.ts
```diff
@@ -259,6 +273,63 @@ export const validateQueryIsPermittedOrThrow = ({
...
+  );
+
+  const indexesOfSelectsForJoinedAlias: number[] = [];
+
+  for (const [_index, joinedAlias] of joinAttributesAliases.entries()) {
+    const entity = expressionMap.aliases.find(
+      (alias) => alias.type === 'join' && alias.name === joinedAlias,
+    )?.metadata;
+
+    if (entity) {
```
Weiko: nit: isDefined

packages/twenty-server/src/engine/twenty-orm/repository/permissions.utils.ts
```diff
@@ -259,6 +273,63 @@ export const validateQueryIsPermittedOrThrow = ({
...
+      for (const [index, select] of expressionMap.selects.entries()) {
+        const regex = /"(\w+)"\."(\w+)"/;
+        const extractedAlias = select.selection.match(regex)?.[1]; // "person"."name" -> "person"
+
+        if (extractedAlias === joinedAlias) {
```
Weiko: should we throw if extractedAlias failed to be extracted and is undefined? I don't think joinedAlias can be undefined so it's probably fine, up to you

packages/twenty-server/src/engine/twenty-orm/repository/permissions.utils.ts
```diff
@@ -259,6 +273,63 @@ export const validateQueryIsPermittedOrThrow = ({
...
+  objectsPermissions: ObjectsPermissions;
+  objectMetadataMaps: ObjectMetadataMaps;
+}) => {
+  const joinAttributesAliases = new Set(
+    expressionMap.joinAttributes.map((join) => join.alias.name),
+  );
+
+  const indexesOfSelectsForJoinedAlias: number[] = [];
+
+  for (const [_index, joinedAlias] of joinAttributesAliases.entries()) {
```
Weiko, line 291: Could use 2 maps instead of nested for loop (O(nm) => O(n+m)) for better performances

Weiko (commented):
Left some comments, some are not related to your work so we can work on it in a later PR, up to you

packages/twenty-server/src/engine/api/common/common-query-runners/common-group-by-query-runner.service.ts
ijreilly: not really useful but "needed" for type as dateGranularity is only defined on date field and target field (from relations) that are date fields. 
I added isGroupByDateFieldOrTargetField for clarity

packages/twenty-server/src/engine/api/graphql/workspace-schema-builder/graphql-type-generators/input-types/relation-field-metadata-gql-type.generator.ts
ijreilly: true!
for now I have just prevented depth 2 (and further) in parse-group-by-args.util #L132:
```
        if (nestedFieldMetadata.type === FieldMetadataType.RELATION) {
          throw new UserInputError(
            `Cannot group by a relation field of the relation field: "${nestedFieldName}" is a relation field of "${targetObjectMetadata.nameSingular}"`,
          );
        }
```
I will see if I find an easy way to remove them from the schema

packages/twenty-server/src/engine/api/common/common-query-runners/common-group-by-query-runner.service.ts
ijreilly: yes ! we can also do that for the filters. I hope there was no good reason I did not do it before when i first implemented the group by with records. will run tests

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
ijreilly: indeed, always defined ! thanks

packages/twenty-server/src/engine/twenty-orm/repository/permissions.utils.ts
ijreilly: We shouldn't throw, select.selection could be equal to just "person" from what i ve seen, not for joins, but here we iterate over all selects to determine which are the ones for joins.
I will add a check only to do the comparison if extractedAlias is defined though.

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
ijreilly: Does the addition of this where clause fix things?

```
const mainQuery = mainQueryQueryBuilder
      .from(`(${subQuery.getQuery()})`, 'ranked_records')
      .setParameters(queryBuilderForSubQuery.expressionMap.parameters)
      .where('rn <= :recordsPerGroupLimit', {
        recordsPerGroupLimit: RECORDS_PER_GROUP_LIMIT,
      })
      .select(groupByAliases)
      .addSelect(
        `JSON_AGG(
        CASE WHEN rn <= ${RECORDS_PER_GROUP_LIMIT} THEN
          JSON_BUILD_OBJECT(
          ...
```

packages/twenty-server/src/engine/twenty-orm/repository/permissions.utils.ts
ijreilly: I need joinedAlias in the expressionMap.selects.entries() iteration so I dont think I can improve performances here

ijreilly: n (where n = joinedAlias) should be a low figure always

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
Weiko: It might help yes!

Weiko (approved):
LGTM

PR #15959, opened 2025-11-20
Add an option to choose to display the second tab as the first tab

### LINKED ISSUE #1801: Pending workflow run's flow is no longer focused when opening record in the side panel
https://github.com/user-attachments/assets/baff1738-14d1-45f1-8e6e-4d94c6da7680

Devessier:
https://github.com/user-attachments/assets/4e9cd2ba-5f1b-469d-8668-f085b3433838

Closes https://github.com/twentyhq/core-team-issues/issues/1801


bosiraphael (approved):
LGTM.
Still unsure about if this information should be stored at the pageLayout level with a boolean `shouldInvertFirstAndSecondTabOnMobile`. The pageLayoutTab approach that you implemented has the benefit to be easily modified to work with every tab as the first tab on mobile, but the other approach is simpler if we want to just be able to invert the first two tabs.
@Bonapara fyi

Devessier:
As discussed, I will move the property to the page layout.

Bonapara:
I have another question/idea. Instead of switching the order, should the setting be `Default tab focus on side panel/mobile`
(keep the same order but focus the second tab instead of the first one)? Might be cleaner?

PR #15962, opened 2025-11-20
Implement workspace cache storage

Weiko:
## Context
Implementing a single service managing all the cache scoped to a workspace, this will be dynamically injected as a WorkspaceContext in the app during a request lifetime (ingested by the future global datasource for example).

Usage:

```typescript
this.globalWorkspaceOrmManager.executeInWorkspaceContext(
        authContext,
        async () => {
           // Everything here will have access to a workspaceContext, containing all the cache data + a ready to use datasource with workspace scoped metadata, permissions, feature flags, etc...
           // Internally will call loadWorkspaceContext
        }
```
Note: executeInWorkspaceContext will probably be owned by a higher level service later and not only the ORM.

```typescript
  private async loadWorkspaceContext(
    authContext: WorkspaceAuthContext,
  ): Promise<WorkspaceContext> {
    const workspaceId = authContext.workspace.id;

    const cache =
      await this.workspaceContextCacheService.get<WorkspaceContextData>(
        workspaceId,
        [
          'objectMetadataMaps',
          'metadataVersion',
          'featureFlagsMap',
          'permissionsPerRoleId',
        ],
      );

    return {
      authContext,
      objectMetadataMaps: cache.objectMetadataMaps,
      metadataVersion: cache.metadataVersion,
      featureFlagsMap: cache.featureFlagsMap,
      permissionsPerRoleId: cache.permissionsPerRoleId,
    };
  }
  ```
  
  The cache retrieval strategy is as followed:
```
- Check if there is an ongoing promise fetching data from the cache => return the promise.
- Check in the local cache entry if lastCheckedAt has expired. If not, return as it is without querying redis.
- Check in redis the cache entry hash and compare with local cache entry hash, if they are the same return the local cache entry data 
- Check in redis the cache entry data, if it's there return it and store it into the local cache entry data and update local cache entry hash. If it's not there recompute the data by querying the DB and update both redis and local cache
```

### CODE REVIEW

File: packages/twenty-server/src/engine/workspace-context-cache/services/workspace-context-cache.service.ts [hunk 1]
```diff
@@ -0,0 +1,350 @@
...
+    const redisEntries = computed.flatMap(({ data, provider }) => {
+      const hash = this.generateHash(data);
+
+      return [
+        { key: provider.dataCacheKey(workspaceId), value: data },
+        { key: provider.hashCacheKey(workspaceId), value: hash },
+      ];
+    });
+
+    await this.cacheStorage.mset(redisEntries);
```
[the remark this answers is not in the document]
Weiko: As of today, the default value is 7 days, not sure what we want to do with that with the new system.
There is no TTL with the local cache though which is a valid point

Weiko:
@greptileai trigger

packages/twenty-server/src/engine/workspace-context-cache/services/workspace-context-cache.service.ts
```diff
@@ -0,0 +1,350 @@
...
+        );
+
+        return { ...result, ...staleResults };
+      },
+    );
+
+    return result as T;
+  }
+
+  invalidate(workspaceId: string, cacheProviderNames?: string[]): void {
```
Weiko: The idea is to push toward a lazy loading instead of calling a "recompute" method like in the old model. However invalidate should probably be able to invalidate redis cache and not local cache only, I will change that in an upcoming PR

packages/twenty-server/src/engine/core-modules/cache-storage/services/cache-storage.service.ts
```diff
@@ -28,6 +28,51 @@ export class CacheStorageService {
...
+    }
+
+    return Promise.all(keys.map((k) => this.get<T>(k)));
+  }
+
+  async mset<T = unknown>(
+    entries: Array<{ key: string; value: T; ttl?: Milliseconds }>,
+  ): Promise<void> {
+    if (this.isRedisCache()) {
+      const pipeline = (this.cache as RedisCache).store.client.multi();
```
Weiko, line 55: According to redis doc
```
Marks the start of a [transaction](https://redis.io/develop/interact/transactions) block. Subsequent commands will be queued for atomic execution using [EXEC](https://redis.io/docs/latest/commands/exec/).
``` 
The issue here is we would prefer using mSet (this one actually exists) but it does not allow us to set ttl apparently so we would need to do multiple requests anyway

etiennejouan (commented):
Nothing to add 🤓  Design looks good good to me, thanks for the concepts have just discovered 🙏

Weiko (commented):
Followups: 
- Deprecate objectMetadataCache and metadata version in the cache and use the new workspace cache system for objectmetadata, permissions and feature flags.
- Add pre-computed entitySchemas to Cache AND WorkspaceContext for global datasource
- Improve global datasource ORM wrapper to expose wrapped methods (with workspace context) to interact with the ORM

Weiko:
@charlesBochet if you want to check again

packages/twenty-server/src/engine/workspace-context-cache/services/workspace-context-cache.service.ts
Weiko: We actually won't do that since we don't have pod stickiness otherwise this will actually overload the DB

packages/twenty-server/src/engine/workspace-cache/decorators/workspace-cache.decorator.ts
```diff
@@ -0,0 +1,6 @@
+import { SetMetadata } from '@nestjs/common';
+
+export const WORKSPACE_CACHE_KEY = 'WORKSPACE_CACHE_KEY';
+
+export const WorkspaceCache = (workspaceCacheKey: string) =>
```
charlesBochet, line 5: workspaceCacheName to keep cacheKey for redis world?

charlesBochet: ok, read the code below, cacheKey seems more natural
ok then: cacheKey = engine:workspace:{workspaceId}:{workspaceCacheKey}:{hash|value}

packages/twenty-server/src/engine/workspace-cache/services/workspace-cache.service.ts
```diff
@@ -0,0 +1,365 @@
...
+    for (const workspaceCacheKey of validFromLocal) {
+      const localKey = this.getCacheKey(workspaceId, workspaceCacheKey);
+      const localEntry = this.localCache.get(localKey);
+
+      if (!isDefined(localEntry)) {
+        continue;
+      }
+
+      result[workspaceCacheKey] = localEntry.data;
+      this.updateLocalCache(
```
charlesBochet, line 166: I prefer this API where we pass everything explicitely

packages/twenty-server/src/engine/workspace-cache/services/workspace-cache.service.ts
```diff
@@ -0,0 +1,365 @@
...
+    workspaceId: string,
+    workspaceCacheKeys: string[],
+  ): Promise<Record<string, unknown>> {
+    const result: Record<string, unknown> = {};
+
+    const { validFromLocal, needsRedisCheck } =
+      await this.partitionKeysByLocalStaleness(workspaceId, workspaceCacheKeys);
+
+    for (const workspaceCacheKey of validFromLocal) {
+      const localKey = this.getCacheKey(workspaceId, workspaceCacheKey);
```
charlesBochet, line 158: not a fan of localKey (which is different from workspaceCacheKey and cacheKey as it's without the hash|data suffix
same for "entry"

I think it could make sense inside localCache but from the outside, I would not name it do things like:
const { localValue, localHash } = this.localCache.get(workspaceId, workspaceCacheKey)

charlesBochet (approved):
Looks way better!

Left a few more comments but I can understand what's happening more easily now :)

PR #15963, opened 2025-11-20
fix: crash in HeaderParser in dicer.

mabdullahabaid:
Resolves [Dependabot Alert 74](https://github.com/twentyhq/twenty/security/dependabot/74).

Upgraded `graphql-upload` from `13.0.0` to `16.0.2`. Type exports changed. API remains the same.

Tested the following upload flows:
- profile picture
- workspace logo
- rich text editor (image, video, file)
- record profile picture
- file associated to record.

They all work as intended, nothing breaks.


packages/twenty-server/src/engine/core-modules/file/file-upload/resolvers/file-upload.resolver.ts
```diff
@@ -1,10 +1,12 @@
 import { UseFilters, UseGuards, UsePipes } from '@nestjs/common';
 import { Args, Mutation, Resolver } from '@nestjs/graphql';
 
-import { FileUpload, GraphQLUpload } from 'graphql-upload';
+import GraphQLUpload from 'graphql-upload/GraphQLUpload.mjs';
```
charlesBochet, line 4: let's avoid this :) why do we need to import the mjs?

charlesBochet (changes requested):
see comment

packages/twenty-server/src/engine/core-modules/file/file-upload/resolvers/file-upload.resolver.ts
mabdullahabaid: Had a conversation with Felix about this and he said `.mjs` is fine. 

The reason for using .mjs:

Versions 15.x and 16.x of graphql-upload removed the root exports, so import { GraphQLUpload, FileUpload } from 'graphql-upload' no longer works. Instead, we must import from the explicit entrypoints with extensions (graphql-upload/GraphQLUpload.js, graphql-upload/processRequest.js, etc.).

For 16.x.x, they changed to .mjs extensions over .js.

image

image

I could not figure out a workaround for importing without extensions, so if you have an idea, that'd be great.

mabdullahabaid:
@charlesBochet replied. Looking into failing CI runs in the meantime.

charlesBochet:
Ok, the library maintainer seems to be the issue, LGTM, we can live wit that.
Let's merge once CI is fixed

mabdullahabaid:
@charlesBochet fixed.

charlesBochet (approved):

PR #15969, opened 2025-11-20
handle HTTP 410,404 in GmailMessagesImportErrorHandler

neo773:
Noticed this in prod, we currently return `undefined` instead of handling it.


neo773:
Can we disable auto review by Copilot? it's not helpful at all since it doesn't takes into account codebase context and guidelines like greptilebot does.

charlesBochet (approved):

PR #15985, opened 2025-11-21
Fix ORM event mixing records in batch updates

charlesBochet:
While investigating the issue a customer was facing, I discovered that before records and after records could be in different order, making the orm event and timeline activity engine mix records


Weiko (approved):
LGTM

PR #15995, opened 2025-11-21
Improve board records first load with group by query

lucasbordeau:
After : 

https://github.com/user-attachments/assets/36b7ef6e-70a5-4da6-a163-bc66ea5f7601

Capture d’écran 2025-11-21 à 17 15 06


charlesBochet:
@lucasbordeau I've merged the other PR, let's fix the conflict and re-open!

PR #16003, opened 2025-11-23
Replace agent handoff system with planning-based router

FelixMalfait:
## Overview

This PR replaces the dynamic agent handoff system with a more predictable planning-based router that decides upfront how to handle multi-agent coordination.

### 🔄 Architecture Shift: Handoffs → Planning

**Removed:**
- `AgentHandoffEntity` and handoff tracking system
- `AgentHandoffService` and `AgentHandoffExecutorService`
- Dynamic agent-to-agent transfers during execution
- Handoff tool generation and description templates

**Added:**
- `AiRouterService` with two strategies: `simple` (single agent) and `planned` (multi-agent)
- `AgentPlanExecutorService` for executing multi-step plans
- Plan validation (cycle detection, dependency resolution)
- `UnifiedRouterResult` type with discriminated union

### 🤖 New Standard Agents

Added two new specialized agents:
- **Researcher Agent**: Web search, fact-finding, competitive intelligence
- **Code Agent**: TypeScript function generation for serverless workflows

### 🏗️ Router Refactoring (Latest)

Split router responsibilities into focused services:
- `AiRouterStrategyDeciderService`: Decides simple vs planned strategy
- `AiRouterPlanGeneratorService`: Generates and validates execution plans
- `AiRouterService`: Coordinates between services (reduced from 426→275 lines)

### ⚙️ Configuration Improvements

- Added `outputStrategy` to agent definitions (`direct` vs `synthesize`)
- Removed hardcoded special cases for workflow-builder
- Added `plannerModel` field to workspace entity
- Increased `MAX_STEPS` from 10 to 25 for complex workflows

### 📝 Agent Prompt Refinements

Significantly simplified prompts for better clarity:
- Workflow Builder: 51→36 lines
- Helper: 49→28 lines
- Data Manipulator: Enhanced with sorting guidance

### 🔍 Enhanced Debugging

- Plan reasoning and step count in data message parts
- Router debug info with token usage tracking
- Better logging throughout execution pipeline

## Benefits

1. **Simpler Mental Model**: Router decides upfront vs dynamic transfers
2. **Better Predictability**: Users see the plan before execution
3. **Cleaner Architecture**: SRP with focused services
4. **Configuration Over Code**: Agent behavior via config, not hardcoded logic
5. **Plan Validation**: Catches invalid dependencies and cycles

## Migration Notes

- Database migration removes `agentHandoff` table
- Adds `plannerModel` column to workspace table
- No API breaking changes (agent endpoints unchanged)

## Testing

- Integration tests updated to remove handoff dependencies
- Agent tool test utilities simplified
- Plan validation covered by new logic

## Next Steps (Future PRs)

- Parallel execution of independent plan steps
- Dynamic re-planning based on results
- Plan caching for common routing patterns
- Error recovery strategies in plan executor


packages/twenty-server/src/engine/core-modules/workspace/workspace.entity.ts
```diff
@@ -286,6 +280,10 @@ export class WorkspaceEntity {
   @Column({ type: 'varchar', nullable: false, default: 'auto' })
   routerModel: ModelId;
 
+  @Field(() => String, { nullable: false })
+  @Column({ type: 'varchar', nullable: false, default: 'auto' })
+  plannerModel: ModelId;
```
charlesBochet: @FelixMalfait this is a breaking change between backend and database currently.

This is fine if the column is not used or if we anticipate the transition state in the code (try catch for instance or fallback behavior). While writing this, I'm wondering if the missing column will break workspace fetch during request hydratation for instance. I'm not sure what's the behavior of typeorm when there is a column in the entitySchema that doesn't exist in db. I would expect it to not query it unless we mention it but for some reason I wonder if it is not trying to match query results with entitySchemas even in select * case. 

Could you double check the behavior when the column is missing? Is this breaking twenty completely or just AI part? Can we avoid it?

If that's an issue:
- Option 1 (what we should do normally): Ideally, we introduce the column (we leave the migration file, and then we start referencing in workspace.entity.ts)
- Option 2: we move database migration before deploy (right now it's after). Note that this will move the problem to the case when we delete a column (here you are deleting a one to many on workspace so we are fine)

packages/twenty-server/src/engine/metadata-modules/agent/agent.service.ts
```diff
@@ -67,21 +67,40 @@ export class AgentService {
     });
   }
 
-  async findOneAgent(id: string, workspaceId: string) {
+  async findOneAgent(
+    workspaceId: string,
+    { id, name }: { id?: string; name?: string },
+  ) {
+    if (!id && !name) {
```
charlesBochet: use isNonEmptyString

packages/twenty-server/src/engine/metadata-modules/ai-router/services/ai-router-plan-generator.service.ts
```diff
@@ -0,0 +1,158 @@
...
+  }: {
+    messages: UIMessage<unknown, UIDataTypes, UITools>[];
+    availableAgents: AgentEntity[];
+    agentDescriptions: string;
+    plannerModel: ModelId;
+  }): Promise<ExecutionPlan> {
+    const model = this.getPlannerModel(plannerModel);
+    const agentNames = availableAgents.map((agent) => agent.name);
+
+    const conversationHistory = messages
```
charlesBochet, line 42: not sure we can do better but this code looks very bug prone (array manipulation)

packages/twenty-server/src/engine/metadata-modules/ai-router/ai-router.service.ts
```diff
@@ -46,160 +29,179 @@ export class AiRouterService {
...
+        }
 
-      if (agentIds.length === 0) {
-        throw new Error('No agent IDs available for routing schema');
-      }
+        const selectedAgent = availableAgents.find(
+          (agent) => agent.name === strategyDecision.agentName,
+        );
+
+        if (!selectedAgent) {
```
charlesBochet: nested ifs, if we can avoid that's for the better! (early returns)

PR #16025, opened 2025-11-24
Prevent NUMERIC field type creation via API

### LINKED ISSUE #16023: Prevent User from creating field NUMERIC via the API
## Bug Description

## Ask

thanks!

2400032275:
Add validation to prevent creation of NUMERIC field type via API.

Fixes #16023

### DIFF
--- a/packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-create-field-input-to-flat-field-metadatas-to-create.util.ts
+++ b/packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-create-field-input-to-flat-field-metadatas-to-create.util.ts
@@ -1,4 +1,5 @@
-import { msg } from '@lingui/core/macro';
+157
+  import { msg } from '@lingui/core/macro';
 import {
   FieldMetadataType,
   type FieldMetadataOptions,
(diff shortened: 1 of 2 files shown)


charlesBochet (changes requested):
@2400032275 thanks for opening a PR, could you actually add your code in FlatFieldMetadataTypeValidatorService (add a callback for NUMERIC case)

charlesBochet:
Also, make sure to run the linter: `npx nx run twenty-server:lint:ci`

Lastly, could you add an integration test next to failing-create-one-field-metadata-enum-v2.integration-spec.ts ? Let's introduce a failing-create-one-field-metadata.integration-spec

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-create-field-input-to-flat-field-metadatas-to-create.util.ts
```diff
@@ -1,4 +1,5 @@
-import { msg } from '@lingui/core/macro';
+157
```
charlesBochet, line 1: @2400032275 ?

charlesBochet:
@2400032275 I'm sorry, this is not the right place, CI are reds and missing test,
I'm going to close it for now, we are trying to keep only ready to merge PR on the repository, feel free to re-open once ready :)

PR #16027, opened 2025-11-24
Morph-add-new-object-destination

### LINKED ISSUE #1898: Add a new target for a Morph Relation
Before working on [migration of timeline activites](https://github.com/twentyhq/twenty/pulls?q=is%3Apr+assignee%3Aguillim+is%3Aclosed), we need to work on this project first. 

## Once the fieldmetadata MORPH_RELATION is created, is it currenlty not possible to add a new relationObject target.

We need that for dynamic relations (pre requisite for #1178 

Tech :
We want to have a dedicated INPUT type for the `updateOne` of `fieldMetadata` since this is a specific use case. This way we will not mix regular updates from a addition of moprh relation target

we want to have a dedicated fieldMetadataupdateMorphInput that will take the morphId instead of the id

Note : 

Currently we have for the create field 
```js
type RelationCreationPayload = {
  type: RelationType;
  targetObjectMetadataId: string;
  targetFieldLabel: string;
  targetFieldIcon: string;
};
```
For the create-field/input.ts
```js
  @IsOptional()
  @Field(() => GraphQLJSON, { nullable: true })
  relationCreationPayload?: RelationCreationPayload;

  @IsOptional()
  @Field(() => [GraphQLJSON], { nullable: true })
  morphRelationsCreationPayload?: RelationCreationPayload[];
```

We suggest to add this to the `UpdateFieldInput` :

```js
type RelationUpdatePayload = {
  targetObjectMetadataId: string;
};

```

guillim:
Add new object target to an existing morph relation (backend only)

Fixes https://github.com/twentyhq/core-team-issues/issues/1898


packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-update-field-input-to-flat-field-metadata.util.ts
```diff
@@ -180,3 +209,184 @@ export const fromUpdateFieldInputToFlatFieldMetadata = ({
...
+    throw new FieldMetadataException(
+      'Target flat object metadata not found',
+      FieldMetadataExceptionCode.FIELD_METADATA_NOT_FOUND,
+    );
+  }
+
+  return sourceFlatObjectMetadata;
+};
+
+const checkConflictInOtherFieldMetadataNames = ({
```
charlesBochet: this file is too long, let's split it :)

packages/twenty-server/test/integration/metadata/suites/field-metadata/morph-relation/successful-add-one-target-to-metadata-morph-relation-v2.integration-spec.ts
```diff
@@ -0,0 +1,188 @@
...
+          },
+        ],
+      },
+      expectToFail: false,
+    });
+
+    createdFieldMetadataId = rawCreateOneField.id;
+  });
+
+  it('When adding one object metadata target to pre-existing morph relation, the relation on the target object metadata should be updated', async () => {
```
charlesBochet, line 173: is there any failing case? like putting the same object on both side of the relation, or putting an object twice?

packages/twenty-shared/src/types/RelationCreationPayload.ts
```diff
@@ -4,5 +4,6 @@ export type RelationCreationPayload = {
   type: RelationType;
   targetObjectMetadataId: string;
   targetFieldLabel: string;
+  targetFieldName?: string;
```
charlesBochet: do we want to allow this? would make sense but it's not possible in the UI I believe

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-update-field-input-to-flat-field-metadata.util.ts
```diff
@@ -180,3 +209,184 @@ export const fromUpdateFieldInputToFlatFieldMetadata = ({
...
+      });
+
+    flatFieldMetadatasToCreate.push(...flatFieldMetadatas);
+    flatIndexMetadatasToCreate.push(...indexMetadatas);
+  });
+
+  return { flatFieldMetadatasToCreate, flatIndexMetadatasToCreate };
+};
+
+const getSourceObjectMetadataFromFieldMetadata = (
```
charlesBochet: not really useful I would say :)

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-update-field-input-to-flat-field-metadata.util.ts
```diff
@@ -180,3 +209,184 @@ export const fromUpdateFieldInputToFlatFieldMetadata = ({
...
+    }
+
+    const computedMorphName = computeMorphRelationFieldName({
+      fieldName: morphNameWithoutObjectName,
+      relationType: fieldMetadataToUpdate.settings.relationType,
+      targetObjectMetadataNameSingular: newTargetObjectMetadata.nameSingular,
+      targetObjectMetadataNamePlural: newTargetObjectMetadata.namePlural,
+    });
+
+    checkConflictInOtherFieldMetadataNames({
```
charlesBochet: we are doing some validation here, is this expected? @prastoin

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-update-field-input-to-flat-field-metadata.util.ts
charlesBochet: this code is already present in other places, let's see what Paul is thinking

charlesBochet (commented):
Left comments, I would need input from @prastoin

packages/twenty-shared/src/types/RelationCreationPayload.ts
guillim: We talked quickly with prastoin yesterday. This is the direction we want. IMO this is the cleanest way to do the current work. But i could also "overwrite what comes off the method `generateMorphOrRelationFlatFieldMetadataPair` from the **from-update-file-input-to-flat-field-metadata.util** if you think this is better

packages/twenty-server/test/integration/metadata/suites/field-metadata/morph-relation/successful-add-one-target-to-metadata-morph-relation-v2.integration-spec.ts
guillim: I will implement that

packages/twenty-server/src/engine/metadata-modules/field-metadata/services/field-metadata.service-v2.ts
```diff
@@ -182,6 +189,7 @@ export class FieldMetadataServiceV2 extends TypeOrmQueryService<FieldMetadataEnt
       flatViewGroupMaps: existingFlatViewGroupMaps,
       flatViewMaps: existingFlatViewMaps,
       flatViewFieldMaps: existingFlatViewFieldMaps,
+      workspaceCustomApplicationId: workspaceCustomFlatApplication.id,
```
prastoin, line 194: ***Question:*** ~~Why is that required ?~~

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-update-field-input-to-flat-field-metadata.util.ts
prastoin: Agreed lets split this and create an equivalent of `packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/validators/utils/validate-morph-relation-creation-payload.util.ts`

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-update-field-input-to-flat-field-metadata.util.ts
prastoin: Indeed, the input transpiler should not validate flat themselve if not only the input integrity required for to flat transpilation

This is checkked in:
- Relation specific validateMorphOrRelationFlatFieldJoinColumName
- Field type generic validateFlatFieldMetadataNameAvailability

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-update-field-input-to-flat-field-metadata.util.ts
prastoin: Indeed redundant, if two morph field result in same name this will be validated at field level name availability incrementally ( first will pass validation second not )

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-update-field-input-to-flat-field-metadata.util.ts
```diff
@@ -130,6 +158,7 @@ export const fromUpdateFieldInputToFlatFieldMetadata = ({
           ...accumulator.flatIndexMetadatasToUpdate,
           ...flatIndexMetadatasToUpdate,
         ],
+        flatFieldMetadatasToCreate: [...accumulator.flatFieldMetadatasToCreate],
```
prastoin, line 160: ***Nitpick:*** We could omit `flatFieldMetadatasToCreate` from reduce accumulator to avoid this, and return at the end of function instead.tbh not important

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-update-field-input-to-flat-field-metadata.util.ts
```diff
@@ -87,9 +102,22 @@ export const fromUpdateFieldInputToFlatFieldMetadata = ({
       fromFlatFieldMetadata: existingFlatFieldMetadataToUpdate,
       rawUpdateFieldInput,
     });
+
+  const { flatFieldMetadatasToCreate, flatIndexMetadatasToCreate } =
+    flatFieldMetadatasToCreateForMorphRelationsPayload({
```
prastoin: ***Remark:*** I think we should integrate this logic in the existing `computeFlatFieldToUpdateAndRelatedFlatFieldToUpdate` that sanitizes the input and so on but also to centralize side effects

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-update-field-input-to-flat-field-metadata.util.ts
```diff
@@ -180,3 +209,184 @@ export const fromUpdateFieldInputToFlatFieldMetadata = ({
...
+  });
+
+  const initialTargetFieldMetadata = findFlatEntityByIdInFlatEntityMaps({
+    flatEntityId: fieldMetadataToUpdate.relationTargetFieldMetadataId ?? '',
+    flatEntityMaps: flatFieldMetadataMaps,
+  });
+  const commonTargetFieldLabel = initialTargetFieldMetadata?.label ?? '';
+  const commonTargetFieldName = initialTargetFieldMetadata?.name ?? '';
+
+  morphRelationsUpdatePayload.forEach((morphRelationUpdatePayload) => {
```
prastoin: ***Nitpick:*** not a fan of the forEach here, would rather have a for of instead or a reduce

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-update-field-input-to-flat-field-metadata.util.ts
```diff
@@ -180,3 +209,184 @@ export const fromUpdateFieldInputToFlatFieldMetadata = ({
...
+    morphRelationFlatFieldMetadata: {
+      name: fieldMetadataToUpdate.name,
+      settings: fieldMetadataToUpdate.settings,
+    },
+    nameSingular: initialFlatFieldMetadataTargetObjectMetadata.nameSingular,
+    namePlural: initialFlatFieldMetadataTargetObjectMetadata.namePlural,
+  });
+
+  const initialTargetFieldMetadata = findFlatEntityByIdInFlatEntityMaps({
+    flatEntityId: fieldMetadataToUpdate.relationTargetFieldMetadataId ?? '',
```
prastoin: ***Suggestion:*** We should have a MorphOrRelation falt field here the `?? ''` is redundant as it's non nullable for such field type

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-update-field-input-to-flat-field-metadata.util.ts
```diff
@@ -180,3 +209,184 @@ export const fromUpdateFieldInputToFlatFieldMetadata = ({
...
+    flatEntityMaps: flatFieldMetadataMaps,
+  });
+  const commonTargetFieldLabel = initialTargetFieldMetadata?.label ?? '';
```
prastoin: ***Question:*** We should rather set null instead of empty string regarding upcoming Etienne's job about null collascing

prastoin: We might wanna early return or have two disctinct block for handling update that result in new field relation pair creation or just an update of an existing pair

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-update-field-input-to-flat-field-metadata.util.ts
```diff
@@ -180,3 +209,184 @@ export const fromUpdateFieldInputToFlatFieldMetadata = ({
...
+      newMorphRelationName: computedMorphName,
+    });
+
+    const { flatFieldMetadatas, indexMetadatas } =
+      generateMorphOrRelationFlatFieldMetadataPair({
+        createFieldInput: {
+          type: FieldMetadataType.MORPH_RELATION,
+          name: computedMorphName,
+          label: morphRelationsCommonLabel,
+          objectMetadataId: initialTargetFieldMetadata?.objectMetadataId ?? '',
```
prastoin: ***Question:*** Unless I'm mistaken this does not work

packages/twenty-shared/src/types/RelationCreationPayload.ts
prastoin: I don't think we should already add this to the creation relation payload directly though, but rather extends in the update relation creation payload

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-update-field-input-to-flat-field-metadata.util.ts
prastoin: Should fallback to an object metadata id in case of update relation creation payload resulting in a field creation

packages/twenty-server/test/integration/metadata/suites/field-metadata/morph-relation/successful-add-one-target-to-metadata-morph-relation-v2.integration-spec.ts
prastoin: We should also add integration coverage to basic relation not only morph relation if possibe

prastoin (commented):
Cool job so far !

packages/twenty-server/src/engine/metadata-modules/field-metadata/services/field-metadata.service-v2.ts
prastoin: Ok my bad it's for creation

packages/twenty-shared/src/types/RelationCreationPayload.ts
guillim: Fine with me. I will remove it for now

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-update-field-input-to-flat-field-metadata.util.ts
guillim: ok. removing the check

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-update-field-input-to-flat-field-metadata.util.ts
guillim: it never is an update of an exisiting pair. But yes, will do an early return for the case of undefined label

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-update-field-input-to-flat-field-metadata.util.ts
guillim: Tried at first, but changes the API quite a bit, and from my first tests, it wasn't working. But if you believe this is where it should belong to, I can try again in that direction.

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-update-field-input-to-flat-field-metadata.util.ts
guillim: why ?

charlesBochet:
Integrations tests are red :)

packages/twenty-server/test/integration/metadata/suites/field-metadata/morph-relation/__snapshots__/failing-add-one-target-to-metadata-morph-relation-v2.integration-spec.ts.snap
```diff
@@ -0,0 +1,189 @@
...
+  "extensions": {
+    "code": "METADATA_VALIDATION_FAILED",
+    "errors": {
+      "cronTrigger": [],
+      "databaseEventTrigger": [],
+      "fieldMetadata": [
+        {
+          "errors": [
+            {
+              "code": "NOT_AVAILABLE",
```
prastoin, line 14: ***Remark:*** Would have expected morph specific errors here and no join col and names collisions

packages/twenty-server/test/integration/metadata/suites/field-metadata/morph-relation/successful-add-one-target-to-metadata-morph-relation-v2.integration-spec.ts
```diff
@@ -0,0 +1,289 @@
...
+import { extractRecordIdsAndDatesAsExpectAny } from 'test/utils/extract-record-ids-and-dates-as-expect-any';
+import { FieldMetadataType } from 'twenty-shared/types';
+import { isDefined } from 'twenty-shared/utils';
+
+import { RelationType } from 'src/engine/metadata-modules/field-metadata/interfaces/relation-type.interface';
+
+import { type FieldMetadataDTO } from 'src/engine/metadata-modules/field-metadata/dtos/field-metadata.dto';
+import { type RelationDTO } from 'src/engine/metadata-modules/field-metadata/dtos/relation.dto';
+
+describe('updateOne FieldMetadataService morph relation fields v2 - Add one target', () => {
```
prastoin, line 17: ***Nitpick:*** Could split in two files one for failing and for success

packages/twenty-server/test/integration/metadata/suites/field-metadata/morph-relation/successful-add-one-target-to-metadata-morph-relation-v2.integration-spec.ts
```diff
@@ -0,0 +1,289 @@
...
+        type
+        targetFieldMetadata {
+          id
+          name
+        }
+      }
+      `,
+    });
+
+    const updateOneField = result.data
```
prastoin, line 209: ***Nitpick:*** Could be testing we have an index on field creation too

packages/twenty-server/test/integration/metadata/suites/field-metadata/morph-relation/successful-add-one-target-to-metadata-morph-relation-v2.integration-spec.ts
```diff
@@ -0,0 +1,289 @@
...
+      `,
+    });
+
+    expect(errors.length).toBe(1);
+    const [firstError] = errors;
+
+    expect(firstError).toMatchSnapshot(
+      extractRecordIdsAndDatesAsExpectAny(firstError),
+    );
+    expect(firstError.extensions.code).not.toBe('INTERNAL_SERVER_ERROR');
```
prastoin: ***Nitpick:*** Could use `expectOneNotInternalServerErrorSnapshot`

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-update-field-input-to-flat-field-metadata.util.ts
```diff
@@ -87,9 +93,30 @@ export const fromUpdateFieldInputToFlatFieldMetadata = ({
...
+    isFlatFieldMetadataOfType(
+      existingFlatFieldMetadataToUpdate,
+      FieldMetadataType.MORPH_RELATION,
+    )
+      ? computeFlatFieldToUpdateFromMorphRelationUpdatePayload({
+          workspaceCustomApplicationId,
+          morphRelationsUpdatePayload:
+            rawUpdateFieldInput?.morphRelationsUpdatePayload,
+          flatFieldMetadataMaps: flatFieldMetadataMaps,
+          fieldMetadataToUpdate: existingFlatFieldMetadataToUpdate,
```
prastoin: ***Remark:*** We should be passing the `flatFieldMetadataFromTo`

packages/twenty-server/test/integration/metadata/suites/field-metadata/morph-relation/successful-add-one-target-to-metadata-morph-relation-v2.integration-spec.ts
prastoin: We should add a new test that covers both the name + morph creation addition in only one input
And expect the added created morph to follow the updated name

prastoin (approved):
I can already approve this very good job !
Left few comments that we've already been talking about together irl
Please let me know !

PR #16030, opened 2025-11-24
Migrate metadata cache

Weiko:
## Context
Deprecating legacy ObjectMetadata from cache in favor of flat entities.
Introducing utils to build byName/byNameSingular/byNamePlural in isolated cases

## Next
- I had to introduce a util to build from flat to legacy objectMetadataMaps, we should instead use flat maps directly when needed (datasource, schema generation, etc)
- Deprecate metadata version in the cache
- Use the new cache strategy for flat entities with permissions and feature flags and inject in the global datasource context


packages/twenty-server/src/engine/api/graphql/workspace-schema.factory.ts
```diff
@@ -51,31 +54,65 @@ export class WorkspaceSchemaFactory {
...
+      metadataVersion = authContext.workspace.metadataVersion ?? 0;
+      await this.workspaceCacheStorageService.setMetadataVersion(
+        workspaceId,
+        metadataVersion,
       );
     }
 
-    const workspaceFeatureFlagsMap =
-      await this.featureFlagService.getWorkspaceFeatureFlagsMap(workspaceId);
+    const { idByNameSingular } = buildObjectIdByNameMaps(
```
Weiko, line 99: generateSchema should consume flat entities separately in the future

packages/twenty-server/src/engine/api/rest/core/handlers/rest-api-base.handler.ts
```diff
@@ -239,6 +264,21 @@ export abstract class RestApiBaseHandler {
       );
     }
 
+    const objectMetadataMaps: ObjectMetadataMaps = {
+      byId: {},
+      idByNameSingular,
+    };
+
+    for (const [id, flatObj] of Object.entries(flatObjectMetadataMaps.byId)) {
```
Weiko, line 269: same, to check if we can avoid this

packages/twenty-server/src/engine/dataloaders/dataloader.service.ts
```diff
@@ -486,20 +485,27 @@ export class DataloaderService {
   private createObjectMetadataLoader() {
     return new DataLoader<
       ObjectMetadataLoaderPayload,
-      ObjectMetadataItemWithFieldMaps | null
+      ObjectMetadataDTO | null
```
Weiko, line 488: This was never used with fields map, much simpler now

packages/twenty-server/src/engine/twenty-orm/factories/workspace-datasource.factory.ts
```diff
@@ -126,15 +129,47 @@ export class WorkspaceDatasourceFactory {
...
 
+          const { idByNameSingular } = buildObjectIdByNameMaps(
+            flatObjectMetadataMaps,
+          );
+          const cachedObjectMetadataMaps: ObjectMetadataMaps = {
+            byId: {},
+            idByNameSingular,
+          };
+
+          for (const [id, flatObj] of Object.entries(
```
Weiko, line 155: same here, entitySchema creation should consume flat entities separately so we don't have to build legacy map

packages/twenty-server/src/engine/twenty-orm/global-workspace-datasource/global-workspace-orm.manager.ts
```diff
@@ -93,13 +99,46 @@ export class GlobalWorkspaceOrmManager {
...
           workspaceId,
+          flatMapsKeys: [
+            'flatObjectMetadataMaps',
+            'flatFieldMetadataMaps',
+            'flatIndexMaps',
+          ],
         },
       );
 
+    const { idByNameSingular } = buildObjectIdByNameMaps(
```
Weiko, line 102: same, this is for the new datasource entity schema build, we should use flat entities separately

charlesBochet (approved):

PR #16033, opened 2025-11-24
Fix and improve chart type selection

bosiraphael:
The new version of the side panel is smaller so it introduced a regression on the chart type selection because there was not enough space for all the chart types to fit with the label.

This PR removes the label to gain some space and display the chart type label in a tooltip.

## Before

CleanShot 2025-11-24 at 17 33 45@2x

### Without disabled charts

https://github.com/user-attachments/assets/74d62b86-e440-41f5-a3e6-c5754940c28e

### With disabled charts

https://github.com/user-attachments/assets/f4ce3a3a-5dee-4f41-8b96-5f671794197d


charlesBochet (approved):
Bravo, what a knowledge of HTML5

PR #16035, opened 2025-11-24
Fix page layout widget deletion

bosiraphael:
With the new side panel, we are able to delete a widget with the side panel still open. This caused the app to crash because we threw when the widget id wasn't defined.

This PR fixes this by closing the side panel in the delete action and by returning null instead of throwing.

## Before

https://github.com/user-attachments/assets/092bfe62-82dc-4d83-9967-1cc753ecf55e

## After

https://github.com/user-attachments/assets/8bed6cc5-961b-4112-8cf5-e587865d14da


bosiraphael:
@greptileai

charlesBochet (approved):

PR #16038, opened 2025-11-24
fix: prevent NUMERIC field type creation via API

### LINKED ISSUE #16023: Prevent User from creating field NUMERIC via the API

neo773:
Block users from creating NUMERIC, POSITION, and TS_VECTOR fields via the API as these are system-only types. Users should use NUMBER instead of NUMERIC.

/closes #16023


charlesBochet (approved):
Love it!

PR #16039, opened 2025-11-25
fix: guard against invalid date and undefined links field value

### LINKED ISSUE #15854: User is shown "Something went wrong" error a lot, several Sentry issues found
## Bug Description
Cannot read properties of undefined (reading 'primaryLinkUrl') and RangeError - Invalid time value
https://github.com/twentyhq/private-issues/issues/370

ehconitin:
closes https://github.com/twentyhq/twenty/issues/15854

- Add `isValid()` check in `formatDateISOStringToDateTime` before calling `formatInTimeZone`
- Add `isDefined()` guard in `getFieldLinkDefinedLinks` (mirrors `phonesUtils` pattern)

before:

https://github.com/user-attachments/assets/1eb89fa4-70b6-4794-8860-0a42522598b5

https://github.com/user-attachments/assets/781c7d37-c435-4832-98d4-8e6925b51e10

after:

https://github.com/user-attachments/assets/b1c22d25-4e8e-45c3-af98-be1684a5962e

https://github.com/user-attachments/assets/ce084a9d-03b8-4f88-8522-a32cb1b7cf2f


packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/input/utils/getFieldLinkDefinedLinks.ts
```diff
@@ -3,6 +3,10 @@ import { isNonEmptyString } from '@sniptt/guards';
 import { isDefined, isValidUrl } from 'twenty-shared/utils';
 
 export const getFieldLinkDefinedLinks = (fieldValue: FieldLinksValue) => {
+  if (!isDefined(fieldValue)) {
```
charlesBochet, line 6: @ehconitin I'm not a fan of this fix as is because your check is not aligned with typing:
- either we should make fieldValue: FieldLinksValue | undefined
- either we should not and this check shouldn't be

I prefer that we keep utils interface as simple as possible (not adding undefined). This reduces the amount  of typing branches in the code. In this case, we should not call the util in the first place at all.

packages/twenty-front/src/modules/localization/utils/formatDateISOStringToDateTime.ts
```diff
@@ -15,14 +16,15 @@ export const formatDateISOStringToDateTime = ({
   timeFormat: TimeFormat;
   localeCatalog: Locale;
 }) => {
+  const parsedDate = new Date(date);
+
+  if (!isValid(parsedDate)) {
```
charlesBochet, line 21: I feel we are not using it right here. We should also prevent this util to be called with a text default value in the first place

packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/input/utils/getFieldLinkDefinedLinks.ts
charlesBochet: here we should not try to load a field value if the record is not loaded yet

packages/twenty-front/src/modules/localization/utils/formatDateISOStringToDateTime.ts
charlesBochet: here we should not try to load the date component with an outdated text value

charlesBochet:
@ehconitin would you have time to work on this one?

prastoin:
Hey @ehconitin is this still in wip ? Closing this until you have some time to come back on this PR, please feel free to reopen

charlesBochet:
Sorry, it's on me, re-opening :)

packages/twenty-front/src/modules/localization/utils/formatDateISOStringToDateTime.ts
lucasbordeau: I'll keep this as it is still a good guard, but I implemented the fix you mention

lucasbordeau (approved):

lucasbordeau:
I couldn't reproduce bug one on main but the added check is still a good thing to have.

PR #16042, opened 2025-11-25
Clicking outside inline fields on record page saves the value.

mabdullahabaid:
Closes #15957

The core problem: `handleChange` calls `setDraftValue`, but when blur fires, `handleClickOutside` runs in the same event turn and uses the `draftValue` captured in its closure from the last render. React hasn’t re-rendered yet, so that captured `draftValue` is stale. Recoil atom writes are synchronous, but the render that would update the closure happens on the next turn.

I considered deferring `handleClickOutside` to the next tick (setTimeout/Promise), but that’s a timing hack: it makes focus/blur ordering unpredictable (inline cells, tables, other listeners), risks double triggers, and can fire after unmount. Another option was to duplicate the `handleChange` logic inside `handleClickOutside` (screenshot), but that defeats having draft state in one place.

<p align="center">
const nextSecondaryLinks = updatedLinks slice (1);
</p>

The clean solution is to read the current draft from a Recoil snapshot at call time inside `handleClickOutside`. Snapshot reads pull the latest atom value (including synchronous `setDraftValue` that just ran) even before a re-render occurs, so they’re not subject to the stale closure. That way, blur uses the up-to-date draft without timing hacks or duplicated logic.

MultiItemFieldInput is used in four places. Each of them contains the updated code.

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-field/ui/hooks/useRecordFieldInput.ts
+++ b/packages/twenty-front/src/modules/object-record/record-field/ui/hooks/useRecordFieldInput.ts
@@ -7,6 +7,19 @@ export const useRecordFieldInput = <FieldValue>() => {
   const recordFieldInputDraftValueCallbackState =
     useRecoilComponentCallbackState(recordFieldInputDraftValueComponentState);
 
+  const getLatestDraftValue = useRecoilCallback(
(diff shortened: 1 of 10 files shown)


lucasbordeau (approved):
LGTM nice description of the problem

PR #16048, opened 2025-11-25
[DASHBOARDS] Pie Chart

### LINKED ISSUE #1536: Connect the pie chart to the group by resolver
Connect the pie chart to the group by resolver

bosiraphael:
## Description

- Connect Pie Chart to the backend
- Update from the old design to the new design
- Switch seamlessly from/to other types of chart from the Pie Chart by converting the configuration

Note: There are a couple things to implement before the pie chart is ready to go live:
- Add a new setting on Bar, Line and Pie chart to hide and display legends (toggle)
- Add data labels next to each slice

## Video QA

https://github.com/user-attachments/assets/b1561e32-0434-43ad-8855-d617b209ce27


ehconitin:
CleanShot 2025-11-25 at 16 56 31

CleanShot 2025-11-25 at 17 00 30

pie chart should handle the edge cases where configuration doesnt make sense -- 
like in this case -- it should probably show full pie? probably greyed out?

also legends has no labels when the config isnt correct

like how when in line and bar chart -- when config doesnt make sense (ie no sensible data to be displayed) - -the axes are still visible? I am unsure how to fo it with pie though

ehconitin:
The legends threshold should be solved with this PR - https://github.com/twentyhq/twenty/pull/16049 -- so no need to worry about legends conditional rendering!

bosiraphael:
@ehconitin it's not that it doesn't make sense, it's just that you showed `percent not empty`, which is always equal to 0 in this case. I don't think we should show anything if all the data i 0.
What so you think @Bonapara?

packages/twenty-front/src/modules/command-menu/pages/page-layout/types/ChartConfigurationSettingIds.ts
```diff
@@ -4,7 +4,7 @@ export enum CHART_CONFIGURATION_SETTING_IDS {
   COLORS = 'COLORS',
   FILTER = 'FILTER',
   GROUP_BY = 'GROUP_BY',
-  SORT_BY_X = 'SORT_BY_X',
+  PRIMARY_BY_X = 'PRIMARY_BY_X',
```
ehconitin: SORT_BY_PRIMARY or PRIMARY_SORT_BY

packages/twenty-front/src/modules/command-menu/pages/page-layout/hooks/useChartSettingsValues.ts
```diff
@@ -145,15 +162,33 @@ export const useChartSettingsValues = ({
             : ''
         }`;
       }
+      case CHART_CONFIGURATION_SETTING_IDS.DATA_ON_DISPLAY_PIE_CHART: {
```
ehconitin, line 165: we should refactor this :) at this point its too hard to understand whats going on

I am creating a ticket

ehconitin: we can keep it for now

packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetPieChart/hooks/useGraphPieChartWidgetData.ts
```diff
@@ -0,0 +1,63 @@
...
+type UseGraphPieChartWidgetDataResult = {
+  data: PieChartDataItem[];
+  loading: boolean;
+  error?: Error;
+  hasTooManyGroups: boolean;
+  objectMetadataItem: ObjectMetadataItem;
+};
+
+// TODO: Remove this once backend returns total group count
+const EXTRA_ITEM_TO_DETECT_TOO_MANY_GROUPS = 1;
```
ehconitin, line 24: duplicate to useGraphBarChartWidgetData?
would like your opinion on this

we have useGraphLineChartWidgetData , useGraphBarChartWidgetData and now useGraphPieChartWidgetData

All looks similar, just with different transformation

should we abstract?

packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetPieChart/hooks/usePieChartHandlers.ts
```diff
@@ -5,21 +5,30 @@ import { isDefined } from 'twenty-shared/utils';
...
     datum: ComputedDatum<{ id: string; value: number; label?: string }>,
   ) => {
     const clickedItem = data.find((d) => d.id === datum.id);
-    if (isDefined(clickedItem?.to)) {
-      window.location.href = clickedItem.to;
+    if (isDefined(clickedItem)) {
+      if (isDefined(onSliceClick)) {
+        onSliceClick(clickedItem);
+      } else if (isDefined(clickedItem.to)) {
+        window.location.href = clickedItem.to;
```
ehconitin, line 25: I will get rid of this `to` in coming PRs

packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetPieChart/utils/transformGroupByDataToPieChartData.ts
```diff
@@ -0,0 +1,101 @@
...
+  const rawResults = groupByData[queryName];
+
+  if (!isDefined(rawResults) || !Array.isArray(rawResults)) {
+    return EMPTY_PIE_CHART_RESULT;
+  }
+
+  // TODO: Add a limit to the query instead of slicing here (issue: twentyhq/core-team-issues#1600)
+  const limitedResults = rawResults.slice(
+    0,
+    PIE_CHART_MAXIMUM_NUMBER_OF_SLICES,
```
ehconitin, line 65: we can add limit no? or is it still not working as expected?

ehconitin (commented):
left some questions! otherwise lgtm

packages/twenty-front/src/modules/command-menu/pages/page-layout/hooks/useChartSettingsValues.ts
bosiraphael: Yes I thought about it too, but didn't want to do this in this PR

packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetPieChart/hooks/useGraphPieChartWidgetData.ts
bosiraphael: No, because I'm adding different logics for the limit. I don't want to force abstraction here

packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetPieChart/utils/transformGroupByDataToPieChartData.ts
bosiraphael: We have the limit, but we have to slice for now because we don't have the total number of groups returned by the group by. So the solution is the meantime is too ask for `limit + 1` in the group by limit and then slice to see if we have exactly the right number of groups of if some are missing.

ehconitin (approved):
lgtm

PR #16050, opened 2025-11-25
Null equivalence - Fix

etiennejouan:
@charlesBochet 
In workflow codebase, NULL (instead of empty object) is expected on workflow version object  step field, when a new workflow is created for example. (packages/twenty-front/src/modules/workflow/workflow-diagram/utils/generateWorkflowDiagram.ts - 42)
This case is not isolated and it creates many issues.

We decided to format NULL value to equivalent (empty string for text field, empty object for raw_json) but it seems it complicates the dev x.

To unlock @Devessier I prefer revert the logic, the time we discuss how to solve this cases.


Devessier (approved):
Thanks. It brings back workflow edition.

PR #16051, opened 2025-11-25
Prevent NUMERIC field type creation via API

### LINKED ISSUE #16023: Prevent User from creating field NUMERIC via the API

2400032275:
Add validation to prevent creation of NUMERIC field type via API.

Fixes #16023

## Changes
- Added validation in `fromCreateFieldInputToFlatFieldMetadatasToCreate` utility to prevent NUMERIC field type creation
- NUMERIC type causes data access issues for users
- Users should use NUMBER type instead
- Includes proper error handling with user-friendly message

## Related
This PR addresses the issue identified in #16023 where users were able to create NUMERIC fields via the API, leading to data access problems. The validation has been placed in the correct location within the field creation flow.


neo773:
Closing this as the issue has already been addressed, thank you for your contribution though.

PR #16056, opened 2025-11-25
Fix chart limits for two-dimensional group by

bosiraphael:
The limit wasn't working properly for two dimensional stacked charts. This PR fixes this.


ehconitin (approved):
lgtm

PR #16057, opened 2025-11-25
[fix] User have no firstName nor lastName

### LINKED ISSUE #14678: Filtering on CREATED BY not returning results
## Bug Description

Filtering a list on Created By / Contains / Adding there the name of the workspace member is not returning results - despite seeing the expected records. 

**Example:** 
Reported by a customer. I could reproduce on her workspace but this is working as expected on mine

### LINKED ISSUE #15692: First name and last name not saved on user
When a user joins a workspace, we ask them for their first name and last name that we save in their workspaceMember, but not on the user.
This is problematic because, among probably other things, we use the user saved name to determine createdByName (cf CreatedByFromAuthContextService). Therefore createdByName is left empty and we will lose the creator information if a user leaves the workspace, as their workspaceMember will be deleted. 

Several things to do here
1. Save first name and last name on user
2. Fix existing: Create and run a command to get name from workspaceMember on user, or in CreatedByFromAuthContextService default to workspaceMember name if userName is an empty string

ijreilly:
Closes https://github.com/twentyhq/twenty/issues/15692, fixes https://github.com/twentyhq/twenty/issues/14678

The issue was:
- When a user is signing up, they are asked for their first name and last name with which we fulfill their workspaceMember entity for the workspace they are signing up to (which is the one they are creating if they are not joining an existing workspace). user.firstName and user.lastName remain empty
- When we create a record, we are using user.firstName and user.lastName to fulfill the text field "createdByName", which intends to save the name of the creator at the moment of the creation. This field is then use 1) when filtering on createdBy (as a trick to be able to filter by this, since we don't have filters on nested fields, ie we dont have filters on person.createdByWorkspaceMember)  2) to display the user creator when a user has left a workspace - otherwise we rely on the dynamic createdByWorkspaceMember.firstName and lastName
- So filtering on createdBy was not working as createdByName is empty. We did not see that in dev mode because we have seeded users with names

The fix
- When we create a record, we use workspaceMember.firstName and workspaceMember.lastName as we should.
- When an existing user joins a workspace, they are asked to give their name again so that we set their workspaceMember name
- This will not fix the history of the records (they will still have createdByName empty so the filter will not work properly), but a command to fix it would be very long as it would iterate over all the records of each active workspace

The long-term view is to get rid of user and to store the name in userWorkspace


Weiko (approved):
LGTM!

PR #16058, opened 2025-11-25
Add import scheduled status to messaging sync

charlesBochet:
We have introduced to new syncStage statuses: `messageChannel.MESSAGES_IMPORT_SCHEDULED` and `calendarChannel.CALENDAR_EVENTS_IMPORT_SCHEDULED`

We need to make sure all existing workspaces have it


prastoin (approved):
Neat

PR #16061, opened 2025-11-25
Don't recompute output schema when moving trigger

### LINKED ISSUE #15982: Moving workflow trigger breaks variable links
https://github.com/user-attachments/assets/6a930a09-31ec-4b58-9326-c45e53fe0892

Devessier:
Position changes don't affect the trigger's output schema, so recomputing it is unnecessary. For webhook triggers specifically, the backend can't reconstruct the schema (it returns `{}`), which overwrites the frontend-built schema from `expectedBody`. Passing `{ computeOutputSchema: false }` prevents this data loss while avoiding wasteful computation.

## Demo

https://github.com/user-attachments/assets/116af221-0d09-4dda-ad40-3301fa610bcd

Closes https://github.com/twentyhq/twenty/issues/15982


Weiko (approved):
LGTM

PR #16063, opened 2025-11-25
Improve board experience 🖼️

lucasbordeau:
This PR improves the general UX and DX of boards, by modifying the query effect to only use paged group by queries.

In this PR we implement two more things in the backend for group by queries : 
- Fixed ORDER BY in the PARTITION BY sub-query (this wasn't working because it was applied in the main query, so it sorted randomly picked records, which was a correct sort on an incorrect dataset returned by the sub-query)
- Added offset paging in PARTITION BY

Miscellaneous, various bug fixes and improvements along the way : 
- Throttled loading of cards to avoid React freeze
- Handling of drag & drop
- Handling of create / delete / update
- Reworked skeleton (the library slows down a lot with hundreds of skeleton for a spinning effect that is hardly noticed)
- Fixed refetch of aggregate queries (I included the new group by aggregates query we use in the existing refetch mechanism)
- Re-trigger queries on filters and sorts changes
- Unselect all record ids when deleting / restoring / detroying
- Fetch only groups that still have records to lighten the group by query.

# What remains to be done 

This is still a naïve fetch more implementation that will work for a few fetch more rounds, but if you scroll and load say 200 cards per column on a board, React will re-render all 200 cards of each column each time.

We would probably need to virtualize the board with paged queries as we did for the table, this could be done after this PR but seems less urgent.

What's nice is that this new query pattern is well designed for virtualization also, drawing from our experience with table virtualization, and adapted to a multi-column request pattern, like a 2:2 matrix of records, for our boards.

So the remaining work would be to design a UI solution for virtualizing this matrix of records, which could be quite different from our table virtualization mechanism.


packages/twenty-front/src/modules/object-record/record-board/components/RecordBoard.tsx
```diff
@@ -48,30 +42,29 @@ export const RecordBoard = () => {
...
-    >
+    <>
       <ScrollWrapper
         componentInstanceId={`scroll-wrapper-record-board-${recordBoardId}`}
       >
         <RecordBoardStickyHeaderEffect />
         <RecordBoardScrollToFocusedCardEffect />
         <RecordBoardDeactivateBoardCardEffect />
+        <RecordBoardQueryEffect />
+        <RecordBoardSelectRecordsEffect />
```
bosiraphael: At this point I'm wondering if we should have a `RecordBoardEffects` which returns all the effects in a fragment

packages/twenty-front/src/modules/object-record/record-board/components/RecordBoardQueryEffect.tsx
```diff
@@ -0,0 +1,88 @@
...
+    queryIdentifier,
+    setRecordBoardCurrentGroupByQueryOffset,
+    scrollWrapperHTMLElement,
+    recordIndexRecordGroupsAreInInitialLoading,
+    shouldFetchMore,
+    recordBoardIsFetchingMore,
+    triggerRecordBoardFetchMore,
+  ]);
+
+  return <></>;
```
bosiraphael: Nit: return `null`

packages/twenty-front/src/modules/object-record/record-board/hooks/useTriggerRecordBoardFetchMore.ts
```diff
@@ -0,0 +1,230 @@
...
+          );
+
+          set(
+            recordIndexRecordIdsByGroupCallbackState(recordGroupDefinition.id),
+            newRecordIds,
+          );
+
+          upsertRecordsInStore(newRecords);
+
+          if (newRecords.length < 10) {
```
bosiraphael: Create a constant for this magic number

packages/twenty-front/src/modules/object-record/record-board/components/RecordBoard.tsx
lucasbordeau: Why not indeed

packages/twenty-front/src/modules/object-record/record-board/components/RecordBoard.tsx
bosiraphael: You could also put this effect inside `RecordBoardEffects`

bosiraphael (approved):
LGTM, left a small nitpick

PR #16065, opened 2025-11-25
[Side Panel V2] Bring back old container

Devessier:
## Before

image

## After

CleanShot 2025-11-25 at 17 35 59@2x


Weiko (approved):
LGTM

PR #16069, opened 2025-11-25
Security - Add complexity max on gql queries

etiennejouan:
closes https://github.com/twentyhq/private-issues/issues/348
closes https://github.com/twentyhq/private-issues/issues/352
closes https://github.com/twentyhq/private-issues/issues/353
closes https://github.com/twentyhq/private-issues/issues/354

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/graphql/hooks/use-compute-complexity.hook.ts
+++ b/packages/twenty-server/src/engine/core-modules/graphql/hooks/use-compute-complexity.hook.ts
@@ -3,7 +3,7 @@ import { type Plugin } from 'graphql-yoga';
 
 import { UserInputError } from 'src/engine/core-modules/graphql/utils/graphql-errors.util';
 
-export const useComputeComplexity = (maximumComplexity = 1000): Plugin => ({
+export const useComputeComplexity = (maximumComplexity: number): Plugin => ({
   onValidate: ({ addValidationRule }) => {
     addValidationRule(
       queryComplexity({
--- a/packages/twenty-server/src/engine/core-modules/graphql/hooks/use-compute-complexity.hook.ts
+++ b/packages/twenty-server/src/engine/core-modules/graphql/hooks/use-compute-complexity.hook.ts
@@ -1,22 +1,29 @@
-import queryComplexity, { simpleEstimator } from 'graphql-query-complexity';
+import { type ValidationContext } from 'graphql';
 import { type Plugin } from 'graphql-yoga';
 
 import { UserInputError } from 'src/engine/core-modules/graphql/utils/graphql-errors.util';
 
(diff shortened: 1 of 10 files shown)


packages/twenty-server/src/engine/core-modules/graphql/hooks/use-compute-complexity.hook.ts
```diff
@@ -0,0 +1,29 @@
...
+      return {
+        Field() {
+          complexity++;
+        },
+        Document: {
+          leave() {
+            if (complexity > maximumComplexity) {
+              context.reportError(
+                new UserInputError(
+                  `Query complexity is too high: ${complexity} - Too many fields requested`,
```
charlesBochet, line 21: should we add translations here?

packages/twenty-server/src/engine/core-modules/graphql/hooks/use-compute-complexity.hook.ts
```diff
@@ -0,0 +1,29 @@
+import { type ValidationContext } from 'graphql';
+import { type Plugin } from 'graphql-yoga';
+
+import { UserInputError } from 'src/engine/core-modules/graphql/utils/graphql-errors.util';
+
```
FelixMalfait, line 6: Very cool!!! Could we add tests?  Not the biggest advocate for tests but it feels like the kind of thing that would deserve it

Is there a limit on the number of objects? I guess that if I do 200 joins it could break the server even if I don't fetch 2000 fields? GroupBy is also protected by this you think?  Sorry I guess you already discussed this with someone but I'm curious

etiennejouan: We're keeping the gql validation very simple since it won't protect the rest api. It simply sums the requested fields. We needed a layer in GraphQL to protect against some DDoS attacks identified in the pen test report.

In a second time, we'll add a complexity calculation layer in the common API, where we can easily compute a score from args and requested fields. 
the rules are not fully defined yet, but it will at least add a multiplier for each requested related object.


I'll add some tests!

FelixMalfait (approved):

packages/twenty-server/src/engine/core-modules/graphql/hooks/use-compute-complexity.hook.ts
FelixMalfait: I feel like doing it at the Common API Layer might be a little over-engineer because this problem is very GraphQL specific. If we ever add ?include=xxx.yyy in rest or similar capabilities in AI/Workflow tools I agree it should move to the common layer, but likely not in the next year imo, that's why I felt taking into account the number of objects now here made sense. But all good for me like this. And if we do move it to the common API layer great too!

PR #16087, opened 2025-11-26
[Workflows] Fix filtering on relative date

### LINKED ISSUE #16054: [Workflows] Filtering on relative date is not working
please see the attached video, reported by a customer
I can reproduce in production

Clicking on the "This"/"Past"/"Next" button does not select the value on which the user clicked. 

He also showed a few other issues during his video

https://share.zight.com/d5udWLbW

Front logo [Front conversations](https://app.frontapp.com/open/top_3rxv12)

Devessier:
https://github.com/user-attachments/assets/4079d59e-0550-401f-a29b-b235edd8ed48

Closes https://github.com/twentyhq/twenty/issues/16054

### DIFF
--- a/packages/twenty-front/src/modules/ui/input/components/internal/date/components/RelativeDatePickerHeader.tsx
+++ b/packages/twenty-front/src/modules/ui/input/components/internal/date/components/RelativeDatePickerHeader.tsx
@@ -4,7 +4,6 @@ import { RELATIVE_DATE_DIRECTION_SELECT_OPTIONS } from '@/ui/input/components/in
 import { RELATIVE_DATE_UNITS_SELECT_OPTIONS } from '@/ui/input/components/internal/date/constants/RelativeDateUnitSelectOptions';
 
 import styled from '@emotion/styled';
-import { useId } from 'react';
 import { type Nullable } from 'twenty-shared/types';
 import {
   relativeDateFilterSchema,
@@ -22,6 +21,7 @@ const StyledContainer = styled.div<{ noPadding: boolean }>`
 `;
 
 type RelativeDatePickerHeaderProps = {
+  instanceId: string;
   direction: RelativeDateFilterDirection;
   amount?: Nullable<number>;
   unit: RelativeDateFilterUnit;
@@ -32,6 +32,7 @@ type RelativeDatePickerHeaderProps = {
 };
 
 export const RelativeDatePickerHeader = ({
+  instanceId,
   direction,
   unit,
   amount,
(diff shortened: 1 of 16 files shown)


martmull (approved):

packages/twenty-front/src/modules/ui/input/components/internal/date/components/RelativeDatePickerHeader.tsx
```diff
@@ -39,6 +40,8 @@ export const RelativeDatePickerHeader = ({
   readonly,
   unitDropdownWidth,
 }: RelativeDatePickerHeaderProps) => {
+  const instanceId = useId();
```
lucasbordeau: I suggest not using useId has it is not resistant to re-renders, we should always use a static if that we can obtain from the context.

lucasbordeau: We already apply this strategy throughout the codebase, there's always information we can use to craft a unique identifier.

lucasbordeau: This is especially true for dropdown ids which should be determined by their environment and not generated.

Devessier: It is resistant to re-renders but not to remounts. I agree that it’s better to use a higher-level identifier. I’ll give it a try.

lucasbordeau: You can add a required props `id` or `dropdownIdSuffix` if there's nothing that allows to craft a unique identifier from the context inside this component.

lucasbordeau: Of course but a remount happens easily if the parent re-renders and recreates its children which can happen very easily for a component that is generic like this one.

lucasbordeau: Better safe than sorry, by not allowing something to bug, it won't bug, that should be a principle like least privilege in security.

lucasbordeau: With our dropdowns, there is a lot of internal logic and hooks to open / close remotely, if the id changes between the open and close events it can be very hard to debug, it has already happened.

Devessier: @lucasbordeau, I found a way to use an existing id everywhere. It continues to work correctly.

Form fields all use `useId` so I kept it here.

Devessier: Merging.

StephanieJoly4:
thank you @Devessier 
when will it please be in production so I can inform the user?

PR #16090, opened 2025-11-26
Improve deactivated objects & fields behaviors.

### LINKED ISSUE #1918: Improve deactivated objects & fields behaviors
# Desired Behavior

### 1. Indicated deactivated objects & fields in table

https://www.figma.com/design/xt8O9mFeLl46C5InWwoMrN/Twenty?node-id=82895-81423&t=G5cvpWztQvHtgQaQ-11

Image

### 2. Indicate if an object with a given name already exists in DB

(Don't display this banner for reserved name, this will be another issue)

Image

### 3. Add a delete button alongside "Disable"

Display a confirmation modal before deletion: “Are you sure you want to delete this object and all the data it contains?” Include a security prompt that asks the user to type the name of the object.

Image

https://www.figma.com/design/xt8O9mFeLl46C5InWwoMrN/Twenty?node-id=34855-565882&t=G5cvpWztQvHtgQaQ-11

mabdullahabaid:
Closes [1918](https://github.com/twentyhq/core-team-issues/issues/1918).

- For the first point in the issue, we just show the deactivated entries along with the deactivated text.

---

- For the second point, we show a banner and control the enabled/disabled state of save button depending on whether we're allowing the user to create table with the typed name. 
- For example, we do not want to allow the user to create a table with reserved name, so we disable the save button without showing a banner.
- Similarly, we do not want the user to create a table with a name that already exists in the database. In this case, we show a banner and we also disable the save button.
- Finally, we do not want to allow the user to create a table where singular and plural name are the same. Therefore, we disable the save button for names like `works`.

---

- For the third point, if we add the delete button, it logically means that we allow the user to delete a custom object/field even it has not been deactivated yet, so did that.
- Upon deleting the object/field, if we wait for the metadata to refetch before we navigate, this is what we see because the path does not exist any longer after deletion and we're waiting for refetch on the path until we navigate away.

https://github.com/user-attachments/assets/dbe0569c-db88-4285-851f-22551b1ca81e

- To avoid this page from appearing, I replaced awaiting refetch to not awaiting refetch and redirecting while the refetch happens in the background.
- Therefore, when we delete something, there is a slight delay for when it is actually cleared out from the list, but the Not Found view does not appear on the screen.

https://github.com/user-attachments/assets/47f49579-ce51-4d6a-b857-72046247bb4b

- I tried optimistically removing the object/field from the metadata, but it leads to some issues (crashes the app) and I have not been able to find a solution for it yet.
- Therefore, instead of getting stuck at perfection and blocking myself, I stopped getting into the issue further and created this PR by ensuring that the desired functionality works.

---

> [!NOTE]
> Show deactivated objects/fields in settings with badges and filters, add delete actions with confirmation, and centralize metadata name computation and reserved-name handling in shared utilities across front/back.
> 
> - **Frontend (Settings UI)**:
>   - Show deactivated state inline ("Deactivated" label) and default to displaying inactive items in objects and fields tables.
>   - Replace field row action menu with a chevron link; keep inactive dropdown for activate/delete.
>   - Add delete actions for custom objects/fields with confirmation modals; navigate immediately while refetch runs to avoid Not Found flashes.
>   - New object form: live conflict detection via `getConflictingObjectMetadataItem` with banner and save disabled on conflicts; validation uses non-throwing `computeMetadataNameFromLabel` wrapper.
>   - Minor: story decorators updated; removed `SettingsObjectFieldActiveActionDropdown`.
> - **Shared (`twenty-shared/metadata`)**:
>   - Add `computeMetadataNameFromLabel`, `RESERVED_METADATA_NAME_KEYWORDS`, and `sanitizeReservedKeyword`; export in `metadata` barrel; add deps `lodash.camelcase`, `transliteration`.
> - **Backend**:
>   - Migrate name computation and reserved keyword checks to `twenty-shared/metadata`; update services, decorators, and relation/morph utilities to use shared util.
>   - Validation: adjust reserved-name validator message; keep safety-net checks; allow deletion validation for non-standard active items (remove active-only deletion block).
> - **Tests/Snapshots**:
>   - Update unit/integration tests and snapshots to reflect new name computation, reserved-name messaging, and behaviors.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 840f68a1e754e644c44be30240fed3680854a2ed. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/twenty-front/src/modules/object-metadata/hooks/useDeleteOneFieldMetadataItem.ts
```diff
@@ -69,8 +69,8 @@ export const useDeleteOneFieldMetadataItem = () => {
       // TODO: see if we can remove this lin altogether
       await resetRecordIndexKanbanAggregateOperation(idToDelete);
 
-      await refreshObjectMetadataItems();
-      await refreshCoreViewsByObjectMetadataId(objectMetadataId);
+      void refreshObjectMetadataItems().catch(() => {});
```
FelixMalfait: Let's try not to shut down. errors like that, bad pattern

mabdullahabaid:
@FelixMalfait fixed the issues discussed on Discord.

For the 404 error, Cursor suggested lifting up the state to parent page instead of the child tab component, which got us the desired functionality without optimistic update to the state.

For reserved keywords, both objects and fields use the `Custom` suffix. Both object and field creation was blocked for reserved keywords, so it felt intuitive to allow for both, but open to feedback if we want to limit for objects only.

mabdullahabaid:
Fixed the issue reported by Cursor bot. 

I also realized that instead of lifting the entire confirmDelete function to the parent, a better approach would be to just lift the isDeleting state up while keeping the confirmDelete function in the child. Therefore, I made that fix.

packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-builder-v2/validators/services/flat-object-metadata-validator.service.ts
```diff
@@ -144,7 +144,11 @@ export class FlatObjectMetadataValidatorService {
         });
       }
 
-      if (!buildOptions.isSystemBuild && flatObjectMetadataToDelete.isActive) {
+      if (
```
FelixMalfait: Should this be removed entirely? I will remove it because I think this check does not make sense anymore

packages/twenty-shared/src/metadata/reserved-metadata-name-keywords.constant.ts
```diff
@@ -0,0 +1,69 @@
+export const CORE_OBJECT_NAMES = [
+  'approvedAccessDomain',
```
FelixMalfait, line 2: @martmull We should maybe in the future add default lint rules to twenty-apps? And have one that outlines reserved keywords? If you find this relevant I'll let you create an issue. Not sure how we could solve this long-term otherwise (maybe helpful message during sync? not sure what would happen today if you used a reserved keyword)

packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-builder-v2/validators/services/flat-object-metadata-validator.service.ts
mabdullahabaid: I agree. It is only blocking our ability to delete active standard objects after the update, but we already have a check to block the deletion of standard objects and we now allow the deletion of active custom objects, so this is redundant and does not add any value.

Deleting it and pushing a commit.

FelixMalfait (approved):
great work!

packages/twenty-shared/src/metadata/compute-metadata-name-from-label.util.ts
```diff
@@ -0,0 +1,26 @@
+import camelCase from 'lodash.camelcase';
+import { slugify } from 'transliteration';
+
+import { sanitizeReservedKeyword } from './sanitize-reserved-keyword.util';
+
+export const computeMetadataNameFromLabel = (label: string): string => {
```
prastoin, line 6: ***Remark:*** naming does reflect throw `computeMetadataNameFromLabelOrThrow`

packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-builder-v2/validators/services/flat-field-metadata-validator.service.ts
```diff
@@ -247,18 +247,6 @@ export class FlatFieldMetadataValidatorService {
       });
     }
 
-    if (
```
prastoin, line 250: ***Question:*** Why removing this validation check here ?

packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-builder-v2/validators/services/flat-field-metadata-validator.service.ts
```diff
@@ -247,18 +247,6 @@ export class FlatFieldMetadataValidatorService {
...
-      !parentObjectMetadataHasBeenDeleted
-    ) {
-      validationResult.errors.push({
-        code: FieldMetadataExceptionCode.INVALID_FIELD_INPUT,
-        message: "Active fields can't be deleted",
-        userFriendlyMessage: msg`Active fields cannot be deleted`,
-      });
-    }
-
     return validationResult;
```
prastoin, line 250: ***Question:*** Why removing this validation check here ?

packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-builder-v2/validators/services/flat-object-metadata-validator.service.ts
```diff
@@ -143,14 +143,6 @@ export class FlatObjectMetadataValidatorService {
           userFriendlyMessage: msg`Standard objects cannot be deleted`,
         });
       }
-
-      if (!buildOptions.isSystemBuild && flatObjectMetadataToDelete.isActive) {
```
prastoin, line 147: ***Question:*** Same here

packages/twenty-server/src/engine/metadata-modules/utils/validate-name-and-label-are-sync-or-throw.util.ts
```diff
@@ -1,12 +1,34 @@
...
 import { isDefined } from 'twenty-shared/utils';
 
 import {
   InvalidMetadataException,
   InvalidMetadataExceptionCode,
 } from 'src/engine/metadata-modules/utils/exceptions/invalid-metadata.exception';
 
+// Server-specific wrapper that converts generic errors to InvalidMetadataException
+// This provides consistent error handling with proper exception codes for the server
+export const computeMetadataNameFromLabel = (label: string): string => {
```
prastoin, line 11: ***Remark:*** Should be in a dedicated file

prastoin (commented):
Hey there, cool job over there ! Wondering the motivations behind v2 validation removals

packages/twenty-server/src/engine/metadata-modules/utils/validate-name-and-label-are-sync-or-throw.util.ts
FelixMalfait: Agree (but fyi this is being moved around in the same file not added debt)

packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-builder-v2/validators/services/flat-field-metadata-validator.service.ts
FelixMalfait: Please double check but the intent is to avoid forcing going through soft-delete since we now introduce a button to directly hard-delete objects (the deactivation process was painful / leading to confusion where users didn't realize a deactivated object still existed)

packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-builder-v2/validators/services/flat-field-metadata-validator.service.ts
FelixMalfait: see above

packages/twenty-shared/src/metadata/compute-metadata-name-from-label.util.ts
FelixMalfait: my bad, @mabdullahabaid could you please address this and the one export = one file comment in a followup PR? No need for a dedicated PR for that, you can just bring those changes into your ongoing PR

FelixMalfait: Actually I'll followup myself later, there is some debt here, I think the cleanup goes in the right direction but it needs more cleanup (we probably don't need to throw at all in the share util so naming is correct but function needs to be fixed)

packages/twenty-server/src/engine/metadata-modules/utils/validate-name-and-label-are-sync-or-throw.util.ts
prastoin: Moved in a dedicated file in https://github.com/twentyhq/twenty/pull/16230 + removed relicas

packages/twenty-shared/src/metadata/compute-metadata-name-from-label.util.ts
prastoin: No worries started to address comments in https://github.com/twentyhq/twenty/pull/16230
Also nothing critical was just investigating some validation assertion removal

packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-builder-v2/validators/services/flat-field-metadata-validator.service.ts
prastoin: Ok thanks for the info ! So the vision is to deprecated the isActive object/fields properties ?

packages/twenty-shared/src/metadata/compute-metadata-name-from-label.util.ts
FelixMalfait: Perfect thanks a lot @prastoin

packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-builder-v2/validators/services/flat-field-metadata-validator.service.ts
FelixMalfait: No we would keep both but now we directly give the choice to users in the UI

PR #16095, opened 2025-11-26
fix: update side panel header title to base font size with baseline alignment

Bonapara:
## Summary
1. Changed the side panel header title font size from small (0.92rem) to base (1rem)
2. Added baseline alignment between the title and subtitle text while keeping the icon centered

## Changes
- Updated `StyledPageInfoTitleContainer` font size from `theme.font.size.sm` to `theme.font.size.md`
- Added `StyledPageInfoTextContainer` wrapper to baseline-align title and subtitle independently from the icon

image


Weiko (approved):
👍

PR #16097, opened 2025-11-26
fix scheduling after folder actions are processed

neo773:


packages/twenty-server/src/modules/messaging/message-import-manager/services/messaging-message-list-fetch.service.ts
```diff
@@ -63,13 +63,29 @@ export class MessagingMessageListFetchService {
...
       if (pendingGroupEmailActionsProcessed) {
+        await this.messageChannelSyncStatusService.scheduleMessageListFetch([
+          messageChannel.id,
+        ]);
+
+        this.logger.log(
+          `WorkspaceId: ${workspaceId}, MessageChannelId: ${messageChannel.id} - Scheduled message list fetch after processing group email action`,
+        );
+
         return;
```
charlesBochet: why do we return and we don't let the function flow instead?

packages/twenty-server/src/modules/messaging/message-import-manager/services/messaging-message-list-fetch.service.ts
```diff
@@ -81,19 +73,42 @@ export class MessagingMessageListFetchService {
...
+        pendingGroupEmailActionsProcessed || pendingFolderActionsProcessed
+          ? await messageChannelRepository.findOne({
+              where: {
+                id: messageChannel.id,
+              },
+              relations: ['connectedAccount', 'messageFolders'],
+            })
+          : messageChannel;
+
+      if (!freshMessageChannel) {
```
charlesBochet: use isDefined

charlesBochet (approved):
left a minor comment LGTM

PR #16100, opened 2025-11-26
[DASHBOARDS] Add pie chart empty state

bosiraphael:
When no data is available, we display a gray circle instead of nothing.

https://github.com/user-attachments/assets/54db403b-739c-43fb-b422-8dcdfb6281fb


Weiko (approved):
LGTM

PR #16102, opened 2025-11-26
Add workspaceMemberId to workflow payload

### LINKED ISSUE #1946: Add metadata infos on manual trigger
## Edit

Add who triggered the workflow, when, etc.

## Scope & Context

**Example:** 
```
I am building a workflow automation that begins with a user input form. 

In the workflow i want to create and assign tasks to the workspace user who submitted the form. 

The input form does not output this workspace user's ID for later ingestion in the workflow.

If the input form outputs the workspace user id who submitted, i can associate all further records with them.

```

I want a "create a new lead" workflow that creates a company, person, opportunity and 5 tasks. All of which assigned to/tied back to the user who submitted the form. This way prospecting SDRs can easily and quickly add leads and have tasks automatically assigned to them.

## Current behavior

Right now the user input forms do not have a data output that includes the workspace user who submitted the form.

**Example:** 
```
The input form does not output this workspace user's ID for later ingestion in the workflow.
```

## Expected behavior
In the workflow editor the workspace user id who submitted the form should be an available field that we can use in other workflow steps so we can properly link any of those steps with the user who submitted the form.

**Example:** 
```
I create a manually triggered workflow with a user form input event.

i create a company and link to workspace user who submitted form as owner.

i create a person and link them to the company via the data output of the create company event.

I create an opportunity and link company from company output, person from person output, user from form output.

i  create 4 tasks for emails, cold calls, etc. and link the workspace user to them as owner from input form output.

```

## Technical inputs

**Example:** 
```
ensure that the form workflow object emits the workspace user who submitted the form as an available data field/output
```

saikumarvasa100-hash:
This change adds the workspaceMemberId to the workflow payload so that subsequent workflow steps can access the user who submitted the form/triggered the workflow.

Fixes twentyhq/core-team-issues#1946

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/workflow/resolvers/workflow-trigger.resolver.ts
+++ b/packages/twenty-server/src/engine/core-modules/workflow/resolvers/workflow-trigger.resolver.ts
@@ -83,7 +83,7 @@ export class WorkflowTriggerResolver {
     return this.workflowTriggerWorkspaceService.runWorkflowVersion({
       workflowVersionId,
       workflowRunId: workflowRunId ?? undefined,
-          payload: {
+        payload: {
           ...(payload ?? {}),
           workspaceMemberId: workspaceMember.id,
         },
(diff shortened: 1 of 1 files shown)


thomtrp:
Hi @saikumarvasa100-hash, thank you for your contribution. I asked clarification about the desired feature on the ticket. Your implementation does not work unfortunately because if the payload contains a field named `workspaceMemberId` it would be overridden. And it would not be available through variables anyway.

PR #16104, opened 2025-11-26
Update install.sh

simonbushell:
fixes Twenty Worker Unhealthy error when installed via 1-click docker

### DIFF
--- a/packages/twenty-docker/scripts/install.sh
+++ b/packages/twenty-docker/scripts/install.sh
@@ -93,6 +93,12 @@ fi
 echo "# === Randomly generated secret ===" >> .env
 echo "APP_SECRET=$(openssl rand -base64 32)" >> .env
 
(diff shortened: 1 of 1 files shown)


packages/twenty-docker/scripts/install.sh
```diff
@@ -93,6 +93,12 @@ fi
 echo "# === Randomly generated secret ===" >> .env
 echo "APP_SECRET=$(openssl rand -base64 32)" >> .env
 
+# Generate random strings for JWT authentication
```
FelixMalfait, line 96: Those are old deprecated variables that shouldn't be used anymore (APP_SECRET is used instead). Would you mind giving more detailed on how you encountered this error/on which version? Thanks!

charlesBochet:
I'm closing it in the meantime, let's keep discussing here and re-open if it makes sense :)

PR #16107, opened 2025-11-26
Fix options menu button height in side panel footer

Bonapara:
## Description
Fixed the height of the options menu button in the side panel footer to be 24px instead of 32px, matching the height of other buttons in the footer.

## Changes
- Added `size="small"` prop to the `Button` component in `OptionsDropdownMenu.tsx`
- This ensures consistent button sizing across the side panel footer

## Before
The options menu button was 32px high (medium size), making it larger than other footer buttons.

image

## After
The options menu button is now 24px high (small size), consistent with other side panel footer buttons.

CleanShot 2025-11-26 at 18 50 43@2x


ehconitin (approved):
LGTM. Please don't come after my job though 🙏

PR #16121, opened 2025-11-27
Update pie chart inner padding

bosiraphael:
## Before

CleanShot 2025-11-27 at 12 13 24@2x

## After

CleanShot 2025-11-27 at 12 13 04@2x


ehconitin (approved):
lgtm

PR #16132, opened 2025-11-27
optimize buildFieldMapsFromFlatObjectMetadata usages

Weiko:
This newly introduced util can be a bit expensive especially when done recursively.
This PR improves that

File | Pattern Fixed | Impact
-- | -- | --
format-result.util.ts | Recursive array/object processing | N array items → 1 call
format-data.util.ts | Recursive array processing | N array items → 1 call
process-nested-relations-v2.helper.ts | Duplicate call in call chain | 2 calls → 1 call
common-result-getters.service.ts | Per-record processing in array | N records → 1 call
compute-relation-connect-query-configs.util.ts | Nested loop (entities × connect fields) | N×M calls → 1 call


etiennejouan (approved):
LGTM

PR #16140, opened 2025-11-27
Currency input field - fix

etiennejouan:
Currency field used to have default value, but default value on field is not mandatory. Defaulf default value logic has been removed.

Also test all field type input when empty. ✅


bosiraphael (approved):
LGTM

PR #15926, opened 2025-11-19
common api - null equivalence

etiennejouan:
closes https://github.com/twentyhq/core-team-issues/issues/1629

To do before requesting review : 
- filter update

Migration to come in an other PR

Strat : 
1/  Null transformation

- [x] Transform NULL equivalent value to NULL in field validation in common api - pre-query - with feature flag
- [ ] Same logic in ORM (Not done, complex to handle feature flag here)
- [x] Transform NULL value to equivalent in data formatting in ORM - post-query

2/ Migration (in other PR) for fieldMetadata not nullable with default defaultValue (empty string, ...)

- [ ] Remove NOT NULL db constraint
- [ ] Update record value to NULL
- [ ] Update field metadata : isNullable:true
- [ ] Update uniqueIndex whereClause (also for standard uniqueIndex) 
- [ ] Activate feature flag

3/ Update metadata creation

- [x] No more default default value
- [x] Update standard field nullability
- [x] Remove index default whereClause for standard field

4/ Update filter

- [x] When filtering on NULL or empty string, be sure all records are returned (the one with NULL + the one with "")

5/ Test

- [ ] Strat. to do

PR #16144, opened 2025-11-27
Update workspace entities to make all TEXT nullable

charlesBochet:
Follow up on #15926


guillim (approved):
LGTM

packages/twenty-server/src/modules/messaging/common/standard-objects/message-folder.workspace-entity.ts
```diff
@@ -47,7 +47,8 @@ export class MessageFolderWorkspaceEntity extends BaseWorkspaceEntity {
     description: msg`Folder name`,
     icon: 'IconFolder',
   })
-  name: string;
+  @WorkspaceIsNullable()
+  name: string | null;
```
guillim, line 51: IMO we shoudl put a default value and & make it mandatory @charlesBochet WDYT ?

packages/twenty-server/src/modules/blocklist/standard-objects/blocklist.workspace-entity.ts
```diff
@@ -34,7 +35,8 @@ export class BlocklistWorkspaceEntity extends BaseWorkspaceEntity {
     description: msg`Handle`,
     icon: 'IconAt',
   })
-  handle: string;
+  @WorkspaceIsNullable()
+  handle: string | null;
```
guillim, line 39: IMO this should be mandatory, and we should declare a default value to  backfill. But it makes no sense to have a "bloclist" without the handle  @charlesBochet

packages/twenty-server/src/modules/workflow/common/standard-objects/workflow-version.workspace-entity.ts
```diff
@@ -90,7 +90,8 @@ export class WorkflowVersionWorkspaceEntity extends BaseWorkspaceEntity {
     description: msg`The workflow version name`,
     icon: 'IconSettingsAutomation',
   })
-  name: string;
+  @WorkspaceIsNullable()
```
etiennejouan, line 93: each time a version is created, a name is filled. It could remain not nullable (without default value - but it would cause null command operation then sync operations)

guillim (approved):

packages/twenty-server/src/database/commands/upgrade-version-command/1-10/1-10-migrate-attachment-type-to-file-category.command.ts
```diff
@@ -62,6 +63,10 @@ export class MigrateAttachmentTypeToFileCategoryCommand extends ActiveOrSuspende
     for (const attachment of attachments) {
       const { id, type } = attachment;
 
+      if (!isDefined(type)) {
```
charlesBochet, line 66: attachment.type default?

packages/twenty-server/src/modules/messaging/message-import-manager/drivers/imap/providers/imap-client.provider.ts
```diff
@@ -83,13 +83,26 @@ export class ImapClientProvider {
     let client: ImapFlow | null = null;
     let timeoutId: NodeJS.Timeout | null = null;
 
+    if (!isDefined(connectionParameters.IMAP?.username)) {
+      throw new CustomError(
```
charlesBochet: throw MessageImportManagerException ?

packages/twenty-server/src/modules/messaging/message-import-manager/drivers/microsoft/services/microsoft-get-message-list.service.ts
```diff
@@ -115,6 +117,13 @@ export class MicrosoftGetMessageListService {
       this.microsoftMessageListFetchErrorHandler.handleError(error);
     });
 
+    if (!isDefined(messageFolder.syncCursor)) {
+      throw new CustomError(
```
charlesBochet: same

packages/twenty-server/src/modules/messaging/common/standard-objects/message-folder.workspace-entity.ts
Weiko: I'm not sure we should, what would be the default value then? 🤔 
However I agree on the fact that it should be mandatory, why are we setting "ALL" TEXT as nullable, I feel like some could be non-nullable (in this case for example) 🤔 @charlesBochet

packages/twenty-server/src/modules/messaging/message-import-manager/services/messaging-messages-import.service.ts
```diff
@@ -114,11 +117,27 @@ export class MessagingMessagesImportService {
         workspaceId,
       );
 
+      if (!isDefined(messageChannel.handle)) {
+        throw new CustomError(
```
charlesBochet: let's use the existing MessageImportDriverException everywhere in these files

PR #16146, opened 2025-11-27
Optimize EntityMetadata caching in GlobalWorkspaceDataSource

Weiko:
## Context
EntityMetadata was being rebuilt from scratch on every findMetadata()/getMetadata() call (~20 times per request). This involved running EntitySchemaTransformer.transform() and EntityMetadataBuilder.build() repeatedly, causing unnecessary CPU overhead.

## Implementation
Cache entityMetadatas in ORMWorkspaceContext: Build EntityMetadata once during workspace context initialization instead of on every metadata lookup
Remove redundant entitySchemas caching: Since flatMetadata is already cached, the additional Redis cache for entitySchemaOptions was unnecessary overhead
Remove WorkspaceEntitiesStorage: Replaced with direct lookup from FlatObjectMetadataMap
Simplify getObjectMetadataFromEntityTarget: Now only accepts string targets, using flat metadata maps directly

Also:
Removed unused injections in some services


packages/twenty-server/src/engine/twenty-orm/utils/get-object-metadata-from-entity-target.util.ts
```diff
@@ -11,27 +7,20 @@ import {
   TwentyORMException,
   TwentyORMExceptionCode,
 } from 'src/engine/twenty-orm/exceptions/twenty-orm.exception';
-import { WorkspaceEntitiesStorage } from 'src/engine/twenty-orm/storage/workspace-entities.storage';
 
 export const getObjectMetadataFromEntityTarget = <T extends ObjectLiteral>(
   entityTarget: EntityTarget<T>,
   internalContext: WorkspaceInternalContext,
 ): FlatObjectMetadata => {
-  const objectMetadataName =
```
Weiko, line 20: I've double checked, it seems this is never reached as an entityTarget but always as a string. I'm keeping the signature as it is for now and throwing to catch early if I'm wrong but otherwise this should simplify this part where WorkspaceEntitiesStorage can be completely deprecated

ijreilly (approved):

PR #992, opened 2023-07-30
Offline screen

Mbistami:
fix #454

PR #16152, opened 2025-11-27
fix: add Node.js 25 compatibility via jwa resolution

### LINKED ISSUE #16148: The backend server crashes on Node.js v25.x because buffer-equal-constant-time (via jsonwebtoken → jwa) uses SlowBuffer.prototype, which was removed in Node 25.
Error:
  TypeError: Cannot read properties of undefined (reading 'prototype')
      at .../buffer-equal-constant-time/index.js:37:35

  Environment: Node v25.1.0, macOS, commit 3b8db734a5

  Root cause: Node 25 removed SlowBuffer. The dependency chain jsonwebtoken → jwa → buffer-equal-constant-time uses this deprecated API.

  Workaround: Use Node 24.x (nvm use 24)

  Suggested fixes:
  1. Update jsonwebtoken to a version without this dependency
  2. Add runtime Node version check with helpful error message
  3. The existing yarn.config.cjs constraint (^24.5.0) only warns at install, not runtime

austinmao:
## Summary

- Adds Node.js 25 compatibility by resolving `jwa` to version 2.0.1
- Updates engine constraint from `^24.5.0` to `>=24.5.0`

Fixes #16148

## Problem

Node.js 25 removed the `SlowBuffer` API, which breaks the `jsonwebtoken` dependency chain:

```
jsonwebtoken → jws → jwa → buffer-equal-constant-time
```

The `buffer-equal-constant-time@1.0.1` package (last updated 12 years ago) uses `SlowBuffer.prototype.equal`, causing this error on Node 25:

```
TypeError: Cannot read properties of undefined (reading 'prototype')
    at .../buffer-equal-constant-time/index.js:37:35
```

## Solution

- Add yarn resolution for `jwa@2.0.1` which uses `crypto.timingSafeEqual()` instead of the deprecated `SlowBuffer` API
- Relax Node engine constraint to allow Node 25+

## References

- [Node 24 compatibility · Issue #992 · auth0/node-jsonwebtoken](https://github.com/auth0/node-jsonwebtoken/issues/992)
- [Node.js v25.0.0 Release Notes](https://nodejs.org/en/blog/release/v25.0.0) - SlowBuffer removal

## Test plan

- [x] Verified backend starts successfully on Node.js v25.1.0
- [x] Verified `yarn install` completes without errors
- [x] Verified `yarn start` runs frontend and backend

### DIFF
--- a/package.json
+++ b/package.json
@@ -194,7 +194,7 @@
     "vite-plugin-svgr": "^4.2.0"
   },
   "engines": {
-    "node": "^24.5.0",
+    "node": ">=24.5.0",
     "npm": "please-use-yarn",
     "yarn": ">=4.0.2"
   },
@@ -207,7 +207,8 @@
     "typescript": "5.9.2",
     "graphql-redis-subscriptions/ioredis": "^5.6.0",
     "prosemirror-view": "1.40.0",
-    "prosemirror-transform": "1.10.4"
+    "prosemirror-transform": "1.10.4",
+    "jwa": "2.0.1"
   },
   "version": "0.2.1",
   "nx": {},
--- a/yarn.lock
+++ b/yarn.lock
@@ -28791,7 +28791,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"buffer-equal-constant-time@npm:1.0.1":
+"buffer-equal-constant-time@npm:^1.0.1":
   version: 1.0.1
   resolution: "buffer-equal-constant-time@npm:1.0.1"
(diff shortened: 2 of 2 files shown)


FelixMalfait:
Hey would it be possible to upgrade jsonwebtoken instead? We try to avoid forcing second-order resolutions, then it creates confusion in the package.json (where we just want first-level dependencies ideally). Thanks!

charlesBochet (changes requested):
agree with @FelixMalfait comment

FelixMalfait:
Closing since this did not receive any answer, thanks for the attempt!

PR #16184, opened 2025-11-28
Augment chart limits and improve padding on bar chart

bosiraphael:
- Maximum from 50 to 100
- Reduce padding
- Make inner padding dynamic


Weiko (approved):
LGTM

PR #16209, opened 2025-12-01
Null equivalence - Activate FF for all

etiennejouan:


Weiko (commented):
PR's name implies you want to activate this for all workspaces however this will only have an effect on dev seeder. Is it expected?

etiennejouan:
> PR's name implies you want to activate this for all workspaces however this will only have an effect on dev seeder. Is it expected?

@Weiko Thanks ! You're right, I'm fixing it

Weiko (approved):
LGTM

PR #16211, opened 2025-12-01
[Dashboards] - fast follows - inverse default value for centre metric and filter count on chart settings

ehconitin:


bosiraphael (approved):
LGTM

PR #16216, opened 2025-12-01
[DASHBOARDS] Add prefix and suffix setting to the aggregate chart

### LINKED ISSUE #1969: [DASHBOARDS] Add prefix and suffix setting in aggregate chart
Image

https://www.figma.com/design/xt8O9mFeLl46C5InWwoMrN/Twenty?node-id=79903-251313&m=dev

bosiraphael:
https://github.com/user-attachments/assets/0103cebe-e426-42d0-a8e4-b82494ef8503


packages/twenty-front/src/modules/command-menu/pages/page-layout/components/ChartSettings.tsx
```diff
@@ -208,6 +216,7 @@ export const ChartSettings = ({ widget }: { widget: PageLayoutWidget }) => {
                     getChartSettingsValues={getChartSettingsValues}
                     onToggleChange={handleItemToggleChange}
                     onInputChange={handleItemInputChange}
+                    onTextInputChange={handleItemTextInputChange}
```
charlesBochet: @bosiraphael general note, it seems that we have a hook to deal with setting update useUpdateChartSettingTextInput, why drilling it down instead of calling it in the child?

charlesBochet (approved):
Left minor comments

packages/twenty-front/src/modules/command-menu/pages/page-layout/components/ChartSettings.tsx
bosiraphael: Good point indeed

PR #16217, opened 2025-12-01
Null equivalence - Empty string default value cleaning

etiennejouan:
- on name standard field on custom object (field metadata + db)
- on text standard fields on standard object (db only)


charlesBochet (approved):

packages/twenty-server/src/engine/metadata-modules/object-metadata/utils/build-default-flat-field-metadatas-for-custom-object.util.ts
Weiko, line 63: Good catch!

Weiko (approved):
LGTM!

PR #16221, opened 2025-12-01
fix(theme): prevent forced light mode switch after login

Anshgrover23:
### Description

This pull request resolves an issue where the application switched to light mode after login even when the user’s system was set to dark mode. Before authentication the app correctly followed system preferences through CSS media queries, but that behavior broke once the user logged in.

**Before**

https://github.com/user-attachments/assets/75e73712-9cb2-42f9-9e25-3a22dc911b8d

**After**

https://github.com/user-attachments/assets/6bba45ce-3817-4e3d-9b45-ff0c7725cf82


charlesBochet (approved):
LGTM indeed, ty!

Issue #11977, opened 2025-05-11
Wrong number of records in deleted record details


BOHEUS:
**Scenario:**
1. Log in
2. Go to People tab (or any other tab)
3. Delete record
4. Open side menu and click "See deleted records"
5. Check deleted record details

**Actual:** Numbers are 0 / (number of all records)

https://github.com/user-attachments/assets/f6901037-4143-4207-91a0-37c5c116d5db

**Expected:** Numbers should be (record index) / (number of all deleted records)

Bonapara:
@thomtrp related to Restore cascade

lucasbordeau:
@charlesBochet @Weiko We would need a ranking function backend side, that allows to get the rank of a given record id in a filtered view.

zeus1959:
@BOHEUS, can you assign me on this issue?

Bonapara:
no assignment @zeus1959 :)

PR #16231, opened 2025-12-01
fix: updated record calculations to display the correct record index

muskanmm:
Issue Number: https://github.com/twentyhq/twenty/issues/11977

## Context: 
As shown in the issue description, the record details for each person were incorrect. For each person, no matter their record index, it would display `0 / (number of all records)`. The intended behavior is to display `(record index) / (number of all records)`. This logic should extend to the deleted records page as well. 

## Changes Made: 
The rankInView was previously being calculated using the cache but this was constantly returning -1. I updated the calculation to query all the records matching the filter and then using `findIndex()` to find the rank in the view. 

## Correct Output: 

https://github.com/user-attachments/assets/48e7aee5-7afe-4f64-8f5f-23bd721a0437

https://github.com/user-attachments/assets/43f9aee6-e35d-4787-8344-bb52b4b23a15


lucasbordeau:
I'm closing this PR as it if open for a long time and this current implementation would involve heavy requests on our servers. 

A backend solution might be preferable but it is not an easy problem to tackle.

Issue #16203, opened 2025-11-30
Add support for docker's secrets feature


frenchki:
## Scope & Context

To make it easier to user the docker secrets feature when deploying for example, a docker swarm it would be great to be able to read secrets from file. If we look at dockers official images this can be implemented in the `docker-entrypoint.sh`

[Postgres docker-entrypoint.sh](https://github.com/docker-library/postgres/blob/master/docker-entrypoint.sh#L5-L25)

[Docker secrets](https://docs.docker.com/engine/swarm/secrets/#use-secrets-in-compose)

## Current behavior

Currently, the secrets can only be read from environment variables that are not files, making it impossible to use this feature in Docker.

## Expected behavior

```yml
services:
  twenty-crm:
    image: twentycrm/twenty:v1.11.5
    volumes:
      - twenty-crm-data:/app/packages/twenty-server/.local-storage
    environment:
      # Twenty CRM postgres secrets (being read from file)
      TWENTY_CRM_POSTGRES_HOST: /run/secrets/TWENTY_CRM_POSTGRES_HOST
      TWENTY_CRM_POSTGRES_PORT: /run/secrets/TWENTY_CRM_POSTGRES_PORT
      TWENTY_CRM_POSTGRES_DATABASE: /run/secrets/TWENTY_CRM_POSTGRES_DATABASE
      TWENTY_CRM_POSTGRES_DATABASE_USER: /run/secrets/TWENTY_CRM_POSTGRES_DATABASE_USER
      TWENTY_CRM_POSTGRES_DATABASE_PASSWORD: /run/secrets/TWENTY_CRM_POSTGRES_DATABASE_PASSWORD
      # Twenty CRM redis secrets
      TWENTY_CRM_REDIS_URL: /run/secrets/TWENTY_CRM_REDIS_URL
      # Twenty CRM application secrets
      TWENTY_CRM_SERVER_URL: /run/secrets/TWENTY_CRM_SERVER_URL
      TWENTY_CRM_APP_SECRET: /run/secrets/TWENTY_CRM_APP_SECRET
      NODE_PORT: 3000
      DISABLE_DB_MIGRATIONS: "false"
      DISABLE_CRON_JOBS_REGISTRATION: "false"
      STORAGE_TYPE: local
    secrets:
      # Twenty CRM postgres secrets
      - TWENTY_CRM_POSTGRES_HOST
      - TWENTY_CRM_POSTGRES_PORT
[...]

PR #16232, opened 2025-12-01
feat(docker): add Docker secrets support to entrypoint

frenchki:
Add support for Docker secrets by implementing _FILE suffix convention for sensitive environment variables. This allows values to be read from files (typically mounted at /run/secrets/) instead of being passed directly as environment variables.

The file_env() function checks for both direct env vars and _FILE variants, reads the file content when the _FILE variant is set, and ensures both aren't set simultaneously to avoid conflicts.

Supported variables with _FILE suffix:
- PG_DATABASE_URL
- REDIS_URL
- APP_SECRET
- AUTH_GOOGLE_CLIENT_ID/SECRET
- AUTH_MICROSOFT_CLIENT_ID/SECRET
- CAPTCHA_SITE_KEY/SECRET_KEY
- ENTERPRISE_KEY
- CLOUDFLARE_API_KEY/WEBHOOK_SECRET
- CLICKHOUSE_URL
- And other authentication/integration credentials

This follows Docker best practices for managing secrets in Swarm mode and Compose, improving security by avoiding secrets in environment variables or compose files.

Fixes issue [16203](https://github.com/twentyhq/twenty/issues/16203)

References:
[Postgres docker-entrypoint.sh](https://github.com/docker-library/postgres/blob/master/docker-entrypoint.sh)
[Mysql docker-entrypoint.sh](https://github.com/docker-library/mysql/blob/master/docker-entrypoint.sh)
[Docker Secrets in MySQL docker image](https://hub.docker.com/_/mysql#docker-secrets)

### DIFF
--- a/packages/twenty-docker/twenty/entrypoint.sh
+++ b/packages/twenty-docker/twenty/entrypoint.sh
@@ -1,6 +1,28 @@
 #!/bin/sh
 set -e
 
+# usage: file_env VAR [DEFAULT]
+#    ie: file_env 'XYZ_DB_PASSWORD' 'example'
+# (will allow for "$XYZ_DB_PASSWORD_FILE" to fill in the value of
+#  "$XYZ_DB_PASSWORD" from a file, especially for Docker's secrets feature)
+file_env() {
+	local var="$1"
+	local fileVar="${var}_FILE"
+	local def="${2:-}"
+	if [ "${!var:-}" ] && [ "${!fileVar:-}" ]; then
+		printf >&2 'error: both %s and %s are set (but are exclusive)\n' "$var" "$fileVar"
+		exit 1
+	fi
+	local val="$def"
+	if [ "${!var:-}" ]; then
+		val="${!var}"
+	elif [ "${!fileVar:-}" ]; then
+		val="$(< "${!fileVar}")"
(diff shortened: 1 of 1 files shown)


charlesBochet (approved):
ok, lgtm :)

charlesBochet (changes requested):
I stamped a bit too fast!

@frenchki I'm OK with the idea as this does not harm the self-host experience for developers that don't want to use this feature. 

1) The only pitfall I see is that Twenty has quite a lot of environment variable and the list `local env_vars` will be quite outdated quickly. Ideally we would like to extract this list from `config-variables.ts` but this looks a bit tricky. 

Another idea, suffix environment variables with _FILE (like in the link from docker swarm dock), get all env variables and select all the one ending with _FILE

2) There is also a typo in your entrypoint.sh making the CI red.

3) Please make sure that twenty docker setup is working as expected with _FILE and without _FILE env variables

charlesBochet:
@frenchki I'm closing in the meantime to only keep active PRs open in the repository but feel free to re-open it and happy to keep moving forward on this

frenchki:
Thank you for all the comments. Will get back to this when I have time 😄

PR #16233, opened 2025-12-02
Workspace member permission tab.

mabdullahabaid:
Introduce a read-only view of permissions similar to agent roles tab. The role selector in the following image feels as if it would lead to a new page, which is not what we want.

Permissions (If easier V1)

Therefore, I added a Select component for now and would love some clarity on what's ideal.

image

Also, the "Add Rule" button changes to enabled when we switch the role from `Admin` to `Member` and clicking it redirects to `/settings/roles/:role-id/add-object-permission`. Do we want to disable the button completely?

One final thing: SettingsAgentRoleTab already re-uses SettingsRolePermissions. I created MemberPermissionsTab to do the same. I don't think we need an abstracted shared component here since both tabs rely on the same shared child anyway. However, if we need a refactor, I can use some direction on how the code should look.

Creating this PR as draft since I think there might be a couple suggested changes on this.

### DIFF
--- a/packages/twenty-front/src/modules/settings/members/components/MemberPermissionsTab.tsx
+++ b/packages/twenty-front/src/modules/settings/members/components/MemberPermissionsTab.tsx
@@ -1,0 +1,127 @@
+import { SettingsRolePermissions } from '@/settings/roles/role-permissions/components/SettingsRolePermissions';
+import { type RoleWithPartialMembers } from '@/settings/roles/types/RoleWithPartialMembers';
+import { useSnackBar } from '@/ui/feedback/snack-bar-manager/hooks/useSnackBar';
+import { Select } from '@/ui/input/components/Select';
+import { type WorkspaceMember } from '@/workspace-member/types/WorkspaceMember';
+import styled from '@emotion/styled';
+import { t } from '@lingui/core/macro';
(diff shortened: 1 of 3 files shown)


charlesBochet:
I'm not sure about the UX here. 

@Bonapara could you take a look?

Bonapara:
@mabdullahabaid Great idea! We should add a confirmation modal too, since it's very sensitive. Also, let's use "Open in Roles" on the button instead of just "Open".

CleanShot 2025-12-02 at 12 11 02

image


https://www.figma.com/design/PNBfTgOVraw557OXChYagk/Explo?node-id=31585-80181&t=aC6igC2zpoeOMNik-11

mabdullahabaid:
@charlesBochet ready for review.

Weiko (approved):
Nice! LGTM 👍

Weiko:
This is already the case on role setting page. What do we want for mobile view here @Bonapara 
Screenshot 2025-12-02 at 14 56 43

Bonapara:
Let’s truncate the text and display a tooltip on hover. The tooltip should be a global pattern, not just for this case. If it’s not already implemented, let’s keep it for another PR.

Weiko:
Thanks! FYI @mabdullahabaid if you are interested!

PR #16239, opened 2025-12-02
Add allow same origin to the iFrame widget

bosiraphael:
Some iFrames were unusable because of this


prastoin (commented):
Hey there, that's pretty risky
I would either opt to pass credentials to the invoked iframe rather than consuming its own cookies 🤔 
I honestly don't know what's the most suitable here

charlesBochet (approved):
IMO this is completely safe for two reasons:
- users are responsible to what they install / setup in their workspace. If they want to hack themselves, they can. (instead of hacking themselves through an iframe, let's just make a chrome extension :p)
- none of the browsers allow any scripting / sharing between a host and an iframe having different hosts (scheme + subdomain + domain + protocol have to be exactly equal). So the only way to hack my-twenty.twenty.com is to host an iframe on... my-twenty.twenty.com), which lead to "hack myself" scenario too

PR #16241, opened 2025-12-02
Null - Second command - Cleaning remaining empty values

etiennejouan:
Because of [this code I forget to clean](https://github.com/twentyhq/twenty/pull/16217), 
- __workspace created between 1.12.0 and 1.12.2 (from friday 28 nov PM > monday 1 dec PM)__ have standard objects with empty string default value set on related TEXT type column (but default value null on field metadata) (ex: jobTitle on person)
- __custom objects created between 1.12.0 and 1.12.2 (from friday 28 nov PM > monday 1 dec PM)__ have "name" TEXT field with empty string default value set + NOT NULL constraint on column

Command cleans data and updates table structure

### DIFF
--- a/packages/twenty-server/src/database/commands/upgrade-version-command/1-13/1-13-clean-empty-string-null-in-text-fields.command.ts
+++ b/packages/twenty-server/src/database/commands/upgrade-version-command/1-13/1-13-clean-empty-string-null-in-text-fields.command.ts
@@ -53,6 +53,26 @@ export class CleanEmptyStringNullInTextFieldsCommand extends ActiveOrSuspendedWo
       );
     }
 
+    const workspace = await this.workspaceRepository.findOne({
+      where: { id: workspaceId },
+    });
+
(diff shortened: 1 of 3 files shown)


packages/twenty-server/src/database/commands/upgrade-version-command/1-13/1-13-clean-empty-string-null-in-text-fields.command.ts
```diff
@@ -0,0 +1,200 @@
...
+        `Workspace ${workspaceId} is not a v1.12.0 or v1.12.1 workspace, skipping`,
+      );
+
+      return;
+    }
+
+    const schemaName = getWorkspaceSchemaName(workspaceId);
+
+    if (isDryRun) {
+      this.logger.log('Dry run mode: No changes will be applied');
```
charlesBochet, line 79: return

charlesBochet (approved):

PR #16249, opened 2025-12-02
Fix SDK/CreateApp CI changed-files-check

Weiko:


.github/workflows/ci-sdk.yaml
```diff
@@ -20,7 +20,6 @@ jobs:
     with:
       files: |
         packages/twenty-sdk/**
-        packages/twenty-server/**
```
martmull, line 23: I think we should keep that

martmull (commented):
thank you

charlesBochet (approved):

PR #16266, opened 2025-12-02
Add message channel reset command

neo773:
In this PR, we are adding a new command to reset a message channel.

We are also refactoring a bit cursor reset as we had multiple implementations at different places in the code base


packages/twenty-server/src/modules/messaging/message-cleaner/commands/messaging-reset-channel.command.ts
```diff
@@ -0,0 +1,199 @@
...
+  ): Promise<void> {
+    const { workspaceId, messageChannelId } = options;
+
+    const messageChannelRepository =
+      await this.twentyORMGlobalManager.getRepositoryForWorkspace<MessageChannelWorkspaceEntity>(
+        workspaceId,
+        'messageChannel',
+      );
+
+    if (messageChannelId) {
```
charlesBochet: isDefined everwyhere

packages/twenty-server/src/modules/messaging/message-cleaner/commands/messaging-reset-channel.command.ts
```diff
@@ -0,0 +1,199 @@
...
+        'messageChannel',
+      );
+
+    const messageChannelMessageAssociationRepository =
+      await this.twentyORMGlobalManager.getRepositoryForWorkspace<MessageChannelMessageAssociationWorkspaceEntity>(
+        workspaceId,
+        'messageChannelMessageAssociation',
+      );
+
+    const associations = await messageChannelMessageAssociationRepository.find({
```
charlesBochet: this whole cleaning logic should be part of a service and be re-usable in this command but also when we remove a message channel from the UI. 
We already have a MessagingConnectedAccountDeletionCleanupJob that's doing:  await this.messageCleanerService.cleanOrphanMessagesAndThreads(
      data.workspaceId,
    );
We should try to unify

charlesBochet (approved):
Left comment but approving

PR #16275, opened 2025-12-03
Upsert based on ID field

### LINKED ISSUE #15929: [Workflow] Create or Update node - the ID cannot be specify which leads to creating duplicates
## Bug Description
When there is a field set as unique, it appears at the top
Image

When there is none, users cannot specify the id of the record being created or updated (in case there is an existing id). 
Image
Image

That would mean that this node is only usable for objects for which users defined a unique field, which I don't think should be enforced given the id is sometimes enough.

**Ask**
Can we let user specify the id of the record as an input? when known, this means this is an update. if not, this is a create. 

Let me know if this is supposed to be used differently but I don't see how at the moment which leads to creating duplicates

thomtrp:
Capture d’écran 2025-12-03 à 11 49 39

- Split CreateRecordBody in two, there are too many differences with CreateRecord
- Add ID field to the upsert multiselect
- Provide the Record id field in upsert action so the user can edit

Fixes https://github.com/twentyhq/twenty/issues/15929


martmull (approved):
lgtm

charlesBochet (approved):
Duplicated code that could be factorized in a common hook (handleFieldChange)

PR #16287, opened 2025-12-03
Add local only cache to cache service and cache typeorm entity metadata

Weiko:
## Problem

buildEntityMetadatas in GlobalWorkspaceOrmManager is computationally expensive and was running on every executeInWorkspaceContext call. This method uses TypeORM's EntitySchemaTransformer and EntityMetadataBuilder to build metadata for all workspace entities (30-50+ objects with many fields each).

The resulting EntityMetadata[] is not serialisable which means it cannot be cached in Redis because they contain:
- Circular references
- Functions/methods
- References to the DataSource instance

## Solution

Extended the workspace cache system to support local-only caching, then created a cache provider for entityMetadatas.

## Implementation details
Updated @WorkspaceCache decorator (workspace-cache.decorator.ts)
- Added localOnly?: boolean option to skip Redis storage for non-serializable data
Created WorkspaceEntityMetadatasCacheService
- Computes entity metadatas from DB to avoid race condition, this is acceptable
Simplified GlobalWorkspaceOrmManager
- Now fetches entityMetadatas from cache instead of rebuilding on every call
Updated Workspace migration runner - the only entry point where metadata can change
- Now invalidate the new 'entityMetadata' local cache when shouldIncrementMetadataGraphqlSchemaVersion is true (== field/object mutations)


packages/twenty-server/src/engine/twenty-orm/global-workspace-datasource/workspace-entity-metadatas-cache.service.ts
```diff
@@ -0,0 +1,138 @@
...
+import { type FlatEntityMaps } from 'src/engine/metadata-modules/flat-entity/types/flat-entity-maps.type';
+import { type FlatFieldMetadata } from 'src/engine/metadata-modules/flat-field-metadata/types/flat-field-metadata.type';
+import { type FlatObjectMetadata } from 'src/engine/metadata-modules/flat-object-metadata/types/flat-object-metadata.type';
+import { ObjectMetadataEntity } from 'src/engine/metadata-modules/object-metadata/object-metadata.entity';
+import { EntitySchemaFactory } from 'src/engine/twenty-orm/factories/entity-schema.factory';
+import { GlobalWorkspaceDataSourceService } from 'src/engine/twenty-orm/global-workspace-datasource/global-workspace-datasource.service';
+import { WorkspaceCache } from 'src/engine/workspace-cache/decorators/workspace-cache.decorator';
+
+@Injectable()
+@WorkspaceCache('entityMetadatas', { localOnly: true })
```
charlesBochet: ormEntityMetadata

packages/twenty-server/src/engine/workspace-cache/services/workspace-cache.service.ts
```diff
@@ -134,307 +157,286 @@ export class WorkspaceCacheService implements OnModuleInit {
...
+  private deleteFromLocalCache(
+    workspaceId: string,
+    cacheKeyNames: WorkspaceCacheKeyName[],
+  ): void {
+    for (const keyName of cacheKeyNames) {
+      const localKey = this.buildCacheKey(workspaceId, keyName);
+      const entry = this.localCache.get(localKey);
+
+      if (isDefined(entry)) {
+        entry.lastCheckedAt = 0;
```
charlesBochet: I don't think this works!

charlesBochet (approved):

PR #16290, opened 2025-12-03
Twenty sdk follow up

martmull:
as title


packages/create-twenty-app/src/utils/app-template.ts
```diff
@@ -125,7 +126,7 @@ const createPackageJson = async ({
       auth: 'twenty auth login',
     },
     dependencies: {
-      'twenty-sdk': '0.1.1',
+      'twenty-sdk': sdkPackageJson.version,
```
prastoin: Shouldn't we be publishing a version of the twenty create app everytime we release a sdk version anw ?

packages/create-twenty-app/src/cli.ts
```diff
@@ -1,13 +1,8 @@
 #!/usr/bin/env node
 import chalk from 'chalk';
 import { Command, CommanderError } from 'commander';
-import { readFileSync } from 'fs';
-import { join } from 'path';
 import { CreateAppCommand } from './create-app.command';
-
```
prastoin, line 7: nice

packages/twenty-sdk/package.json
```diff
@@ -69,5 +67,8 @@
   "engines": {
     "node": "^24.5.0",
     "yarn": "^4.0.2"
+  },
+  "typesVersions": {
```
prastoin: ***Polish:*** Could be removed

prastoin (approved):
Cool !!

packages/create-twenty-app/src/utils/app-template.ts
martmull: Indeed, i was wondering. next does that actually

packages/twenty-sdk/package.json
martmull: oups forget that one

packages/create-twenty-app/src/utils/app-template.ts
martmull: good remark, let's see that in another PR

Issue #16286, opened 2025-12-03
Clicking on "See runs" shows executions from all workflows


Devessier:
## Bug Description

When I click on "See runs", I see workflow runs from many different workflows.

https://github.com/user-attachments/assets/25874e0b-ae79-4d25-ae46-6a03e5a96061

The link of the "See runs" button is `https://twenty-eng.twenty.com/objects/workflowRuns?filter%5Bworkflow%5D%5BIS%5D%5BselectedRecordIds%5D%5B0%5D=47cd9272-1243-4c36-b45b-f9893accb903`.

However, once I click on the link, the URL becomes `https://twenty-eng.twenty.com/objects/workflowRuns`.

## Expected behavior

I would expect seeing only executions from the workflow I was viewing.

abk404:
This is same as https://github.com/twentyhq/twenty/issues/16229 and is now fixed.

charlesBochet:
Thank you! Closing

PR #16300, opened 2025-12-03
fix(workflow): clicking on 'See runs' shows executions from all workflows

### LINKED ISSUE #16229: [Workflows] Runs shown from a given workflow are no longer filtered to this workflow
## Bug Description

When you click on See Runs from a workflow, it used to show only the runs of this workflow (which was helpful and what a user would expect).
Now you click, you see all runs from all workflows
cc @thomtrp

abk404:
Fixes: #16229 and #16286

This PR fixes the bug where click on 'See All Runs' button for a specific workflow was not filtering the runs for that speciific workflow. 

**Reason** : The filter logic in `getFilterFilterableFieldMetadataItems` was excluding ALL system fields(except field id) from being filterable, which prevented these essential business relations from being used in filters. 

**Fix** : Added a targeted exception to allow specific system relation fields (`workflow` and `workflowVersion`) to be filterable, similar to how the `id` system field is already whitelisted.

### Before

https://github.com/user-attachments/assets/50b98626-ccdc-468a-96f9-0c911fec9a39

Screenshot 2025-12-03 at 11 52 50 PM

### After

https://github.com/user-attachments/assets/856bbd61-4958-43de-8fa5-3f52ca9e7260

Screenshot 2025-12-04 at 12 33 32 AM

### DIFF
--- a/packages/twenty-front/src/modules/object-metadata/utils/getFilterFilterableFieldMetadataItems.ts
+++ b/packages/twenty-front/src/modules/object-metadata/utils/getFilterFilterableFieldMetadataItems.ts
@@ -11,6 +11,10 @@ export const getFilterFilterableFieldMetadataItems = ({
     const isFieldActive = field.isActive;
     const isIdField = field.name === 'id';
 
+    const isWorkflowRelationField =
+      field.type === FieldMetadataType.RELATION &&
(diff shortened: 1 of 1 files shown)


charlesBochet (approved):
ty!

PR #16303, opened 2025-12-03
fix : slash menu overflow causing page scroll

kpark1208:
Issue : #15794 

## Context
As shown in the linked issue, slash menu dropdown for rich text editor at bottom of notes editor screen extended beyond the page container, causing main page to scroll.

## Changes Made
Added floating UI's `flip()` middleware to the component in `CustomSlashMenu.tsx` so that the menu automatically flips upward when detecting not enough space for dropdown. This prevents overflow beyond the page container. 

Additionally, added offset to menu when flipping up to prevent overlapping of the current line / caret.

## Results

https://github.com/user-attachments/assets/47a79468-779c-4fd7-8e22-2b496456ea92


etiennejouan (commented):
Thank you for your contribution. It works fine, nice job ! 👏 

During testing, I wondered if we could have a better logic for the offset. If we open slash menu on a heading block, slash menu overlaps with cursor and text.
I'll check for dynamic solution before approving. If you find one, do not hesitate to commit. It's nitpick, if we don't find anything, we'll merge as is.

Image

kpark1208:
Thanks for catching that!

Added a dynamic offset logic that reads the current block’s heading level from the cursor's noteblock and adjusts the vertical offset accordingly. This resolves the overlap issue for each respective heading fonts. It's currently failing one of the frontend CI on `npx nx reset:env twenty-front` but couldn't locate what the error was.

Please review and let me know if any other changes would be needed!

etiennejouan (approved):
Thank you for your contribution, it works great ! 👏

PR #16317, opened 2025-12-04
Null - Fix emails field

### LINKED ISSUE #16270: [API - Create Customer] Sending an empty string as email leads to an error "duplicate"
Precondition:
- CRM has already a customer with an empty E-Mail

Steps to reproduce:
- Cerate a customer, where you send the e-mail element but with an empty value ""-
What happens:
- Duplicate error is send
Expected result:
- Duplicate check will be not done on E-Mail

etiennejouan:
fixes https://github.com/twentyhq/twenty/issues/16270

It happens with api user only.

Integration tests were ok because we expect a returned empty string (check in test) even if null value is expected in db (not test). How can we improve this ?


Weiko (approved):
LGTM

PR #16319, opened 2025-12-04
Page layout tab v2

prastoin:
# Introduction
Migrating `pageLayoutTab` to the v2 engine
- Types and constants
- Builder and validate
- Runner

Introduced a new `StrictSyncableEntity` that enforces that `universalIdentifier` and `applicationId` are defined
As these entities are brand new we could enforce this rule already
This still requires a migration command to associate the existing entities to custom workspace application instance and define universalIdentifier

Handled retro-comp of the migration through a migration as upgrade command fallback


packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/workspace-migration-builder-v2/validators/services/flat-page-layout-tab-validator.service.ts
```diff
@@ -0,0 +1,105 @@
...
+  >): FailedFlatEntityValidation<FlatPageLayoutTab> {
+    const validationResult: FailedFlatEntityValidation<FlatPageLayoutTab> = {
+      type: 'create_page_layout_tab',
+      errors: [],
+      flatEntityMinimalInformation: {
+        id: flatPageLayoutTab.id,
+        title: flatPageLayoutTab.title,
+      },
+    };
+
```
prastoin, line 32: We will have to add page layout validation once we've migrated everything in v2

prastoin:
@greptileai I've refactored the migration and upgrade to be handling everything in a single migration, could you please re-review

packages/twenty-server/src/database/typeorm/core/migrations/common/1764854150563-addUniversalIdentifierAndApplicationIdToPageLayoutTab.ts
```diff
@@ -0,0 +1,60 @@
+import { type MigrationInterface, type QueryRunner } from 'typeorm';
+
+export class AddUniversalIdentifierAndApplicationIdToPageLayoutTab1764854150563
```
prastoin, line 3: Hey @Weiko are you ok with that ?
Pleaser refer newly created `StrictSyncableEntity`

packages/twenty-server/src/engine/workspace-manager/workspace-sync/interfaces/strict-syncable-entity.interface.ts
```diff
@@ -0,0 +1,21 @@
+import { Column, Index, JoinColumn, ManyToOne, Relation } from 'typeorm';
+
+import type { ApplicationEntity } from 'src/engine/core-modules/application/application.entity';
+
+@Index(['workspaceId', 'universalIdentifier'], {
+  unique: true,
+})
+export abstract class StrictSyncableEntity {
```
Weiko, line 8: This is temporary right? In the end all syncable entities should have those non nullable

packages/twenty-server/src/database/typeorm/core/migrations/common/1764854150563-addUniversalIdentifierAndApplicationIdToPageLayoutTab.ts
Weiko: Yes this is definitely something we want to sync, same for tabs and widgets. I'm not sure this is a top priority though

Weiko: Yes this is definitely something we want to sync, same for tabs and widgets. I'm not sure this is a top priority though

prastoin: Sorry I meant mutating the metadata schema in a typeorm migration

prastoin: It's as it's becoming a blocking point for @bosiraphael

packages/twenty-server/src/engine/workspace-manager/workspace-sync/interfaces/strict-syncable-entity.interface.ts
prastoin: Yep it is, it's for incremental entity upgrade

packages/twenty-server/src/database/typeorm/core/migrations/common/1764854150563-addUniversalIdentifierAndApplicationIdToPageLayoutTab.ts
Weiko: > Sorry I meant mutating the metadata schema in a typeorm migration

Ah, missed that. We prefer to avoid running non-DDL in migrations because they are supposed to represent schemas and not business logic and more importantly they need to be deterministic/repeatable across different systems (and those are sometimes harder to revert). 
However, I'd say this could be the only use case where it's acceptable. I'd prefer a command tbh but I can understand here so I'm not strongly against. Wdyt @charlesBochet ?

bosiraphael (approved):
LGTM

prastoin:
@bosiraphael could you please fix unit tests dependency injection ?

bosiraphael:
Tested page layout tab creation update and delete and it works

PR #16330, opened 2025-12-04
use micros converters in FormCurrencyFieldInput

LuckyGoyal039:
Fixes: #15887

The currency fields were expecting values in micros (multiplied by 10^6), but
this requirement was not indicated anywhere in the workflow UI. Users had to
manually add a code node to multiply amounts by 10^6 before using the new
currency value.

This change integrates the micros conversion logic directly into
FormCurrencyFieldInput by using convertCurrencyAmountToCurrencyMicros and
convertCurrencyMicrosToCurrencyAmount utilities. This allows users to input
human-readable decimal amounts (e.g., 3.21 for $3.21) in the form, which are
automatically converted to micros internally for storage and processing.

This eliminates the need for users to add separate code nodes for conversion
and improves the user experience by making the expected input format clear.


LuckyGoyal039:
@thomtrp , @Bonapara could I get a review on the PR?

thomtrp (approved):
LGTM, thank you for your contribution @LuckyGoyal039!

PR #16296, opened 2025-12-03
[Dashboards] Server: Rich Text Widget

ehconitin:
closes https://github.com/twentyhq/core-team-issues/issues/1893

sync in the transformation layer - 
with just the blocknote format -

CleanShot 2025-12-04 at 16 43 27

with just the mardown format - 
CleanShot 2025-12-04 at 16 45 03

attaching response since in screen shot its getting cut - 
```
{
  "id": "7189a10a-177c-4cc2-8946-950b8c125913",
  "pageLayoutTabId": "82ac56dd-4351-43ad-bf3d-49624c057e17",
  "workspaceId": "20202020-1c25-4d02-bf25-6aeccf7ea419",
  "title": "awdawd awdawdawd",
  "type": "STANDALONE_RICH_TEXT",
  "objectMetadataId": null,
  "gridPosition": {
    "row": 1,
    "column": 1,
    "rowSpan": 1,
    "columnSpan": 1
  },
  "configuration": {
    "body": {
      "markdown": "# Hello World",
      "blocknote": "[{\"id\":\"2e744f21-7c71-414d-b87f-fb48400c593c\",\"type\":\"heading\",\"props\":{\"textColor\":\"default\",\"backgroundColor\":\"default\",\"textAlignment\":\"left\",\"level\":1},\"content\":[{\"type\":\"text\",\"text\":\"Hello World\",\"styles\":{}}],\"children\":[]}]"
    }
  },
  "createdAt": "2025-12-04T11:14:52.212Z",
  "updatedAt": "2025-12-04T11:14:52.212Z",
  "deletedAt": null
}
```

PR #16332, opened 2025-12-04
Fix dashboard seeds

bosiraphael:
Fixes a regression on seeding introduced by https://github.com/twentyhq/twenty/pull/16296.
In the seeding process an async function was used without awaiting. This caused all the seeded configs to be empty.

Note: The need to await is because `transformRichTextV2Value` is async due to `const { ServerBlockNoteEditor } = await import('@blocknote/server-util');`. @etiennejouan do you know why is it done this way? Is it to reduce bundle size? Would it be possible to create another util which is not async and which imports `@blocknote/server-util` at the top?

### DIFF
--- a/packages/twenty-server/src/engine/workspace-manager/dev-seeder/core/utils/seed-page-layout-widgets.util.ts
+++ b/packages/twenty-server/src/engine/workspace-manager/dev-seeder/core/utils/seed-page-layout-widgets.util.ts
@@ -17,26 +17,30 @@ export const seedPageLayoutWidgets = async ({
   objectMetadataItems: ObjectMetadataEntity[];
   isDashboardV2Enabled: boolean;
 }) => {
-  const pageLayoutWidgets = getPageLayoutWidgetDataSeeds(
+  const widgetSeeds = getPageLayoutWidgetDataSeeds(
     workspaceId,
     objectMetadataItems,
     isDashboardV2Enabled,
-  ).map((widget) => {
-    const validatedConfiguration = widget.configuration
-      ? validateAndTransformWidgetConfiguration({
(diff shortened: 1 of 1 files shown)


ehconitin (approved):
Did not test but code lgtm!

charlesBochet (approved):

etiennejouan:
> Fixes a regression on seeding introduced by #16296. In the seeding process an async function was used without awaiting. This caused all the seeded configs to be empty.
> 
> Note: The need to await is because `transformRichTextV2Value` is async due to `const { ServerBlockNoteEditor } = await import('@blocknote/server-util');`. @etiennejouan do you know why is it done this way? Is it to reduce bundle size? Would it be possible to create another util which is not async and which imports `@blocknote/server-util` at the top?

@bosiraphael  I know that we had issues with the blocknote/server-util package import. Felix (not sure but I'm not the author of these lines) find this workaround as we don't succeed 'standard' import at the top

bosiraphael:
@FelixMalfait fyi

> > Fixes a regression on seeding introduced by #16296. In the seeding process an async function was used without awaiting. This caused all the seeded configs to be empty.
> > Note: The need to await is because `transformRichTextV2Value` is async due to `const { ServerBlockNoteEditor } = await import('@blocknote/server-util');`. @etiennejouan do you know why is it done this way? Is it to reduce bundle size? Would it be possible to create another util which is not async and which imports `@blocknote/server-util` at the top?
> 
> @bosiraphael I know that we had issues with the blocknote/server-util package import. Felix (not sure but I'm not the author of these lines) find this workaround as we don't succeed 'standard' import at the top

PR #16333, opened 2025-12-04
Fix dev mode

### LINKED ISSUE #1922: [Extensibility] [v1] Fix watch mode exit on error
as title, should not exit

martmull:
Fixes infinite loop in dev mode

Solution -> We stop generating in dev mode

### DIFF
--- a/packages/twenty-apps/hello-world/src/actions/create-new-post-card.ts
+++ b/packages/twenty-apps/hello-world/src/actions/create-new-post-card.ts
@@ -10,6 +10,7 @@ export const main = async (params: { recipient?: string }) => {
         Authorization: `Bearer ${process.env.TWENTY_API_KEY}`,
       },
     });
+
     const createPostCard = await client.mutation({
       createPostCard: {
         __args: {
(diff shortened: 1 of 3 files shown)


packages/twenty-apps/hello-world/src/actions/create-new-post-card.ts
```diff
@@ -21,6 +22,9 @@ export const main = async (params: { recipient?: string }) => {
         id: true,
       },
     });
+
+    console.log('createPostCard result', createPostCard);
```
prastoin, line 26: ***Polish:*** could be removed ?

packages/twenty-sdk/package.json
prastoin, line 70: ***Polish:*** could be removed

prastoin (approved):
Looks good !

packages/twenty-sdk/package.json
martmull: it is generated automatically

packages/twenty-apps/hello-world/src/actions/create-new-post-card.ts
martmull: I have added that so it provides an example for the `twenty app logs` command

prastoin:
By calling the sync application service directly you're bypassing the sdk generation ?

PR #16337, opened 2025-12-04
Fix seeding perf + batch role targets creation

Weiko:


packages/twenty-server/src/engine/metadata-modules/role-target/services/role-target.service.ts
```diff
@@ -37,40 +37,68 @@ export class RoleTargetService {
...
-        },
-        flatRoleTargetMaps,
-        workspaceId,
-      });
+    const allFlatRoleTargetsToCreate: FlatRoleTarget[] = [];
+    const allFlatRoleTargetsToDelete: FlatRoleTarget[] = [];
+
+    for (const createRoleTargetInput of createRoleTargetInputs) {
+      const { flatRoleTargetToCreate, flatRoleTargetsToDelete } =
+        fromCreateRoleTargetInputToFlatRoleTargetToCreate({
```
prastoin, line 79: ***Nitpick:*** This will throw and fail fast on first input invalid transpilation

packages/twenty-server/src/engine/metadata-modules/user-role/user-role.service.ts
```diff
@@ -208,57 +215,72 @@ export class UserRoleService {
...
+      }
+
+      userWorkspaceIdsToAssign.push(userWorkspaceId);
     }
 
-    await this.validateMoreThanOneWorkspaceMemberHasAdminRoleOrThrow({
-      workspaceId,
-      adminRoleId: currentRole.id,
-    });
+    if (isDefined(adminRoleIdToValidate)) {
```
prastoin, line 276: ***Vision:*** in the future we should be able to pass userWorkspaceMember to the validate build and run and at some point same for the runner so it will be to have workspace scoped validation errors

prastoin (approved):
Nice job ! 

Image

PR #16343, opened 2025-12-04
Refactor validate build and run for configuration to be less verbose and more reliable

prastoin:
# Introduction
Refactored the api `validateBuildAndRunWorkspaceMigration` to be require less configuration but to infer required args dynamically depending on provided metadata maps to compare

## `inferDeletionFromMissingEntities`
Is not dynamically computed avoiding any miss configuration issue and any missleading devxp

## Maps computation

Making only one call to redis to build both dependency and to be compared entity maps. It does not matter to avoid passing a about to compared flat entity maps to could also be a depedency, it's handled directly in the builder setup optimistic cache logic

Please note that the flat maps used for the service input transpilation might differ from the one that we will dynamically compute and inject in the builder. Leading to do 2 redis calls but also race condition prone validation error
We prefer that this occurs at the builder rather than at the runner level as the pg instance is not cache and reflect the real state of a given workspace
In a nutshell, there's a possible race condition between cache invalidation and computation in both service input transpilers and builder but we're totally ok with that


packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/services/workspace-migration-validate-build-and-run-service.ts
```diff
@@ -19,14 +38,138 @@ export class WorkspaceMigrationValidateBuildAndRunService {
...
+          (metadataName) =>
+            Object.keys(
+              ALL_METADATA_REQUIRED_METADATA_FOR_VALIDATION[metadataName],
+            ) as AllMetadataName[],
+        ),
+      ),
+    ];
+    const allMetadataNameCacheToCompute = [
+      ...new Set(...allMetadataNameToCompare, allDependencyMetadataName),
+    ];
```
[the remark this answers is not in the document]
prastoin: Good catch

packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/services/workspace-migration-build-orchestrator.service.ts
```diff
@@ -65,7 +65,10 @@ export class WorkspaceMigrationBuildOrchestratorService {
       (allFlatEntityMaps, currFlatMaps) => {
         const fromToOccurence = fromToAllFlatEntityMaps[currFlatMaps];
 
-        if (!isDefined(fromToOccurence)) {
+        if (
+          !isDefined(fromToOccurence) ||
+          isDefined(allFlatEntityMaps[currFlatMaps])
+        ) {
           return allFlatEntityMaps;
         }
```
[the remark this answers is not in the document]
prastoin: No it's partial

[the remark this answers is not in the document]
prastoin: No it's partial

packages/twenty-server/src/engine/metadata-modules/flat-entity/types/flat-entity-to-create-delete-update.type.ts
```diff
@@ -0,0 +1,9 @@
+import { type AllMetadataName } from 'twenty-shared/metadata';
+
+import { type MetadataFlatEntity } from 'src/engine/metadata-modules/flat-entity/types/metadata-flat-entity.type';
+
+export type FlatEntityToCreateDeleteUpdate<T extends AllMetadataName> = {
```
thomtrp, line 5: naming sounds weird. `FlatEntityChangeset` or `FlatEntityMutations`?

packages/twenty-server/src/engine/core-modules/application/application-sync.service.ts
```diff
@@ -990,54 +988,37 @@ export class ApplicationSyncService {
...
-    });
-
     await this.workspaceMigrationValidateBuildAndRunService.validateBuildAndRunWorkspaceMigration(
       {
-        fromToAllFlatEntityMaps: {
-          flatObjectMetadataMaps: {
-            from: fromFlatObjectMetadataMaps,
-            to: createEmptyFlatEntityMaps(),
+        allFlatEntities: {
+          objectMetadata: {
```
thomtrp, line 1004: is `allFlatEntities` well named? I feel it rather group operations to perform. Not entities

packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/services/workspace-migration-validate-build-and-run-service.ts
```diff
@@ -1,15 +1,35 @@
 import { Injectable, Logger } from '@nestjs/common';
 
-import { WorkspaceMigrationV2ExceptionCode } from 'twenty-shared/metadata';
+import {
+  AllMetadataName,
```
thomtrp, line 4: Not sure I understand what AllMetadataName represent

packages/twenty-server/src/engine/metadata-modules/flat-entity/types/flat-entity-to-create-delete-update.type.ts
prastoin: You have a Matrice too, but create delete update are very wide spread naming across the builder and runner.
No strong opinion, not fan of mutation as we have create operations
Maybe operations are pretty decent ?

packages/twenty-server/src/engine/core-modules/application/application-sync.service.ts
prastoin: IMO this is related to previous comment, technically we have `allFlatEntityCreateDeleteUpdateByMetadataName`.
Maybe `allFlatEntitiesByMetadataName` would be more realistic ? But still quite technical

packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/services/workspace-migration-validate-build-and-run-service.ts
prastoin: These are all the metadata entities name that are v2 compatible, we were thinking of renaming it to `allSyncableMetadataName` at some point

packages/twenty-server/src/engine/workspace-manager/workspace-migration-v2/services/workspace-migration-validate-build-and-run-service.ts
```diff
@@ -19,14 +39,145 @@ export class WorkspaceMigrationValidateBuildAndRunService {
...
+          [metadataFlatEntityMapsKey]:
+            allRelatedFlatEntityMaps[metadataFlatEntityMapsKey],
+        };
+      },
+      initialAccumulator,
+    );
+
+    return {
+      allRelatedFlatEntityMaps,
+      allMetadataNameToCompare,
```
prastoin: Could be removed

packages/twenty-server/src/engine/metadata-modules/flat-entity/types/flat-entity-to-create-delete-update.type.ts
thomtrp: operations sound good

thomtrp (approved):

PR #16346, opened 2025-12-04
Fix twenty server main typecheck

prastoin:
Two PRs have been merged at the same time leading to type error on default branch


Weiko:
Already fixed in another PR, thanks though 🙏

PR #16351, opened 2025-12-04
Improvement on messaging

charlesBochet:
In this PR:
- change messaging / calendar stale duration check to 30minutes (cron is running every 1h, duration check was 1h, so evaluation was flaky)
- when temporary error (throttling), preserve syncStageStartedAt as this is necessary to assess exponential throttling

### DIFF
--- a/packages/twenty-server/src/modules/calendar/calendar-event-import-manager/crons/jobs/calendar-event-list-fetch.cron.job.ts
+++ b/packages/twenty-server/src/modules/calendar/calendar-event-import-manager/crons/jobs/calendar-event-list-fetch.cron.job.ts
@@ -53,7 +53,7 @@ export class CalendarEventListFetchCronJob {
         const now = new Date().toISOString();
 
         const [calendarChannels] = await this.coreDataSource.query(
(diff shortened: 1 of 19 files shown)


packages/twenty-server/src/modules/calendar/calendar-event-import-manager/crons/jobs/calendar-event-list-fetch.cron.job.ts
```diff
@@ -53,7 +53,7 @@ export class CalendarEventListFetchCronJob {
         const now = new Date().toISOString();
 
         const [calendarChannels] = await this.coreDataSource.query(
-          `UPDATE ${schemaName}."calendarChannel" SET "syncStage" = '${CalendarChannelSyncStage.CALENDAR_EVENT_LIST_FETCH_SCHEDULED}', "syncStageStartedAt" = '${now}'
+          `UPDATE ${schemaName}."calendarChannel" SET "syncStage" = '${CalendarChannelSyncStage.CALENDAR_EVENT_LIST_FETCH_SCHEDULED}', "syncStageStartedAt" = COALESCE("syncStageStartedAt", '${now}')
```
charlesBochet, line 56: we update syncStageStartedAt only if nulll

PR #16365, opened 2025-12-05
1895 extensibility v1 application tokens

### LINKED ISSUE #1895: [Extensibility] [v1] Application tokens
-

martmull:
First PR to implement application tokens
- add new application role in twenty-server
- move duplicated constants and types to twenty-shared
- will  add role configuration utils into twenty-sdk in another PR


packages/twenty-server/src/engine/core-modules/application/application-sync.service.ts
```diff
@@ -124,6 +134,104 @@ export class ApplicationSyncService {
 
     let serverlessFunctionLayerId = application.serverlessFunctionLayerId;
 
+    const applicationRole = manifest.application.applicationRole;
+
+    if (
+      isDefined(applicationRole) &&
+      isDefined(applicationRole.universalIdentifier)
+    ) {
```
thomtrp, line 195: I would extract in a separated function

packages/twenty-shared/src/types/HttpMethod.ts
```diff
@@ -0,0 +1,7 @@
+export enum HTTPMethod {
```
thomtrp, line 1: did you move the one used in routeTrigger entity?

thomtrp (approved):
LGTM

packages/twenty-shared/src/types/HttpMethod.ts
martmull: nice catch thank you

PR #16367, opened 2025-12-05
Refactor page layout tab and widgets migration

prastoin:
# Introduction
Making columns nullable instead of required + metadata on the fly migration in typeorm migration that could affect other breaking change migrations to be run


Weiko (approved):
LGTM! 👍

PR #16371, opened 2025-12-05
Fix unique standard field

### LINKED ISSUE #15925: Cannot set a standard field from a custom object unique
## Bug Description
When I try to set the Name field of a custom object to unique, I get this error when trying to save this configuration (which is therefore not saved). 

```
         "message": "Cannot edit standard field metadata properties: isUnique",
            "extensions": {
                "subCode": "FIELD_MUTATION_NOT_ALLOWED",
                "userFriendlyMessage": "An error occurred.",
                "code": "FORBIDDEN"
            }
```

cc @prastoin  as Etienne said this might be related to one of your ongoing topics

thank you!

etiennejouan:
Fixes https://github.com/twentyhq/twenty/issues/15925

- update field metadata update logic
- uniformize the way index are named
- command to migrate v1-named unique index
- add integration testing


packages/twenty-server/src/database/commands/upgrade-version-command/1-13/1-13-rename-unique-index.command.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    const uniqueIndexes = await this.indexMetadataRepository.find({
+      where: {
+        workspaceId,
+        isUnique: true,
+      },
+      relations: ['objectMetadata', 'indexFieldMetadatas.fieldMetadata'],
+    });
+
```
prastoin: ***Suggestion:*** We could use workspaceCacheService directly here, might be costy to re-construction field relation from maps though TBD

packages/twenty-server/src/database/commands/upgrade-version-command/1-13/1-13-rename-unique-index.command.ts
```diff
@@ -0,0 +1,117 @@
...
+    for (const indexMetadata of uniqueIndexes) {
+      if (indexMetadata.name.includes('UNIQUE')) {
```
prastoin: ***Question:*** I'm wondering if we could compute the index name a again using v2 method and compare to existing, I'm afraid the difference might not only be the UNIQUE prefix

packages/twenty-server/src/database/commands/upgrade-version-command/1-13/1-13-rename-unique-index.command.ts
```diff
@@ -0,0 +1,117 @@
...
+
+      if (!isDryRun) {
+        await this.renameIndexOnDatabase(
+          dataSource,
+          schemaName,
+          indexMetadata.name,
+          newIndexName,
+        );
+
+        await this.indexMetadataRepository.update(indexMetadata.id, {
```
prastoin: ***Remark:*** We should invalidate field object and index cache entries after updating all indeces

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/constants/flat-field-metadata-editable-properties.constant.ts
```diff
@@ -21,5 +21,6 @@ export const FLAT_FIELD_METADATA_EDITABLE_PROPERTIES = {
     'label',
     'options',
     'settings',
+    'isUnique', // not editable for standard fields with standard unique constraint
```
prastoin: At least from the api metadata

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/handle-index-changes-during-field-update.util.ts
```diff
@@ -135,11 +152,31 @@ const handleExistingIndexes = ({
...
+      status: 'success',
+      result: {
+        ...FIELD_METADATA_UPDATE_INDEX_SIDE_EFFECT,
+        flatIndexMetadatasToDelete: uniqueIndexToDelete
+          ? [uniqueIndexToDelete]
+          : [],
+      },
     };
   }
   const updatedIndexes = recomputeIndexOnFlatFieldMetadataNameUpdate({
```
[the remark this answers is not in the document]
prastoin: This is true, not all entities have the applicationId set already
We might be checking the isCustom as a fallback assertion

packages/twenty-server/src/engine/twenty-orm/decorators/workspace-field-index.decorator.ts
```diff
@@ -30,10 +30,18 @@ export function WorkspaceFieldIndex(
     ];
 
     metadataArgsStorage.addIndexes({
-      name: `IDX_${generateDeterministicIndexName([
-        convertClassNameToObjectMetadataName(target.constructor.name),
-        ...columns,
-      ])}`,
+      name: generateDeterministicIndexNameV2({
```
prastoin, line 33: Nice addition, we should also note that sync metadata does not handle index update so it has no impact on existing indeces, which is good ( only used atm at workspace creation )

packages/twenty-server/test/integration/metadata/suites/field-metadata/failing-update-one-standard-field-metadata.integration-spec.ts
```diff
@@ -103,3 +103,59 @@ describe('Standard field metadata update should be ignored', () => {
     },
   );
 });
+
+describe('Standard field with standard unique index update should fail on isUnique change', () => {
```
prastoin: neat

packages/twenty-server/test/integration/metadata/suites/field-metadata/successful-update-one-standard-field-metadata.integration-spec.ts
```diff
@@ -253,3 +256,148 @@ describe('Standard field metadata update should succeed', () => {
...
+      input: customObject,
+      gqlFields: `
+        id
+        nameSingular
+      `,
+    });
+
+    customObjectId = data.createOneObject.id;
+
+    const { objects } = await findManyObjectMetadata({
```
prastoin, line 289: ***Suggestion:*** We could use `findManyObjectWithIndexes` and expect the index to be created and as custom

packages/twenty-server/test/integration/metadata/suites/field-metadata/successful-update-one-standard-field-metadata.integration-spec.ts
```diff
@@ -253,3 +256,148 @@ describe('Standard field metadata update should succeed', () => {
...
+        isCustom
+      `,
+    });
+
+    expect(data.updateOneField.id).toBe(nameFieldMetadata.id);
+    expect(data.updateOneField.name).toBe('name');
+    expect(data.updateOneField.isUnique).toBe(true);
+  });
+
+  it('should set isUnique back to false on standard field', async () => {
```
prastoin: ***Remark:*** That's a bit of a test leakage, I would rather handle that in an after jest hook
Because if you only run the first test the `isUnique` value is not updated restored which could have an impact on other tests

packages/twenty-server/test/integration/metadata/utils/warn-if-error-but-not-expected-to-fail.util.ts
```diff
@@ -16,7 +16,7 @@ export const warnIfErrorButNotExpectedToFail = ({
       process.env.LOG_LEVELS.includes('debug')
     ) {
       // eslint-disable-next-line no-console
-      console.log(JSON.stringify(response.body.errors, null, 2));
+      console.log(JSON.stringify(response.body.errors));
```
prastoin: ***Question:*** Why's that ? pretty useful for json to be formatted when logged

prastoin (commented):
Good job ! We might wanna update the frontend to also be comparing the index applicationId to the workspaceCustomApplicationId in order to disable dyanmically the input instead of checking the isCustom boolean
That's polishing

packages/twenty-server/test/integration/metadata/utils/warn-if-error-but-not-expected-to-fail.util.ts
etiennejouan: Should not have been committed. Modify during test

prastoin (approved):

PR #16372, opened 2025-12-05
Show friendly message for non-previewable file previews

### LINKED ISSUE #15771: File preview - Improve zip file preview
## Scope & Context

The preview of uploaded files is working well, apart for the zip files where the preview is "weird". 
Could we instead show a message to tell users that such preview is NA because of the nature of the file, telling them to download the file instead? 
A customer reported this as a bug (because he was surprised of what this looked like). I don't think this is necessarily a bug but a good feedback or where we can improve the user's experience.

abdulrahmancodes:
Screenshot 2025-12-06 at 12 50 42 AM

Closes #15771


Devessier (approved):
Great job! 💪

Issue #16376, opened 2025-12-06
Deleted user instead of Untitled in Created by field


BOHEUS:
Ticket: https://discord.com/channels/1130383047699738754/1446119214669889596

Right now, if creator of record is deleted, "Untitled" is shown instead of "Deleted user". Technically speaking, it's possible to show "Deleted user" as "Untitled" appears only when there's no name of user by modifying the ActorDisplay.tsx, however, in newest version, data regarding deleted user are preserved in DB as there's no cascade data removal upon deleting workspace member so change would probably work only for records migrated from older versions where creator doesn't exist now.

BOHEUS:
@Bonapara what should be the correct approach to solve this problem since creator data appears in records table view, record details view and in API?

Bonapara:
I was thinking literally display "deleted user", is that what you had in mind @BOHEUS?

BOHEUS:
Closing as user in ticket didn't respond and according to comment from other similar issue, it's been already resolved https://github.com/twentyhq/twenty/issues/18194#issuecomment-4103285885

PR #16377, opened 2025-12-06
Deleted user placeholder

BOHEUS:
This change is only for old records with no data of deleted workspace member responsible for creating a record, new records have data of removed workspace member saved in DB in columns `createdByWorkspaceMemberId` and `createdByName`

Related issue #16376

### DIFF
--- a/packages/twenty-front/src/modules/ui/field/display/components/ActorDisplay.tsx
+++ b/packages/twenty-front/src/modules/ui/field/display/components/ActorDisplay.tsx
@@ -17,6 +17,7 @@ import {
   IconSettingsAutomation,
   IconWebhook,
 } from 'twenty-ui/display';
+import { isNonEmptyString } from '@sniptt/guards';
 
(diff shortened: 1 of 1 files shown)


etiennejouan (commented):
@BOHEUS thank you for your contribution 👍 

It seems good but i've not been able to test. I tried after deleting my workspace member, but 'createdByName' remains stored.
Besides, name should never be null or undefined in ActorDisplay, can be empty string. Then, if name is empty string `name ?? tDeleted user` should not work well. I would prefer `isNonEmptyString(name)?name:tDeleted user`.

BOHEUS:
Done

etiennejouan:
@BOHEUS, thanks. how do you test to be sure it's working ?

BOHEUS:
@etiennejouan after some thought, I think it's better to close this PR and go back to drawing board for time being as it'll only fix the problem in one place (table view of records) while no info still will be in record details view as well as in API so better approach must be taken to solve this once and properly (which I didn't realize in the beginning)

PR #16382, opened 2025-12-07
[fix] Fix switch view type from "Layout" switcher

ijreilly:
In this previous work aiming at eliminating the creation of viewGroups being separated from the views, I removed the creation of viewGroups from the FE at view creation, but forgot to do it at view update when the user chooses "change Layout" option (from Table/Kanban/Calendar to another).

In this PR
- I removed creation, deletion and destroy of viewGroups from usePersistViewGroupRecords. I left update because it useful for visibility and position
- since usePersistViewGroups record was also handling optimistic, I moved optimistic logic to a new hook useViewsSideEffectsOnViewGroups. I realized that so far we optimistic is only useful for creation. For the update though we do need to compute the view groups that will be created to be able to call loadRecordIndexState, so I created a function for that, but I did not seek to perform real optimistic, with writing and deleting groups from the cache as it was a bit heavier and not in use for now and I want to merge this asap as it fixes the creation of viewGroups.

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-board/states/switchesToKanbanViewTypeComponentState.ts
+++ b/packages/twenty-front/src/modules/object-record/record-board/states/switchesToKanbanViewTypeComponentState.ts
@@ -1,0 +1,9 @@
+import { ContextStoreComponentInstanceContext } from '@/context-store/states/contexts/ContextStoreComponentInstanceContext';
+import { createComponentState } from '@/ui/utilities/state/component-state/utils/createComponentState';
+
+export const switchesToKanbanViewTypeComponentState =
+  createComponentState<boolean>({
+    key: 'switchesToKanbanViewTypeComponentState',
+    componentInstanceContext: ContextStoreComponentInstanceContext,
+    defaultValue: false,
+  });
--- a/packages/twenty-front/src/modules/object-record/record-index/hooks/useHandleRecordGroupField.ts
+++ b/packages/twenty-front/src/modules/object-record/record-index/hooks/useHandleRecordGroupField.ts
@@ -5,7 +5,6 @@ import { useSetRecordGroups } from '@/object-record/record-group/hooks/useSetRec
 import { useLoadRecordIndexStates } from '@/object-record/record-index/hooks/useLoadRecordIndexStates';
 import { useRecoilComponentCallbackState } from '@/ui/utilities/state/component-state/hooks/useRecoilComponentCallbackState';
 import { usePersistView } from '@/views/hooks/internal/usePersistView';
-import { usePersistViewGroupRecords } from '@/views/hooks/internal/usePersistViewGroup';
 import { useGetViewFromPrefetchState } from '@/views/hooks/useGetViewFromPrefetchState';
 import { type ViewGroup } from '@/views/types/ViewGroup';
 import { convertCoreViewToView } from '@/views/utils/convertCoreViewToView';
@@ -16,8 +15,6 @@ import { type CoreView } from '~/generated/graphql';
 import { isUndefinedOrNull } from '~/utils/isUndefinedOrNull';
 
 export const useHandleRecordGroupField = () => {
-  const { deleteViewGroups } = usePersistViewGroupRecords();
-
   const currentViewIdCallbackState = useRecoilComponentCallbackState(
     contextStoreCurrentViewIdComponentState,
   );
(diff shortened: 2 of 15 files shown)


packages/twenty-front/src/modules/object-record/record-index/hooks/useHandleRecordGroupField.ts
```diff
@@ -159,32 +156,14 @@ export const useHandleRecordGroupField = () => {
...
-        if (!isDefined(view.mainGroupByFieldMetadataId)) {
-          throw new Error('mainGroupByFieldMetadataId is required');
-        }
-
-        setRecordGroupsFromViewGroups({
-          viewId: view.id,
-          mainGroupByFieldMetadataId: view.mainGroupByFieldMetadataId,
-          viewGroups: [],
-          objectMetadataItem,
+        await updateView({
```
ijreilly, line 165: this feature has been disabled anyway for now so this is not callable. 
It aims at removing the grouping from a view. I will fully take care of it next

packages/twenty-front/src/modules/views/utils/convertUpdateViewInputToCore.ts
```diff
@@ -46,5 +46,8 @@ export const convertUpdateViewInputToCore = (
       calendarFieldMetadataId: view.calendarFieldMetadataId,
     }),
     ...(isDefined(view.visibility) && { visibility: view.visibility }),
+    ...(view.mainGroupByFieldMetadataId !== undefined && {
```
lucasbordeau, line 49: isDefined

packages/twenty-front/src/modules/views/hooks/useViewsSideEffectsOnViewGroups.ts
```diff
@@ -0,0 +1,149 @@
...
+    return { viewGroupsToCreate };
+  };
+
+  return {
+    triggerViewGroupSideEffectAtViewCreation,
+    getViewGroupsToCreateAtViewUpdate,
+  };
+};
+
+export default useViewsSideEffectsOnViewGroups;
```
lucasbordeau: Avoid export default

packages/twenty-front/src/modules/views/hooks/useViewsSideEffectsOnViewGroups.ts
```diff
@@ -0,0 +1,149 @@
...
+            updatedAt: new Date().toISOString(),
+            deletedAt: null,
+          }) as Omit<CoreViewGroup, 'workspaceId'>,
+      ),
+    });
+
+    return { viewGroupsToCreate };
+  };
+
+  const getViewGroupsToCreateAtViewUpdate = ({
```
lucasbordeau: Can we extract this in a separate file to avoid barrel hooks ?

packages/twenty-front/src/modules/views/hooks/useViewsSideEffectsOnViewGroups.ts
```diff
@@ -0,0 +1,149 @@
...
+import { v4 } from 'uuid';
+import { type CoreViewGroup } from '~/generated/graphql';
+
+const useViewsSideEffectsOnViewGroups = () => {
+  const { triggerViewGroupOptimisticEffect } =
+    useTriggerViewGroupOptimisticEffect();
+
+  const objectMetadataItems = useRecoilValue(objectMetadataItemsState);
+
+  const triggerViewGroupSideEffectAtViewCreation = ({
```
lucasbordeau: I would use a more self-explanatory naming than side effect.

packages/twenty-front/src/modules/object-record/record-board/states/switchesToKanbanViewTypeComponentState.ts
```diff
@@ -0,0 +1,9 @@
+import { ContextStoreComponentInstanceContext } from '@/context-store/states/contexts/ContextStoreComponentInstanceContext';
+import { createComponentState } from '@/ui/utilities/state/component-state/utils/createComponentState';
+
+export const switchesToKanbanViewTypeComponentState =
```
lucasbordeau: isSwitching, to respect our convention with similar states

packages/twenty-front/src/modules/views/utils/convertUpdateViewInputToCore.ts
ijreilly: The behaviour between undefined and null is different here which is why i did not use isDefined.
There's a different between nullifying view.mainGroupByFieldMetadataId and the undefined value which we will have if we don't update view.mainGroupByFieldMetadataId

lucasbordeau (approved):
LGTM, let's see isDefined later and edge case also.

Issue #414, opened 2023-06-26
I see a checkbox on cards


charlesBochet:

Bonapara:
I can select a card by clicking on the checkbox OR on the header (unless the card name that allows to change a card company / People relation)

PR #16385, opened 2025-12-08
Fix: Persist hide empty groups setting on backend (BREAKING: deploy server first)

### LINKED ISSUE #13754: [Group by view] Remove empty group configuration is not saved
## Bug Description

I have a group by view
I chose the option to not display empty group
If i navigate and come back to this page, empty groups are not shown. But anytime you truly reload, the empty groups are back.
As a user, you want the configurations of the view to be saved as we have people creating the views on behalf of the users - who are not expected to change the set up

**Example:** 
Twenty for Twenty workspace, view with active workspaces 5+ users

## Expected behavior

The view remains the same, not showing empty groups, until someone change the toggle.

abdulrahmancodes:
Closes #13754 and [core team issue #414](https://github.com/twentyhq/core-team-issues/issues/414)


packages/twenty-front/src/modules/views/types/GraphQLView.ts
```diff
@@ -16,6 +16,7 @@ export type GraphQLView = {
   type: ViewType;
   key: ViewKey | null;
   mainGroupByFieldMetadataId?: string | null;
+  shouldHideEmptyGroups?: boolean;
```
ijreilly: shouldHideEmptyGroups is non nullable so should never be undefined

packages/twenty-front/src/modules/views/utils/convertCoreViewToView.ts
```diff
@@ -39,6 +39,7 @@ export const convertCoreViewToView = (
     ),
     viewSorts: coreView.viewSorts,
     mainGroupByFieldMetadataId: coreView.mainGroupByFieldMetadataId ?? null,
+    shouldHideEmptyGroups: coreView.shouldHideEmptyGroups ?? false,
```
ijreilly: coreView.shouldHideEmptyGroups is never undefined in my understanding

ijreilly (commented):
looks good, although the first time i tested the toggle wasnt changing anything (it was not hiding the empty group in the current view), but i was not able to reproduce. 

Also note that this will cause a downtime if the front is deployed before the back.

ijreilly (approved):
lgtm !

prastoin:
Hey there
We should invalidate flat-view cache after releasing this

PR #16386, opened 2025-12-08
Ensure searchVector is generated from searchFieldMetadata records and expose mutations for searchFieldMetadata.

mabdullahabaid:
Related to [1428](https://github.com/twentyhq/core-team-issues/issues/1428).

**GraphQL Mutations for Search Field Management.**
- addSearchField: Adds a field to the search vector for a given object
- removeSearchField: Removes a field from the search vector
- Both mutations are protected by DATA_MODEL permission guard
- Returns SearchFieldMetadataDTO with full metadata

**Automatic Search Vector Regeneration**
- When fields are added or removed:
- Workspace sync updates metadata to reflect the new search field configuration
- Direct column regeneration drops and recreates the searchVector PostgreSQL generated column with the updated expression
- Associated indexes are preserved during regeneration

**Why need direct column regeneration?**
- PostgreSQL doesn't allow altering asExpression on generated columns, so the column is dropped and recreated. 
- The regeneration process:
- Fetches current search fields from SearchFieldMetadataEntity
- Builds the to_tsvector SQL expression from searchable fields
- Drops existing indexes on searchVector (if any)
- Drops the searchVector column
- Recreates the column with the new expression
- Recreates indexes with original configuration

**Migration Command for Existing Workspaces**
- upgrade:1-13:backfill-search-field-metadata: Backfills the searchFieldMetadata table from existing hardcoded SEARCH_FIELDS_FOR_* configurations
- Supports both standard objects (Person, Company, Opportunity, etc.) and custom objects
- Skips objects that already have search field metadata entries

**Next Steps**
- Remove SEARCH_FIELDS_FOR_* arrays from the entities (once command 1.13 has run).
- Build the UI on the Frontend (ObjectSettings) to addSearchField and removeSearchField.
- Determine if we need to allow more types beyond the ones specified by SEARCHABLE_FIELD_TYPES array.


mabdullahabaid:
Created a design document for this feature: 
https://github.com/mabdullahabaid/design-docs/blob/main/design-dynamic-search-field-management.md

**Challenge:**

PostgreSQL doesn't allow altering the asExpression of a generated column directly, so the searchVector column must be dropped and recreated. This creates a challenge:
- workspaceSyncMetadataService.synchronize() manages its own transaction and updates metadata
- regenerateSearchVectorColumn() performs DDL operations (DROP/ADD column) in its own transaction

This introduces the following issue highlighted by Greptile: No transaction handling across metadata save, workspace sync, and column regeneration (can leave system in inconsistent state).

I am unable to figure out how to solve for the transactional issue. Also, I can use some thoughts on the approach to solving the problem.

PR #16399, opened 2025-12-08
Refactor global datasource part 2

Weiko:
## Context
Deprecating TwentyORMManager in favor of TwentyORMGlobalManager (temporarily, as this will simplify the ultimate goal to later replace all usages with the new TwentyORMGlobalManagerV2 which will have a similar signature)
This means this PR had to refactor a bit of code to pass down the workspaceId when not available directly as it is now a requirement, meaning we also deprecated scopedWorkspaceContextFactory to have a less obscure way to fetch the workspaceId and have something more declarative.

Step 3 will be to update TwentyORMGlobalManager to use a featureFlag toggling and use the new GlobalWorkspaceOrmManager internally using the new cache service

Step 4 will be to remove the feature flag and pg_pool patch


packages/twenty-server/src/modules/messaging/common/query-hooks/message/message-find-one.post-query.hook.ts
```diff
@@ -31,8 +31,15 @@ export class MessageFindOnePostQueryHook
       throw new ForbiddenError('User is required');
     }
 
+    const workspace = authContext.workspace;
+
+    if (!workspace) {
```
charlesBochet, line 36: nitpick: not specifically here but let's use isDefined

charlesBochet (approved):

PR #16404, opened 2025-12-08
Fix messaging errors on message folder list

charlesBochet:
This is a known quirk of many Google APIs, including Gmail. Some error responses use numeric code fields (e.g., 403), while others may return them as strings (e.g., "403"). This depends on which internal service returns the error and the language client you’re using.

### DIFF
--- a/packages/twenty-server/src/modules/messaging/message-import-manager/drivers/gmail/mocks/gmail-api-error-mocks.ts
+++ b/packages/twenty-server/src/modules/messaging/message-import-manager/drivers/gmail/mocks/gmail-api-error-mocks.ts
@@ -63,118 +63,111 @@ const gmailApiErrorMocks = {
     code: 500,
     message: 'backendError',
   },
-
-  convertToErrorWithErrorCodeStringOrNumber: function ({
-    error,
-    errorCodeAsString,
-  }: {
-    error: GmailApiError;
-    errorCodeAsString: boolean;
-  }): GmailApiError {
-    return {
-      code: errorCodeAsString ? error.code.toString() : error.code,
-      message: error.message,
-    };
-  },
-
-  getError: function ({
-    code,
-    type,
-    errorCodeAsString = false,
-  }: {
-    code: number;
-    type?: string;
-    errorCodeAsString?: boolean;
-  }): GmailApiError {
-    switch (code) {
-      case 400:
-        switch (type) {
-          case 'invalid_grant':
-            return this.convertToErrorWithErrorCodeStringOrNumber({
-              error: this.invalidGrant,
-              errorCodeAsString,
-            });
-          case 'failedPrecondition':
-            return this.convertToErrorWithErrorCodeStringOrNumber({
-              error: this.failedPrecondition,
-              errorCodeAsString,
-            });
-          default:
-            return this.convertToErrorWithErrorCodeStringOrNumber({
-              error: this.badRequest,
-              errorCodeAsString,
-            });
-        }
-      case 401:
-        return this.convertToErrorWithErrorCodeStringOrNumber({
-          error: this.invalidCredentials,
-          errorCodeAsString,
-        });
-      case 403:
-        switch (type) {
-          case 'dailyLimit':
-            return this.convertToErrorWithErrorCodeStringOrNumber({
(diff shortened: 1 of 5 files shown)


charlesBochet:
@neo773 can you take a look?

packages/twenty-server/src/modules/messaging/message-import-manager/drivers/gmail/types/gmail-api-error.type.ts
```diff
@@ -1,4 +1,4 @@
 export type GmailApiError = {
-  code: string;
+  code: number | string;
```
charlesBochet, line 2: changing type here

packages/twenty-server/src/modules/messaging/message-import-manager/drivers/gmail/utils/is-gmail-api-error-error.util.ts
```diff
@@ -7,7 +7,7 @@ export const isGmailApiError = (error: unknown): error is GmailApiError => {
 
   if (
     !('code' in error) ||
-    typeof error.code !== 'string' ||
+    (typeof error.code !== 'number' && typeof error.code !== 'string') ||
```
charlesBochet, line 10: accepting both

packages/twenty-server/src/modules/messaging/message-import-manager/drivers/gmail/utils/parse-gmail-api-error.util.ts
```diff
@@ -9,8 +9,10 @@ export const parseGmailApiError = (
 ): MessageImportDriverException => {
   const { code, message } = error;
 
-  switch (code) {
-    case '400':
+  const codeAsNumber = Number(code);
```
charlesBochet, line 12: parsing both as number

FelixMalfait (approved):

guillim (approved):
Just thinking it's a bit weird to declare methods in a mock file but othewise LGTM

packages/twenty-server/src/modules/messaging/message-import-manager/drivers/gmail/mocks/gmail-api-error-mocks.ts
```diff
@@ -3,112 +3,171 @@ import { type GmailApiError } from 'src/modules/messaging/message-import-manager
...
+  errorCodeAsString: boolean;
+}): GmailApiError => {
+  return {
+    code: errorCodeAsString ? error.code.toString() : error.code,
+    message: error.message,
+  };
 };
 
-export default gmailApiErrorMocks;
+export const getGmailApiError = ({
```
Weiko, line 81: nit: maybe instead of calling the util on every return we can extract that logic outside of the method.
Or maybe we can refactor the nested switch + util to a config map like so
```typescript
type ErrorKey = keyof typeof gmailApiErrorMocks;

const ERROR_MAP: Record<
  number,
  Record<string | "default", ErrorKey> | ErrorKey
> = {
  400: {
    invalid_grant: "invalidGrant",
    failedPrecondition: "failedPrecondition",
    default: "badRequest",
  },
  401: "invalidCredentials",
  403: {
    dailyLimit: "dailyLimitExceeded",
    userRateLimit: "userRateLimitExceeded",
    rateLimit: "rateLimitExceeded",
    domainPolicy: "domainPolicyError",
    default: "rateLimitExceeded",
  },
  404: "notFound",
  410: "gone",
  429: {
    concurrent: "tooManyConcurrentRequests",
    default: "tooManyConcurrentRequests",
  },
  500: "backendError",
};

export const getGmailApiError = ({
  code,
  type,
  errorCodeAsString = false,
}: {
  code: number;
  type?: string;
  errorCodeAsString?: boolean;
}): GmailApiError => {
  const mapping = ERROR_MAP[code];

  if (!mapping) {
    throw new Error(`Unknown error code: ${code}`);
  }

  const key =
    typeof mapping === "string"
      ? mapping
      : mapping[type] ?? mapping._default;

  return convertToErrorWithErrorCodeStringOrNumber({
    error: gmailApiErrorMocks[key],
    errorCodeAsString,
  });
};
```

Weiko (approved):
LGTM

PR #16407, opened 2025-12-08
Fixes on views

### LINKED ISSUE #15640: Kanban View showing Twice the same column that cannot be removed
Kanban view on opp shows twice the column "No Value". Once this column is removed from the view and then we refresh, the two (same) columns are back.
thsi was reported by the customer and I can reproduce

Front logo [Front conversations](https://app.frontapp.com/open/top_3pe91y)

ijreilly:
Fixes

- Fixes https://github.com/twentyhq/twenty/issues/15640 : we have some viewGroups that were in the past wrongly migrated; eg deleteing an enum's option did not lead to deleting the associated viewGroup. We have not cleaned that (we could do a command), but we can identify them and choose not to display them - otherwise currently they are shown as a duplicate of "No value" column
- Fixes buggy edge case introduced by https://github.com/twentyhq/twenty/pull/16382 : after updating a view from Table/Calendar to Kanban, then visiting another view, when coming back to the newly updated kanban view the view groups were empty. This is due to to the view groups being optimistically created with on-the-fly computed ids when the view is updated, while we call refreshCoreViewsByObjectMetadataId() after. The view groups we get from the refresh obviously have different ids. It is not possible to indicate the ids of the view group through the update of the view as they are a side effect of the view update.


packages/twenty-front/src/modules/object-record/record-board/components/RecordBoardQueryEffect.tsx
```diff
@@ -63,13 +69,19 @@ export const RecordBoardQueryEffect = () => {
   const { triggerRecordBoardInitialQuery } =
     useTriggerRecordBoardInitialQuery();
 
+  const recordGroupdIds = useRecoilComponentValue(recordGroupIdsComponentState);
+  const recordGroupIdsHaveChanged = recordGroupdIds !== lastRecordGroupIds;
```
lucasbordeau: Shouldn't we do a deep compare here ? Otherwise this could be an array reference comparison.

lucasbordeau (approved):
LGTM

PR #16411, opened 2025-12-09
fix: ensure unique GraphQL schema caching per API key

abdulrahmancodes:
## Description

This PR fixes an issue where the GraphQL schema was being incorrectly cached and shared across different API keys within the same workspace. This resulted in the `createdBy` field (Actor) from the first API key's request being erroneously attributed to subsequent requests made by different API keys.

## Changes

- Updated the `@graphql-yoga/nestjs` patch to include the request's `Authorization` header in the schema cache key generation logic.
- This ensures that every unique authentication token (and thus every unique API key) generates a distinct cache entry, preventing schema context collisions.

Closes #15093


charlesBochet (changes requested):
@abdulrahmancodes thanks!

Question / feedbacks:
1) Could you document how you were able to generate the patch, I was struggling with it last time I try, if you found a reliable way to generate it, that's great, we should document it somewhere :)
2) I'm surprised about all the changes, I would expect that we just need to add the apiKeyId instead of workspaceUserId in the cacheKey.
3) storing authHeader as cacheKey is a bad idea, security wise

abdulrahmancodes:
@charlesBochet Thanks for the feedback.
I manually edited the patch file. The patch format is a standard git diff, so I located the cache key generation sections, added the `authHeader` extraction and included it in the `cacheKey`. Then I ran yarn to apply the patch and verify it worked. I wasn't aware of the autogeneration process

charlesBochet (approved):

PR #16414, opened 2025-12-09
Facebook links: display profile instead of full url

Bonapara:
Like we do for LinkedIn/Twitter


packages/twenty-front/src/modules/ui/field/display/components/LinkDisplay.tsx
```diff
@@ -26,9 +26,15 @@ export const LinkDisplay = ({ value }: LinkDisplayProps) => {
     ? LinkType.LinkedIn
     : displayedValue.startsWith('twitter.')
       ? LinkType.Twitter
-      : LinkType.Url;
-
-  if (type === LinkType.LinkedIn || type === LinkType.Twitter) {
+      : displayedValue.startsWith('facebook.')
+        ? LinkType.Facebook
+        : LinkType.Url;
```
abk404: not a blocker but some switch case statement would be better than tertiary conditions from readability standpoint.

charlesBochet (approved):

PR #16420, opened 2025-12-09
Fix flat entity maps date serialization

charlesBochet:
Changes:
- as we store date in redis as serialized, let's make all flatEntity dates as string. This requires changing FlatEntity types and making sure that entity are converted to flatEntity and flatEntity to dtos


packages/twenty-server/src/engine/metadata-modules/cron-trigger/services/workspace-flat-cron-trigger-map-cache.service.ts
```diff
@@ -46,6 +46,8 @@ export class WorkspaceFlatCronTriggerMapCacheService extends WorkspaceCacheProvi
         ]),
         universalIdentifier:
           cronTriggerEntity.universalIdentifier ?? cronTriggerEntity.id,
+        createdAt: cronTriggerEntity.createdAt.toISOString(),
```
Weiko, line 49: we could almost have a util
```typescript
return isDefined(date) ? new Date(date).toISOString() : new Date().toISOString()
```
maybe not worth it though...

Weiko (approved):
LGTM

PR #16421, opened 2025-12-09
Fix - revert standard sync logic on index + fix command

etiennejouan:


packages/twenty-server/src/database/commands/upgrade-version-command/1-13/1-13-rename-unique-index.command.ts
prastoin: ***Suggestion:*** We might wanna want a queryRunner here in order to avoid any issues

prastoin (approved):
Very good job !
Image

PR #16424, opened 2025-12-09
Fix is empty and non empty filter for numbers

thomtrp:
isNonEmptyString does not work for a number. We need to properly handle null and undefined


prastoin (approved):
Discussed and approved !

PR #16436, opened 2025-12-09
Standard views tooling

prastoin:
# Introduction
In this PR we've introduced the type safe generation tooling for standard view, view fields, view filters and view groups.
Also generated the ids before such as what was done for object and fields, it's still temporary as will be replaced by universalIdentifier mapping afterwards

## Task views
In this PR only implemented the `task` views and view groups, filters and fields in order to battle test all the generation tools
Will create a dedicated PR that will implement all the remaining views.

Related https://github.com/twentyhq/core-team-issues/issues/1995


packages/twenty-server/src/engine/workspace-manager/twenty-standard-application/types/twenty-standard-all-flat-entity-maps.type.ts
```diff
@@ -0,0 +1,12 @@
+import { type AllFlatEntityMaps } from 'src/engine/metadata-modules/flat-entity/types/all-flat-entity-maps.type';
+
+export type TwentyStandardAllFlatEntityMaps = Pick<
```
thomtrp, line 3: do we really need twenty-standard prefix rather than standard?

packages/twenty-server/src/engine/workspace-manager/twenty-standard-application/utils/field-metadata/compute-calendar-event-participant-standard-flat-field-metadata.util.ts
```diff
@@ -16,7 +16,7 @@ export const buildCalendarEventParticipantStandardFlatFieldMetadatas = ({
   now,
   objectName,
   workspaceId,
-  standardFieldMetadataIdByObjectAndFieldName,
+  standardObjectMetadataRelatedEntityIds,
```
thomtrp, line 19: what is an object related entity ?

thomtrp (approved):

packages/twenty-server/src/engine/workspace-manager/twenty-standard-application/utils/field-metadata/compute-calendar-event-participant-standard-flat-field-metadata.util.ts
prastoin: A given object, views, fields, view filter and so on

packages/twenty-server/src/engine/workspace-manager/twenty-standard-application/types/twenty-standard-all-flat-entity-maps.type.ts
prastoin: Tbh both sounds good to me, will remove in a next PR
Also note that this as code configuration will be deprecated in favor of an twenty-app instance in the future

PR #16444, opened 2025-12-09
Fix workspace invite onboarding loop

### LINKED ISSUE #16440: Users cannot join their workspace
## Bug Description
https://github.com/twentyhq/private-issues/issues/388

charlesBochet:
- The issue is that the listener is not triggered if they are no change to the workspaceMember (which happens when name is already filled through GoogleSSO)

Fixes https://github.com/twentyhq/twenty/issues/16440


Weiko (approved):
LGTM

PR #16447, opened 2025-12-09
Refactor global datasource part 3

Weiko:
## Context
Following https://github.com/twentyhq/twenty/pull/16399
Now using the new global orm manager everywhere and returning a GlobalDatasource/WorkspaceDatasource based on a feature flag.
This means we now need to wrap all our ORM calls within executeInWorkspaceContext callback (at least for now) so the global datasource can dynamically hydrate its context via the new store (the global datasource does not store anything related to workspaces as it is now a unique singleton). If feature flag is off it still uses local data stored in the workspace datasource.


packages/twenty-server/src/database/commands/command-runners/workspaces-migration.command-runner.ts
```diff
@@ -138,24 +140,32 @@ export abstract class WorkspacesMigrationCommandRunner<
...
+                  workspaceId,
+                )
+              : undefined;
+
+            await this.runOnWorkspace({
+              options,
               workspaceId,
-            })
-          : undefined;
+              dataSource,
```
Weiko, line 162: This shouldn't be needed later, datasource won't be workspace scoped so it won't be passed as parameter here but dynamically injected in the command via its context

charlesBochet (approved):
LGTM, let's goo!

PR #16448, opened 2025-12-09
fix messaging error parsing

neo773:


charlesBochet (approved):
Ty, Next steps:
1) merge isGmailApiError and isGmailAPIBatchError. This utils should typeguard and be very precise (make sure it's all error we know, we should check status, reasons, ...)
2) add test on top of isGmailAPIError with all known cases explictely declared  (like the prettified log you sent me above we should document them completely, the current generation in the logs is not good enough as the Gmail API is inconsistent. We should add all errors payloads for each of the use cases (folder fetch, messages, message-list, calendar-events...)
3) I would also use these mocks to test specifically: GmailFoldersErrorHandlerService, GmailMessagesImportErrorHandler, GmailEmailAliasErrorHandlerService, GmailMessageListFetchErrorHandler

PR #16452, opened 2025-12-10
fix(settings): preserve select option value when renaming label

Saikrishna1876:
This pull request addresses #16388.

Previously, when editing a select field option's label in the data model settings, the internal value was also being recalculated from the new label. This caused records with the old value to show blank cells in table views until a page refresh, because the stored value no longer matched any option.

For example, renaming 'Lead' to 'Hot Lead' would change the value from 'LEAD' to 'HOT_LEAD', breaking the mapping for existing records that still had 'LEAD' stored.

This fix changes the behavior so that:
- For NEW options: value is auto-generated from the label (existing behavior)
- For EXISTING options: value is preserved when editing the label

The fix uses the isNewRow prop to determine whether to auto-generate the value. This ensures that label changes are immediately reflected in the UI without breaking the value-to-option mapping for existing records.

Fixes: select field option rename causing blank cells in table view


ijreilly (changes requested):
Hi @Saikrishna1876 thanks for contributing ! 
Even if your implementation solves the current issues, it introduces a new behaviour we don't want to adopt at the moment: decorrelating name and label for options. It could produce behaviours that would be hard to understand / fix for some users; for instance if I update a label' name then create a new one with the same previous name: (her I toggle on / off the advanced settings which allow me to see how the names are created/updated): 

https://github.com/user-attachments/assets/52ab73dd-c0aa-433d-bea6-473e3d909d61

Instead, we should probably invalidate the cache we have associated to findManyOpportunities (or else).

Saikrishna1876:
Yes, you are right.

ijreilly:
are you willing to work on it again or shall i close the pr @Saikrishna1876 ?

Saikrishna1876:
> are you willing to work on it again or shall i close the pr @Saikrishna1876 ?

I have tried a few solutions, but I have failed to solve the bug. I have nothing left to try, you can close the PR.
Thank you for you cooperation @ijreilly.

PR #16455, opened 2025-12-10
Feat/wrap nav drawer with overflowing text with tooltip

Abhay-86:
# **feat: Add tooltips for truncated names in navigation components**

### **Summary**

Adds hover tooltips across navigation components so users can view full names when text is truncated due to limited space.

---

# **Changes Made**

### **Sidebar Navigation**

* Added tooltips to navigation drawer items and menu labels.
* Long workspace or company names now show full text on hover.

### **Top Bar**

* Added tooltip support to the View Picker dropdown header.
* Long view names (e.g., *“AuraML, very very long name”*) now reveal their full value on hover.

---

# **Implementation**

* Integrated `OverflowingTextWithTooltip` from `twenty-ui/display`.
* Wrapped navigation and view labels with the component.
* Updated layout styles to allow accurate overflow detection.
* Tooltip appears **only** when truncation occurs; short names show no tooltip.

---

# **User Experience**

Users can hover over any truncated workspace, company, navigation, or view name to see the complete text—improving clarity, accessibility, and overall navigation usability.

[https://github.com/user-attachments/assets/97f9ea6c-409c-4ec5-a27c-a7b1c5aa5682](https://github.com/user-attachments/assets/97f9ea6c-409c-4ec5-a27c-a7b1c5aa5682)

---

### **Fixes**

Closes **#16412**

---

### DIFF
--- a/packages/twenty-front/src/modules/ui/navigation/navigation-drawer/components/NavigationDrawerItem.tsx
+++ b/packages/twenty-front/src/modules/ui/navigation/navigation-drawer/components/NavigationDrawerItem.tsx
@@ -17,6 +17,7 @@ import {
   AppTooltip,
   type IconComponent,
   Label,
+  OverflowingTextWithTooltip,
   type TablerIconsProps,
   TooltipDelay,
   TooltipPosition,
@@ -345,7 +346,9 @@ export const NavigationDrawerItem = ({
 
           <StyledLabelParent>
             <StyledEllipsisContainer>
-              <StyledItemLabel>{label}</StyledItemLabel>
+              <StyledItemLabel>
+                <OverflowingTextWithTooltip text={label} />
+              </StyledItemLabel>
               {secondaryLabel && (
                 <StyledItemSecondaryLabel>
                   {' · '}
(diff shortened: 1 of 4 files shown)


Devessier:
@Abhay-86, I dived into the `OverflowingTextWithTooltip` component and the `delay={TooltipDelay.mediumDelay}` parameter does absolutely nothing.

The component tries to fix several issues at once and is now a bit convoluted. Are these my understandings:

- Passing `isOpen={true}` turns the tooltip into a controlled component. The delay property will have no effect in this mode.
- We use a portal so that tooltips are always displayed on top of any content, including nested dropdowns.

I think we should go with a fully controlled component. Let's keep that update for later.

For now, we'll go with no delay at all.

Devessier (approved):

Abhay-86:
> @Abhay-86, I dived into the `OverflowingTextWithTooltip` component and the `delay={TooltipDelay.mediumDelay}` parameter does absolutely nothing.
> 
> The component tries to fix several issues at once and is now a bit convoluted. Are these my understandings:
> 
> * Passing `isOpen={true}` turns the tooltip into a controlled component. The delay property will have no effect in this mode.
> * We use a portal so that tooltips are always displayed on top of any content, including nested dropdowns.
> 
> I think we should go with a fully controlled component. Let's keep that update for later.
> 
> For now, we'll go with no delay at all.

Thanks for confirming! I noticed the same behavior while debugging — the delay wasn’t having any effect because the component becomes controlled when isOpen={true}.

I didn’t comment earlier as you were working on that, but I had reached the same conclusion.

Going with no delay for now makes sense. 
Also, if there’s anything else I can help with in the meantime, feel free to let me know - I’m already looking for more issues to contribute to.

Thank you 😊

PR #16458, opened 2025-12-10
Prevent seeding empty configuration for dashboard widgets

Devessier:
We used to seed widgets with empty configuration when the fields they required weren't defined. I suggest we only seed properly defined widgets.


bosiraphael (approved):
LGTM

PR #16459, opened 2025-12-10
Fix sync role v1

Weiko:
## Context
SyncMetadata for role update was broken when those roles had permission flags in them.
This PR https://github.com/twentyhq/twenty/issues/16365 introduced a change in all existing standard roles which probably triggered the update of faulty roles for the first time. Down the line, some properties from flat were not omit before querying the DB (in this case permissionFlagIds does not exist as a column in role table)

The fix is definitely not great, this also probably broke the update of permissionFlags for existing STANDARD roles but it's not subject to change (It actually never changed. Again, what has been updated was not the permission flag but a new boolean introduced 2 days ago which forced the "UPDATE" path in the sync which was probably never triggered before, or at least not for roles with permission flags which are "quite" new). Also we are about to remove this whole code in favor of the new migration v2

### DIFF
--- a/packages/twenty-server/src/engine/metadata-modules/flat-role/utils/from-standard-role-definition-to-flat-role.util.ts
+++ b/packages/twenty-server/src/engine/metadata-modules/flat-role/utils/from-standard-role-definition-to-flat-role.util.ts
@@ -1,3 +1,4 @@
+import { removePropertiesFromRecord } from 'twenty-shared/utils';
 import { v4 } from 'uuid';
 
 import { type FlatRole } from 'src/engine/metadata-modules/flat-role/types/flat-role.type';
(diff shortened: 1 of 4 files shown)


packages/twenty-server/src/engine/metadata-modules/flat-role/utils/from-standard-role-definition-to-flat-role.util.ts
```diff
@@ -10,7 +11,7 @@ export const fromStandardRoleDefinitionToFlatRole = (
   const createdAt = new Date().toISOString();
 
   return {
-    ...standardRoleDefinition,
```
Weiko, line 13: fromStandardRoleDefinitionToFlatRole was returning a FlatRole with permissionFlags in it which is wrong

charlesBochet (approved):

prastoin (commented):
Ok with that !
Note that everything here is going to be deprecated with the sync metadata

PR #16462, opened 2025-12-10
Fix line break in richTextV2 field workflow editor

### LINKED ISSUE #14976: Emails, Tasks and Notes body no longer support line break
## Bug Description

Notes and tasks and emails were created but without the line break since a few days. 

Image
Image

etiennejouan:
Api expect markdown format where '\n' is escape (but not '  \n')

fixes https://github.com/twentyhq/twenty/issues/14976


packages/twenty-front/src/modules/workflow/workflow-variables/utils/parseEditorContent.ts
```diff
@@ -3,6 +3,7 @@ import { isDefined } from 'twenty-shared/utils';
 
 export const parseEditorContent = (json: JSONContent): string => {
   const parseNode = (node: JSONContent): string => {
+    console.log(node);
```
thomtrp: remove log

thomtrp (approved):

PR #16473, opened 2025-12-10
fix process group emails

neo773:
Added guard check and replaced offset pagination with cursor pagination


charlesBochet (approved):

packages/twenty-server/src/modules/messaging/message-import-manager/services/messaging-delete-group-email-messages.service.ts
```diff
@@ -119,9 +136,7 @@ export class MessagingDeleteGroupEmailMessagesService {
             break;
```
prastoin, line 136: ***Question:*** Unless I'm mistaken this now won't ever be reached ?

prastoin (commented):
Hey there, thanks for the fix
Please provide minimal PR description explaining the initial root cause and motivations and the fix it adds too
It will be useful if anyone deploying this commit to prod ( Will always prefer a human one-liner sentence rather than a whole auto-generated mardown file ) 
Thanks !

PR #16479, opened 2025-12-10
Message folders optimization

neo773:
- Batch Gmail API calls using `googleapis-batcher` for folder processing
  - Add concurrency limit for Microsoft Graph folder processing 
  - Skip IMAP folder sync when no new messages (checks UIDVALIDITY/MODSEQ)
  - Refactored `syncMessageFolders` to return folder state directly, avoiding extra DB round-trips
  - Refactored `processPendingFolderActions` to reuse state instead of querying DB again
  - Add unique index on message folders entity


packages/twenty-server/src/modules/messaging/message-folder-manager/services/sync-message-folders.service.ts
```diff
@@ -223,55 +80,160 @@ export class SyncMessageFoldersService {
...
-    });
-
-    const existingFolderMap = new Map<string, MessageFolderWorkspaceEntity>();
+  private async syncFolderChanges(
+    discoveredFolders: DiscoveredMessageFolder[],
+    existingFolders: MessageFolder[],
+    messageChannelId: string,
+    workspaceId: string,
+  ): Promise<MessageFolder[]> {
+    const existingFoldersByExternalId = new Map(
```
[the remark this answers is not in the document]
neo773: yeah except this isn't possible API will always return id

packages/twenty-server/src/engine/core-modules/auth/services/create-message-channel.service.ts
charlesBochet: let's use insert here

packages/twenty-server/src/engine/core-modules/auth/services/create-message-channel.service.ts
```diff
@@ -81,17 +70,20 @@ export class CreateMessageChannelService {
...
+        const createdMessageChannel = await messageChannelRepository.findOne({
+          where: { id: newMessageChannel.id },
+          relations: ['connectedAccount', 'messageFolders'],
+        });
+
+        if (!isDefined(createdMessageChannel)) {
+          throw new Error('Message channel not found');
         }
 
+        await this.syncMessageFoldersService.syncMessageFolders(
```
charlesBochet: when we pass more than one Param, let's pass an object

packages/twenty-server/src/modules/messaging/common/standard-objects/message-folder.workspace-entity.ts
```diff
@@ -39,6 +40,10 @@ registerEnumType(MessageFolderPendingSyncAction, {
 })
 @WorkspaceIsNotAuditLogged()
 @WorkspaceIsSystem()
+@WorkspaceIndex(['messageChannelId', 'externalId'], {
```
charlesBochet, line 43: syncMetadata is deprecated, we will need to address this later

packages/twenty-server/src/modules/messaging/message-folder-manager/services/sync-message-folders.service.spec.ts
```diff
@@ -0,0 +1,519 @@
...
+        const messageChannel = createMockMessageChannel({
+          messageFolders: [],
+        });
+
+        gmailGetAllFoldersService.getAllMessageFolders.mockResolvedValue(
+          discoveredFolders,
+        );
+
+        const result = await service.syncMessageFolders(
+          messageChannel as never,
```
charlesBochet: as never!?

packages/twenty-server/src/modules/messaging/message-folder-manager/services/sync-message-folders.service.spec.ts
```diff
@@ -0,0 +1,519 @@
...
+import {
+  MessageFolderPendingSyncAction,
+  type MessageFolderWorkspaceEntity,
+} from 'src/modules/messaging/common/standard-objects/message-folder.workspace-entity';
+import { GmailGetAllFoldersService } from 'src/modules/messaging/message-folder-manager/drivers/gmail/services/gmail-get-all-folders.service';
+import { ImapGetAllFoldersService } from 'src/modules/messaging/message-folder-manager/drivers/imap/services/imap-get-all-folders.service';
+import { MicrosoftGetAllFoldersService } from 'src/modules/messaging/message-folder-manager/drivers/microsoft/services/microsoft-get-all-folders.service';
+import { SyncMessageFoldersService } from 'src/modules/messaging/message-folder-manager/services/sync-message-folders.service';
+
+type SyncedMessageFolder = Pick<
```
charlesBochet, line 23: to be honest, I don't believe in unit testing services and mocking all there dependencies. We should aim for:
- unit tests on utils
- integration tests on jobs (which is not here yet but soon)

packages/twenty-server/src/modules/messaging/message-folder-manager/services/sync-message-folders.service.spec.ts
charlesBochet: typing fixed

packages/twenty-server/src/modules/messaging/message-folder-manager/services/sync-message-folders.service.spec.ts
charlesBochet: we can keep for now, but let's not invest further on these tests

packages/twenty-server/src/modules/messaging/common/standard-objects/message-folder.workspace-entity.ts
charlesBochet: let's keep this in mind as a follow up

charlesBochet (approved):

PR #16483, opened 2025-12-11
Display a CTA to view the existing duplicate when adding a duplicate email/domain.

### LINKED ISSUE #89: Display a CTA to view the existing duplicate when adding a duplicate email/domain
We recently added the ability to enforce unicity constraints, which we've applied on `company.domainNames` and `person.emails`.

It's a bit frustrating when your record doesn't get saved, and you don't really know what record is causing the issue.

We should fetch the conflicting record and display a CTA in the notification popup that would lead the user to the conflicting record.

mabdullahabaid:
Closes [89](https://github.com/twentyhq/core-team-issues/issues/89)

## Problem

When users attempt to create or update a record with a duplicate value for a unique field (e.g., duplicate email or domain name), they receive a generic error message: "This record already exists. Please check your data and try again." This provides no actionable way to locate and view the existing conflicting record, forcing users to manually search for it.

## Solution

This PR enhances duplicate key constraint error handling to automatically detect the conflicting record and display a "View existing record" link in the error notification. When clicked, users are navigated directly to the existing record's detail page.

### 1. PostgreSQL Error Parsing (`parse-postgres-constraint-error.util.ts`)
- Extracts structured information from PostgreSQL `QueryFailedError` messages.

### 2. Conflicting Record Lookup (`find-conflicting-record.util.ts`)
- Queries the database to find the existing record with the conflicting value

### 3. Error Handling Orchestration (`handle-duplicate-key-error.util.ts`)
- Parses PostgreSQL error to extract column name and conflicting value
- Attempts to find the conflicting record
- Enriches `TwentyORMException` with `conflictingRecordId` and `conflictingObjectNameSingular` if found

### 4. Exception Computation Updates (`compute-twenty-orm-exception.ts`)
- Made function `async` and added optional `entityManager` and `internalContext` parameters
- Needed to support async database queries for conflicting record lookup

### 5. GraphQL Error Handler (`twenty-orm-graphql-api-exception-handler.util.ts`)
- **Changes**: Enhanced `DUPLICATE_ENTRY_DETECTED` case to include `conflictingRecordId` and `conflictingObjectNameSingular` in GraphQL error extensions

### 1. Error Extraction Utility (`get-conflicting-record-from-apollo-error.util.ts`)
- Accesses GraphQL error extensions
- Validates that both `conflictingRecordId` and `conflictingObjectNameSingular` exist and are strings
- Returns `null` if validation fails

### 2. SnackBar Enhancement (`useSnackBar.ts`)
- Extracts conflicting record info from Apollo error
- Constructs URL using `getAppPath` utility
- Adds link object to snackbar options with text "View existing record"

image

### DIFF
--- a/packages/twenty-front/src/modules/ui/feedback/snack-bar-manager/hooks/useSnackBar.ts
+++ b/packages/twenty-front/src/modules/ui/feedback/snack-bar-manager/hooks/useSnackBar.ts
@@ -8,12 +8,11 @@ import {
   snackBarInternalComponentState,
   type SnackBarOptions,
 } from '@/ui/feedback/snack-bar-manager/states/snackBarInternalComponentState';
+import { buildErrorAction } from '@/ui/feedback/snack-bar-manager/utils/build-error-action.util';
 import { useAvailableComponentInstanceIdOrThrow } from '@/ui/utilities/state/component-state/hooks/useAvailableComponentInstanceIdOrThrow';
 import { type ApolloError } from '@apollo/client';
 import { t } from '@lingui/core/macro';
-import { AppPath } from 'twenty-shared/types';
-import { getAppPath, isDefined } from 'twenty-shared/utils';
-import { getConflictingRecordFromApolloError } from '~/utils/get-conflicting-record-from-apollo-error.util';
+import { isDefined } from 'twenty-shared/utils';
 import { getErrorMessageFromApolloError } from '~/utils/get-error-message-from-apollo-error.util';
 
 export const useSnackBar = () => {
@@ -145,26 +144,13 @@ export const useSnackBar = () => {
           ? getErrorMessageFromApolloError(apolloError)
           : t`An error occurred.`;
 
-      const conflictingRecord = apolloError
-        ? getConflictingRecordFromApolloError(apolloError)
-        : null;
-
-      const link = conflictingRecord
-        ? {
-            href: getAppPath(AppPath.RecordShowPage, {
-              objectNameSingular:
-                conflictingRecord.conflictingObjectNameSingular,
-              objectRecordId: conflictingRecord.conflictingRecordId,
-            }),
-            text: t`View existing record`,
-          }
-        : undefined;
+      const errorAction = buildErrorAction(apolloError);
 
       setSnackBarQueue({
         id: uuidv4(),
         message: errorMessage,
+        ...errorAction,
         ...options,
-        link,
         variant: SnackBarVariant.Error,
       });
     },
--- a/packages/twenty-front/src/modules/ui/feedback/snack-bar-manager/hooks/useSnackBar.ts
+++ b/packages/twenty-front/src/modules/ui/feedback/snack-bar-manager/hooks/useSnackBar.ts
@@ -11,7 +11,9 @@ import {
 import { useAvailableComponentInstanceIdOrThrow } from '@/ui/utilities/state/component-state/hooks/useAvailableComponentInstanceIdOrThrow';
 import { type ApolloError } from '@apollo/client';
 import { t } from '@lingui/core/macro';
-import { isDefined } from 'twenty-shared/utils';
+import { AppPath } from 'twenty-shared/types';
+import { getAppPath, isDefined } from 'twenty-shared/utils';
+import { getConflictingRecordFromApolloError } from '~/utils/get-conflicting-record-from-apollo-error.util';
 import { getErrorMessageFromApolloError } from '~/utils/get-error-message-from-apollo-error.util';
 
 export const useSnackBar = () => {
@@ -143,10 +145,26 @@ export const useSnackBar = () => {
           ? getErrorMessageFromApolloError(apolloError)
           : t`An error occurred.`;
 
+      const conflictingRecord = apolloError
+        ? getConflictingRecordFromApolloError(apolloError)
+        : null;
+
+      const link = conflictingRecord
+        ? {
+            href: getAppPath(AppPath.RecordShowPage, {
+              objectNameSingular:
+                conflictingRecord.conflictingObjectNameSingular,
+              objectRecordId: conflictingRecord.conflictingRecordId,
+            }),
+            text: t`View existing record`,
+          }
+        : undefined;
+
       setSnackBarQueue({
         id: uuidv4(),
         message: errorMessage,
         ...options,
+        link,
         variant: SnackBarVariant.Error,
       });
     },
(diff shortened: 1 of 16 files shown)


FelixMalfait:
Hey great, thanks!
This record already exist is a bit vague, can we be explicit about the duplicate field's label instead? 
Also tests are broken? 
(didn't look a the code, will leave the review to someone else)

packages/twenty-front/src/modules/ui/feedback/snack-bar-manager/hooks/useSnackBar.ts
```diff
@@ -143,10 +145,26 @@ export const useSnackBar = () => {
           ? getErrorMessageFromApolloError(apolloError)
           : t`An error occurred.`;
 
+      const conflictingRecord = apolloError
```
FelixMalfait: This is too specific to your problem, try to generalize this pattern. There are probably other use-cases for a CTA in the snackbar

packages/twenty-server/src/engine/api/graphql/workspace-query-runner/utils/handle-duplicate-key-error.util.ts
```diff
@@ -1,27 +1,63 @@
...
+  );
+
+  const exception = new TwentyORMException(
     `A duplicate entry was detected`,
     TwentyORMExceptionCode.DUPLICATE_ENTRY_DETECTED,
     {
       userFriendlyMessage: msg`This record already exists. Please check your data and try again.`,
     },
-  );
+  ) as DuplicateKeyErrorWithMetadata;
```
FelixMalfait: Can you type the exception directly instead of using AS?

packages/twenty-server/src/engine/twenty-orm/repository/workspace-select-query-builder.ts
```diff
@@ -126,24 +126,24 @@ export class WorkspaceSelectQueryBuilder<
   }
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  override getRawOne<U = any>(): Promise<U | undefined> {
+  override async getRawOne<U = any>(): Promise<U | undefined> {
     try {
       this.validatePermissions();
 
-      return super.getRawOne();
+      return await super.getRawOne();
```
FelixMalfait: These async await changes could be quite dangerous/impactful (affect lots of areas). Are you sure this is the way to go?

packages/twenty-server/src/engine/twenty-orm/utils/twenty-orm-graphql-api-exception-handler.util.ts
```diff
@@ -13,10 +20,24 @@ export const twentyORMGraphqlApiExceptionHandler = (
...
-    case TwentyORMExceptionCode.TOO_MANY_RECORDS_TO_UPDATE:
-      throw new UserInputError(error.message, {
+    case TwentyORMExceptionCode.TOO_MANY_RECORDS_TO_UPDATE: {
+      const errorWithMetadata = error as DuplicateKeyErrorWithMetadata;
+      const extensions: Record<string, unknown> = {
         userFriendlyMessage: error.userFriendlyMessage,
-      });
+      };
+
+      if (
```
FelixMalfait: Why do an if if you're within a switch case for TwentyORMExceptionCode already, seems odd

packages/twenty-server/src/engine/api/graphql/workspace-query-runner/utils/handle-duplicate-key-error.util.ts
```diff
@@ -1,27 +1,63 @@
...
 }
 
-export const handleDuplicateKeyError = (
-  _error: PostgreSQLError,
-  _objectMetadata: FlatObjectMetadata,
-) => {
-  // Since we no longer have indexMetadatas in FlatObjectMetadata,
-  // we provide a generic error message
-  throw new TwentyORMException(
+export const handleDuplicateKeyError = async (
```
FelixMalfait, line 23: @Weiko could you please review this PR? I made a few minor comments but I'm not sure if it's the right approach overall?

FelixMalfait: For the upsert we probably have tooling to find any potential conflict with an index? One approach is to extract the index from the error message but the other approach could be to do a clean request to find the conflicting record when we catch an exception - I guess that code can be mutualized with something existing for upserts (?)

mabdullahabaid:
> Hey great, thanks! This record already exist is a bit vague, can we be explicit about the duplicate field's label instead? Also tests are broken? (didn't look a the code, will leave the review to someone else)

Hey, yes! Resolved most of the comments, moving to the one about the Snackbar now.

Not entirely sure about the approach though - awaiting response from @Weiko on that.

packages/twenty-server/src/engine/api/graphql/workspace-query-runner/utils/handle-duplicate-key-error.util.ts
mabdullahabaid: Done.

packages/twenty-server/src/engine/twenty-orm/utils/twenty-orm-graphql-api-exception-handler.util.ts
mabdullahabaid: Fixed.

packages/twenty-server/src/engine/twenty-orm/repository/workspace-select-query-builder.ts
mabdullahabaid: We do not need await for getRawOne, getRawMany, and getCount because we can return an unresolved promise for the call-sites to await them.

However, we do need await for computeTwentyORMException because we need to return the Error object, not the Promise.

Removed the await pointed out in the comment and fixed.

packages/twenty-front/src/modules/ui/feedback/snack-bar-manager/hooks/useSnackBar.ts
mabdullahabaid: I have tried generalizing with actionText, actionOnClick, actionTo props. It's a pattern used by Info.tsx component in twenty-ui, which has props buttonTitle, onClick, and to props.

build-error-action allows introducing more types of actions that the useSnackBar hook then passes into SnackBar (e.g. retry, refetch or something else).

Weiko:
@mabdullahabaid this approach LGTM. I think the tradeoff is acceptable as long as the extra request is done in the "catch" part

Weiko (approved):
There is now a coupling between error handling and I/O (`throw new computeTwentyORMException` will query the DB, I'm not sure about that pattern tbh 🤔. There is a fallback behavior in case the query fails which is good though 👍 
LGTM, I might have a second look later this week as I'm refactoring some parts for RLS integration and I will let you know

mabdullahabaid:
@Weiko sounds good, thanks!

PR #16488, opened 2025-12-11
1895 extensibility v1 application tokens 2

### LINKED ISSUE #1895: [Extensibility] [v1] Application tokens

martmull:
- moves applicationRoleId to application entity
- add new `APPLICATION` FieldActorSource and `APPLICATION` JwtTokenTypeEnum value
- create a new token with applicationId when executing a function
- when applicationId is in token, check for application.defaultRole permissions 
-use twenty-shared types in `twenty-sdk/application`

TODO in another PR -> do the same roleIds check when using an API KEY token

Result
image

image


martmull:
closing as handled by https://github.com/twentyhq/twenty/pull/16504

PR #16490, opened 2025-12-11
Fix overflow text with tooltip

Devessier:
This PR allows the `OverflowingTextWithTooltip` component to take complex content as parameter. It fixes the buggy display of long records pinned as favorites.

https://github.com/user-attachments/assets/0224da09-8547-4def-927d-3f545406287e

Created extensive stories to test every case. Cleaned the component.

## Demo

https://github.com/user-attachments/assets/a8f257b5-2baf-42da-897b-a76b8a9ab664


guillim (approved):

guillim (approved):
LGTM

PR #16491, opened 2025-12-11
Remove sync metadata from upgrade

prastoin:
# Introduction
Related https://github.com/twentyhq/core-team-issues/issues/1910
From now on the upgrade won't integrate any sync metadata as it's going to be deprecated very soon
Any updates to about to removes workspace-entity or standard flat entity will require a dedicated upgrade command, what we've already started doing during the 1.13 sprint, until we have totally migrated the v2 to be workspace agnostic


Weiko (approved):
LGTM!

PR #16492, opened 2025-12-11
Fix default order by

bosiraphael:
Always apply an order by on the group by

### DIFF
--- a/packages/twenty-front/src/modules/page-layout/widgets/graph/utils/generateGroupByQueryVariablesFromBarOrLineChartConfiguration.ts
+++ b/packages/twenty-front/src/modules/page-layout/widgets/graph/utils/generateGroupByQueryVariablesFromBarOrLineChartConfiguration.ts
@@ -1,5 +1,6 @@
 import { type ObjectMetadataItem } from '@/object-metadata/types/ObjectMetadataItem';
 import { GRAPH_DEFAULT_DATE_GRANULARITY } from '@/page-layout/widgets/graph/constants/GraphDefaultDateGranularity.constant';
+import { GRAPH_DEFAULT_ORDER_BY } from '@/page-layout/widgets/graph/constants/GraphDefaultOrderBy';
 import { getGroupByOrderBy } from '@/page-layout/widgets/graph/utils/getGroupByOrderBy';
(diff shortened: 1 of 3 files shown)


lucasbordeau:
Is this what is reflected in the UI ? Do we have asc by default selected ?

packages/twenty-front/src/modules/page-layout/widgets/graph/utils/generateGroupByQueryVariablesFromBarOrLineChartConfiguration.ts
```diff
@@ -117,24 +118,21 @@ export const generateGroupByQueryVariablesFromBarOrLineChartConfiguration = ({
...
-    );
-  }
-  if (
-    isDefined(groupByFieldY) &&
-    isDefined(chartConfiguration.secondaryAxisOrderBy)
-  ) {
+  orderBy.push(
+    getGroupByOrderBy({
+      graphOrderBy:
+        chartConfiguration.primaryAxisOrderBy ?? GraphOrderBy.FIELD_ASC,
```
ehconitin: the best thing here would be to have a `const DEFAULT_ORDER_BY = GraphOrderBy.FIELD_ASC;`
and use it here

packages/twenty-front/src/modules/page-layout/widgets/graph/utils/generateGroupByQueryVariablesFromBarOrLineChartConfiguration.ts
ehconitin, line 1: we can remove the dead code on line 165
` ...(orderBy.length > 0 && { orderBy }),`

since now orderBy is always populated -- but harmless

ehconitin (approved):
LGTM! left 2 comments

packages/twenty-front/src/modules/page-layout/widgets/graph/utils/generateGroupByQueryVariablesFromBarOrLineChartConfiguration.ts
bosiraphael: Yes!

PR #16498, opened 2025-12-11
fix: constrain side column width

Devessier:
## Before

https://github.com/user-attachments/assets/18bfd8b0-7636-40b2-aa64-24c84ab74009

## After

CleanShot 2025-12-11 at 14 13 23@2x


Weiko (approved):
LGTM

PR #16502, opened 2025-12-11
Fix raw json field display in read only

etiennejouan:
Preview
Screenshot 2025-12-11 at 15 54 07

Fixes https://github.com/twentyhq/twenty/issues/16311


Devessier (approved):
Great job, Mister Proust 💪

PR #16504, opened 2025-12-11
1895 extensibility v1 application tokens 3

### LINKED ISSUE #1895: [Extensibility] [v1] Application tokens

martmull:
- moves applicationRoleId to application entity
- add new `APPLICATION` FieldActorSource and `APPLICATION` JwtTokenTypeEnum value
- create a new token with applicationId when executing a function
- when applicationId is in token, check for application.defaultRole permissions 
-use twenty-shared types in `twenty-sdk/application`

Result
image

image

image


packages/twenty-server/@types/jest.d.ts
```diff
@@ -15,7 +15,6 @@ declare module '@jest/types' {
       INVALID_ACCESS_TOKEN: string;
       APPLE_JONY_MEMBER_ACCESS_TOKEN: string;
       APPLE_PHIL_GUEST_ACCESS_TOKEN: string;
-      ACME_JONY_MEMBER_ACCESS_TOKEN: string;
```
Weiko, line 18: why removing those?

packages/twenty-server/src/engine/core-modules/application/application.entity.ts
```diff
@@ -60,6 +61,12 @@ export class ApplicationEntity {
   @Column({ nullable: true, type: 'uuid' })
   serverlessFunctionLayerId: string | null;
 
+  @Column({ nullable: true, type: 'uuid' })
+  defaultRoleId: string | null;
+
+  @Field(() => RoleDTO, { nullable: true })
+  defaultRole: RoleDTO | null;
```
Weiko: Not clear to me why it's called "default" role. Don't we expect to have 1 role per application?

packages/twenty-server/src/engine/core-modules/application/application.exception.ts
```diff
@@ -10,3 +10,8 @@ export enum ApplicationExceptionCode {
   APPLICATION_NOT_FOUND = 'APPLICATION_NOT_FOUND',
   FORBIDDEN = 'FORBIDDEN',
 }
+
+export const ApplicationNotFoundDefaultError = new ApplicationException(
```
Weiko: We don't really have this pattern, prefer to have a new error code in ApplicationExceptionCode or maybe I'm missing something?

packages/twenty-server/src/engine/core-modules/file/guards/file-path-guard.ts
```diff
@@ -28,7 +25,6 @@ export class FilePathGuard implements CanActivate {
     try {
       const payload = await this.jwtWrapperService.verifyJwtToken(
         fileSignature,
-        JwtTokenTypeEnum.FILE,
```
Weiko, line 31: this info is already in the token?

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
```diff
@@ -52,30 +47,23 @@ export class JwtWrapperService {
...
+    // tokens. Changing that would break the existing api_key, so we leave it
+    // like this for now
+    // (see api-key.service.ts generateApiKeyToken method)
+    const type =
+      payload.type === JwtTokenTypeEnum.API_KEY
+        ? JwtTokenTypeEnum.ACCESS
+        : payload.type;
 
     // TODO: check if this is really needed
     if (type !== 'FILE' && !payload.sub) {
```
Weiko: nit: can compare with enum instead of string

packages/twenty-server/src/engine/core-modules/actor/utils/build-created-by-from-application.util.ts
```diff
@@ -0,0 +1,15 @@
...
+import { type ApplicationEntity } from 'src/engine/core-modules/application/application.entity';
+
+type BuildCreatedByFromApplicationArgs = {
+  application: ApplicationEntity;
+};
+export const buildCreatedByFromApplication = ({
+  application,
+}: BuildCreatedByFromApplicationArgs): ActorMetadata => ({
+  source: FieldActorSource.APPLICATION,
+  name: application.name,
```
Weiko, line 12: application.name is mutable, should we include the id as well?

packages/twenty-server/src/engine/core-modules/application/application-sync.service.ts
```diff
@@ -106,6 +104,14 @@ export class ApplicationSyncService {
       });
     }
 
+    if (isDefined(application.defaultRoleId)) {
+      await this.syncApplicationRolePermissions({
```
Weiko: This means we don't sync role permissions the first time we sync the app because defaultRoleId will be empty at this point. This will force 2 sync instead of one 🤔

Weiko (commented):
Note:
- Seems we don't handle field permission sync?
- Why did we remove applicationId in target? I'm not fully sure but this would mean permission handling would be done differently for application then (we are caching targets and we base permissions on that for now)
- As mentioned, not sure about the naming for defaultRole in application
- Issue with permissions not being synced the first time due to defaultRoleId missing

packages/twenty-server/src/engine/core-modules/file/guards/file-path-guard.ts
Weiko: Ok this was for legacy tokens indeed, should be fine to drop these now 👍

packages/twenty-server/@types/jest.d.ts
martmull: this is not used

packages/twenty-server/src/engine/core-modules/application/application.exception.ts
martmull: it is done by other exceptions like in `workspace.exception.ts`
I just followed the existing patterns

packages/twenty-server/src/engine/core-modules/file/guards/file-path-guard.ts
martmull: Yes, but we still need to maintain a specific case for apiKey which where generated with ACCESS type secrets

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
martmull: I was not sure so I did not applied the changes everywhere, but indeed

packages/twenty-server/src/engine/core-modules/actor/utils/build-created-by-from-application.util.ts
martmull: Yes, nice idea. We currently don't do that for the other actors like apiKey or fullName (see `packages/twenty-server/src/engine/core-modules/actor/utils`)
I think this can be done in another PR, for all actors

packages/twenty-server/src/engine/core-modules/application/application-sync.service.ts
martmull: The role is created or updated in `syncApplication`, at first sync.
This line is creating role **permissions**. We do it after `syncObjects` call in case it applies permissions on object created by the application

packages/twenty-server/src/engine/core-modules/application/application.entity.ts
martmull: Indeed, when discussing that with Félix, "default" was refering to the `workspace.defaultRoleId` column. But it might be different. 
We may call it `roleId` indeed, or use the standard architecture for roles (using the roleTarget)
@FelixMalfait what do you think? Do you confirm `defaultRole`?

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
martmull: updated

martmull:
> * Seems we don't handle field permission sync?
Yes indeed, I tried to keep PR not too big

packages/twenty-server/src/engine/core-modules/application/application.exception.ts
martmull: will do

packages/twenty-server/src/engine/core-modules/file/guards/file-path-guard.ts
martmull: add 2 decoding strategy for apiKeys and clean legacy choice of secret with access type or api keys

packages/twenty-server/src/engine/core-modules/application/application.entity.ts
martmull: application will be able to declare multiple roles
- defaultROle -> defaultServerlessFunctionRole
A role cannot target an application

martmull:
Check if we can merge application_TOKEN into access_token

martmull:
> application_TOKEN

Will do in another PR as this is not straighforward and this is already a big PR

.yarn/patches/@genql-cli-npm-3.0.5-6ca1fa6f53.patch
```diff
@@ -0,0 +1,334 @@
+diff --git a/dist/render/client/renderClient.js b/dist/render/client/renderClient.js
+index ac9532b2f6a4ff2cbb443f1d7cba39eee5d803ec..bee7e4d1e8655e3498851c397a9694801d84a9eb 100644
+--- a/dist/render/client/renderClient.js
++++ b/dist/render/client/renderClient.js
+@@ -19,6 +19,46 @@ function(options${url ? '?' : ''}: ClientOptions): Client {
```
FelixMalfait: Let's try to do without a patch

packages/twenty-apps/hello-world/application.config.ts
```diff
@@ -1,22 +1,45 @@
...
-      universalIdentifier: 'ef8ab489-e68a-4841-b402-261f440e6185',
-      description: 'Twenty API Url',
+    DEFAULT_RECIPIENT_NAME: {
+      universalIdentifier: '19e94e59-d4fe-4251-8981-b96d0a9f74de',
+      description: 'Default recipient name for postcards',
+      value: 'Alex Karp',
       isSecret: false,
     },
   },
+  role: {
```
FelixMalfait: defaultRoleId only here, folder for multiples roles

FelixMalfait: Only id should be in application.config ; roles should be in a dedicated folder, with the same metadata sync mechanism as other objects

packages/twenty-server/src/database/typeorm/core/migrations/common/1765206100942-updateRoleColumns.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export class UpdateRoleColumns1765206100942 implements MigrationInterface {
+  name = 'UpdateRoleColumns1765206100942';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "core"."roleTarget" DROP COLUMN "targetApplicationId"`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "core"."application" ADD "defaultServerlessFunctionRoleId" uuid`,
```
Weiko, line 11: Still not sure why we put the prefix "default" here @FelixMalfait @martmull 
As for why we don't use roleTarget like other entities, I'm now fine with that but that means this will be handled a bit differently in the code base (it won't use the same cache and permission logic) so you need to be sure about that, not a strong blocker for me

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
```diff
@@ -54,54 +49,67 @@ export class JwtWrapperService {
...
-        throw new AuthException(
-          'Invalid token type',
-          AuthExceptionCode.INVALID_JWT_TOKEN_TYPE,
-        );
+      // This is due to an unfortunate mistake in the secret generation of API_KEY
+      // tokens. We used to sign with ACCESS Jwt Token Type instead of API_KEY.
+      // Now we need to check both cases not to break the existing api keys
+      // See this PR for context -> https://github.com/twentyhq/twenty/pull/16504
+      // This code block can be deleted, but all api keys created before
+      // 12/12/2025 will be broken
```
Weiko, line 97: @FelixMalfait wdyt?

Weiko (approved):
LGTM, left some comments also I agree with @FelixMalfait let's try to avoid patching if possible

PR #16508, opened 2025-12-11
prevent filtering out messages sent by user when their handle resembes group email

neo773:


packages/twenty-server/src/modules/messaging/message-import-manager/utils/filter-emails.util.ts
```diff
@@ -24,9 +28,29 @@ export const filterEmails = (
...
+      return true;
+    }
+
+    const isSenderGroupEmail = isGroupEmail(senderHandle);
+
+    if (!isSenderGroupEmail) {
+      return true;
+    }
+
+    const isSentByUser = isMessageFromUser(message, userHandles);
```
charlesBochet: I would do it the other way:

isMessageFromUser ? => if yes return true (we don't filter out)
!isGroupEmail ? => if yes return true

return false

charlesBochet: also isMessageFromUser => isMessageSenderMatchingHandles

packages/twenty-server/src/modules/messaging/message-import-manager/utils/is-message-from-user.util.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const isMessageFromUser = (
+  message: MessageWithParticipants,
+  userHandles: string[],
+): boolean => {
+  const fromParticipant = message.participants?.find(
+    (participant) => participant.role === MessageParticipantRole.FROM,
+  );
+
+  if (!fromParticipant?.handle) {
```
charlesBochet: isDefined

charlesBochet (approved):
LGTM with minor comments

charlesBochet (approved):

PR #16529, opened 2025-12-12
fix(server): non blocking CleanEmptyStringNullInTextFieldsCommand error

prastoin:
# Introduction
Getting query timeout on prod upgrade for workspace that has huge timeline activity table


etiennejouan (approved):
Thanks, sorry ! 🙏

PR #16530, opened 2025-12-12
Compute output schema on frontend

### LINKED ISSUE #1382: Refacto - Move step output schema logic at runtime
- Use existing states
- When populating the states, calculate the output either by using step settings (code step) either by using step types
- Remove existing endpoints

thomtrp:
Fixes https://github.com/twentyhq/core-team-issues/issues/1382

Current issue : all step output schemas are computed and stored on backend side. Which means that, when the database schema is updated - like a field creation - steps needs to be deleted an recreated. Which is invisible to users.

Solution : schema generation is moved on frontend side

1. Coming on the page the first time, the schema is populated for all steps except a few ones that are handled differently (Code, Webhook, http node, Agent)

2. A separated state allow to determine if a step needs a recomputation. 

3. The user only needs a refresh to see the whole schema re-computed

Follow-up:
- check if remaining backend steps could be moved to runtime computation. But Code will still require storage.
- Clean backend service that is not used anymore


packages/twenty-front/src/modules/workflow/workflow-variables/utils/generate/computeStepOutputSchema.ts
```diff
@@ -0,0 +1,221 @@
...
+export const computeStepOutputSchema = ({
+  step,
+  objectMetadataItems,
+}: {
+  step: WorkflowTrigger | WorkflowAction;
+  objectMetadataItems: ObjectMetadataItem[];
+}): OutputSchemaV2 | undefined => {
+  const stepType = step.type;
+
+  if (PERSISTED_OUTPUT_SCHEMA_TYPES.includes(stepType)) {
```
martmull, line 65: use utils `shouldComputeOutput...`

packages/twenty-front/src/modules/workflow/workflow-variables/types/RecordOutputSchemaV2.ts
```diff
@@ -25,6 +27,7 @@ export type FieldOutputSchemaV2 = RecordFieldLeaf | RecordFieldNode;
 
 export type RecordOutputSchemaV2 = {
   object: {
+    icon?: string;
     label: string;
     objectMetadataId: string;
     isRelationField?: boolean;
```
martmull, line 33: Would be great to factorize all those types in twenty-shared (in another PR obviously)

packages/twenty-front/src/modules/workflow/workflow-variables/utils/generate/generateFakeValue.ts
```diff
@@ -0,0 +1,92 @@
...
+    const elementType = valueType.replace('[]', '');
+
+    return Array.from({ length: 3 }, () => generateFakeValue(elementType));
+  } else if (valueType.startsWith('{') && valueType.endsWith('}')) {
+    const objData: Record<string, FakeValueTypes> = {};
+
+    const properties = valueType
+      .slice(1, -1)
+      .split(';')
+      .map((property) => property.trim())
```
martmull, line 34: We should move that to twenty-shared instead of copy paste

packages/twenty-front/src/modules/workflow/workflow-variables/utils/generate/generateFindRecordsOutputSchema.ts
```diff
@@ -0,0 +1,32 @@
+import { type ObjectMetadataItem } from '@/object-metadata/types/ObjectMetadataItem';
+import { type FindRecordsOutputSchema } from '@/workflow/workflow-variables/types/FindRecordsOutputSchema';
+import { generateRecordOutputSchema } from '@/workflow/workflow-variables/utils/generate/generateRecordOutputSchema';
+
+export const generateFindRecordsOutputSchema = (
```
martmull, line 5: same

packages/twenty-front/src/modules/workflow/workflow-variables/utils/generate/generateRecordEventOutputSchema.ts
```diff
@@ -0,0 +1,184 @@
...
+} from 'twenty-shared/types';
+import { capitalize } from 'twenty-shared/utils';
+import { DatabaseEventAction } from '~/generated/graphql';
+
+const camelToTitleCase = (camelCaseText: string): string =>
+  capitalize(
+    camelCaseText
+      .replace(/([A-Z])/g, ' $1')
+      .replace(/^./, (str) => str.toUpperCase()),
+  );
```
martmull, line 22: same go to twenty-shared

martmull (commented):
Hey nice changes. I see that some code has been duplicated I think it worth factorize directly using twenty-shared when possible

martmull (approved):

PR #16532, opened 2025-12-12
fix: ensure deactivated object records do not appear in search

mabdullahabaid:
Deactivating an object still allowed querying its records in search.

image

<br />
<br />

This PR introduces a minor improvement to not query records of deactivated objects.

image


etiennejouan (approved):
Hi @mabdullahabaid Thank you for your contribution ! 🙏  Looks good to me !

PR #16533, opened 2025-12-12
e2e tests

ijreilly:
In this PR, 
- current basic E2E tests are fixed, and some were added, covering some basic scenarios
- some tests avec been commented out, until we decide whether they are worth fixing 

The next steps are
- evaluate the flakiness of the tests. Once they've proved not to be flaky, we should add more tests + re-write the current ones not using aria-label (cf @lucasbordeau indication). 
- We will add them back to the development flow


lucasbordeau:
I'm concerned that aria-labels won't be maintained, much like comments, and that it risks bloating our codebase.

We could consider React locators : https://playwright.dev/docs/other-locators#react-locator

Devessier:
> I'm concerned that aria-labels won't be maintained, much like comments, and that it risks bloating our codebase.
> 
> We could consider React locators : https://playwright.dev/docs/other-locators#react-locator

I agree that we will likely never update the accessibility labels unless accessibility becomes a team concern. However, if we use React locators, the E2E tests will be dependent on implementation details.

We should think about the best compromise.

packages/twenty-e2e-testing/tests/authentication/signup_invite_email.spec.ts
```diff
@@ -49,12 +49,12 @@ test('Sign up with invite link via email', async ({
...
+    await expect(page.getByText('Account Deletion')).toBeVisible();
     await confirmationModal.typePlaceholderToInput();
+    await confirmationModal.clickConfirmButton();
 
     await Promise.all([
-      page.waitForURL('/welcome'),
-
-      confirmationModal.clickConfirmButton(),
+      page.waitForURL('**/welcome'),
     ]);
```
Devessier:
```suggestion
    await page.waitForURL('**/welcome');
```

packages/twenty-e2e-testing/tests/create-kanban-view.spec.ts
```diff
@@ -0,0 +1,59 @@
+import { expect, test } from '../lib/fixtures/screenshot';
+
+if (process.env.LINK) {
+  const baseURL = new URL(process.env.LINK).origin;
+  test.use({ baseURL });
+}
```
Devessier: I'm unsure we want to write that in every test.

packages/twenty-e2e-testing/tests/create-kanban-view.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+  for (const element of byIndustryElements) {
+    await expect(element).toBeVisible();
+  }
+  await page.getByText('Options').click();
+  await page.getByText('Group', { exact: true }).click();
+  await Promise.all([page.getByRole('button', { name: 'Hide group null', exact: true }).click(),
+    page.waitForRequest((req) => {
+    return req.url().includes('/metadata') &&
+           req.method() === 'POST'
+  })]);
```
Devessier, line 51: The problem with this kind of assertion is that we can make dozens of calls to `/metadata` in a single navigation. This assertion proves nothing.

We can leave it as is, but since it doesn’t prove that the action we expect was performed, it could lead to flakiness.

packages/twenty-front/src/modules/object-record/record-field-list/components/RecordFieldList.tsx
```diff
@@ -90,7 +90,7 @@ export const RecordFieldList = ({
         instanceId,
       }}
     >
-      <PropertyBox>
+      <PropertyBox ariaLabel="Record fields list">
```
Devessier: I don’t think this is a proper use of `aria-label`. `aria-label` is meant to provide a textual label for an element that would otherwise be unclear or meaningless to someone using a screen reader, such as a blind person.

For example, you might use `aria-label` to explain the meaning of an icon. An icon alone doesn’t convey anything to a blind user, so you must add additional information so that everyone can understand the page.

If you set an `aria-label` on a container that already contains text, the `aria-label` will override all the natural text inside it for screen readers. **It is really harmful.**

In your case, it seems you want to add an `id` to an element so you can locate it in end-to-end tests. Instead, you should use the `data-testid` attribute. Playwright provides methods to select elements by their `data-testid` attribute.

See: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label.

packages/twenty-front/src/modules/object-record/record-field-list/record-detail-section/components/RecordDetailSectionContainer.tsx
```diff
@@ -51,6 +52,7 @@ type RecordDetailSectionContainerProps = {
   rightAdornment?: React.ReactNode;
   hideRightAdornmentOnMouseLeave?: boolean;
   areRecordsAvailable?: boolean;
+  ariaLabel?: string;
```
Devessier: I think this should be `data-testid` instead.

packages/twenty-front/src/modules/object-record/record-field-list/record-detail-section/relation/components/RecordDetailRelationSection.tsx
```diff
@@ -185,6 +185,7 @@ export const RecordDetailRelationSection = ({
       }}
     >
       <RecordDetailSectionContainer
+        ariaLabel={`${fieldDefinition.label} (relation)`}
```
Devessier: I think this should be `data-testid` instead.

packages/twenty-front/src/modules/object-record/record-group/components/RecordGroupMenuItemDraggable.tsx
```diff
@@ -36,6 +36,9 @@ export const RecordGroupMenuItemDraggable = ({
     const iconButtons = [
       {
         Icon: recordGroup.isVisible ? IconEyeOff : IconEye,
+        ariaLabel: recordGroup.isVisible
+          ? `Hide group ${recordGroup.value}`
+          : `Show group ${recordGroup.value}`,
```
Devessier, line 41: That's a proper use of `aria-label` 👍

packages/twenty-front/src/modules/object-record/record-inline-cell/property-box/components/PropertyBox.tsx
```diff
@@ -3,6 +3,7 @@ import styled from '@emotion/styled';
 interface PropertyBoxProps {
   children: React.ReactNode;
   className?: string;
+  ariaLabel?: string;
```
Devessier: I think this should be `data-testid` instead.

Devessier (changes requested):
Thank you for working on bringing back meaningful end-to-end tests!

One main concern: I think we should use `data-testid` instead of `aria-label`.

If possible, it would be great to remove the use of `process.env.LINK`.

Devessier (approved):

packages/twenty-e2e-testing/tests/create-kanban-view.spec.ts
ijreilly: I don't think we expect other POST requests in this setup, nor that for now we actually do any /metadata post requests while navigating only

ijreilly: But I intent to make sure this is not flaky so i m keeping that in mind

packages/twenty-front/src/modules/object-record/record-field-list/components/RecordFieldList.tsx
ijreilly: Thank you, I replaced this with data-testid!

Issue #15916, opened 2025-11-19
Destroy Many is much longer than before


StephanieJoly4:
## Bug Description

I've tried many times since 2 days to batch destroy records. It takes a really long time compared to before. 
I've checked with @ijreilly and I don't think we can assume this is the db being stressed at a given time, since this happens every time I try since 3 days. 

It took about 4 to 5 minutes to delete 907 records and queries take 5sec or more 

<img width="871" height="597" alt="Image" src="https://github.com/user-attachments/assets/f4374c6e-ab66-4bb3-9edc-4cc38721672e" />

<img width="876" height="202" alt="Image" src="https://github.com/user-attachments/assets/a7cf4090-ca52-40d0-bd02-87e086e739af" />

Did we change anything ? I don't have the times in mind but feel it was much faster before.
thank you!

FelixMalfait:
@abdulrahmancodes could you please try to replicate and diagnose/fix this? Thanks!

etiennejouan:
@StephanieJoly4 closing the issue.
The DestroyMany works on its own. 
It collides with other tasks that are paralyzing the server. I'm investigating these costly tasks for optim.

PR #16535, opened 2025-12-12
Investigate - Slow transformRichTextV2 function

etiennejouan:
Investigate CPU overloading ...

related to https://github.com/twentyhq/twenty/issues/15916


charlesBochet (commented):
const t0 = performance.now();
doSomething();
const t1 = performance.now();
console.log(`Call to doSomething took ${t1 - t0} milliseconds.`);

charlesBochet (approved):

PR #16540, opened 2025-12-12
Fix APIKey in search resolver for get role

Weiko:
Fixes https://github.com/twentyhq/twenty/issues/16534

Typing was wrong, apiKey from request object is ApiKeyEntity and not a string which was then failing when using 
```typescript
const roleId = apiKeyRoleMap[apiKeyId];

    if (!isDefined(roleId)) {
      throw new ApiKeyException(
        `API key ${apiKeyId} has no role assigned`,
        ApiKeyExceptionCode.API_KEY_NO_ROLE_ASSIGNED,
      );
    }
```

error
```
API key [object Object] has no role assigned"
```

## Before
Screenshot 2025-12-12 at 17 37 06

## After
Screenshot 2025-12-12 at 17 36 44


etiennejouan (approved):
LGTM 👍

PR #16541, opened 2025-12-12
Fix APIKey typing

Weiko:
Following this https://github.com/twentyhq/twenty/pull/16540
Those were not failing (yet) but were wrongly typed and error prone


packages/twenty-server/src/engine/api/mcp/controllers/mcp-core.controller.ts
```diff
@@ -36,7 +37,7 @@ export class McpCoreController {
   async handleMcpCore(
     @Body() body: JsonRpc,
     @AuthWorkspace() workspace: WorkspaceEntity,
-    @AuthApiKey() apiKey: string | undefined,
+    @AuthApiKey() apiKey: ApiKeyEntity | undefined,
```
guillim, line 40: this is going to break any current MCP users right ? are we ok with that ?

packages/twenty-server/src/engine/core-modules/billing/billing.resolver.ts
```diff
@@ -87,12 +88,12 @@ export class BillingResolver {
       plan,
       requirePaymentMethod,
     }: BillingCheckoutSessionInput,
-    @AuthApiKey() apiKey?: string,
```
guillim, line 90: Same subject here, just worried if we have webhook calls defined in stripe that we need to adjust. (i am not familiar with this)

packages/twenty-server/src/engine/core-modules/workspace/workspace.resolver.ts
```diff
@@ -136,7 +137,7 @@ export class WorkspaceResolver {
     @Args('data') data: UpdateWorkspaceInput,
     @AuthWorkspace() workspace: WorkspaceEntity,
     @AuthUserWorkspaceId() userWorkspaceId: string,
-    @AuthApiKey() apiKey?: string,
```
guillim, line 139: same concern about external calls

guillim (commented):
LGTM, expect a need for synchronisation with our providers where we may have defined webhooks that have API key

packages/twenty-server/src/engine/api/mcp/controllers/mcp-core.controller.ts
Weiko: This is only a typing change, this won't break the actual logic behind it. Also, request object already types it correctly but this was not typed properly in controllers/resolvers.
The typing was wrong originally already, code expected things to be strings but they were objects. They were rarely accessed as values but simply checked if they were defined but I'd prefer to correct the typing right now to avoid any issue in the future. We had a real bug that I've fixed in this PR if you want an example https://github.com/twentyhq/twenty/pull/16540

guillim: My bad, I did not read 16540. Then it's 100% logic

guillim (approved):

PR #16543, opened 2025-12-12
WaffleChart implementation based on PieChart code

seljudas:
### Summary

This PR implements the nivo/ResponsiveWaffle Waffle Chart widget to the Twenty dashboard system.

I've used the implementation of the PieChart widget as a template.

Changes
• 	Added WaffleChart component under twenty-front/src/modules/page-layout/widgets/graph/graphWidgetWaffleChart..
• 	Updated schema definitions to include WAFFLE_CHART type.
• 	Validated TypeScript handler signatures against operational lineage.
• 	Added unit tests for schema relationships and DOM event propagation.

Validation
• 	Verified backend schema enums and database migration for WAFFLE.
• 	Confirmed operational lineage from widget config → runtime event firing.

Notes
Tooltip and right click does not work and I do not know why.
Verified that the widget can be saved the database.
Edit menu not quite consistent with what I expect, not sure why.
CenterMetric code similar to PieChart exists, but not in use - if included displays for instance total items on top of chart instead of in the middle. I assume that you would want a different name for this.


bosiraphael (commented):
Hello @seljudas! Thank you very much for this cool contribution :) Good work overall.
One remark is that a lot of code is duplicated for the new waffle chart, but essentially since it behaves the same as the pie chart, we could factorize a lot of this code between the two implementations. Same as what we did for line and bar chart which shares common utils and config.
I know this PR is a work in progress, but basically what is left to do to make it mergeable:
- Factorize the code
- Update the settings (`Show value in center`: should become `Show value below`, and we should show the total aggregate value below)
- Implement the data labels
- Update the each slice represents setting icon, which could be a square
- Make date granularity settings work (it throws an error for now)
- Make tooltip work @ehconitin can help you with this

charlesBochet:
@seljudas Thanks a lot for opening a PR, I'm going to close it as we have no news but feel free to address the feedbacks and re-open it once ready, the team will keep reviewing it :)

PR #16544, opened 2025-12-12
Refactored Date to Temporal in critical date zones

### LINKED ISSUE #16110: in dashboards graphbar, when set "creationdate" as x axis, get date minus 1 (day or month)
## Bug Description

in dashaboards graph bar, when you set "creation date" as X axis, shows always the granularity minus 1, if granularity day then shows as yesterday records from today, if granularity is month the shows as lastmonth records from this month.

lucasbordeau:
Fixes https://github.com/twentyhq/twenty/issues/16110

This PR implements Temporal to replace the legacy Date object, in all features that are time zone sensitive. (around 80% of the app)

Here we define a few utils to handle Temporal primitives and obtain an easier DX for timezone manipulation, front end and back end.

This PR deactivates the usage of timezone from the graph configuration, because for now it's always UTC and is not really relevant, let's handle that later.

Workflows code and backend only code that don't take user input are using UTC time zone, the affected utils have not been refactored yet because this PR is big enough.

# New way of filtering on date intervals

As we'll progressively rollup Temporal everywhere in the codebase and remove `Date` JS object everywhere possible, we'll use the way to filter that is recommended by Temporal.

This way of filtering on date intervals involves half-open intervals, and is the preferred way to avoid edge-cases with DST and smallest time increment edge-case.

## Filtering endOfX with DST edge-cases

Some day-light save time shifts involve having no existing hour, or even day on certain days, for example Samoa Islands have no 30th of December 2011 : https://www.timeanddate.com/news/time/samoa-dateline.html, it jumps from 29th to 31st, so filtering on `< next period start` makes it easier to let the date library handle the strict inferior comparison, than filtering on `≤ end of period` and trying to compute manually the end of the period.

For example for Samoa Islands, is end of day `2011-12-29T23:59:59.999` or is it `2011-12-30T23:59:59.999` ? If you say I don't need to know and compute it, because I want everything strictly before `2011-12-29T00:00:00 + start of next day (according to the library which knows those edge-cases)`, then you have a 100% deterministic way of computing date intervals in any timezone, for any day of any year.

Of course the Samoa example is an extreme one, but more common ones involve DST shifts of 1 hour, which are still problematic on certain days of the year.

## Computing the exact _end of period_

Having an open interval filtering, with `[included - included]` instead of half-open `[included - excluded)`, forces to compute the open end of an interval, which often involves taking an arbitrary unit like minute, second, microsecond or nanosecond, which will lead to edge-case of unhandled values.

For example, let's say my code computes endOfDay by setting the time to `23:59:59.999`, if another library, API, or anything else, ends up giving me a date-time with another time precision `23:59:59.999999999` (down to the nanosecond), then this date-time will be filtered out, while it should not.

The good deterministic way to avoid 100% of those complex bugs is to create a half-open filter : 

`≥ start of period` to `< start of next period` 

For example : 

`≥ 2025-01-01T00:00:00` to `< 2025-01-02T00:00:00` instead of  `≥ 2025-01-01T00:00:00` to `≤ 2025-01-01T23:59:59.999` 

Because, `2025-01-01T00:00:00` = `2025-01-01T00:00:00.000` = `2025-01-01T00:00:00.000000` = `2025-01-01T00:00:00.000000000` => no risk of error in computing start of period

But `2025-01-01T23:59:59` ≠ `2025-01-01T23:59:59.999` ≠ `2025-01-01T23:59:59.999999` ≠ `2025-01-01T23:59:59.999999999` => existing risk of error in computing end of period

This is why an half-open interval has no risk of error in computing a date-time interval filter.

Here is a link to this debate : https://github.com/tc39/proposal-temporal/issues/2568

> For this reason, we recommend not calculating the exact nanosecond at the end of the day if it's not absolutely necessary. For example, if it's needed for <= comparisons, we recommend just changing the comparison code. So instead of <= zdtEndOfDay your code could be < zdtStartOfNextDay which is easier to calculate and not subject to the issue of not knowing which unit is the right one. 
>
> [Justin Grant](https://github.com/justingrant), top contributor of Temporal

## Application to our codebase

Applying this half-open filtering paradigm to our codebase means we would have to rename `IS_AFTER` to `IS_AFTER_OR_EQUAL` and to keep `IS_BEFORE` (or even `IS_STRICTLY_BEFORE`) to make this half-open interval self-explanatory everywhere in the codebase, this will avoid any confusion. 

See the relevant issue : https://github.com/twentyhq/core-team-issues/issues/2010

In the mean time, we'll keep this operand and add this semantic in the naming everywhere possible.

## Example with a different user timezone 

Example on a graph grouped by week in timezone Pacific/Samoa, on a computer running on Europe/Paris : 

image

Then the associated data in the table view, with our **half-open date-time filter** : 

image

And the associated SQL query result to see how DATE_TRUNC in Postgres applies its internal start of week logic : 

image

The associated SQL query without parameters to test in your SQL client : 

```SQL
SELECT "opportunity"."closeDate" as "close_date", TO_CHAR(DATE_TRUNC('week', "opportunity"."closeDate", 'Pacific/Samoa') AT TIME ZONE 'Pacific/Samoa', 'YYYY-MM-DD') AS "DATE_TRUNC by week start in timezone Pacific/Samoa", "opportunity"."name" FROM "workspace_1wgvd1injqtife6y4rvfbu3h5"."opportunity" "opportunity" ORDER BY "opportunity"."closeDate" ASC NULLS LAST
```

# Date picker simplification (not in this PR)

Our DatePicker component, which is wrapping `react-datepicker` library component, is now exposing plain dates as string instead of Date object. The Date object is still used internally to manage the library component, but since the date picker calendar is only manipulating plain dates, there is no need to add timezone management to it, and no need to expose a handleChange with Date object. 

The timezone management relies on date time inputs now.

The modification has been made in a previous PR : https://github.com/twentyhq/twenty/issues/15377 but it's good to reference it here.

# Calendar feature refactor

Calendar feature has been refactored to rely on Temporal.PlainDate as much as possible, while leaving some date-fns utils to avoid re-coding them. 

Since the trick is to use utils to convert back and from Date object in exec env reliably, we can do it everywhere we need to interface legacy Date object utils and Temporal related code.

## TimeZone is now shown on Calendar :

image

## Month picker has been refactored 

image

Since the days weren't useful, the picker has been refactored to remove the days.

# Miscellaneous 
- Fixed a bug with drag and drop edge-case with 2 items in a list.

# Improvements 

## Lots of chained operations
It would be nice to create small utils to avoid repeated chained operations, but that is how Temporal is designed, a very small set of primitive operations that allow to compose everything needed. Maybe we'll have wrappers on top of Temporal in the coming years.

## Creation of Temporal objects is throwing errors

If the input is badly formatted Temporal will throw, we might want to adopt a global strategy to avoid that.

Example : 

```ts
const newPlainDate = Temporal.PlainDate.from('bad-string'); // Will throw
```


bosiraphael:
@cubic-dev-ai review this PR

packages/twenty-front/src/modules/object-record/advanced-filter/command-menu/components/AdvancedFilterCommandMenuValueFormInput.tsx
```diff
@@ -179,6 +182,9 @@ export const AdvancedFilterCommandMenuValueFormInput = ({
     metadata: fieldDefinition?.metadata as FieldMetadata,
   };
 
+  const shouldUseUTCTimeZone = isWorkflowFindRecords === true;
+  const timeZone = shouldUseUTCTimeZone ? 'UTC' : undefined;
```
bosiraphael: I'm wondering if we could put 'UTC' in a constant to avoid typos when someone wants to use this timezone

packages/twenty-front/src/modules/ui/input/components/internal/date/components/DatePickerWithoutCalendar.tsx
```diff
@@ -0,0 +1,457 @@
...
+        <Suspense
+          fallback={
+            <StyledDatePickerFallback>
+              <SkeletonTheme
+                baseColor={theme.background.tertiary}
+                highlightColor={theme.background.transparent.lighter}
+                borderRadius={2}
+              >
+                <Skeleton
+                  width={DATE_PICKER_CONTAINER_WIDTH - 16}
```
bosiraphael: We could create a constant for the number 16

packages/twenty-shared/src/utils/filter/dates/utils/resolveRelativeDateFilter.ts
```diff
@@ -1,53 +1,86 @@
...
   relativeDateFilter: RelativeDateFilter,
+  referenceTodayZonedDateTime: Temporal.ZonedDateTime,
 ) => {
   const { direction, amount, unit, firstDayOfTheWeek } = relativeDateFilter;
 
-  const referenceDate = new TZDate();
-
   switch (direction) {
-    case 'NEXT':
+    case 'NEXT': {
```
bosiraphael, line 17: I know this wasn't introduced in your PR, but we tend to use enums for this in the codebase rather than string literals

packages/twenty-shared/src/utils/filter/turnRecordFilterIntoGqlOperationFilter.ts
```diff
@@ -174,20 +168,53 @@ export const turnRecordFilterIntoRecordGqlOperationFilter = ({
           );
       }
     case 'DATE': {
-      const resolvedFilterValue = resolveDateFilter(recordFilter);
```
bosiraphael, line 177: Not your PR, but this file could benefit some refactoring later to use sub utils to make it more readable

bosiraphael (commented):
Great job, just left a couple of nitpicks

packages/twenty-shared/src/utils/filter/turnRecordFilterIntoGqlOperationFilter.ts
lucasbordeau: Duly noted, this would indeed cause a very tedious QA due to its central role in the codebase

packages/twenty-shared/src/utils/filter/dates/utils/resolveRelativeDateFilter.ts
lucasbordeau: Let's clarify this after this PR because it would cause a lot of changes in twenty-front for something that is not essential.

bosiraphael (approved):

PR #16558, opened 2025-12-15
Fix cursor-based pagination

### LINKED ISSUE #6344: Order by currency bug with cursor
## Bug Description

When we have a view with an order by on a currency field type, if we go to the show page and try to navigate with the buttons, if we have an empty amount on the previous record, the request says there's nothing before, while there is actually a record.

Example (the previous arrow is deactivated) : 

image

## Expected behavior

The backend should return the previous record, even if the currency field is empty.

abdulrahmancodes:
Closes #6344 

### Issues Fixed

1. **Cursor encoding was broken:** The frontend only requested the `id` field in `recordGqlFields`, but cursor encoding requires all orderBy field values. 

2. **Cursor encoding was broken**. The frontend only requested the `id` field in `recordGqlFields` for pagination queries, but cursor encoding requires all orderBy field values to generate valid cursors.
3. **Navigation buttons were not disabled when navigation was unavailable**. The next/previous record navigation buttons remained enabled even when `recordAfter`/`recordBefore` was not available, leading to a poor user experience where clicking the buttons would have no effect.

### Changes

- **Updated `buildColumnsToSelect`** to automatically include orderBy fields in the SELECT statement:
  - Extract column names from orderBy (including composite field sub-columns)
  - Add these columns to the select clause so cursor encoding has access to all required fields, even if not explicitly requested by the frontend

- **Updated `buildCursorCompositeFieldWhereCondition`** to properly handle null composite field values:
  - Use `is: 'NULL'` instead of `eq: null` for equality conditions
  - Use `is: 'NOT NULL'` or `is: 'NULL'` for comparison conditions when all sub-fields are null
  - Include NULL values in comparison filters based on sort order (`NullsFirst`/`NullsLast`) and pagination direction
  - Filter out null values from cursor entries to avoid invalid SQL comparisons
  
- **Added disabled state support to Action component system**:
  - Updated `Action`, `ActionDisplay`, `ActionButton`, `ActionListItem`, and `ActionDropdownItem` to accept and handle `disabled` prop
  - Updated `NavigateToNextRecordSingleRecordAction` and `NavigateToPreviousRecordSingleRecordAction` to use `canNavigateToNextRecord` and `canNavigateToPreviousRecord` flags to disable buttons when navigation is unavailable

### Known Limitations

**Navigation with Null Composite Field Values:**

Navigation backward when records have null composite field values (e.g., amount) may skip records in certain scenarios. This occurs when:
- Records are sorted by a composite field with `AscNullsLast` ordering
- Navigating backward from a record with a value to a record with a null value
- The filter logic for handling nulls in cursor-based pagination needs further refinement

Multiple approaches were attempted to fix this issue (including explicit `IS NOT NULL` checks, restructuring OR conditions, and separate null handling), but none fully resolved the skipping behavior without introducing other issues. This is a known limitation that requires further investigation. The current implementation handles most cases but may skip intermediate records in specific scenarios involving null composite field values.


lucasbordeau:
Hi @abdulrahmancodes sorry for the delay.

My main concern while reviewing this PR is that the filter built from the cursor looks odd.

For example, if I navigate from a simple view to a show page, I should have indeed a request to find the next and previous record id, by using cursor after and before with limit 1, respectively. 

But if I look at the generated filters for this I end up with something like that : 

```JSON
[
  {
    "position": {
      "lt": 22
    }
  },
  {
    "and": [
      {
        "position": {
          "eq": 22
        }
      },
      {
        "id": {
          "lt": "50505050-0016-4e7c-8001-123456789abc"
        }
      }
    ]
  }
]
```

The problem is that a `eq` operator on the position field inside an `and` makes the other filters in this `and` useless. Moreover in this precise example `"50505050-0016-4e7c-8001-123456789abc"` is the record that has `position = 22` so the second filter cannot return any data from an SQL request.

It works because the first filter is ok, but the second being an impossible filter, it will create technical debt and lead to bugs.

I don't know if this comes from your PR originally but since we're working on this part of the code it is the best PR to fix this problem.

lucasbordeau (changes requested):
See my comment

lucasbordeau:
@abdulrahmancodes I'm closing this PR because there are too many problems with the generated queries.

Here is an example : 

Test set : 

image

What I get with the SQL query for previous button (with a LIMIT set to 4 to see more records) : 

image


And here is the generated query for the previous button : 

```SQL
SELECT
  "person"."id" AS "person_id",
  "person"."nameFirstName" AS "person_nameFirstName",
  "person"."nameLastName" AS "person_nameLastName",
  "person"."position" AS "person_position"
FROM
  "workspace_1wgvd1injqtife6y4rvfbu3h5"."person" "person"
WHERE
  (
    (
      (
        (
          "person"."id" != '20202020-be4b-4b3f-826b-2d91511a0631'
        )
        AND (
          (
            (
              (
                (
                  (
                    (
                      "person"."nameFirstName" < 'Aaron'
                      AND "person"."nameFirstName" < 'Aaron'
                    )
                    OR (
                      "person"."nameFirstName" IS NULL
                      OR "person"."nameFirstName" = ''
                      AND "person"."nameFirstName" IS NULL
                      OR "person"."nameFirstName" = ''
                    )
                  )
                )
                OR (
                  (
                    (
                      "person"."nameFirstName" = 'Aaron'
                      AND "person"."nameFirstName" = 'Aaron'
                    )
                    AND (
                      (
                        (
                          "person"."nameLastName" < 'Griffin'
                          AND "person"."nameLastName" < 'Griffin'
                        )
                        OR (
                          "person"."nameLastName" IS NULL
                          OR "person"."nameLastName" = ''
                          AND "person"."nameLastName" IS NULL
                          OR "person"."nameLastName" = ''
                        )
                      )
                    )
                  )
                )
              )
            )
            OR (
              (
                (
                  "person"."nameFirstName" = 'Aaron'
                  AND "person"."nameFirstName" = 'Aaron'
                  AND "person"."nameLastName" = 'Griffin'
                )
                AND ("person"."position" < 112)
              )
            )
            OR (
              (
                (
                  "person"."nameFirstName" = 'Aaron'
                  AND "person"."nameFirstName" = 'Aaron'
                  AND "person"."nameLastName" = 'Griffin'
                )
                AND ("person"."position" = 112)
                AND (
                  "person"."id" < '20202020-be4b-4b3f-826b-2d91511a0631'
                )
              )
            )
          )
        )
      )
    )
  )
  AND ("person"."deletedAt" IS NULL)
ORDER BY
  "person_nameFirstName" DESC NULLS FIRST,
  "person_nameLastName" DESC NULLS FIRST,
  "person_position" DESC NULLS FIRST,
  "person_id" DESC NULLS FIRST
LIMIT
  2
```

There are already two problems : 
- Duplicated conditions : `("person"."nameFirstName" < 'Aaron' AND "person"."nameFirstName" < 'Aaron')`
- `DESC` instead of `ASC` in the order by

Let's discuss this with the core team and the recent change made to order by relation nested field also. 

I'm pasting this here for reference, let's open this again when we'll have a clearer vision.

PR #16561, opened 2025-12-15
fix(workflow): line break not supported by Send Email Nodes

abk404:
Closes #16557

Tiptap Editor (which the Send Email Node uses) , creates a content json with type 'hardBreak' for line breaks. 

The was no rederer defined for this `hardBreak` node type,  so the `renderNode` function was ignoring that node (returning null).

**Fix :** Added a renderer for  `hardBreak` node type.


abk404:
Added test as coverage dropped below threshold.

etiennejouan (approved):
Hi @abk404 , Thank you for your contribution. It works fine 👏

PR #16592, opened 2025-12-16
Vendor twenty-shared into twenty-sdk

martmull:
twenty-shared is not bundled properly when deploying twenty-sdk to npm. This aims to fix that issue

### DIFF
--- a/packages/twenty-sdk/vite.config.ts
+++ b/packages/twenty-sdk/vite.config.ts
@@ -1,3 +1,4 @@
+import * as fs from 'fs-extra';
 import path from 'path';
 import { defineConfig } from 'vite';
 import dts from 'vite-plugin-dts';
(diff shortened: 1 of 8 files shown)


packages/twenty-sdk/vite.config.ts
```diff
@@ -44,7 +60,26 @@ export default defineConfig(() => {
...
+          return {
+            filePath,
+            content: content.replace(
+              /(from\s+["'])twenty-shared(\/[^"']*)?(["'])/g,
+              `$1${rel}/twenty-shared$2$3`,
+            ),
+          };
+        },
       }),
+      copySharedDist(),
```
prastoin: Shouldn't we run the copy plugin before the dts ?

packages/twenty-sdk/vite.config.ts
```diff
@@ -44,7 +60,26 @@ export default defineConfig(() => {
...
+          let rel = path
+            .relative(fromDir, vendorDir)
+            .split(path.sep)
+            .join(path.posix.sep);
+          if (!rel.startsWith('.')) rel = `./${rel}`;
+
+          return {
+            filePath,
+            content: content.replace(
+              /(from\s+["'])twenty-shared(\/[^"']*)?(["'])/g,
```
prastoin, line 77: Everything here is quite complex, I'm wondering if we couldn't handle that through tsconfig path alias instead 🤔 so we don't have to handle relative path mapping ourselves

prastoin (commented):
Hey there, as said I'm not sure about the most idiomatic solution here
I think we might be able to find native implementation or least rely on some of their basis:
- We could try to hack through the typescript custom paths alias in order to handle relative path handler resolver
- I'm not sure too but I think we can either configure vite or rollup through vite lib config to generate a dedicated [manual chunk](https://rollupjs.org/configuration-options/#output-manualchunks) for the twenty shared. Also might be quite trickier to setup as we're hoisting dependencies in root node_modules

packages/twenty-sdk/vite.config.ts
martmull: I tried but did not succeed

prastoin (approved):
As discussed lets start from there and improve the solution to be more idiomatic to used tools afterwards

PR #16593, opened 2025-12-16
fix: hide GraphQL stack traces and messages in production

mabdullahabaid:
Resolves [Code Scanning Alert 180](https://github.com/twentyhq/twenty/security/code-scanning/180).

- Normalize unexpected GraphQL errors in convertExceptionToGraphql to a generic "Internal Server Error" instead of exposing exception.name directly to clients.
- Only attach stack and response (original error message) in development, so production responses don’t leak internal class names, implementation details, or stack traces, while observability is preserved via `ExceptionHandlerService`/Sentry.
- Keep behavior consistent with `convertHttpExceptionToGraphql`, which also only exposes detailed response and stack information when `NODE_ENV === DEVELOPMENT`.


mabdullahabaid:
Don't really see why it's being reported as a breaking change - error.message is still defined, just not set to exception.name to hide internal details.

For stack trace and response, we do not expose to the client (and it's not consuming them either).

guillim (approved):

guillim (approved):
LGTM

PR #16604, opened 2025-12-16
Fix records cache invalidation on field metadata update

etiennejouan:
Fixes https://github.com/twentyhq/twenty/issues/16388


etiennejouan:
Cache invalidation is not the solution as data on table view is lazy loaded, then cache invalidation or refetching do not trigger table view rendering.

[Other solution discussed on parent issue](https://github.com/twentyhq/twenty/issues/16388#issuecomment-3664752756)

PR #16612, opened 2025-12-17
feat: add resizable navigation drawer and command menu panels

FelixMalfait:
## Summary

Adds Notion-style resizable panels for the navigation drawer (left sidebar) and command menu (right panel).

## Behavior

- **Hover** at panel edge → resize cursor appears
- **Click** → collapse/close the panel
- **Drag** → resize the panel (5px movement threshold to distinguish from click)

## Constraints

| Panel | Min | Max | Default | Collapse Threshold |
|-------|-----|-----|---------|-------------------|
| Navigation Drawer | 180px | 350px | 220px | 150px |
| Command Menu | 320px | 600px | 400px | 250px |

## Performance Optimizations

- **CSS variables** for smooth 60fps resize (no React re-renders during drag)
- **Table resize observer disabled** during panel resize to prevent expensive recalculations
- **React.memo wrapper** on page body to prevent unnecessary re-renders

## Architecture

- `useResizablePanel` hook following the same pattern as `useResizeTableHeader`
- `ResizablePanelEdge` - resize handle positioned at panel edge
- `ResizablePanelGap` - resize handle in the gap between panels
- `cssVariableEffect` - Recoil effect to sync CSS variables with state

## Refactoring

- Split `recoil-effects.ts` into separate files in `utils/recoil/` (one export per file)
- Persist panel widths to localStorage via existing `localStorageEffect`


packages/twenty-front/src/modules/object-record/components/CommandMenuPageLayout.tsx
```diff
@@ -62,18 +72,28 @@ const StyledModalContainer = styled.div`
   z-index: 1;
 `;
 
+const GAP_WIDTH = 8;
+
+// Memoized wrapper to prevent re-renders of children during resize
+const MemoizedPageBody = memo(({ children }: { children: ReactNode }) => (
```
FelixMalfait: Let's remove that

packages/twenty-front/src/utils/recoil/cssVariableEffect.ts
```diff
@@ -0,0 +1,22 @@
+import { type AtomEffect } from 'recoil';
+
+export const cssVariableEffect =
```
lucasbordeau: Let's use an EffectComponent like `RecordTableColumnWidthEffect` to avoid creating another pattern for the same type of logic. The less patterns we have in the code base the better.

packages/twenty-front/src/modules/ui/layout/resizable-panel/types/ResizablePanelTypes.ts
```diff
@@ -0,0 +1,7 @@
+export type ResizablePanelSide = 'left' | 'right';
```
lucasbordeau: We prefer having one type per file and not creating a -Types barrel file.

packages/twenty-front/src/modules/ui/layout/resizable-panel/hooks/useResizablePanel.ts
```diff
@@ -0,0 +1,147 @@
...
+  onWidthChange: (width: number) => void;
+  onCollapse: () => void;
+  cssVariableName?: string;
+  onResizeStart?: () => void;
+};
+
+const clampWidth = (width: number, min: number, max: number): number =>
+  Math.min(max, Math.max(min, width));
+
+export const useResizablePanel = ({
```
lucasbordeau, line 23: That's a pattern that is not adapted to a code base that use atomic state management, we avoid doing that and instead prefer setting recoil states synchronously from event handlers, and then components should listen to those states.

packages/twenty-front/src/modules/ui/layout/resizable-panel/components/ResizablePanelGap.tsx
```diff
@@ -0,0 +1,57 @@
...
+  constraints: ResizablePanelConstraints;
+  currentWidth: number;
+  onWidthChange: (width: number) => void;
+  onCollapse: () => void;
+  gapWidth: number;
+  cssVariableName?: string;
+  onResizeStart?: () => void;
+};
+
+export const ResizablePanelGap = ({
```
lucasbordeau, line 26: Isn't it the same logic as ResizablePanelEdge ?

lucasbordeau:
There is a small effect with a width that is incorrect for a few seconds that makes the right drawer look strange while resizing : 

https://github.com/user-attachments/assets/2b23f7ee-dcc8-4c6f-aad1-76059dd8597d

packages/twenty-front/src/modules/command-menu/components/CommandMenuWidthEffect.tsx
```diff
@@ -0,0 +1,20 @@
+import { useEffect } from 'react';
+import { useRecoilValue } from 'recoil';
+
+import {
+  COMMAND_MENU_WIDTH_VAR,
+  commandMenuWidthState,
+} from '../states/commandMenuWidthState';
```
charlesBochet, line 7: we don't use relative imports usually (I thought there was a lint rule for that at some point!)

packages/twenty-front/src/modules/ui/layout/resizable-panel/components/ResizablePanelGap.tsx
```diff
@@ -0,0 +1,55 @@
+import styled from '@emotion/styled';
+
+import { useResizablePanel } from '../hooks/useResizablePanel';
```
charlesBochet, line 3: same everywhere

packages/twenty-front/src/modules/ui/navigation/navigation-drawer/components/NavigationDrawer.tsx
```diff
@@ -69,40 +93,62 @@ export const NavigationDrawer = ({
     setIsHovered(false);
   };
 
-  const desktopWidth = isNavigationDrawerExpanded
-    ? NAV_DRAWER_WIDTHS.menu.desktop.expanded
-    : NAV_DRAWER_WIDTHS.menu.desktop.collapsed;
+  const handleCollapse = useCallback(() => {
```
charlesBochet, line 96: all these useCallback are not necessary IMO

packages/twenty-front/src/utils/recoil/localStorageEffect.ts
```diff
@@ -0,0 +1,21 @@
+import { type AtomEffect } from 'recoil';
+
+export const localStorageEffect =
+  <T>(key?: string): AtomEffect<T> =>
+  ({ setSelf, onSet, node }) => {
+    const savedValue = localStorage.getItem(key ?? node.key);
+    if (savedValue != null) {
```
charlesBochet, line 7: use guards isDefined

charlesBochet (commented):
LGTM on the code, left minors comments

PR #16582, opened 2025-12-16
Feature 15797 add secondary action button

Lakshayyy-m:
This is a fix for #15797 

This pull request is to replace PR #16307 and to extend #16265 

Just to repeat, this PR does the following -> 
**Table Cell Button and Edit Button Improvements**
- Enhanced RecordTableCellButton to support a secondary action and icon, enabling both the primary and secondary actions based on the selected action mode.
- Updated RecordTableCellEditButton to determine the action mode for actionable fields, and provide both copy and navigate actions as primary/secondary buttons, with appropriate feedback. 

## When primary function is to copy
image

## When primary function is to open link
image

Hey @etiennejouan, please have a look!
Thank you

PR #16613, opened 2025-12-17
Second Action Button - fix

etiennejouan:
Merged https://github.com/twentyhq/twenty/pull/16582 without testing case when fieldDefinition.metadata.settings?.clickAction is not set (default value)

To test : 
When 
Screenshot 2025-12-17 at 09 37 26
is set (or default)
In table view you can copy cell content
Screenshot 2025-12-17 at 09 36 58

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-table/record-table-cell/hooks/useGetSecondaryRecordTableCellButton.ts
+++ b/packages/twenty-front/src/modules/object-record/record-table/record-table-cell/hooks/useGetSecondaryRecordTableCellButton.ts
@@ -32,9 +32,12 @@ export const useGetSecondaryRecordTableCellButton = () => {
     return [];
   }
 
+  const mainActionOnClick =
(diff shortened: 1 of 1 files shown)


packages/twenty-front/src/modules/object-record/record-table/record-table-cell/hooks/useGetSecondaryRecordTableCellButton.ts
```diff
@@ -32,9 +32,12 @@ export const useGetSecondaryRecordTableCellButton = () => {
     return [];
   }
 
+  const mainActionOnClick =
+    fieldDefinition.metadata.settings?.clickAction ||
```
ijreilly: It should give the same output in that precise case where clickAction values are restricted to the enum, but I think `??` is more accurate

(`||` treats as falsy and uses the fallback: false, 0, '', null, undefined, NaN
`??` only uses the fallback for: null, undefined)

ijreilly (approved):

PR #16401, opened 2025-12-08
Fix Zapier

charlesBochet:

PR #16616, opened 2025-12-17
Remove position input from zapier

martmull:
As title
Fixes zapier tests


prastoin (commented):
Hey there, a while ago I've seen https://github.com/twentyhq/twenty/pull/16401 pass by
Have you checked with Charles ?

martmull:
@prastoin I might miss something but I guess this is not the purpose of this ticket isn't it?

prastoin:
@martmull which ticket ?

packages/twenty-zapier/src/creates/crud_record.ts
```diff
@@ -1,12 +1,18 @@
...
 import { computeInputFields } from '../utils/computeInputFields';
-import { InputData } from '../utils/data.types';
+import { type InputData } from '../utils/data.types';
 import handleQueryParams from '../utils/handleQueryParams';
 import requestDb, { requestSchema } from '../utils/requestDb';
 import { DatabaseEventAction } from '../utils/triggers/triggers.utils';
-import { capitalize } from 'twenty-shared/utils';
+import { isNonEmptyString } from '@sniptt/guards';
+
+const capitalize = (stringToCapitalize: string) => {
```
prastoin, line 11: ***Question:*** Twenty-shared import were failing ?

martmull:
> @martmull which ticket ?

I mean PR

packages/twenty-zapier/CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
+## 2.1.1
+* Add changelog
+
+## 2.1.0
```
prastoin, line 4: ***Question:*** Is this required ?

prastoin (commented):
Hey there ! What was the initial issue ? `twenty-shared` not being found at build time ?

packages/twenty-zapier/CHANGELOG.md
martmull: yes

packages/twenty-zapier/src/creates/crud_record.ts
martmull: yes, "zapier push" command was failing

prastoin: It might become a bottleneck duplicating the field metadata type and so on at some point ?
Just challenging the fix versus trying to fix the dependency injection
Seems like this is a vendoring issue too such as for `twenty-sdk` yesterday

prastoin (approved):
Approving this right now, might not be the most ideal fix as we might wanna be vendoring or investigating the twenty-shared not found root cause instead of duplicating it

PR #16622, opened 2025-12-17
Fix drag and drop in dropdown

bosiraphael:
## Bug description

Due to the dropdown using floating ui, there was an offset on the draggable item when the drag and drop was implemented inside a scrollable dropdown.

### Before

https://github.com/user-attachments/assets/fe4b4c36-39ae-4d26-9e19-85d5ffd23b30

### After

https://github.com/user-attachments/assets/1f00b9b0-c231-49cb-a232-75ca42a98e8c


lucasbordeau (approved):
LGTM

PR #16633, opened 2025-12-17
[POC] Real-Time

thomtrp:
https://github.com/user-attachments/assets/3ad7a4f3-d08b-4a1c-b3c2-bb42ef9f3575

Real time POC.

Next steps:
- find out the correct API for subscription. Should be triggered directly in query hooks (useFindManyRecords, useLazy...)
- improve query matching


packages/twenty-front/src/modules/object-record/hooks/useLazyFindManyRecords.ts
```diff
@@ -145,9 +145,68 @@ export const useLazyFindManyRecords = <T extends ObjectRecord = ObjectRecord>({
     ],
   );
 
+  const refetchRecords = useRecoilCallback(
```
thomtrp: waiting to know if I should comment or remove that part

packages/twenty-front/src/modules/object-record/record-table/virtualization/hooks/useTriggerInitialRecordTableDataLoad.ts
```diff
@@ -224,6 +256,19 @@ export const useTriggerInitialRecordTableDataLoad = () => {
     ],
   );
 
+  useSubscribeToRefetch({
```
thomtrp: same

packages/twenty-server/src/engine/subscriptions/subscriptions.resolver.ts
```diff
@@ -46,4 +52,46 @@ export class SubscriptionsResolver {
...
+      if (filteredIds.length === 0) {
+        return null;
+      }
+
+      return {
+        subscriptionIds: filteredIds,
+      };
+    },
+  })
+  onRefetchSignal(
```
thomtrp: not really fan of the naming. To be discussed

packages/twenty-front/src/modules/object-record/record-table/virtualization/hooks/useTriggerInitialRecordTableDataLoad.ts
```diff
@@ -224,6 +256,19 @@ export const useTriggerInitialRecordTableDataLoad = () => {
...
+    query: findManyRecordsQuery,
```
ijreilly: this reminds me that we are other queries subject to value changes from external updates: aggregate queries
on table view they are triggered from RecordTableColumnAggregateFooterValue
Not a very visible case but worth tackling. Its likely that a refresh is already triggered by table data refresh, but not sure!

packages/twenty-front/src/modules/subscription/components/SubscriptionProvider.tsx
```diff
@@ -0,0 +1,89 @@
...
+    const token = tokenPair?.accessOrWorkspaceAgnosticToken?.token;
+
+    if (!token) {
+      return null;
+    }
+
+    return createClient({
+      url: `${REACT_APP_SERVER_BASE_URL}/graphql`,
+      headers: {
+        Authorization: `Bearer ${token}`,
```
ijreilly: this feels a bit off, maybe we can create a util?

packages/twenty-front/src/modules/subscription/components/SubscriptionProvider.tsx
```diff
@@ -0,0 +1,89 @@
...
+          }
+
+          for (const subscriptionId of data.onRefetchSignal.subscriptionIds) {
+            const entry = registry.get(subscriptionId);
+            entry?.onRefetch();
+          }
+        },
+        error: (error) => {
+          // eslint-disable-next-line no-console
+          console.error('Subscription error:', error);
```
ijreilly: is this intentional?

packages/twenty-front/src/modules/subscription/components/SubscriptionProvider.tsx
```diff
@@ -0,0 +1,89 @@
...
+  }, [registry]);
+
+  useEffect(() => {
+    if (!sseClient || subscriptions.length === 0) {
+      return;
+    }
+
+    const unsubscribe = sseClient.subscribe(
+      {
+        query: print(ON_REFETCH_SIGNAL),
```
ijreilly: I dont see print() used anywhere else (maybe ive overlooked)

ijreilly (commented):
you can ignore my comment as this is actually a poc - the check on aggregate queries can be done though !

packages/twenty-front/src/modules/subscription/components/SubscriptionProvider.tsx
thomtrp: yes error cannot be null and I found it useful to see the logs for now

packages/twenty-front/src/modules/subscription/components/SubscriptionProvider.tsx
thomtrp: this is actually quite current, at least in tests. Queries are in AST and I want to pass a string

packages/twenty-front/src/modules/subscription/components/SubscriptionProvider.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const { sseClient } = useSseClient();
+
+  const subscriptions = useMemo(() => {
+    return Array.from(subscriptionRegistry.values()).map((entry) => ({
+      id: entry.id,
+      query: entry.query,
+    }));
+  }, [subscriptionRegistry]);
+
+  useEffect(() => {
```
lucasbordeau: Can we put this in an effect component as a sibling of this one, because the risk is that the whole application re-renders when the subscription registry changes, while this Provider role is essentially to initialize the SSE client.

packages/twenty-front/src/modules/subscription/components/SubscriptionProviderEffect.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { useRecoilValue } from 'recoil';
+import { isDefined } from 'twenty-shared/utils';
+
+export const SubscriptionProviderEffect = () => {
+  const subscriptionRegistry = useRecoilValue(subscriptionRegistryState);
+
+  const { sseClient } = useSseClient();
+
+  const subscriptions = useMemo(() => {
+    return Array.from(subscriptionRegistry.values()).map((entry) => ({
```
guillim: ~~what is the point of the .map here ?~~ removing the `onRefetch` property. I get it now :)

lucasbordeau (approved):
LGTM

PR #16642, opened 2025-12-17
Filter expected errors from sentry

abdulrahmancodes:


packages/twenty-front/src/modules/apollo/services/apollo.factory.ts
```diff
@@ -247,8 +247,14 @@ export class ApolloFactory<TCacheShape> implements ApolloManager<TCacheShape> {
...
                 }
+                case 'GRAPHQL_VALIDATION_FAILED':
+                case 'NOT_FOUND':
                 case 'BAD_USER_INPUT':
-                case 'FORBIDDEN': {
+                case 'FORBIDDEN':
+                case 'METHOD_NOT_ALLOWED':
+                case 'TIMEOUT':
+                case 'CONFLICT':
+                case 'METADATA_VALIDATION_FAILED': {
```
abdulrahmancodes, line 254: These were already filtered in the BE

packages/twenty-front/src/modules/apollo/services/apollo.factory.ts
```diff
@@ -247,8 +247,14 @@ export class ApolloFactory<TCacheShape> implements ApolloManager<TCacheShape> {
                   console.log('UNAUTHENTICATED, triggering token renewal');
                   return handleTokenRenewal(operation, forward);
                 }
+                case 'GRAPHQL_VALIDATION_FAILED':
+                case 'NOT_FOUND':
                 case 'BAD_USER_INPUT':
-                case 'FORBIDDEN': {
+                case 'FORBIDDEN':
+                case 'METHOD_NOT_ALLOWED':
```
charlesBochet: @abdulrahmancodes I think the frontend should report to Sentry the error code that should never happen:
- METHOD_NOT_ALLOWED (the FE should never do that)
- TIMEOUT (this is a BE TIMEOUT, right? this is not normal)

What the difference between: 
- GRAPHQL_VALIDATION_FAILED and BAD_USER_INPUT?

charlesBochet (changes requested):
@abdulrahmancodes left comments

packages/twenty-front/src/modules/apollo/services/apollo.factory.ts
abdulrahmancodes: Ah yes, you’re right. I’ve updated the filtering.
The FE should never use an incorrect http method. If that happens, it’s a bug on the FE side.

Regarding the graphql errors: I’ve removed GRAPHQL_VALIDATION_FAILED from filtering. This error indicates that the graphql query structure itself is invalid, which points to a FE bug in how the query is constructed.

BAD_USER_INPUT remains filtered because, in that case, the query structure is valid but the provided input violates business rules. This is expected user-level validation and not a FE bug.

For TIMEOUT, yes, this represents a BE timeout and should not be considered normal.

charlesBochet (approved):

charlesBochet:
ty @abdulrahmancodes

PR #16643, opened 2025-12-17
Fix/close filter by enter

### LINKED ISSUE #16636: I should be able to press "enter" to exit input
When adding or modifying a filter, I should be able to close it if I press "enter".

Image

Image

It works if I click "escape", but not "enter".

ManikanthMartha:
Fixes #16636 

Added useCloseDropdown() hook and set onEnter prop to onEnter={closeDropDown()} using dropdownID

EDIT from @charlesBochet after refactoring:
- ObjectDropdownFilters are used in 3 places: Main Filter menu, EditableChip, AdvancedFilters
- deprecate vectorSearch in view filter area, we are not using them, we are doing a anyField filter now. While refactoring the points below, I did not want to maintain vectorSearch as it was not used anymore
- stop confusing the dropdownId (which is an id to interact with a specific dropdown) and componentInstanceIds (which is used to scope component states) for EditableFilter case
- I haven't fixed the confusion for MainFilter case
- It was already handled for AdvancedFilter case


guillim:
Looking at your PR

guillim:
your code (before the refactor of chares) was not working @ManikanthMartha 

https://github.com/user-attachments/assets/25c51ab9-56df-4c0c-b87d-536431346a18

@charlesBochet you added some logic but i see this doing the QA

lucasbordeau (approved):
LGTM

PR #16647, opened 2025-12-18
feat: soft disable LOCAL code interpreter driver in production

FelixMalfait:
## Summary

Instead of throwing an error at server startup when LOCAL code interpreter is configured in production, we now return a DisabledDriver that only throws when the code interpreter is actually used.

## Changes

- Created `DisabledDriver` class that implements `CodeInterpreterDriver` but throws an error only when `execute()` is called
- Added `DISABLED` to the `CodeInterpreterDriverType` enum
- Updated the factory to return a `DISABLED` driver config instead of throwing when LOCAL is used in production
- Updated the module to handle the new `DISABLED` driver type

## Motivation

Many users don't need the code interpreter feature and want to deploy to production without configuring E2B. Previously, the server would crash at startup if `CODE_INTERPRETER_TYPE=LOCAL` was set in production.

**Before:** Server crashes at startup in production if `CODE_INTERPRETER_TYPE=LOCAL`

**After:** Server starts fine. The error only occurs if someone actually tries to **use** the code interpreter feature, at which point they get a clear error message explaining how to configure E2B.

### DIFF
--- a/packages/twenty-server/.env.example
+++ b/packages/twenty-server/.env.example
@@ -4,6 +4,7 @@ PG_DATABASE_URL=postgres://postgres:postgres@localhost:5432/default
 REDIS_URL=redis://localhost:6379
 APP_SECRET=replace_me_with_a_random_string
 SIGN_IN_PREFILLED=true
+CODE_INTERPRETER_TYPE=local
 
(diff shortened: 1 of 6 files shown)


charlesBochet (commented):
LGTM

PR #16649, opened 2025-12-18
fix: ensure QueryRunner is released in workspace schema operations

FelixMalfait:
## Summary

Fixes a database connection leak in `WorkspaceDataSourceService` where `QueryRunner.release()` was not being called when schema operations failed.

## Problem

The `createWorkspaceDBSchema` and `deleteWorkspaceDBSchema` methods use TypeORM's QueryRunner but didn't wrap the operations in try-catch-finally blocks. When schema operations fail (e.g., permission denied, schema conflicts), the `QueryRunner.release()` method was never called.

**Impact:** Failed schema operations leak database connections, which can exhaust the connection pool and cause the application to hang or crash under load.

## Solution

Wrap both methods in try-finally blocks to ensure `queryRunner.release()` is always called, regardless of whether the operation succeeds or fails.

## Changes

- `createWorkspaceDBSchema`: Wrapped in try-finally to ensure connection release
- `deleteWorkspaceDBSchema`: Wrapped in try-finally to ensure connection release


charlesBochet (commented):
LGTM

PR #16652, opened 2025-12-18
Add pattern for variable tag in tiptap

thomtrp:
Since we now store rich text value in blocknote rather than markdown, variables need to be resolved accordingly. 

Replacing the variable tag pattern `{"type":"variableTag","attrs":\{"variable":"(\{\{[^{}]+\}\})"\}\}` by a blocknote text `{"type":"text","text":"${escapedText}"}`

Fixes https://github.com/twentyhq/twenty/issues/16583

To test :
- build a workflow that creates a note/ sends an email with a variable in the body
- make sure the result is properly formatted once run

### DIFF
--- a/packages/twenty-server/src/modules/workflow/workflow-executor/workflow-actions/record-crud/create-record.workflow-action.ts
+++ b/packages/twenty-server/src/modules/workflow/workflow-executor/workflow-actions/record-crud/create-record.workflow-action.ts
@@ -10,11 +10,13 @@ import {
   RecordCrudExceptionCode,
 } from 'src/engine/core-modules/record-crud/exceptions/record-crud.exception';
 import { CreateRecordService } from 'src/engine/core-modules/record-crud/services/create-record.service';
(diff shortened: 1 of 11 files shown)


packages/twenty-server/src/modules/workflow/workflow-executor/workflow-actions/record-crud/create-record.workflow-action.ts
```diff
@@ -37,8 +40,25 @@ export class CreateRecordWorkflowAction implements WorkflowAction {
...
+      );
+
+    const inputWithResolvedRichText = {
+      ...rawInput,
+      objectRecord: resolveRichTextFieldsInRecord(
+        rawInput.objectRecord,
+        objectMetadataInfo,
+        context,
+      ),
+    };
```
Devessier, line 58: 💪

Devessier (approved):
Great!

PR #16663, opened 2025-12-18
fix: prevent writing in command menu input when modal is open

### LINKED ISSUE #13467: When a modal is open and the command menu search is focused, we can still write in the input
## Bug Description

When a modal is open and the command menu search is focused, we can still write in the input

https://github.com/user-attachments/assets/630e6555-f5ca-4d45-a708-e9bb93719b42

## Expected behavior

Keyboard actions other than the one defined by the modal shouldn't do anything.

abdulrahmancodes:
Closes #13467 

https://github.com/user-attachments/assets/e96814a8-fa2b-48b6-bf0f-922dc9cdb95f

### DIFF
--- a/packages/twenty-front/src/modules/command-menu/components/CommandMenuTopBarInputFocusEffect.tsx
+++ b/packages/twenty-front/src/modules/command-menu/components/CommandMenuTopBarInputFocusEffect.tsx
@@ -1,6 +1,6 @@
 import { useRecoilValue } from 'recoil';
 
-import { SIDE_PANEL_FOCUS_ID } from '@/command-menu/constants/SidePanelFocusId';
+import { COMMAND_MENU_SEARCH_INPUT_FOCUS_ID } from '@/command-menu/constants/CommandMenuSearchInputFocusId';
 import { commandMenuPageState } from '@/command-menu/states/commandMenuPageState';
 import { CommandMenuPages } from '@/command-menu/types/CommandMenuPages';
 import { currentFocusIdSelector } from '@/ui/utilities/focus/states/currentFocusIdSelector';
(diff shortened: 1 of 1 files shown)


packages/twenty-front/src/modules/command-menu/components/CommandMenuTopBarInputFocusEffect.tsx
```diff
@@ -12,15 +14,24 @@ export const CommandMenuTopBarInputFocusEffect = ({
   inputRef,
 }: CommandMenuTopBarInputFocusEffectProps) => {
   const commandMenuPage = useRecoilValue(commandMenuPageState);
+  const focusStack = useRecoilValue(focusStackState);
+  const isModalInStack = focusStack.some(
+    (item) => item.componentInstance.componentType === FocusComponentType.MODAL,
+  );
 
   useEffect(() => {
+    if (isModalInStack) {
```
charlesBochet: @abdulrahmancodes let's avoid adding behaviors in effect. Here is what we should do:
1) Make sure that we can only type in the command menu top bar when the field has the focus (we should put the focus in the focusStack when we focus the field, and remove it from the stack when we blur
2) when we open the modal, we should push a focus for the modal in the focusStack
3) when we close the modal, it should be removed

@bosiraphael do you agree?

charlesBochet: Actually discussed with @bosiraphael  and modal and text focus are correctly pushed to the stack. 
And it's not great to blur the input so we might not have a better fix...

charlesBochet: Let's keep the effect then, but change your check isModalInStack by a check: isTextInputFocused (and we check if the TextInput focusId is the currentItem in the focusStack)

lucasbordeau:
I'm closing this because this bug cannot be reproduced with the new right drawer, plus we don't have a confirmation modal for soft delete anymore. 

Let's re-open this if we still experience this.

PR #16664, opened 2025-12-18
Always enable drag selection when side panel is opened

bosiraphael:
We disabled that before because the side panel was always on top of the content. But now it makes sense to enable this.


Weiko (approved):

Devessier (approved):
LGTM!

PR #16665, opened 2025-12-18
Fixed command menu and main container layout

lucasbordeau:
This PR fixes https://github.com/twentyhq/twenty/issues/16645

It solves two problems : 
- Command menu for mobile was outside of its proper place, it should have been portaled instead of lifted that high in the hierarchy, which is done here, thus avoiding context issues.
- CSS was odd due to a code path removing main container styling for mobile display, everything has been cleaned with explicit and durable naming. I used "main container layout" instead of  "page layout" to disambiguate from page layout feature.

## Before 

image

## After 

image


Devessier (approved):
That's a true masterpiece. Approving and sending to production immediately.

PR #16670, opened 2025-12-18
Fix subscription cross tenant issue

martmull:
As title


packages/twenty-server/src/engine/subscriptions/subscription.service.ts
```diff
@@ -0,0 +1,50 @@
+import { Inject, Injectable } from '@nestjs/common';
+
+import { RedisPubSub } from 'graphql-redis-subscriptions';
+
+export enum SUBSCRIPTION_CHANNEL {
```
charlesBochet: should be in a separate file

charlesBochet (approved):
Minor comment, LGTM

PR #16672, opened 2025-12-18
Fix view field updates not persisting

### LINKED ISSUE #16381: when changing size of column in companies section , a toast error with this text pops up : An error occurred.
## Bug Description

Image

### LINKED ISSUE #16417: changing column size throw error
when changing the size of column an error toast pops up and says : An Error Occurred .

Image

RajdeepDs:
Fixes an issue where column width changes and other view field modifications were not persisted, and a toast error was shown.

**Root Causes:**
1. Frontend was using client-generated IDs instead of database IDs for updates
2. Backend cache was stale, causing `View field to update not found` error

**Changes:**
- Use the existing database ID when updating view fields
- Exclude client-generated IDs when creating new view fields  
- Invalidate flat entity maps cache before/after view field operations
- Refresh both view and view field caches to prevent validation errors

Fixes #16417 
Closes #16381


prastoin (changes requested):
Hello there, thanks for your contribution
`twenty-server` modifications are redundant as the validate build and run will invalidate the required cache
Front update seems legit though !

RajdeepDs:
Hello @prastoin thanks for the clarification!
I re-tested the fix after removing the twenty-server changes, and the issue still works correctly with the frontend update alone.
I've removed the server-side cache invalidation as it was indeed redundant.

prastoin (approved):
Thanks !

prastoin (changes requested):
Wait we need to keep the id in the create view field input though

packages/twenty-front/src/modules/views/hooks/useSaveCurrentViewFields.ts
prastoin: ***Bug:*** Unless I'm mistaken this brings a regression for any created view field, pre-generated id won't be send to the server anymore

RajdeepDs: Fixed 👍
The regression was caused by stripping the client-generated `id` during creation. I’ve preserved the `id` in `CreateViewFieldInput`, while keeping updates bound to the existing DB id. I also updated the equality check to include `aggregateOperation` to avoid skipped updates.

prastoin (approved):
Looks good ! Thanks for your contribution !

PR #16677, opened 2025-12-18
Read some endpoints on replica db (behind feature flag)

ijreilly:
Next steps: 
- move some workers' activities to replica db


packages/twenty-server/src/engine/twenty-orm/global-workspace-datasource/global-workspace-datasource.service.ts
```diff
@@ -47,7 +47,32 @@ export class GlobalWorkspaceDataSourceService
...
+        url: this.twentyConfigService.get('PG_DATABASE_REPLICA_URL'),
+        type: 'postgres',
+        logging: this.twentyConfigService.getLoggingConfig(),
+        entities: [],
+        ssl: this.twentyConfigService.get('PG_SSL_ALLOW_SELF_SIGNED')
+          ? {
+              rejectUnauthorized: false,
+            }
+          : undefined,
+        poolSize: this.twentyConfigService.get('PG_POOL_MAX_CONNECTIONS'),
```
ijreilly: shall we keep the same as on the master?
We have ~3400 connections capacity (right @charlesBochet ?)
With the 13 pods (8 workers + 5 server), 40 connections for write + 40 connections for read = 80*13 = 1040

packages/twenty-server/src/engine/twenty-orm/global-workspace-datasource/global-workspace-datasource.service.ts
```diff
@@ -47,7 +47,32 @@ export class GlobalWorkspaceDataSourceService
...
+        extra: {
+          query_timeout: 10000, // 10 seconds,
```
ijreilly: should we increase the timeout ? To allow for very large queries for analytics (not useful now though i believe)

packages/twenty-server/src/engine/core-modules/twenty-config/config-variables.ts
```diff
@@ -834,6 +834,22 @@ export class ConfigVariables {
...
+    isEnvOnly: true,
+  })
+  @IsDefined()
+  @IsUrl({
+    protocols: ['postgres', 'postgresql'],
+    require_tld: false,
+    allow_underscores: true,
+    require_host: false,
+  })
+  PG_DATABASE_REPLICA_URL: string;
```
Weiko, line 851: I think this one shouldn't be required (that's what IsDefined() is doing here, probably IsOptional is what you want.

packages/twenty-server/src/engine/twenty-orm/global-workspace-datasource/global-workspace-datasource.service.ts
```diff
@@ -47,7 +47,32 @@ export class GlobalWorkspaceDataSourceService
       this.workspaceEventEmitter,
     );
 
+    this.globalWorkspaceDataSourceReplica = new GlobalWorkspaceDataSource(
+      {
+        url: this.twentyConfigService.get('PG_DATABASE_REPLICA_URL'),
```
Weiko: We should fallback to PG_DATABASE_URL if this one is undefined (for self-hoster, we probably won't spawn a replica DB in the image we provide to them)

packages/twenty-server/src/engine/twenty-orm/global-workspace-datasource/global-workspace-datasource.service.ts
Weiko: I think it's fine then!

packages/twenty-server/src/engine/twenty-orm/global-workspace-datasource/global-workspace-datasource.service.ts
Weiko: We could. I'd at least set it as an ENV variable instead of raw here

packages/twenty-server/src/engine/twenty-orm/global-workspace-datasource/global-workspace-orm.manager.ts
```diff
@@ -50,7 +92,16 @@ export class GlobalWorkspaceOrmManager {
       );
     }
 
-    const globalDataSource = await this.getGlobalWorkspaceDataSource();
+    const isReadOnReplicaEnabledForWorkspace =
+      await this.featureFlagService.isFeatureEnabled(
```
Weiko: nit: You already have access to `this.workspaceCacheService.getOrRecompute` in this service. The result will be the same but this will avoid an extra import. Not a strong requirement, just a suggestion!

Weiko (approved):
Left some comments, otherwise LGTM!

packages/twenty-server/src/engine/twenty-orm/global-workspace-datasource/global-workspace-datasource.service.ts
```diff
@@ -60,10 +98,22 @@ export class GlobalWorkspaceDataSourceService
     return this.globalWorkspaceDataSource;
   }
 
+  public getGlobalWorkspaceDataSourceReplica(): GlobalWorkspaceDataSource {
+    if (!this.globalWorkspaceDataSourceReplica) {
```
Weiko: nit: isDefined

packages/twenty-server/src/engine/api/common/common-query-runners/common-base-query-runner.service.ts
```diff
@@ -334,14 +337,29 @@ export abstract class CommonBaseQueryRunnerService<
...
-      queryRunnerContext.flatObjectMetadata.nameSingular,
-      rolePermissionConfig,
-    );
+    const isReadOnReplicaEnabled =
+      await this.featureFlagService.isFeatureEnabled(
+        FeatureFlagKey.IS_READ_ON_REPLICA_ENABLED,
+        workspaceId,
+      );
+
+    const repository =
```
Weiko: Not a blocker and not on you but I feel like preparing the repository early here does not bring much value, especially if you pass your datasource as a parameter as well.
I feel getReadRepository method from the orm manager is not very useful, it's just a shortcut to `this.getGlobalDatasource().getRepository()`.
Maybe something to think about later!

packages/twenty-server/src/engine/core-modules/twenty-config/config-variables.ts
```diff
@@ -1449,6 +1465,24 @@ export class ConfigVariables {
...
   AWS_SES_ACCOUNT_ID: string;
+
+  @ConfigVariablesMetadata({
+    group: ConfigVariablesGroup.SERVER_CONFIG,
+    description: 'Timeout in milliseconds for primary database queries',
+    type: ConfigVariableType.NUMBER,
+  })
+  @CastToPositiveNumber()
+  @IsOptional()
+  PG_DATABASE_PRIMARY_TIMEOUT_MS: number = 10000;
```
Weiko, line 1477: since this is now used during onModuleInit (in comparison with workspace-datasource.factory that was doing new datasources on the fly), we should have isEnvOnly for those as well (meaning they can't be changed in the admin panel)

Weiko (approved):
LGTM!

packages/twenty-server/src/engine/core-modules/twenty-config/config-variables.ts
ijreilly: ok thanks, did not know about this arg !

packages/twenty-server/src/engine/api/common/common-query-runners/common-base-query-runner.service.ts
ijreilly: You are right I can at least do directly 
```
    const globalWorkspaceDataSource =
      this.isReadOnly && isReadOnReplicaEnabled
        ? await this.globalWorkspaceOrmManager.getGlobalWorkspaceDataSourceReplica()
        : await this.globalWorkspaceOrmManager.getGlobalWorkspaceDataSource();

    const repository = globalWorkspaceDataSource.getRepository(
      queryRunnerContext.flatObjectMetadata.nameSingular,
      rolePermissionConfig,
    );
```

here

PR #16680, opened 2025-12-18
fix: preserve relations during optimistic cache update when modifying task assignee

### LINKED ISSUE #16624: On a Record page (ex: Company), adding a task and modifying the assignee is resetting the relations (FE only)
## Bug Description

To reproduce:
- go to a company record page
- New task
- change title + change assignee
- see that relation disappear

## Expected behavior

- relation should not disappear

Image

akramcodez:
## What does this PR do?

Fixes a bug where changing a task's assignee on a Record page caused relations (like taskTargets) to disappear from the UI.

### Explain How the Feature Works

The optimistic cache update now preserves all existing relation fields when modifying task properties. Previously, only the fields being updated were included in the cache update, causing other relations to be dropped.

**Technical fix:**
Merge `computedRecordGqlFields` (all depth-1 fields) with fields from the optimistic input to ensure relation fields aren't lost during cache updates.

### Relevant User Scenarios

- Users editing task assignees on Company/Person record pages
- Any scenario where updating one field shouldn't affect other relations
- Prevents unexpected UI state loss during optimistic updates

fixes #16624

### DIFF
--- a/packages/twenty-front/src/modules/object-record/hooks/useUpdateOneRecord.ts
+++ b/packages/twenty-front/src/modules/object-record/hooks/useUpdateOneRecord.ts
@@ -118,12 +118,15 @@ export const useUpdateOneRecord = <
       isDefined(cachedRecordWithConnection);
 
     if (shouldHandleOptimisticCache) {
-      const recordGqlFields = generateDepthRecordGqlFieldsFromRecord({
-        objectMetadataItem,
-        objectMetadataItems,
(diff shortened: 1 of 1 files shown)


charlesBochet:
Hi @akramcodez, thanks for opening a PR. I believe it has been fixed by this PR: https://github.com/twentyhq/twenty/pull/16707

thanks for giving a try, this is much appreciated and sorry we are not merging your PR this time

PR #16686, opened 2025-12-18
Fix standard agent and roles deletion command

prastoin:
# Introduction
Caught red handed, introduced a failing command in https://github.com/twentyhq/twenty/pull/16499 that was failing even in system build which is should not


Weiko (approved):
LGTM

martmull (approved):
did you test it? 😅

PR #16694, opened 2025-12-18
Add useUpdateManyRecords hook and message folders sync status mutation

neo773:
- Added new `useUpdateManyRecords` hook for batch record updates with optimistic cache updates
- Added `updateMessageFoldersSyncStatus` hook for managing message folder sync state
- Redesigned Message Folders List with BreadCrumb and Animations


packages/twenty-front/src/modules/object-metadata/utils/generateUpdateManyRecordsMutation.ts
```diff
@@ -0,0 +1,29 @@
...
+  const capitalizedObjectNamePlural = capitalize(objectMetadataItem.namePlural);
+
+  const mutationResponseField = getUpdateManyRecordsMutationResponseField(
+    objectMetadataItem.namePlural,
+  );
+
+  const updateManyRecordsMutation = gql`
+    mutation UpdateMany${capitalizedObjectNamePlural}($filter: ${capitalizedObjectNameSingular}FilterInput!, $data: ${capitalizedObjectNameSingular}UpdateInput!) {
+      ${mutationResponseField}(filter: $filter, data: $data) {
+        id
```
charlesBochet, line 23: note: I guess we should pass the fields we want to have too but that can be done later

charlesBochet (approved):

PR #16697, opened 2025-12-19
Fix stale relative date displays

abdulrahmancodes:
Closes #14697


guillim:
as discussed in the related issue, we do not want to handle the fix this way. 

Note: thanks @abdulrahmancodes because your suggestion was actually working. it's just not going the right direction for the codebase.

PR #16702, opened 2025-12-19
fix: Next has a denial of service with server components

mabdullahabaid:
The alerts regarding CVE-2025-55182, CVE-2025-55183, CVE-2025-55184 are false-positive given Twenty only imported Next 15.2.4 via `react-emails` as a devDependency, so it should have never made it to the production build - it was reported due to yarn.lock file containing the version. However, in order to remove the alerts, updated react-emails to 4.0.17 (latest patch in 4.0 minor release) and Next version it imports to 15.5.9.

[Dependabot Alert 337](https://github.com/twentyhq/twenty/security/dependabot/337), [Dependabot Alert 343](https://github.com/twentyhq/twenty/security/dependabot/343), [Dependabot Alert 344](https://github.com/twentyhq/twenty/security/dependabot/344).

Additionally, updated Next 14.2.33 to 14.2.35 to resolve another couple alerts reported in CVE-2025-55184.

[Dependabot Alert 345](https://github.com/twentyhq/twenty/security/dependabot/345), [Dependabot Alert 346](https://github.com/twentyhq/twenty/security/dependabot/346).


guillim:
You have some typing error @mabdullahabaid . Let me know when you think it's ready for review

mabdullahabaid:
@guillim ready for review.

The comment about `render` function by Sentry is false since server uses `@react-emails/render` package and it maintains the definition for `plainText: true`.

`react-emails` is only used as a dev server to display email templates. After version 4.0.8, they separated out preview-server and react-emails. The former provides the UI, while the latter provides the `email` CLI command - we need both.

guillim (approved):

PR #16709, opened 2025-12-19
Prevent /graphql access when not authenticated

charlesBochet:
Fixes https://github.com/twentyhq/twenty/issues/13103


charlesBochet:
Not ready, will re-open soon

PR #16718, opened 2025-12-19
Fix composite upsert

thomtrp:
Doing an upsert on existing value, composite field not updated properly.

```
Input: {
  id: "08ca34fe-fc39-474f-adac-4d89f844e922",
  name: "tom",
  linkedinLink: {
    primaryLinkUrl: "https://www.linkedin.com/in/etienneyaouni1982",
    primaryLinkLabel: "etienne",
    secondaryLinks: null,
  },
}
```

Building `overwrites` for upsert forgets `linkedinLink` because column names are not flattened yet.
We don't want to call formatData yet on the input, because this is heavy.

Overriding `overwrites` on execute.

### DIFF
--- a/packages/twenty-server/src/engine/twenty-orm/repository/workspace-insert-query-builder.ts
+++ b/packages/twenty-server/src/engine/twenty-orm/repository/workspace-insert-query-builder.ts
@@ -115,6 +115,40 @@ export class WorkspaceInsertQueryBuilder<
         shouldBypassPermissionChecks: this.shouldBypassPermissionChecks,
       });
 
+      // Fix overwrites for composite fields - valuesSet contains formatted/flattened column names
+      // but overwrites was computed before formatData, missing composite field columns
+      if (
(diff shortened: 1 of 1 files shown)


charlesBochet (approved):
This should go through commonAPI but we can merge this for now as you are the only user of this API

packages/twenty-server/src/engine/twenty-orm/repository/workspace-insert-query-builder.ts
```diff
@@ -115,6 +115,40 @@ export class WorkspaceInsertQueryBuilder<
...
+              (col) =>
+                allValueKeys.has(col.databaseName) &&
+                !this.expressionMap.onUpdate.overwrite!.includes(
+                  col.databaseName,
+                ),
+            )
+            .map((col) => col.databaseName);
+
+          this.expressionMap.onUpdate.overwrite = [
+            ...this.expressionMap.onUpdate.overwrite,
```
charlesBochet, line 146: you seem to end up with both

PR #16723, opened 2025-12-19
Added option to prevent sending requests to twenty-icons

BOHEUS:
Solution for #15891

https://github.com/user-attachments/assets/82ce5c4b-0a78-45a8-8196-413473887820


BOHEUS:
I'd be more than happy if someone could tell me how to solve problem with improper React hook as I don't see passing value several functions as a proper solution

packages/twenty-server/src/engine/core-modules/workspace/workspace.entity.ts
FelixMalfait: The way we would do it is through a CONFIG VARIABLE not via a database field (server-level not workspace level).  I think I'll close the PR for now because it requires a different approach but feel free to re-open starting from scratch. 

Also "allowExternalRequest" is too broad, we try to name things so that there's no doubt what this is about when you read them.


Thanks!!

BOHEUS: Okay, thanks for information, I'll work on it in nearest future then :+1:

BOHEUS:
@FelixMalfait solution changed to client config available in Admin Panel, also covers cases with the non-chip avatars like in CommandMenuPageInfoLayout or ShowPageSummaryCard

https://github.com/user-attachments/assets/5c0d10ac-aff0-4cae-bc43-59cbb05afaef

charlesBochet (approved):
LGTM but maybe a nicer approach would have been a service worker here, this could be also useful for offline mode or to capture and control any outgoing traffic from twenty UI

PR #16729, opened 2025-12-20
fix onboarding for messaging

neo773:


charlesBochet:
@neo773 I've tested this one and I have the following issues:

- when I sign up and create a new workspace, I connect my account during the flow
- then I have clicked on "Settings > Account" and the account was briefly showed as Not synced, not sure why?
- then when I refresh the page, I'm always proposed to invite my team
image

charlesBochet (approved):

PR #16740, opened 2025-12-22
Enable read on replica for all 

ijreilly:
Removing the feature flag to enable read on replica for all workspaces.
It will still be possible to toggle off the feature by removing the env variable `PG_DATABASE_REPLICA_URL`.


Weiko (approved):
LGTM!

PR #16757, opened 2025-12-22
feat: add customizable field ordering for detail views

fatih4159:
This commit implements a new feature that allows users to reorder fields in detail views through a drag-and-drop interface in settings.

Changes:
- Added detailFieldOrder column to objectMetadata table to store custom field order
- Updated ObjectMetadataEntity and GraphQL DTOs to include detailFieldOrder
- Created SettingsDetailFieldOrderSection component with drag-and-drop functionality
- Updated useFieldListFieldMetadataItems to use custom field order when available
- Integrated field ordering UI into ObjectFields settings tab

The field order is persisted per object type and automatically applied to all detail views. If no custom order is set, fields display alphabetically (existing behavior).


FelixMalfait:
Hey @fatih4159 sorry but we'll have to close this because there's a broader vision for this feature which we are currently working on at the moment. It will come early next year!

PR #16759, opened 2025-12-22
Fix database event emission

etiennejouan:
Fixes https://github.com/twentyhq/private-issues/issues/395

When calling `removeUserFromWorkspaceAndPotentiallyDeleteWorkspace` from `user.service`, 
```
        await workspaceMemberRepository.delete({
          userId: userWorkspace.userId,
        });
```
related database event is not emitted.

Same issue with update has been fixed [here](https://github.com/twentyhq/twenty/pull/13287/changes)

The database event is not emitted because `await eventSelectQueryBuilder.getOne()` in `workspace-delete-query-builder` returns `null`. This happens because the `selectQueryBuilder` has no entity—the database request is sent and the raw result is not `null`, but the entity is `null`, causing the final result to be null.

It can be fixed the same way update has been fixed, updating the `workspace-entity-manager`
- Pros : consistant with update but we should not forget to fix softDelete and restore
- Cons : `workspace-entity-manager` is a copy of typeORM logic + permission injection. Should it be more ?
 
Alternatively (as featured in this PR), it can be fixed by updating computeEventSelectQueryBuilder, inspired by TypeORM's logic in typeorm/query-builder/QueryBuilder.js at line 67.
- Pros : it fit with typeORM logic + It fixes all repository operations

### DIFF
--- a/packages/twenty-server/src/engine/twenty-orm/utils/compute-event-select-query-builder.util.ts
+++ b/packages/twenty-server/src/engine/twenty-orm/utils/compute-event-select-query-builder.util.ts
@@ -46,5 +46,14 @@ export const computeEventSelectQueryBuilder = <T extends ObjectLiteral>({
   eventSelectQueryBuilder.expressionMap.aliases = expressionMap.aliases;
   eventSelectQueryBuilder.setParameters(expressionMap.parameters);
 
+  if (
+    eventSelectQueryBuilder.expressionMap.selects.length === 0 &&
+    eventSelectQueryBuilder.expressionMap.mainAlias
+  ) {
+    eventSelectQueryBuilder.expressionMap.selects = [
+      { selection: eventSelectQueryBuilder.expressionMap.mainAlias.name },
+    ];
+  }
+
   return eventSelectQueryBuilder;
 };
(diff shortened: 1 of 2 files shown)


packages/twenty-server/src/engine/twenty-orm/utils/compute-event-select-query-builder.util.ts
```diff
@@ -46,5 +46,14 @@ export const computeEventSelectQueryBuilder = <T extends ObjectLiteral>({
   eventSelectQueryBuilder.expressionMap.aliases = expressionMap.aliases;
   eventSelectQueryBuilder.setParameters(expressionMap.parameters);
 
+  if (
```
ijreilly, line 49: Im in favor of this which replicates [typeorm's behaviour](https://github.com/typeorm/typeorm/blob/2359039daa4d7337c9152cc214ffa38b4576ebfe/src/query-builder/QueryBuilder.ts#L176) when switching from any query builder to a select query builder

If we choose this approach we could even want to revert the [changes made here](https://github.com/twentyhq/twenty/pull/13287/files) that are fixed by this (i think)

Weiko: Interesting. I agree with the strategy 👍

Weiko (approved):
LGTM

PR #16760, opened 2025-12-22
Adding one How To article to the workflow section

StephanieJoly4:
- added an image, one article
- updated the 2 files under the Navigation folder but not the docs.json
- no need to redirect the links given this is a new article


Weiko (approved):
LGTM

PR #16765, opened 2025-12-22
Auto-focus workflow title field when creating new workflow

### LINKED ISSUE #14135: Prompt Workflow Naming Upon Workflow Creation
## Current Behavior

When we create a new workflow now, we are not prompted to name our workflow. This results in us having a lot of untitled workflows as it is not necessary to name them before saving progress either. 

https://github.com/user-attachments/assets/28bfccd4-4d37-4a66-942e-dda620e1f7eb

## Expected behavior

A seamless way to make sure users name their workflows is to have the focus on the workflow name to prompt users to name their workflow before they start configuring it.

Image

abdulrahmancodes:
Closes #14135 

https://github.com/user-attachments/assets/4170c33d-4a2e-450a-98c0-20163d4b2382

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-table/hooks/useCreateNewIndexRecord.ts
+++ b/packages/twenty-front/src/modules/object-record/record-table/hooks/useCreateNewIndexRecord.ts
@@ -101,10 +101,21 @@ export const useCreateNewIndexRecord = ({
             });
           }
         } else {
-          navigate(AppPath.RecordShowPage, {
-            objectNameSingular: objectMetadataItem.nameSingular,
-            objectRecordId: recordId,
-          });
+          const isNewWorkflow =
+            objectMetadataItem.nameSingular === 'workflow' &&
+            !createdRecord.name;
+
+          navigate(
+            AppPath.RecordShowPage,
+            {
+              objectNameSingular: objectMetadataItem.nameSingular,
+              objectRecordId: recordId,
+            },
+            undefined,
+            isNewWorkflow
+              ? { state: { isNewWorkflow: true, objectRecordId: recordId } }
+              : undefined,
+          );
         }
 
         if (isDefined(recordIndexGroupFieldMetadataItem)) {
--- a/packages/twenty-front/src/modules/object-record/record-table/hooks/useCreateNewIndexRecord.ts
+++ b/packages/twenty-front/src/modules/object-record/record-table/hooks/useCreateNewIndexRecord.ts
@@ -101,9 +101,8 @@ export const useCreateNewIndexRecord = ({
             });
           }
         } else {
-          const isNewWorkflow =
-            objectMetadataItem.nameSingular === 'workflow' &&
-            !createdRecord.name;
+          const labelIdentifierFieldMetadataItem =
+            getLabelIdentifierFieldMetadataItem(objectMetadataItem);
 
           navigate(
             AppPath.RecordShowPage,
@@ -112,9 +111,14 @@ export const useCreateNewIndexRecord = ({
               objectRecordId: recordId,
             },
             undefined,
-            isNewWorkflow
-              ? { state: { isNewWorkflow: true, objectRecordId: recordId } }
-              : undefined,
+            {
+              state: {
+                isNewRecord: true,
+                objectRecordId: recordId,
+                labelIdentifierFieldName:
+                  labelIdentifierFieldMetadataItem?.name,
+              },
+            },
           );
         }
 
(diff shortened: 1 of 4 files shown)


packages/twenty-front/src/modules/object-record/record-table/hooks/useCreateNewIndexRecord.ts
```diff
@@ -101,10 +101,21 @@ export const useCreateNewIndexRecord = ({
             });
           }
         } else {
-          navigate(AppPath.RecordShowPage, {
-            objectNameSingular: objectMetadataItem.nameSingular,
-            objectRecordId: recordId,
-          });
+          const isNewWorkflow =
+            objectMetadataItem.nameSingular === 'workflow' &&
```
FelixMalfait: Use CoreObjectNameSingular instead of hardcoding 'workflow'

packages/twenty-front/src/modules/object-record/record-title-cell/hooks/useOpenNewRecordTitleCell.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  const openNewWorkflowTitleCell = useRecoilCallback(
+    ({ set }) =>
+      ({ recordId }: { recordId: string }) => {
+        const instanceId = getRecordFieldInputInstanceId({
+          recordId,
+          fieldName: 'name',
+          prefix: RecordTitleCellContainerType.PageHeader,
+        });
+
```
FelixMalfait, line 19: Did you check the existing behavior for focus when you create a new record that opens in the side panel? 
I feel like this issue should have been re-interpreted as: when I create a new record that does not open in the side panel then it should still autofocus on the title, like it does in the side panel (so no need to even refer to "workflow" anywhere!)

Check useCreateNewIndexRecord.ts ; useRecordTitleCell.tsx ; RecordTitleCellContainer.tsx ; RecordTitleCellTextFieldInput.tsx (autofocus)

packages/twenty-front/src/modules/object-record/record-table/hooks/useCreateNewIndexRecord.ts
FelixMalfait: Forget this, see below comment

packages/twenty-front/src/modules/object-record/record-title-cell/hooks/useOpenNewRecordTitleCell.ts
abdulrahmancodes: I've updated the PR to generalize this for all new records, not just workflows

abdulrahmancodes: Yeah, I checked the existing code. We aren't using the existing `useRecordTitleCell` hook because the it depends on `useContextStoreObjectMetadataItem` which throws if `objectMetadataItem` is `undefined`, and in `PageChangeEffect` the context store isn't set up yet for the new page.
As for why we're handling this in `PageChangeEffect` instead of `useCreateNewIndexRecord`: unlike the side panel which opens synchronously, navigating to the full page triggers `PageChangeEffect` which resets the focus stack. If we called `openRecordTitleCell` right after `navigate()`, the focus would get wiped by that reset. By handling it in `PageChangeEffect` after the reset, we ensure the focus is pushed in the correct order

packages/twenty-front/src/modules/object-record/record-table/hooks/useCreateNewIndexRecord.ts
```diff
@@ -101,10 +101,25 @@ export const useCreateNewIndexRecord = ({
...
+
+          navigate(
+            AppPath.RecordShowPage,
+            {
+              objectNameSingular: objectMetadataItem.nameSingular,
+              objectRecordId: recordId,
+            },
+            undefined,
+            {
+              state: {
```
FelixMalfait, line 115: This is not the most idiomatic way of doing things. Try to take more inspiration from what exists instead of adding new behavior (e.g. recoil state here). I'll do a quick POC to show you

FelixMalfait: Ok I spent 2 hours trying to do it cleanly with recoil but I had an issue with the focus stack which was really difficult to solve, I've been banging my head on it. I think we can merge like that even if it's not perfect.

FelixMalfait (approved):

PR #16768, opened 2025-12-23
fix: enable API key management of workspace views + fix permission bypass vulnerability

eeeggg:
Fixes #16739

- Remove empty string coercion in createCoreView that caused PostgreSQL UUID errors for API keys
- Add permission check allowing API keys with VIEWS permission to manage workspace views they created

API keys with 'Manage Views' permission can now create, update, and delete workspace views via both GraphQL and REST APIs.


FelixMalfait (approved):
Great work thank you! I did a small style update as we prefer small functions with early return to using else / nested if (even though there were some in the code already!)

FelixMalfait:
Ah damn merged too quickly without my commit, will open a new PR!

PR #16769, opened 2025-12-23
Fix: Record label identifier setting not saving on first attempt

abdulrahmancodes:


packages/twenty-front/src/modules/settings/data-model/objects/forms/components/SettingsDataModelObjectIdentifiersForm.tsx
```diff
@@ -59,10 +59,6 @@ export const SettingsDataModelObjectIdentifiersForm = ({
   const handleSave = async (
     formValues: SettingsDataModelObjectIdentifiersFormValues,
   ) => {
-    if (!formConfig.formState.isDirty) {
```
abdulrahmancodes, line 62: Removed the unnecessary isDirty check since we explicitly call `handleSubmit(handleSave)()` in response to `onChange`, we already know the user changed something. The check was redundant

FelixMalfait (approved):

PR #16770, opened 2025-12-23
fix: prevent API keys from creating UNLISTED views

FelixMalfait:
## Summary
UNLISTED views are personal views tied to a specific user, so API keys should not be able to create them.

## Changes
- Added check in `canUserCreateView` to block API keys from creating UNLISTED views
- Refactored the service to use smaller functions with early returns (no nested if/else)

### Creating Views
| Caller | Visibility | Has VIEWS Permission | Result |
|--------|------------|---------------------|--------|
| User | UNLISTED | (not checked) | ✅ Allow |
| User | WORKSPACE | Yes | ✅ Allow |
| User | WORKSPACE | No | ❌ Denied |
| **API Key** | **UNLISTED** | (not checked) | **❌ Denied** |
| API Key | WORKSPACE | Yes | ✅ Allow |
| API Key | WORKSPACE | No | ❌ Denied |

### DIFF
--- a/packages/twenty-server/src/engine/metadata-modules/view-permissions/services/view-access.service.ts
+++ b/packages/twenty-server/src/engine/metadata-modules/view-permissions/services/view-access.service.ts
@@ -1,7 +1,7 @@
 import { Injectable } from '@nestjs/common';
 
-import { isDefined } from 'twenty-shared/utils';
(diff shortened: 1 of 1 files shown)


FelixMalfait:
Merging as this was a FIX commit I forgot to push on a contributor PR I reviewed

PR #16773, opened 2025-12-23
Refactor command menu input

bosiraphael:
- Open input on item click
- `CommandMenuItemNumberInput` and `CommandMenuItemTextInput` don't only define the input but the whole menu item now
- Fixed behavior on escape

## Video QA

https://github.com/user-attachments/assets/bf2f03e9-d07c-4a1e-9bc8-6606839269ff


Devessier (approved):
Good for me! As usual.

PR #16777, opened 2025-12-23
[DASHBOARDS] Update iFrame error message

bosiraphael:
`Invalid URL` instead of `no data` for iFrame widgets.

CleanShot 2025-12-23 at 14 13 17@2x

### DIFF
--- a/packages/twenty-front/src/modules/page-layout/widgets/graph/components/GraphWidgetRenderer.tsx
+++ b/packages/twenty-front/src/modules/page-layout/widgets/graph/components/GraphWidgetRenderer.tsx
@@ -10,12 +10,7 @@ type GraphWidgetRendererProps = {
 
 export const GraphWidgetRenderer = ({ widget }: GraphWidgetRendererProps) => {
   if (!widget.configuration || !('graphType' in widget.configuration)) {
-    return (
-      <PageLayoutWidgetNoDataDisplay
-        widgetId={widget.id}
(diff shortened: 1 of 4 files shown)


packages/twenty-front/src/modules/page-layout/widgets/graph/components/GraphWidgetRenderer.tsx
```diff
@@ -10,7 +10,12 @@ type GraphWidgetRendererProps = {
 
 export const GraphWidgetRenderer = ({ widget }: GraphWidgetRendererProps) => {
   if (!widget.configuration || !('graphType' in widget.configuration)) {
-    return <PageLayoutWidgetNoDataDisplay widgetId={widget.id} />;
+    return (
+      <PageLayoutWidgetNoDataDisplay
+        widgetId={widget.id}
+        widgetType={widget.type}
```
Devessier: We have `useCurrentWidget` now to prevent prop drilling 🥸

Devessier (approved):
Good. I think we can refactor a bit to prevent prop drilling.

packages/twenty-front/src/modules/page-layout/widgets/graph/components/GraphWidgetRenderer.tsx
bosiraphael: Indeed ! I forgot about this masterpiece

PR #16781, opened 2025-12-23
Ensure record fields uniqueness across widgets

Devessier:
This PR dissociates record inputs so that several inputs for the same record field can live on the same page.

## Simple values demo

https://github.com/user-attachments/assets/a8c224d8-4cd6-4fe3-9cf8-01f6bdd2fca9

## Relations demo

https://github.com/user-attachments/assets/87267a8b-8ce5-41ce-8a78-cba2384828bb

## Doesn't break Record Table

https://github.com/user-attachments/assets/bcdd2b05-7cb4-4ed2-9093-58be3e04e43e

## Doesn't break Show Page

https://github.com/user-attachments/assets/f9ab9e14-012c-451e-a35f-01a165523f95


packages/twenty-front/src/modules/object-record/record-field-list/record-detail-section/relation/components/RecordDetailRelationSectionDropdownToOne.tsx
```diff
@@ -30,10 +30,12 @@ import { IconForbid, IconPencil } from 'twenty-ui/display';
 import { LightIconButton } from 'twenty-ui/input';
 
 type RecordDetailRelationSectionDropdownToOneProps = {
+  instanceId: string;
   dropdownTriggerClickableComponent?: ReactNode;
 };
 
 export const RecordDetailRelationSectionDropdownToOne = ({
+  instanceId,
```
FelixMalfait: nit: I feel instanceId is very generic and maybe it's better to be very explicit even if it's verbose.
E.g. in the code I see recordPickerInstanceId, selectableListInstanceId, etc. Sometimes it's just instanceId but I'd say we should move away from it for clarity

FelixMalfait (approved):

packages/twenty-front/src/modules/object-record/record-field-list/record-detail-section/relation/components/RecordDetailRelationSectionDropdownToOne.tsx
Devessier: I agree. I discussed this with Lucas, and he said these changes are too low-level. We should think about a global pattern to scope fields and allow multiple instances of them.

We agreed that I would replace my changes with a React context to provide a scoping identifier. I’ll do that tomorrow. Thanks again for your review.

PR #16789, opened 2025-12-23
fix: 🐛 visible fields count updates when toggling field

### LINKED ISSUE #16734: Linkedin field in Companies table is active even though it's not visible at all
**Scenario:**
1. Log in
2. Go to Companies table

**Actual:** Technically there are 8 visible fields in view but only 7 are available to user, Linkedin field is not available, cannot be added to view, filtered or sorted by even though it exists in Data model

https://github.com/user-attachments/assets/dfefbd8c-ad03-478e-990c-fbb70c465d31

**Expected:** Linkedin field is available to end user

Bhoomaahamso:
Fixes: #16734

Solution:
The ObjectOptionsDropdownDefaultView was reading from recordIndexFieldDefinitionsState via useObjectOptionsForBoard, while field visibility changes update currentRecordFieldsComponentState. This caused the "X shown" count to remain stale after hiding fields.

Changed to use visibleRecordFieldsComponentSelector (same state source used by ViewFieldsVisibleDropdownSection) so both components react to the same state updates.

https://github.com/user-attachments/assets/62eb5c98-f15f-4ee8-bdce-1ab4e4752f66


Bhoomaahamso:
@BOHEUS @thomtrp Could I get a review for this PR?

BOHEUS:
I mean, that's not the fix for a problem I reported but for another one I missed :sweat_smile: I'll leave the review to core team

etiennejouan (approved):
Hi @Bhoomaahamso, thank you for your contribution. Looks good to me and it works well ! 👏

PR #16668, opened 2025-12-18
Correctly separate widgets in side-column mode.

mabdullahabaid:
Closes [2006](https://github.com/twentyhq/core-team-issues/issues/2006)

Removed the top-border in FieldsWidget and added buttom-border to WidgetCard.

Beyond this, I tried making the spacing of these widgets even at the top and the bottom, but it turns out the widgets are also used in activity views, so configuring them for side-column messes up those activity views/tabs.

I feel that in order to make these widgets cleaner in side-column, we have to make the widgets context aware - I tried figuring out some other solution, but since some cards have spacing defined within, we do not really have a clean and quick solution imo.

If we do not align spacing, the following delta remains in bottom-spacing of the widgets and it looks ugly.

image

image

Let me know your thoughts.

PR #16804, opened 2025-12-24
Correctly separate widgets in side-column mode.

mabdullahabaid:
Had to close the other PR since I messed up re-basing somehow: https://github.com/twentyhq/twenty/pull/16668/files.

Moved changes to this new PR in order to resolve comments from the previous PR and also match the field-design to that on Figma.


mabdullahabaid:
I simplified FieldsWidgetSectionContainer because FieldsWidget only displays scalar fields (text, numbers, dates, etc.), not relation fields.

FieldsWidget filters out relation and morph relation fields at the field metadata level, so they never reach FieldsWidgetSectionContainer.

The original component (RecordDetailSectionContainer) includes features for relation fields: links to view all related records, dropdowns for managing relations, and conditional rendering based on whether records exist.

Since FieldsWidget excludes relation fields, FieldsWidgetSectionContainer doesn’t need those features, at least at this point.

packages/twenty-front/src/modules/page-layout/widgets/widget-card/components/WidgetCard.tsx
```diff
@@ -79,6 +79,8 @@ const StyledWidgetCard = styled.div<{
     if (variant === 'side-column' && !isEditable) {
       return css`
         background: ${theme.background.secondary};
+        padding: ${theme.spacing(2)};
+        border-bottom: 1px solid ${theme.border.color.light};
```
Devessier, line 83: Let's display a border bottom only if the widget isn't the last widget of its parent tab.

You could create a `useIsCurrentWidgetLastOfTab` hook that would get the tab that contains the current widgets, sorts its widgets and returns whether the last of them is the current widget.

Devessier: I created https://github.com/twentyhq/core-team-issues/issues/2030 for the follow up.

Devessier (approved):
Good job! Thank you. I created a follow up issue.

packages/twenty-front/src/modules/page-layout/widgets/widget-card/components/WidgetCard.tsx
mabdullahabaid: Okay, thank youu

PR #16807, opened 2025-12-25
feat: add updatedBy field to track last record modifier

abdulrahmancodes:
> [!NOTE]
> Implements last-modifier tracking and unified actor injection.
> 
> - New `ActorFromAuthContextService` replaces createdBy-only logic; pre-query hooks now inject both `createdBy` and `updatedBy` on create and `updatedBy` on update
> - Add `updatedBy` standard field to core/custom objects (e.g., `person`, `company`, `task`, `note`, `attachment`, `dashboard`, `workflow`, `workflowRun`) with IDs, metadata builders, and ORM entity fields
> - Record CRUD: `create` now sets both `createdBy` and `updatedBy`; `update` sets `updatedBy`; workflow actions use a shared workflow actor builder; REST base handler uses the new actor service
> - Seeder data and snapshots updated to include `updatedBy`; GraphQL result formatting adds defaults/handling for empty composite fields (incl. actor)
> - Frontend `useUpdateOneRecord` upserts returned record into the local store after mutation
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1f283778e9cb28a35bf84632a1a2997250bb3131. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/twenty-server/src/modules/dashboard/standard-objects/dashboard.workspace-entity.ts
```diff
@@ -91,6 +91,17 @@ export class DashboardWorkspaceEntity extends BaseWorkspaceEntity {
   @WorkspaceIsFieldUIReadOnly()
   createdBy: ActorMetadata;
 
+  @WorkspaceField({
+    standardId: DASHBOARD_STANDARD_FIELD_IDS.updatedBy,
+    type: FieldMetadataType.ACTOR,
+    label: msg`Updated by`,
+    icon: 'IconUserCircle',
+    description: msg`The user who last updated the record`,
```
FelixMalfait: nit: we refer to it as workspace member mostly (not user)

FelixMalfait (commented):
This looks good, great work again! Did you think about the impact on deployment? What if we we deploy but the sync metadata takes 2 hour to run on all workspaces (thousands). Would that create a down time / errors for operations during that meantime? (I did not check)

abdulrahmancodes:
@FelixMalfait The code checks if the field exists in object metadata before injection. If sync hasn't run yet for a workspace, it skips the injection. So, no errors during the transition period

FelixMalfait (commented):
Probably we should backfill views for existing workspaces in the future (not needed now).

Let's create an issue for later and we'll see when to prioritize it

FelixMalfait (approved):

PR #16808, opened 2025-12-25
feat: helm chart

dnplkndll:
# Add Helm Chart

- Introduces a Twenty Helm chart with sensible defaults: internal Postgres/Redis, auto DB creation/user, migrations, TLS via cert-manager, and quickstart docs.

## Feedback requested
- Handling replicas > 1 with local storage (warn/force S3?).
- Defaults/guards for ephemeral pods + S3.


charlesBochet (approved):
Thank you @dnplkndll. I could not test it but as this will be community maintained, I'm happy to merge it.

It's likely that the core team will maintain it at some point seems we are using helm internally and we will publish our own charts at some point so everybody can benefit for it (but right now they are not abstract enough, so this should be a good base)

PR #16031, opened 2025-11-24
Introduce a workspace member page.

mabdullahabaid:
- Refactored workspace member details into a focused Infos-only page.
- Aligned the flow with SettingsProfile, including controlled name inputs, debounced saves, and stable instance IDs.
- Added a dedicated member-picture upload flow. Introduced the MemberPictureUploader, connected to the uploadWorkspaceMemberProfilePicture mutation.
- Backend now includes a workspace-member resolver/module for profile-picture uploads. The endpoint is permission-guarded, streams files through FileUploadService, and returns the signed file without modifying the member entity.

---

> [!NOTE]
> Adds a workspace member detail page with picture/name management, integrates a new avatar upload mutation, and updates list routing; replaces the old profile picture uploader across profile and onboarding.
> 
> - **Frontend**
>   - **Settings Members**:
>     - Add `pages/settings/members/SettingsWorkspaceMember` with `MemberInfosTab`, `MemberNameFields`, and `MemberEmailField` for viewing/editing member info.
>     - Update routes in `SettingsRoutes` and add `SettingsPath.WorkspaceMemberPage`.
>     - Update `SettingsWorkspaceMembers` to navigate to member detail on row click and simplify row actions (remove dropdown menu).
>   - **Avatar Upload**:
>     - Introduce `WorkspaceMemberPictureUploader` using `uploadWorkspaceMemberProfilePicture` mutation.
>     - Replace `ProfilePictureUploader` in `SettingsProfile` and `onboarding/CreateProfile`.
>   - **GraphQL (client)**:
>     - Add `uploadWorkspaceMemberProfilePicture` mutation types/hooks in `generated(-metadata)/graphql.ts`.
> - **Backend**
>   - Add `UserWorkspaceResolver` with `uploadWorkspaceMemberProfilePicture` mutation guarded by `WorkspaceAuthGuard` and `SettingsPermissionGuard` (WORKSPACE_MEMBERS), using `FileUploadService`.
>   - Register resolver and `PermissionsModule` in `UserWorkspaceModule`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 359652f8c94d093a69469969874131e525d3dc6f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #16812, opened 2025-12-25
Allow user to update profile picture

AVtheking:
Fixes [#16805](https://github.com/twentyhq/twenty/issues/16805)

User was not able to update his own profile picture it showed permisson error while doing so.

Updated permission so that user is able to edit profile picture

---

> [!NOTE]
> Ensures profile picture uploads succeed only with appropriate permissions and clearer errors.
> 
> - Tightens `UploadProfilePicturePermissionGuard` to handle missing `workspace`, allow during workspace creation, and permit uploads when user has `WORKSPACE_MEMBERS` or `PROFILE_INFORMATION` settings; otherwise throws a `PermissionsException` with a user-friendly message
> - In `user-workspace.resolver.ts`, validates the upload result and throws an error if no files were returned
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3766bac15e340bdd467de3a9d923dbd697aab3db. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


FelixMalfait:
@mabdullahabaid can you please check this? I thought it was using the other resolver / you had split the upload endpoint into 2 resolvers. 

@AVtheking thank you! That looks good but I want to check @mabdullahabaid first

mabdullahabaid:
@FelixMalfait [this](https://github.com/twentyhq/twenty/pull/16031/commits/542946405b95fec5ed458ccb204b58b3c799da86) commit tried to unify again, leading to the bug (cursor reported it too 😅).

Currently, the `uploadProfilePicture` mutation is unused since `ProfilePictureUploader.tsx` was removed in the referenced commit. There are two ways to solve this - one of them is addressed by this PR, the other would be to remove the `uploadWorkspaceMemberProfilePicture` resolver altogether and rely on `uploadProfilePicture` instead as you recommended back in the day.

This would not have any impact UI-wise - workspace member page is not visible without the permission, so a "Member" cannot update profile picture of peers except for one workaround - sending a request to production API using the playground or CURL or something i.e. outside the Frontend.

I think that's not too big an issue, so I recommend we fall back to using `uploadProfilePicture`. It helps remove redundant code and if we need to refactor again, we can when we make the users table redundant.

FelixMalfait:
Thanks @mabdullahabaid @AVtheking, I will take over creating a custom guard for the 2 use-cases then

FelixMalfait (approved):

PR #16815, opened 2025-12-26
Improve userFriendlyMessage devX

ijreilly:
Two challenges with error messages
- always provide a useful/meaningful error message for the end user instead of the generic one. eg: show "Wrong password" and not "An error occured"
- avoid technical details unless error regards a technical feature. eg: show "An error occured" and not "Invalid post-hook payload."; but do show "Invalid issuer URL." as it occurs while configuring SSO

What this PR does
- Make userFriendlyMessage mandatory for widely used GraphqlQueryRunnerException and CommonQueryRunnerException, so that developers are forced to ask themselves what the error message should be, and as it contains very wide error codes (eg: "Bad request") which should not be mapped to just one default message
- Keep userFriendlyMessage optional for service-specific exceptions (eg: workflowStepExecutorException), but convert the error code to userFriendlyMessage mapper to a switch case function with a typecheck ensuring that all codes are mapped to a message. These default messages are still overridable where they are thrown.


packages/twenty-server/src/engine/api/rest/input-request-parsers/depth-parser-utils/parse-depth-rest-request.util.ts
```diff
@@ -22,6 +23,7 @@ export const parseDepthRestRequest = (request: AuthenticatedRequest): Depth => {
         ', ',
       )}`,
       RestInputRequestParserExceptionCode.INVALID_DEPTH_QUERY_PARAM,
+      { userFriendlyMessage: STANDARD_ERROR_MESSAGE },
```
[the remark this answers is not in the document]
ijreilly: when using the api, the user see both the original error message and the userFriendlyMessage, so no need to duplicate it in the userFriendlyMessage. Also, this keeps the userFriendlyMessage non-technical which could otherwiser be misleading for developers

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/validator-utils/validate-date-and-date-time-field-or-throw.util.ts
```diff
@@ -23,5 +24,6 @@ export const validateDateAndDateTimeFieldOrThrow = (
   throw new CommonQueryRunnerException(
     `Invalid value ${inspect(value)} for date or date-time field "${fieldName}"`,
     CommonQueryRunnerExceptionCode.INVALID_ARGS_DATA,
+    { userFriendlyMessage: msg`Invalid value for date.` },
```
FelixMalfait: Why make a message that's less helpful to users? If there are several fields of the same type it's useful to let the user know why value is wrong or at least which field?

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/validator-utils/validate-emails-field-or-throw.util.ts
```diff
@@ -31,6 +32,7 @@ export const validateEmailsFieldOrThrow = (
         throw new CommonQueryRunnerException(
           `Invalid subfield ${subField} for emails field "${fieldName}"`,
           CommonQueryRunnerExceptionCode.INVALID_ARGS_DATA,
+          { userFriendlyMessage: msg`Invalid value for emails.` },
```
FelixMalfait, line 35: Not linked to your changes but why wasn't this caught in the first place? I thought that with my changes the userFriendly message mandatory but they did not? Thanks!

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/validator-utils/validate-phones-field-or-throw.util.ts
```diff
@@ -35,6 +36,7 @@ export const validatePhonesFieldOrThrow = (
         throw new CommonQueryRunnerException(
           `Invalid subfield ${subField} for phones field "${fieldName}"`,
           CommonQueryRunnerExceptionCode.INVALID_ARGS_DATA,
+          { userFriendlyMessage: msg`Invalid value for phones.` },
```
FelixMalfait, line 39: (won't comment every instance but same comment as above, I feel like keeping the field name and/or value is important to guide users)

FelixMalfait: For me the more details the better (field + value), but if you disagree and want to do something else that's fine with me too

FelixMalfait (approved):
Thanks for tackling this! Made a few comments but approving so you can merge when you feel it's resolved

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/validator-utils/validate-emails-field-or-throw.util.ts
ijreilly: Your changes did make it mandatory to map every code with a userFriendlyMessage, but not to add a userFriendlyMessage to every error thrown (see changes in file common-query-runner.exception.ts ), so the userFriendlyMessage was going to be the same for all INVALID_ARGS_DATA, "Invalid data provided." in your implementation. I think for CommonQueryRunnerException, we can make the userFriendlyMessage mandatory in the devX (as it is done here), since codes can group together a large variety of errors (eg: INVALID_ARGS_DATA here or BAD_REQUEST).

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/validator-utils/validate-phones-field-or-throw.util.ts
ijreilly: I think I feared that it would get technical as fieldName is not the same as fieldLabel. 
Also looking closer at this specific case here, this error is thrown when an unexpected subField is in the args, which I don't think users will be able to do anything about. So not sure what's the best to do here

ijreilly: But in other cases I agree we should still point the value in the error message

ijreilly: I left it as is as there is no value to communicate on and I think "subfield" is too technical
Anyway if it happens on the product there won't be able to do anything I think. 
I checked and in the other validation files we do display the problematic value in the message, eg for text field we have
```
    throw new CommonQueryRunnerException(
      `Invalid string value ${inspectedValue} for text field "${fieldName}"`,
      CommonQueryRunnerExceptionCode.INVALID_ARGS_DATA,
      { userFriendlyMessage: msg`Invalid value: "${inspectedValue}"` },
    );
```

Edit: its missing in some (date, number...) so updating them to add the value to the error message

PR #16818, opened 2025-12-26
[Fix] Fix NaN position in notes resulting in not being able to create notes

ijreilly:
Following [discord thread](https://discord.com/channels/1130383047699738754/1453910755387899996/1453910755387899996), reproductible on twenty-eng

https://github.com/user-attachments/assets/ae7f363d-87e1-44fa-8fe2-ee78412d62a7

Records positions are computed at record creation, depending on the position arg from the request, being equal to `last`, `first`, or not present. 
When being equal to last, as it is done when creating a note from the product, the position is calculated using `.maximum()` function which uses postgres' MAX function. If there is a `NaN` value among the list, the MAX will return `NaN` too. So if for some reason there is a NaN somewhere in the position column, all subsequent records being created with last position argument will be created with NaN value. Until [this PR](https://github.com/twentyhq/twenty/pull/16630), where we introduced a validation on position at record creation which throws when NaN is trying to be introduced as a position, this was going silent. (fyi @etiennejouan , not on you at all but for info)

Looking into twenty-eng workspace, I found note records with NaN position dating back to august 2025, making it hard to understand and debug why they were introduced with NaN position. So I did not find the real root cause, but I suggest to
- update record-position.service to fix the issue for subsequent records that go through this service (which is what is currently broken)
- run a command to fix the existing records with NaN position for Notes, as it is where the issue happened for both the user reporting the issue on discord, and us on twenty-eng. So hopefully the problem was limited to Notes

### DIFF
--- a/packages/twenty-server/src/database/commands/upgrade-version-command/1-15/1-15-fix-nan-position-values-in-notes.command.ts
+++ b/packages/twenty-server/src/database/commands/upgrade-version-command/1-15/1-15-fix-nan-position-values-in-notes.command.ts
@@ -10,11 +10,10 @@ import { RunOnWorkspaceArgs } from 'src/database/commands/command-runners/worksp
 import { WorkspaceEntity } from 'src/engine/core-modules/workspace/workspace.entity';
 import { DataSourceService } from 'src/engine/metadata-modules/data-source/data-source.service';
 import { GlobalWorkspaceOrmManager } from 'src/engine/twenty-orm/global-workspace-datasource/global-workspace-orm.manager';
-import { getWorkspaceSchemaName } from 'src/engine/workspace-datasource/utils/get-workspace-schema-name.util';
 
 @Command({
   name: 'upgrade:1-15:fix-nan-position-values-in-notes',
-  description: 'Fix NaN position values in notes by replacing them with 0',
+  description: 'Fix NaN position values in notes by replacing them with 2',
 })
 export class FixNanPositionValuesInNotesCommand extends ActiveOrSuspendedWorkspacesMigrationCommandRunner {
   protected readonly logger = new Logger(
@@ -43,22 +42,21 @@ export class FixNanPositionValuesInNotesCommand extends ActiveOrSuspendedWorkspa
       );
     }
 
-    const schemaName = getWorkspaceSchemaName(workspaceId);
-
     if (isDryRun) {
       this.logger.log('Dry run mode: No changes will be applied');
     }
 
     try {
       // Count NaN position values in notes
-      const countResult = await dataSource.query(
-        `SELECT COUNT(*) as count FROM "${schemaName}"."note" WHERE "position" = 'NaN'::float8`,
-        [],
-        undefined,
+      const noteRepository = await this.globalWorkspaceOrmManager.getRepository(
+        workspaceId,
+        'note',
         { shouldBypassPermissionChecks: true },
       );
 
-      const nanCount = parseInt(countResult[0]?.count || '0', 10);
+      const nanCount = await noteRepository.count({
+        where: { position: 'NaN' },
+      });
 
       this.logger.log(
         `Found ${nanCount} note(s) with NaN position values in workspace ${workspaceId}`,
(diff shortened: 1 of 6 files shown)


packages/twenty-server/src/engine/core-modules/record-position/services/record-position.service.ts
```diff
@@ -26,7 +26,7 @@ export class RecordPositionService {
     workspaceId,
     index = 0,
   }: RecordPositionServiceCreateArgs): Promise<number> {
-    if (typeof value === 'number') {
+    if (isNumber(value) && !Number.isNaN(value)) {
```
Weiko, line 30: nit: we could introduce a util/typeguard, up to you

packages/twenty-server/src/engine/core-modules/record-position/services/record-position.service.ts
```diff
@@ -230,19 +238,30 @@ export class RecordPositionService {
...
+        },
+      );
 
-        return await repository.maximum('position');
-      },
-    );
+    return this.sanitizePositionValue(result);
+  }
+
+  private sanitizePositionValue(value: number | null): number | null {
```
Weiko: nit: could be a sanitizeNumber util as mentioned earlier

packages/twenty-server/src/database/commands/upgrade-version-command/1-15/1-15-fix-nan-position-values-in-notes.command.ts
```diff
@@ -0,0 +1,97 @@
...
+    const schemaName = getWorkspaceSchemaName(workspaceId);
+
+    if (isDryRun) {
+      this.logger.log('Dry run mode: No changes will be applied');
+    }
+
+    try {
+      // Count NaN position values in notes
+      const countResult = await dataSource.query(
+        `SELECT COUNT(*) as count FROM "${schemaName}"."note" WHERE "position" = 'NaN'::float8`,
```
Weiko: I'm not sure but is the cast to float necessary here?

packages/twenty-server/src/database/commands/upgrade-version-command/1-15/1-15-fix-nan-position-values-in-notes.command.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    const schemaName = getWorkspaceSchemaName(workspaceId);
+
+    if (isDryRun) {
+      this.logger.log('Dry run mode: No changes will be applied');
+    }
+
+    try {
+      // Count NaN position values in notes
+      const countResult = await dataSource.query(
```
Weiko: nit: why using raw queries here? Not a strong blocker, just curious

packages/twenty-server/src/database/commands/upgrade-version-command/1-15/1-15-fix-nan-position-values-in-notes.command.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  override async runOnWorkspace({
+    workspaceId,
+    options,
+    dataSource,
+  }: RunOnWorkspaceArgs): Promise<void> {
+    const isDryRun = options.dryRun || false;
+
+    if (!isDefined(dataSource)) {
```
Weiko, line 39: note: could this really be undefined? I think we should throw in ActiveOrSuspendedWorkspacesMigrationCommandRunner then 🤔 
No need to change in this PR, just a thought

packages/twenty-server/src/database/commands/upgrade-version-command/1-15/1-15-fix-nan-position-values-in-notes.command.ts
```diff
@@ -0,0 +1,97 @@
...
+import { WorkspaceEntity } from 'src/engine/core-modules/workspace/workspace.entity';
+import { DataSourceService } from 'src/engine/metadata-modules/data-source/data-source.service';
+import { GlobalWorkspaceOrmManager } from 'src/engine/twenty-orm/global-workspace-datasource/global-workspace-orm.manager';
+import { getWorkspaceSchemaName } from 'src/engine/workspace-datasource/utils/get-workspace-schema-name.util';
+
+@Command({
+  name: 'upgrade:1-15:fix-nan-position-values-in-notes',
+  description: 'Fix NaN position values in notes by replacing them with 2',
+})
+export class FixNanPositionValuesInNotesCommand extends ActiveOrSuspendedWorkspacesMigrationCommandRunner {
```
Weiko, line 18: Could we have the same issue on other objects?

Weiko (approved):
Left some nit comments. Strategy LGTM @ijreilly

packages/twenty-server/src/database/commands/upgrade-version-command/1-15/1-15-fix-nan-position-values-in-notes.command.ts
ijreilly: Yes we could as I don't know what introduced the NaN positions in the first place (date back to at least august from my investigation). But the case that was reported by a user was on note and I was also able to reproduce on Notes. I think if it was on another object we would have had more complaints as it results in impossibility to create a record at all

packages/twenty-server/src/database/commands/upgrade-version-command/1-15/1-15-fix-nan-position-values-in-notes.command.ts
ijreilly: I changed it to use typeorm!

PR #16822, opened 2025-12-26
feat: use side-column variant for all widgets on mobile and side panel

### LINKED ISSUE #1957: Use side-column variant for all widgets on mobile and side panel
The `getWidgetCardVariant` function returns the variant to use to style widget cards.

We have a bunch of different widget variants:

Image

## Currently

The `side-column` is only used when displaying widgets of the pinned tab.

## Expected

We shouldn't modify the rendering conditions of the `canvas` and `dashboard` variants.

However, the `side-column` variant should be preferred to `record-page` when viewing the page on a mobile or inside the side panel.

**We must update stories accordingly. I think we should have one story to assert we use the `side-column` variant on mobile and another for side panel. Feel free to create as many stories as you find necessary.**

mabdullahabaid:
Closes [1957](https://github.com/twentyhq/core-team-issues/issues/1957).

`getWidgetCardVariant` now returns `side-panel` for mobile and right drawer instead of returning `record-page`.

Added stories to WidgetRenderer.stories.tsx because WidgetRenderer is where getWidgetCardVariant is called and the variant is applied. These stories test the actual behavior (mobile/side panel triggering side-column variant) rather than just visual appearance, following the existing pattern of individual behavior-testing stories in this file.


Devessier (approved):
Great job 👍

PR #16833, opened 2025-12-29
feat: Add If/Else node

### LINKED ISSUE #1265: 🔀 If/Else Node
## Description:
If/else node makes sure that workflow always continues with the path changing depending on the condition.
- Default behavior: If/else nodes split into two paths → If (true) or Else (false).
- Else if: Add more routes; conditions are checked in order, and the first match is taken.
- Else branch will always exist and takes care of all the other cases not specified (does not need to be configured).
- At runtime, the flow will follow only one path.

Key difference from filters:
1) Filters: Gatekeepers meaning workflow may stop if the condition fails.
2) If/else: Always continues, just down different paths.

Figma file: https://www.figma.com/design/xt8O9mFeLl46C5InWwoMrN/Twenty?node-id=76218-40586&t=u6a2zdcUZjY6ziCO-1

Default state when if/else node is added:
Image

When additional else if route is added:
Image

### Frontend - Step content

The frontend structure will be similar to the Filter Action. Check component `WorkflowEditActionFilter`
For filters we have:
- an array of `filter groups`. Check `StepFilters.ts`
```
export type StepFilterGroup = {
  id: string;
  logicalOperator: StepLogicalOperator;
  parentStepFilterGroupId?: string;
  positionInStepFilterGroup?: number;
};
```
- an array of `filters`. Each filters being linked to a group.

We will re-use the structure here. We should simply add a new one:
```
type StepIfElseBranch = {
  id: string;
  nextStepIds: string[];
  filterGroupId?: string;  // ONE group per branch. ELSE branch = undefined
};
```
The filter group position will determine the if/else operand. First one is IF, middles are ELSE_IF, last one is ELSE.

### Backend - Step content

If-Else should be a similar logic to `FilterWorkflowAction`. Main difference is:
- filter action executes all groups and filters at once and returns a boolean `matchesFilter``
- If-Else will execute its branches - and the associated filters - one by one. Once a branche returns matchesFilter, the step will return the nextStepIds associated to the branch.
- ELSE branch handling: The ELSE branch should have no filters (or empty filter groups). When no previous branch matches, the ELSE branch is selected by default. 
- the result of the step will be `matchingBranchId`

### Backend - Routing

Since the matchingBranchId will be stored in step result, the workflow executor will be able to retrieve the next step ids to execute. 
We simply need to update `getNextStepIdsToExecute`. If the executed step is IfElse, then we look for the `matchingBranchId` in result. Then, in the settings, retrieve the nextStepIds linked to that branch.

```
async getNextStepIdsToExecute({
  executedStep,
  executedStepResult,
}: {
  executedStep: WorkflowAction;
  executedStepResult: WorkflowActionOutput;
}): Promise<string[] | undefined> {
  // ... existing iterator logic ...

  const isIfElseStep = isWorkflowIfElseAction(executedStep);

  if (isIfElseStep) {
    const ifElseResult = executedStepResult.result as
      | WorkflowIfElseResult
      | undefined;

    if (ifElseResult?.matchingBranchId) {
      const matchingBranch = executedStep.settings.input.branches.find(
        (branch) => branch.id === ifElseResult.matchingBranchId
      );
      return matchingBranch?.nextStepIds;
    }
  }

  return executedStep.nextStepIds;
}
```

### Frontend - Routing

This may be the trickiest part. Behavior:
- Single handle at bottom for all branch edges
- No step-level nextStepIds — routing lives in branches[].nextStepIds
- On creation: Auto-creates IF + ELSE branches, each with an empty node
- IF/ELSE branches: Mandatory, cannot be deleted
- ELSE-IF branches: Added/deleted via step settings UI, creates empty node on add, orphans steps on delete
- Edges: Non-editable (not deletable, movable, or reconnectable) — controlled entirely by branch settings
- Labels: if / else if / else

What should be updated:

1. Diagram Generation. Create `generateNodesAndEdgesForIfElseNode.ts`. Loop through branches, create one edge per `branch.nextStepIds` entry. Set `deletable: false`, `selectable: false`, `reconnectable: false`. Add label based on branch position (first = "if", last = "else", middle = "else if")

2. Node Creation (Backend). When creating If/Else step, auto-create 2 empty nodes - see how this is done for iterators. Initialize with IF branch with filterGroupId + ELSE branch without filterGroupId. Return `additionalCreatedSteps` like Iterator does in `runStepCreationSideEffectsAndBuildStep`

3. Branch Management (Frontend UI)
- Add ELSE-IF: Insert branch before ELSE, create empty node, shift positions
- Delete ELSE-IF: Remove branch from array, orphan connected steps
- No edge creation/deletion APIs used — all changes go through step settings update

4. No changes to edge hooks
- useCreateEdge / useDeleteEdge not used for If/Else edges
- Edge existence derived from branch settings at render time

Not sure about the full start. This is a global idea.

abdulrahmancodes:
Closes [#1265](https://github.com/twentyhq/core-team-issues/issues/1265)

---

> [!NOTE]
> Enables conditional branching in workflows via a new `IF_ELSE` action, guarded by `IS_IF_ELSE_ENABLED`, spanning UI, builder, executor, and shared schemas.
> 
> - Adds feature flag `IS_IF_ELSE_ENABLED` and GraphQL/enum updates; seeded and exposed in public flags
> - Frontend: shows `IF_ELSE` in Flow actions when flag enabled; new editor (`WorkflowEditActionIfElse*`) to manage branches and conditions; workflow diagram generation for `IF_ELSE` with labeled edges; centers edge labels and hides edge controls when `deletable` is false; step create/delete logic handles `IF_ELSE` and its empty child nodes; refactors filter UI into `workflow-steps/filters/*`
> - Backend: builder creates `IF_ELSE` with initial step filters, branches, and two positioned `EMPTY` child steps; executor adds `IfElseWorkflowAction` that evaluates filters and routes to matching branch; factory/module wiring and next-step resolution updated
> - Shared: introduces `StepIfElseBranch`, `IF_ELSE_BRANCH_POSITION_OFFSETS`, and schemas for `if-else` and step filters; updates `workflowActionSchema` exports and types
> - Minor: action icon/color support; new hook export (`updateWorkflowVersionPosition`); storybook paths updated
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 1ea3277a72bc1e5aa02897a42e257c28dd46d16a. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


FelixMalfait:
Great work!!!
- When we remove the "If/else" node, can we remove any node that is still in "Add action" ; it's strange to leave them behind. Maybe that's why I got the error below 
Screenshot 2025-12-31 at 14 46 17
- Let's remove the "Else section" since it doesn't appear in Figma?

FelixMalfait (commented):
I added a test for the Quick Lead workflow. Could you add one or two workflows to be tested that contain these conditions in a followup PR? Since it isn't seeded like Quick Lead you will also have to take care of workflow creation in the PR.

FelixMalfait (approved):
Really great work! Please add tests next (check quick-lead-workflow.integration-spec.ts)

packages/twenty-front/src/modules/workflow/workflow-steps/hooks/useDeleteStep.ts
```diff
@@ -25,11 +27,35 @@ export const useDeleteStep = () => {
...
-      );
+    const steps = workflow?.currentVersion?.steps;
+    const stepToDelete = isDefined(steps)
+      ? steps.find((step) => step.id === stepId)
+      : undefined;
+
+    if (
+      isDefined(stepToDelete) &&
+      isDefined(steps) &&
+      stepToDelete.type === 'IF_ELSE'
```
thomtrp, line 38: This hook should remain generic. For both if-Else and AI agent, we should have a scalable behavior done on backend side. You can check the `removeStep` util

packages/twenty-front/src/modules/workflow/workflow-steps/filters/components/WorkflowStepFilterColumn.tsx
```diff
@@ -1,18 +1,24 @@
...
 import styled from '@emotion/styled';
 import { useContext } from 'react';
 import { type StepFilter, type StepFilterGroup } from 'twenty-shared/types';
+import { isDefined } from 'twenty-shared/utils';
 
 type WorkflowStepFilterColumnProps = {
   stepFilterGroup: StepFilterGroup;
   stepFilter: StepFilter;
   stepFilterIndex: number;
+  isIfBranch?: boolean;
```
thomtrp, line 20: we should not see if-else related informations here. The readonly, provided by WorkflowIfElseBranchEditor, should included the logic of isLastFilterInIfBranch.

thomtrp (commented):
Great work @abdulrahmancodes! A few comments but looks nice already 🔥

thomtrp:
@Bonapara this is hard to understand what else-if filter matches what branch. Are you ok if:
- we use else-if filter when there is only one branch
- we use else-if 1 / else-if 2 / ... labels when there are more than 1 else-if branch?

Capture d’écran 2026-01-05 à 17 51 53

Bonapara:
CleanShot 2026-01-06 at 18 11 01@2x

@thomtrp we can add smth like that

https://www.figma.com/design/xt8O9mFeLl46C5InWwoMrN/Twenty?node-id=79460-312740&t=98UPRHyR6pPfzMcJ-11

New issue?

thomtrp:
I think it can remain the same since this is still in lab. Thanks for the design! @abdulrahmancodes could you have a look? 🙏

PR #16838, opened 2025-12-29
Fix E2E tests + Skip chromatic 

ijreilly:
- Always skip chromatic job (we dont check the outcome)
- Fix workflow creation test
- Run E2E tests before merge (to enable through repo rulesets)


Weiko (approved):
Ok!

PR #16841, opened 2025-12-29
prevents the click event from bubbling up to parent elements that mig…

Murkyskate:
…ht have their own click handlers

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-table/record-table-cell/components/RecordTableCellCheckbox.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-table/record-table-cell/components/RecordTableCellCheckbox.tsx
@@ -33,7 +33,6 @@ export const RecordTableCellCheckbox = () => {
 
   const handleClick = useCallback(
     (event?: React.MouseEvent<HTMLDivElement>) => {
-      event?.stopPropagation();
       setCurrentRowSelected({
         newSelectedState: !isSelected,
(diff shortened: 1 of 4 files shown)


packages/twenty-front/src/modules/object-record/record-table/record-table-cell/components/RecordTableCellCheckbox.tsx
```diff
@@ -33,6 +33,7 @@ export const RecordTableCellCheckbox = () => {
 
   const handleClick = useCallback(
     (event?: React.MouseEvent<HTMLDivElement>) => {
+      event?.stopPropagation();
```
charlesBochet: using stopPropagation is usually not a good idea as this is very sensible to dom changes. We prefer using the notion of focusStack to understand what element should have the focus

Murkyskate:
Hi @charlesBochet , I have updated all table components to use data-select-disable="true" removing event?.stopPropagation()

charlesBochet:
@Murkyskate I'm very sorry for the slow review over the holidays!

I'm actually quite confused, what problem are we solving here? The issue commented is about an issue with record merge which seems unrelated.

Could you link the right issue or explain what we are fixing here (with screen recording of the issue, it would be ideal)

I'm going to close this PR in the meantime, but feel free to comment and I'll re-open it (or you can re-open it your self by adding a comment: "/twenty pr open")

PR #16844, opened 2025-12-29
Redesign Object/Field tables

FelixMalfait:
Redesign the data model pages for more clarity / better distinction between fields and relations

### DIFF
--- a/packages/twenty-front/src/modules/object-record/read-only/utils/__tests__/isObjectMetadataSettingsReadOnly.test.ts
+++ b/packages/twenty-front/src/modules/object-record/read-only/utils/__tests__/isObjectMetadataSettingsReadOnly.test.ts
@@ -12,13 +12,12 @@ describe('isObjectMetadataSettingsReadOnly', () => {
         isUIReadOnly: false,
         isRemote: false,
       },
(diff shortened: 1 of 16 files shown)


FelixMalfait:
@mabdullahabaid could you please review/test? Thanks!

Feel free to merge if it looks good to you / to take over for small changes

mabdullahabaid:
Looks good to me! I really like the separation between relations and fields.

We can now create custom fields for both standard and system objects. Removing the `applicationId` dependency allows objects from installed apps to be editable, while the readonly flag safeguards the structural integrity.

There is one issue I can identify: I can add fields to system objects, but when I navigate to records page for the system object and then try to edit the field value, I get an error. It's not something introduced by this PR, so we're good in terms of the changes made in this PR.

https://github.com/user-attachments/assets/cf0cf785-aa8d-4444-af73-48c68c49a6cb

FelixMalfait:
@mabdullahabaid sorry I didn't answer to this. Could you please open a new issue for that bug? In your screen you don't show the network tab, what the actual error is (maybe for example i could be the "new metadata / refresh page" error message and then there are two issues: the fact that it displays a generic error, and the fact that it's isn't refreshed automatically). Thanks!

PR #16857, opened 2025-12-30
feat(workflow): use authContext in CRUD services for Common API migration

FelixMalfait:
## Summary

This PR migrates workflow CRUD operations to properly use the Common API layer's authentication context, addressing the issues from the reverted PR #15875.

The original PR was reverted because the Common API required passing either a User or an API Key for authentication, which was problematic for workflows. Since then, the "Application" concept was introduced in the Common API layer, allowing for token injection in serverless functions.

This PR leverages the "Twenty Standard Application" concept for non-manual workflow triggers, providing a clean authentication path without the issues of user impersonation.

### Core Infrastructure
- **RecordCrudExecutionContext**: Replace `workspaceId` with full `authContext`
- **WorkflowExecutionContext**: Add `authContext` field to carry authentication info
- **ToolGeneratorContext/ToolSpecification**: Add optional `authContext` support for tool generation

### Authentication Flow
- **WorkflowExecutionContextService**: Build appropriate auth context based on trigger type:
  - **Manual triggers**: Use user's workspace auth context with their role permissions
  - **Non-manual triggers**: Use Twenty Standard Application auth context (bypasses permission checks or uses default serverless function role)
- **ApplicationService**: Add `findTwentyStandardApplicationOrThrow` method to retrieve the system application
- **UserWorkspaceService**: Make relations configurable in `getUserWorkspaceForUserOrThrow` to load only what's needed

### CRUD Services Migration
All 5 record CRUD services now receive `authContext` instead of `workspaceId`:
- `CreateRecordService`
- `UpdateRecordService`
- `DeleteRecordService`
- `FindRecordsService`
- `UpsertRecordService`

### Workflow Actions
All record CRUD workflow actions pass `executionContext.authContext` to the services:
- `CreateRecordWorkflowAction`
- `UpdateRecordWorkflowAction`
- `DeleteRecordWorkflowAction`
- `FindRecordsWorkflowAction`
- `UpsertRecordWorkflowAction`

### AI Agent Integration
- AI agent workflow action passes auth context to agent executor
- Tool provider and MCP protocol service support auth context propagation

## Benefits
- ✅ Proper authentication for workflow CRUD operations via Common API
- ✅ Non-manual triggers use system application context (no user impersonation issues)
- ✅ Manual triggers preserve user permissions correctly
- ✅ Foundation for better permission handling in automated workflows
- ✅ Cleaner separation between user-initiated and system-initiated operations

## Related
- Reverted PR: #15875


packages/twenty-server/src/engine/api/mcp/services/mcp-protocol.service.ts
```diff
@@ -97,6 +99,43 @@ export class McpProtocolService {
     return roleId;
   }
 
+  private buildAuthContext(
```
Weiko, line 90: I'm surprised we don't have that logic somewhere else, should we create a util for that?

packages/twenty-server/src/engine/api/mcp/services/mcp-protocol.service.ts
```diff
@@ -97,6 +99,43 @@ export class McpProtocolService {
...
+    workspace: WorkspaceEntity,
+    userWorkspaceId?: string,
+    apiKey?: ApiKeyEntity,
+  ): WorkspaceAuthContext {
+    if (isDefined(apiKey)) {
```
Weiko: I feel like we could simply return
```typescript
  private buildAuthContext(
    workspace: WorkspaceEntity,
    userWorkspaceId?: string,
    apiKey?: ApiKeyEntity,
  ): WorkspaceAuthContext {
    return {
      user: null,
      apiKey: apiKey ?? null,
      application: null,
      workspace,
      workspaceMemberId: undefined,
      userWorkspaceId: userWorkspaceId ?? undefined,
    };
  }
```
the conditions don't bring much or am I missing something? Do you think we could have both api and userWorkspace and you want to make sure your object only contains apiKey if it's there?

packages/twenty-server/src/engine/core-modules/application/application.service.ts
```diff
@@ -146,6 +146,38 @@ export class ApplicationService {
     });
   }
 
+  async findTwentyStandardApplicationOrThrow(workspaceId: string): Promise<{
```
Weiko, line 149: Why not using findWorkspaceTwentyStandardAndCustomApplicationOrThrow?

packages/twenty-server/src/engine/core-modules/application/application.service.ts
```diff
@@ -146,6 +146,38 @@ export class ApplicationService {
...
+    });
+
+    if (!isDefined(workspace)) {
+      throw new ApplicationException(
+        `Could not find workspace ${workspaceId}`,
+        ApplicationExceptionCode.APPLICATION_NOT_FOUND,
+      );
+    }
+
+    const application = await this.applicationRepository.findOne({
```
Weiko: We prefer to use the cache when it's available. (this will go through the local (pod) cache instead of querying the DB). Tradeoff here is you'll have to filter yourself and not use pg filter (which is most likely more efficient) but we are talking about applications here so probably 2-3 max. Up to you!

packages/twenty-server/src/modules/workflow/workflow-executor/services/workflow-execution-context.service.ts
```diff
@@ -30,35 +34,88 @@ export class WorkflowExecutionContextService {
...
-    const rolePermissionConfig = roleId
-      ? { unionOf: [roleId] }
+    return this.buildApplicationExecutionContext(
+      workflowRun,
+      runInfo.workspaceId,
+    );
+  }
+
+  private async buildUserExecutionContext(
+    workflowRun: Awaited<
```
Weiko: Why not using the return type directly? Probably
```typescript
workflowRun: WorkflowRunWorkspaceEntity
```

packages/twenty-server/src/modules/workflow/workflow-executor/services/workflow-execution-context.service.ts
```diff
@@ -30,35 +34,88 @@ export class WorkflowExecutionContextService {
...
+    return {
+      isActingOnBehalfOfUser: true,
+      initiator: workflowRun.createdBy,
+      rolePermissionConfig: { unionOf: [roleId] },
+      authContext,
+    };
+  }
+
+  private async buildApplicationExecutionContext(
+    workflowRun: Awaited<
```
Weiko: same

Weiko (commented):
Left some comments! 👍 @FelixMalfait

PR #16858, opened 2025-12-30
Create query runner context from request context instead of schema builder internal context

Weiko:
## Context
This PR refactors how authentication context is handled in the GraphQL resolver layer. Previously, the auth context was baked into the schema builder context at schema creation time. Now, the auth context is extracted from the request at query execution time, enabling better schema caching.

## Implementation
- Removed user-specific data from schema cache key: The schema cache key no longer includes userId and apiKeyId, allowing the same schema to be shared across all users in a workspace
- Cache key simplified from ${workspaceId}-${userId}-${apiKeyId}-${url}-${cacheVersion} to ${workspaceId}-${url}-${cacheVersion}
- Removed authContext from WorkspaceSchemaBuilderContext: The schema builder context is now purely about object/field metadata, not about who is accessing it
- Created createQueryRunnerContext utility: A new helper that combines the schema builder context with the auth context from the actual request
- Updated all resolver factories: All 15 resolver factories now use createQueryRunnerContext to build the full context needed by query runners at execution time

## Benefits
- Improved cache efficiency: GraphQL schemas are now cached per workspace rather than per user, significantly reducing memory usage and schema regeneration
- Cleaner separation of concerns: Schema building (metadata-driven) is now separate from query execution (auth-driven)
- Fresh auth context: Auth context is always retrieved from the current request, ensuring it reflects the latest state
- Reduced complexity: Simplified the schema creation path by removing unnecessary auth checks

## Next steps
- Introduce a hash in redis and expose it in the req object so the yoga patch can access the hash and use it during its own cache key computation. This way we will be able to invalidate the local cache in yoga graphql schema generation without restarting pods.


packages/twenty-server/src/engine/api/graphql/graphql-config/graphql-config.service.ts
```diff
@@ -161,15 +144,15 @@ export class GraphQLConfigService
           throw handleExceptionAndConvertToGraphQLError(
             error,
             this.exceptionHandlerService,
-            user
```
Weiko, line 164: There was a bug here, user was never defined (only in the "try" block). I've moved the context extraction outside of it

FelixMalfait (approved):

PR #16863, opened 2025-12-30
`[DO_NOT_RELEASE_MAIN_UNTIL_MERGED]` Prevent migration failure due to workspace orphan metadata rows

prastoin:
# Introduction
The `AddWorkspaceForeignKeys1767002571103` migration would fail when released in production right now, as `foreignKey` be applicable as there's a lof of orphan entries in database

As a workaround in order not to block any patch release we're fallbacking the migration using save point and an upgrade command that will attempt to apply the `foreignKey` on every workspace upgrade until it succeed

We should keep in mind that any new fresh self installation will have the foreignKey double checked that it would not implies regression on workspace deletion using the integration tests

## Cleaning upgrade command
We won't implement the cleaning command in this PR yet either will I as discussed with @Weiko someone else might be taking the subject starting next week

---

> [!NOTE]
> Strengthens workspace data integrity and makes the FK migration resilient.
> 
> - Adds `upgrade:1-16:add-workspace-foreign-keys-migration` command to apply `workspaceId` FKs once per run; wires into `V1_16_UpgradeVersionCommandModule` and 1.16 upgrade sequence
> - Refactors migration `1767002571103` to use `addWorkspaceForeignKeysQueries` util and wrap in a savepoint, swallowing errors to avoid blocking releases
> - Extracts FK DDL into `utils/1767002571103-addWorkspaceForeignKeys.util` for reuse by command and migration
> - Removes duplicate `workspaceId` columns from entities (e.g., `cronTrigger`, `databaseEventTrigger`, `indexMetadata`, `objectMetadata`, `roleTarget`, `role`, `serverlessFunction`) relying on `SyncableEntity`; keeps indexes/relations
> - Marks legacy delete paths as deprecated; temporarily extends `WorkspaceManagerService.delete` to also delete `serverlessFunction` by `workspaceId`
> - Updates wiring to inject `ServerlessFunctionEntity` repository in `workspace-manager` module/service and corresponding unit test
> - Extends integration tests and adds GraphQL helpers to create serverless functions and triggers; verifies cascade deletion of related metadata on workspace removal
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6805bf5d1b32828b4bb1e9f130bfe6e478f66aee. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/twenty-server/src/database/commands/upgrade-version-command/1-16/1-16-add-workspace-foreign-keys-migration.command.ts
```diff
@@ -0,0 +1,67 @@
...
+
+      return;
+    }
+
+    const queryRunner = this.coreDataSource.createQueryRunner();
+
+    await queryRunner.connect();
+    await queryRunner.startTransaction();
+
+    if (!options.dryRun) {
```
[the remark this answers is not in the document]
prastoin: Nice catch

packages/twenty-server/src/database/commands/upgrade-version-command/1-16/1-16-add-workspace-foreign-keys-migration.command.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  constructor(
+    @InjectRepository(WorkspaceEntity)
+    protected readonly workspaceRepository: Repository<WorkspaceEntity>,
+    protected readonly twentyORMGlobalManager: GlobalWorkspaceOrmManager,
+    protected readonly dataSourceService: DataSourceService,
+    @InjectDataSource()
+    private readonly coreDataSource: DataSource,
+  ) {
+    super(workspaceRepository, twentyORMGlobalManager, dataSourceService);
```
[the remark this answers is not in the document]
prastoin: yes quite paradoxical but expected

packages/twenty-server/src/database/typeorm/core/migrations/common/1767002571103-addWorkspaceForeignKeys.ts
```diff
@@ -1,59 +1,37 @@
...
-    );
-    await queryRunner.query(
-      `ALTER TABLE "core"."workspaceMigration" ADD CONSTRAINT "FK_bebfcf8dcb299fd39ad04ed4c7f" FOREIGN KEY ("workspaceId") REFERENCES "core"."workspace"("id") ON DELETE CASCADE ON UPDATE NO ACTION`,
-    );
-    await queryRunner.query(
-      `ALTER TABLE "core"."remoteTable" ADD CONSTRAINT "FK_30b429b14ddc6aab7495fa884f3" FOREIGN KEY ("workspaceId") REFERENCES "core"."workspace"("id") ON DELETE CASCADE ON UPDATE NO ACTION`,
-    );
-    await queryRunner.query(
-      `ALTER TABLE "core"."remoteServer" ADD CONSTRAINT "FK_d1293835c5a0bb89fc0ed45713f" FOREIGN KEY ("workspaceId") REFERENCES "core"."workspace"("id") ON DELETE CASCADE ON UPDATE NO ACTION`,
-    );
```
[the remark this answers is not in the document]
prastoin: That's acceptable as we don't really handle downgrading

packages/twenty-server/src/engine/workspace-manager/workspace-manager.service.ts
```diff
@@ -182,8 +185,10 @@ export class WorkspaceManagerService {
     });
   }
 
-  // TODO investigate why some entities are not on cascade delete
-  // Are foreign keys correctly applied ?
+  /**
+   * @deprecated Should be removed once AddWorkspaceForeignKeysMigrationCommand has been run successfully in production
```
Weiko, line 189: not entirely, we still need to trigger the schema deletion @prastoin

packages/twenty-server/src/engine/core-modules/workspace/services/workspace.service.ts
```diff
@@ -276,6 +276,10 @@ export class WorkspaceService extends TypeOrmQueryService<WorkspaceEntity> {
     });
   }
 
+  /**
+   * @deprecated Should be removed once AddWorkspaceForeignKeysMigrationCommand has been run successfully in production
+   * As we will be able to rely on foreignKey delete cascading
```
Weiko, line 283: We still need to cancel the billing calling stripe. Comment could be misleading if a user delete the whole method after that

Weiko (commented):
Left some comments @prastoin

Weiko (commented):
Left some comments about your comments being misleading. Otherwise LGTM!

Weiko (approved):

packages/twenty-server/src/engine/workspace-manager/workspace-manager.service.ts
prastoin: Indeed you're right 
+1 this should be covered in the integration test too

packages/twenty-server/src/engine/core-modules/workspace/services/workspace.service.ts
prastoin: That's a critical bug right here, you're right !
this should be done even before removing the workspace
As it's now a workspaceRelationEntity
```ts
export class BillingSubscriptionEntity extends WorkspaceRelatedEntity {
```
If not nothing will be there to be cancelled

prastoin: So in the end no critical bug as the workspace is deleted after the operations
But still extracted the logic that we will have to keep out the marked as deprecated method for consistency

packages/twenty-server/src/engine/workspace-manager/__tests__/workspace-manager.service.spec.ts
```diff
@@ -3,6 +3,7 @@ import { getDataSourceToken, getRepositoryToken } from '@nestjs/typeorm';
 
 import { type DataSource, type Repository } from 'typeorm';
 
+import { describe } from 'node:test';
```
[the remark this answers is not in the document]
prastoin: fixed

Issue #16526, opened 2025-12-12
viewField records not soft-deleted when parent view is deleted


mydalat:
Description:
Problem
When deleting a view in Twenty CRM, the parent view record is soft-deleted (deletedAt is set), but the related viewField records remain active (deletedAt = NULL).
This creates orphan records in the database.
Steps to Reproduce

Create a custom view (e.g., KANBAN view for People)
Delete the view from UI
Check database:

`-- View is soft-deleted
SELECT id, name, "deletedAt" FROM core.view WHERE id = '<view_id>';
-- deletedAt = '2025-12-12 12:59:55'

-- But viewFields still active
SELECT count(*) FROM core."viewField" WHERE "viewId" = '<view_id>' AND "deletedAt" IS NULL;
-- count = 23
`

Expected Behavior

When a view is soft-deleted, all related child records (viewField, viewFilter, viewSort, viewGroup, viewFilterGroup) should also be soft-deleted.

Workaround
Manual cleanup:

`UPDATE core."viewField" 
SET "deletedAt" = NOW() 
WHERE "viewId" = '<view_id>' AND "deletedAt" IS NULL;`

Environment

Twenty version: v1.12.0
Database: PostgreSQL (twenty-postgres-spilo)

s3arthak:
I would like to work on this . Assign this to me

BOHEUS:
That's not a problem for end-user as view isn't available in UI if it's soft-deleted though now it poses few questions:
- why view is soft-deleted and not hard-deleted when user deletes view in UI?
- how to hard-delete a view in UI?
- what's the correct logic here? should related records from other tables be also soft-deleted or not? (considering that in API they'll be visible as active)

Records in core."viewField" and other linked tables are not orphaned as they're still linked to view in core."view" and they're deleted when view is permanently deleted

Weiko:
We rarely implement cascade soft deletion because this can be complex when you want to restore those records recursively (for example when you have unicity constraints that don't rely on deletedAt, the restore can be tricky).
I don't think having soft-deletion on that table brings much value though, it's probably due to its history when Views was a standard object and we simply kept the deletedAt column during the migration. The real fix would probably be to remove that behavior (cc @Bonapara wdyt?)

Bonapara:
Let's destroy the views instead of soft-deleting them, indeed. We thus don't need the deleted at column anymore.

PR #16868, opened 2025-12-30
Destroy core view

BOHEUS:
Solution for #16526

### DIFF
--- a/packages/twenty-front/src/modules/views/hooks/internal/usePersistView.ts
+++ b/packages/twenty-front/src/modules/views/hooks/internal/usePersistView.ts
@@ -11,9 +11,11 @@ import { v4 } from 'uuid';
 import {
   type CreateCoreViewMutationVariables,
   type DeleteCoreViewMutationVariables,
+  type DestroyCoreViewMutationVariables,
   type UpdateCoreViewMutationVariables,
   useCreateCoreViewMutation,
   useDeleteCoreViewMutation,
+  useDestroyCoreViewMutation,
   useUpdateCoreViewMutation,
   ViewType,
 } from '~/generated/graphql';
@@ -22,6 +24,7 @@ export const usePersistView = () => {
   const [createCoreViewMutation] = useCreateCoreViewMutation();
   const [updateCoreViewMutation] = useUpdateCoreViewMutation();
   const [deleteCoreViewMutation] = useDeleteCoreViewMutation();
+  const [destroyCoreViewMutation] = useDestroyCoreViewMutation();
   const { triggerViewGroupOptimisticEffectAtViewCreation } =
     useViewsSideEffectsOnViewGroups();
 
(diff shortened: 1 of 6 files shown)


BOHEUS:
Not sure if destroyView function in usePersistView.ts should be removed if it's not used anywhere

BOHEUS:
Also, should be a upgrade command removing all views with deletedAt != null?

packages/twenty-front/src/modules/views/view-picker/components/ViewPickerOptionDropdown.tsx
```diff
@@ -48,7 +48,7 @@ export const ViewPickerOptionDropdown = ({
   const { t } = useLingui();
   const { closeDropdown } = useCloseDropdown();
   const { getIcon } = useIcons();
-  const { deleteViewFromCurrentState } = useDeleteViewFromCurrentState();
+  const { deleteViewFromCurrentState } = useDestroyViewFromCurrentState();
```
prastoin: ***Polish:*** naming

packages/twenty-front/src/modules/views/hooks/internal/usePersistView.ts
```diff
@@ -22,6 +24,7 @@ export const usePersistView = () => {
   const [createCoreViewMutation] = useCreateCoreViewMutation();
   const [updateCoreViewMutation] = useUpdateCoreViewMutation();
   const [deleteCoreViewMutation] = useDeleteCoreViewMutation();
```
prastoin: ***Remark:*** Even wondering if we should keep the delete here

prastoin (commented):
Hey there thanks for your contribution ! 
About to push a fix for the naming

prastoin:
Even though we might end up renaming the destroy ( hard delete ) to delete ( currently soft delete ) in the future, for the moment it's still two distinct operations so lets keep naming explicit for the moment

packages/twenty-front/src/modules/views/hooks/internal/usePersistView.ts
```diff
@@ -154,12 +154,12 @@ export const usePersistView = () => {
...
+    [destroyCoreViewMutation, handleMetadataError, enqueueErrorSnackBar],
   );
 
   return {
     createView,
     updateView,
-    deleteView,
+    destroyView,
   };
 };
```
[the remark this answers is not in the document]
prastoin: expected

prastoin:
> Also, should be a upgrade command removing all views with deletedAt != null?

For the moment I think it's over-engineered, lets consider implementing such a command when we decide to totally deprecate soft deletion from now `core` view-xx entities

prastoin (approved):
Image

Weiko (approved):
LGTM

Weiko (approved):

PR #16870, opened 2025-12-30
fix: use white background for widgets on mobile and in side panel

### LINKED ISSUE #2029: [Record Page Layouts] Widgets should have a white background on mobile and in the side panel
Currently, widgets have a slightly gray background in the side panel and on mobile. It should be white instead.

**The background should remain gray when widgets are displayed in the side column.**

Ideally, the background color should be defined by the parent container. If it's too difficult, stick with the current approach.

Image

mabdullahabaid:
Closes [2029](https://github.com/twentyhq/core-team-issues/issues/2029)

Widgets now use a white background on mobile and in the side panel, while keeping gray in the side column. Background colors are set at the parent container level (PageLayoutVerticalListViewer/Editor) so widgets inherit the correct background, centralizing the background logic in the container components.

Other variants (dashboard, record-page) specify their own background color inside `WidgetCard.tsx` file, so we should not have any regressions.


packages/twenty-front/src/modules/page-layout/components/PageLayoutVerticalListViewer.tsx
```diff
@@ -15,8 +24,17 @@ type PageLayoutVerticalListViewerProps = {
 export const PageLayoutVerticalListViewer = ({
   widgets,
 }: PageLayoutVerticalListViewerProps) => {
+  const isMobile = useIsMobile();
+  const { isInRightDrawer } = useLayoutRenderingContext();
+  const { isInPinnedTab } = useIsInPinnedTab();
+
+  const shouldUseWhiteBackground =
+    (isMobile || isInRightDrawer) && !isInPinnedTab;
```
Devessier: Let's extract this common logic into a custom hook.

Devessier (commented):
Great job. Thank you.
Let's slightly refactor the code before merging.

packages/twenty-front/src/modules/page-layout/components/PageLayoutVerticalListViewer.tsx
mabdullahabaid: Done

packages/twenty-front/src/modules/page-layout/components/PageLayoutGridLayout.tsx
```diff
@@ -104,6 +113,13 @@ export const PageLayoutGridLayout = ({ tabId }: PageLayoutGridLayoutProps) => {
 
   const gridContainerRef = useRef<HTMLDivElement>(null);
 
+  const isMobile = useIsMobile();
+  const { isInRightDrawer } = useLayoutRenderingContext();
+  const { isInPinnedTab } = useIsInPinnedTab();
+
+  const shouldUseWhiteBackground =
+    (isMobile || isInRightDrawer) && !isInPinnedTab;
```
Devessier: @mabdullahabaid, I feel these lines could also be refactored

mabdullahabaid: Ahh, yes missed it, thanks. Refactored.

Devessier (approved):

PR #16880, opened 2025-12-31
Fix timeline activities display

prastoin:
Fixes https://github.com/twentyhq/twenty/issues/16809

# Introduction
Front was expected timeline activities field to be relation and not morph relation

Also fixed the new v2 workspace creation that had a bug in the createStandardRelation util ( it's not in production yet so everything's fine )

close https://github.com/twentyhq/twenty/issues/16809


Weiko (approved):
LGTM, thx

PR #16916, opened 2026-01-02
If else node tests

abdulrahmancodes:


packages/twenty-server/test/integration/graphql/suites/workflow/if-else-workflow.integration-spec.ts
```diff
@@ -0,0 +1,401 @@
...
+      getWorkflowResponse.body.data.workflow.versions.edges.length,
+    ).toBeGreaterThan(0);
+    createdWorkflowVersionId =
+      getWorkflowResponse.body.data.workflow.versions.edges[0].node.id;
+
+    const manualTrigger = {
+      name: 'Manual Trigger',
+      type: 'MANUAL',
+      settings: {
+        outputSchema: {},
```
FelixMalfait: I was expecting to see "number" around here, it doesn't break without it? Probably we should still add it with the right format?

packages/twenty-server/test/integration/graphql/suites/workflow/if-else-workflow.integration-spec.ts
```diff
@@ -0,0 +1,401 @@
...
+        payload: { number: 5 },
+      });
+
+      const workflowRun = await waitForWorkflowCompletion(workflowRunId);
+
+      expect(workflowRun).toBeDefined();
+      expect(workflowRun?.status).toBe('COMPLETED');
+      expect(workflowRun?.state?.stepInfos).toBeDefined();
+      expect(workflowRun?.state?.stepInfos?.trigger?.status).toBe('SUCCESS');
+      expect(workflowRun?.state?.stepInfos?.[ifElseStepId!]?.status).toBe(
```
FelixMalfait: Sorry I don't see the different with previous check, where do you check it has taken the right branch?

packages/twenty-server/test/integration/graphql/suites/workflow/if-else-workflow.integration-spec.ts
abdulrahmancodes: yes

charlesBochet:
@abdulrahmancodes 
image


Tests seems to be failing, could you please fix them, re-push and re-open :)

FelixMalfait:
@abdulrahmancodes feel free to keep using the Github CI as running these locally are very painful. You can re-open as draft if you want

FelixMalfait (approved):

PR #16918, opened 2026-01-03
Fix navigation memory after custom object rename

### LINKED ISSUE #11291: Object Rename Causes 404 on Redirect
## Bug Description

When renaming an object in Settings → Data Models while on the object's Details Page, the updated name is not reflected in the redirect URL after clicking Exit Settings. This results in a 404 error since the system still tries to access the object using its old name.

## Example

Go to the Details Page of any object (e.g., Pets).

Click Settings while on the object's Details Page.

Navigate to Data Models.

Rename the object (e.g., change Pet to PetDetail).

Click Exit Settings.

Observe that the redirect URL still references the old object name, leading to a 404 error.

Possible Cause
The issue appears to be caused by the Recoil state (navigationMemorisedUrl), which does not update after renaming an object.

https://github.com/user-attachments/assets/9d5e1392-e71c-43bb-ba21-1ad7a10fc683

RajdeepDs:
This PR fixes an issue where exiting Settings after renaming a custom object could redirect to a stale object URL and result in a `404`. When a custom object name was updated, the memorized navigation URL was not kept in sync, causing redirects to use the old object route. 
The navigation state is now updated only when the memorized URL belongs to the renamed object, ensuring redirects always point to the correct route while preserving unrelated navigation context.

Fixes https://github.com/twentyhq/twenty/issues/11291

### DIFF
--- a/packages/twenty-front/src/modules/settings/data-model/object-details/components/SettingsUpdateDataModelObjectAboutForm.tsx
+++ b/packages/twenty-front/src/modules/settings/data-model/object-details/components/SettingsUpdateDataModelObjectAboutForm.tsx
@@ -6,6 +6,7 @@ import {
   type SettingsDataModelObjectAboutFormValues,
   settingsDataModelObjectAboutFormSchema,
 } from '@/settings/data-model/validation-schemas/settingsDataModelObjectAboutFormSchema';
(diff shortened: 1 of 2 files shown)


etiennejouan (approved):
Hi @RajdeepDs thank you so much for your contribution ! It works well, I took the liberty to slightly re-organize.
Thanks ! 🙏

PR #16922, opened 2026-01-03
Fix: time format issue in datepicker mask

vasu1303:
Fixes: #16872 

## Summary

Fixes the DateTimePicker to respect user's 12H/24H time format preference. Previously, the time display at the top of the DateTimePicker always showed 24-hour format (e.g., "14:30") regardless of the user's time format setting.

## Screenshots

_After:_ Time respects user preference, showing 12H with AM/PM (e.g., "03:28 PM") or 24H format
image

### Changes Made:

1. **TimeMask.ts** - Added `getTimeMask()` to return appropriate mask pattern based on time format
2. **TimeBlocks.ts** - Restored as constant file per linting requirements
3. **getTimeBlocks.ts** (new) - Dynamic block generator supporting 12H (1-12 + AM/PM) and 24H (0-23)
4. **DateTimeBlocks.ts** - Simplified to static constant
5. **getDateTimeMask.ts** - Updated to accept and use `timeFormat` parameter
6. **DateTimePickerInput.tsx** - Dynamically generates mask and blocks based on user's time format preference, increased input width to accommodate AM/PM
7. **useParseJSDateToIMaskDateTimeInputString.ts** - Formats dates with correct time pattern
8. **useParseDateTimeInputStringToJSDate.ts** - Parses dates with correct time pattern
9. **date-utils.ts** - Added `getTimePattern()` helper (returns 'hh:mm a' for 12H, 'HH:mm' for 24H)
10. **parseDateTimeToString.ts** - Added optional `timeFormat` parameter for consistency

### Technical Approach:

- Leverages existing `useDateTimeFormat()` hook to get user's `timeFormat` preference
- Supports `TimeFormat.SYSTEM`, `TimeFormat.HOUR_12`, and `TimeFormat.HOUR_24`
- Uses IMask blocks with appropriate ranges: 1-12 for 12H, 0-23 for 24H
- Adds 'aa' (AM/PM) block for 12-hour format with enum validation

### DIFF
--- a/packages/twenty-front/src/modules/localization/hooks/useDateTimeFormat.ts
+++ b/packages/twenty-front/src/modules/localization/hooks/useDateTimeFormat.ts
@@ -1,6 +1,7 @@
 import { useRecoilValue } from 'recoil';
 
 import { workspaceMemberFormatPreferencesState } from '@/localization/states/workspaceMemberFormatPreferencesState';
+import { resolveDateFormat } from '@/localization/utils/resolveDateFormat';
 import { resolveTimeFormat } from '@/localization/utils/resolveTimeFormat';
 
 export const useDateTimeFormat = () => {
@@ -10,7 +11,7 @@ export const useDateTimeFormat = () => {
 
   return {
     timeZone: workspaceMemberFormatPreferences.timeZone,
-    dateFormat: workspaceMemberFormatPreferences.dateFormat,
+    dateFormat: resolveDateFormat(workspaceMemberFormatPreferences.dateFormat),
     timeFormat: resolveTimeFormat(workspaceMemberFormatPreferences.timeFormat),
     calendarStartDay: workspaceMemberFormatPreferences.calendarStartDay,
   };
--- a/packages/twenty-front/src/modules/localization/hooks/useDateTimeFormat.ts
+++ b/packages/twenty-front/src/modules/localization/hooks/useDateTimeFormat.ts
@@ -1,6 +1,7 @@
 import { useRecoilValue } from 'recoil';
 
 import { workspaceMemberFormatPreferencesState } from '@/localization/states/workspaceMemberFormatPreferencesState';
+import { resolveTimeFormat } from '@/localization/utils/resolveTimeFormat';
 
 export const useDateTimeFormat = () => {
   const workspaceMemberFormatPreferences = useRecoilValue(
@@ -10,7 +11,7 @@ export const useDateTimeFormat = () => {
   return {
     timeZone: workspaceMemberFormatPreferences.timeZone,
     dateFormat: workspaceMemberFormatPreferences.dateFormat,
-    timeFormat: workspaceMemberFormatPreferences.timeFormat,
+    timeFormat: resolveTimeFormat(workspaceMemberFormatPreferences.timeFormat),
     calendarStartDay: workspaceMemberFormatPreferences.calendarStartDay,
   };
 };
--- a/packages/twenty-front/src/modules/localization/utils/resolveTimeFormat.ts
+++ b/packages/twenty-front/src/modules/localization/utils/resolveTimeFormat.ts
@@ -1,0 +1,10 @@
+import { TimeFormat } from '@/localization/constants/TimeFormat';
+import { detectTimeFormat } from '@/localization/utils/detectTimeFormat';
+
+export const resolveTimeFormat = (timeFormat: TimeFormat): TimeFormat => {
+  if (timeFormat === TimeFormat.SYSTEM) {
+    return TimeFormat[detectTimeFormat()];
+  }
+
+  return timeFormat;
+};
--- a/packages/twenty-front/src/modules/localization/utils/resolveTimeFormat.ts
+++ b/packages/twenty-front/src/modules/localization/utils/resolveTimeFormat.ts
@@ -3,7 +3,8 @@ import { detectTimeFormat } from '@/localization/utils/detectTimeFormat';
 
 export const resolveTimeFormat = (timeFormat: TimeFormat): TimeFormat => {
   if (timeFormat === TimeFormat.SYSTEM) {
-    return TimeFormat[detectTimeFormat()];
+    const detectedFormat = detectTimeFormat();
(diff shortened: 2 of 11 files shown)


packages/twenty-front/src/modules/localization/hooks/useDateTimeFormat.ts
```diff
@@ -10,7 +11,7 @@ export const useDateTimeFormat = () => {
   return {
     timeZone: workspaceMemberFormatPreferences.timeZone,
     dateFormat: workspaceMemberFormatPreferences.dateFormat,
-    timeFormat: workspaceMemberFormatPreferences.timeFormat,
+    timeFormat: resolveTimeFormat(workspaceMemberFormatPreferences.timeFormat),
```
charlesBochet, line 15: why do we have this inconsistency with dateFormat? it seems wrong that timeFormat is computed through an util but not dateFormat

packages/twenty-front/src/modules/localization/utils/resolveTimeFormat.ts
```diff
@@ -0,0 +1,11 @@
+import { TimeFormat } from '@/localization/constants/TimeFormat';
+import { detectTimeFormat } from '@/localization/utils/detectTimeFormat';
+
+export const resolveTimeFormat = (timeFormat: TimeFormat): TimeFormat => {
```
charlesBochet, line 4: see comment above

packages/twenty-front/src/utils/date-utils.ts
```diff
@@ -182,17 +183,23 @@ export const formatToHumanReadableDate = (date: Date | string) => {
   return i18n.date(parsedJSDate, { dateStyle: 'medium' });
 };
 
+const getTimePattern = (timeFormat: TimeFormat) => {
+  return timeFormat === TimeFormat.HOUR_12 ? 'hh:mm a' : 'HH:mm';
+};
+
 export const getDateTimeFormatStringFoDatePickerInputMask = (
```
charlesBochet: let's pass an object: { dateFormat, timeFormat } instead, these eases the devXP

packages/twenty-front/src/modules/ui/input/components/internal/date/utils/getDateTimeMask.ts
```diff
@@ -1,8 +1,11 @@
...
-
 import { type DateFormat } from '@/localization/constants/DateFormat';
+import { type TimeFormat } from '@/localization/constants/TimeFormat';
 import { getDateMask } from './getDateMask';
+import { getTimeMask } from './getTimeMask';
 
-export const getDateTimeMask = (dateFormat: DateFormat): string => {
-  return `${getDateMask(dateFormat)} ${TIME_MASK}`;
+export const getDateTimeMask = (
+  dateFormat: DateFormat,
```
charlesBochet: same comment here (see below)

charlesBochet:
@vasu1303 Thanks for opening this PR—much appreciated.

We only accept PRs with passing (green) tests. Since the overall direction looks good, I’ve added a few comments to help guide you.

I’m going to close this PR for now because the tests are failing, but feel free to reopen it if you have the bandwidth to continue working on it. We’ll be happy to keep reviewing it.

To reopen the PR, you can comment:
"/twenty pr open"

packages/twenty-front/src/modules/localization/hooks/useDateTimeFormat.ts
vasu1303: resolved, added `resolveDateFormat()` too

packages/twenty-front/src/utils/date-utils.ts
vasu1303: updated to pass objects, thanks!

vasu1303:
Hi @charlesBochet,  sorry for the delay in getting back to this! I have now addressed all the feedback: I unified the date and time resolution using utilities and refactored the components to pass { dateFormat, timeFormat } as an object to improve the DevXP. I've also made sure the tests are passing.

Let me know if this looks good!

lucasbordeau:
@vasu1303 Nice work.

After testing this, there's a small UX problem.

The `AM` or `PM` shouldn't reset until we delete it, right now it is deleted if we change the time part, which forces a user to put back `AM` or `PM` each time he changes the time part.

The input has been developed to not destroy any other part when modifying only a part of the date time input text. 

You can find the right lifecycle with the mask and the input to avoid this annoying effect.

lucasbordeau:
https://github.com/user-attachments/assets/5ce7ead5-0327-486d-b628-91890a8004e8

packages/twenty-front/src/modules/ui/input/components/internal/date/hooks/useParseJSDateToIMaskDateTimeInputString.ts
```diff
@@ -1,13 +1,19 @@
...
 import { getDateTimeFormatStringFoDatePickerInputMask } from '~/utils/date-utils';
 
 export const useParseJSDateToIMaskDateTimeInputString = () => {
-  const { dateFormat } = useDateTimeFormat();
+  const { dateFormat, timeFormat } = useDateTimeFormat();
 
   const parseJSDateToDateTimeInputString = (date: Date) => {
-    const parsingFormat =
-      getDateTimeFormatStringFoDatePickerInputMask(dateFormat);
+    if (!date || !isValid(date)) {
```
lucasbordeau, line 9: I wouldn't bring validation with the Date object, as we're trying to avoid its usage. Prefer using Temporal.

lucasbordeau (approved):

lucasbordeau:
Hi @vasu1303 nice work. I've decided to preserve the input instead of forcing the enum, since the final validation is still here, it's less painful to be able to enter a wrong first character than to lose the AM/PM part every time we change another block.

vasu1303:
Hi @lucasbordeau, thank you for the appreciation and you taking the time out to fix that UX issue. Sorry I couldn't circle back sooner due to a busy streak on my end, but I'm glad to see it merged. Thanks!

PR #16934, opened 2026-01-05
Fix command menu input click outside listener blocking other elements

ehconitin:
click outside listeners for `CommandMenuItemNumberInput` and `CommandMenuItemTextInput` were always active, even when the input wasn't focused. This blocked clicks on other elements like toggles in chart settings.

fixed by only enabling click outside listener when input is actually focused


Weiko (approved):
LGTM

bosiraphael (changes requested):
No need to add a local state, we can use the focus stack state

bosiraphael (approved):
LGTM

PR #16937, opened 2026-01-05
Clarify non-implemented calendar layouts

Weiko:
## Context
Some users are confused with the dropdown showing other calendar layouts. We've decided to hide it for now since their implementation is not planned yet.

Before
Screenshot 2026-01-05 at 15 06 36

After
Screenshot 2026-01-05 at 15 00 15


FelixMalfait (approved):
LGTM but needs lint

PR #16943, opened 2026-01-05
Publish batch events for sse subscriptions

thomtrp:
- publish batch events to avoid multiple redis / graphql sse publications 
- add event to the endpoint output


lucasbordeau (approved):
LGTM

PR #16945, opened 2026-01-05
fix: Remove incorrect background logic from PageLayoutGridLayout

mabdullahabaid:
Removes the background color logic that was incorrectly added to PageLayoutGridLayout in PR #16870. Grid layouts are only used for dashboards where widgets have their own background colors set in WidgetCard.tsx, so the container background was causing visual mismatches in padding/gap areas. The background logic remains correctly applied in PageLayoutVerticalListViewer and PageLayoutVerticalListEditor where widgets need to inherit the container background.


Devessier (changes requested):
Thanks for the quick fix.

Could we add a check in `useShouldUseWhiteBackground` to ensure the page layout type is `RECORD_PAGE`? It would make it explicit that the hook is only meant to be used in this case:

```ts
import { useIsInPinnedTab } from '@/page-layout/widgets/hooks/useIsInPinnedTab';
import { useLayoutRenderingContext } from '@/ui/layout/contexts/LayoutRenderingContext';
import { useIsMobile } from '@/ui/utilities/responsive/hooks/useIsMobile';
import { PageLayoutType } from '~/generated/graphql';

export const useShouldUseWhiteBackground = () => {
  const isMobile = useIsMobile();
  const { isInRightDrawer, layoutType } = useLayoutRenderingContext();
  const { isInPinnedTab } = useIsInPinnedTab();

  const shouldUseWhiteBackground =
    (isMobile || isInRightDrawer) &&
    !isInPinnedTab &&
    layoutType === PageLayoutType.RECORD_PAGE;

  return { shouldUseWhiteBackground };
};
```

Furthermore, I'm unsure if the check `!isInPinnedTab` is necessary.

Devessier (approved):

PR #16949, opened 2026-01-06
Fix metadata relation field settings item type tag display

prastoin:
# Introduction
followup https://github.com/twentyhq/twenty/pull/16880

Fixing the item tag type displayal of the relation field within an object
Inferring the application id at field level instead of object level

## From
image

## To
image


martmull (approved):

Weiko (approved):
LGTM

PR #16950, opened 2026-01-06
Clean workflow run stoppage flag

thomtrp:
Flag has never been removed. Feature allows to stop a running workflow run.


Weiko (approved):
LGTM

PR #16952, opened 2026-01-06
Billing - Update alert when meter price is updated

etiennejouan:


packages/twenty-server/src/engine/core-modules/billing/services/billing-subscription-update.service.ts
FelixMalfait, line 217: Is this similar?

FelixMalfait (approved):

PR #16953, opened 2026-01-06
Fix add related records to activity target after object renaming

### LINKED ISSUE #16928: Cannot add Relations to a Note for a given workspace
## Bug Description
https://github.com/twentyhq/private-issues/issues/400

ijreilly:
Fixes https://github.com/twentyhq/twenty/issues/16928 and https://github.com/twentyhq/private-issues/issues/400

When an object is renamed, its corresponding relation field on noteTarget and taskTarget are not renamed. This caused an issue as in the FE, we used the related record object name to compute the field name used in the payload.
This PR addresses it by, instead, finding the field name after identifying it thanks to the object id. 

To reproduce, you can rename an object then try to tie it to a note.


etiennejouan (commented):
Looks good but seems it still needs some front logic update. 

Cf when trying to open Note tabs from a renamed object record

https://github.com/user-attachments/assets/26add61c-c886-431c-bc7e-b13e2d832c7d

packages/twenty-front/src/modules/activities/utils/getActivityTargetObjectRecords.ts
```diff
@@ -36,27 +38,50 @@ export const getActivityTargetObjectRecords = ({
...
+    : CoreObjectNameSingular.TaskTarget;
+
+  const activityTargetObjectMetadata = objectMetadataItems.find(
+    (item) => item.nameSingular === activityTargetObjectNameSingular,
+  );
+
+  const activityTargetRelationFields =
+    activityTargetObjectMetadata?.fields.filter(
+      (field) =>
+        isDefined(field.relation?.targetObjectMetadata.id) &&
```
prastoin, line 52: ***Polish:*** Would verify namingSingular is defined in order to avoid passing undefined or null inside the includes

packages/twenty-front/src/modules/activities/utils/getActivityTargetFieldNameForObject.ts
```diff
@@ -0,0 +1,38 @@
...
+      objectMetadataItem.nameSingular === activityTargetObjectNameSingular,
+  );
+
+  if (!isDefined(activityTargetObjectMetadata)) {
+    return undefined;
+  }
+
+  const targetField = activityTargetObjectMetadata.fields.find(
+    (field) =>
+      field.relation?.targetObjectMetadata.id === targetObjectMetadataId,
```
prastoin, line 34: ***Polish:*** would also assert isDefined here, honestly that's very capilo tractee as `targetObjectMetadataId` should always be defined but we never know what could be returned from the graphql api in the `morphRelation` in the end at runtime

packages/twenty-front/src/modules/activities/inline-cell/hooks/useUpdateActivityTargetFromCell.ts
```diff
@@ -159,8 +173,8 @@ export const useUpdateActivityTargetFromCell = ({
           await createOneActivityTarget({
             ...activityTarget,
             [targetObjectName]: undefined,
-            [pickedObjectMetadataItem.nameSingular]: undefined,
-            [`${pickedObjectMetadataItem.nameSingular}Id`]: morphItem.recordId,
+            [targetFieldName]: undefined,
+            [`${targetFieldName}Id`]: morphItem.recordId,
```
prastoin, line 177: ***remark:*** out of scope we might wanna expose `computeMorphOrRelationFieldJoinColumnName` in twenty-shared

prastoin (approved):
Looks solid !

PR #16964, opened 2026-01-06
Update updatedat field on dashboards after edition

### LINKED ISSUE #1896: Update updatedAt field on dashboards after edition
When editing a dashboard, the underlying page layout is updated. But this update is not reflected in the updatedAt field in the dashboard.
When updating a page layout, tab or widget, we should update the linked dashboard updatedAt field.

bosiraphael:
Closes https://github.com/twentyhq/core-team-issues/issues/1896


packages/twenty-server/src/engine/metadata-modules/dashboard/services/dashboard-sync.service.ts
```diff
@@ -0,0 +1,110 @@
...
+          const dashboardRepository =
+            await this.globalWorkspaceOrmManager.getRepository(
+              workspaceId,
+              'dashboard',
+              { shouldBypassPermissionChecks: true },
+            );
+
+          await dashboardRepository.update(
+            { pageLayoutId },
+            { updatedAt: new Date() },
```
Weiko: Note: do we want the exact same value as the pagelayout?

packages/twenty-server/src/engine/metadata-modules/page-layout-tab/services/page-layout-tab.service.ts
```diff
@@ -327,6 +344,11 @@ export class PageLayoutTabService {
       );
     }
 
+    await this.dashboardSyncService.updateLinkedDashboardsUpdatedAtByTabId({
```
Weiko, line 350: Don't we want to actually destroy the dashboard in this case?

packages/twenty-server/src/engine/metadata-modules/page-layout-tab/services/page-layout-tab.service.ts
```diff
@@ -274,6 +286,11 @@ export class PageLayoutTabService {
...
```
Weiko: Don't we want to update the deletedAt here?

packages/twenty-server/src/engine/metadata-modules/dashboard/services/dashboard-sync.service.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  async updateLinkedDashboardsUpdatedAtByTabId({
+    tabId,
+    workspaceId,
+  }: {
+    tabId: string;
+    workspaceId: string;
+  }): Promise<void> {
+    const { flatPageLayoutTabMaps } =
```
Weiko: This is reading from the cache but it seems you re calling your new service before the getOrRecompute

Weiko: Nevermind, this should be good, no race condition because we invalidate in the runner 👍

packages/twenty-server/src/engine/metadata-modules/page-layout-tab/services/page-layout-tab.service.ts
bosiraphael: It's the tab service, but for the page layout service yes it could make sense

packages/twenty-server/src/engine/metadata-modules/page-layout-tab/services/page-layout-tab.service.ts
bosiraphael: It is already handled: in the page layout service
 ```
if (flatPageLayoutToDestroy.type === PageLayoutType.DASHBOARD) {
      await this.destroyAssociatedDashboards(id, workspaceId);
    }
```

Weiko (approved):
LGTM!

PR #16966, opened 2026-01-06
Implemented update-one event optimistic effect on table

lucasbordeau:
This PR implements update-one event optimistic effect on table.

It will ignore all other events. 

This is a POC, not a final implementation so things have not been cleaned yet.

This simple implementation already brings some questions.

# Optimistic update of relation ?

When updating a relation, the event that is sent does not contain the nested record that is update, only the recordId. 

Solutions proposed : 
- The frontend gives a parameter like `attachNestedRecordIdentifier`, then `workspace-event-emitter.resolver.ts` computes what's needed and modifies the event record.
- The frontend computes the record identifier, and makes a find one request if needed, not ideal, since the backend could do and avoid subsequent network round trips.

⚠️ Due to this problem, relation update event is not handled for now.

# Returning only the modified fields ? 

When an update event is sent, it would be interesting to filter out the fields that are not updated, to optimize the network load.

# Avoid sending event to the original application ?

It would be interesting to send events only to the other browser tabs. Because the original browser tab that makes the request doesn't need an event. 

The problem is that we don't have a trace id from the original mutation down to the subscription resolver, because in-between there is an abstract catch-all database event listener.

We could guess that an event is not interesting for a given frontend application, but that would still involve additional network round trip.

Another risk is that the logic that processes the even on the frontend ends up creating weird bugs due to doubling the optimistic effect, but since our optimistic engine should be idempotent, that could be ok.

# Demo

https://github.com/user-attachments/assets/5554a1ae-3f84-468d-aac3-01c2efa156e1

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-table/virtualization/components/RecordTableVirtualizedSSEEffect.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-table/virtualization/components/RecordTableVirtualizedSSEEffect.tsx
@@ -1,0 +1,203 @@
+import { triggerUpdateRecordOptimisticEffect } from '@/apollo/optimistic-effect/utils/triggerUpdateRecordOptimisticEffect';
+import { useApolloCoreClient } from '@/object-metadata/hooks/useApolloCoreClient';
+import { useObjectMetadataItems } from '@/object-metadata/hooks/useObjectMetadataItems';
+import { useGetRecordFromCache } from '@/object-record/cache/hooks/useGetRecordFromCache';
+import { getObjectTypename } from '@/object-record/cache/utils/getObjectTypename';
+import { getRecordNodeFromRecord } from '@/object-record/cache/utils/getRecordNodeFromRecord';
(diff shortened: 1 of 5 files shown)


packages/twenty-front/src/modules/object-record/record-table/virtualization/components/RecordTableVirtualizedSSEEffect.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+          const matchesForThisSubscription = matches.filter((match) => {
+            return match.subscriptionIds.includes(subscriptionId);
+          });
+
+          const updateEvents = matchesForThisSubscription.filter((match) => {
+            return match.event.action === DatabaseEventAction.UPDATED;
+          });
+
+          if (updateEvents.length !== 1) {
```
thomtrp, line 99: can't we lose events since we handle these by batches?

lucasbordeau:
Closing this PR as we've decided to adopt another pattern

PR #16971, opened 2026-01-06
Remove tests run on push on main 

ijreilly:
Since tests are now run in the pre-merge queue with the latest main version, they need not to be run again when merged into main, it would be the exact same thing

### DIFF
--- a/.github/workflows/ci-front.yaml
+++ b/.github/workflows/ci-front.yaml
@@ -1,10 +1,6 @@
 name: CI Front and E2E
 
 on:
-  push:
-    branches:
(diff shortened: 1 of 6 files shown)


Weiko (approved):

.github/workflows/ci-front.yaml
```diff
@@ -1,10 +1,6 @@
 name: CI Front and E2E
 
 on:
-  push:
```
Weiko, line 4: The same change is probably needed for twenty-server workflows and others I guess?

PR #16972, opened 2026-01-06
fix purple color palette

ehconitin:
before - 

CleanShot 2026-01-06 at 23 20 50

now - 

CleanShot 2026-01-06 at 23 20 03


bosiraphael (approved):
LGTM

PR #16974, opened 2026-01-06
If else node followup changes

abdulrahmancodes:


packages/twenty-front/src/modules/workflow/workflow-steps/hooks/useDeleteWorkflowVersionStep.ts
```diff
@@ -23,7 +30,16 @@ export const useDeleteWorkflowVersionStep = () => {
   const deleteWorkflowVersionStep = async (
     input: DeleteWorkflowVersionStepInput,
   ) => {
-    const result = await mutate({ variables: { input } });
+    const result = await mutate({
+      variables: { input },
+      awaitRefetchQueries: true,
+      refetchQueries: [
```
thomtrp, line 36: if we stop removing the empty node, we can remove that refetch logic. Updating cache will be enough

packages/twenty-server/src/modules/workflow/workflow-builder/workflow-version-step/utils/remove-step.ts
```diff
@@ -48,17 +66,71 @@ export const removeStep = ({
...
+  const allRemovedStepIds = [stepIdToDelete, ...emptyChildStepIds];
+
   const updatedSteps =
     existingSteps
-      ?.filter((step) => step.id !== stepIdToDelete)
+      ?.filter((step) => !allRemovedStepIds.includes(step.id))
       .map((step) => {
+        if (
+          step.type === WorkflowActionType.IF_ELSE &&
+          isWorkflowIfElseAction(step)
```
thomtrp, line 94: This behavior looks huge and is not consistent with iterators. For iterators, we let the user remove the empty node. The user can delete it by himself. I think we can do that for now and see if user complain. Wdyt?

packages/twenty-front/src/modules/workflow/workflow-steps/hooks/useDeleteWorkflowVersionStep.ts
abdulrahmancodes: @FelixMalfait suggested that we should remove the empty nodes. LMK what should we do @thomtrp ?

packages/twenty-front/src/modules/workflow/workflow-diagram/components/WorkflowDiagramCanvasBase.tsx
```diff
@@ -501,13 +502,31 @@ export const WorkflowDiagramCanvasBase = ({
...
+
+      const bounds = containerRef.current.getBoundingClientRect();
+
       setConnectionStartInfo({
         nodeId: params.nodeId,
         handleId: params.handleId,
+        startPosition: {
+          x: clientX - bounds.left,
+          y: clientY - bounds.top,
+        },
```
thomtrp, line 529: why is this useful ? can't we simply deactivate the handle when the node is an if-else?

packages/twenty-front/src/modules/workflow/workflow-steps/utils/getWorkflowPreviousSteps.ts
```diff
@@ -18,6 +18,12 @@ const isParentStep = ({
     );
   }
 
+  if (potentialParentStep.type === 'IF_ELSE') {
+    return !!potentialParentStep.settings.input.branches?.some((branch) =>
+      branch.nextStepIds?.includes(currentStep.id),
+    );
+  }
```
thomtrp, line 25: 👌

packages/twenty-server/src/modules/workflow/workflow-builder/workflow-version-step/utils/remove-step.ts
thomtrp: ok let's keep it for now. We'll see if we do the same for iterators at some point

packages/twenty-front/src/modules/workflow/workflow-steps/hooks/useDeleteWorkflowVersionStep.ts
thomtrp: let's keep it for now

packages/twenty-front/src/modules/workflow/workflow-diagram/components/WorkflowDiagramCanvasBase.tsx
abdulrahmancodes: The startPosition is needed to differentiate between a simple click on the handle and an actual drag (when the user drags from the handle to create a new node). Without tracking the start position, we can’t measure the drag distance, and every handle click would end up creating a node. This behavior applies to all node types, not just if-else.

thomtrp (approved):
LGTM

PR #16977, opened 2026-01-06
Fix the crash of the app page when object view is configured to open only on record page.

### LINKED ISSUE #16577: Crash when navigating to record page with command menu open
## Steps to Reproduce

1. Go to any record index page (e.g., `/objects/people`, `/objects/companies`, `/objects/dashboards`)
2. Make sure the object's view is configured to open only on record page
   - Note: Dashboards and Workflows always open in the record page
3. Open the command menu (Cmd/Ctrl+K) and **keep it open**
4. Click on any record from the table, which will navigate to a full page
5. **The app crashes**

## Expected behavior

It shouldn't crash -- should handle the navigation gracefully

araj00:
### Problem:
If the command menu page is kept opened while navigating to record show page then the app crashes.

### Root cause

When navigatiing to a record show page while the command menu page is kept opened, it tries to open the record show page with command menu open, as its state were set to true , before navigating to a new page along with the instance id of previous context which are used in the current context while rendering the page. Hence, leading to an error page.

### Changes

while navigation to a record show page the command menu open state is set to false. This is to handle the navigation gracefully

Fixes: #16577


packages/twenty-front/src/modules/object-record/record-calendar/components/RecordCalendarAddNew.tsx
```diff
@@ -70,8 +73,9 @@ export const RecordCalendarAddNew = ({
 
   return (
     <StyledButton
-      onClick={() => {
-        createNewIndexRecord({
+      onClick={async () => {
+        closeCommandMenu();
```
charlesBochet: I feel:
- this should actually be part of createNewIndexRecord
- and we should call it only if the view is set to navigate to a new record page

packages/twenty-front/src/modules/object-record/record-index/hooks/useOpenRecordFromIndexView.ts
```diff
@@ -82,6 +85,7 @@ export const useOpenRecordFromIndexView = () => {
             resetNavigationStack: true,
           });
         } else {
+          closeCommandMenu();
```
charlesBochet, line 88: this one looks good

packages/twenty-front/src/modules/object-record/record-calendar/components/RecordCalendarAddNew.tsx
araj00: I have done the suggested changes. Please check on it.

charlesBochet (approved):

PR #16984, opened 2026-01-07
Fix memory crash when creating record in table view

### LINKED ISSUE #16980: Workspace cannot create new People record
## Bug Description
https://github.com/twentyhq/private-issues/issues/402

etiennejouan:
**Bug Fixed**
Creating a new record in the People table view caused a browser memory crash ("Paused before potential out of memory crash").
**Root Cause**
In useResetVirtualizationBecauseDataChanged.ts, when creating a record, a loop iterated from 0 to totalNumberOfRecordsToVirtualize (the total database count). 
**Fix Applied**
Changed the loop to only iterate through indices from actually loaded pages

Fixes https://github.com/twentyhq/twenty/issues/16980


etiennejouan:
After discussing it with @charlesBochet, we choose to remove individual row loading states resetting.
We accept the potential visual glitches (stale data or incorrect loading indicators) until the refetch completes.

prastoin (approved):
As discussed should also be reviewed by @lucasbordeau 
We might wanna be iterating over the displayed data total count instead of none instead of all of them in db previously
Sounds good to me though !

PR #16987, opened 2026-01-07
Fix view picker small bugs

### LINKED ISSUE #15422: [Views] - Editing the icon of the view is no longer shown immediately
## Bug Description

Before yesterday's release (i think), when a user edited the icon of a view, the new icon appeared immediately
Now, the icon does not seem updated: the user needs to reload the page to see the new one. 
It gives the impression the user request is not taken into account. 
Likely like to the optimistic

**Example:** 
I can provide a video if need be

thanks!

lucasbordeau:
This PR solves small bugs around the view picker.

- Couldn’t obtain optimistic update after a re-order of a view by drag and drop, we needed to refresh the page
- Picking a new icon wouldn’t trigger optimistic update (same problem)
- Picking a new icon would change the view (difficult to understand behavior)
- Picking a new icon would trigger left drawer collapse (z-index problem)

Since core views are not being handled by object metadata items anymore, and that all view logic is plugged on coreViewsState, this PR implemented optimistic effect by upserting into this state.

Fixes https://github.com/twentyhq/twenty/issues/15422
Fixes https://github.com/twentyhq/twenty/issues/16986

# Before

https://github.com/user-attachments/assets/64099c21-df9f-4772-ab0d-9ea449aed761

# After

https://github.com/user-attachments/assets/f4e844b3-6530-4178-abdb-b7a10d2327b8


packages/twenty-front/src/modules/views/hooks/internal/usePerformViewAPIPersist.ts
```diff
@@ -91,39 +88,6 @@ export const usePersistView = () => {
     ],
   );
 
-  const updateView = useCallback(
```
charlesBochet, line 96: I would leave it in usePersistView

Current naming convention:
- persistViewFilter = create, update, delete in backend
- saveViewFilter = transform the recordFilters to viewFilters and call persistViewFilter
- remove / upsertRecordFilter = update record filters

charlesBochet: New naming proposal:
- performViewFilterAPICreate / performViewFilterAPIUpdate / performViewFilterAPIUpdate viewFilter => wrapping of api calls
- saveRecordFiltersToViewFilters
- upsert / remove recordFilter => update recordFilters state

charlesBochet: so this hook should performViewAPIUpdate

charlesBochet (approved):

PR #16991, opened 2026-01-07
Provide executed step output instead of result

thomtrp:
For huge workflows, after 20 steps, we stop the job and re-trigger a new one. 
But some of those workflows actually never stop.
See https://github.com/twentyhq/private-issues/issues/401 

Here is a first issue. We resuming the workflow, we were providing the result instead of the output


martmull (approved):
free stats

PR #16992, opened 2026-01-07
fix: composite field matching in relation field orderBy with groupBy

ehconitin:
- Fix groupBy field matching for composite fields nested within relation fields
- Add `nestedSubFieldName` comparison when the nested field is a composite type (CURRENCY, FULL_NAME, ADDRESS, etc.)


packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-order/utils/prepare-for-order-by-relation-field-parsing.util.ts
```diff
@@ -86,12 +87,38 @@ export const prepareForOrderByRelationFieldParsing = ({
...
-  ) as GroupByRelationField | undefined;
+  let compositeSubFieldName: string | undefined;
+
+  if (
+    isCompositeFieldMetadataType(nestedFieldMetadata.type) &&
+    isObject(nestedFieldOrderByValue)
+  ) {
+    const compositeSubFields = Object.keys(nestedFieldOrderByValue);
+
+    if (compositeSubFields.length === 1) {
```
etiennejouan, line 98: If more than 1, it should throw no ?

etiennejouan: Ok, we can forget, it would throw just after

etiennejouan (approved):
LGTM !

PR #14923, opened 2025-10-06
[groupBy] Date ranges

ijreilly:
Closes https://github.com/twentyhq/core-team-issues/issues/1559

2 kinds of date ranges: absolute (day, month, quarter, year) and "cyclic" (day of the year, month of the year, quarter of the year. ex dates in january 2024 and january 2025 will be grouped together in "Q1").

Absolute dates are expressed in datetime format to ease their handling, while cyclic dates are expressed with more human readable labels ("Q1") because datetime does not make sense anyway

Capture d’écran 2025-10-06 à 18 13 29

Capture d’écran 2025-10-06 à 18 14 02

Capture d’écran 2025-10-06 à 18 14 14

PR #16803, opened 2025-12-24
Reduce complexity on groupBy query

etiennejouan:
fixes https://github.com/twentyhq/core-team-issues/issues/2009

PR #16994, opened 2026-01-07
Fix: Kanban view crashes with 'Cannot convert undefined or null to object'

FelixMalfait:
## Problem
Kanban views with a 'No Value' group (representing records with null values) crash with:
```
TypeError: Cannot convert undefined or null to object
```

This occurs when the filter includes `null` in the `in` clause (e.g., `{status: {in: ['TODO', 'IN_PROGRESS', 'DONE', null]}}`), which happens automatically when a Kanban view includes a 'No Value' column.

## Root Cause
The `checkForDeletedAtFilter` function in `graphql-query.parser.ts` recursively traverses filter objects but didn't handle the case where a nested value is `null`. When recursing into the filter structure, it would call `Object.entries(null)`, which throws the error.

**Timeline:**
- Sept 19, 2024: Bug introduced (latent) in PR #7133 - `checkForDeletedAtFilter` was added without null check
- Oct 7, 2025: Bug became active in PR #14923 - `applyDeletedAtToBuilder` was added to groupBy resolver

## Fix
1. Added null check at the beginning of `checkForDeletedAtFilter` to return early if filter is null/undefined
2. Changed `isDefined` import in `group-by-resolver.factory.ts` from `class-validator` to `twenty-shared/utils` (the class-validator version only checks for `undefined`, not `null`)

## Testing
Manually verified that Kanban views with 'No Value' columns now load correctly.

### DIFF
--- a/packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query.parser.ts
+++ b/packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query.parser.ts
@@ -77,6 +77,10 @@ export class GraphqlQueryParser {
   private checkForDeletedAtFilter = (
     filter: FindOptionsWhere<ObjectLiteral> | FindOptionsWhere<ObjectLiteral>[],
   ): boolean => {
+    if (!filter) {
+      return false;
+    }
+
     if (Array.isArray(filter)) {
       return filter.some((subFilter) =>
         this.checkForDeletedAtFilter(subFilter),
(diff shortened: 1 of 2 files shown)


FelixMalfait:
Screenshot 2026-01-07 at 17 26 56
AI is wrong this bug has been spiking only since the release. Or maybe it's the flag we just enabled?

FelixMalfait:
https://twenty-v7.sentry.io/issues/7168406768/?environment=prod-eu&environment=prod&project=4507072499810304&query=is%3Aunresolved%20Cannot%20convert%20undefined&referrer=issue-stream

etiennejouan (commented):
This PR didn't totally fix the issue I introduce with this [PR](https://github.com/twentyhq/twenty/pull/16803/changes#diff-0e1f4c7ec8d517450bf1c1e6e05bb04a093d5930cd45f6a109a88f397cb8ef56) (groupBy query doesn't fail anymore with your fix but null select records are not returned)

Not sure we should merge it

Fix in this [PR](https://github.com/twentyhq/twenty/pull/16998)

FelixMalfait:
Thanks @etiennejouan!

PR #16794, opened 2025-12-24
[DASHBOARDS] Manual and position-based sorting for chart widgets

bosiraphael:
## Description

SELECT fields have a defined option order that users expect to see reflected in charts.
This PR allows sorting by that position and also enables custom manual ordering.

### Reordering on primary axis

https://github.com/user-attachments/assets/994f515e-19cb-4a5e-b745-e8c77e92ae0b

### Reordering on secondary axis

https://github.com/user-attachments/assets/444c16f2-1920-4dc4-8b42-312d520ab43b

Note: The colors in the graph will match the colors of the select options, but this will be done in another PR

---

> [!NOTE]
> Introduces new sort modes and UI for chart groupings, with full FE/BE support and updated GraphQL schema.
> 
> - Extend `GraphOrderBy` with `FIELD_POSITION_ASC/DESC` and `MANUAL`; add corresponding fields in configs: `primaryAxisManualSortOrder`, `secondaryAxisManualSortOrder`, and `manualSortOrder` (pie)
> - New UI: dropdown options filtered by field type, icons, and a draggable submenu (`ChartManualSortSubMenuContent`) to reorder select options; integrates with widget edit flow
> - Sorting logic added/refactored: `sortChartData`, `sortByManualOrder`, `sortBySelectOptionPosition`, `sortLineChartSeries`, plus updates to bar/line/pie transformers to honor new modes and manual orders
> - Default behaviors: select fields default to `FIELD_POSITION_ASC`; query variable builders skip `orderBy` when using manual/position sorts
> - Update GraphQL generated types/fragments/queries and backend DTOs/schemas to persist new fields; add tests for sorting utilities and snapshots; add sorting icons in `twenty-ui`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 78c9b56c0f1f2d45f7f8b270bb59ca599a005abe. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #16996, opened 2026-01-07
Fix chart sorting

bosiraphael:
I introduced a bug in the chart sorting in this PR: https://github.com/twentyhq/twenty/pull/16794
This PR fixes this.

The sorting on the first axis is already done by the group by for FIELD_ASC and FIELD_DESC. It is sorted also for the second axis but in each group.
For instance, if I create a graph that displays the amount of sales on each day of the week grouped by vendor
I can have for:
Monday: Vendor B, Vendor C
Tuesday: Vendor A, Vendor B
Wednesday: Vendor A, Vendor C
...
Inside each day the order of the vendor is correct, but by looking at only one day, I can't know the order of all the vendors.
That is why we always need to sort the second axis.


packages/twenty-front/src/modules/page-layout/widgets/graph/utils/compareDimensionValues.ts
```diff
@@ -0,0 +1,71 @@
...
+    if (isFieldMetadataDateKind(fieldType)) {
+      const dateA = parseDate(rawValueA, fieldType);
+      const dateB = parseDate(rawValueB, fieldType);
+      if (isDefined(dateA) && isDefined(dateB)) {
+        return applyDirection(Temporal.PlainDate.compare(dateA, dateB));
+      }
+    }
+
+    if (fieldType === FieldMetadataType.CURRENCY) {
+      if (subFieldName === 'amountMicros') {
```
ehconitin, line 56: feels like subFieldName could be typed if we are comparing it like this?
I see its typed as a string

packages/twenty-front/src/modules/page-layout/widgets/graph/utils/__tests__/sortTwoDimensionalChartPrimaryAxisDataByFieldOrManually.test.ts
```diff
@@ -16,15 +16,15 @@ const formattedToRawLookup = new Map([
...
+  it('should return data unchanged by FIELD_ASC, since it is already sorted by the backend', () => {
     const result = sortTwoDimensionalChartPrimaryAxisDataByFieldOrManually({
       data: testData,
       orderBy: GraphOrderBy.FIELD_ASC,
       formattedToRawLookup,
       getFormattedValue,
     });
 
-    expect(result.map((i) => i.label)).toEqual(['Alpha', 'Beta', 'Gamma']);
+    expect(result.map((i) => i.label)).toEqual(['Beta', 'Alpha', 'Gamma']);
```
ehconitin: could the test data itself be sorted then? 
because now we are expecting sorted data from backend?

packages/twenty-front/src/modules/page-layout/widgets/graph/utils/sortTwoDimensionalChartPrimaryAxisDataByFieldOrManuallyIfNeeded.ts
ehconitin, line 1: this util returns data unchanged or `not sorted` like before

naming is now outdated

packages/twenty-front/src/modules/page-layout/widgets/graph/utils/__tests__/sortTwoDimensionalChartPrimaryAxisDataByFieldOrManually.test.ts
ehconitin: best would be to remove this test case -- since there is no 'sorting' in the util ?

packages/twenty-front/src/modules/page-layout/widgets/graph/utils/compareDimensionValues.ts
bosiraphael: Good catch

ehconitin (approved):
lgtm

PR #16998, opened 2026-01-07
Fix kanban view when grouping select field has null value

etiennejouan:


packages/twenty-front/src/modules/object-record/record-group/utils/computeRecordGroupOptionsFilter.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  if (!hasNullValue) {
+    return {
+      [fieldName]: {
+        in: recordGroupValues,
+      },
+    };
+  }
+
+  const nonNullValues = recordGroupValues.filter(
```
prastoin, line 21: ***Question:*** Couldn't we directly insert te `{is: 'NULL'}` format higher ? instead of passing null and re-transpile here ?

packages/twenty-front/src/modules/object-record/record-group/utils/computeRecordGroupOptionsFilter.ts
```diff
@@ -0,0 +1,42 @@
...
+  recordGroupValues: RecordGroupDefinition['value'][];
+}): RecordGqlOperationFilter => {
+  if (!isDefined(recordGroupFieldMetadata) || recordGroupValues.length === 0) {
+    return {};
+  }
+
+  const fieldName = recordGroupFieldMetadata.name;
+  const hasNullValue = recordGroupValues.some(isNull);
+
+  if (!hasNullValue) {
```
prastoin: ***Polish:*** Could facto but totally subjective

prastoin (approved):
Looks good to me !

prastoin:
Could be cool to later add a story book test on this too

PR #17001, opened 2026-01-07
BREAKING CHANGE: Removing remote integration feature

Weiko:
## Context
The feature has not been maintained for more than a year and was never officially launched. 
This PR removes its code due to the upcoming refactoring of the sync-metadata that will break its logic if not handled properly and it would be too costly for the time being.

TODO:
- remove isRemote from object/field metadata


charlesBochet (approved):

prastoin (approved):
🎉

PR #17003, opened 2026-01-07
feat: improve the design of the fields widget

### LINKED ISSUE #2005: [Record Page Layouts] Improve the design of the Fields widget
## Currently

Image

## Expected

- No top border
- Fix section title style
- Sections should be expandable
- Remove padding
- **No need to implement the More button**

Image

Figma: https://www.figma.com/design/xt8O9mFeLl46C5InWwoMrN/Twenty?node-id=79879-145979&t=k9OHC560hshtlA4N-0

mabdullahabaid:
Closes [2005](https://github.com/twentyhq/core-team-issues/issues/2005).

This is how it looks like and I have a feeling that it matches the Figma design. I am not sure if we need to remove more padding as mentioned in the issue since removing it takes it away from the Figma design. Please review and let me know.

image

<br />
<br />

Figma design itself:

image

In terms of data that we receive, I believe it would be handled by Fields Configuration coming from the Backend, so General and Other are decided at that layer, unless I am missing something.

### DIFF
--- a/packages/twenty-front/src/modules/page-layout/widgets/fields/components/FieldsWidgetSectionContainer.tsx
+++ b/packages/twenty-front/src/modules/page-layout/widgets/fields/components/FieldsWidgetSectionContainer.tsx
@@ -62,7 +62,11 @@ export const FieldsWidgetSectionContainer = ({
           stroke={theme.icon.stroke.sm}
         />
       </StyledHeader>
-      <AnimatedExpandableContainer isExpanded={isExpanded}>
+      <AnimatedExpandableContainer
+        isExpanded={isExpanded}
+        initial={false}
+        mode="fit-content"
+      >
         {children}
       </AnimatedExpandableContainer>
     </StyledFieldsWidgetSectionContainer>
--- a/packages/twenty-front/src/modules/page-layout/widgets/fields/components/FieldsWidgetSectionContainer.tsx
+++ b/packages/twenty-front/src/modules/page-layout/widgets/fields/components/FieldsWidgetSectionContainer.tsx
@@ -1,14 +1,22 @@
+import { useTheme } from '@emotion/react';
 import styled from '@emotion/styled';
-import { Section } from 'twenty-ui/layout';
+import { useState } from 'react';
+import { IconChevronDown, IconChevronUp } from 'twenty-ui/display';
+import { AnimatedExpandableContainer, Section } from 'twenty-ui/layout';
 
 const StyledFieldsWidgetSectionContainer = styled(Section)`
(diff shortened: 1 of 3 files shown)


mabdullahabaid:
Maybe we need to reduce left-right padding instead of completely removing it, but open to feedback.

packages/twenty-front/src/modules/page-layout/widgets/fields/components/FieldsWidgetSectionContainer.tsx
```diff
@@ -29,12 +38,33 @@ export const FieldsWidgetSectionContainer = ({
...
+        ) : (
+          <IconChevronDown
+            size={theme.icon.size.md}
+            stroke={theme.icon.stroke.sm}
+            color={theme.font.color.tertiary}
+          />
+        )}
       </StyledHeader>
-      {children}
+      <AnimatedExpandableContainer isExpanded={isExpanded}>
```
Devessier: The animation happens when the component mounts. If I remember well, there is an option to ask the library not to animate on mount.

https://github.com/user-attachments/assets/48754d97-a931-48b1-95bc-900877a5a511

packages/twenty-front/src/modules/page-layout/widgets/fields/components/FieldsWidgetSectionContainer.tsx
```diff
@@ -29,12 +38,33 @@ export const FieldsWidgetSectionContainer = ({
...
-      <StyledHeader>
+      <StyledHeader onClick={handleToggleSection}>
         <StyledTitleLabel>{title}</StyledTitleLabel>
+        {isExpanded ? (
+          <IconChevronUp
+            size={theme.icon.size.md}
+            stroke={theme.icon.stroke.sm}
+            color={theme.font.color.tertiary}
+          />
+        ) : (
```
Devessier: I feel it would be better to rotate the icon instead of going from chevron up to chevron down. This is a pattern we widely use in the codebase, I guess you can take inspiration from there.

https://github.com/user-attachments/assets/75598165-de55-490e-8241-e1ae768a56d5

Devessier (commented):
Thanks for your work. It looks way better like this.

Indeed, the sections will come from the backend. The changes you made to populate the Other sections seem sufficient to me for the moment.

I wrote two small recommendations. Let's work on them before merging the PR.

mabdullahabaid:
@Devessier done.

No animation on first render but during toggles took time to figure out, but got it!

Devessier (approved):
Great 💪

PR #17011, opened 2026-01-08
[Dashboards] Fix chart axis label behavior

ehconitin:
- Axis labels now respect `axisNameDisplay` config when chart has no data (previously always showed both) - fixed for bar and line charts
- Horizontal bar charts now correctly map axis settings to visual positions:
    - "X axis" setting → bottom axis (value)
    - "Y axis" setting → left axis (category)
    
video QA

https://github.com/user-attachments/assets/5bf32294-9a5e-4aa6-b3fd-0d4e33608d14


bosiraphael (approved):
LGTM

PR #17015, opened 2026-01-08
Enable switching field to group records by on existing kanban view

### LINKED ISSUE #16982: Kanban view - cannot select the field with which data is grouped by
## Bug Description
https://github.com/twentyhq/private-issues/issues/403

"When I change layout of a view to Kanban, it automatically adjust the "group" to a fixed field and I cannot edit it. Which ruins all my views"
Reported by a customer and I can reproduce on another workspace, which makes the creation of new Kanban view not available at all

ijreilly:
Fixes https://github.com/twentyhq/twenty/issues/16982 and https://github.com/twentyhq/private-issues/issues/403

Re-introducing the feature to allow to switch field to group records by on a kanban view

https://github.com/user-attachments/assets/a552d3f2-7900-4771-b512-98cb99cdd8de


packages/twenty-front/src/modules/object-record/object-options-dropdown/components/ObjectOptionsDropdownRecordGroupsContent.tsx
```diff
@@ -75,6 +76,10 @@ export const ObjectOptionsDropdownRecordGroupsContent = () => {
     handleHideEmptyRecordGroupChange,
   } = useRecordGroupVisibility();
 
+  const { availableFieldsForKanban } = useGetAvailableFieldsForKanban();
```
charlesBochet: this naming seems outdated?

charlesBochet (approved):
Left comment + let's fix tests

PR #17017, opened 2026-01-08
Implemented SSE subscription mechanism on the frontend

lucasbordeau:
This PR is a follow-up of https://github.com/twentyhq/twenty/pull/16966 and implements a new mechanism to handle SSE events.

It creates only one event stream per browser tab, then use mutations to tell the backend which query to listen to, without re-mounting the event stream connexion.  

Then each event that comes from this unique subscription is then dispatched in a new JavaScript CustomEvent, per queryId, on which specific hooks add an event listener.

This PR introduces the generic tooling as well as the handling of update events on table.


packages/twenty-front/src/modules/object-record/record-table/virtualization/components/RecordTableVirtualizedOnObjectRecordEventsEffect.tsx
```diff
@@ -0,0 +1,164 @@
...
+        updatedRecordsWithUpdatedFieldsOnly.push(
+          Object.fromEntries(
+            updatedFields.map((fieldName) => {
+              return [fieldName, updatedRecord[fieldName]];
+            }),
+          ) as ObjectRecord,
+        );
+      }
+
+      registerObjectOperation(objectMetadataItem, {
```
charlesBochet, line 153: don't know if this is temporary but I feel we could directly add the logic of packages/twenty-front/src/modules/object-record/record-table/virtualization/components/RecordTableVirtualizedDataChangedEffect.tsx here and remove the RecordTableVirtualizedDataChangedEffect

packages/twenty-front/src/modules/sse-db-event/graphql/subscriptions/OnDbEventsRelatedToQueriesSubscription.ts
```diff
@@ -0,0 +1,25 @@
+import { gql } from '@apollo/client';
+
+export const ON_DB_EVENTS_RELATED_TO_QUERIES_SUBSCRIPTION = gql`
+  subscription OnDbEventsRelatedToQueries(
```
charlesBochet: not sure it helps to have relatedToQueries here (every where)

we could just have OnDbEvent({ eventStreamId } {
dbEvents {}

Just to clarify: if an event is matching several queries, the backend is sending it multiple times? otherwise, I would expect to have matchingQueryIds in the return

packages/twenty-front/src/modules/sse-db-event/graphql/subscriptions/OnDbEventsRelatedToQueriesSubscription.ts
```diff
@@ -0,0 +1,25 @@
...
+  ) {
+    onDbEventsRelatedToQueries(parameters: $parameters) {
+      eventStreamId
+      dbEventsWithRelatedQueryIds {
+        queryIds
+        dbEvent {
+          action
+          objectNameSingular
+          eventDate
+          record
```
charlesBochet: this should be batchAPI

charlesBochet (commented):
Looks very promising, left comments / questions

packages/twenty-front/src/modules/object-record/record-table/virtualization/components/RecordTableVirtualizedOnObjectRecordEventsEffect.tsx
lucasbordeau: No because it is used also if we change the data from another place in the same application.

packages/twenty-front/src/modules/sse-db-event/components/SSEQuerySubscribeEffect.tsx
```diff
@@ -0,0 +1,128 @@
...
+    boolean,
+    { input: AddQuerySubscriptionInput }
+  >(ADD_QUERY_TO_EVENT_STREAM_MUTATION, { client: apolloCoreClient });
+
+  const [removeQueryFromEventStream] = useMutation<
+    void,
+    { input: RemoveQueryFromEventStreamInput }
+  >(REMOVE_QUERY_FROM_EVENT_STREAM_MUTATION, { client: apolloCoreClient });
+
+  const requiredQueryListeners = useRecoilValue(requiredQueryListenersState);
```
charlesBochet, line 36: not totally against it but I'm not sure about the value TBH.

We could make this synchornously without having to write in a state and then read from the state and then compute create / remove

charlesBochet: why do we need it?

packages/twenty-front/src/modules/sse-db-event/utils/getObjectRecordEventsForQueryEventName.ts
```diff
@@ -0,0 +1,3 @@
+export const getObjectRecordEventsForQueryEventName = (queryId: string) => {
```
charlesBochet, line 1: naming is not clear: event here is DOM event I think

but I feel we could have all queries listening to the same eventName and filter after that. I'm challening the whole query stateful approach, looks to be a bit complex

charlesBochet (approved):

PR #17021, opened 2026-01-08
feat: add sorting on relation fields (Many-to-One)

FelixMalfait:
## Summary

This PR enables sorting records by fields of related objects. For example, sorting **People by their Company's name**.

### Before
Only scalar and composite fields could be sorted. Relation fields showed in the sort dropdown but produced errors.

### After
Many-to-One relation fields can now be sorted using the related object's **label identifier field** (e.g., Company's `name`).

---

### Frontend
- Added `RELATION` to sortable field types (restricted to `MANY_TO_ONE` relations)
- New `getOrderByForRelationField()` generates nested orderBy structures using the related object's label identifier
- Updated `turnSortsIntoOrderBy()` to handle relation fields by looking up related object metadata

### Backend
- Extended `GraphqlQueryOrderFieldParser.parse()` to detect nested relation ordering like `{ company: { name: 'AscNullsLast' } }`
- Returns `ParseOrderByResult` containing both `orderBy` conditions and `relationJoins` info
- Added LEFT JOINs for relation ordering in `applyOrderToBuilder()`
- Added `addRelationOrderColumnsToBuilder()` for TypeORM DISTINCT compatibility

### Tests
- Added unit tests for `filterSortableFieldMetadataItems`, `getOrderByForRelationField`, and `turnSortsIntoOrderBy`
- Added integration tests covering ascending/descending order and composite label identifiers

---

## TypeORM Bug Workaround

We encountered a significant TypeORM limitation when implementing this feature. When using `getMany()` with `ORDER BY` on joined relation columns, TypeORM generates a DISTINCT subquery that has specific requirements:

### Issue 1: Alias Parsing
TypeORM's `orderBy()` method fails with **"alias not found"** when using quoted SQL identifiers like `"company"."name"`. TypeORM internally expects unquoted property paths (e.g., `company.name`) for its alias resolution mechanism.

### Issue 2: setFindOptions Clears addSelect
`setFindOptions({ select })` **clears any previously added `addSelect()` columns**. This caused `"column distinctAlias.company_name does not exist"` errors because the relation columns needed for ORDER BY were being removed.

### Solution
We split the logic into two methods:
1. `applyOrderToBuilder()` - adds JOINs and ORDER BY (before `setFindOptions`)
2. `addRelationOrderColumnsToBuilder()` - adds relation columns for SELECT (AFTER `setFindOptions`)

This ensures the relation columns are present in the final SQL query's SELECT clause with the proper underscore aliases (`company_name`) that TypeORM's DISTINCT subquery expects.

**Related TypeORM issue**: https://github.com/typeorm/typeorm/issues/9921

---

## Screenshots/Demo

_Add screenshots if applicable_


packages/twenty-front/src/modules/object-metadata/utils/__tests__/filterSortableFieldMetadataItems.test.ts
```diff
@@ -0,0 +1,125 @@
...
+  it('should allow TEXT field type', () => {
+    const field = {
+      id: 'field-1',
+      name: 'name',
+      type: FieldMetadataType.TEXT,
+      isSystem: false,
+      isActive: true,
+    };
+
+    expect(filterSortableFieldMetadataItems(field as any)).toBe(true);
```
charlesBochet: any here
let's filterSortableFieldMetadataItems should Pick<>

packages/twenty-front/src/modules/object-metadata/utils/__tests__/filterSortableFieldMetadataItems.test.ts
```diff
@@ -0,0 +1,125 @@
+import { filterSortableFieldMetadataItems } from '@/object-metadata/utils/filterSortableFieldMetadataItems';
+import { FieldMetadataType, RelationType } from '~/generated-metadata/graphql';
+
+describe('filterSortableFieldMetadataItems', () => {
```
charlesBochet, line 4: nice :)

packages/twenty-front/src/modules/object-metadata/utils/filterSortableFieldMetadataItems.ts
```diff
@@ -9,5 +10,14 @@ export const filterSortableFieldMetadataItems = (field: FieldMetadataItem) => {
     field.type,
   );
 
-  return !isSystemField && isFieldActive && isFieldTypeSortable;
+  // Allow MANY_TO_ONE relations to be sorted (sorts by related object's label identifier)
```
charlesBochet: comment not useful

packages/twenty-front/src/modules/object-record/object-sort-dropdown/utils/__tests__/turnSortsIntoOrderBy.test.ts
```diff
@@ -165,4 +166,144 @@ describe('turnSortsIntoOrderBy', () => {
...
+          createdAt: '2021-01-01',
+          updatedAt: '2021-01-01',
+          isActive: true,
+          relation: {
+            type: RelationType.MANY_TO_ONE,
+            targetObjectMetadata: {
+              nameSingular: 'company',
+            },
+          },
+        } as unknown as FieldMetadataItem,
```
charlesBochet, line 221: if we could avoid this would be nice (same pb as any) but I guess we already have some tests doing it in the same file

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-order/utils/parse-composite-field-for-order.util.ts
```diff
@@ -38,7 +38,8 @@ export const parseCompositeFieldForOrder = (
         );
       }
 
-      const fullFieldName = `"${objectNameSingular}"."${fieldMetadata.name}${capitalize(subFieldKey)}"`;
+      // Use unquoted property path for TypeORM's getMany() alias resolution
```
charlesBochet: let's get rid of all these comments!

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-order/graphql-query-order.parser.ts
```diff
@@ -77,58 +86,220 @@ export class GraphqlQueryOrderFieldParser {
...
-
-            acc[`"${objectNameSingular}"."${columnName}"${orderByCasting}`] =
-              convertOrderByToFindOptionsOrder(
-                orderByDirection as OrderByDirection,
-                isForwardPagination,
-              );
+        // Handle nested relation ordering like { company: { name: 'AscNullsLast' } }
+        if (
+          isMorphOrRelationFlatFieldMetadata(fieldMetadata) &&
+          isObject(orderByDirection) &&
```
charlesBochet: this I'm not a big fan, we should pilot by metadata not by the shape of orderByDirection (isObject)

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-order/graphql-query-order.parser.ts
```diff
@@ -77,58 +86,220 @@ export class GraphqlQueryOrderFieldParser {
...
+            orderByDirection: orderByDirection as Record<string, unknown>,
+            isForwardPagination,
+          });
+
+          if (relationOrderResult) {
+            Object.assign(orderByConditions, relationOrderResult.orderBy);
+
+            // Add join if not already added
+            if (!addedJoinAliases.has(relationOrderResult.joinInfo.joinAlias)) {
+              relationJoins.push(relationOrderResult.joinInfo);
```
charlesBochet, line 103: btw, this should change the computed query complexity @etiennejouan FYI (usage protection)

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-order/graphql-query-order.parser.ts
```diff
@@ -77,58 +86,220 @@ export class GraphqlQueryOrderFieldParser {
...
-      },
-      {} as Record<string, OrderByCondition>,
+  private isOrderByDirection(value: unknown): value is OrderByDirection {
+    return (
+      typeof value === 'string' &&
+      Object.values(OrderByDirection).includes(value as OrderByDirection)
     );
   }
 
+  private parseRelationFieldOrder({
```
charlesBochet, line 151: this file is too long, we should split it

Note: all this code looks quite hacky.
I'm not against it if we have it well tested unitary

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-order/types/order-by.types.ts
```diff
@@ -0,0 +1,13 @@
+export type OrderByCondition = {
```
FelixMalfait: One export per file

packages/twenty-server/src/engine/api/common/common-query-runners/common-find-many-query-runner.service.ts
```diff
@@ -275,4 +309,31 @@ export class CommonFindManyQueryRunnerService extends CommonBaseQueryRunnerServi
       );
     }
   }
+
+  protected override computeQueryComplexity(
```
FelixMalfait, line 313: @etiennejouan could you please check?

packages/twenty-front/src/modules/apollo/optimistic-effect/utils/sortCachedObjectEdges.ts
```diff
@@ -4,8 +4,23 @@ import { type ReadFieldFunction } from '@apollo/client/cache/core/types/common';
...
 import { type RecordGqlOperationOrderBy } from '@/object-record/graphql/types/RecordGqlOperationOrderBy';
 import { type OrderBy } from '@/types/OrderBy';
-import { sortAsc, sortDesc, sortNullsFirst, sortNullsLast } from '~/utils/sort';
 import { isDefined } from 'twenty-shared/utils';
+import { sortAsc, sortDesc, sortNullsFirst, sortNullsLast } from '~/utils/sort';
+
+// Extracts the OrderBy direction from a nested orderBy structure
+// Handles up to 3 levels: field, field.subField, or relation.compositeField.subField
+const extractOrderByDirection = (value: unknown): OrderBy | null => {
+  if (typeof value === 'string') {
```
charlesBochet: use isNonEmptyString

packages/twenty-front/src/modules/apollo/optimistic-effect/utils/sortCachedObjectEdges.ts
```diff
@@ -36,15 +57,31 @@ export const sortCachedObjectEdges = ({
...
+          subFieldValue as Reference | StoreObject,
+        ) ?? null;
+      return subSubFieldValue;
+    }
+
+    return subFieldValue as string | number | null;
   };
 
-  const orderByValue = orderBySubFieldValue || (orderByFieldValue as OrderBy);
+  const orderByValue =
```
charlesBochet: naming: value the left, direction on the right

packages/twenty-front/src/modules/apollo/optimistic-effect/utils/sortCachedObjectEdges.ts
```diff
@@ -4,8 +4,23 @@ import { type ReadFieldFunction } from '@apollo/client/cache/core/types/common';
...
 import { isDefined } from 'twenty-shared/utils';
+import { sortAsc, sortDesc, sortNullsFirst, sortNullsLast } from '~/utils/sort';
+
+// Extracts the OrderBy direction from a nested orderBy structure
+// Handles up to 3 levels: field, field.subField, or relation.compositeField.subField
+const extractOrderByDirection = (value: unknown): OrderBy | null => {
+  if (typeof value === 'string') {
+    return value as OrderBy;
+  }
+  if (typeof value === 'object' && value !== null) {
```
charlesBochet, line 17: same here

packages/twenty-front/src/modules/apollo/optimistic-effect/utils/sortCachedObjectEdges.ts
charlesBochet: still not a big fan of getting the direction by checking the shape of the variable.

I would prefer to have something like
{ company: { name: 'Asc' }, email: 'Desc' } => we transform based on field metadata and end up with:
[{ fieldMetadataId: xxx, direction, 'Asc' }, { fieldMetadataId: yyy, direction 'Desc' }]

and then we don't have to have this kind of assertion that will break at some point as nothing is properly typed.

As long as we have unit tests on this that's acceptable, but this is not robust

packages/twenty-front/src/modules/apollo/optimistic-effect/utils/sortCachedObjectEdges.ts
```diff
@@ -36,15 +57,31 @@ export const sortCachedObjectEdges = ({
...
         fieldValue as Reference | StoreObject,
       ) ?? null;
 
-    return subFieldValue;
+    // Handle 3-level nesting (relation.compositeField.subField)
+    if (isDefined(subFieldValue) && isDefined(orderBySubSubFieldName)) {
+      const subSubFieldValue =
+        readCacheField<string | number | null>(
+          orderBySubSubFieldName,
+          subFieldValue as Reference | StoreObject,
```
charlesBochet, line 79: a bit confusing that we are treating both composite and relations together

relation => should be a ref indeed in cache
composite => should be an object

packages/twenty-front/src/modules/object-metadata/utils/__tests__/getOrderByForRelationField.test.ts
```diff
@@ -0,0 +1,150 @@
...
+          id: 'name-field-id',
+          name: 'name',
+          type: FieldMetadataType.TEXT,
+          isActive: true,
+        },
+      ],
+    };
+
+    const result = getOrderByForRelationField(
+      field as any,
```
charlesBochet: still not a fan of this as any, we should pick<> here

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-order/types/order-by-condition.type.ts
```diff
@@ -0,0 +1,4 @@
+export type OrderByCondition = {
```
charlesBochet: naming looks a bit suspect: condition?

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-order/utils/get-optional-order-by-casting.util.ts
```diff
@@ -0,0 +1,16 @@
...
+import { type FlatFieldMetadata } from 'src/engine/metadata-modules/flat-field-metadata/types/flat-field-metadata.type';
+
+export const getOptionalOrderByCasting = (
+  fieldMetadata: Pick<FlatFieldMetadata, 'type'>,
+): string => {
+  if (
+    fieldMetadata.type === FieldMetadataType.SELECT ||
+    fieldMetadata.type === FieldMetadataType.MULTI_SELECT
+  ) {
+    return '::text';
```
charlesBochet, line 12: I'm a bit surprise this is part of "graphql tooling" regarding folders. It looks more low level postgres stuff

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-order/utils/is-order-by-direction.util.ts
```diff
@@ -0,0 +1,10 @@
+import { OrderByDirection } from 'twenty-shared/types';
+
+export const isOrderByDirection = (
+  value: unknown,
+): value is OrderByDirection => {
+  return (
+    typeof value === 'string' &&
```
charlesBochet: use isNonEmptyString

charlesBochet (approved):
LGTM, I have added a few comments but they are not that importants. I think the tech debt introduced is not hurting maintenance too much and code quality is overall good

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-order/utils/get-optional-order-by-casting.util.ts
FelixMalfait: This was already in the code I just moved it

FelixMalfait: I get your point but I think it might be challenging to move this at a lower-level, I don't see an easy path

PR #17022, opened 2026-01-08
[SSE] Add backend for SSE subscriptions

thomtrp:
- moved a few gql types to twenty shared to re-use in server
- added a new endpoint, onEventSubscription that expect a streamId to create a connection
- two new endpoints to store queries in Redis
- updated the existing batch channel to directly use object record type


packages/twenty-server/src/engine/subscriptions/dtos/object-record-event.dto.ts
```diff
@@ -0,0 +1,36 @@
+import { Field, ObjectType } from '@nestjs/graphql';
+
+import GraphQLJSON from 'graphql-type-json';
+
+@ObjectType('ObjectRecordEventProperties')
+export class ObjectRecordEventPropertiesDTO {
```
Weiko: Could we have a dedicated file for each dto?

packages/twenty-server/src/engine/subscriptions/event-stream.service.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    return matchedQueryIds;
+  }
+
+  private isQueryMatchingEvent(
+    operationSignature: RecordGqlOperationSignature,
+    event: ObjectRecordSubscriptionEvent,
+  ): boolean {
+    // to be improved
+    return operationSignature.objectNameSingular === event.objectNameSingular;
```
Weiko, line 107: Don't we need to know if it's an update/create/delete etc?

Weiko (commented):
Questions:
How do we handle permissions here? It seems users can listen to anything regardless of permissions?
Followup question: How do we check which users own which eventStream? I'm assuming the cross-workspace security is handled but as a user from the same workspace I can subscribe to an eventStream created by another user and receive events on objects I don't have access to?

packages/twenty-server/src/engine/subscriptions/event-stream.service.ts
```diff
@@ -0,0 +1,109 @@
...
+    return new Map(Object.entries(data || {}));
+  }
+
+  async matchQueriesWithEvent(
+    queries: Map<string, RecordGqlOperationSignature>,
+    event: ObjectRecordSubscriptionEvent,
+  ): Promise<string[]> {
+    const matchedQueryIds: string[] = [];
+
+    for (const [queryId, operationSignature] of queries.entries()) {
```
Weiko, line 93: Do we expect a lot of queries? Should we actually store a map of objectNameSingular instead?

Weiko (commented):
How do we handle cache growth? Abandoned subscriptions / closed socket does not seem to properly cleanup cache or am I missing something (appart from calling removeQueryFromEventStream which is not enough imho)? Should we set a TTL at least?
Where do we generate the eventStreamId? Is it client-side? Is it safe to collisions?

packages/twenty-server/src/engine/subscriptions/event-stream.service.ts
thomtrp: we return the full event to frontend

packages/twenty-server/src/engine/subscriptions/event-stream.service.ts
thomtrp: I don't expect more than 10-20 queries per connection

thomtrp:
@Weiko planning on adding TTL but not for that first version to unblock frontend
Great catch for permission!. I think we also need the userId added to the event stream it he does not get notified. Will also add it to a separated PR.

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.resolver.ts
```diff
@@ -104,8 +114,94 @@ export class WorkspaceEventEmitterResolver {
...
+            event: eventWithObjectName,
+          });
+        }
+      }
+
+      return { eventStreamId, eventWithQueryIdsList };
+    },
+  })
+  onEventSubscription(
+    @Args('eventStreamId') _: string,
```
Weiko, line 170: Why are we not using eventStreamId for the subscription?

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.resolver.ts
```diff
@@ -104,8 +114,94 @@ export class WorkspaceEventEmitterResolver {
...
+    @AuthWorkspace() workspace: WorkspaceEntity,
+  ) {
+    return this.subscriptionService.subscribe({
+      channel: SubscriptionChannel.WORKSPACE_EVENT_BATCH_CHANNEL,
```
Weiko, line 174: Oh ok, I think we have an N+1 issue here. I will ping you

Weiko (commented):
More comments!

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.resolver.ts
thomtrp: it is used. When the subscription is established, NestJS keeps track of the original arguments for that connection. So we can use it when we receive an event

Weiko (approved):
LGTM but we need to address the security/perf issues in the next PR

PR #17031, opened 2026-01-08
fix(frontend): filter null records in record index query

eslteacher902010:
Filters out null records returned from `useFindManyRecords` at the record index query boundary.

This prevents frontend crashes in record list views (e.g. notes and tasks) when unexpected null entries appear in record collections. The change is intentionally scoped and does not affect downstream components or API behavior.


charlesBochet:
@eslteacher902010 I'm not a fan of this code change! The API should not return null records so I don't understand the change

What bug are we trying to solve?

Closing the PR but we can keep discussing here

PR #17038, opened 2026-01-09
Made the code more resilient to stale value prop in date filter

lucasbordeau:
Fixes : https://github.com/twentyhq/twenty/issues/17035

There was a bad pattern which risked introduce this bug in `turnRecordFilterIntoRecordGqlOperationFilter` after recent refactor of date logic and date filters, which didn't create any bug during dev time because it wasn't tested with value combinations that existed before the refactor.

Code has been re-organized to make it resilient to any wrong value combination.

Also fixed a small bug that appeared during QA with `DATE` filter type, which was blocking the save button from disappearing after a save.

### DIFF
--- a/packages/twenty-shared/src/utils/filter/turnRecordFilterIntoGqlOperationFilter.ts
+++ b/packages/twenty-shared/src/utils/filter/turnRecordFilterIntoGqlOperationFilter.ts
@@ -168,7 +168,10 @@ export const turnRecordFilterIntoRecordGqlOperationFilter = ({
           );
       }
     case 'DATE': {
-      if (recordFilter.operand === RecordFilterOperand.IS_RELATIVE) {
+      const itsARelativeDateFilter =
+        recordFilter.operand === RecordFilterOperand.IS_RELATIVE;
+
+      if (itsARelativeDateFilter) {
         const relativeDateFilterValue = resolveRelativeDateFilterStringified(
           recordFilter.value,
         );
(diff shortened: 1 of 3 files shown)


packages/twenty-shared/src/utils/filter/turnRecordFilterIntoGqlOperationFilter.ts
```diff
@@ -168,7 +168,10 @@ export const turnRecordFilterIntoRecordGqlOperationFilter = ({
           );
       }
     case 'DATE': {
-      if (recordFilter.operand === RecordFilterOperand.IS_RELATIVE) {
```
Weiko, line 171: Seems there is a few duplicated logic between date and datetime in this file, what do you think?

Weiko (approved):
Left a comment about a potential refactoring but non blocking
LGMT! @lucasbordeau

packages/twenty-shared/src/utils/filter/turnRecordFilterIntoGqlOperationFilter.ts
lucasbordeau: I'll do it in a follow-up PR.

lucasbordeau: @Weiko I did look at what could be refactored and there seems to be enough subtle differences between how plain date and date times are handled for each operator to justify this structure that looks like it could be refactored. I think it's not worth it. Charles mentioned earlier that we could split this big function into smaller functions for each field type though, which would be interesting to do.

Issue #15231, opened 2025-10-21
Users with view-only permissions cannot use workflows


mlamboia:
### Bug Description

When a user only has view permissions for workflows in Twenty, they are unable to use a workflow, even though the workflow is visible and active.

#### Steps to reproduce:

1 - Create and activate a workflow (for example, with a manual trigger).

2 - Set the workflow permissions to view only.
<img width="513" height="243" alt="Image" src="https://github.com/user-attachments/assets/501ae6d7-622e-45a3-972e-751ffa363d5f" />

3 - Enable “Manage workflow” and open the workflows page — the workflow is visible and can be used.
<img width="443" height="36" alt="Image" src="https://github.com/user-attachments/assets/4f4097f2-0194-4df3-a345-8b2efb987544" />

<img width="404" height="79" alt="Image" src="https://github.com/user-attachments/assets/bc38e188-2c96-4f05-b362-15bf505aaa70" />

4 - Disable “Manage workflow” and open the workflows page again — the workflow is not visible and cannot be used.
<img width="443" height="36" alt="Image" src="https://github.com/user-attachments/assets/59f815dd-cc29-44af-8d10-0f54332aee22" />

<img width="404" height="79" alt="Image" src="https://github.com/user-attachments/assets/920aaf0a-b260-48b7-b0c0-5db1fcdffbf6" />

#### Expected behavior:
Users with view-only access should still be able to use active workflows, as long as they have the necessary permissions to view them. They should not be able to edit, delete, or destroy workflows.

#### Current behavior:
Users with view-only permissions cannot use workflows at all.

#### Technical inputs

* This issue likely relates to the permission handling logic for workflows.

* The visibility logic works correctly (workflows are shown), but the execution logic is probably tied to the “manage workflow” permission instead of the view permission.

pArv-2029:
Hi @StephanieJoly4 @mlamboia  , I'm a contributor for Hacktoberfest and would love to work on this bug. Could you please assign it to me

StephanieJoly4:
thank you @pArv-2029, we might need to wait for @Bonapara and the team what needs to done from a product perspective before diving right in

Bonapara:
Manual trigger "Run" shouldn’t be gated by the `manage workflows `permission. We should pull manual execution out of that permission. (No specific permission needed for manual trigger at that stage)

Bhoomaahamso:
Hi @Bonapara, I’d like to help with this issue.

[...]

PR #17039, opened 2026-01-09
Introduce WorkflowManualTrigger entity to decouple workflow execution from permissions

abdulrahmancodes:
## Summary

This is the first of 3 PRs addressing [#15231](https://github.com/twentyhq/twenty/issues/15231). It introduces a `WorkflowManualTrigger` entity to decouple manual workflow execution from `WorkflowVersion` table visibility permissions.

## Next Steps

PR 2: Create workflow-specific roles and copy permissions
PR 3: Remove "Manage workflow" settings permission entirely


packages/twenty-server/src/modules/workflow/common/standard-objects/workflow-manual-trigger.workspace-entity.ts
```diff
@@ -0,0 +1,83 @@
...
+@WorkspaceEntity({
+  standardId: STANDARD_OBJECT_IDS.workflowManualTrigger,
+  namePlural: 'workflowManualTriggers',
+  labelSingular: msg`Workflow Manual Trigger`,
+  labelPlural: msg`Workflow Manual Triggers`,
+  description: msg`A workflow manual trigger`,
+  icon: STANDARD_OBJECT_ICONS.workflowManualTrigger,
+})
+@WorkspaceIsSystem()
+export class WorkflowManualTriggerWorkspaceEntity extends BaseWorkspaceEntity {
```
FelixMalfait: I don't think we should include "workflow" in the name, because like other triggers this primitive will be re-used outside the workflow context. Twenty applications will be able to declare commands like that (I'm an app developer I create a "Twenty App", I want to be able to register something in the command menu -> I'm declaring my app as code, not using the Workflow product which is a drag-n-drop UI for end-user to build their thing)

packages/twenty-server/src/modules/workflow/common/standard-objects/workflow-manual-trigger.workspace-entity.ts
```diff
@@ -0,0 +1,83 @@
...
+  workflowName: string;
+
+  @WorkspaceField({
+    standardId: WORKFLOW_MANUAL_TRIGGER_STANDARD_FIELD_IDS.settings,
+    type: FieldMetadataType.RAW_JSON,
+    label: msg`Settings`,
+    description: msg`Manual trigger settings (icon, isPinned, availability, outputSchema)`,
+  })
+  @WorkspaceIsFieldUIReadOnly()
+  settings: ManualTriggerSettings;
```
FelixMalfait: Let's no do JSON here but do a nice flat structure. You tried to migrate things as-is without thinking too much about why we're doing this and the possibilities it enables. We should have a flag structure here with name, icon, etc as first-class fields

FelixMalfait: Maybe label more than name actually if we are consistent with our naming in other places in the app (label is user-facing while name is more of a key/slug identifier)

packages/twenty-front/src/modules/workflow/hooks/useWorkflowManualTriggers.ts
```diff
@@ -0,0 +1,68 @@
...
+  objectNameSingular: string,
+): boolean => {
+  const availability = manualTrigger.settings?.availability;
+
+  if (!isDefined(availability)) {
+    return false;
+  }
+
+  if (
+    availability.type === 'SINGLE_RECORD' ||
```
FelixMalfait: If we expose things cleanly (flat field structure not json) we'd get this enum from the graphql codegen

packages/twenty-front/src/modules/object-metadata/utils/mapFieldMetadataToGraphQLQuery.ts
```diff
@@ -135,6 +135,10 @@ ${mapObjectMetadataToGraphQLQuery({
     fieldType === FieldMetadataType.RELATION &&
     fieldMetadata.relation?.type === RelationType.MANY_TO_ONE
   ) {
+    if (gqlField === fieldMetadata.settings?.joinColumnName) {
```
FelixMalfait, line 138: Why move this? It's not obvious when reading

packages/twenty-front/src/modules/workflow/hooks/useWorkflowManualTriggers.ts
```diff
@@ -0,0 +1,68 @@
...
+    });
+
+  const records = isDefined(objectMetadataItem)
+    ? manualTriggers.filter((manualTrigger) =>
+        isObjectSpecificTrigger(manualTrigger, objectMetadataItem.nameSingular),
+      )
+    : manualTriggers.filter(isGlobalTrigger);
+
+  return { records };
+};
```
FelixMalfait, line 156: It feels risky to have the migration command and deprecate this in the same PR, I think you should find a way to make it a bit smoother deployment (e.g. both with a flag?).

Otherwise if your command take 2 hours to run on the thousands of workspaces then it means we might break existing commands for that period

packages/twenty-shared/src/metadata/standard-object-ids.ts
```diff
@@ -49,4 +49,5 @@ export const STANDARD_OBJECT_IDS = {
   workflowVersion: '20202020-d65d-4ab9-9344-d77bfb376a3d',
   workspaceMember: '20202020-3319-4234-a34c-82d5c0e881a6',
   workflowAutomatedTrigger: '20202020-3319-4234-a34c-7f3b9d2e4d1f',
```
FelixMalfait, line 51: I see this was named with workflow too so I understand why you went that way but this was a mistake, those primitives should not be exclusively tied to workflows since they will be used in apps (@thomtrp fyi I think you're aligned on this)

packages/twenty-server/src/engine/workspace-manager/twenty-standard-application/utils/field-metadata/compute-workflow-manual-trigger-standard-flat-field-metadata.util.ts
```diff
@@ -0,0 +1,190 @@
...
+        onDelete: RelationOnDeleteAction.CASCADE,
+        joinColumnName: 'workflowVersionId',
+      },
+    },
+    standardObjectMetadataRelatedEntityIds,
+    dependencyFlatEntityMaps,
+    twentyStandardApplicationId,
+    now,
+  }),
+  workflow: createStandardRelationFieldFlatMetadata({
```
FelixMalfait: Do we need a relation to both workflow and workflow version? I'm usually not a fan of this, since it can be retrieved dynamically. By copying the data you create risks of inconsistency. But in some rare cases doing what you did is justified (for perf reason etc)

packages/twenty-front/src/modules/action-menu/actions/record-actions/run-workflow-actions/hooks/useRunWorkflowRecordActions.tsx
```diff
@@ -7,17 +7,24 @@ import { type ObjectMetadataItem } from '@/object-metadata/types/ObjectMetadataI
 import { recordStoreFamilyState } from '@/object-record/record-store/states/recordStoreFamilyState';
 import { useSnackBar } from '@/ui/feedback/snack-bar-manager/hooks/useSnackBar';
 import { useRecoilComponentValue } from '@/ui/utilities/state/component-state/hooks/useRecoilComponentValue';
-import { useActiveWorkflowVersionsWithManualTrigger } from '@/workflow/hooks/useActiveWorkflowVersionsWithManualTrigger';
 import { useRunWorkflowVersion } from '@/workflow/hooks/useRunWorkflowVersion';
-
```
FelixMalfait, line 12: Can you please merge recent main? I see quite a lot of conflicts already

charlesBochet:
@abdulrahmancodes closing since this is staled, let's re-open when working on it :)

packages/twenty-front/src/modules/object-metadata/utils/mapFieldMetadataToGraphQLQuery.ts
abdulrahmancodes: This check handles the case where we are querying the join column field (e.g. `workflowVersionId`) rather than the relation field. Without this early return, the code would incorrectly try to build a nested query for what should be a simple scalar field.

packages/twenty-front/src/modules/workflow/hooks/useWorkflowManualTriggers.ts
abdulrahmancodes: Would splitting into 2 PRs (BE+migration first, then FE) work instead of a feature flag?

packages/twenty-server/src/engine/workspace-manager/twenty-standard-application/utils/field-metadata/compute-workflow-manual-trigger-standard-flat-field-metadata.util.ts
abdulrahmancodes: yeah, you're right. We only need `workflowId` as denormalized field. We don't query manualTriggers by workflow. I'll remove the workflow relation and keep just the flat field for convenience. The full workflow can be accessed via `workflowVersion.workflow` if ever needed

packages/twenty-shared/src/metadata/standard-object-ids.ts
```diff
@@ -49,4 +49,5 @@ export const STANDARD_OBJECT_IDS = {
...
+  manualTrigger: '20202020-3319-4234-a34c-8f2a1b3c4d5e',
```
FelixMalfait, line 52: I just realized you made it a Workspace Schema object, this is not what we wanted, the goal was to move it to the core schema so that it becomes part of the toolbox for apps (metadata not data)

PR #17040, opened 2026-01-09
Fix cleaning command

etiennejouan:
Previous behavior:

Used billingSubscription.updatedAt (when the subscription record was last modified)
Problem: updatedAt changes for ANY update, not just status changes, making it unreliable for tracking when payment problems started

Current behavior:

Uses currentPeriodStart when subscription status is Unpaid or Canceled
Why it works: WORKSPACE_INACTIVE_DAYS_BEFORE_SOFT_DELETION is shorter than the minimum billing interval (1 month). Then if a workspace is unpaid for the entire current billing period, it will always exceed the deletion threshold before the next period starts

Ideal fix:
Track workspace.suspendedAt explicitly, giving you the exact timestamp of when payment problems began, regardless of billing periods.

### DIFF
--- a/packages/twenty-server/src/engine/workspace-manager/workspace-cleaner/services/cleaner.workspace-service.ts
+++ b/packages/twenty-server/src/engine/workspace-manager/workspace-cleaner/services/cleaner.workspace-service.ts
@@ -70,25 +70,34 @@ export class CleanerWorkspaceService {
       );
   }
 
-  async computeWorkspaceBillingInactivity(
+  async computeDaysSinceSubscriptionUnpaid(
     workspace: WorkspaceEntity,
   ): Promise<number> {
     try {
       const lastSubscription =
         await this.billingSubscriptionRepository.findOneOrFail({
-          where: { workspaceId: workspace.id },
+          where: {
+            workspaceId: workspace.id,
+          },
           order: { updatedAt: 'DESC' },
         });
 
-      const daysSinceBillingInactivity = differenceInDays(
+      if (
+        lastSubscription.status !== SubscriptionStatus.Unpaid &&
+        lastSubscription.status !== SubscriptionStatus.Canceled
+      ) {
+        throw new Error();
+      }
+
+      const daysSinceSubscriptionUnpaid = differenceInDays(
         new Date(),
-        lastSubscription.updatedAt,
+        lastSubscription.currentPeriodStart,
       );
 
-      return daysSinceBillingInactivity;
+      return daysSinceSubscriptionUnpaid;
     } catch {
       throw new WorkspaceCleanerException(
-        `No billing subscription found for workspace ${workspace.id} ${workspace.displayName}`,
+        `No cancelled or unpaid billing subscription found for workspace ${workspace.id} ${workspace.displayName}`,
         WorkspaceCleanerExceptionCode.BILLING_SUBSCRIPTION_NOT_FOUND,
       );
     }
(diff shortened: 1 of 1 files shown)


packages/twenty-server/src/engine/workspace-manager/workspace-cleaner/services/cleaner.workspace-service.ts
```diff
@@ -70,25 +70,34 @@ export class CleanerWorkspaceService {
...
+          },
           order: { updatedAt: 'DESC' },
         });
 
-      const daysSinceBillingInactivity = differenceInDays(
+      if (
+        lastSubscription.status !== SubscriptionStatus.Unpaid &&
+        lastSubscription.status !== SubscriptionStatus.Canceled
+      ) {
+        throw new Error();
```
Weiko: Could we throw a custom billing error? (even if it's not user facing)

packages/twenty-server/src/engine/workspace-manager/workspace-cleaner/services/cleaner.workspace-service.ts
```diff
@@ -70,25 +70,34 @@ export class CleanerWorkspaceService {
...
+      }
+
+      const daysSinceSubscriptionUnpaid = differenceInDays(
         new Date(),
-        lastSubscription.updatedAt,
+        lastSubscription.currentPeriodStart,
```
Weiko, line 96: This is not clear to me why it works

packages/twenty-server/src/engine/workspace-manager/workspace-cleaner/services/cleaner.workspace-service.ts
```diff
@@ -70,25 +70,34 @@ export class CleanerWorkspaceService {
...
       );
```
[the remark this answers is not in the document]
etiennejouan: subscription became unpaid at period start when paiement is proceeded

packages/twenty-server/src/engine/workspace-manager/workspace-cleaner/services/cleaner.workspace-service.ts
```diff
@@ -70,25 +70,34 @@ export class CleanerWorkspaceService {
...
 
-      return daysSinceBillingInactivity;
+      return daysSinceSubscriptionUnpaid;
     } catch {
       throw new WorkspaceCleanerException(
```
prastoin, line 101: ***Remark:*** we're swallowing this exception grain
```ts
      if (
        lastSubscription.status !== SubscriptionStatus.Unpaid &&
        lastSubscription.status !== SubscriptionStatus.Canceled
      ) {
        throw new Error();
      }
```

packages/twenty-server/src/engine/workspace-manager/workspace-cleaner/services/cleaner.workspace-service.ts
Weiko: Ok, I missed your description!

packages/twenty-server/src/engine/workspace-manager/workspace-cleaner/services/cleaner.workspace-service.ts
charlesBochet: should we throw here?
if yes, the method should be named orThrow

if we throw we should add an error message, ideally custom exception but that's acceptable to keep Error

etiennejouan: I thought this error would be catch in try/catch

etiennejouan: If workspace is suspended, we should not have other subscription status. The error message is the one in try/catch. I'm renaming it

charlesBochet (approved):

PR #17042, opened 2026-01-09
[DASHBOARDS] Allow dashboards to be restored

### LINKED ISSUE #2033: Add the ability to see soft deleted dashboards and to restore them.

bosiraphael:
This PR introduces a few changes:
- Add three actions: see deleted dashboards, destroy dashboard and restore dashboard
- Remove the soft delete and restore on all the page layout entities
- Cascade the destruction of a dashboard to a page layout

Video QA:

https://github.com/user-attachments/assets/ab993b11-dd9c-4e88-880c-92691a521cc2


charlesBochet (approved):

packages/twenty-server/src/modules/dashboard/query-hooks/dashboard-destroy-many.pre-query.hook.ts
```diff
@@ -0,0 +1,40 @@
...
+  async execute(
+    authContext: AuthContext,
+    _objectName: string,
+    payload: DestroyManyResolverArgs<{ id: { in: string[] } }>,
+  ): Promise<DestroyManyResolverArgs<{ id: { in: string[] } }>> {
+    const workspace = authContext.workspace;
+
+    assertIsDefinedOrThrow(workspace, WorkspaceNotFoundDefaultError);
+
+    await this.dashboardToPageLayoutSyncService.destroyPageLayoutsForDashboards(
```
Weiko, line 31: This could lead to orphan pageLayout if this part fails / event is lost but I don't think we can do better with current devX (for the first case we can add a try catch but that won't be fully bullet proof)

packages/twenty-server/src/modules/dashboard/services/dashboard-to-page-layout-sync.service.ts
```diff
@@ -0,0 +1,86 @@
...
+          );
+
+        const dashboards = await dashboardRepository.find({
+          where: {
+            id: In(dashboardIds),
+          },
+          withDeleted: true,
+        });
+
+        for (const dashboard of dashboards) {
```
Weiko, line 72: nit: perf: could use a destroyMany for better perf

packages/twenty-server/src/engine/metadata-modules/page-layout-tab/controllers/page-layout-tab.controller.ts
```diff
@@ -83,15 +89,22 @@ export class PageLayoutTabController {
...
+    return this.pageLayoutTabService.update({
+      id,
+      workspaceId: workspace.id,
+      updateData: input,
+    });
   }
 
   @Delete(':id')
   @UseGuards(SettingsPermissionGuard(PermissionFlagType.LAYOUTS))
   async delete(
```
Weiko, line 101: should it be named "destroy" then? See what's standard with REST first

Weiko (commented):
LGTM

packages/twenty-server/src/modules/dashboard/query-hooks/dashboard-destroy-many.pre-query.hook.ts
bosiraphael: Yes indeed

PR #17043, opened 2026-01-09
Fix MS Office preview for private/local URLs

### LINKED ISSUE #16900: Can't see file preview
**Scenario:**
1. Log in
2. Go to People records > any record > Files
3. Upload file (.doc, .docx, .ppt, .pptx, .xls, .xlsx)

**Actual:** Some file types can't be previewed

Image

**Expected:** All previewable file types can be previewed

FelixMalfait:
## Summary

Fixes #16900 - MS Office documents (doc, docx, ppt, pptx, xls, xlsx, odt) cannot be previewed when hosted on local/private networks.

## Problem

Microsoft's Office Online viewer (`view.officeapps.live.com`) requires documents to be publicly accessible from the internet. For self-hosted Twenty instances or local development, files are not reachable by Microsoft's servers, causing the viewer to fail with "An error occurred".

## Solution

Detect private/local URLs upfront and show a helpful message with a download button instead of letting the viewer fail silently.

**Detected private URLs:**
- `localhost`, `127.0.0.1`, `[::1]`
- Private IP ranges: `10.x.x.x`, `172.16-31.x.x`, `192.168.x.x`
- Local domain patterns: `.local`, `.localhost`, `.internal`

## Alternatives Explored (that don't work)

We investigated more sophisticated detection approaches:

1. **postMessage events** - Microsoft's viewer doesn't send any error messages we can listen to
2. **Fetching the embed URL** - Blocked by CORS (Microsoft doesn't set `Access-Control-Allow-Origin`)
3. **Reading iframe content** - Cross-origin restrictions prevent JavaScript access

The URL-based detection is the most reliable approach for catching the common cases where preview will definitely fail.

## Testing

1. Upload an Office file (docx, pptx, xlsx) on a local Twenty instance
2. Try to preview it
3. Should see "This file cannot be previewed because it is hosted locally" with a Download button


FelixMalfait:
Screenshot 2026-01-09 at 14 51 27


This is really not ideal as it won't catch all cases, but I guess it should still reduce the support volume

PR #17045, opened 2026-01-09
Improve application ast 2

martmull:
fix some issues with cli tools

### DIFF
--- a/packages/twenty-sdk/src/application/functions/define-function.ts
+++ b/packages/twenty-sdk/src/application/functions/define-function.ts
@@ -35,12 +35,8 @@ export const defineFunction = <T extends FunctionConfig>(config: T): T => {
     throw new Error('Function must have a handler');
   }
 
-  if (!config.triggers || config.triggers.length === 0) {
-    throw new Error('Function must have at least one trigger');
-  }
-
   // Validate each trigger
-  for (const trigger of config.triggers) {
+  for (const trigger of config.triggers ?? []) {
     if (!trigger.universalIdentifier) {
       throw new Error('Each trigger must have a universalIdentifier');
     }
--- a/packages/twenty-sdk/src/cli/utils/load-manifest.ts
+++ b/packages/twenty-sdk/src/cli/utils/load-manifest.ts
@@ -153,7 +153,10 @@ const loadSources = async (appPath: string): Promise<Sources> => {
   const sources: Sources = {};
 
   // Get all TypeScript files in src/ folder
-  const tsFiles = await loadFiles(['src/**/*.ts'], appPath);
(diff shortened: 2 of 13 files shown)


packages/twenty-sdk/src/application/functions/define-function.ts
```diff
@@ -35,12 +35,8 @@ export const defineFunction = <T extends FunctionConfig>(config: T): T => {
     throw new Error('Function must have a handler');
   }
 
-  if (!config.triggers || config.triggers.length === 0) {
```
Weiko, line 38: Why removing this validation? It seemed right or am I missing something?

packages/twenty-sdk/src/cli/utils/load-manifest.ts
```diff
@@ -63,7 +63,10 @@ const loadFiles = async (
  * Load all object definitions from src/app/ (any *.object.ts file).
  */
 const loadObjects = async (appPath: string): Promise<ObjectManifest[]> => {
-  const objectFiles = await loadFiles(['src/app/**/*.object.ts'], appPath);
+  const objectFiles = await loadFiles(
+    ['src/app/**/*.object.ts', 'generated/**/**.ts'],
```
Weiko: Why do we load generated classes for the manifest?

packages/twenty-sdk/src/cli/utils/validate-manifest.ts
```diff
@@ -242,13 +242,6 @@ const validateFunctions = (
       });
     }
 
-    if (!fn.triggers || fn.triggers.length === 0) {
-      errors.push({
-        path: fnPath,
-        message: 'Function must have at least one trigger',
```
Weiko, line 248: same

Weiko (commented):
Left some comments @martmull

packages/twenty-sdk/src/application/functions/define-function.ts
martmull: no we should be able to sync function without triggers I think, it avoid frustrating dev experience with an error after yarn sync after yarn create-entity function

packages/twenty-sdk/src/cli/utils/load-manifest.ts
martmull: we need it right now to be able to use the generated Twenty client in functions. Soon we will generate directly serverside and we won't need to do it client side

packages/twenty-sdk/src/cli/utils/validate-manifest.ts
martmull: see response above

Weiko (approved):
LGTM

PR #17046, opened 2026-01-09
Fix sort direction toggle when clicking on existing sort

gayanMatch:
Closes #17032

  Summary

  When a sort is already applied to a column, clicking "Sort" from the column header dropdown now toggles the sort direction (ASC → DESC or DESC → ASC) instead of doing nothing.

  Problem

  Previously, clicking "Sort" on a column that already had a sort applied would always try to create a new sort with ASC direction. Since the upsertRecordSort function updates an existing sort with the same field, this effectively did nothing when the sort was already ASC—which felt unintuitive.

  Solution

  Modified useHandleToggleColumnSort to:
  - Check if a sort already exists for the clicked field
  - If it exists, toggle the direction
  - If it doesn't exist, create a new sort with ASC direction (existing behavior)

---

Contribution by Gittensor, see my contribution statistics at https://gittensor.io/miners/details?githubId=132382032


packages/twenty-front/src/modules/object-record/record-index/hooks/useHandleToggleColumnSort.ts
```diff
@@ -24,6 +26,10 @@ export const useHandleToggleColumnSort = ({
 
   const { upsertRecordSort } = useUpsertRecordSort();
 
+  const currentRecordSorts = useRecoilComponentValue(
```
charlesBochet: we should not subscribe to the state using useRecoilComponentValue

this makes the hook reactive to this state which the caller will not expect as we are implementing a sync action.

Instead, make handleToggleColumnSort a recoilCallback and read existing sorts from the snapshot, they are many examples in the code

charlesBochet (changes requested):
Thanks @gayanMatch, I've left an important comment

gayanMatch:
I updated my PR. Now it's using useRecoilCallback like it does in useUpsertRecordSort.

charlesBochet (approved):
Thank you!

PR #17049, opened 2026-01-09
feat: drop empty states and instead render nothing

Devessier:
Dropping the empty state as currently seen in production.

## Before

CleanShot 2026-01-09 at 16 09 37@2x

## After

https://github.com/user-attachments/assets/32566a17-d5b6-4d62-80d5-144ba0050276


bosiraphael (approved):
LGTM

PR #17055, opened 2026-01-09
Center lock icon when widget isn't accessible

Devessier:
## Before

CleanShot 2026-01-09 at 18 11 43@2x

## After

CleanShot 2026-01-09 at 18 11 16@2x


bosiraphael (approved):
LGTM

PR #17056, opened 2026-01-09
[Extensibility] Support relation fields

### LINKED ISSUE #2043: [Extensibility] Support relation fields
Today, we only handle regular fields and not relation fields when synchronizing an object defined in an application.
We should be able to manage relation fields in applications.

ijreilly:
Closes https://github.com/twentyhq/core-team-issues/issues/2043

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/application/application-sync.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/application/application-sync.service.ts
@@ -357,6 +357,12 @@ export class ApplicationSyncService {
     }
   }
 
+  private isRelationFieldManifest(
+    field: FieldManifest | RelationFieldManifest,
+  ): field is RelationFieldManifest {
+    return field.type === FieldMetadataType.RELATION;
+  }
+
   private async syncFields({
     objectId,
     fieldsToSync,
@@ -366,9 +372,36 @@ export class ApplicationSyncService {
     objectId: string;
     workspaceId: string;
     applicationId: string;
-    fieldsToSync?: FieldManifest[];
+    fieldsToSync?: (FieldManifest | RelationFieldManifest)[];
   }) {
     if (!isDefined(fieldsToSync)) {
+      return;
+    }
+
+    const regularFields = fieldsToSync.filter(
+      (field): field is FieldManifest => !this.isRelationFieldManifest(field),
+    );
+
+    await this.syncRegularFields({
+      objectId,
+      fieldsToSync: regularFields,
+      workspaceId,
+      applicationId,
+    });
+  }
+
+  private async syncRegularFields({
+    objectId,
+    fieldsToSync,
+    workspaceId,
+    applicationId,
+  }: {
+    objectId: string;
+    workspaceId: string;
+    applicationId: string;
+    fieldsToSync: FieldManifest[];
+  }) {
+    if (fieldsToSync.length === 0) {
       return;
     }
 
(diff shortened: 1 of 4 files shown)


Weiko (commented):
I think this will be outdated very quickly. 
The current implementation is calling different services (field-metadata, object-metadata, etc...) many times when syncing the whole manifest.json (which contains, fields, objects, relations, etc). This will create many isolated migrations (one for each type of action.) and if one fail we will have a workspace with a broken state where some parts have been synced but not all. 
The vision is to have syncManifest that transpiles the whole manifest into an object that will be sent as a unique call to workspaceMigrationValidateBuildAndRunService (which is called today in each independent service) and execute a unique migration (isolated in its own transaction). 
This will be easier to validate the manifest as well (also if you introduce a relation AND an object at the same time (so the object does not exist yet in the cache), I'm not sure how this will behave currently, or you probably have to reimplement a lot of things to make it work)

packages/twenty-shared/src/application/objectManifestType.ts
```diff
@@ -9,4 +10,5 @@ export type ObjectManifest = SyncableEntityOptions & {
   description?: string;
   icon?: string;
   fields: FieldManifest[];
+  relations?: RelationFieldManifest[];
```
martmull: I think I would prefer 
```suggestion
fields: (FieldManifest | RelationFieldManifest)[] |;
```

martmull:
> I think this will be outdated very quickly. The current implementation is calling different services (field-metadata, object-metadata, etc...) many times when syncing the whole manifest.json (which contains, fields, objects, relations, etc). This will create many isolated migrations (one for each type of action.) and if one fail we will have a workspace with a broken state where some parts have been synced but not all. The vision is to have syncManifest that transpiles the whole manifest into an object that will be sent as a unique call to workspaceMigrationValidateBuildAndRunService (which is called today in each independent service) and execute a unique migration (isolated in its own transaction)

will be handled but htis blocked ticket -> https://github.com/twentyhq/core-team-issues/issues/1917

Weiko:
> > I think this will be outdated very quickly. The current implementation is calling different services (field-metadata, object-metadata, etc...) many times when syncing the whole manifest.json (which contains, fields, objects, relations, etc). This will create many isolated migrations (one for each type of action.) and if one fail we will have a workspace with a broken state where some parts have been synced but not all. The vision is to have syncManifest that transpiles the whole manifest into an object that will be sent as a unique call to workspaceMigrationValidateBuildAndRunService (which is called today in each independent service) and execute a unique migration (isolated in its own transaction)
> 
> will be handled but htis blocked ticket -> [twentyhq/core-team-issues#1917](https://github.com/twentyhq/core-team-issues/issues/1917)

Yes I wanted to give more context to @ijreilly 👍

packages/twenty-server/src/engine/core-modules/application/application-sync.service.ts
```diff
@@ -366,12 +372,39 @@ export class ApplicationSyncService {
...
   }) {
     if (!isDefined(fieldsToSync)) {
       return;
     }
 
+    const regularFields = fieldsToSync.filter(
+      (field): field is FieldManifest => !this.isRelationFieldManifest(field),
+    );
+
+    await this.syncRegularFields({
```
ijreilly, line 385: I have to handle relation fields after objects are created, otherwise it breaks because of independency between objects (through targetObjectUniversalIdentifier)

martmull (approved):
Nice thank you

PR #17063, opened 2026-01-09
Enable editing for calendar event custom fields

salmonumbrella:
## Summary
- Enable editing for custom calendar event fields while keeping standard fields read-only
- Load calendar event fields dynamically so custom field values are editable everywhere they appear
- Preserve calendar event participants rendering

## Testing
- npx nx run twenty-front:lint:diff-with-main --skip-nx-cache --output-style=stream
- npx jest --config packages/twenty-front/jest.config.mjs --runTestsByPath packages/twenty-front/src/modules/activities/calendar/hooks/__tests__/useCalendarEvents.test.tsx

## Notes
- Full nx lint/test are very slow locally after barrel generation; CI should run the full suite


salmonumbrella:
Updated the calendar event update hook to track the async update and surface the loading state instead of hardcoding false.

salmonumbrella:
@cubic-dev-ai moved the loading state to the component level and now return it via a memoized update hook, so no hooks are called inside the nested function.

charlesBochet:
@salmonumbrella Thanks for opening a PR, much appreciated

However, I'm not sure what we are trying to do here. Is there any issue describing the change? 
First look, I would say that we don't want users to modify calendarEvents

charlesBochet:
I'm going to close the PR for now but let's keep discussing here and you can also re-open it by commenting /twenty pr open

salmonumbrella:
@charlesBochet Thanks for the quick review!

To clarify the use case: I created a custom field (via the API) that links to a Notion meeting page for a calendar event. This custom field lives entirely within Twenty and is separate from the synced Google Calendar data.

The issue is that I'm unable to edit the URL for this custom field on the frontend. Since custom fields don't touch the actual Google Calendar variables, there's no risk of data corruption or sync conflicts—it's purely Twenty-side metadata.

This PR enables editing only for custom fields while keeping the standard calendar fields (title, dates, participants, etc.) read-only as intended.

Example of frontend UI:
https://cln.sh/647ctNkwnqjxPHSsTPKM

packages/twenty-front/src/modules/activities/calendar/components/CalendarEventDetails.tsx
```diff
@@ -113,11 +122,51 @@ export const CalendarEventDetails = ({
...
-    recordId: calendarEvent.id,
-    objectMetadataId: objectMetadataItem.id,
+  const { updateOneRecord } = useUpdateOneRecord({
+    objectNameSingular: CoreObjectNameSingular.CalendarEvent,
   });
 
+  const [isUpdating, setIsUpdating] = useState(false);
+  const updateEntity = useCallback(
+    ({ variables }: RecordUpdateHookParams) => {
+      if (!updateOneRecord) return;
```
charlesBochet: why this check? updateOneRecord cannot be undefined

packages/twenty-front/src/modules/activities/calendar/components/CalendarEventDetails.tsx
```diff
@@ -98,6 +106,7 @@ export const CalendarEventDetails = ({
     'location',
     'description',
   ];
+  const standardFieldNames = new Set(standardFieldOrder);
```
charlesBochet: this optimization does not looks useful, let's just use on standardFieldOrder

packages/twenty-front/src/modules/activities/calendar/components/CalendarEventDetails.tsx
```diff
@@ -113,11 +122,51 @@ export const CalendarEventDetails = ({
...
+    objectMetadataItem.id,
+  );
+  const isRecordDeleted = useIsRecordDeleted({ recordId: calendarEvent.id });
+
+  const isFieldReadOnly = (fieldMetadataItem: FieldMetadataItem) => {
+    if (standardFieldNames.has(fieldMetadataItem.name)) {
+      return true;
+    }
+
+    const fieldReadOnlyByPermissions = isFieldMetadataReadOnlyByPermissions({
```
charlesBochet: i'm surprised we don't already have a hook to compute permissions based on objectPermissions, isRecordDeleted

Let's check how other implementations are done (for workflow for instance) and if we really need to do that. I think these fields are already isUIReadOnly which should be enough, no need to check standardFieldNames (let's double check that)

packages/twenty-front/src/modules/command-menu/pages/calendar-event/components/CommandMenuCalendarEventPage.tsx
```diff
@@ -1,23 +1,48 @@
...
   const viewableRecordId = useRecoilComponentValue(
     viewableRecordIdComponentState,
   );
 
+  const { recordGqlFields } = useGenerateDepthRecordGqlFieldsFromObject({
+    objectNameSingular: CoreObjectNameSingular.CalendarEvent,
+    depth: 1,
+  });
+
+  const calendarEventRecordGqlFields = useMemo(
```
charlesBochet: memoization is not useuful here

packages/twenty-front/src/modules/command-menu/pages/calendar-event/components/CommandMenuCalendarEventPage.tsx
```diff
@@ -1,23 +1,48 @@
...
+    }),
+    [recordGqlFields],
+  );
+
   const { record: calendarEvent } = useFindOneRecord<CalendarEvent>({
-    objectNameSingular:
-      FIND_ONE_CALENDAR_EVENT_OPERATION_SIGNATURE.objectNameSingular,
+    objectNameSingular: CoreObjectNameSingular.CalendarEvent,
     objectRecordId: viewableRecordId ?? '',
-    recordGqlFields: FIND_ONE_CALENDAR_EVENT_OPERATION_SIGNATURE.fields,
```
charlesBochet, line 20: is FIND_ONE_CALENDAR_EVENT_OPERATION_SIGNATURE use elsewhere? If not we should remove it, if yes, there is likely a regression as signatures are here to make sure queries are consistent while making cache udpates

salmonumbrella:
Thanks @charlesBochet! Addressed the notes: removed the updateOneRecord guard + standardFieldNames Set, switched to the shared read-only helper, dropped the unnecessary memoization, and removed the unused operation signature. Pushed in 346c2d5.

packages/twenty-front/src/modules/activities/calendar/components/CalendarEventDetails.tsx
```diff
@@ -113,11 +121,46 @@ export const CalendarEventDetails = ({
...
+      setIsUpdating(true);
+      void updateOneRecord({
+        idToUpdate: variables.where.id as string,
+        updateOneRecordInput: variables.updateOneRecordInput,
+      }).finally(() => setIsUpdating(false));
+    },
+    [updateOneRecord],
+  );
+
+  const useUpdateOneCalendarEventRecordMutation: RecordUpdateHook = useCallback(
```
charlesBochet: do we still need this?

packages/twenty-front/src/modules/activities/calendar/components/CalendarEventDetails.tsx
```diff
@@ -113,11 +121,46 @@ export const CalendarEventDetails = ({
...
+
+  const objectPermissions = useObjectPermissionsForObject(
+    objectMetadataItem.id,
+  );
   const isRecordReadOnly = useIsRecordReadOnly({
     recordId: calendarEvent.id,
     objectMetadataId: objectMetadataItem.id,
   });
 
+  const isFieldReadOnly = (fieldMetadataItem: FieldMetadataItem) => {
```
charlesBochet: do we still need this method?

charlesBochet (commented):
Left a last round of comment, please re-test and then we are good to go, ty!

charlesBochet:
Please also fix the lint :)

salmonumbrella:
@charlesBochet fixed lint by making the boolean predicate explicit in the workflow story (commit 6c6b9d5). Ran  successfully (only warnings about ignored generated files).

salmonumbrella:
@charlesBochet (follow-up) lint now clean on diff: ran nx lint:diff-with-main successfully; only warnings are for ignored generated files.

charlesBochet:
@salmonumbrella I've left two comments (still not resolved) above too

salmonumbrella:
@charlesBochet addressed the two latest comments in CalendarEventDetails: removed the memoized wrapper for useUpdateRecord and inlined the read-only helper (commit 7153a3d).

charlesBochet (approved):

charlesBochet:
Thank you!

charlesBochet:
@salmonumbrella Just tried on our pre-prod environment, and I can't edit a textField on calendarEvent, could you check?
image

salmonumbrella:
I created a custom field called "Notion" on the CalendarEvent object via the API. The field appears in the Calendar Event details panel, but it's not editable - it shows as read-only like the standard fields.

**Expected:** Custom fields (where `isCustom: true` and `isUIReadOnly: false`) should be editable.
**Actual:** The custom field appears but cannot be edited.

![Custom field not editable [screenshot]](https://cln.sh/F0dWyjwDHWvVnRWQHpd8)

**Possible causes I investigated:**
- The `isRecordFieldReadOnly` logic looks correct - it checks `fieldMetadataItem.isUIReadOnly ?? false`
- The `customFields` filter correctly identifies `field.isCustom` fields
- Unclear if the field was created with `isUIReadOnly: true` by default, or if there's an object-level permission issue

@charlesBochet Could you help investigate?

PR #17065, opened 2026-01-10
IMAP edge cases

neo773:
Fixes IMAP edge case. We had a maximum call stack reached crash if the array was very large for some folders. Plus some minor perf improvement.

https://twenty-v7.sentry.io/issues/6997180505/events/652c69da1be944aab1c87ac2e6a06c30/

### DIFF
--- a/packages/twenty-server/src/modules/messaging/message-import-manager/drivers/imap/services/imap-find-sent-folder.service.ts
+++ b/packages/twenty-server/src/modules/messaging/message-import-manager/drivers/imap/services/imap-find-sent-folder.service.ts
@@ -140,23 +140,15 @@ export class ImapFindSentFolderService {
     folderPath: string,
   ): Promise<number> {
     try {
-      const lock = await client.getMailboxLock(folderPath);
+      const status = await client.status(folderPath, {
+        messages: true,
+      });
 
-      try {
-        const status = await client.status(folderPath, {
-          messages: true,
-        });
+      const messageCount = status?.messages;
 
-        const messageCount = status?.messages;
+      this.logger.debug(`Folder "${folderPath}" has ${messageCount} messages`);
 
-        this.logger.debug(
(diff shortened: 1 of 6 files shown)


packages/twenty-server/src/modules/messaging/message-import-manager/drivers/imap/services/imap-find-sent-folder.service.ts
```diff
@@ -140,23 +140,15 @@ export class ImapFindSentFolderService {
     folderPath: string,
   ): Promise<number> {
     try {
-      const lock = await client.getMailboxLock(folderPath);
```
charlesBochet, line 143: @neo773 could you explain when you do this kind of change? why don't we need the lock anymore? why was it introduced in the first place.

This helps for the review

packages/twenty-server/src/modules/messaging/message-import-manager/drivers/imap/utils/create-sync-cursor.util.ts
```diff
@@ -2,17 +2,20 @@ import { type MailboxState } from './extract-mailbox-state.util';
...
   const { uidValidity, highestModSeq } = mailboxState;
   const lastSeenUid = previousCursor?.highestUid ?? 0;
 
-  const highestUid =
-    messages.length > 0
-      ? Math.max(...messages.map((message) => message.uid))
-      : lastSeenUid;
+  let highestUid = lastSeenUid;
+
+  for (let i = 0; i < messageUids.length; i++) {
```
charlesBochet, line 14: ok, we usually avoid the for loop in the code base for reduce API in case we want to extract one aggregate (here the max). This avoid mutating variable (here let highestUid)

charlesBochet (commented):
Left comments :)

charlesBochet (approved):

PR #17066, opened 2026-01-10
fix: add main entity columns to SELECT for DISTINCT compatibility with mixed ordering

FelixMalfait:
## Problem
When combining relation field ordering (e.g., `{ company: { name: 'DescNullsLast' } }`) with regular field ordering (e.g., `{ position: 'AscNullsFirst' }`), TypeORM creates a DISTINCT subquery that expects ALL orderBy columns to be selected with underscore aliases.

We were only adding relation columns (like `company_name`), not main entity columns (like `person_position`), causing the error:
```
column distinctAlias.person_position does not exist
```

## Solution
Now we also add main entity columns to the SELECT when there are relation orderBy fields present (which is the condition that triggers DISTINCT behavior).

## Test
Added integration test for mixed ordering (relation + regular field) scenario.

### DIFF
--- a/packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query.parser.ts
+++ b/packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query.parser.ts
@@ -143,35 +143,19 @@ export class GraphqlQueryParser {
     parsedOrderBy: Record<string, OrderByClause>,
     objectNameSingular: string,
   ): void {
-    // Add ORDER BY columns with underscore alias for DISTINCT compatibility
+    // Add relation ORDER BY columns with underscore alias for DISTINCT compatibility
     // This must be called AFTER setFindOptions because setFindOptions clears addSelect
-    const orderByKeys = Object.keys(parsedOrderBy);
-
-    // Check if there are any relation columns (different alias than main entity)
-    const hasRelationOrderBy = orderByKeys.some((key) => {
-      const parts = key.split('.');
-
-      return parts.length === 2 && parts[0] !== objectNameSingular;
-    });
-
-    for (const orderByKey of orderByKeys) {
+    for (const orderByKey of Object.keys(parsedOrderBy)) {
       const parts = orderByKey.split('.');
 
       if (parts.length === 2) {
         const [alias, column] = parts;
 
-        // Add relation columns always, and main entity columns only when
-        // there are relation columns (to support DISTINCT subquery compatibility)
+        // Only add select for joined relation columns, not main entity columns
         if (alias !== objectNameSingular) {
           queryBuilder.addSelect(
             `"${alias}"."${column}"`,
             `${alias}_${column}`,
-          );
-        } else if (hasRelationOrderBy) {
-          // Main entity columns need underscore alias when DISTINCT is used
-          queryBuilder.addSelect(
-            `"${objectNameSingular}"."${column}"`,
-            `${objectNameSingular}_${column}`,
           );
         }
       }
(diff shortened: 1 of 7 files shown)


packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query.parser.ts
charlesBochet, line 139: could we clean that?

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query.parser.ts
```diff
@@ -143,20 +143,36 @@ export class GraphqlQueryParser {
...
   ): void {
-    // Add relation ORDER BY columns with underscore alias for DISTINCT compatibility
+    // Add ORDER BY columns with underscore alias for DISTINCT compatibility
     // This must be called AFTER setFindOptions because setFindOptions clears addSelect
-    for (const orderByKey of Object.keys(parsedOrderBy)) {
+    const orderByKeys = Object.keys(parsedOrderBy);
+
+    // Check if there are any relation columns (different alias than main entity)
+    const hasRelationOrderBy = orderByKeys.some((key) => {
+      const parts = key.split('.');
```
charlesBochet: parts naming is not clear to me, what is a part?

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query.parser.ts
```diff
@@ -143,20 +143,36 @@ export class GraphqlQueryParser {
...
         const [alias, column] = parts;
 
-        // Only add select for joined relation columns, not main entity columns
         if (alias !== objectNameSingular) {
+          // Always add relation columns
           queryBuilder.addSelect(
             `"${alias}"."${column}"`,
             `${alias}_${column}`,
           );
+        } else if (hasRelationOrderBy && column !== 'id') {
```
charlesBochet: this code is not clear to me, why are we checking hasRelationOrderBy globally within all orderByKeys and then re-looping on them?

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query.parser.ts
charlesBochet: something more clear would be to use a regex to extract {objectNameSingular}.{column} for instance

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query.parser.ts
```diff
@@ -143,20 +143,36 @@ export class GraphqlQueryParser {
...
+
+    for (const orderByKey of orderByKeys) {
       const parts = orderByKey.split('.');
 
       if (parts.length === 2) {
         const [alias, column] = parts;
 
-        // Only add select for joined relation columns, not main entity columns
         if (alias !== objectNameSingular) {
+          // Always add relation columns
```
charlesBochet: i don't understand this comment

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query.parser.ts
charlesBochet: Example of something that would be more readable:
1. orderByKeys => we transform to an array of {objectNameSingular, column} 
2.  const columnsInObject =  filter alias === objectNameSingular, then we add the columns to queryBuilder from this list
3. const columnsInRelatedObject = filter alias !== objectNameSingular, then we add the columns to queryBuilder from this list

packages/twenty-server/test/integration/graphql/suites/order-by-relation-field.integration-spec.ts
```diff
@@ -453,6 +453,46 @@ describe('Order by relation field (e2e)', () => {
     expect(edges.length).toBeGreaterThan(0);
   });
 
+  it('should support mixed ordering (relation field + regular field)', async () => {
```
charlesBochet, line 456: thanks!

PR #17077, opened 2026-01-11
feat: upgrade Storybook to version 9

mabdullahabaid:
Upgraded from 8.6.15 to 9.1.17 in two steps: 
- 8.6.15 -> 9.0.0 
- 9.0.0  -> 9.1.17

I had to disable `storybook-addon-cookie` since it is not supported for Storybook 9. However, I do intend to upgrade to Storybook 10 when this is merged, so we can replace the aforementioned add-on with this fork specifically created to support Storybook 10 and above: https://www.npmjs.com/package/@storybook-community/storybook-addon-cookie.

Additionally, once we upgrade to Version 10 successfully, I will start looking into integrating the official Vitest add-on.


FelixMalfait (approved):
Great!!!

PR #17078, opened 2026-01-11
fix: removed scrollbar, elipsified overflow text supporting tooltip

### LINKED ISSUE #15152: Visible scrolls while adding new field to an object

carbonFibreCode:
fixes #15152 

Added CSS to ellipsify overflowing text with tooltip support, consistent with existing dropdown implementations such as the address dropdown for long country names.
This approach resolves the issue locally without modifying global components, avoiding the limitations of previous attempts.

Screenshot 2026-01-11 at 4 09 45 PM

Before :

Screenshot 2026-01-11 at 4 09 11 PM

After :

Screenshot 2026-01-11 at 4 08 28 PM

### DIFF
--- a/packages/twenty-front/src/modules/settings/components/SettingsMorphRelationMultiSelect.tsx
+++ b/packages/twenty-front/src/modules/settings/components/SettingsMorphRelationMultiSelect.tsx
@@ -68,20 +68,6 @@ const StyledDescription = styled.span`
   font-size: ${({ theme }) => theme.font.size.sm};
 `;
 
-const StyledMenuItemWrapper = styled.div`
-  width: 100%;
-  display: flex;
-  align-items: center;
-  gap: ${({ theme }) => theme.spacing(2)};
-  min-width: 0;
-
-  > div > div {
-    overflow: hidden;
-    text-overflow: ellipsis;
-    white-space: nowrap;
-  }
-`;
-
 export const SettingsMorphRelationMultiSelect = ({
   className,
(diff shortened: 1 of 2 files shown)


packages/twenty-front/src/modules/settings/components/SettingsMorphRelationMultiSelect.tsx
```diff
@@ -68,6 +68,20 @@ const StyledDescription = styled.span`
...
 `;
 
+const StyledMenuItemWrapper = styled.div`
+  width: 100%;
+  display: flex;
+  align-items: center;
+  gap: ${({ theme }) => theme.spacing(2)};
+  min-width: 0;
+
+  > div > div {
```
charlesBochet: let's avoid doing it. Is this the way it's done in other dropdowns? 

I feel we should be using Existing TextWithOverflowingTooltip component. If not possible, we should add this style to the container itself and not using > div > div that will break when we change the DOM structure

charlesBochet (approved):

charlesBochet:
I have made the fix, ty!

carbonFibreCode:
> I have made the fix, ty!

hey, thanks for the fix, i was trying to push changes rn but it's already merged... lol

PR #17079, opened 2026-01-11
Fix orderBy columns missing from SELECT in DISTINCT subquery

FelixMalfait:
## Description

Fixes a bug where queries with relation field + scalar field ordering would fail with:
```
column distinctAlias.person_position does not exist
```

## Root Cause

When a GraphQL query orders by columns that are **not in the selected fields**, TypeORM's DISTINCT subquery fails because it expects those columns in the inner SELECT with alias format (e.g., `person_position`).

The issue only manifests when:
1. A filter is applied (triggers DISTINCT path in TypeORM)
2. OrderBy includes columns NOT in the GraphQL selection
3. Example: query selects only `id`, but orders by `position`

## The Fix

`addRelationOrderColumnsToBuilder` now accepts `columnsToSelect` and adds orderBy columns via `addSelect()` only if they're **NOT** already in the selected columns:

- **Relation orderBy columns**: Always added (never in columnsToSelect)
- **Main entity orderBy columns**: Added only when not already selected

This ensures all orderBy columns are present in TypeORM's inner SELECT for the DISTINCT subquery.

## Test Plan

Added integration test for the exact failing scenario:
- Filter with `neq` (triggers DISTINCT path)
- Multiple orderBy: relation field + scalar field  
- Minimal field selection (only `id`, not `position`)

## Related

Regression introduced in #17021 which added nested sort support.

### DIFF
--- a/packages/twenty-server/src/engine/api/common/common-query-runners/common-find-many-query-runner.service.ts
+++ b/packages/twenty-server/src/engine/api/common/common-query-runners/common-find-many-query-runner.service.ts
@@ -31,10 +31,6 @@ import {
 } from 'src/engine/api/common/types/common-query-args.type';
 import { CommonSelectedFieldsResult } from 'src/engine/api/common/types/common-selected-fields-result.type';
 import { getPageInfo } from 'src/engine/api/common/utils/get-page-info.util';
-import {
-  GraphqlQueryRunnerException,
-  GraphqlQueryRunnerExceptionCode,
-} from 'src/engine/api/graphql/graphql-query-runner/errors/graphql-query-runner.exception';
 import { ProcessAggregateHelper } from 'src/engine/api/graphql/graphql-query-runner/helpers/process-aggregate.helper';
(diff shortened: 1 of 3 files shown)


packages/twenty-server/src/engine/api/common/common-query-runners/common-find-many-query-runner.service.ts
```diff
@@ -112,11 +108,12 @@ export class CommonFindManyQueryRunnerService extends CommonBaseQueryRunnerServi
           fieldIdByName,
         )
       ) {
-        throw new GraphqlQueryRunnerException(
+        // Not throwing exception because still used on record show page
+        /* throw new GraphqlQueryRunnerException(
           'Cursor-based pagination is not supported with relation field ordering. Use offset pagination instead.',
```
FelixMalfait, line 113: Temporary, will need to be brought back

FelixMalfait:
Merging now as this is fixing a critical bug on main. We still need to see what we want to do with cursor-based pagination on record pages

PR #17080, opened 2026-01-11
fix: hide "Add new" button for workspace member relations

AyeshaWaseemAhmad:
**Description**

Fixes the "Add new" button appearing in relation picker dropdowns for workspace member relations (e.g., Account Owner on companies). Workspace members cannot be created from relation pickers, so this button should not appear.

**Changes Made**

- Modified RelationManyToOneFieldInput.tsx to conditionally pass the onCreate prop to SingleRecordPicker only when createNewRecordAndOpenRightDrawer is defined.
- Moved the conditional check to the JSX level for clarity.

**Technical Details**

The useAddNewRecordAndOpenRightDrawer hook already returns undefined for workspace member relations, but the component was still passing a handler function to SingleRecordPicker, causing the "Add new" button to appear. The fix ensures that when createNewRecordAndOpenRightDrawer is undefined, onCreate is also undefined, which prevents SingleRecordPickerMenuItemsWithSearch from rendering the button.

Testing

- Verified that the "Add new" button no longer appears in the Account Owner dropdown for companies.
- Confirmed that the "Add new" button still appears for other relation fields where it's appropriate (e.g., People, Opportunities).
- Tested that existing functionality for selecting workspace members from the dropdown still works correctly.

Fixes #17060

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/input/components/RelationManyToOneFieldInput.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/input/components/RelationManyToOneFieldInput.tsx
@@ -110,7 +110,7 @@ export const RelationManyToOneFieldInput = () => {
       EmptyIcon={IconForbid}
       emptyLabel={t`No ${fieldLabel}`}
       onCancel={onCancel}
-      onCreate={handleCreateNew}
+      onCreate={createNewRecordAndOpenRightDrawer ? handleCreateNew : undefined}
       onMorphItemSelected={handleMorphItemSelected}
       objectNameSingulars={[
         fieldDefinition.metadata.relationObjectMetadataNameSingular,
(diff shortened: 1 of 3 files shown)


packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/input/components/RelationManyToOneFieldInput.tsx
```diff
@@ -110,7 +110,7 @@ export const RelationManyToOneFieldInput = () => {
       EmptyIcon={IconForbid}
       emptyLabel={t`No ${fieldLabel}`}
       onCancel={onCancel}
-      onCreate={handleCreateNew}
+      onCreate={createNewRecordAndOpenRightDrawer ? handleCreateNew : undefined}
```
ehconitin: use isDefined

ehconitin (commented):
This file has the same issue -- `packages/twenty-front/src/modules/object-record/record-field-list/record-detail-section/relation/components/RecordDetailRelationSectionDropdownToMany.tsx` also passes onCreate={handleCreateNew} without checking if `createNewRecordAndOpenRightDrawer` is defined. Could you fix that here too?

thanks

packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/input/components/RelationManyToOneFieldInput.tsx
AyeshaWaseemAhmad: Addressed.

ehconitin (approved):
lgtm! thankyou

PR #17098, opened 2026-01-12
[Dashboards] add primary axis select gap fill for bar and line charts

### LINKED ISSUE #2056: [DASHBOARDS] Always display all the select options even if the value is null
When we group by a select value on the first axis, we should display all select options, even if the value is null.

We should only hide those groups if the setting "Hide empty values" is set to `true`.

ehconitin:
closes https://github.com/twentyhq/core-team-issues/issues/2056

https://github.com/user-attachments/assets/44eb2704-9ea9-4f2b-8941-6dd5c12d6276


packages/twenty-front/src/modules/page-layout/widgets/graph/types/FillSelectGapsResult.ts
```diff
@@ -0,0 +1,5 @@
+import { type GroupByRawResult } from '@/page-layout/widgets/graph/types/GroupByRawResult';
+
+export type FillSelectGapsResult = {
+  data: GroupByRawResult[];
```
charlesBochet: why is it useful to have result.data?

charlesBochet (approved):
LGTM regarding the code, I've left one nitpick comment, I'm testing locally now

packages/twenty-front/src/modules/page-layout/widgets/graph/types/FillSelectGapsResult.ts
ehconitin: thanks for the catch -- its no longer required
I tried to abstract date gap fill and select gap fill into one (date gap fill requires wasTruncated, which is in result.data)

But later decided on not to overabstract and keep things separate 

cleaning it up

charlesBochet:
Tested the behavior, LGTM, merge when you are ready :)

packages/twenty-front/src/modules/page-layout/widgets/graph/utils/fillSelectGapsInTwoDimensionalChartData.ts
```diff
@@ -0,0 +1,52 @@
...
+  for (const item of data) {
+    const primaryDimensionValue = item.groupByDimensionValues?.[0];
+
+    if (!isDefined(primaryDimensionValue)) {
+      continue;
+    }
+
+    const secondDimensionValue = (item.groupByDimensionValues?.[1] ?? null) as
+      | string
+      | null;
```
bosiraphael: It is not necessarily a string, you can extract the type `DimensionValue` in a type file. It is defined in `packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetBarChart/utils/createEmptyDateGroup.ts`

PR #17099, opened 2026-01-12
fix(kanban): optimistic drag & drop update to prevent jitter

nr0225:
Delayed endRecordDrag() execution to ensure smooth UI transition during optimistic updates.


charlesBochet:
@nr0225 What problem are we solving here, could you post a before / after as a comment?

Also they are Chinese comment in your PR

I'm going to close this PR too as I don't understand it but we can also keep discussing here (same as the other one) and will re-open if it makes sense

PR #17104, opened 2026-01-12
Fix settings data model tables to take full width and align columns

Bonapara:
## Summary
- Updated Relations and Fields tables in the object detail settings page to use flexible width (`1fr`) for the Name column instead of fixed pixels
- Aligned column widths between both tables (`1fr 148px 148px 36px`) so the App and Type/Data type columns line up

## Test plan
- [ ] Navigate to Settings > Data model > select any object (e.g., Companies)
- [ ] Verify the Relations table rows take the full available width
- [ ] Verify the Fields table rows take the full available width
- [ ] Verify the App and Type/Data type columns are aligned between both tables


Bonapara:
# After 

CleanShot 2026-01-12 at 18 47 10@2x


# Before

CleanShot 2026-01-12 at 18 47 26@2x

charlesBochet (approved):

PR #17113, opened 2026-01-13
fix rich text widget's side menu handle

ehconitin:
before - 

https://github.com/user-attachments/assets/283a1ed6-19e9-4a4d-ba4f-5bac7d4a23ef

after - 

https://github.com/user-attachments/assets/778a45af-fb59-4e21-aa53-290461c4d159


bosiraphael (approved):
LGTM

PR #17120, opened 2026-01-13
Allow `isUnique` mutation on standard field

prastoin:
# Introduction
As we've just [identified](https://github.com/twentyhq/twenty/pull/16981) the standard field metadata in production we can not enable the is unique comparison even for standard entities


etiennejouan (approved):
Looks gooooood to me

PR #17121, opened 2026-01-13
Move query matching before publication

thomtrp:
- new channel EVENT_STREAM_CHANNEL based on event stream id
- on event, perform the matching and publish only to the right streams
- store a list of active streams per workspace
- store the user id along with the queries for each stream

Bonus:
- remove onSubscriptionMatch


charlesBochet:
Leaving it to you @thomtrp as tests are red :)

packages/twenty-server/src/engine/subscriptions/event-stream.service.ts
```diff
@@ -1,96 +1,176 @@
...
+    userWorkspaceId,
+  }: {
+    workspaceId: string;
+    eventStreamChannelId: string;
+    userId: string;
+    userWorkspaceId: string;
+  }): Promise<void> {
+    const key = this.getEventStreamKey(workspaceId, eventStreamChannelId);
+    const streamData: EventStreamData = {
+      userId,
```
Weiko: What about api keys / agents mode (that are not running behind a user). Should we use an authContext here instead?

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.service.ts
```diff
@@ -42,10 +44,78 @@ export class WorkspaceEventEmitterService {
...
-    });
+    await this.publishToEventStreams(workspaceEventBatch);
+  }
+
+  private async publishToEventStreams(
+    workspaceEventBatch: WorkspaceEventBatch<ObjectRecordEvent>,
+  ): Promise<void> {
+    const workspaceId = workspaceEventBatch.workspaceId;
+
+    const activeStreamIds =
```
Weiko, line 55: This might not get cleared and have stale data due to missing TTL

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.service.ts
```diff
@@ -42,10 +44,78 @@ export class WorkspaceEventEmitterService {
...
+      await this.eventStreamService.getActiveStreamIds(workspaceId);
+
+    if (activeStreamIds.length === 0) {
+      return;
+    }
+
+    const streamsData = await this.eventStreamService.getStreamsData(
```
Weiko, line 62: This however would be cleared, meaning you might iterate below hover dead steam data with your
```typescript
if (!isDefined(streamData)) {
        continue;
      }
```
for no reason and this would add up.

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.service.ts
```diff
@@ -42,10 +44,78 @@ export class WorkspaceEventEmitterService {
...
+      workspaceId,
+      activeStreamIds,
+    );
+
+    for (const [streamChannelId, streamData] of streamsData) {
+      if (!isDefined(streamData)) {
+        continue;
```
Weiko, line 72: instead of continue, you could also add some "pruning" logic here for the issue I've mentioned above

Weiko (approved):
This is much cleaner imho!

Note: activeStreams does not have a TTL, if destroy is never called (connection drop, tab crash, etc) this would stay in the cache forever, see my comments about this issue.

Otherwise LGTM! 👏 @thomtrp

PR #17123, opened 2026-01-13
Hardcode workflow objects and dashboards to be last in the navigation drawer

### LINKED ISSUE #2058: Fix workspace favorites order
Workspace favorites are ordered by created at.
We should order them by position.

Standard objects should always be first.
Then custom objects.
Then workflows and dashboards.

Create a command to update the position of the workspace favorites to make them match the current sort.

bosiraphael:
Closes https://github.com/twentyhq/core-team-issues/issues/2058

We settled on a simpler solution by just hardcoding the objects order for now. We will change this once it's possible to reorder the workspace favorites with drag and drop.

### DIFF
--- a/packages/twenty-front/src/modules/object-metadata/components/NavigationDrawerSectionForObjectMetadataItems.tsx
+++ b/packages/twenty-front/src/modules/object-metadata/components/NavigationDrawerSectionForObjectMetadataItems.tsx
@@ -20,8 +20,6 @@ const ORDERED_FIRST_STANDARD_OBJECTS: string[] = [
 
 const ORDERED_LAST_STANDARD_OBJECTS: string[] = [
   CoreObjectNameSingular.Workflow,
-  CoreObjectNameSingular.WorkflowVersion,
-  CoreObjectNameSingular.WorkflowRun,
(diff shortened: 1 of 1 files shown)


packages/twenty-front/src/modules/object-metadata/components/NavigationDrawerSectionForObjectMetadataItems.tsx
```diff
@@ -8,15 +8,23 @@ import { NavigationDrawerSection } from '@/ui/navigation/navigation-drawer/compo
...
   CoreObjectNameSingular.Company,
   CoreObjectNameSingular.Opportunity,
   CoreObjectNameSingular.Task,
   CoreObjectNameSingular.Note,
 ];
 
+const ORDERED_LAST_STANDARD_OBJECTS: string[] = [
+  CoreObjectNameSingular.Workflow,
+  CoreObjectNameSingular.WorkflowVersion,
+  CoreObjectNameSingular.WorkflowRun,
```
thomtrp: you don't need run and version here. These do not have workspace favorites

thomtrp (approved):

Weiko (approved):

PR #17126, opened 2026-01-13
Invalidate legacy cache after flats

prastoin:
# Introduction

Some legacy caches are based on the flat ones
So we need to seq invalidate before invalidating others as it could result in inter dep cache invalidation race condition


charlesBochet (approved):

charlesBochet:
Makes sense, LGTM

PR #17127, opened 2026-01-13
refactor google refresh token service error handling

neo773:
Refactors to be consistent with Microsoft service
Handles scenarios like temporary error which was not handled before
Moved `IsGmailNetworkError` from root orchestrator to driver level


charlesBochet:
@neo773 could you please explain why you make the change? what was the issue and how this is solving it

Generally always add PR description, unless the title is enough

neo773:
@charlesBochet 
Apologies, will come up with a system to ensure this doesn't happen again

charlesBochet (approved):
LGTM

PR #17129, opened 2026-01-13
[Apps] Small improvements

ijreilly:
- Fix undefined in breadcrumbs
breadcrumbs_bad
- Make role required in app definition


charlesBochet (approved):

packages/twenty-sdk/src/application/objects/define-object.ts
```diff
@@ -36,27 +36,55 @@ import { validateFieldsOrThrow } from './validate-fields';
...
+
+    if (!isDefined(field.type)) {
+      throw new Error(`Field "${field.label}" must have a type`);
+    }
+
+    if (!isDefined(field.name)) {
+      throw new Error('Field must have a name');
+    }
+
+    if (!isDefined(field.label)) {
```
charlesBochet: use isNonEmptyString?

PR #17131, opened 2026-01-13
fix: record does not open in side panel after returning from fullscreen

### LINKED ISSUE #17089: Can't reopen record after having navigated to its show page
## Bug Description

1. Click on a Company record on the record index
2. Choose to open the record in its fullscreen show page
3. Go back to the record index and click on the same record **-> it does nothing**

https://github.com/user-attachments/assets/cba473f7-cf6c-4900-b5fd-e31f17b70484

## Expected behavior

I should be able to open a given record many times.

abdulrahmancodes:
Closes #17089 

### 1. Can't reopen record after having navigated to its show page
After opening a record in the show page from the command menu and going back to the index, clicking the same record again did nothing. The command menu navigation stack was not cleared when opening in the show page, so the "already open" check skipped reopening. We now clear the command menu navigation stack before navigating to the show page (in `RecordShowRightDrawerOpenRecordButton`), so the same record can be reopened from the index.

### 2. Row doesn't highlight when opening command menu after return from show page
After returning from the record show page to the index, the first row click opened the command menu but the row did not highlight. The "side panel close" event was emitted not only when the panel actually closed, but also when opening the command menu (cleanup ran with `isCommandMenuClosing` and always emitted the event). Listeners like `RecordTableDeactivateRecordTableRowEffect` then deactivated the row. We now emit the side panel close event only when the close animation actually completes (`CommandMenuSidePanelForDesktop`), and skip emitting it when cleanup is run from the open path (`useNavigateCommandMenu` passes `emitSidePanelCloseEvent: false`). The table still deactivates the row when the user closes the panel, but no longer when they open the command menu by clicking a row.


Devessier:
Hey @bosiraphael,
I don’t have a computer right now. Could you please review this PR tomorrow? You have a good knowledge of this part of the codebase. Thank you!

packages/twenty-front/src/modules/object-record/record-table/components/RecordTableDeactivateRecordTableRowEffect.tsx
```diff
@@ -1,12 +1,26 @@
...
+          .getValue();
+
+        if (!isCommandMenuOpened) {
+          deactivateRecordTableRow();
+        }
+      },
+    [deactivateRecordTableRow],
+  );
+
+  useListenToSidePanelClosing(() => setTimeout(deactivateRowIfMenuClosed, 0));
```
bosiraphael: Not a fan of the setTimeout, I think it could be done in a cleaner way

ijreilly:
any update on this @abdulrahmancodes? does it need to be reviewed again (cc @bosiraphael) ?

abdulrahmancodes:
@ijreilly This slipped my mind while working on the navbar customization. I'll take care of the review comment today and update the PR shortly.

abdulrahmancodes:
Hey @bosiraphael! I've addressed the review comment. Please review if it's fine now. Thanks

packages/twenty-front/src/modules/object-record/record-table/components/RecordTableDeactivateRecordTableRowEffect.tsx
```diff
@@ -1,11 +1,27 @@
...
+        if (!isCommandMenuOpened) {
+          deactivateRecordTableRow();
+        }
+      },
+    [deactivateRecordTableRow],
+  );
+
   useListenToSidePanelClosing(() => {
-    deactivateRecordTableRow();
+    queueMicrotask(deactivateRowIfMenuClosed);
```
bosiraphael: This is more or less the same as doing a setTimeout 0, it would be better to try to refactor this to make it synchronous rather than using an effect

packages/twenty-front/src/modules/command-menu/hooks/useNavigateCommandMenu.ts
```diff
@@ -50,7 +50,9 @@ export const useNavigateCommandMenu = () => {
           .getValue();
 
         if (isCommandMenuClosing) {
-          commandMenuCloseAnimationCompleteCleanup();
+          commandMenuCloseAnimationCompleteCleanup({
+            emitSidePanelCloseEvent: false,
+          });
```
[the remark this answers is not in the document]
abdulrahmancodes: We only use emitSidePanelCloseEvent: false when we’re about to open the menu (we always set isCommandMenuOpened to true right after). So when the close transition finishes, the menu is already open and handleTransitionEnd correctly doesn’t run cleanup or emit. Not emitting is intended.

bosiraphael (approved):
LGTM

PR #16962, opened 2026-01-06
dashboard workspace standard seeds

ehconitin:
# Introduction

In this pull request we're introducing new standard page layout, tabs and widget ( 1 page layout, 1 tab and 8 widgets ) and also a new opportunity field

Also now prefilling new records, 6 opportunities and a dashboard.

### New workspace creation
Relies on existing standard declaration builder

### Backfill command
We've been hacking through the standard builder in order to extract only the standard page layout entities, updated their entity dependencies to match the workspace ids so the validation passes

## Remark
- Refactored the `PageLayoutWidget` configuration type to be dynamically typed through a generic discriminated union

PR #17141, opened 2026-01-14
Fix dev seed page layout cache invalidation

prastoin:
# Introduction

followup https://github.com/twentyhq/twenty/pull/16962

Root cause: as now the twenty standard app installation in seeded workspace invalidates and set the page layout-xxx caches the inserted through direct repo access data in the prefill legacy page layout methods do not interact with the cache
The cache being set in prior when recompute does not result in db introspection as before as when was unset


bosiraphael (approved):
LGTM

PR #17145, opened 2026-01-14
Filter valid fields in record steps

### LINKED ISSUE #16775: [Workflow] Field deletion can break workflow runs
## Bug Description

To reproduce:
- fill a field in a create record step
- delete the field in model
- try to run the workflow

thomtrp:
Fixes https://github.com/twentyhq/twenty/issues/16775

When a field is deleted from the model, the workflow action step still stores the deleted field name in `step.settings.input.objectRecord`. 

We need to filter out the fields that are not valid anymore. This logic existed before but had been removed with the migration to tool services.


etiennejouan (approved):
LGTM !

PR #17146, opened 2026-01-14
fix: e2e login test - handle optional Continue with Email button

FelixMalfait:
## Summary
The e2e login test was failing because it unconditionally tried to click 'Continue with Email' button, but this button doesn't exist when password is the only auth method.

## Root Cause
In `SignInUpWorkspaceScopeFormEffect.tsx`, when a workspace only has password authentication (no Google/Microsoft/SSO), the effect automatically calls `continueWithEmail()` which skips the Init step and shows the email field directly.

## Changes
1. **loginPage.ts**: Added `clickLoginWithEmailIfVisible()` method that only clicks the button if it exists
2. **login.setup.ts**: 
   - Replaced `clickLoginWithEmail()` with `clickLoginWithEmailIfVisible()` 
   - Updated regex from `/Welcome to .+/` to `/Welcome, .+/` to match the recent UI change

### DIFF
--- a/packages/twenty-e2e-testing/lib/pom/loginPage.ts
+++ b/packages/twenty-e2e-testing/lib/pom/loginPage.ts
@@ -85,6 +85,13 @@ export class LoginPage {
     await this.loginWithEmailButton.click();
   }
 
+  async clickLoginWithEmailIfVisible() {
+    const isVisible = await this.loginWithEmailButton.isVisible();
+    if (isVisible) {
+      await this.loginWithEmailButton.click();
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


charlesBochet (approved):

FelixMalfait:
@charlesBochet I'm not sure if it will work, didn't test locally - wanted to use this PR to iterate, let's see if it's rejected from merge queue

charlesBochet:
let me run the e2e tests locally, should take 2mins

charlesBochet:
image

All good!

PR #17150, opened 2026-01-14
[Dashboards] disable chart drill-down and pointer cursor for relation group-by fields

ehconitin:


packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetBarChart/components/GraphWidgetBarChartRenderer.tsx
```diff
@@ -79,6 +80,11 @@ export const GraphWidgetBarChartRenderer = () => {
     }),
   );
 
+  const primaryGroupByField = objectMetadataItem.fields.find(
+    (field) => field.id === configuration.primaryAxisGroupByFieldMetadataId,
+  );
+  const isDrillDownSupported = isChartDrillDownSupported(primaryGroupByField);
```
bosiraphael: Could we rename this? I would prefer something like `isFilteredViewRedirectionSupported`

bosiraphael (approved):
LGTM

PR #17152, opened 2026-01-14
Add feature flag for sse

thomtrp:
As title


charlesBochet (approved):

packages/twenty-front/src/modules/sse-db-event/components/SSEProviderEffect.tsx
```diff
@@ -2,23 +2,27 @@ import { SseClientContext } from '@/sse-db-event/contexts/SseClientContext';
...
 import { isNonEmptyString } from '@sniptt/guards';
 import { type ExecutionResult, print } from 'graphql';
 import { useContext, useEffect } from 'react';
 import { useRecoilState } from 'recoil';
 import { isDefined } from 'twenty-shared/utils';
 import { v4 } from 'uuid';
-import { type EventSubscription } from '~/generated/graphql';
+import { type EventSubscription, FeatureFlagKey } from '~/generated/graphql';
 
 export const SSEProviderEffect = () => {
```
martmull, line 13: if you update that you might update SSEQuerySubscribeEffect isn't it?
(I would change none of those 2 files personally, change in `packages/twenty-front/src/modules/sse-db-event/components/SSEProvider.tsx` is enough to me)

Issue #16978, opened 2026-01-07
Domain cannot be re-added to new Company if already used by deleted company


jeremymoll:
## Bug Description

When trying to recreate a deleted Company in Twenty, the domain field shows an error "This Domain Name value is already in use. Please check your data and try again." even though the original Company was deleted and is no longer visible in the UI.

### Steps to reproduce:

1. Create a Company with a domain (e.g. "example.com").
2. Delete the Company.
3. Try to create a new Company with the same domain name.
4. Error appears when typing in the domain field.

### Expected behavior

Deleted Companies should not block reuse of their domain names, OR deleted companies should simply become Archived so they can be restored at a later time since the data is kept in the database anyway.

### Technical inputs

Domain uniqueness check likely includes soft-deleted records (with deletedAt).
Workaround: Manually remove deletedAt from database to restore visibility.

LTan-101104:
Hi @jeremymoll, I would like to fix this issue. 

I have tried running locally and make some tests. As you said, the issue lies in domain uniqueness checking does not exclude soft-deleted records, adding another condition to only check when deletedAt is null should work.

LTan-101104:
Already made the PR #17159 

Feel free to make any comments/ suggestions and I will try to fix it. Appreciate your attention/ help.

etiennejouan:
Hi @jeremymoll 

If you're blocked by a soft deleted company, you can destroy it by accessing 'soft deleted records' (available in command menu in UI). Also, soft deleted records are destroy after a 14 days retention period.

We deliberately included soft deleted records in the uniqueness constraints. Otherwise, we would have the same issue when restoring records. There is no solution that fits perfectly. In the longer term, we will allow the creation of uniqueness constraints with filters, which will be able to resolve the issue according to business needs.

PR #17159, opened 2026-01-15
[Company record] fix unable to reuse the domain from a deleted record for the current company

LTan-101104:
## Summary ##
This aims to fix issue #16978 
When soft-deleting a company in Twenty, the domainName unique constraint was still enforcing uniqueness against deleted records. This prevented users from creating a new company with the same domain name as a previously deleted company.

## Root Cause ##
The domainNameUniqueIndex on the Company entity was defined without a partial index condition, meaning it applied to ALL records regardless of their deletedAt status. Since Twenty uses soft delete (setting deletedAt timestamp rather than actually deleting), the unique constraint blocked legitimate operations.

## Changes ##
```compute-company-standard-flat-index-metadata.util.ts```:

Added ``` indexWhereClause: '"deletedAt" IS NULL' ``` to the domainNameUniqueIndex definition
This makes the unique constraint only apply to active (non-deleted) records.


etiennejouan (commented):
Hi @LTan-101104, thank you so much for your PR but the soft deleted record are included in uniqueness constraint on purpose. 
Soft deleted records are destroy periodically. 
If you're blocked you can still destroy the record (in UI 'See deleted records' in command menu)

LTan-101104:
@etiennejouan I see, thank you for your review and explanation.

PR #17160, opened 2026-01-15
[Breaking: DEPLOY SERVER BEFORE FRONT] fix: allow custom domain without Cloudflare API key

### LINKED ISSUE #17101: Setting workspace custom domain fails unless CLOUDFLARE_API_KEY is set
## Bug Description

When self-hosting TwentyCRM in **multi-workspace** mode, trying to set a **workspace custom domain** fails with:

> `Environment variable CLOUDFLARE_API_KEY must be defined to use this feature.`

This blocks using workspace custom domains unless Cloudflare credentials are provided, even though the UI text for “Custom Domain” implies manual DNS configuration should be sufficient (“Set the name of your custom domain and configure your DNS records.”).

### Context / what I’m trying to do

- Multi-workspace base domains work fine:
  - `foo.crm.example.com`
  - `bar.crm.example.com`
  - `baz.crm.example.com`
- I want an additional alias custom domain for one workspace:
  - `crm.baz.com` → should map to workspace `baz`

### Steps to reproduce

1. Self-host TwentyCRM with multi-workspace enabled (e.g. `IS_MULTIWORKSPACE_ENABLED=true`, `DEFAULT_SUBDOMAIN=app`, `SERVER_URL=https://app.crm.example.com`).
2. Ensure multi-workspace subdomains work (e.g. `baz.crm.example.com` opens the `baz` workspace).
3. Go to workspace **baz** → **Settings** → **Domains** → **Custom Domain**.
4. Enter `crm.baz.com` and save.

### Actual behavior

Saving shows the error:

- `Environment variable CLOUDFLARE_API_KEY must be defined to use this feature.`

This prevents saving the custom domain unless Cloudflare env vars are configured.

**Example:**  
- Attempt to set workspace Custom Domain to `crm.baz.com`
- Result: error popup/banner `"Environment variable CLOUDFLARE_API_KEY must be defined to use this feature."`

## Expected behavior

Workspace **Custom Domain** should be usable with manual DNS configuration (CNAME/A record) and reverse proxy / TLS handled externally, without requiring Cloudflare-specific environment variables.

If Cloudflare is only intended for an *optional* “automated provisioning” path, the UI should either:
- not require Cloudflare for workspace custom domains, or
- clearly label the feature as “Cloudflare-only”, or
- provide a non-Cloudflare/manual path.

## Technical inputs

**Example:**  
- Deployment: Self-hosted (Docker / docker-compose)
- Mode: Multi-workspace enabled
- Working workspace hostnames: `foo.crm.example.com`, `bar.crm.example.com`, `baz.crm.example.com`
- Desired workspace custom domain: `crm.baz.com`
- DNS provider: NOT Cloudflare
- Reverse proxy / TLS termination: external nginx
- Twenty version / image tag: v1.12.0
- Browser: Firefox
- Relevant logs: `Setting workspace custom domain fails unless CLOUDFLARE_API_KEY is set`

Notes:
- `CLOUDFLARE_API_KEY` is intentionally NOT set because Cloudflare is not used.
- Error appears specifically when saving workspace Custom Domain.

VedantMadane:
## Summary

Fixes #17101

When self-hosting TwentyCRM, users can now set workspace custom domains without requiring CLOUDFLARE_API_KEY to be configured. This enables manual DNS configuration for those not using Cloudflare.

## Changes

- Added isCloudflareConfigured method to DnsManagerService
- Modified all Cloudflare-dependent methods to gracefully handle missing configuration
- Added getManualDnsRecords helper that provides DNS configuration instructions when Cloudflare is not available
- isHostnameWorking returns true in manual mode, allowing the domain to be saved and enabled
- Added comprehensive tests for non-Cloudflare scenarios

## Behavior

When CLOUDFLARE_API_KEY is set: Works exactly as before with automatic Cloudflare provisioning

When CLOUDFLARE_API_KEY is NOT set:
- Custom domain can be saved to the database
- User receives manual DNS configuration instructions
- Domain is marked as working, user is responsible for external DNS and TLS configuration

## Testing

Added tests covering non-Cloudflare scenarios. Full test suite requires Docker which was not run locally.

---

> [!NOTE]
> Introduces a Cloudflare integration feature flag and wires it through server and front to gate the custom domain UI.
> 
> - Server: adds `isCloudflareIntegrationEnabled` to `ClientConfig`, computed from `CLOUDFLARE_API_KEY` and `CLOUDFLARE_ZONE_ID` in `client-config.service`; updates GraphQL schema and unit tests.
> - Frontend: adds `isCloudflareIntegrationEnabledState`, extends `ClientConfig` type, sets the flag in `useClientConfig`, and conditionally renders `SettingsCustomDomain` in `SettingsDomain` when enabled.
> - Updates mocked client config to include the new flag.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0c76dde03b1ae940a9dae65e8673c3ba4cfec9dc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


FelixMalfait:
Thanks @VedantMadane! What you did was good but I still think we don't want to support this feature for now. This is a very niche use-case and we're trying to make Twenty much more stable and bug-free, that means reducing the surface area and not exposing this which wouldn't be properly tested (our only production use-case is with Cloudflare so if we were to break this we wouldn't see it).

FelixMalfait (approved):

nexusai97-gif:
I've analyzed the GitHub content, and it appears to be written in English. Here's my reply: 

Okay, I own an API with multiple models and high performance at a cheap price. Message me on Telegram @Nexusai97.

PR #17162, opened 2026-01-15
[Dashboards] [Performance] Bar chart re renders fix

ehconitin:
before - 

https://github.com/user-attachments/assets/c9b4a60a-d4bd-43d5-8621-54777a326171

after - 

https://github.com/user-attachments/assets/74358810-067e-44f7-82f1-f4dcd7956956


packages/twenty-front/src/modules/page-layout/widgets/graph/graphWidgetBarChart/components/GraphWidgetBarChart.tsx
```diff
@@ -107,6 +107,15 @@ export const GraphWidgetBarChart = ({
   const [chartHeight, setChartHeight] = useState<number>(0);
   const containerRef = useRef<HTMLDivElement>(null);
 
+  const debouncedSetChartWidth = useDebouncedCallback(
```
ehconitin: @bosiraphael unrelated to chart rerenders -- but wdyt about this? 
In terms of reducing these state calls when resizing?

bosiraphael (approved):
LGTM

PR #17167, opened 2026-01-15
[Apps] App misc - fixes + settings permissions for apps + uploadFile

ijreilly:
In this PR

- handle settings permission check for applications. Until then this was unhandled and applications could not perform actions requiring settings permissions, even if they were granted them

- fix ties to attachment, noteTarget etc: 
When an object had their fields synchronized in the app, the system fields created as a side-effect of the object creation (relations to noteTarget, attachment, taskTarget, favorites, timelineActivities - created with `isCustom: true`, not sure that is correct btw) were then deleted because they are not declared in the app, and identified as deletable because of `isCustom: true`. 
Updating the logic to exclude system fields from the logic that detects fields to delete.
I think this outline the confusion we have around isCustom, isSystem etc.

- introduce uploadFile util in generated twenty client as it cannot be handled by the client's query / mutation. I had to use this for my invoicing app

### DIFF
--- a/packages/twenty-sdk/src/cli/services/generate.service.ts
+++ b/packages/twenty-sdk/src/cli/services/generate.service.ts
@@ -112,9 +112,9 @@ export default class Twenty {
     fileBuffer: Buffer,
     filename: string,
     contentType: string = 'application/octet-stream',
+    fileFolder: string = 'Attachment',
   ): Promise<{ path: string; token: string }> {
     const form = new FormData();
-    const fileFolder = "Attachment";
 
     form.append('operations', JSON.stringify({
       query: \`mutation UploadFile($file: Upload!, $fileFolder: FileFolder) {
--- a/packages/twenty-sdk/src/cli/services/generate.service.ts
+++ b/packages/twenty-sdk/src/cli/services/generate.service.ts
@@ -1,12 +1,12 @@
-import { generate } from '@genql/cli';
-import chalk from 'chalk';
-import { join, resolve } from 'path';
 import { ApiService } from '@/cli/services/api.service';
 import { ConfigService } from '@/cli/services/config.service';
+import { generate } from '@genql/cli';
+import chalk from 'chalk';
 import * as fs from 'fs-extra';
+import { join, resolve } from 'path';
 import {
-  DEFAULT_API_URL_NAME,
   DEFAULT_API_KEY_NAME,
+  DEFAULT_API_URL_NAME,
 } from 'twenty-shared/application';
 
 export const GENERATED_FOLDER_NAME = 'generated';
@@ -106,6 +106,41 @@ export default class Twenty {
 
   mutation<R extends MutationGenqlSelection>(request: R & { __name?: string }) {
     return this.client.mutation(request);
+  }
+
+  async uploadFile(
+    fileBuffer: Buffer,
+    filename: string,
+    contentType: string = 'application/octet-stream',
+  ): Promise<{ path: string; token: string }> {
+    const form = new FormData();
+    const fileFolder = "Attachment";
+
+    form.append('operations', JSON.stringify({
+      query: \`mutation UploadFile($file: Upload!, $fileFolder: FileFolder) {
+        uploadFile(file: $file, fileFolder: $fileFolder) { path token }
+      }\`,
+      variables: { file: null, fileFolder },
+    }));
+    form.append('map', JSON.stringify({ '0': ['variables.file'] }));
+    form.append('0', new Blob([fileBuffer], { type: contentType }), filename);
+
+
+    const response = await fetch(\`\${process.env.TWENTY_API_URL}/graphql\`, {
+      method: 'POST',
+      headers: {
+        Authorization: \`Bearer \${process.env.TWENTY_API_KEY}\`,
+      },
+      body: form,
+    });
+
+    const result = await response.json();
+
+    if (result.errors) {
+      throw new GenqlError(result.errors, result.data);
+    }
+
+    return result.data.uploadFile;
   }
 }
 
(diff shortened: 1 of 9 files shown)


packages/twenty-server/src/engine/core-modules/application/application-sync.service.ts
```diff
@@ -690,12 +691,18 @@ export class ApplicationSyncService {
...
         objectId: objectToUpdate.id,
         workspaceId,
         applicationId,
       });
+
+      await this.syncRelations({
+        objectsToSync: [objectToSync],
+        workspaceId,
+        applicationId,
+      });
```
martmull: is this expected? I though it was already called isn't it?

ijreilly: I had only done it for objectToCreate, not objectToUpdate !

martmull: I don't get it, I though we could not call this syncRelation in syncObjects method because syncRelations needs all objects to be created

ijreilly: you are right je me suis emmêlé les pinceaux, je revert

packages/twenty-sdk/src/cli/services/generate.service.ts
```diff
@@ -107,6 +107,41 @@ export default class Twenty {
   mutation<R extends MutationGenqlSelection>(request: R & { __name?: string }) {
     return this.client.mutation(request);
   }
+
+  async uploadFile(
```
ijreilly, line 111: genQL's query and mutation above only accept JSON which is not suitable for uploading files
This is the way this is done by twenty

packages/twenty-server/src/engine/guards/settings-permission.guard.ts
```diff
@@ -47,6 +47,7 @@ export const SettingsPermissionGuard = (
           setting: requiredPermission,
           workspaceId,
           apiKeyId: ctx.getContext().req.apiKey?.id,
+          applicationId: ctx.getContext().req.application?.id,
```
ijreilly, line 50: @FelixMalfait you did not have to do something like this for AI because they work with a apiKey being granted a role, is that right ? 

Double checking before I extend this to everywhere `userHasWorkspaceSettingPermission` is called

packages/twenty-sdk/src/cli/services/generate.service.ts
```diff
@@ -107,6 +107,41 @@ export default class Twenty {
...
+    fileBuffer: Buffer,
+    filename: string,
+    contentType: string = 'application/octet-stream',
+  ): Promise<{ path: string; token: string }> {
+    const form = new FormData();
+    const fileFolder = "Attachment";
```
martmull: can this be a parameter?

packages/twenty-sdk/src/cli/services/generate.service.ts
```diff
@@ -107,6 +107,41 @@ export default class Twenty {
   mutation<R extends MutationGenqlSelection>(request: R & { __name?: string }) {
     return this.client.mutation(request);
   }
+
+  async uploadFile(
+    fileBuffer: Buffer,
+    filename: string,
+    contentType: string = 'application/octet-stream',
+  ): Promise<{ path: string; token: string }> {
+    const form = new FormData();
```
martmull, line 117: what is `FormData`?

packages/twenty-sdk/src/cli/services/generate.service.ts
```diff
@@ -107,6 +107,41 @@ export default class Twenty {
...
+      query: \`mutation UploadFile($file: Upload!, $fileFolder: FileFolder) {
+        uploadFile(file: $file, fileFolder: $fileFolder) { path token }
+      }\`,
+      variables: { file: null, fileFolder },
+    }));
+    form.append('map', JSON.stringify({ '0': ['variables.file'] }));
+    form.append('0', new Blob([fileBuffer], { type: contentType }), filename);
+
+
+    const response = await fetch(\`\${process.env.TWENTY_API_URL}/graphql\`, {
```
martmull, line 129: should not be process.env.TWENTY_API_URL but should be `merged` (see in Twenty constructor)

packages/twenty-sdk/src/cli/services/generate.service.ts
```diff
@@ -107,6 +107,41 @@ export default class Twenty {
...
+      method: 'POST',
+      headers: {
+        Authorization: \`Bearer \${process.env.TWENTY_API_KEY}\`,
```
martmull, line 132: same

martmull (commented):
Nice code, some comments

packages/twenty-sdk/src/cli/services/generate.service.ts
ijreilly: Yes I hesitated - I think for now we only accept the `Attachment` fileFolder but it may change with @etiennejouan 's work on file field type. I will make it a parameter

packages/twenty-sdk/src/cli/services/generate.service.ts
ijreilly: it's a node-native object that is used to build a multipart/form-data HTTP body request to upload a file (this format is required, we cant use a json).
In our FE implementation, our Apollo client is taking care of transforming the request in multipart/form-data.

packages/twenty-sdk/src/cli/services/generate.service.ts
ijreilly: `merged` is not exposed from the Client at the moment from my understanding. should it be ?

charlesBochet (approved):
@ijreilly I'm merging this PR as we will re-organize a bit the sdk folder

packages/twenty-sdk/src/cli/services/generate.service.ts
martmull: merged can be exposed if saved in a private attribute in the constructor

PR #17172, opened 2026-01-15
File v2 - update core.file table

etiennejouan:
For the following, I'll create the issues first


packages/twenty-server/src/engine/core-modules/file/entities/file.entity.ts
```diff
@@ -17,18 +22,28 @@ export class FileEntity extends WorkspaceRelatedEntity {
   @PrimaryGeneratedColumn('uuid')
   id: string;
 
-  @Column({ nullable: false })
-  name: string;
+  @Column({ nullable: false, type: 'uuid' })
+  applicationId: string;
+
+  @ManyToOne('ApplicationEntity', {
+    onDelete: 'CASCADE',
```
etiennejouan: @charlesBochet @Weiko 

what do you think of this one ? nullable true + set null better ?

Concerning workspace files (the ones linked to file field), they would point to the application defining the file field.

Still don't know how we can trigger file deletion in storage when file is cascade delete because of an application deletion. Maybe to early ?

prastoin: ~~I would tend to delete cascade on this, that's what all `syncableEntity` have too~~
Having a cron that will iterate over all the files with an applicationId null is always better than comparing the contrary
So yeah maybe null cascade with cleaning cron would be pretty good, though it would also mean that the universalIdentifier is still present and could implies constraint collisions though

packages/twenty-server/src/engine/core-modules/file/entities/file.entity.ts
```diff
@@ -17,18 +22,28 @@ export class FileEntity extends WorkspaceRelatedEntity {
...
+    nullable: false,
+  })
+  @JoinColumn({ name: 'applicationId' })
+  application: Relation<ApplicationEntity>;
 
   @Column({ nullable: false })
-  fullPath: string;
+  path: string;
```
prastoin, line 36: Unless I'm mistaken we should have a constraint on this at `workspaceId` + `applicationId` + `path` ?

packages/twenty-server/src/engine/core-modules/file/entities/file.entity.ts
prastoin: Having a `deletedAt` would make the soft deletion more explicit

etiennejouan: We have not discuss of soft deletion logic. Maybe to add when it will be used ?

Regarding null cascade, it could be an option ! I don't know how we can then enforce dev to provide applicationId, if not files would be deleted

packages/twenty-server/src/engine/core-modules/file/entities/file.entity.ts
etiennejouan: It could cause erorr if applicationId is nullable

packages/twenty-server/src/engine/core-modules/file/entities/file.entity.ts
prastoin: ApplicationId being required and non nullable is enough no ?

packages/twenty-server/src/engine/core-modules/file/entities/file.entity.ts
```diff
@@ -5,30 +5,49 @@ import {
...
-  type: string;
-
   @CreateDateColumn({ type: 'timestamptz' })
   createdAt: Date;
+
+  @UpdateDateColumn({ type: 'timestamptz' })
+  updatedAt: Date;
+
+  @Column({ nullable: false, default: false })
+  isStaticAsset: boolean;
```
Weiko, line 51: Still not a great fan of this bool name but we said we were ok for now to change later if needed

packages/twenty-server/src/engine/core-modules/file/services/file-metadata.service.ts
```diff
@@ -45,10 +45,8 @@ export class FileMetadataService {
     }
 
     const createdFile = this.fileRepository.create({
-      name: filename,
-      fullPath: files[0].path,
+      path: files[0].path,
       size: file.length,
-      type: mimeType,
```
Weiko, line 51: Same, not a great fan of removing type, this let MIME type inference to the caller (through the extension) or some header parsing in the file before serving it. Again we apparently were ok for that but just wanted to share my thoughts

packages/twenty-server/src/engine/core-modules/file/entities/file.entity.ts
```diff
@@ -5,30 +5,49 @@ import {
...
 
   @Column({ nullable: false, type: 'bigint' })
   size: number;
 
-  @Column({ nullable: false })
-  type: string;
-
   @CreateDateColumn({ type: 'timestamptz' })
   createdAt: Date;
+
```
Weiko, line 43: should we introduce a label already? Since path is unique, this might break things like renaming if that's something we want but I'm guessing this will be ok once we introduce a label?

packages/twenty-server/src/engine/core-modules/file/entities/file.entity.ts
etiennejouan: You can have null cascade and non nullable ? 🤔

packages/twenty-server/src/engine/core-modules/file/entities/file.entity.ts
etiennejouan: I understood label would be hold by FILE-type field value and only for workspace use

packages/twenty-server/src/engine/core-modules/file/services/file-metadata.service.ts
etiennejouan: I'm not sure to understand. At upload can't we deduct the MIME type and use it for the path ?

packages/twenty-server/src/engine/core-modules/file/entities/file.entity.ts
prastoin: You're right, there's a middle to find
We could add a constraint making it nullable only if soft deleted

prastoin:
Hey, small recap of what we've discussed irl so far ( can change ofc )
As discussed we will land on having cascade restrict policy on files and appId
We will always soft delete files before removing an app ( set deletedAt + nullify applicationId, making app id nullable + a constraint that assert it's only nullable for soft deleted rows )
A cron job or an async job will scan and hard delete the soft deleted and their s3 file entries

You won't be able to remove an app until all its related files have been soft deleted or hard deleted in prior

etiennejouan:
> Hey, small recap of what we've discussed irl so far ( can change ofc ) As discussed we will land on having cascade restrict policy on files and appId We will always soft delete files before removing an app ( set deletedAt + nullify applicationId, making app id nullable + a constraint that assert it's only nullable for soft deleted rows ) A cron job or an async job will scan and hard delete the soft deleted and their s3 file entries
> 
> You won't be able to remove an app until all its related files have been soft deleted or hard deleted in prior

- cascade restrict
- nullable applicationId (and not only for soft deleted file, in case a "file field" file is upload but not yet link to record)

packages/twenty-shared/src/utils/files/extractFolderPathFilenameAndType.util.ts
```diff
@@ -0,0 +1,19 @@
+export const extractFolderPathFilenameAndType = (
+  fullPath: string,
+): {
+  folderPath: string;
+  filename: string;
+  type: string;
+} => {
+  if (!fullPath || typeof fullPath !== 'string') {
```
prastoin: ***Polish:*** `isNonEmptyString`

packages/twenty-shared/src/utils/files/extractFolderPathFilenameAndTypeOrThrow.util.ts
```diff
@@ -0,0 +1,19 @@
+export const extractFolderPathFilenameAndType = (
+  fullPath: string,
+): {
+  folderPath: string;
+  filename: string;
+  type: string;
+} => {
+  if (!fullPath || typeof fullPath !== 'string') {
+    throw new Error('Invalid fullPath provided');
```
prastoin, line 11: ***Question:*** Should this return null instead or a success failure result in order to let the caller determine what to do ?

packages/twenty-shared/src/utils/files/extractFolderPathFilenameAndType.util.ts
```diff
@@ -0,0 +1,19 @@
+export const extractFolderPathFilenameAndType = (
```
prastoin: ***Polish:*** Should be called `orThrow`

prastoin (approved):
Looks good to me !

packages/twenty-shared/src/utils/files/extractFolderPathFilenameAndTypeOrThrow.util.ts
etiennejouan: Should be enhance after all refactor. Prefer minimizing change in this PR

PR #17177, opened 2026-01-15
Add get current workflow version tool

thomtrp:
This is required so an agent can easily understand what version should be updated from the workflow show page.


FelixMalfait (approved):

Weiko (approved):
LGTM

PR #17180, opened 2026-01-15
[Dashboards] [ai] update dashboard tools schema and skill with correct config field names

ehconitin:
CleanShot 2026-01-16 at 00 55 56


FelixMalfait (approved):
Nice!

PR #17175, opened 2026-01-15
Front Extensibility: Introduce Front Component Entity

charlesBochet:
As part of the extensibility effort, we are introducing a new engine entity called "Front Component". This represents a dynamic react component that will be rendered in CommandMenu actions or in PageLayout widgets

This PR introduce the entity and all the necessary boilerplate to make it syncable and cachable in the engine

PR #17181, opened 2026-01-15
Add command menu item entity

abdulrahmancodes:


FelixMalfait:
Nice, could you please check Paul's comments and Charles' changes here? https://github.com/twentyhq/twenty/pull/17175 ; see if any applies to you. Thanks!

charlesBochet:
@abdulrahmancodes the other PR is ready to merge: https://github.com/twentyhq/twenty/pull/17175 you can replicate everything

Also: I'm not sure about having workflowVersionId on commandMenuItem long term, relation to workspaceSchema won't be supported and dependencies are in the wrong way IMO. This highlight a design issue: having workflowVersion in the workspaceSchema feels wrong as this seems to be an engine capability.

Idea to keep it fully generic while keeping workflows in the workspaceSchema (vision, I'm not saying we should do it now):
- CommandMenuItem always point to a FrontComponent
- Workflow module actually register a FrontComponent that triggers a workflow serverless that launch the workflow

I'm fine keeping workflowVersionId for now to ease migration and keep this for later but let's not try to add a real relation on top of that then. Let's keep it basic uuid

@FelixMalfait wdyt?

packages/twenty-server/src/engine/metadata-modules/command-menu-item/entities/command-menu-item.entity.ts
```diff
@@ -0,0 +1,58 @@
...
+  @Column({
+    type: 'enum',
+    enum: CommandMenuItemAvailabilityType,
+    nullable: false,
+    default: CommandMenuItemAvailabilityType.GLOBAL,
+  })
+  availabilityType: CommandMenuItemAvailabilityType;
+
+  @Column({ nullable: true, type: 'varchar' })
+  availabilityObjectNameSingular: string | null;
```
FelixMalfait: Why objectNameSingular and not a FK to the object metadata? It would be more robust

abdulrahmancodes: Updated to use a foreign key

FelixMalfait (approved):

FelixMalfait:
(tests broken)

charlesBochet (approved):
Very nice!

charlesBochet:
@abdulrahmancodes I'm sorry, I've just merged my PR which created conflicts :( should be quite straightforward to fix hopefully!

PR #17185, opened 2026-01-16
[Dashboards] [ai] expand widget schemas with subfield grouping, color, and styling options

ehconitin:
Got rid of filters for now -- filters would need extra tailoring since they depend on field names/types and the RecordGqlOperationFilter DSL (nested AND/OR and composite subfields), which the AI can’t reliably construct without additional metadata and skills 

created a followup to tackle the same https://github.com/twentyhq/core-team-issues/issues/2108

### DIFF
--- a/packages/twenty-server/src/modules/dashboard/tools/schemas/widget.schema.ts
+++ b/packages/twenty-server/src/modules/dashboard/tools/schemas/widget.schema.ts
@@ -358,9 +358,6 @@ const richTextConfigSchema = z.object({
       blocknote: z.string().nullable().optional(),
       markdown: z.string().nullable().optional(),
     })
-    .refine((value) => value.blocknote != null || value.markdown != null, {
-      message: 'body must include blocknote or markdown',
-    })
     .describe('Rich text content (RichTextV2Body)'),
 });
 
--- a/packages/twenty-server/src/modules/dashboard/tools/schemas/widget.schema.ts
+++ b/packages/twenty-server/src/modules/dashboard/tools/schemas/widget.schema.ts
@@ -1,3 +1,4 @@
+import { isNumber } from '@sniptt/guards';
 import { ObjectRecordGroupByDateGranularity } from 'twenty-shared/types';
 import { z } from 'zod';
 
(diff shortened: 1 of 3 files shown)


packages/twenty-server/src/modules/dashboard/tools/schemas/widget.schema.ts
```diff
@@ -1,9 +1,174 @@
...
 import { AggregateOperations } from 'src/engine/api/graphql/graphql-query-runner/constants/aggregate-operations.constant';
+import { AxisNameDisplay } from 'src/engine/metadata-modules/page-layout-widget/enums/axis-name-display.enum';
+import { BarChartGroupMode } from 'src/engine/metadata-modules/page-layout-widget/enums/bar-chart-group-mode.enum';
+import { BarChartLayout } from 'src/engine/metadata-modules/page-layout-widget/enums/bar-chart-layout.enum';
+import { GraphOrderBy } from 'src/engine/metadata-modules/page-layout-widget/enums/graph-order-by.enum';
 import { WidgetConfigurationType } from 'src/engine/metadata-modules/page-layout-widget/enums/widget-configuration-type.type';
 import { WidgetType } from 'src/engine/metadata-modules/page-layout-widget/enums/widget-type.enum';
 
+// Chart color options (MAIN_COLOR_NAMES plus 'auto').
+// should we export MAIN_COLOR_NAMES from twenty-ui to twenty-shared and use that here?
```
FelixMalfait, line 14: Yes!

FelixMalfait: Ah no actually, didn't realize it was twenty-ui not twenty-shared already. 
Move it to twenty-shared but don't "export from twenty-ui" (maybe that's what you meant already) as we don't want to create a dependency between twenty-ui and twenty-server

FelixMalfait (approved):

packages/twenty-server/src/modules/dashboard/tools/schemas/widget.schema.ts
ehconitin: was thinking of some workaround/ hacky way -- simply import the MAIN_COLOR_NAMES const from twenty ui to shared -- and then re export it -- and then import it from shared here 😅

ehconitin: merging this as it is - -will decide what to do of this comment in followups

PR #17186, opened 2026-01-16
Add messageFolderId to MessageChannelMessageAssociation

neo773:
## Changes
- Extended Redis Client to add `HSET` command
- Added `messageFolderId` to `MessageChannelMessageAssociation`
- Refactored `MessagingDeleteFolderMessagesService` to use `messageFolderId` instead of making network calls

## Why
- Current `FOLDER_DELETION` approach doesn't work for IMAP which relied on querying remote origin.

Screenshot 2026-01-16 at 5 33 16 PM


charlesBochet:
@neo773 Do we want to go this way? If we do, we will need to keep them in sync, I think it's too heavy to maintain at this point?

Why is it needed?
It also does not make sense for messages on gmail that have multiple folders. If we want to make it clean, we should have a many to many relation between message and messageFolder

Can't we avoid the edge case and reschedule a full message list fetch under some condition (trade off)

charlesBochet (changes requested):
Don't think it's a good idea :)

neo773:
Closing this  to minimize noise, will reopen a new PR soon

PR #17188, opened 2026-01-16
Improve image upload error handling and validation

Achieve3318:
- Add URL validation in getImageBufferFromUrl utility
- Add response status validation and content-type checking
- Add timeout and connection error handling with specific error messages
- Validate buffer is not empty before processing
- Validate file type detection results before proceeding
- Ensure detected file type is actually an image format
- Add proper type safety for Axios error handling

This improves robustness when uploading images from URLs by:
- Preventing invalid URLs from being processed
- Providing clear error messages for different failure scenarios
- Ensuring only valid image files are processed
- Handling network errors gracefully

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/file/file-upload/services/file-upload.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/file/file-upload/services/file-upload.service.ts
@@ -118,10 +118,6 @@ export class FileUploadService {
       this.httpService.axiosRef,
     );
 
-    if (buffer.length === 0) {
-      throw new Error('Received empty image buffer from URL');
-    }
-
     const type = await FileType.fromBuffer(buffer);
 
     if (!type || !type.ext || !type.mime) {
--- a/packages/twenty-server/src/engine/core-modules/file/file-upload/services/file-upload.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/file/file-upload/services/file-upload.service.ts
@@ -118,12 +118,28 @@ export class FileUploadService {
       this.httpService.axiosRef,
     );
 
+    if (buffer.length === 0) {
+      throw new Error('Received empty image buffer from URL');
+    }
+
     const type = await FileType.fromBuffer(buffer);
+
+    if (!type || !type.ext || !type.mime) {
+      throw new Error(
+        'Unable to detect image type from buffer. The file may not be a valid image format.',
+      );
+    }
+
+    if (!type.mime.startsWith('image/')) {
+      throw new Error(
+        `Detected file type is not an image: ${type.mime}. Please provide a valid image URL.`,
+      );
+    }
 
     return await this.uploadImage({
       file: buffer,
-      filename: `${v4()}.${type?.ext}`,
-      mimeType: type?.mime,
+      filename: `${v4()}.${type.ext}`,
+      mimeType: type.mime,
       fileFolder,
       workspaceId,
     });
--- a/packages/twenty-server/src/utils/image.ts
+++ b/packages/twenty-server/src/utils/image.ts
@@ -68,18 +68,17 @@ export const getImageBufferFromUrl = async (
     }
     
     if (error instanceof Error) {
-      const axiosError = error as AxiosError;
       if (
         axiosError.code === 'ECONNABORTED' ||
-        (error instanceof Error && error.message.includes('timeout'))
+        error.message.includes('timeout')
       ) {
         throw new Error(`Request timeout while fetching image from URL: ${url}`);
       }
       if (
         axiosError.code === 'ENOTFOUND' ||
         axiosError.code === 'ECONNREFUSED' ||
-        (error instanceof Error && error.message.includes('ENOTFOUND')) ||
-        (error instanceof Error && error.message.includes('ECONNREFUSED'))
+        error.message.includes('ENOTFOUND') ||
+        error.message.includes('ECONNREFUSED')
       ) {
         throw new Error(`Failed to connect to image URL: ${url}`);
       }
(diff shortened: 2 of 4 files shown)


packages/twenty-server/src/engine/core-modules/file/file-upload/services/file-upload.service.ts
```diff
@@ -120,10 +120,22 @@ export class FileUploadService {
 
     const type = await FileType.fromBuffer(buffer);
 
+    if (!type || !type.ext || !type.mime) {
+      throw new Error(
+        'Unable to detect image type from buffer. The file may not be a valid image format.',
+      );
+    }
+
+    if (!type.mime.startsWith('image/')) {
```
etiennejouan, line 129: Not sure we want to be too much restricted on image type. What about SVG ? or old format ?

packages/twenty-server/src/utils/image.ts
```diff
@@ -26,9 +26,64 @@ export const getImageBufferFromUrl = async (
...
+    const axiosResponse = axiosError.response;
+    if (axiosResponse) {
+      throw new Error(
+        `Failed to fetch image: HTTP ${axiosResponse.status} from ${url}`,
+      );
+    }
+    
+    if (error instanceof Error) {
+      if (
+        axiosError.code === 'ECONNABORTED' ||
```
etiennejouan, line 74: why do we have axiosError here ? It should be filter in previous condition no ?

packages/twenty-server/src/utils/image.ts
```diff
@@ -26,9 +26,64 @@ export const getImageBufferFromUrl = async (
...
+        error.message.includes('timeout')
+      ) {
+        throw new Error(`Request timeout while fetching image from URL: ${url}`);
+      }
+      if (
+        axiosError.code === 'ENOTFOUND' ||
```
etiennejouan, line 82: same here

packages/twenty-server/src/utils/image.ts
```diff
@@ -26,9 +26,64 @@ export const getImageBufferFromUrl = async (
...
+      ? response.data.length
+      : response.data.byteLength;
 
-  return Buffer.from(response.data, 'binary');
+    if (bufferLength === 0) {
+      throw new Error('Received empty response from image URL');
+    }
+
+    const contentType = response.headers['content-type'];
+    if (contentType && !contentType.startsWith('image/')) {
```
etiennejouan, line 55: And what happens if server doesn't send content type ?

etiennejouan (changes requested):
Hi @Angel98518 
Thank you so much for your contribution.
I let you some comments. 
The CI is not passed because of lint issue. 
Good PR adding some guard on image downloading. I would prefer the util error handling to be as light as possible. Not sure we need all this nuance of error messages.

Achieve3318:
Hi.
@charlesBochet 
can you review my PR?
Thanks

packages/twenty-server/src/engine/core-modules/file/file-upload/services/file-upload.service.ts
Achieve3318: SVG files: SVG uses image/svg+xml, which starts with `image/`, so it passes this check. The codebase also has specific SVG handling in _sanitizeFile() (line 61).
Older formats: The `startsWith('image/')` check follows standard MIME conventions. Common formats (PNG, JPEG, GIF, BMP, TIFF, WebP, SVG, ICO, etc.) use MIME types starting with image/.

If `FileType.fromBuffer()` detects a type with a valid image extension, we could make this more flexible. Options:
Keep the current approach — rely on MIME type validation (standard and safe).
Make it more flexible — also accept files if FileType detects an image extension (e.g., `['jpg', 'jpeg', 'png', 'gif', 'svg', 'webp', 'bmp', 'ico', 'tiff'].includes(type.ext)`).
Which do you prefer? We want to avoid blocking legitimate formats while maintaining security.

packages/twenty-server/src/utils/image.ts
Achieve3318: The logic handles two types of Axios errors:

1. Axios errors with response (lines 66-70): HTTP errors (4xx, 5xx) that include a response object with a status code.
2. Axios errors without response (lines 73-88): Network-level errors (timeout, DNS failure, connection refused) where response is undefined, but the error still has an `axiosError.code` property (e.g., 'ECONNABORTED', 'ENOTFOUND', 'ECONNREFUSED').

If an Axios error has a response, we throw immediately in the first block. The second block handles Axios errors that don't have a response (network failures) as well as other Error types that might have timeout/DNS errors in their message.
So `axiosError.code` is only checked for network-level Axios errors that didn't have a response. Does this make sense, or would you prefer a different approach?

packages/twenty-server/src/utils/image.ts
Achieve3318: if content-type is missing, we'll still catch non-image files through the `FileType.fromBuffer()` validation later.

packages/twenty-server/src/utils/image.ts
Achieve3318: same as above

Achieve3318:
@charlesBochet 
I sent you comment.
can you give me feedback?

etiennejouan (approved):
Hi @Angel98518  Thank you for your answers and fixing lint ! Approved, thank you for your contribution

Achieve3318:
@etiennejouan 
I have fixed only image.ts and file-upload.service.ts.
but github-merge-queue bot checked this and show me e2e test failure (workflow-creation.spec.ts, login.setup.ts).
I didn't change these two files.
please let me know the reason

Achieve3318:
@etiennejouan 
can you solve my issue?

Achieve3318:
@github-actions please re-run and merge

PR #17193, opened 2026-01-16
Invoicing app

ijreilly:


packages/twenty-shared/src/constants/QueryDefaultLimitRecords.ts
```diff
@@ -1 +1 @@
-export const QUERY_DEFAULT_LIMIT_RECORDS = 60;
+export const QUERY_DEFAULT_LIMIT_RECORDS = 1000;
```
martmull, line 1: we should remove that change i think

packages/twenty-sdk/package.tgz
martmull, line 1: this file should not be commited

PR #17194, opened 2026-01-16
fix(workflow): enable variable mapping for phone country code

### LINKED ISSUE #17182: Workflow Create Record cannot fully populate Phone field (country selector not mappable)
### Summary

After the fix in #12343 (E.164 phone handling with "No country"), phone numbers can now be written correctly via the API.
However, the same is **not possible via Workflows**, because the Phone field’s country selector is not mappable from workflow variables.

This creates a functional gap between API usage and Workflow automation for international phone numbers.

---

### Environment

- Deployment: Self-hosted (official Docker image)
- Twenty version: latest Docker image
- Feature used: Workflows → Trigger: Webhook → Action: Create Record
- Field type: Phone
- Phone field config: Default country = "No country"

---

### Steps to reproduce

1. Create a Phone field on an object (e.g. Opportunities or Parents)
   - Set default country to **No country**
2. Create a Workflow:
   - Trigger: Webhook
   - Action: Create Record
3. Send a webhook payload with an E.164 phone number, e.g. "+919999999999"
4. Map this payload value to the Phone field in the workflow

---

### Actual behavior

- The phone number is stored as the **national number only** (e.g. `9999999999`)
- The country selector remains unset / default
- The E.164 calling code (`+91`) is silently dropped

---

### Expected behavior

One of the following should be possible:

- Workflow Create Record accepts full E.164 values and preserves them when default country is "No country"
**or**
- Workflow mapping exposes the Phone field’s country selector as a writable/mappable input

This is especially important for international lead ingestion.

---

### Why this matters

- #12343 fixed E.164 handling at the API level, but workflows still cannot achieve parity
- This forces users to abandon the Phone field type and use Text fields for WhatsApp / phone numbers
- It breaks international automation use cases and makes Workflows less capable than direct API usage

---

### Workaround (current)

- Use a Text field to store WhatsApp / phone numbers in E.164 format
- Avoid the Phone field entirely for workflow-driven ingestion

This works but loses the benefits of the Phone field type.

---

### Related issue

- #12343 – Updating phone numbers via API adds incorrect country prefix when "No country" is selected

---

Happy to help test or provide additional details if needed.

AzimKrishna:
# PR Title
fix(workflow): enable variable mapping for phone country code and resolve calling code inference

# Description
Closes #17182

## Problem
As reported in #17182, the **Phone field's country selector was not mappable** from workflow variables in the "Create Record" action. Users could not dynamically set the country based on incoming data (e.g., from a webhook), forcing them to either hardcode the country or face data loss where the calling code (e.g., `+91`) was stripped.

Attempts to simply enable mapping revealed a gap: Since the frontend **cannot determine the numeric calling code** (e.g., `91`) from a variable string (e.g., `{{webhook.country}}`) at design time, it required a mechanism to defer this logic to the backend. Without this, the calling code field remained invalid or unpopulated.

## Solution
This PR solves both issues to fully enable dynamic phone number persistence in workflows.

### Frontend (`packages/twenty-front`)
-   **Enabled Variable Mapping**: Updated `FormCountryCodeSelectInput` to accept variable strings.
-   **Fix Variable Handling**: Modified `FormPhoneFieldInput` to detect when a variable is selected. Instead of passing the ISO code (e.g., "IN") to the calling code field, it now sends an empty string `''`. This delegates the inference logic to the backend.

### Backend (`packages/twenty-server`)
-   **Improved Inference**: Updated `transform-phones-value.util.ts` to treat an empty string `''` calling code as "missing" rather than valid-but-empty.
-   **Constraint Solver**: The transformer now correctly infers the numeric calling code (e.g., `+91`) from the provided `countryCode` (e.g., `IN`) when the calling code is empty.

## Rationale
We implemented **Option 2** (Enabling Country Selector Mapping) over **Option 1** (Full E.164 Parsing in Phone Field) for the following reasons:

1.  **Architecture Alignment**: It preserves the existing UI/Data model where Country Code (`primaryPhoneCountryCode`) and Phone Number (`primaryPhoneNumber`) are distinct fields.
2.  **Ambiguity Resolution**: Inferring Calling Code from Country Code (e.g., `IN` → `+91`) is deterministic. Inferring Country from a raw E.164 string on the backend is ambiguous (e.g., `+1` could be US, CA, or others).
3.  **Variable Validation**: By utilizing the variable mapping mechanism, we bypass the client-side validation in `FormNumberFieldInput` that essentially stripped `+` signs (via JS type coercion), allowing the backend to receive and process the raw data correctly.

## Root Cause Analysis
-   **Frontend Limitation**: The `FormNumberFieldInput` component validates manual input using `+inputValue` (JavaScript number coercion). This treats `+91999` as a positive number and **strips the `+` sign** before sending it to the backend.
-   **Variable Bypass**: When a workflow variable is used, this validation is bypassed, allowing the raw string (e.g. `+91...`) to reach the backend.
-   **Backend Bug**: Previously, if the frontend sent an empty string `""` for the calling code (which happens when no country is selected or when we try to support variables), the backend treated `""` as a valid-but-empty value and prioritized it over any inferred calling code. This caused it to discard the `+91` derived from the number.

## Expected Usage
After this fix, the Country Code selector behaves as follows:

### 1. Mapping Country & Number (Standard)
Map the fields separately. The backend will infer the Calling Code.
```json
{
  "primaryPhoneCountryCode": "IN",      // Variable resolving to ISO code "IN"
  "primaryPhoneNumber": "9876543210"    // National number
}
// Result: Stored with +91 calling code.
```

### 2. Mapping E.164 Number (Bonus Feature)
Map a full E.164 string to the number field. Leave Country as "No country".
```json
{
  "primaryPhoneCountryCode": "",        // No Country / Empty
  "primaryPhoneNumber": "+919876543210" // Variable resolving to E.164
}
// Result: Backend infers Country=IN, CallingCode=+91.
```
*Note: This only works for mapped variables because they bypass the frontend `+` stripping.*

## Testing
-   [x] **Unit Test**: Added `transform-phones-value.util.spec.ts` (new file) to verify:
    -   Inference of `+91` from `{ countryCode: 'IN', callingCode: '' }`.
    -   Preservation of explicitly provided calling codes.
    -   Inference from E.164 numbers (Bonus case).
-   [x] **Manual Verification**:
    -   configured a workflow with a country variable `{{webhook.country}}`.
    -   Verified the "Invalid calling code" error is resolved.
    -   Confirmed the record is created with the correct inferred calling code.

---

> [!NOTE]
> Enables dynamic phone inputs with workflow variables and robust server-side inference.
> 
> - **Backend**: Update `transform-phones-value.util` to treat empty `callingCode` as missing, infer `callingCode` from `countryCode` or from an E.164 `number`, and coerce `number` to string; add unit tests covering preservation, inference, and null/undefined cases.
> - **Frontend**: Allow variable strings in `FormCountryCodeSelectInput` and pass `VariablePicker`; in `FormPhoneFieldInput` detect variable country selections and send empty `primaryPhoneCallingCode` to delegate inference; update Storybook to reflect two variable pickers and variable-driven changes.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 79572724339dc77eb17a3d08ac148bbeeb8602ae. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


FelixMalfait:
Thanks. alot ! This doesn't seem like the clean solution (strange type conversions etc). I'll close since it's probably not the right direction but we can discuss your plan in the issue :)

PR #17195, opened 2026-01-16
Various layout fixes so that RPL v1 look the same as production version

Devessier:
> [!WARNING]
> Most of the quick fixes will be dropped once we release the feature and deprecate the old show pages code. 

This PR fixes padding and alignment issues so that the new Record Page Layout feature looks like the old show pages.

## This PR

CleanShot 2026-01-16 at 17 24 40@2x

CleanShot 2026-01-16 at 17 27 14@2x

## The production

CleanShot 2026-01-16 at 17 25 11@2x

CleanShot 2026-01-16 at 17 27 19@2x


Weiko (approved):
LGTM

PR #17201, opened 2026-01-16
Add permissions to SSE

thomtrp:
Flow:
- fetch user role from context stored in cache - do not support api context yet
- check object permissions
- filter restricted fields on event
- fetch role rls predicate and combine it with the query filter

Do not support dynamic predicates yet.

### DIFF
--- a/packages/twenty-server/src/engine/workspace-event-emitter/__tests__/workspace-event-emitter.service.spec.ts
+++ b/packages/twenty-server/src/engine/workspace-event-emitter/__tests__/workspace-event-emitter.service.spec.ts
@@ -1,0 +1,974 @@
+import { Test, type TestingModule } from '@nestjs/testing';
+
+import {
+  FieldMetadataType,
+  type ObjectsPermissionsByRoleId,
+  type RecordGqlOperationFilter,
+} from 'twenty-shared/types';
+
+import { type FlatEntityMaps } from 'src/engine/metadata-modules/flat-entity/types/flat-entity-maps.type';
+import { type FlatFieldMetadata } from 'src/engine/metadata-modules/flat-field-metadata/types/flat-field-metadata.type';
+import { type FlatObjectMetadata } from 'src/engine/metadata-modules/flat-object-metadata/types/flat-object-metadata.type';
+import { EventStreamService } from 'src/engine/subscriptions/event-stream.service';
+import { SubscriptionService } from 'src/engine/subscriptions/subscription.service';
+import { type EventStreamData } from 'src/engine/subscriptions/types/event-stream-data.type';
+import { WorkspaceCacheService } from 'src/engine/workspace-cache/services/workspace-cache.service';
+import { type WorkspaceEventBatch } from 'src/engine/workspace-event-emitter/types/workspace-event-batch.type';
+import { WorkspaceEventEmitterService } from 'src/engine/workspace-event-emitter/workspace-event-emitter.service';
+
+jest.mock(
+  'src/engine/twenty-orm/utils/build-row-level-permission-record-filter.util',
+  () => ({
+    buildRowLevelPermissionRecordFilter: jest.fn(),
+  }),
+);
+
+jest.mock(
+  'src/engine/twenty-orm/utils/is-record-matching-rls-row-level-permission-predicate.util',
+  () => ({
+    isRecordMatchingRLSRowLevelPermissionPredicate: jest.fn(),
+  }),
+);
+
+const {
+  buildRowLevelPermissionRecordFilter,
+} = require('src/engine/twenty-orm/utils/build-row-level-permission-record-filter.util');
+const {
+  isRecordMatchingRLSRowLevelPermissionPredicate,
+} = require('src/engine/twenty-orm/utils/is-record-matching-rls-row-level-permission-predicate.util');
+
+type MockObjectRecordEvent = {
+  recordId: string;
+  userId?: string;
+  workspaceMemberId?: string;
+  properties: {
+    before?: object;
+    after?: object;
+    updatedFields?: string[];
+    diff?: object;
+  };
+};
+
+describe('WorkspaceEventEmitterService', () => {
+  let service: WorkspaceEventEmitterService;
+  let mockSubscriptionService: jest.Mocked<
+    Pick<SubscriptionService, 'publish' | 'publishToEventStream'>
+  >;
+  let mockEventStreamService: jest.Mocked<
+    Pick<
+      EventStreamService,
+      'getActiveStreamIds' | 'getStreamsData' | 'removeFromActiveStreams'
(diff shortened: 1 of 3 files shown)


packages/twenty-server/src/engine/workspace-event-emitter/__tests__/workspace-event-emitter.service.spec.ts
```diff
@@ -0,0 +1,974 @@
...
+  };
+
+  const workspaceId = 'test-workspace-id';
+  const streamChannelId = 'test-stream-channel-id';
+  const userWorkspaceId = 'test-user-workspace-id';
+  const roleId = 'test-role-id';
+  const objectMetadataId = 'test-object-metadata-id';
+  const fieldMetadataId = 'test-field-metadata-id';
+
+  const createMockObjectMetadata = (): FlatObjectMetadata =>
```
charlesBochet: for tests we should not create the mocks manually, this makes them hard to maintain when we do breaking change. Instead, we should derived mocks from autogenerated mocks against the server (kind of like a codegen).

==> use getMockObjectMetadataItemOrThrow & co tooling

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.service.ts
```diff
@@ -126,4 +215,201 @@ export class WorkspaceEventEmitterService {
...
+
+    return isRecordMatchingRLSRowLevelPermissionPredicate({
+      record,
+      filter: combinedFilter,
+      flatObjectMetadata: objectMetadata,
+      flatFieldMetadataMaps,
+    });
+  }
+
+  private async fetchPermissionsContext(workspaceId: string): Promise<{
```
charlesBochet, line 381: naming: fetch makes me thing of a db / remote call

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.service.ts
```diff
@@ -126,4 +215,201 @@ export class WorkspaceEventEmitterService {
...
+      filtersToApply.push(subscriberRLSFilter);
+    }
+
+    const combinedFilter = combineFilters(filtersToApply);
+
+    if (Object.keys(combinedFilter).length === 0) {
+      return true;
+    }
+
+    return isRecordMatchingRLSRowLevelPermissionPredicate({
```
charlesBochet, line 373: LGTM but I wonder if we shouldn't rather:
- combine the queryFilter with RLS filter and check if isRecordMatchingFilter instead

would be simpler and avoid separating RLS from filters as they are actually the same

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.service.ts
```diff
@@ -126,4 +215,201 @@ export class WorkspaceEventEmitterService {
...
+      ),
+    };
+
+    return {
+      ...event,
+      properties: filteredProperties,
+    } as ObjectRecordEvent & { objectNameSingular: string };
+  }
+
+  private matchQueriesWithEventAndRLS(
```
charlesBochet: naming not clear to me: matchXXXAndRLS ?

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.service.ts
```diff
@@ -126,4 +215,201 @@ export class WorkspaceEventEmitterService {
...
+      ...(filteredBefore !== undefined && { before: filteredBefore }),
+      ...(filteredAfter !== undefined && { after: filteredAfter }),
+      ...(filteredDiff !== undefined && { diff: filteredDiff }),
+      updatedFields: properties.updatedFields?.filter(
+        (field) => !restrictedFieldNames.has(field),
+      ),
+    };
+
+    return {
+      ...event,
```
charlesBochet, line 298: we should also filter all events that have no updates after having applied fieldPermissions

charlesBochet (commented):
Left comments :)

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.service.ts
```diff
@@ -89,23 +113,88 @@ export class WorkspaceEventEmitterService {
...
+      flatRowLevelPermissionPredicateGroupMaps: FlatRowLevelPermissionPredicateGroupMaps;
+      flatFieldMetadataMaps: FlatEntityMaps<FlatFieldMetadata>;
+      userWorkspaceRoleMap: Record<string, string>;
+      rolesPermissions: ObjectsPermissionsByRoleId;
+    },
   ): Promise<void> {
+    const { userWorkspaceId } = streamData.authContext;
+
+    if (!isDefined(userWorkspaceId)) {
+      this.logger.warn(
```
Weiko: I think we should avoid logs there, this will likely spam us without much value, right?

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.service.ts
Weiko: > fetch makes me thing of a db / remote call

I think that's what it's doing, right? It's calling redis here (or the local cache in some cases but that's still fine imho)

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.service.ts
Weiko: Good point 👍 I can work on that

Weiko (commented):
LGTM!

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.service.ts
thomtrp: this is already done right after calling that function
```
const filteredEvent = this.filterRestrictedFieldsFromEvent(
  eventWithObjectName,
  restrictedFields,
  permissionsContext.flatFieldMetadataMaps,
);

const filteredProperties = filteredEvent.properties as {
  updatedFields?: string[];
};

if (
  isDefined(filteredProperties.updatedFields) &&
  filteredProperties.updatedFields.length === 0
) {
  continue;
}
```

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.service.ts
thomtrp: renaming `getMatchingQueryIds` to simplify. Parameters already explain that it uses event and subscriber permissions

Weiko (approved):

PR #17203, opened 2026-01-16
Move tools/eslint-rules to packages/twenty-eslint-rules

FelixMalfait:
## Summary

Moves the custom ESLint rules from `tools/eslint-rules` to `packages/twenty-eslint-rules` for better organization within the monorepo packages structure.

## Changes

- Move `eslint-rules` from `tools/` to `packages/twenty-eslint-rules`
- Use `loadWorkspaceRules` from `@nx/eslint-plugin` to load custom rules
- Update all ESLint configs to use the `twenty/` rule prefix instead of `@nx/workspace-`
- Update `project.json`, `jest.config.mjs` with new paths
- Update `package.json` workspaces and `nx.json` cache inputs
- Update Dockerfile reference

## Technical Details

The custom ESLint rules are now loaded using Nx's `loadWorkspaceRules` utility which:
- Handles TypeScript transpilation automatically
- Allows loading workspace rules from any directory
- Provides a cleaner approach than the previous `@nx/workspace-` convention

## Testing

- Verified all 17 custom ESLint rules load correctly from the new location
- Verified linting works on dependent packages (twenty-front, twenty-server, etc.)


FelixMalfait:
Also the big change is that I had to upgrade NX

FelixMalfait:
Also removed I18FrontDecorator (now applied by default on all stories, less overhead for devs)

PR #17205, opened 2026-01-16
Implemented SSE update across main front components

lucasbordeau:
This PR implements SSE update events across the main components of the application : tables, boards, calendars, show pages.

There is still work to do on other event type and on making sure everything works fine, but this first implementation should be robust enough to start with.

Some problems encountered along the way : 
- Events are returning raw Postgres output, because they are not normalized by the GraphQL layer, so we ended up with amountMicros as string values, which the frontend does not like, so I implemented a small util in our `formatResult` generic pipeline to turn amountMicros to a number if it's a string value. We could implement other formatters for composite fields if we see problems with events.
-`action` property was missing in SSE events, which is required by the frontend to know what kind of event it is.

# QA

https://github.com/user-attachments/assets/393b45ac-59d2-48b0-855b-2ce9c4b8ae57

https://github.com/user-attachments/assets/cb214e7a-1595-4b85-bdb6-87630993d2e2


charlesBochet:
This looks awesome!

packages/twenty-front/src/modules/sse-db-event/hooks/useTriggerOptimisticEffectFromSseUpdateEvents.ts
```diff
@@ -0,0 +1,122 @@
...
+      objectRecordEvents: ObjectRecordEvent[];
+      objectMetadataItem: ObjectMetadataItem;
+    }) => {
+      const recordGqlFields = generateDepthRecordGqlFieldsFromObject({
+        objectMetadataItem,
+        objectMetadataItems,
+        depth: 1,
+      });
+
+      const updateEvents = objectRecordEvents.filter((objectRecordEvent) => {
```
charlesBochet, line 41: we should also treat creation and other cases in upcoming PRs

charlesBochet (approved):

mejih08-prog (approved):

Issue #16963, opened 2026-01-06
When an object is renamed, in relation picker, we still show old name


ijreilly:

AyeshaWaseemAhmad:
Hi @ijreilly, I would like to work on this issue.

ijreilly:
@AyeshaWaseemAhmad yes go ahead thank you! is the issue clear enough? i can provide more details

AyeshaWaseemAhmad:
@ijreilly It would be helpful if you could share more details. I tried reproducing this by updating the company name and then checking the people view, and the updated name was reflected correctly for me.

ijreilly:
<img width="910" height="356" alt="Image" src="https://github.com/user-attachments/assets/e169f657-1325-4eba-bf01-ce0f991d92b2" />

The issue is on morph relations. To reproduce you can create a relation between an object and two other objects (just tick more than one object when setting it). then it's in the relation picker that the issue occurs - see on my screenshot it says "Company" instead of "Listing". Maybe its because it is using the object's name instead of label.

araj00:
<img width="1398" height="589" alt="Image" src="https://github.com/user-attachments/assets/7b293b3f-cde5-406b-bc93-8015939841fe" />

@ijreilly I have one question, if we do change the standard object( not custom object) name then is it required to reflect that change in the app path or setting path. Because, right now, if we do change the standard object like person, company etc then in url, it stilll shows objectNamePlural. If it is necessary to reflect the changed name in path also then we need to change it from hardcoding those standard app path to dynamic handling in app and setting path. What is your thought on this?

PR #17209, opened 2026-01-16
fix: display current object name in morph relation picker after rename

AyeshaWaseemAhmad:
**Summary**

Issue #16963: When an object is renamed, the morph relation picker still shows the old name.

**Root cause**

The picker used searchRecord.objectNameSingular from the GraphQL search response, which can be stale after a rename.
The search record stores the object name at query time, not the current metadata.

**Solution**

- Updated SingleRecordPickerMenuItem:
- Added useObjectMetadataItems to access current object metadata.
- Look up the current object metadata by morphItem.objectMetadataId.
- Use labelSingular (or nameSingular as fallback) instead of searchRecord.objectNameSingular for display.
- Updated MultipleRecordPickerMenuItemContent:
- Use objectMetadataItem.labelSingular (already available as a prop) instead of searchRecord.objectNameSingular.


AyeshaWaseemAhmad:
@ijreilly Could you please review this to help fix the issue: https://github.com/twentyhq/twenty/issues/16963

packages/twenty-front/src/modules/settings/data-model/fields/forms/morph-relation/components/SettingsDataModelFieldRelationForm.tsx
```diff
@@ -62,17 +62,27 @@ export type SettingsDataModelFieldMorphRelationFormValues = z.infer<
...
   existingFieldMetadataId,
+  sourceObjectMetadataId,
   disabled = false,
 }: SettingsDataModelFieldRelationFormProps) => {
   const { t } = useLingui();
-  const { control } =
-    useFormContext<SettingsDataModelFieldMorphRelationFormValues>();
+  const { control, watch } = useFormContext();
+
+  const currentIds = watch('morphRelationObjectMetadataIds') as
```
ijreilly, line 78: didnt manage to use react hook form errors

ijreilly:
thanks @AyeshaWaseemAhmad for your contribution !

packages/twenty-server/src/engine/core-modules/search/dtos/search-record.dto.ts
```diff
@@ -16,6 +16,11 @@ export class SearchRecordDTO {
   @IsNotEmpty()
   objectNameSingular: string;
 
+  @Field(() => String)
```
ijreilly, line 19: @AyeshaWaseemAhmad I [removed](https://github.com/twentyhq/twenty/pull/17209/commits/d4da81b63204a08ab935748d1f636da8b82baece) your implementation on SingleRecordPickerMenuItem to avoid adding a dependency to objectMetadataItems state
Instead I think we should get the information from the search request.
To avoid breaking changes between FE and BE (that are not deployed in an exact simultaneity), I am adding the BE in this PR, and once we will have deployed it I will deploy the FE part (that adds objectLabelSingular to the query and uses it)

ijreilly:
Fixes https://github.com/twentyhq/core-team-issues/issues/2049

ijreilly (approved):
Thank you, i also added related fixes in your pr (not described in the original issue)

AyeshaWaseemAhmad:
@ijreilly Just to confirm, you’ve fixed two issues in this PR. I noticed you referred to another issue [here](https://github.com/twentyhq/twenty/pull/17209#issuecomment-3773024377), whereas I had worked on this [issue](https://github.com/twentyhq/twenty/issues/16963).

ijreilly:
@AyeshaWaseemAhmad Yes I know, and your fix works, thank you ! i just took the advantage of working on this pr (as i slightly updated your implementation as i explained in the comment on the file diff of this PR) to fix something else that is on the same surface

PR #17211, opened 2026-01-17
feat: add TypeScript Go (tsgo) for faster type checking

FelixMalfait:
## Summary

- Add `@typescript/native-preview` (tsgo) for dramatically faster type checking on frontend projects
- Configure tsgo as default for frontend projects (twenty-front, twenty-ui, twenty-shared, etc.)
- Keep tsc for twenty-server (faster for NestJS decorator-heavy code)
- Fix type imports for tsgo compatibility (DOMPurify, AxiosInstance)
- Remove deprecated `baseUrl` from tsconfigs where safe

## Performance Results

| Project | tsgo | tsc -b | Speedup |
|---------|------|--------|---------|
| **twenty-front** | 1.4s | 60.7s | **43x faster** |
| **twenty-server** | 2m42s | 1m10s | tsc is faster (decorators) |

tsgo excels at modern React/JSX codebases but struggles with decorator-heavy NestJS backends, so we use the optimal checker for each.

## Usage

```bash
# Default (tsgo for frontend, tsc for backend)
nx typecheck twenty-front
nx typecheck twenty-server

# Force tsc fallback if needed
nx typecheck twenty-front --configuration=tsc

# Force tsgo on backend (slower, not recommended)
nx typecheck twenty-server --configuration=tsgo
```

## Test plan

- [x] `nx typecheck twenty-front` passes with tsgo
- [x] `nx typecheck twenty-server` passes with tsc
- [x] `nx run-many -t typecheck --exclude=fireflies` passes
- [ ] CI tests pass


FelixMalfait:
@charlesBochet fyi. Upgrading on twenty-front since MS says it's okay tu use, but didn't manage to upgrade twenty-server, I think they need to make progress on decorator support

prastoin:
Nice very cool initiative
Looks weird for twenty-server, tsgo is retro-comp with all dependencies natively ?

packages/twenty-front/tsconfig.build.json
```diff
@@ -1,8 +1,7 @@
 {
   "extends": "./tsconfig.json",
   "compilerOptions": {
-    "types": ["node"],
-    "baseUrl": "."
```
prastoin, line 5: ***Remark:*** wondering why this why required, same for above tsconfig

prastoin: Ok nevermind it has been deprecated

packages/twenty-server/project.json
```diff
@@ -41,7 +41,20 @@
         "updateBuildableProjectDepsInPackageJson": true
       }
     },
-    "typecheck": {},
+    "typecheck": {
```
prastoin, line 44: ***Question:*** Can't we use the global nx configuration typecheck `tsc` configuration instead of redeclaring here ?

FelixMalfait: Maybe but I think this is pretty straightforward to understand/read, and hopefully this overwrite is only temporary! I think they have some issues with decorators and NestJS uses a lot of decorator, it should get better within a few months

FelixMalfait:
Thanks for the review @prastoin - I'd love to migrate twenty-server to, the impact is huge when it works, but it doesn't seem as straightforward as it was for frontend, I think it's best to wait.

packages/twenty-server/project.json
prastoin: I would recommend removing the tsc configuration at the top nx configuration then, and only keep the tsc at the twenty-server lvl
That's nitpicking though

prastoin (approved):

prastoin:
Removing `baseUrl` would break the `tsc` compilation ?

charlesBochet:
Awesome! Great that this TS built with TS non-sense is coming to an end :D
Tested locally for frontend

charlesBochet:
I have the impression that it's also way less RAM intensive (~800MB vs ~3GB before), hard to confirm while the whole project is not fully migrated. This could also be game changer for linter ts rules (this is what's slowing down the linter mostly right now)

PR #17214, opened 2026-01-17
fix: refresh virtualized table when field metadata is updated ( #16388 )

carbonFibreCode:
Fixes #16388 

Now we have used the metadata fetching from network using resetVirtualizationBecauseDataChanged(), as we navigate back to the table after updating the field content. As the virtualised table uses the "cache-first" strategy, it fails to give fresh data from the data base

Please let me know i we need to add the loading state (and it's UI requirements) while it fetches the cell content, as currently the cell is empty until it's populated by fresh data

https://github.com/user-attachments/assets/901085ba-4337-4a5d-86c8-15ac84250e8f


packages/twenty-front/src/modules/object-record/record-table/virtualization/components/RecordTableVirtualizedInitialDataLoadEffect.tsx
```diff
@@ -103,5 +108,45 @@ export const RecordTableVirtualizedInitialDataLoadEffect = () => {
...
+    setLastProcessedFieldMetadataUpdate,
+  ] = useRecoilComponentState(lastProcessedFieldMetadataUpdateComponentState);
+
+  useEffect(() => {
+    if (!lastFieldMetadataItemUpdate) {
+      return;
+    }
+
+    if (
+      lastFieldMetadataItemUpdate.timestamp === lastProcessedFieldMetadataUpdate
```
etiennejouan: `lastFieldMetadataItemUpdate.timestamp < lastProcessedFieldMetadataUpdate` Shouldn't be more logic like that ?

etiennejouan (commented):
Hi @carbonFibreCode, thank you so much for your contribution. I've tested it and it works well.

Concerning [the implementation strategy](https://github.com/twentyhq/twenty/issues/16388#issuecomment-3664752756), we plan to have an app-scoped state holding all sorts of metadata updates—a higher level one than the one you built.

Another issue I see: when first loading another table view based on another view showing the field, it would go through the useEffect (RecordTableVirtualizedInitialDataLoadEffect.tsx file), resetting virtualization when we don't need it to.

@lucasbordeau, what do you think? Would you have more precise guidance to help fix this? Or too complicated to fix without context/experience on twenty codebase ?

carbonFibreCode:
> Hi @carbonFibreCode, thank you so much for your contribution. I've tested it and it works well.
> 
> Concerning [the implementation strategy](https://github.com/twentyhq/twenty/issues/16388#issuecomment-3664752756), we plan to have an app-scoped state holding all sorts of metadata updates—a higher level one than the one you built.
> 
> Another issue I see: when first loading another table view based on another view showing the field, it would go through the useEffect (RecordTableVirtualizedInitialDataLoadEffect.tsx file), resetting virtualization when we don't need it to.
> 
> @lucasbordeau, what do you think? Would you have more precise guidance to help fix this? Or too complicated to fix without context/experience on twenty codebase ?

hey @etiennejouan  thanks for review,  i am working on it right now to follow the implementation strategy for using appScoped state, open to @lucasbordeau 's comment
I'd prefer the guidance to get familiar with the codebase.

etiennejouan:
> > Hi @carbonFibreCode, thank you so much for your contribution. I've tested it and it works well.
> > Concerning [the implementation strategy](https://github.com/twentyhq/twenty/issues/16388#issuecomment-3664752756), we plan to have an app-scoped state holding all sorts of metadata updates—a higher level one than the one you built.
> > Another issue I see: when first loading another table view based on another view showing the field, it would go through the useEffect (RecordTableVirtualizedInitialDataLoadEffect.tsx file), resetting virtualization when we don't need it to.
> > @lucasbordeau, what do you think? Would you have more precise guidance to help fix this? Or too complicated to fix without context/experience on twenty codebase ?
> 
> hey @etiennejouan thanks for review, i am working on it right now to follow the implementation strategy for using appScoped state, open to @lucasbordeau 's comment I'd prefer the guidance to get familiar with the codebase.

@carbonFibreCode 
To make sure you don't work for nothing, I would wait wait for @lucasbordeau  to get back.

carbonFibreCode:
> @carbonFibreCode To make sure you don't work for nothing, I would wait wait for @lucasbordeau to get back.

waiting for @lucasbordeau 's review,

In the mean time i tried switching to using a generic lastAppOperationState as requested, for the table refresh logic, I opted for a pure event driven approach, if a relevant event hasn't been processed by the component, we trigger a refresh.

packages/twenty-front/src/modules/object-record/record-table/virtualization/components/RecordTableVirtualizedInitialDataLoadEffect.tsx
```diff
@@ -103,5 +108,45 @@ export const RecordTableVirtualizedInitialDataLoadEffect = () => {
     visibleRecordFields,
   ]);
 
+  const lastFieldMetadataItemUpdate = useRecoilValue(
```
lucasbordeau: The logic works for me, could you just extract this into a separate effect component ? We try to keep one useEffect per component to avoid side effects that can happen when multiple useEffects share the same reactive dependencies.

carbonFibreCode: Yeah sure, i'll do it and push the changes by 9:30 PM IST  (4:30 PM CET) -> i hope correct about the time zone.. haha

carbonFibreCode: Hi @lucasbordeau, I’ve updated the PR with the separate effect component as you suggested and all checks are passing. Would you mind having another look and sharing any further guidance, given the current blocked: tech spec needed label?

packages/twenty-front/src/modules/object-metadata/states/lastFieldMetadataItemUpdateState.ts
```diff
@@ -0,0 +1,13 @@
+import { createState } from 'twenty-ui/utilities';
+
+export type LastFieldMetadataItemUpdate = {
```
lucasbordeau: Ok I think it's a good pattern : 
- The table updates only when it mounts back after settings change, so that's asynchronous as it should be
- We reset virtualization as we already do for other update events, and it is designed to handle this frequently right now. (We would like to optimize those small state effects but that requires unlocking other improvements before.)
- We already use this pattern elsewhere

Two changes though : 
- Could you use an uuid v4 instead of a timestamp for the referential integrity of this object ?
- Extract this type in a standalone file

packages/twenty-front/src/modules/object-record/record-table/virtualization/states/lastProcessedFieldMetadataUpdateComponentState.ts
```diff
@@ -0,0 +1,9 @@
+import { ContextStoreComponentInstanceContext } from '@/context-store/states/contexts/ContextStoreComponentInstanceContext';
+import { createComponentState } from '@/ui/utilities/state/component-state/utils/createComponentState';
+
+export const lastProcessedFieldMetadataUpdateComponentState =
```
lucasbordeau: lastProcessedFieldMetadataUpdateIdComponentState => use an id property instead

packages/twenty-front/src/modules/object-record/record-table/virtualization/components/RecordTableVirtualizedFieldMetadataUpdateEffect.tsx
```diff
@@ -0,0 +1,62 @@
...
+    setLastProcessedFieldMetadataUpdate,
+  ] = useRecoilComponentState(lastProcessedFieldMetadataUpdateComponentState);
+
+  useEffect(() => {
+    if (!lastFieldMetadataItemUpdate) {
+      return;
+    }
+
+    if (
+      lastFieldMetadataItemUpdate.timestamp === lastProcessedFieldMetadataUpdate
```
lucasbordeau: Use id here for comparison

carbonFibreCode: hmmmm.... ok

carbonFibreCode: @lucasbordeau pushed the changes

- now eventId is compared as per the request
- pulled out the types in separate file as per existing design pattern

lucasbordeau (approved):
LGTM, nice contribution.

I've just added some minor renaming to make it more aligned with our conventions.

carbonFibreCode:
> LGTM, nice contribution.
> 
> I've just added some minor renaming to make it more aligned with our conventions.

Thanks for the feedback, Lucas, i've noted those naming conventions and will make sure to apply them to my future contributions. Appreciate the guidance on this one.

carbonFibreCode:
> lucasbordeau

@lucasbordeau
Hey, some CI checks are failing blocking auto merge, let me know if it's the pipeline/config issue or something's on me to fix

PR #17233, opened 2026-01-19
fix: E2E login test flakiness

FelixMalfait:
## Summary
- Fix E2E login test flakiness by using `click()` auto-waiting instead of `isVisible()` check
- The login form shows a loader while GraphQL data loads. The previous `isVisible()` check returned immediately (no waiting) and would fail while the loader was showing
- Using `click()` which has built-in auto-waiting for elements to be visible and actionable fixes this

## Test plan
- E2E tests should pass more reliably in CI
- Login setup test should no longer timeout waiting for the email field


FelixMalfait:
Not 100% sure it was the real issue but I think this could fix the flakiness I've seen in merge queue?

bosiraphael (approved):

PR #17240, opened 2026-01-19
Support referencing updatedFields array in databaseEvent triggers 

### LINKED ISSUE #1838: [Extensibility] [v1] Support referencing updatedFields array in databaseEvent triggers
In application function defineFunction method, we want to be able to listen to specific fields update, not necessary all updates, of database events.

Update `packages/twenty-shared/src/application/serverlessFunctionManifestType.ts` `DatabaseEventTrigger` so it have an optional `updatedFields?: string[]` parameter. This parameter should be taken into account in twenty-server syncApplication endpoint

ijreilly:
Closes https://github.com/twentyhq/core-team-issues/issues/1838

DatabaseEventTriggers can now define a field-level granularity on which updates they should react to.
After a discussion with the team, we have decided to rely on field names and not UID, just as we do for objects. The rationale is that we want to keep a pleasant devX and consider it's not on twenty to ensure continuity of api usage around an object if their name is updated: for instance if a user has based their webhook on the name of an object, if they decide to update it, they have to take care of updating their webhook.

### DIFF
--- a/packages/twenty-server/src/engine/metadata-modules/database-event-trigger/utils/transform-event-batch-to-event-payloads.ts
+++ b/packages/twenty-server/src/engine/metadata-modules/database-event-trigger/utils/transform-event-batch-to-event-payloads.ts
@@ -1,11 +1,13 @@
+import { isDefined } from 'twenty-shared/utils';
+
 import type {
   DatabaseEventPayload,
   ObjectRecordEvent,
 } from 'twenty-shared/database-events';
 
-import type { WorkspaceEventBatch } from 'src/engine/workspace-event-emitter/types/workspace-event-batch.type';
 import { type DatabaseEventTriggerEntity } from 'src/engine/metadata-modules/database-event-trigger/entities/database-event-trigger.entity';
 import { type ServerlessFunctionTriggerJobData } from 'src/engine/metadata-modules/serverless-function/jobs/serverless-function-trigger.job';
+import type { WorkspaceEventBatch } from 'src/engine/workspace-event-emitter/types/workspace-event-batch.type';
 
 export const transformEventBatchToEventPayloads = ({
   workspaceEventBatch,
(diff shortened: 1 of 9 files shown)


packages/twenty-server/src/engine/metadata-modules/database-event-trigger/utils/transform-event-batch-to-event-payloads.ts
```diff
@@ -32,3 +42,35 @@ export const transformEventBatchToEventPayloads = ({
 
   return result;
 };
+
+const filterEventsByUpdatedFields = ({
```
ijreilly, line 46: I implemented this logic here instead of CallDatabaseEventTriggerJobsJob's shouldTriggerJob because shouldTriggerJob is about the batch of events, while we don't want to trigger the serverless for an event that is not concerned but that shares the batch with one that is.

martmull, line 46: looks good but this needs a little unit test to test (and document the behavior)

martmull (commented):
lgtm but small unit test required I think

ijreilly:
@martmull ready for review again, thank you !

martmull (approved):
lgtm

prastoin (commented):
Hey there, shouldn't we implement a side effect on an field name update that will search for any related database event trigger settings to update too ?
Is it overkill or are we going to handle that later with many to many relations ?
Unless I'm mistaken right now if a user creates a database event trigger from the api and renames the related custom field name in the UI it will introduce desync both

Just making sure we're okay with that !
Are we expecting to make the settings `updatedFields` required at some point and backfill existing occurences ? Maybe we should require the `updatedFields` in the input settings from now on already ?

We should also validate that the fields provided in the settings names are foundable in the workspace

Sorry hopping on after this has been already merged !

ijreilly:
Yes that was the exact point of the conversation we followed this morning, should we use names or UID - UID being more robust but making the dev experience less readable. 
We chose names and stated that it's not our job to fix the users' integration if they choose to change a name

updatedFields should not be required and should not be, it's an additional option we give to users, but it's also possible to listen to any update.

prastoin:
> We chose names and stated that it's not our job to fix the users' integration if they choose to change a name

Even through the api metadata ? We've been handling those side effect for them so far

ijreilly:
Yes, even through api metadata

PR #17241, opened 2026-01-19
SSE - Protect query api

thomtrp:
- Ensure user can only add/remove queries of its own stream
- Prevent stream from being highjacked

### DIFF
--- a/packages/twenty-server/src/engine/subscriptions/event-stream.service.ts
+++ b/packages/twenty-server/src/engine/subscriptions/event-stream.service.ts
@@ -10,6 +10,10 @@ import { InjectCacheStorage } from 'src/engine/core-modules/cache-storage/decora
 import { CacheStorageService } from 'src/engine/core-modules/cache-storage/services/cache-storage.service';
 import { CacheStorageNamespace } from 'src/engine/core-modules/cache-storage/types/cache-storage-namespace.enum';
 import { EVENT_STREAM_TTL_MS } from 'src/engine/subscriptions/constants/event-stream-ttl.constant';
+import {
+  EventStreamException,
+  EventStreamExceptionCode,
+} from 'src/engine/subscriptions/event-stream.exception';
 import { type EventStreamData } from 'src/engine/subscriptions/types/event-stream-data.type';
 import { type ObjectRecordSubscriptionEvent } from 'src/engine/subscriptions/types/object-record-subscription-event.type';
 
@@ -31,6 +35,16 @@ export class EventStreamService {
     authContext: SerializableAuthContext;
   }): Promise<void> {
     const key = this.getEventStreamKey(workspaceId, eventStreamChannelId);
+
+    const existing = await this.cacheStorageService.get<EventStreamData>(key);
+
+    if (isDefined(existing)) {
+      throw new EventStreamException(
+        'Event stream already exists',
+        EventStreamExceptionCode.EVENT_STREAM_ALREADY_EXISTS,
+      );
+    }
+
     const streamData: EventStreamData = {
       authContext,
       workspaceId,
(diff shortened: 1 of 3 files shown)


packages/twenty-server/src/engine/subscriptions/event-stream.service.ts
```diff
@@ -31,6 +35,16 @@ export class EventStreamService {
     authContext: SerializableAuthContext;
   }): Promise<void> {
     const key = this.getEventStreamKey(workspaceId, eventStreamChannelId);
+
+    const existing = await this.cacheStorageService.get<EventStreamData>(key);
+
```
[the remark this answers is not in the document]
thomtrp: nope. Eventstream should be unique so no need for lock

packages/twenty-server/src/engine/subscriptions/event-stream.service.ts
```diff
@@ -126,6 +140,37 @@ export class EventStreamService {
...
+    );
+
+    if (!isDefined(streamData)) {
+      return false;
+    }
+
+    if (isDefined(authContext.userWorkspaceId)) {
+      return (
+        streamData.authContext.userWorkspaceId === authContext.userWorkspaceId
+      );
```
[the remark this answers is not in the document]
thomtrp: user workspace is enough

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.resolver.ts
```diff
@@ -148,9 +170,26 @@ export class WorkspaceEventEmitterResolver {
...
+      workspaceId: workspace.id,
+      eventStreamChannelId,
+      authContext: {
+        userWorkspaceId,
+        apiKeyId: apiKey?.id,
+      },
+    });
+
+    if (!isAuthorized) {
+      throw new ForbiddenException(
```
prastoin: ***Question:*** Will this get translated ?

prastoin (approved):
Hey there ! Looks good 
Would it possible to setup an integration testing env for such stream interactions ? + it's hard to test without such test env
This is critical and should covered
Maybe not in this PR though
Still approving !

packages/twenty-server/src/engine/subscriptions/event-stream.service.ts
```diff
@@ -31,6 +35,16 @@ export class EventStreamService {
     authContext: SerializableAuthContext;
   }): Promise<void> {
     const key = this.getEventStreamKey(workspaceId, eventStreamChannelId);
+
+    const existing = await this.cacheStorageService.get<EventStreamData>(key);
+
+    if (isDefined(existing)) {
+      throw new EventStreamException(
+        'Event stream already exists',
```
prastoin, line 43: ***Remark:*** I'm not even sure we might wanna return this much information, maybe only returning an auth exception would be enough ? What do you think ?

thomtrp: will be better for debugging. Not a lot of chance the error is thrown unless we do not implement it correctly

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.resolver.ts
thomtrp: good point. Using EventStreamException instead

packages/twenty-server/src/engine/subscriptions/event-stream.service.ts
prastoin: Ok sounds legit !
I meant from an attacker point of view

Weiko (approved):
LGTM!

PR #17242, opened 2026-01-19
BILLING - Send dis-sync error to sentry

etiennejouan:
Investigate Stripe activity when workspace not found, creating [alert](https://twenty-v7.sentry.io/issues/alerts/rules/twenty-server/16611050/details/) in Sentry

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-subscription.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-subscription.service.ts
@@ -6,7 +6,6 @@ import { InjectRepository } from '@nestjs/typeorm';
 import { isDefined } from 'twenty-shared/utils';
 import { WorkspaceActivationStatus } from 'twenty-shared/workspace';
 import { In, Repository } from 'typeorm';
-
 import { msg } from '@lingui/core/macro';
 
 import type Stripe from 'stripe';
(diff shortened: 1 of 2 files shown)


packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-subscription.service.ts
```diff
@@ -75,14 +76,26 @@ export class BillingWebhookSubscriptionService {
...
     }
 
     if (
       isDefined(workspace.deletedAt) &&
       type !== BillingWebhookEvent.CUSTOMER_SUBSCRIPTION_DELETED
     ) {
-      return { noWorkspace: true };
+      throw new BillingException(
+        `Workspace not found for subscription event ${event.id} / workspaceId: ${workspaceId}`,
+        BillingExceptionCode.BILLING_SUBSCRIPTION_EVENT_WORKSPACE_NOT_FOUND,
```
prastoin, line 94: ***Question:*** Should this be the exact same exception ? Might wanna have the grain over exception from soft deleted workspace versus destroyed ones ?

prastoin (approved):
Looks good to me ! What was the purpose of the previously return record ?
```ts
{ noWorkspace: true };
```
The method seems to return diff return type

packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-subscription.service.ts
etiennejouan: I don't know what was the purpose. I think when you're confident stripe sub events are no longer useful because workspace is deleted, you respond a `{ noWorkspace: true };` response.

Issue is that we may have other cases where sub is still alive but workspace deleted. This PR is temporary for debugging/investigation purpose.

PR #17243, opened 2026-01-19
fix rls create new record with composite

Weiko:
## Context
buildValueFromFilter was not handling composite filters which was needed for RLS. This PR implements that and fix some issues with RLS

Tested with a few composite + relation fields
Screenshot 2026-01-19 at 15 38 42

### DIFF
--- a/packages/twenty-front/src/modules/object-record/hooks/useBuildRecordInputFromRLSPredicates.ts
+++ b/packages/twenty-front/src/modules/object-record/hooks/useBuildRecordInputFromRLSPredicates.ts
@@ -14,7 +14,15 @@ import { type ObjectRecord } from '@/object-record/types/ObjectRecord';
 import { isUndefined } from '@sniptt/guards';
 import { useRecoilValue } from 'recoil';
 import { RelationType } from 'twenty-shared/types';
-import { isDefined } from 'twenty-shared/utils';
+import { isDefined, isPlainObject } from 'twenty-shared/utils';
+
+const mergeCompositeValues = (
+  existingValue: unknown,
+  incomingValue: unknown,
+) =>
+  isPlainObject(existingValue) && isPlainObject(incomingValue)
+    ? { ...existingValue, ...incomingValue }
+    : incomingValue;
 
 export const useBuildRecordInputFromRLSPredicates = ({
   objectMetadataItem,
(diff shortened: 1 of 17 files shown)


packages/twenty-front/src/modules/object-record/hooks/useBuildRecordInputFromRLSPredicates.ts
```diff
@@ -67,35 +122,37 @@ export const useBuildRecordInputFromRLSPredicates = ({
...
+            throw new Error(
+              `Subfield name not found for composite field: ${fieldMetadataItem.name}`,
+            );
+          }
+
+          const compositeValue = buildCompositeValueFromSubField({
+            compositeFieldType: fieldMetadataItem.type,
+            subFieldName: filter.subFieldName,
+            value: currentWorkspaceMemberFieldValue,
+          });
```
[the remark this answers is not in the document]
Weiko: This is done in buildValueFromFilter

packages/twenty-front/src/modules/object-record/hooks/useBuildRecordInputFromRLSPredicates.ts
```diff
@@ -37,6 +39,59 @@ export const useBuildRecordInputFromRLSPredicates = ({
...
+    name: string;
+    type: string;
+    settings?: { relationType?: RelationType };
+  }) =>
+    fieldMetadataItem.type === 'RELATION' &&
+    fieldMetadataItem.settings?.relationType === RelationType.MANY_TO_ONE
+      ? `${fieldMetadataItem.name}Id`
+      : fieldMetadataItem.name;
+
+  const getWorkspaceMemberFieldValue = ({
```
thomtrp, line 60: not sure why we need to use the workspace member prefix everywhere. Looks like working for any field and subfield

thomtrp (approved):

packages/twenty-front/src/modules/object-record/hooks/useBuildRecordInputFromRLSPredicates.ts
Weiko: It's true that this could be a util that would work with any object but in the case of RLS we are always using it with workspace member. I'll rename it anyway 👍

PR #17246, opened 2026-01-19
Watch command skeleton

charlesBochet:
Summary

Rewrites the app:dev command to use Vite's dev server instead of manual file watching with chokidar. This provides better file watching capabilities and aligns with the existing build tooling.

image


packages/twenty-sdk/src/cli/build/index.ts
martmull, line 1: can you remove that?

packages/twenty-sdk/src/cli/build/build.service.ts
martmull, line 1: can you move this service to src/cli/utilities/build/service?

martmull (approved):

packages/twenty-sdk/src/cli/build/build.service.ts
charlesBochet: yes, I will actually remove all this when I'm done, I'm extracting piece of code of this for now

PR #17252, opened 2026-01-19
Twenty SDK watch build of functions

charlesBochet:
## Summary

This PR adds function build support to the `twenty dev` command, enabling tree-shaken production builds of serverless functions during development with Vite's incremental build capabilities.

### New Features
- **Function building in dev mode**: Serverless functions are now compiled to tree-shaken bundles in `.twenty/functions/` during development
- **Automatic restart on entry point changes**: When functions are added or removed, the watcher automatically restarts with the new configuration
- **Function entry point logging**: Dev mode now displays detected function entry points with their names and paths


packages/twenty-sdk/src/cli/utilities/vite/__tests__/function-builder.spec.ts
```diff
@@ -0,0 +1,66 @@
+import { computeFunctionOutputPath } from '../function-paths';
+
+describe('computeFunctionOutputPath', () => {
```
Weiko: Copy pasting issue? Both function-builder and function-path are testing computeFunctionOutputPath (and have the same tests)

packages/twenty-sdk/src/cli/commands/app/app-dev.ts
```diff
@@ -1,136 +1,140 @@
...
+      return;
     }
-  }
 
-  private async writeManifestToOutput(
-    result: BuildManifestResult,
-  ): Promise<void> {
-    try {
-      const outputDir = path.join(this.appPath, OUTPUT_DIR);
+    this.isRestarting = true;
```
Weiko, line 108: nit: potential race condition where isRestarting is set too late (while multiple quick scheduleRestart won't early return as expected). Not that important though but FYI.

charlesBochet: ty!

Weiko (approved):

PR #17259, opened 2026-01-19
Add Front components to SDK

charlesBochet:
## Add `frontComponent` entity type to SDK

This PR adds a new `frontComponent` entity type at parity with `serverlessFunction`. Front components are React components that can be defined and bundled as part of Twenty applications.

#### New Entity Type
- Added `FRONT_COMPONENT` to `SyncableEntity` enum
- Front components use `*.front-component.tsx` file naming convention
- CLI command `twenty app:add` now includes `front-component` as an option

#### SDK Application Layer (`twenty-sdk`)
- Added `defineFrontComponent()` function for defining front component configurations with validation
- Added `FrontComponentConfig` type for component configuration
- Added `getFrontComponentBaseFile()` template generator for scaffolding new front components
- Added `loadFrontComponentModule()` to load front component modules and extract component metadata

#### Shared Types (`twenty-shared`)
- Added `FrontComponentManifest` type with `universalIdentifier`, `name`, `description`, `componentPath`, and `componentName` fields
- Updated `ApplicationManifest` to include optional `frontComponents` array

#### Manifest Build System
- Updated `manifest-build.ts` to discover and load `*.front-component.tsx` files
- Updated `manifest-validate.ts` to validate front components and check for duplicate IDs
- Updated `manifest-display.ts` to display front component count in build summary
- Updated `manifest-plugin.ts` to display front component entry points and watch `.tsx` files

#### Template (`create-twenty-app`)
- New applications now include a sample `hello-world.front-component.tsx` file

#### Tests
- Added unit tests for `defineFrontComponent()`
- Added unit tests for `getFrontComponentBaseFile()`


packages/twenty-sdk/src/cli/utilities/file/utils/file-config-loader.ts
```diff
@@ -48,19 +49,31 @@ const getTsconfigAliases = async (
   }
 };
 
+/**
+ * Transform JSX/TSX files using esbuild before loading.
+ */
+const transformJsxFile = async (filepath: string): Promise<string> => {
+  // Use esbuild for fast JSX transformation
+  const esbuild = await import('esbuild');
```
martmull: why using esbuild for front components instead of vite?

packages/twenty-sdk/src/cli/utilities/file/utils/file-config-loader.ts
```diff
@@ -48,19 +49,31 @@ const getTsconfigAliases = async (
...
+  const source = await fs.readFile(filepath, 'utf8');
+
+  const result = await esbuild.transform(source, {
+    loader: 'tsx',
+    jsx: 'automatic',
+    format: 'esm',
```
martmull: you need to pass the tsconfig so it resolves aliases

packages/twenty-sdk/src/cli/utilities/file/utils/file-config-loader.ts
```diff
@@ -303,3 +316,100 @@ export const loadFunctionModule = async (
...
+      );
+    };
+
+    const config = findConfigExport<{ component: Function }>(mod, hasComponent);
+
+    if (!config) {
+      throw new Error(
+        `Front component file ${filepath} must export a config object with a "component" property`,
+      );
+    }
```
martmull: that check should be in defineFrontComponent

packages/twenty-sdk/package.json
```diff
@@ -38,6 +38,7 @@
     "chokidar": "^4.0.0",
     "commander": "^12.0.0",
     "dotenv": "^16.4.0",
+    "esbuild": "^0.24.0",
```
martmull: why vite cannot do it?

packages/twenty-sdk/src/cli/utilities/file/utils/file-config-loader.ts
charlesBochet: good point, let's try with vite. Based on what we discuss, this code will likely disppear, this is to extract defineFrontComponent out of jsx file

packages/twenty-sdk/src/cli/utilities/file/utils/file-config-loader.ts
charlesBochet: I'm not sure it's needed this is to extract defineFrontComponent not to parse the actual code

charlesBochet: actually we don't need esbuild and I can use jiti here too

packages/twenty-sdk/src/cli/utilities/file/utils/file-config-loader.ts
charlesBochet: can actually be done using jiti

martmull (approved):

PR #17264, opened 2026-01-20
fix: implement polymorphic relation detach when selecting 'No {Field}'

### LINKED ISSUE #17253: Fix: "No {Field}" option in polymorphic relation picker does nothing
## Bug Description

When clicking the "No {Field}" option (e.g., "No Polymorphic Owner") in a polymorphic relation picker, nothing happens. The relation is not cleared/detached.

## Root Cause

The detach logic was never implemented in the handlers:

### 1. `RecordDetailMorphRelationSectionDropdownManyToOne.tsx` (lines 89-102)

```typescript
const handleRelationPickerEntitySelected = (
  selectedMorphItem?: RecordPickerPickableMorphItem,
) => {
  closeDropdown(dropdownId);

  if (!selectedMorphItem?.recordId) return;  // ← Returns early without clearing the relation

  onSubmit?.({
    newValue: {
      id: selectedMorphItem.recordId,
      objectMetadataId: selectedMorphItem.objectMetadataId,
    },
  });
};
```

### 2. `MorphRelationManyToOneFieldInput.tsx` (lines 35-41)

```typescript
const handleMorphItemSelected = async (
  selectedMorphItem: RecordPickerPickableMorphItem | null | undefined,
) => {
  if (!isDefined(selectedMorphItem)) {
    // Handle detach
    return;  // ← Just returns - the "detach" logic was never implemented!
  }
  // ...
};
```

## Expected Behavior

Clicking "No {Field}" should clear/detach the polymorphic relation by setting the foreign key fields to `null`.

## Files to Update

- `packages/twenty-front/src/modules/object-record/record-field-list/record-detail-section/relation/components/RecordDetailMorphRelationSectionDropdownManyToOne.tsx`
- `packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/input/components/MorphRelationManyToOneFieldInput.tsx`

## Implementation Notes

When `selectedMorphItem` is undefined, the handlers should:
1. Call `onSubmit` with a null value, OR
2. Call a mutation to clear the polymorphic relation fields (the `{relation}Id` and `{relation}ObjectMetadataId` columns)

Reference similar patterns in regular relation pickers for how detaching is handled.

bitloi:
Fixes #17253

When clicking 'No {Field}' in a polymorphic relation picker, the relation was not being cleared. This commit implements the missing detach logic:

- RecordDetailMorphRelationSectionDropdownManyToOne: Call onSubmit with newValue: null when no item is selected
- MorphRelationManyToOneFieldInput: Call persistMorphManyToOne with valueToPersist: null for detach case
- useMorphPersistManyToOne: Implement actual detach logic that sets all morph relation ID fields to null via updateOneRecord

Also adds unit tests for the useMorphPersistManyToOne hook.

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/input/hooks/useMorphPersistManyToOne.ts
+++ b/packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/input/hooks/useMorphPersistManyToOne.ts
@@ -35,13 +35,43 @@ export const useMorphPersistManyToOne = ({
         recordId: string;
         fieldDefinition: FieldDefinition<FieldMetadata>;
         valueToPersist: string | null | undefined;
-        targetObjectMetadataNameSingular: string;
+        targetObjectMetadataNameSingular?: string;
       }) => {
         assertFieldMetadata(
           FieldMetadataType.MORPH_RELATION,
           isFieldMorphRelation,
           fieldDefinition,
         );
+
+        const fieldName = fieldDefinition.metadata.fieldName;
+
+        if (!isDefined(valueToPersist)) {
+          const allNullRecordInput: Record<string, null> =
+            fieldDefinition.metadata.morphRelations.reduce(
+              (acc, morphRelation) => {
+                const computedFieldName = computeMorphRelationFieldName({
+                  fieldName,
+                  relationType: fieldDefinition.metadata.relationType,
+                  targetObjectMetadataNameSingular:
+                    morphRelation.targetObjectMetadata.nameSingular,
+                  targetObjectMetadataNamePlural:
+                    morphRelation.targetObjectMetadata.namePlural,
+                });
+                acc[`${computedFieldName}Id`] = null;
+                return acc;
+              },
+              {} as Record<string, null>,
+            );
+
+          updateOneRecord?.({
+            objectNameSingular: objectMetadataNameSingular,
+            idToUpdate: recordId,
+            updateOneRecordInput: allNullRecordInput,
+          });
+
+          return;
+        }
+
         const targetObjectMetadataItem = objectMetadataItems.find(
           (objectMetadataItem) =>
             objectMetadataItem.nameSingular ===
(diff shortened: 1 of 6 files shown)


packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/input/hooks/useMorphPersistManyToOne.ts
```diff
@@ -35,13 +35,43 @@ export const useMorphPersistManyToOne = ({
...
         assertFieldMetadata(
           FieldMetadataType.MORPH_RELATION,
           isFieldMorphRelation,
           fieldDefinition,
         );
+
+        const fieldName = fieldDefinition.metadata.fieldName;
+
+        if (!isDefined(valueToPersist)) {
+          const allNullRecordInput: Record<string, null> =
```
lucasbordeau: Ok I would add context because I had to read the reducer to understand what it does, something like `recordWithAllMorphObjectIdsToNull` would make more self-explanatory.

packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/input/hooks/useMorphPersistManyToOne.ts
```diff
@@ -35,13 +35,43 @@ export const useMorphPersistManyToOne = ({
...
+            fieldDefinition.metadata.morphRelations.reduce(
```
lucasbordeau: Also clean code nitpick here, it can be very nice to create a util that we put into the reduce call, so at the end it looks like : 

```tsx
const recordWithAllMorphObjectIdsToNull = morphRelations.reduce(setAllMorphObjectIdsToNull)
```

packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/input/hooks/useMorphPersistManyToOne.ts
bitloi: Done!

packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/input/hooks/useMorphPersistManyToOne.ts
bitloi: Done!

lucasbordeau (approved):
LGTM, nice contribution.

bitloi:
@lucasbordeau Can we merge the PR and close the issue? :)

PR #17267, opened 2026-01-20
Simplify config file script in sdk

charlesBochet:
Some simplifications about loading the manifest as there is no difference between all entities


packages/twenty-sdk/src/cli/utilities/manifest/utils/manifest-config-loader.ts
```diff
@@ -0,0 +1,192 @@
...
+    importPath ?? path.relative(appPath, filepath).replace(/\\/g, '/');
+
+  const { [entryProperty]: _, ...configWithoutEntry } = config;
+
+  return {
+    ...defaults,
+    ...configWithoutEntry,
+    [`${entryProperty}Name`]: entryName,
+    [`${entryProperty}Path`]: entryPath,
+  } as T;
```
martmull: did you check it works for object or extendObjects? I was expecting no entryName would exist for them isn't it?

martmull (approved):

PR #17271, opened 2026-01-20
Stop workflow in any status

thomtrp:
Workflows can now be cancelled when not started or enqueued.
Also displaying the stop option when selecting all.

We got the case of a client that did an infinite loop. So he had a lot of workflows to stop. Supporting select all would have avoid him to wait for 4 hours so we process the runs throttled. This is not something that is supposed to happen often so I did not see the point of refactoring the endpoint. But I wanted the users to have this option just in case

### DIFF
--- a/packages/twenty-front/src/modules/action-menu/actions/record-actions/constants/WorkflowRunsActionsConfig.tsx
+++ b/packages/twenty-front/src/modules/action-menu/actions/record-actions/constants/WorkflowRunsActionsConfig.tsx
@@ -61,7 +61,9 @@ export const WORKFLOW_RUNS_ACTIONS_CONFIG = inheritActionsFromDefaultConfig({
       scope: ActionScope.RecordSelection,
       Icon: IconPlayerStop,
       shouldBeRegistered: ({ selectedRecord }) => {
-        return selectedRecord?.status === 'RUNNING';
+        const stoppableStatuses = ['NOT_STARTED', 'ENQUEUED', 'RUNNING'];
+
+        return stoppableStatuses.includes(selectedRecord?.status);
       },
       availableOn: [
         ActionViewType.SHOW_PAGE,
--- a/packages/twenty-front/src/modules/action-menu/actions/record-actions/constants/WorkflowRunsActionsConfig.tsx
+++ b/packages/twenty-front/src/modules/action-menu/actions/record-actions/constants/WorkflowRunsActionsConfig.tsx
@@ -60,9 +60,12 @@ export const WORKFLOW_RUNS_ACTIONS_CONFIG = inheritActionsFromDefaultConfig({
       type: ActionType.Standard,
       scope: ActionScope.RecordSelection,
       Icon: IconPlayerStop,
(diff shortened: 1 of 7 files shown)


packages/twenty-front/src/modules/action-menu/actions/record-actions/constants/WorkflowRunsActionsConfig.tsx
```diff
@@ -60,8 +60,13 @@ export const WORKFLOW_RUNS_ACTIONS_CONFIG = inheritActionsFromDefaultConfig({
...
       scope: ActionScope.RecordSelection,
       Icon: IconPlayerStop,
-      shouldBeRegistered: ({ selectedRecord }) => {
-        return selectedRecord?.status === 'RUNNING';
+      shouldBeRegistered: ({ selectedRecord, isSelectAll }) => {
+        if (isSelectAll === true) {
+          return true;
+        }
+
+        const stoppableStatuses = ['NOT_STARTED', 'ENQUEUED', 'RUNNING'];
```
Weiko, line 68: What "STOP" means for a NOT_STARTED workflow run?

packages/twenty-front/src/modules/action-menu/actions/record-actions/single-record/workflow-run-actions/components/StopWorkflowRunSingleRecordAction.tsx
```diff
@@ -1,27 +1,67 @@
...
-      await stopWorkflowRun(selectedRecordId);
+    if (contextStoreTargetedRecordsRule.mode === 'selection') {
+      for (const selectedRecordId of contextStoreTargetedRecordsRule.selectedRecordIds) {
+        await stopWorkflowRun(selectedRecordId);
+      }
+    } else {
+      const records = await fetchAllRecordIds();
+
+      for (const record of records) {
+        await stopWorkflowRun(record.id);
```
Weiko, line 63: Note: Would it be easy to implement a bulk mutation instead? (with filters)

packages/twenty-front/src/modules/action-menu/actions/record-actions/constants/WorkflowRunsActionsConfig.tsx
thomtrp: Not started means the job has not been created yet. The status will be updated to stop and the job will never be created.
Enqueued means the job has been created but not run yet. It will return immediately if the current run is stopped status.

packages/twenty-front/src/modules/action-menu/actions/record-actions/single-record/workflow-run-actions/components/StopWorkflowRunSingleRecordAction.tsx
thomtrp: not complicated. But you're not supposed to stop workflows by batches unless you did something wrong (like an infinite loop). That's why we did it this way in first place

martmull (commented):
lgtm

Weiko (approved):

PR #17272, opened 2026-01-20
fix gmail message import policy

neo773:
Gmail sync was importing messages from folders users had explicitly disabled because the folder filtering logic lived in the parent service but the Gmail driver wasn't aware of the import policy when building its API queries.                                                                                                                                                                                               
                                                                                                                                                                                                             
Moved messageFolderImportPolicy down to the driver level so Gmail can                                                                                                                                      
- Build proper -label: exclusions during initial sync
- Added `buildGmailLabelSearchName` as our syntax for nested folders was wrong
- Filter out messages from disabled folders during incremental sync via history API


packages/twenty-server/src/modules/messaging/message-import-manager/drivers/microsoft/services/microsoft-get-message-list.service.ts
```diff
@@ -41,17 +38,24 @@ export class MicrosoftGetMessageListService {
     connectedAccount,
     messageFolders,
   }: GetMessageListsArgs): Promise<GetMessageListsResponse> {
-    if (messageFolders.length === 0) {
```
neo773, line 44: This is intentional, we don't want to throw anymore if user has no folders selected we just leave a WARN level log.

packages/twenty-server/src/modules/messaging/message-import-manager/drivers/gmail/utils/compute-gmail-exclude-search-filter.util.ts
```diff
@@ -1,12 +1,15 @@
...
     MessageFolderWorkspaceEntity,
-    'externalId' | 'isSynced'
+    'externalId' | 'isSynced' | 'name' | 'parentFolderId'
   >[],
 ) =>
   messageFolders
     .filter((folder) => !folder.isSynced)
-    .map((folder) => `-label:${folder.externalId}`)
+    .map((folder) => buildGmailLabelSearchName(folder, messageFolders))
+    .filter((name): name is string => name !== null)
```
prastoin: ```Question:``` Can't name be undefined here too ? would have used `filter.isDefined`

packages/twenty-server/src/modules/messaging/message-import-manager/drivers/gmail/utils/build-gmail-label-search-name.util.ts
```diff
@@ -0,0 +1,36 @@
...
+  const folderMap = new Map(
+    allFolders
+      .filter((folder) => folder.externalId)
+      .map((folder) => [folder.externalId, folder]),
+  );
+
+  const pathParts: string[] = [];
+  let current: FolderInput | undefined = folder;
+
+  while (current?.name) {
```
prastoin: ***Question:*** Can't this infinite loop in malicious folder configuration ? Can folder have circular deps or symlink and are we validating the folder max depth ?

neo773: circular deps can't happen here. `GmailGetAllFoldersService` is the only thing that creates these entities, and it derives `parentFolderId` by parsing label names—A/B/C's parent is always A/B. parent is always a prefix of child, so cycles are impossible by construction.

packages/twenty-server/src/modules/messaging/message-import-manager/drivers/gmail/utils/compute-gmail-exclude-search-filter.util.ts
neo773: resolved

packages/twenty-server/src/modules/messaging/message-import-manager/drivers/gmail/utils/build-gmail-label-search-name.util.ts
```diff
@@ -0,0 +1,36 @@
...
+  if (!folder.name) {
+    return null;
+  }
+
+  const folderMap = new Map(
+    allFolders
+      .filter((folder) => folder.externalId)
+      .map((folder) => [folder.externalId, folder]),
+  );
+
```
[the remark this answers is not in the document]
neo773: Pending folder actions are processed synchronously before message list fetching

packages/twenty-server/src/modules/messaging/message-import-manager/drivers/gmail/utils/build-gmail-label-search-name.util.ts
prastoin: Ok I see makes sense
Could we still at least define `MAXIMUM_GMAIL_FOLDER_DEPTH` ? What do you think ?

neo773: Can't hurt I guess, pushed a new commit

prastoin (approved):

prastoin:
Cool thanks ! If @charlesBochet can still have an eye on this\

PR #17277, opened 2026-01-20
Post syncable entities identification cleanup

prastoin:
# Introduction

following https://github.com/twentyhq/twenty/pull/17279
As we've finally identified all the syncable metadata entities, which means they're expected to have non nullable applicationId and universalIdentifier at pg_level we can remove previous retro comp universalIdentifier fallbacking and update the dto too


prastoin:
Will reopen after this has been merged https://github.com/twentyhq/twenty/pull/17279

PR #17286, opened 2026-01-20
2093 extensibility add twenty auth switch command

### LINKED ISSUE #2093: [Extensibility] Add twenty auth switch and list commands
See https://github.com/twentyhq/core-team-issues/issues/2346 for context

martmull:
add auth:switch and auth:list commands


packages/twenty-sdk/src/cli/commands/auth/auth-list.ts
```diff
@@ -0,0 +1,55 @@
...
+            '⚠ No workspaces configured. Use `twenty auth:login` to create one.',
+          ),
+        );
+        return;
+      }
+
+      console.log(chalk.blue('Available workspaces:\n'));
+
+      for (const workspace of availableWorkspaces) {
+        // Set active workspace temporarily to read its config
```
Weiko: Can't we avoid a global side effect like this?
Can't getConfig take a parameter instead or am I missing something?

Weiko (commented):
Do we want to keep "profile"? This seems to be mixed with "workspace" in terms of naming

packages/twenty-sdk/src/cli/commands/auth/auth-list.ts
martmull: will look at it. It works well be indeed that way to list is strange

martmull:
> Do we want to keep "profile"? This seems to be mixed with "workspace" in terms of naming

where do you see "profile"?

packages/twenty-sdk/src/cli/commands/auth/auth-list.ts
martmull: done

Weiko (approved):
LGTM

PR #17287, opened 2026-01-20
Handle dynamic predicates

thomtrp:
Fetch and send full workspace member to handle dynamic predicate. I only tested that it did not break the current logic since the frontend does not yet send queries we dynamic predicates.


Weiko (approved):
What do we do with apiKey authContext?
LGTM

PR #17293, opened 2026-01-20
Handled SSE create event

lucasbordeau:
This PR handles SSE create event.

It also fixes a regression on board, which was making it flash with skeletons on any update / create.

This PR was a good test case to experience how each main components of the app : table, board, calendar, reacts to a create event. 

It is not straightforward for now, because each component handles records with its own state management to turn those records into a coherent display of rows or cards. 

The requests for each component are also different : fetch more, group by, multiple requests in parallel, so the cleanest way to handle optimistic effect requires to create one small optimistic engine per component tuned for its internal data logic.

For now we've decided to implement what's doable in a reasonable amount of time and that includes not handling table with groups for now. 

Creating a clean optimistic logic for each component will be done later.

# QA

## Importing 100 records via the API (could be a script, a workflow, an AI calling tools, a CSV import, etc.)

https://github.com/user-attachments/assets/d38a3770-8b0a-4f83-8275-5e7d1be0a5c6

## Creating from different components and seeing the SSE event being processed by other components

https://github.com/user-attachments/assets/106b2f49-19cb-4190-92b7-0653c8373366


packages/twenty-front/src/modules/object-record/hooks/useCreateOneRecord.ts
```diff
@@ -186,17 +191,31 @@ export const useCreateOneRecord = <
 
     await refetchAggregateQueries();
 
+    const positionToUse =
+      recordInput.position === 'first'
+        ? 'first'
+        : recordInput.position === 'last'
+          ? 'last'
+          : null;
```
thomtrp, line 199: is it a copy from an existing backend logic? And if yes, should we have it in twenty-shared?

thomtrp (approved):

PR #17298, opened 2026-01-21
[Dashboards] Dynamic paddings

### LINKED ISSUE #2055: [DASHBOARDS] Make the padding dynamic according to the size of the chart
## Current situation

The graph padding is always the same, it is independent of the chart size
Image

## Desired behavior

The chart padding should be dynamic.

ehconitin:
closes https://github.com/twentyhq/core-team-issues/issues/2055

https://github.com/user-attachments/assets/7a19066f-e891-4dfa-92ee-b85690765e60


ehconitin:
@cubic-dev-ai review please

ehconitin:
@greptileai review please

packages/twenty-front/src/modules/page-layout/widgets/graph/utils/getChartMarginsFromText.ts
```diff
@@ -0,0 +1,168 @@
...
+const getMaxLabelDimensions = ({
+  labels,
+  fontSize,
+  fontFamily,
+}: {
+  labels?: string[];
+  fontSize: number;
+  fontFamily?: string;
+}) =>
+  (labels ?? []).reduce(
```
bosiraphael: Not very readable, return empty array if labels is empty

packages/twenty-front/src/modules/page-layout/widgets/graph/utils/getChartMarginsFromText.ts
```diff
@@ -0,0 +1,168 @@
...
+
+      return {
+        width: Math.max(maxDimensions.width, width),
+        height: Math.max(maxDimensions.height, height),
+      };
+    },
+    { width: 0, height: 0 },
+  );
+
+const estimateRotatedLabelHeight = ({
```
bosiraphael: A lot of logic seems duplicated from `packages/twenty-front/src/modules/page-layout/widgets/graph/utils/computeBottomLegendOffsetFromText.ts‎`

packages/twenty-front/src/modules/page-layout/widgets/graph/utils/measureTextDimensions.ts
```diff
@@ -0,0 +1,71 @@
...
+
+type TextDimensions = {
+  width: number;
+  height: number;
+};
+
+const textCache = new Map<string, TextDimensions>();
+let cachedCanvas: HTMLCanvasElement | null = null;
+
+const getCanvasContext = () => {
```
bosiraphael: Seems overly complicated, I think we could just use a heuristic in this case

lucasbordeau: Yes I agree, measuring an element to derive values should be avoided because it kind of break the pattern of React which is to compute everything into the data state world and then give it to React so it applies visual changes. 

In this case I would just use an array of thresholds with hardcoded paddings, even if it does not look perfectly smooth.

It will be more maintainable. 

Plus overall the chart area has already a lot of utils and complex computation, I think that we should try to head over simplistic solution to keep refactoring affordable in the future.

ehconitin:
@greptileai review please

ehconitin:
@cubic-dev-ai review please

lucasbordeau (approved):
Ok for me from a code perspective, but two points.

First it would nice to have screenshots comparing main and this PR because it's hard to see the difference for a reviewer who is not used to this part of the app. 

Second, this solution is quite complex, but as it works and it is details, let's merge like that, but the maintenance cost is high. Could we think about a less dynamic solution ? For example : computing the total width of the grid on the screen with the div's width, then computing a grid cell's width then using multipliers with this cell width and the size of the graph ? 

Otherwise LGTM

ehconitin:
thanks for the review! Sorry for no screenshots

agree the complexity is higher than ideal. Merging as-is since it works, but creating a followup to explore simplifications

PR #17300, opened 2026-01-21
[Dashboards] deep-clone widget configuration on duplication to avoid shared draft state

ehconitin:
closes 4th issue from this bug bucket - https://github.com/twentyhq/core-team-issues/issues/2109

CleanShot 2026-01-21 at 17 55 59


ehconitin:
This isnt right!

after further debugging, the issue lies only on line charts duplication??? -- which is weird!
I am digging more

ehconitin:
this isnt the right fix -- the issue liees in apollo caching where -- fix should be adding unique ids on line chart

PR #17304, opened 2026-01-21
Fix identify remaining entities command to cover soft-deleted rows

prastoin:
# Introduction
Some entities aren't passing the migration due to not covering the soft deleted rows

## Entities that implements soft deletion
Inserted with universal programmatically
- rowLevelPermissionPredicate
- rowLevelPermissionPredicateGroup
- pageLayout
- pageLayoutTab
- pageLayoutWidget

Legacy might contains null
- viewFilterGroup
- serverlessFunction
- viewSort


Weiko (approved):
LGTM

PR #17306, opened 2026-01-21
[Apps] Get rid of .yarn binaries in apps

### LINKED ISSUE #1956: [Extensibility] [v2] Stop commiting .yarn in applications
This is a dangerous attack vector we should not commit the .yarn/releases/** folder of applications

Instead we want to use `create-twenty-app` to install a new application or an existing application. `create-twenty-app` will itself provide the .yarn/release folder.

ijreilly:
Fixes https://github.com/twentyhq/core-team-issues/issues/1956

**Problem**
Within an app, the `.yarn/releases/` folder contains executable Yarn binaries that run when executing any yarn command (`.yarnrc` file indicates yarn path to be `.yarn/releases/yarn-4.9.2.cjs `.)
This is a supply chain attack vector: a malicious actor could submit a PR with a compromised `yarn-4.9.2.cjs binary`, which would execute arbitrary code on developers' machines or CI systems.

**Fix**
Actually, thanks to Corepack, we don't need to store and execute this binary. 
Corepack can be seen as the manager of a package manager: in `package.json` we indicate a packageManager version like `"packageManager": "yarn@4.9.2"`, and when executing `yarn` Corepack will securely fetch the verified version from npm, avoiding the risk of executing a compromised binary committed to the repository. This was already in our app's package.json template but we were not using it!

We can now
- remove the folder containing the binary from our app template base-application (that is scaffolded when creating an app through cli), `.yarn/releases/`, and remove `yarnPath: .yarn/releases/yarn-4.9.2.cjs` from its .yarnrc
- remove them from the community apps that were already published in the repo
- add .yarn to gitignore 

**Tested**
This has been tested and works for app created in the repo, outside the repo, and existing apps in the repo


packages/create-twenty-app/src/constants/base-application/.yarn/releases/yarn-4.9.2.cjs
martmull, line 1: on doit le garder celui là

packages/create-twenty-app/src/constants/base-application/.yarn/install-state.gz
martmull, line 1: celui là on peut le supprimer oui je pense

packages/create-twenty-app/src/constants/base-application/.yarnrc.yml
martmull, line 1: faut garder

packages/twenty-docs/developers/extend/capabilities/apps.mdx
```diff
@@ -86,8 +83,6 @@ my-twenty-app/
   .nvmrc
   .yarnrc.yml
```
martmull, line 87: faut garder

martmull (changes requested):

martmull (approved):

PR #17309, opened 2026-01-21
Use new SSE setup for workflow runs

thomtrp:
- Add a specific subscriber for workflow run
- Workflow runs use apollo cache. Adding updateRecordFromCache on record updates


lucasbordeau (approved):
LGTM

PR #17313, opened 2026-01-21
fix: prevent default object re-selection in relation field form

### LINKED ISSUE #17111: Default value selected on change for Object destination on Relation field
## Bug Description

This is a minor issue with the user interface. When you create a Relation field on an object, when you select the "Object destination", after de-selecting the preselected object (e.g. Companies), when you select a new object, the initial one is also selected, and "2 Objects" is shown.

**Steps to reproduce:** 

1. From Data Model in Settings, add a new Relation field on an object.
2. Notice "Company" is pre-selected under "Object destination"
3. Expand the drop-down, de-select the default object and select a different object (e.g. Opportunity)
4. Notice both values are selected (the initial one is selected again).

Image

## Expected behavior

Only the one option you select should be selected, not two.

## Workaround

Remove the initial default selection after it's re-added.

claytonlin1110:
Fixes #17111

### Problem
When creating a Relation field, after deselecting the pre-selected default object (e.g., Company) and selecting a different object (e.g., Opportunity), both objects would end up being selected, showing "2 Objects" instead of just the newly selected one.

### Root Cause
The `initialMorphRelationsObjectMetadataIds` array was being recreated on every component render, causing react-hook-form's `Controller` to treat the `defaultValue` prop as a new value and re-apply it after user changes.

### Solution
1. **Memoized the initial value**: Used `useMemo` to ensure `initialMorphRelationsObjectMetadataIds` has a stable reference across renders
2. **Added initialization guard**: Used `useEffect` with a `useRef` flag to ensure the default value is only set once during component mount
3. **Maintained Controller defaultValue**: Kept the `defaultValue` prop on the Controller, which now works correctly with the stable memoized value

### Changes
- `SettingsDataModelFieldRelationForm.tsx`: Added memoization and initialization guard to prevent default value re-application

### DIFF
--- a/packages/twenty-front/src/modules/settings/components/SettingsMorphRelationMultiSelect.tsx
+++ b/packages/twenty-front/src/modules/settings/components/SettingsMorphRelationMultiSelect.tsx
@@ -118,7 +118,7 @@ export const SettingsMorphRelationMultiSelect = ({
     }));
 
   const selectedOptions = options.filter((option) =>
-    (selectedObjectMetadataIds ?? []).includes(option.objectMetadataId),
+    localSelectedObjectMetadataIds.includes(option.objectMetadataId),
   );
 
   const filteredOptions = useMemo(
--- a/packages/twenty-front/src/modules/settings/components/SettingsMorphRelationMultiSelect.tsx
+++ b/packages/twenty-front/src/modules/settings/components/SettingsMorphRelationMultiSelect.tsx
@@ -1,5 +1,5 @@
 import styled from '@emotion/styled';
-import { plural } from '@lingui/core/macro';
+import { plural, t } from '@lingui/core/macro';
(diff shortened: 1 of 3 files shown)


claytonlin1110:
Can anyone review this?

etiennejouan:
Hi @claytonlin1110, thank you for your contribution. 
I've tested your PR, it does not work. Also you have some CI issue, you need to run the linter fix command.
I close the PR you can re-open it by commenting "/twenty PR open" when done.

claytonlin1110:
/twenty PR open
let me reopen it and make it a draft. once i'm done, i will reopen this PR

etiennejouan (approved):
Hi @claytonlin1110 thank you for your contribution, the first one, congrats ! 👏 
I tested and it works fine ! but it looks a bit too complicated for an issue like this, so I took the liberty to update.

PR #17315, opened 2026-01-21
fix: remove empty else-if branches and add numbering

abdulrahmancodes:


thomtrp (approved):
LGTM, thanks @abdulrahmancodes!

PR #17319, opened 2026-01-21
fix isCalendarFieldReadOnly function to check if calender field is re…

Mazen-Saeed:
This PR fixes the inconsistency between Calendar and Table views when trying to edit createdAt date field. 

Previously, calendar cards could be dragged even though the createdAt field are UI read-only, resulting in the confusing and inconsistent behavior compared to the Table view where the field is not editable.

The issue was caused by the drag logic checking only user permissions and not the field’s UI read-only metadata. This change updates the calendar drag logic to also check for isUIReadOnly, ensuring that records are not draggable when the selected calendar date field is read-only.


Mazen-Saeed:
Could i get a review here?

lucasbordeau (approved):

lucasbordeau (approved):

PR #17327, opened 2026-01-22
Add vscode tasks for integration test run and inputs to debug mode

prastoin:
# Introduction
Created a task that allow spawning a term that will run the opened file and run its integration tests

## Motivation
Jest extension is quite buggy lately, this might just be a tmp workaround but we can't get `run test` within the test file directly. Also passing by task allows giving inputs

## Usage

`cmd+shift+P`-> `run task` -> `twenty server -  run integration test file` -> fill inputs prompts:
- watch: will rerun on every related files updates
- updateSnapshot

Note: you can add a custom shortcut to the task too, also it will appear in task history

Watch mode won't re-create the server instance on each run, meaning that nest won't hit the `createServer` making very fast to iterate with
Down side of this is that any module injection won't get hydrated

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -68,6 +68,7 @@
         "./jest-integration.config.ts",
         "${relativeFile}",
         "--silent=false",
+        "${input:updateSnapshot}"
       ],
       "cwd": "${workspaceFolder}/packages/twenty-server",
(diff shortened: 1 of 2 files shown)


Devessier (approved):
👍

Weiko (approved):

PR #17329, opened 2026-01-22
Disable query logs if I want to

martmull:
As tile

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/twenty-config/config-variables.ts
+++ b/packages/twenty-server/src/engine/core-modules/twenty-config/config-variables.ts
@@ -10,6 +10,7 @@ import {
   validateSync,
 } from 'class-validator';
 import { isDefined } from 'twenty-shared/utils';
+import { type LoggerOptions } from 'typeorm/logger/LoggerOptions';
 
 import { type AwsRegion } from 'src/engine/core-modules/twenty-config/interfaces/aws-region.interface';
 import { NodeEnvironment } from 'src/engine/core-modules/twenty-config/interfaces/node-environment.interface';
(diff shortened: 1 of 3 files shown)


packages/twenty-server/src/engine/core-modules/twenty-config/twenty-config.service.ts
```diff
@@ -202,7 +202,7 @@ export class TwentyConfigService {
   getLoggingConfig(): LoggerOptions {
     switch (this.get('NODE_ENV')) {
       case NodeEnvironment.DEVELOPMENT:
-        return ['query', 'error'];
+        return this.get('TYPEORM_LOGGING');
```
charlesBochet: if we do that, we should remove the switch about the node environment.

packages/twenty-server/src/engine/core-modules/twenty-config/config-variables.ts
```diff
@@ -589,6 +591,17 @@ export class ConfigVariables {
...
+  @ConfigVariablesMetadata({
+    group: ConfigVariablesGroup.LOGGING,
+    description:
+      'TypeORM logging options for development mode. Accepts comma-separated values: query, schema, error, warn, info, log, migration',
+    type: ConfigVariableType.ARRAY,
+    options: ['query', 'schema', 'error', 'warn', 'info', 'log', 'migration'],
+  })
+  @CastToTypeORMLogLevelArray()
+  @IsOptional()
+  TYPEORM_LOGGING: LoggerOptions = ['query', 'error'];
```
charlesBochet: default should be error for production case

martmull: ok

packages/twenty-server/src/engine/core-modules/twenty-config/twenty-config.service.ts
martmull: ok

packages/twenty-server/src/engine/core-modules/twenty-config/decorators/cast-to-typeorm-log-level-array.decorator.ts
```diff
@@ -0,0 +1,29 @@
+import { Transform } from 'class-transformer';
+
+export const CastToTypeORMLogLevelArray = () =>
+  Transform(({ value }: { value: string }) => toTypeORMLogLevelArray(value));
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const toTypeORMLogLevelArray = (value: any) => {
+  if (typeof value === 'string') {
```
charlesBochet: use type guard isNonEmptyString

charlesBochet (approved):

PR #17332, opened 2026-01-22
Switch coverage provider from Istanbul to V8

mabdullahabaid:
Switched Storybook coverage collection to the V8 provider while keeping NYC for merging/reporting. V8 uses native runtime coverage (lower instrumentation overhead) but still emits Istanbul‑compatible output, so NYC continues to merge shards and enforce thresholds. 

This should theoretically make coverage runs faster/lighter without changing the overall reporting pipeline. Replacing NYC does not offer benefits except less tools to maintain - ultimately, it's just merging JSON reports and reporting the statistics. Therefore, I believe we're good with NYC for now.


mabdullahabaid:
Closing the PR for now since performance tests are not fully stable when the coverage provider is switched from Istanbul to V8. This would require a rework of test cases and a relatively deeper dive into ensuring we do not introduce regressions.

PR #17337, opened 2026-01-22
[SSE] Event stream TTL refresh

thomtrp:
This PR update the redis subscription iterator wrapper with an heartbeat system. Heartbeat interval are based on event stream TTL. Those refresh the event stream TTL and active streams in redis.

I also cleaned a few functions that were not used anymore.


packages/twenty-server/src/engine/subscriptions/event-stream.service.ts
```diff
@@ -214,40 +204,22 @@ export class EventStreamService {
...
+      this.cacheStorageService.expire(activeStreamsKey, EVENT_STREAM_TTL_MS),
+    ]);
 
-  private isQueryMatchingEvent(
-    operationSignature: RecordGqlOperationSignature,
-    event: ObjectRecordSubscriptionEvent,
-  ): boolean {
-    // to be improved
-    return operationSignature.objectNameSingular === event.objectNameSingular;
+    return streamRefreshed;
```
Weiko: This only returns the result of your first expire, should probably deconstruct both promises and return 1 && 2?

packages/twenty-server/src/engine/workspace-event-emitter/utils/wrap-async-iterator-with-lifecycle.ts
```diff
@@ -0,0 +1,61 @@
...
+      heartbeatInterval = null;
+    }
+    if (onCleanup) {
+      await onCleanup();
+    }
+  };
+
+  return {
+    next: () => iterator.next(),
+    return: async () => {
```
Weiko, line 60: Race condition: The only way to clear the interval is to call cleanup.
Since you start the interval before the first iteration, if something happens between the interval start and the iterator start, the interval might never be cleaned up and call onHeartbeat forever. The fix would be to lazily call startInterval in the first next().

Weiko (commented):
Left some comments @thomtrp

packages/twenty-server/src/engine/workspace-event-emitter/utils/wrap-async-iterator-with-lifecycle.ts
```diff
@@ -0,0 +1,61 @@
...
+      clearInterval(heartbeatInterval);
+      heartbeatInterval = null;
+    }
+    if (onCleanup) {
+      await onCleanup();
+    }
+  };
+
+  return {
+    next: () => iterator.next(),
```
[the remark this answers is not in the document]
thomtrp: Redis pub/sub iterators never complete normally. They wait for messages indefinitely. So unless return or throw are called, no need for cleanup

packages/twenty-server/src/engine/subscriptions/event-stream.service.ts
thomtrp: TBH this is not used but still more rigorous 👍

packages/twenty-server/src/engine/workspace-event-emitter/utils/wrap-async-iterator-with-lifecycle.ts
thomtrp: Good catch. Updated!

Weiko (approved):
LGTM!

PR #17338, opened 2026-01-22
Fix menu item toggle

bosiraphael:
## Before

https://github.com/user-attachments/assets/c26ad780-3a6a-4ccf-91a9-1c7e1bd1413d

## After

https://github.com/user-attachments/assets/40f951c7-6bcf-45a3-98bc-c631f5ec883b


ehconitin (approved):
lgtm

PR #17340, opened 2026-01-22
Fix blank page layout after navigation

Devessier:
Until now, it wasn’t possible to navigate between page layouts. Dashboards don’t contain links to other dashboards. With record page layouts, however, a page layout can now contain a link to another page layout. If the user opens a record in the side panel and then clicks a link to another record, the current page layout is replaced with the page layout for the clicked record.

In this scenario, the `PageLayoutRenderer` component is not remounted. As a result, the `isInitialized` state was not reset to `false`, and the corresponding initialization `useEffect` was not triggered again.

All page layout states are component states bound to `PageLayoutComponentInstanceContext`. For example, after navigating to another record, `pageLayoutPersistedComponentState` was actually `undefined` because the context's `instanceId` had changed.

Replacing the `isInitialized` state with a component state fixes the bug and is consistent with the existing pattern of using component states to store everything related to page layouts.

## Before

https://github.com/user-attachments/assets/24217145-51e5-49ef-8180-de62a6acfe10

## After

https://github.com/user-attachments/assets/e0ffe107-8fae-4f3a-9016-72c85bc735f4


bosiraphael (approved):
LGTM

PR #17355, opened 2026-01-22
Remove useUpdateOneRecordV2

ijreilly:
Previously we introduced useUpdateOneRecordV2 to handle morph relations.
In this PR we are removing it to only use useUpdateOneRecord which has been adapted not to requires objectMetadataNameSingular in its props.


charlesBochet (approved):
LGTM

PR #17357, opened 2026-01-22
Refactor rls backend simplify add tests

Weiko:
## Context
Removing full CRUD to RLS since we are actually using an upsert only over the role resolver, this removes a lot of unused code (could be re-added later but unlikely). Simplified the folder arch at the same time
Add simple integration tests to RLS


thomtrp (approved):

Weiko:
This is not a breaking change for the frontend, only generated has been updated but there were no usage in the frontend (hence why this refactoring). 
Also I'm assuming nobody was consuming this API since the feature is still early and behind a feature flag

PR #17360, opened 2026-01-22
Shrink table rows

Devessier:
We often use the `fr` unit in a CSS grid thinking that it will force the max-width of the grid's items. However, [`1fr` is always equal to `minmax(auto, 1fr)`](https://stackoverflow.com/a/52861514). The consequences are:

- By default, the width of the element will be `1fr`.
- However, if the size of the element becomes wider than the resolved dimension of `1fr`, the item will expand to match its `auto` width. This can be easily encountered when an item has a really long text. The `auto` size will be the dimension of the long text displayed on a single line.

If you want your element not to overflow, you can use `minmax(0, 1fr)` instead of `1fr`.

This is a known behavior in the community. CSS frameworks like Tailwind CSS even made their utility classes setting grid columns use `minmax(0, xfr)` by default.

CleanShot 2026-01-22 at 16 11 00@2x

See: https://www.bigbinary.com/blog/understanding-the-automatic-minimum-size-of-flex-items.

Another fix is to use `min-width: 0;` on items themselves, instead of using `minmax(0, 1fr)` at the grid level. This would make it possible to create a reusable `Td` component.

### Before
CleanShot 2026-01-22 at 15 46 48@2x

### After
CleanShot 2026-01-22 at 15 47 00@2x

### Before
CleanShot 2026-01-22 at 16 00 33@2x

### After
CleanShot 2026-01-22 at 16 00 44@2x

### Before
CleanShot 2026-01-22 at 15 50 16@2x

### After
CleanShot 2026-01-22 at 15 50 01@2x


martmull (commented):
Nice

martmull (approved):

PR #17361, opened 2026-01-22
Allow variables with dots and keys

thomtrp:
Fixes https://github.com/twentyhq/private-issues/issues/410#issuecomment-3781085655

Currently, JSON keys with spaces like { "toto toto": 123 } are rejected with "JSON keys cannot contain spaces" error. This is problematic for HTTP requests and webhook triggers where users cannot control the response structure.

We use Handlebars to eval variables, segment-literal bracket notation to escape keys with special characters:
Normal: {{step.normalKey}}
With spaces: `{{step.[key with space]}}`

So we simply need to wrap segments with spaces with brackets.

This PR: 
- Create shared path utilities to wrap the variable segments when needed
- Use it in all variable generation places
- Remove the restrictions

This body is now supported:
Capture d’écran 2026-01-22 à 16 10 13

### DIFF
--- a/packages/twenty-shared/src/workflow/utils/variable-path.util.ts
+++ b/packages/twenty-shared/src/workflow/utils/variable-path.util.ts
@@ -1,0 +1,70 @@
+// Characters that require bracket escaping in variable paths
+// Spaces, dots, and brackets would break the dot-notation parsing
+const SPECIAL_CHARS_REGEX = /[\s.[]/;
+
+export const needsEscaping = (key: string): boolean =>
+  SPECIAL_CHARS_REGEX.test(key);
+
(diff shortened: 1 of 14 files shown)


packages/twenty-shared/src/workflow/utils/variable-path.util.ts
```diff
@@ -0,0 +1,70 @@
...
+    current += char;
+    segmentIndex++;
+  }
+
+  if (current.length > 0) {
+    segments.push(current);
+  }
+
+  return segments;
+};
```
Devessier, line 70: This is a good looking parser state machine 👀

Devessier (approved):
LGTM

PR #17364, opened 2026-01-22
Follow up dev and build

martmull:
fresh eye report

- Improve template entities
- Ignore .twenty properly
- Add base src alias to tsconfig

### DIFF
--- a/packages/create-twenty-app/src/constants/base-application/tsconfig.json
+++ b/packages/create-twenty-app/src/constants/base-application/tsconfig.json
@@ -22,6 +22,9 @@
     "skipLibCheck": true,
     "skipDefaultLibCheck": true,
     "resolveJsonModule": true,
+    "paths": {
+      "src/*": ["./src/*"]
+    }
   },
   "exclude": ["node_modules", "dist", "**/*.test.ts", "**/*.spec.ts"]
 }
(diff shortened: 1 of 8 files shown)


packages/create-twenty-app/src/constants/base-application/tsconfig.json
```diff
@@ -22,6 +22,9 @@
     "skipLibCheck": true,
     "skipDefaultLibCheck": true,
     "resolveJsonModule": true,
+    "paths": {
+      "src/*": ["./src/*"]
```
charlesBochet: I think we should put a link to ./ and not to ./src which is not enforced

why not ~/ pointing to "." ?

packages/twenty-sdk/.gitignore
```diff
@@ -1,2 +1 @@
 node_modules
-.twenty
```
charlesBochet, line 2: that's useful for integration tests who are building apps?

martmull: mmh, maybe, checking

packages/create-twenty-app/src/constants/base-application/tsconfig.json
martmull: yes indeed

packages/twenty-sdk/.gitignore
martmull: yes indeed

charlesBochet (approved):

Weiko (approved):
LGTM

PR #17372, opened 2026-01-22
fix: allow board column tag to shrink so aggregate value remains visible

bitloi:
## Summary
When a select option label is long, it was pushing the aggregate value (e.g., '12.6m') out of view because the tag had `flex-shrink: 0`.

Changed to `min-width: 0` and `overflow: hidden` to allow the tag to shrink and truncate with ellipsis, keeping the aggregate value visible.

## Changes
- Modified `StyledTag` in [RecordBoardColumnHeader.tsx](cci:7://file:///root/74/bronze/twenty/packages/twenty-front/src/modules/object-record/record-board/record-board-column/components/RecordBoardColumnHeader.tsx:0:0-0:0) to allow shrinking

The [Tag](cci:1://file:///root/74/bronze/twenty/packages/twenty-ui/src/components/tag/Tag.tsx:100:0-141:2) component already has built-in text truncation with `OverflowingTextWithTooltip`, so this change enables that functionality to work properly.

Fixes #17350
```**


bitloi:
To reproduce the aggregate values like "5.5m" shown in the screenshot:
1. Click on the "-" next to any column header
2. Select "Sum" (or another aggregate operation)
3. Choose the "Amount" field

The aggregate value will then display instead of "-".

allow board column

@lucasbordeau Can you please review the changes?

bitloi:
@lucasbordeau Please let me know the result.

lucasbordeau:
Nice, added max-width so overflow can trigger : 
image

lucasbordeau (approved):

PR #17381, opened 2026-01-23
Migrate attachments to morph relations + fix morph join column filtering

### LINKED ISSUE #1744: Attachements

mabdullahabaid:
Closes [1744](https://github.com/twentyhq/core-team-issues/issues/1744).

This PR migrates attachments to morph relations behind a feature flag, following the TimelineActivity pattern. it introduces the `IS_ATTACHMENT_MIGRATED` flag, updates standard field metadata and indexes to use morph relations, adds a workspace migration that renames `attachment.*Id` columns to `target*Id` and converts the corresponding field metadata to `MORPH_RELATION` with a shared `morphId`. On the frontend, attachment read/write paths now switch to `target*Id` when the flag is enabled.

It also fixes optimistic filtering for morph join columns. The metadata API deduplicates morph fields, so attachments now expose a single target field of type `MORPH_RELATION` plus a `morphRelations` array listing each target object. Because only one `settings.joinColumnName` is returned (e.g. `targetRocketId`), filters like `targetCompanyId` don’t map to any field and the optimistic cache code throws. `doesMorphRelationJoinColumnMatch` resolves this by computing all valid join column names from `morphRelations` using `computeMorphRelationFieldName` and comparing them to the filter key. That makes filters like `targetCompanyId` resolvable even with a single target field, so attachment uploads and list matching no longer crash.

image

<br />
<br />

Today the metadata API returns one morph field called `target` and a list of possible targets (`morphRelations`), but it does not tell us the join column for each target. That’s why the Frontend had to compute join column names.

If we want to fix this at the API level, there are two options:

- Add join column names to each target in `morphRelations` (e.g. company → `targetCompanyId`). This is additive and low‑risk.
- Return each target as its own field (`targetCompany`, `targetPerson`, etc.) instead of a single target. This is a larger change because it changes the shape of metadata and would require more UI updates.

---

> [!NOTE]
> Introduces morph relations for attachments behind `IS_ATTACHMENT_MIGRATED`, aligning server schema/metadata and frontend behavior.
> 
> - Adds `IS_ATTACHMENT_MIGRATED` flag (frontend/server) and seeds/defaults; updates generated GraphQL enums
> - New workspace upgrade `1.17` command migrates data: renames `attachment.*Id` → `target*Id` and converts related fields to `MORPH_RELATION` with shared `morphId`
> - Updates standard field metadata and indexes to `target*` (attachment + related objects), dev seeds, snapshots, and workspace entity types
> - Frontend: switches attachment read/write filters via `getActivityTargetObjectFieldIdName` using the feature flag; updates hooks/components (`useAttachments`, `useUploadAttachmentFile`, editors); expands `Attachment` type
> - Fixes optimistic cache filtering to recognize morph join columns in `isRecordMatchingFilter` by computing valid join-column keys from `morphRelations`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f208fa23b1135af114f484f6d65041a816d6e718. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
```diff
@@ -0,0 +1,258 @@
...
+            error,
+          );
+
+          throw error;
+        }
+      }
+
+      this.logger.log(`✅ Successfully migrated attachment fieldmetadata`);
+
+      await queryRunner.commitTransaction();
```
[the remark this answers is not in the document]
mabdullahabaid: 1-13-migrate-timeline-activity-to-morph-relations.command.ts followed the same pattern. 

Once executed, I believe the command will never run again on a specific server, so we do not need to over-engineer imo.

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
```diff
@@ -0,0 +1,258 @@
...
+        { old: 'note', new: 'targetNote' },
+        { old: 'person', new: 'targetPerson' },
+        { old: 'company', new: 'targetCompany' },
+        { old: 'opportunity', new: 'targetOpportunity' },
+        { old: 'dashboard', new: 'targetDashboard' },
+        { old: 'workflow', new: 'targetWorkflow' },
+        ...customObjectMetadataNames.map((customObjectName) => ({
+          old: `${customObjectName}`,
+          new: `target${capitalize(customObjectName)}`,
+        })),
```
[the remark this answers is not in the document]
mabdullahabaid: Attachments for custom objects are many‑to‑one relations on the attachment side, so the join columns live on the attachment table and are expected to exist. Attachments for custom objects are not managed via a reverse relationship as you mentioned.

FelixMalfait (approved):
Great work!!!
Let's do taskTarget/noteTarget but only consider these as MORPH for now (as if noteTarget was a normal object, don't try to solve the migration to MANY-TO-MANY in the same PR, we will try it late). So noteTarget+taskTarket should be very similar to this PR (maybe one PR for both is enough)

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
```diff
@@ -0,0 +1,258 @@
...
+        where: {
+          workspaceId,
+          isCustom: true,
+        },
+      });
+      const customObjectMetadataNames = customObjectMetadata.map(
+        (objectMetadata) => objectMetadata.nameSingular,
+      );
+
+      const fieldNameMigrations = [
```
prastoin, line 89: ***Suggestion:*** We can directly identify such field, at least if existing for the given workspace, using their `universalIdentifier`

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
```diff
@@ -0,0 +1,258 @@
...
+      ];
+
+      const fieldMigrations = fieldNameMigrations.map(
+        ({ old: oldFieldName, new: newFieldName }) => ({
+          old: `${oldFieldName}Id`,
+          new: `${newFieldName}Id`,
+        }),
+      );
+
+      const attachmentObjectMetadata =
```
prastoin, line 110: same

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
```diff
@@ -0,0 +1,258 @@
...
+
+      await queryRunner.commitTransaction();
+
+      await this.featureFlagService.enableFeatureFlags(
+        [FeatureFlagKey.IS_ATTACHMENT_MIGRATED],
+        workspaceId,
+      );
+
+      await this.workspaceCacheService.invalidateAndRecompute(workspaceId, [
+        'flatFieldMetadataMaps',
```
prastoin, line 231: ***Suggestion:*** We should use such pattern in order to not forget any related cache entries
```ts
    const relatedMetadataNames = getMetadataRelatedMetadataNames('viewField');
    const relatedCacheKeysToInvalidate = relatedMetadataNames.map(
      getMetadataFlatEntityMapsKey,
    );

    this.logger.log(
      `Invalidating caches: ${relatedCacheKeysToInvalidate.join(' ')}`,
    );
    if (!options.dryRun) {
      await this.workspaceCacheService.invalidateAndRecompute(workspaceId, [
        'flatViewFieldMaps',
        ...relatedCacheKeysToInvalidate,
      ]);
    }
```

packages/twenty-server/src/engine/workspace-manager/workspace-migration/constant/default-feature-flags.ts
```diff
@@ -2,4 +2,5 @@ import { FeatureFlagKey } from 'src/engine/core-modules/feature-flag/enums/featu
 
 export const DEFAULT_FEATURE_FLAGS = [
   FeatureFlagKey.IS_TIMELINE_ACTIVITY_MIGRATED,
+  FeatureFlagKey.IS_ATTACHMENT_MIGRATED,
```
prastoin, line 5: Nice

prastoin (commented):
Hey there ! Solid work over there
We should be sure to test ( sorry if redundant also a reminder to myself ):
- False Feature flag implem is working as expected on existing workspaces
- True Feature flag implem is working as expected on new workspaces
- Refactor as possible using new universalIdentifier the backfill command on a prod extract ( lets reach out on monday )

mabdullahabaid:
@prastoin thank you for the review and suggestions, will look into them tomorrow!

PR #17385, opened 2026-01-23
Add multiple recipients (To/CC/BCC) to Send Email workflow action

neo773:
Adds support for multiple recipients in the Send Email workflow action

image

/closes #17383

### DIFF
--- a/packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/components/EmailRecipientsInput.tsx
+++ b/packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/components/EmailRecipientsInput.tsx
@@ -1,6 +1,7 @@
 import { FormFieldInputContainer } from '@/object-record/record-field/ui/form-types/components/FormFieldInputContainer';
 import { FormTextFieldInput } from '@/object-record/record-field/ui/form-types/components/FormTextFieldInput';
 import { InputLabel } from '@/ui/input/components/InputLabel';
+import { extractNonEmptyEmails } from '@/workflow/workflow-steps/workflow-actions/email-action/utils/extractNonEmptyEmails';
 import { WorkflowVariablePicker } from '@/workflow/workflow-variables/components/WorkflowVariablePicker';
 import styled from '@emotion/styled';
 import { t } from '@lingui/core/macro';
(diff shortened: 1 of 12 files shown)


packages/twenty-server/src/modules/workflow/workflow-executor/workflow-actions/mail-sender/types/workflow-send-email-action-input.type.ts
```diff
@@ -1,6 +1,13 @@
+export type EmailRecipients = {
```
FelixMalfait, line 1: This is exported but re-defined elsewhere. 
Also ideally 1 file = 1 export

packages/twenty-server/src/engine/core-modules/tool/tools/send-email-tool/send-email-tool.ts
```diff
@@ -121,6 +121,50 @@ export class SendEmailTool implements Tool {
...
+  } {
+    if (parameters.recipients && parameters.recipients.to.length > 0) {
+      return {
+        to: parameters.recipients.to,
+        cc: parameters.recipients.cc ?? [],
+        bcc: parameters.recipients.bcc ?? [],
+      };
+    }
+
+    if (parameters.email) {
```
FelixMalfait, line 137: Is this only for legacy and to be cleaned up later? Maybe add a comment for context for when this should be cleaned up?

packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/components/WorkflowEditActionSendEmail.tsx
```diff
@@ -82,12 +89,29 @@ export const WorkflowEditActionSendEmail = ({
...
+    const legacyEmail = action.settings.input.email;
+
+    const recipients: EmailRecipients = inputRecipients
+      ? {
+          to: inputRecipients.to ?? [],
+          cc: inputRecipients.cc ?? [],
+          bcc: inputRecipients.bcc ?? [],
+        }
+      : {
+          to: legacyEmail ? [legacyEmail] : [],
```
FelixMalfait, line 103: Same, add comment for cleanup later?

packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/components/WorkflowEditActionSendEmail.tsx
```diff
@@ -56,9 +57,15 @@ type WorkflowFile = {
   createdAt: string;
 };
 
+type EmailRecipients = {
```
FelixMalfait, line 60: twenty-shared?

packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/components/EmailRecipientsInput.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  return (
+    <FormFieldInputContainer>
+      <InputLabel>{label}</InputLabel>
+      <StyledContainer>
+        {entries.map((entry) => {
+          const isLastEntry = entry.id === entries[entries.length - 1].id;
+
+          return (
+            <StyledEmailRow key={entry.id} readonly={readonly}>
```
FelixMalfait, line 88: As discussed I feel like this UX works well for static inputs, but does not work well when you have a variable which contains a list of email. That's why I would be inclined to keeping a single field and be smart automatically (explode if comma separated, if it's an array then consider all emails in array, if it's a string then consider it the single email)

PR #17387, opened 2026-01-23
Rename `PageLayoutWidget` configuration `DTO` -> `ValidationSchema`

prastoin:
# Introduction
Replaced `Dto` naming to `ValidationSchema` as these class validators are only used in inputs validation
Still inferring types for the entity `configuration` prop

## Motivation
About to standardize entities jsonb Property, narrowed and non narrowed typing, in order to implement `universalSettings`


Weiko (commented):
As discussed ValidationSchema should not have ObjectType decorator, otherwise LGTM

prastoin:
Well in the end would either require everything to be duplicated as both `Dto` and `ValidationSchema`
Can't find a satisfying pattern right now
Closing for the moment

PR #17388, opened 2026-01-23
[DASHBOARDS] Fix pie chart gap when there is only one slice

bosiraphael:
## Before
CleanShot 2026-01-23 at 11 00 44@2x

## After
CleanShot 2026-01-23 at 11 00 28@2x


ehconitin (approved):
lgtm

PR #17392, opened 2026-01-23
Add prometheus exporter

thomtrp:
Create a metric endpoint and expose prometheus gauge for event stream count.


packages/twenty-server/src/engine/core-modules/cache-storage/services/cache-storage.service.ts
```diff
@@ -206,6 +206,38 @@ export class CacheStorageService {
...
 
+  async scanAndCountSetMembers(scanPattern: string): Promise<number> {
+    if (!this.isRedisCache()) {
+      throw new Error(
+        'scanAndCountSetMembers is only supported with Redis cache',
+      );
+    }
+
+    const redisClient = (this.cache as RedisCache).store.client;
+    let cursor = 0;
```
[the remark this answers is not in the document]
thomtrp: this would require a way larger database than we have

etiennejouan (commented):
Looks good but I wonder why we don't rely on OTEL API (for all metrics collection) ?

etiennejouan (approved):
Tested and it works great ! 👏 

(Difference with previous implementation is that all metrics are exported on /metrics endpoint, not only the one you want)

packages/twenty-server/src/engine/core-modules/cache-storage/services/cache-storage.service.ts
```diff
@@ -206,6 +206,38 @@ export class CacheStorageService {
...
+        MATCH: `${this.namespace}:${scanPattern}`,
+        COUNT: 100,
+      });
+
+      cursor = result.cursor;
+      const keys = result.keys;
+
+      if (keys.length > 0) {
+        const counts = await Promise.all(
+          keys.map((key) => redisClient.sCard(key)),
```
Weiko, line 231: This could be painful (N network calls), can we use Redis pipeline for that instead?

Weiko: Probably even better with LUA

PR #17394, opened 2026-01-23
Encrypt/decrypt app secret variables

### LINKED ISSUE #1724: [Extensibility] Offuscated secret applicationVariables
as title, From a @Weiko comment in https://github.com/twentyhq/twenty/pull/15143

Fix and merge https://github.com/twentyhq/twenty/pull/15283

ijreilly:
Closes https://github.com/twentyhq/core-team-issues/issues/1724
From PR https://github.com/twentyhq/twenty/pull/15283, followed same implementation
- Introduce EnvironmentModule to provide type safety for env-only variables
- Encrypt secret variables
- When querying app secret variable, display up to first 5 characters (depending on secret length) then `******`

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/applicationVariable/application-variable.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/applicationVariable/application-variable.service.ts
@@ -1,9 +1,9 @@
 import { Injectable } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
 
+import { ApplicationVariables } from 'twenty-shared/application';
 import { isDefined } from 'twenty-shared/utils';
 import { In, Not, Repository } from 'typeorm';
-import { ApplicationVariables } from 'twenty-shared/application';
 
 import { ApplicationVariableEntity } from 'src/engine/core-modules/applicationVariable/application-variable.entity';
 import {
@@ -36,16 +36,10 @@ export class ApplicationVariableEntityService {
       return applicationVariable.value;
     }
 
-    const decryptedValue = this.secretEncryptionService.decrypt(
-      applicationVariable.value,
-    );
-
-    const visibleCharsCount = Math.min(
-      5,
-      Math.floor(decryptedValue.length / 10),
-    );
-
-    return `${decryptedValue.slice(0, visibleCharsCount)}${SECRET_APPLICATION_VARIABLE_MASK}`;
+    return this.secretEncryptionService.decryptAndMask({
+      value: applicationVariable.value,
+      mask: SECRET_APPLICATION_VARIABLE_MASK,
+    });
   }
 
   async update({
@@ -122,6 +116,7 @@ export class ApplicationVariableEntityService {
             applicationId,
           },
           {
+            value: encryptedValue,
             description: description ?? '',
             isSecret: isSecretValue,
           },
--- a/packages/twenty-server/src/engine/core-modules/applicationVariable/application-variable.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/applicationVariable/application-variable.service.ts
@@ -6,6 +6,12 @@ import { In, Not, Repository } from 'typeorm';
 import { ApplicationVariables } from 'twenty-shared/application';
 
 import { ApplicationVariableEntity } from 'src/engine/core-modules/applicationVariable/application-variable.entity';
+import {
+  ApplicationVariableEntityException,
+  ApplicationVariableEntityExceptionCode,
+} from 'src/engine/core-modules/applicationVariable/application-variable.exception';
+import { SECRET_APPLICATION_VARIABLE_MASK } from 'src/engine/core-modules/applicationVariable/constants/secret-application-variable-mask.constant';
+import { SecretEncryptionService } from 'src/engine/core-modules/secret-encryption/secret-encryption.service';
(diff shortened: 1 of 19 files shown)


packages/twenty-server/src/engine/core-modules/twenty-config/storage/config-storage.service.ts
```diff
@@ -77,21 +69,9 @@ export class ConfigStorageService implements ConfigStorageInterface {
...
-      } catch (error) {
-        this.logger.debug(
-          `${isDecrypt ? 'Decryption' : 'Encryption'} failed for key ${
-            key as string
-          }: ${error.message}. Using original value.`,
-        );
-
-        return convertedValue;
-      }
+      return isDecrypt
```
ijreilly, line 72: I dont understand why we had that try / catch behaviour before, swallowing the encryption/decryption error with a fallback which seems problematic to me - if we mean to encrypt a value but leave it decrypted, we will have errors when trying to decrypt again, won't we ? And getting the encrypted value when we try to fetch the decrypted value does not help either - if i understood correctly whats happening

packages/twenty-server/src/engine/core-modules/secret-encryption/secret-encryption.service.ts
```diff
@@ -0,0 +1,58 @@
...
+    if (!isDefined(value)) {
+      return value;
+    }
+
+    try {
+      const appSecret = this.getAppSecret();
+
+      return encryptText(value, appSecret);
+    } catch (error) {
+      this.logger.debug(
```
Weiko: I think we should not swallow here as this is not expected to happen. If this fails we will save non-encrypted value and we won't know (this should not happen but still)

packages/twenty-server/src/engine/core-modules/secret-encryption/secret-encryption.service.ts
```diff
@@ -0,0 +1,58 @@
...
+    if (!isDefined(value)) {
+      return value;
+    }
+
+    try {
+      const appSecret = this.getAppSecret();
+
+      return decryptText(value, appSecret);
+    } catch (error) {
+      this.logger.debug(
```
Weiko: same

packages/twenty-server/src/engine/core-modules/secret-encryption/secret-encryption.service.ts
```diff
@@ -0,0 +1,58 @@
...
+import { isDefined } from 'twenty-shared/utils';
+
+import {
+  decryptText,
+  encryptText,
+} from 'src/engine/core-modules/auth/auth.util';
+import { EnvironmentConfigDriver } from 'src/engine/core-modules/twenty-config/drivers/environment-config.driver';
+
+@Injectable()
+export class SecretEncryptionService {
```
Weiko, line 12: I know I originally introduced this service in my PR but I'm wondering if it's bringing much value 🤔

packages/twenty-server/src/engine/core-modules/applicationVariable/application-variable.service.ts
```diff
@@ -14,21 +20,64 @@ export class ApplicationVariableEntityService {
...
+
+    return this.secretEncryptionService.encrypt(value);
+  }
+
+  getDisplayValue(applicationVariable: ApplicationVariableEntity): string {
+    if (!applicationVariable.isSecret) {
+      return applicationVariable.value;
+    }
+
+    const decryptedValue = this.secretEncryptionService.decrypt(
```
Weiko: Note: Should we introduce a decryptAndMask method? Separating this concern into 2 methods might let a variable holding the full decrypted value outside secretEncryptionService and could cause errors later (if someone decide to do something else without / log / etc)

Weiko:
Seems we removed the save button and it saves on blur, regardless of that we should make sure that if tmr we add more editable stuff in the application settings, we don't send the "encrypted" value that is returned by the backend. Can we flag that somewhere?

Screenshot 2026-01-27 at 12 00 00

packages/twenty-server/src/engine/core-modules/secret-encryption/secret-encryption.service.ts
Weiko: Open to discussion

Weiko (commented):
Left some comments but overall LGTM! @ijreilly

packages/twenty-server/src/engine/core-modules/secret-encryption/secret-encryption.service.ts
ijreilly: totally agree, i have changed the behavior in config-storage (see my comment - would like some validation on this btw :)) but did not think to do here too

packages/twenty-server/src/engine/core-modules/secret-encryption/secret-encryption.service.ts
ijreilly: kept it and added decryptAndMask in it

Weiko (approved):
LGTM

packages/twenty-server/src/engine/core-modules/secret-encryption/secret-encryption.service.ts
Weiko: Ah yes perfect 👍 @ijreilly

PR #17396, opened 2026-01-23
[TYPES] `UniversalEntity` `JsonbProperty` and `SerializedRelation`

prastoin:
# Introduction

In this PR we're introducing mainly two branded type signatures for both `JsonbProperty` entities properties and `SerializedRelation` (jsonb serialized property storing another entity id).

Allowing to dynamically map over them later in order to build universal `jsonb` `serialized` relations.

## `JsonbProperty`

A branded wrapper type that marks entity properties stored as PostgreSQL JSONB columns. It adds a phantom brand `__JsonbPropertyBrand__` to object types while leaving primitives unchanged. The branded key is optional and typed as never, also omitted when transpiled to `UniversalFlat`

**Should be used at entities lvl only:**
```typescript
@Column({ type: 'jsonb', nullable: false })
gridPosition: JsonbProperty<GridPosition>;

@Column({ nullable: false, type: 'jsonb', default: [] })
publishedVersions: JsonbProperty<string[]>;
```

## `SerializedRelation`

A branded string type that marks foreign key IDs stored inside JSONB objects. These are entity references serialized within a JSONB column rather than being a regular database foreign key.

**Usage in jsonb property generic***
```ts
type FieldMetadataRelationSettings = {
  relationType: RelationType;
  onDelete?: RelationOnDeleteAction;
  joinColumnName?: string | null;
  junctionTargetFieldId?: SerializedRelation;
};
```

## `FormatJsonbSerializedRelation<T>`

A transformation type that processes JSONB properties for universal entity mapping. It:
1. Detects properties with the `JsonbProperty` brand
2. Finds `SerializedRelation` properties
3. Renames them from `*Id` to `*UniversalIdentifier`
4. Removes the brand from the output type ( optional though )

```typescript
// Input: JsonbProperty<{ targetFieldMetadataId: SerializedRelation }>
// Output: { targetFieldMetadataUniversalIdentifier: SerializedRelation }
```

## Result
An example of the dynamic type mapping, through a type-test example
```ts
type SettingsTestCase = UniversalFlatFieldMetadata<
    | FieldMetadataType.RELATION
    | FieldMetadataType.NUMBER
    | FieldMetadataType.TEXT
  >['settings']

type SettingsExpectedResult =
  | {
      relationType: RelationType;
      onDelete?: RelationOnDeleteAction | undefined;
      joinColumnName?: string | null | undefined;
      junctionTargetFieldUniversalIdentifier?: SerializedRelation | undefined;
    }
  | {
      dataType?: NumberDataType | undefined;
      decimals?: number | undefined;
      type?: FieldNumberVariant | undefined;
    }
  | {
      displayedMaxRows?: number | undefined;
    }
  | null;

type Assertions = [
  Expect<Equal<SettingsTestCase, SettingsExpectedResult>>,
]
```

## Remarks

- Removed duplicated twenty-server and twenty-shared typed
- Removed class validator instances for default value that were not used at runtime, we will refactor that to add validation across all entities following a same pattern


packages/twenty-shared/src/types/ExtractSerializedRelationProperties.type.ts
```diff
@@ -0,0 +1,13 @@
+import { SERIALIZED_RELATION_BRAND } from './SerializedRelation.type';
+
+export type ExtractSerializedRelationProperties<T> = T extends unknown
+  ? NonNullable<
+      {
+        [P in keyof T]-?: [NonNullable<T[P]>] extends [never]
+          ? never
+          : typeof SERIALIZED_RELATION_BRAND extends keyof NonNullable<T[P]>
```
prastoin: create abstraction

packages/twenty-shared/src/types/ExtractSerializedRelationProperties.type.ts
```diff
@@ -0,0 +1,13 @@
+import { SERIALIZED_RELATION_BRAND } from './SerializedRelation.type';
+
+export type ExtractSerializedRelationProperties<T> = T extends unknown
+  ? NonNullable<
+      {
```
prastoin: might require object to descope

packages/twenty-shared/src/types/FieldMetadataDefaultValue.ts
```diff
@@ -46,266 +11,107 @@ export const fieldMetadataDefaultValueFunctionName = {
...
+  primaryPhoneNumber: string | null;
+  primaryPhoneCountryCode: string | null;
+  primaryPhoneCallingCode: string | null;
+  additionalPhones: object | null;
+};
 
-export type FieldMetadataFunctionDefaultValue = ExtractValueType<
-  FieldMetadataDefaultValueUuidFunction | FieldMetadataDefaultValueNowFunction
->;
+export type FieldMetadataDefaultValueMapping = {
```
prastoin, line 68: Could factorize to have only one JsonbProperty call or at least reduce as we handle distributive type map

prastoin: In the best of the world should even be at entity lvl

packages/twenty-shared/src/types/RowLevelPermissionPredicateValue.ts
```diff
@@ -12,5 +12,4 @@ export type RowLevelPermissionPredicateValue =
   | number
   | RelationPredicateValue
   | Record<string, unknown>
-  | null
-  | undefined;
+  | null;
```
prastoin, line 15: Note: discussed with coco, was an error to set undefined

packages/twenty-server/src/engine/metadata-modules/flat-entity/types/extract-entity-one-to-many-entity-relation-properties.type.ts
```diff
@@ -9,7 +7,7 @@ export type ExtractEntityOneToManyEntityRelationProperties<
 > = NonNullable<
   {
     [P in keyof T]: NonNullable<T[P]> extends Array<infer U>
-      ? U extends Relation<TTarget>
+      ? U extends TTarget
```
prastoin, line 10: ***Note:*** `typeorm` Relation was not necessary at all to make a discriminated union here
In the end in order to extract relation we use `WorkspaceRelationEntity` or other potential discrimator basis

.cursor/rules/creating-syncable-entity.mdc
```diff
@@ -237,6 +238,160 @@ export abstract class WorkspaceRelatedEntity {
...
+```typescript
+import { JsonbProperty } from 'src/engine/workspace-manager/workspace-migration/universal-flat-entity/types/jsonb-property.type';
+
+@Entity('myEntity')
+export class MyEntityEntity extends SyncableEntity {
+  // Simple JSONB column - wrap the type
+  @Column({ type: 'jsonb', nullable: true })
+  settings: JsonbProperty<MyEntitySettings> | null;
+
+  // JSONB column with complex type
```
charlesBochet, line 258: what's the difference with previous case?

packages/twenty-shared/src/types/SerializedRelation.type.ts
```diff
@@ -0,0 +1,5 @@
+export const SERIALIZED_RELATION_BRAND = '__SerializedRelationBrand__' as const;
+
+export type SerializedRelation = string & {
```
charlesBochet, line 3: I don't really understand this one and especially the never

charlesBochet (approved):

packages/twenty-shared/src/types/SerializedRelation.type.ts
prastoin: Discussed during a call

.cursor/rules/creating-syncable-entity.mdc
prastoin: Discussed during a call

PR #17400, opened 2026-01-23
Files v2 - FILES field update/create in common API

etiennejouan:
## FILES Field update interface

```
mutation {
  updateCompany(
    id: "company-uuid"
    data: {
      filesfield: [
          { fileId: "new-file-uuid", label: "new-document.pdf" }
        ]
    }
  ) {
    id
    filesfield {
      fileId
      label
      extension
    }
  }
}
```


packages/twenty-server/src/engine/api/common/common-query-runners/common-update-many-query-runner.service.ts
```diff
@@ -68,6 +78,50 @@ export class CommonUpdateManyQueryRunnerService extends CommonBaseQueryRunnerSer
...
+        .getMany()) as (ObjectRecord & { id: string })[];
+
+      if (existingRecords.length > 1) {
+        throw new CommonQueryRunnerException(
+          'File can be associated with only one record',
+          CommonQueryRunnerExceptionCode.INVALID_ARGS_DATA,
+          {
+            userFriendlyMessage: msg`File can be associated with only one record`,
+          },
+        );
```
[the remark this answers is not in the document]
etiennejouan: a file can be linked to only one record

packages/twenty-server/src/engine/api/common/common-query-runners/common-create-many-query-runner/common-create-many-query-runner.service.ts
```diff
@@ -99,6 +107,13 @@ export class CommonCreateManyQueryRunnerService extends CommonBaseQueryRunnerSer
       rolePermissionConfig,
     });
 
+    await this.processFilesFieldSyncOperationsIfNeeded({
+      originalData,
+      flatObjectMetadata,
+      flatFieldMetadataMaps,
+      workspace: authContext.workspace,
+    });
```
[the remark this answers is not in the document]
etiennejouan: True ! Need to discuss it

packages/twenty-server/src/engine/api/graphql/workspace-schema-builder/services/type-mapper.service.ts
```diff
@@ -105,13 +108,43 @@ export class TypeMapperService {
     return this.baseTypeScalarMapping.get(fieldMetadataType);
   }
 
-  mapToPreBuiltGraphQLInputType({
+  mapToPreBuiltGraphQLCreateInputType({
```
Weiko: mapToPreBuiltGraphQLCreateInputType and mapToPreBuiltGraphQLUpdateInputType seem really similar

Weiko: mapToPreBuiltGraphQLCreateInputType and mapToPreBuiltGraphQLUpdateInputType seem really similar. Should we refactor this?

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/services/files-field-sync.service.ts
```diff
@@ -0,0 +1,533 @@
...
+      throw new CommonQueryRunnerException(
+        `File should be a temporary files field file: ${fileId}`,
+        CommonQueryRunnerExceptionCode.INVALID_ARGS_DATA,
+        {
+          userFriendlyMessage: msg`${fileId} file is already associated with a permanent files field. Re-upload the file.`,
+        },
+      );
+    }
+
+    (toAddFileItem as FileItemOutput).extension = path.extname(fileEntity.path);
```
Weiko: You are losing data here I think
For example
```sh
> path.extname("test.tar.gz")
'.gz'
```
instead of tar.gz

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/services/files-field-sync.service.ts
```diff
@@ -0,0 +1,533 @@
...
+
+    if (
+      toAddFileIds.length === 0 &&
+      toUpdateFileIds.length === 0 &&
+      toRemoveFileIds.length === 0
+    ) {
+      return;
+    }
+
+    if ([...new Set(toAddFileIds)].length !== toAddFileIds.length) {
```
Weiko: Should we check for update as well?

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/services/files-field-sync.service.ts
```diff
@@ -0,0 +1,533 @@
...
+      flatFieldMetadataMaps,
+    );
+
+    if (filesFields.length === 0) {
+      return;
+    }
+
+    const operations: Promise<void>[] = [];
+
+    for (const record of data) {
```
Weiko: We might have perf issues with large dataset. We could batch those steps

packages/twenty-server/src/engine/api/common/common-query-runners/common-create-many-query-runner/common-create-many-query-runner.service.ts
```diff
@@ -61,6 +67,8 @@ export class CommonCreateManyQueryRunnerService extends CommonBaseQueryRunnerSer
       );
     }
 
+    const originalData = structuredClone(args.data);
```
Weiko: deep copy might be intensive with large dataset. You are doing this for the whole object, can we instead extract only the necessary data for files and create a new object with only that info?

Weiko (commented):
As discussed, not fully sure about the API where we delegate diff to the consumer but I can understand, not a strong opinion on that.
I've left some comments. For your question about failure, indeed you might leave things in inconsistent state, this could be solved with a "transaction" pattern (but might be a bit heavy to implement with file operations)
Note: I'm assuming multiple concurrent requests could be manipulating the same files simultaneously and you might face some race conditions? Maybe we want some locking mechanism for that?
Note2: FilesFieldSyncService is doing a lot of things, maybe we can split a bit things like validation etc? Up to you, don't do it if it's not relevant or too complex

Nice one @etiennejouan ! 👏

charlesBochet:
@etiennejouan 

I'm sorry about this one, I pushed you in a wrong direction. I think we cannot update files without checking the value of the record, therefore validation must happen in the ORM layer. we should do something like you've done for connect.

and if we do that we can simplify the API

packages/twenty-server/src/engine/api/common/common-query-runners/common-update-many-query-runner.service.ts
```diff
@@ -33,6 +33,10 @@ export class CommonUpdateManyQueryRunnerService extends CommonBaseQueryRunnerSer
   UpdateManyQueryArgs,
   ObjectRecord[]
 > {
+  constructor() {
```
Weiko: Why this change?

packages/twenty-server/src/engine/core-modules/file-storage/file-storage.service.ts
```diff
@@ -126,6 +130,32 @@ export class FileStorageService {
...
+  async deleteByFileId({
+    fileId,
+    workspaceId,
+    fileFolder,
+  }: {
+    fileId: string;
+    workspaceId: string;
+    fileFolder: FileFolder;
+  }): Promise<void> {
+    const file = await this.fileRepository.findOneOrFail({
```
Weiko, line 152: This method will throw a typeorm exception. Do we want our own exception here? Will it be user facing?

packages/twenty-server/src/engine/core-modules/file/types/file-settings.types.ts
```diff
@@ -0,0 +1,6 @@
+export type FileFieldSettings = {
```
Weiko, line 1: Does it need to be in twenty-shared or we don't plan to expose this to the FE in the future?

Weiko: (if it is the plan but through gql, then it's fine because it will be in the schema so no need to put it in twenty-shared either)

packages/twenty-server/src/engine/twenty-orm/field-operations/files-field-sync/files-field-sync.ts
```diff
@@ -0,0 +1,603 @@
...
+    const toUpdate: FileItem[] = [];
+    const toRemove: FileItem[] = [];
+
+    for (const newFile of newFiles) {
+      const existingFile = existingFileMap.get(newFile.fileId);
+
+      if (!isDefined(existingFile)) {
+        toAdd.push(newFile);
+      } else {
+        toUpdate.push(newFile);
```
Weiko: Are we also pushing if only the label changed? What do we want to do if that's the case? (probably a lighter process?)

packages/twenty-server/src/engine/twenty-orm/field-operations/files-field-sync/files-field-sync.ts
```diff
@@ -0,0 +1,603 @@
...
+          throw new TwentyORMException(
+            `Application ID not found for file ${fileId}`,
+            TwentyORMExceptionCode.INVALID_INPUT,
+          );
+        }
+
+        await this.fileRepository
+          .createQueryBuilder()
+          .update()
+          .set({
```
Weiko: I think typeorm handles that format as well? Can you confirm?
```typescript
    .set({
      settings: {
        isTemporaryFile: false,
        toDelete: false,
      },
      applicationId,
    })
```
This would be better if that's true (type safe, real boolean instead of 'false', easier to read overall)

packages/twenty-server/src/engine/twenty-orm/field-operations/files-field-sync/files-field-sync.ts
```diff
@@ -0,0 +1,603 @@
...
+            settings: () => `jsonb_set(settings, '{isTemporaryFile}', 'false')`,
+            applicationId,
+          })
+          .where('id = :id', { id: fileId })
```
Weiko: Can we do a batch update instead? Here we are doing 1 update per file, this seems avoidable if I'm not mistaken

Weiko (commented):
Left some comments. The API looks better!
We still have some potential race conditions / side-effects that we don't handle properly but this should be fine for this version as discussed earlier

packages/twenty-server/src/engine/api/common/common-query-runners/common-update-many-query-runner.service.ts
etiennejouan: removed

packages/twenty-server/src/engine/core-modules/file-storage/file-storage.service.ts
etiennejouan: For the moment it's in a job, but I create a custom exception for the fileField service using it

packages/twenty-server/src/engine/core-modules/file/types/file-settings.types.ts
etiennejouan: Don't know for the moment, I think we can move it when needed

packages/twenty-server/src/engine/twenty-orm/field-operations/files-field-sync/files-field-sync.ts
etiennejouan: we are pushing in all case.
The more logic would be to append label in existing file item (and keep existing file item if nothing is updated) but not sure it is lighter.

packages/twenty-server/src/engine/twenty-orm/field-operations/files-field-sync/files-field-sync.ts
etiennejouan: I like 
```          .set({
            settings: () => `jsonb_set(settings, '{isTemporaryFile}', 'false')`,
```
it enables to only update the isTemporaryFile key in jsonb column. But it's true, we don't have other prop

packages/twenty-server/src/engine/twenty-orm/field-operations/files-field-sync/files-field-sync.ts
etiennejouan: Yes good point, it needs to be done by applicationId

Weiko (approved):
LGTM! 👏 @etiennejouan

PR #17403, opened 2026-01-23
fix(twenty-server): add SSRF protection to webhook requests

FelixMalfait:
## Summary

- Adds SSRF (Server-Side Request Forgery) protection to webhook requests by using the same secure axios adapter already used by HTTP workflow actions
- Prevents webhooks from making requests to private/internal IP addresses (10.x, 192.168.x, 172.16-31.x, 169.254.x, localhost)
- Adds specific error logging when a webhook fails due to SSRF protection

## Context

The HTTP workflow tool (`HTTP_REQUEST` action) already had SSRF protection via `HTTP_TOOL_SAFE_MODE_ENABLED`, but webhooks were using `HttpService` directly without this protection. This inconsistency meant users could potentially configure webhooks to probe internal infrastructure.

### What's protected now:

| Feature | Before | After |
|---------|--------|-------|
| HTTP Workflow Action | Protected (secure adapter) | Protected (secure adapter) |
| Webhooks | **Unprotected** | Protected (secure adapter) |

### The secure adapter validates:
1. Protocol must be `http:` or `https:`
2. DNS resolution of hostname
3. Resolved IP must not be in private ranges

## Test plan

- [ ] Configure a webhook with an external URL (e.g., `https://webhook.site`) - should work
- [ ] Configure a webhook with `http://localhost:3000` - should fail with SSRF error in audit log
- [ ] Configure a webhook with `http://10.0.0.1/test` - should fail with SSRF error in audit log
- [ ] Configure a webhook with a domain that resolves to a private IP - should fail

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/tool/utils/__tests__/get-secure-axios-adapter.util.spec.ts
+++ b/packages/twenty-server/src/engine/core-modules/tool/utils/__tests__/get-secure-axios-adapter.util.spec.ts
@@ -1,9 +1,10 @@
+import * as http from 'http';
+import * as https from 'https';
+
 import { AxiosHeaders, type InternalAxiosRequestConfig } from 'axios';
 
-import {
-  getSecureAdapter,
-  type SecureAdapterDependencies,
-} from 'src/engine/core-modules/tool/utils/get-secure-axios-adapter.util';
+import { type SecureAdapterDependencies } from 'src/engine/core-modules/tool/utils/get-secure-axios-adapter.types';
+import { getSecureAdapter } from 'src/engine/core-modules/tool/utils/get-secure-axios-adapter.util';
 
 describe('getSecureAdapter', () => {
   let mockDnsLookup: jest.Mock;
@@ -160,7 +161,7 @@ describe('getSecureAdapter', () => {
   });
 
   describe('DNS rebinding protection', () => {
-    it('should replace hostname with resolved IP in the request URL', async () => {
+    it('should preserve original hostname in URL for TLS validation', async () => {
       mockDnsLookup.mockResolvedValue({
         address: '93.184.216.34',
         family: 4,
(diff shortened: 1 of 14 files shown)


packages/twenty-server/src/engine/core-modules/tool/utils/get-secure-axios-adapter.util.ts
```diff
@@ -1,14 +1,28 @@
...
+  AxiosHeaders,
   type AxiosAdapter,
   type InternalAxiosRequestConfig,
 } from 'axios';
 
 import { isPrivateIp } from 'src/engine/core-modules/tool/utils/is-private-ip.util';
-const httpAdapter = axios.getAdapter('http');
 
-export const getSecureAdapter = (): AxiosAdapter => {
+export type SecureAdapterDependencies = {
```
charlesBochet: exported types should be in its own file

charlesBochet: if not exported it can stay here

packages/twenty-server/src/engine/core-modules/twenty-config/config-variables.ts
```diff
@@ -69,10 +69,12 @@ export class ConfigVariables {
   @IsOptional()
   IS_EMAIL_VERIFICATION_REQUIRED = false;
 
+  // TODO: Consider renaming to OUTBOUND_HTTP_SAFE_MODE_ENABLED as this now
```
charlesBochet: agree let's do it now :)

packages/twenty-server/src/engine/core-modules/webhook/jobs/call-webhook.job.ts
```diff
@@ -84,7 +85,15 @@ export class CallWebhookJob {
           .toString('hex');
       }
 
-      const response = await this.httpService.axiosRef.post(
+      const isSafeModeEnabled = this.twentyConfigService.get(
+        'HTTP_TOOL_SAFE_MODE_ENABLED',
+      );
+
+      const axiosClient = isSafeModeEnabled
```
charlesBochet: LGTM but would even be better to expose our own HTTPAxiosClient that would be a service checking HTTP_TOOL_SAFE_MODE_ENABLED and returning the right client directly. would make sure upcoming  implementtions are not breaking the security

packages/twenty-server/test/integration/metadata/suites/developers/webhooks.integration-spec.ts
```diff
@@ -308,4 +404,147 @@ describe('webhooksResolver (e2e)', () => {
...
   });
+
+  describe('webhook delivery', () => {
+    const WEBHOOK_RECEIVER_PORT = 4317;
+
+    it('should block delivery to private IP when safe mode is enabled (SSRF protection)', async () => {
+      const receiver = await createWebhookReceiver(WEBHOOK_RECEIVER_PORT);
+
+      try {
+        const createWebhookResponse = await makeMetadataAPIRequest({
```
charlesBochet: let's create tooling for that in integration tests so it's easy to interact with webhooks in integ tests.
we have example for other metadata endpoints

packages/twenty-server/test/integration/metadata/suites/developers/webhooks.integration-spec.ts
```diff
@@ -1,10 +1,106 @@
...
+      id
+      name {
+        firstName
+        lastName
+      }
+    }
+  }
+`;
+
+// Helper to create a simple HTTP server for testing webhook delivery
```
charlesBochet: not useful :)

packages/twenty-server/src/engine/core-modules/tool/utils/__tests__/get-secure-axios-adapter.util.spec.ts
```diff
@@ -0,0 +1,371 @@
...
+      const adapter = getSecureAdapter(dependencies);
+      const config = {
+        url: 'https://example.com',
+        headers: { 'Content-Type': 'application/json' },
+      } as unknown as InternalAxiosRequestConfig;
+
+      await adapter(config);
+
+      expect(mockHttpAdapter).toHaveBeenCalled();
+      expect(config.headers['Host']).toBe('example.com');
```
charlesBochet: this file is a bit long, still OK but let's split in case you add more tests

charlesBochet (approved):
Approved with some comments (most nitpicks)

PR #17407, opened 2026-01-23
Change `formatResult` to return string instead of `Date` object for `DATE_TIME`

lucasbordeau:
This PR modifies our broadly used `formatResult` util to counter act TypeORM transforming any date time to a `Date` object.

Instead we return the ISO string for any `DATE_TIME`, this way we're not transporting Date object from one function to another in the backend. 

We do this because there was problems working with events utils that take string date time in parameters and received Date objects. 

As this is a recurring problem and because it's an opinionated choice from TypeORM, we chose to switch to string only in our codebase, from TypeORM's output to frontend.


Weiko (approved):
LGTM
Hopefully we are not expecting those being Date objects somewhere in the codebase 🤞 (I'm pretty sure we don't)

PR #17410, opened 2026-01-24
Do not delete archived messages

BOHEUS:
Related to https://github.com/twentyhq/core-team-issues/issues/1125

Only backend part as there are no mocks for frontend or I haven't found them


charlesBochet:
@BOHEUS this does not seem ready from a product perspective.
we would need to check this from a product perspective. I think it's tied to the whole problematic to what's happening when a userWorkspace get deleted:
- should we actually keep the workspceMember but hide it?, etc.

It's not only about messages

I'm going to close it, we can keep discussing it here or on the original issue but we will need input from product

BOHEUS:
@charlesBochet no problem, please let me know if my approach was correct, I'll wait for response from product to tackle this problem

PR #17412, opened 2026-01-24
fix(twenty-server): preserve input order in createMany response

FelixMalfait:
## Summary
- The `createMany` API was returning records in arbitrary order because `fetchUpsertedRecords` used `WHERE id IN (...)` without `ORDER BY`
- This caused flaky tests that assumed input order was preserved (e.g., `people-merge-many.integration-spec.ts`)
- Fixed by reordering the fetched records to match the original input order from `generatedMaps`

## Root Cause
```typescript
// Before: No ORDER BY, so SQL returns in arbitrary order
const upsertedRecords = await queryBuilder
  .where({ id: In(objectRecords.generatedMaps.map((record) => record.id)) })
  .getMany();  // Returns in arbitrary order!
```

## Fix
```typescript
// After: Reorder results to match original input order
const orderedIds = objectRecords.generatedMaps.map((record) => record.id);
const upsertedRecords = await queryBuilder
  .where({ id: In(orderedIds) })
  .getMany();

// Preserve original input order
const recordsById = new Map(upsertedRecords.map((record) => [record.id, record]));
return orderedIds
  .map((id) => recordsById.get(id))
  .filter((record) => record !== undefined);
```

## Test plan
- [x] Run `people-merge-many.integration-spec.ts` multiple times - passes consistently
- [x] Lint passes

### DIFF
--- a/packages/twenty-server/src/engine/api/common/common-query-runners/common-create-many-query-runner/common-create-many-query-runner.service.ts
+++ b/packages/twenty-server/src/engine/api/common/common-query-runners/common-create-many-query-runner/common-create-many-query-runner.service.ts
@@ -419,18 +419,33 @@ export class CommonCreateManyQueryRunnerService extends CommonBaseQueryRunnerSer
       flatFieldMetadataMaps,
     });
 
+    const orderedIds = objectRecords.generatedMaps.map((record) => record.id);
+
     const upsertedRecords = await queryBuilder
       .setFindOptions({
         select: columnsToSelect,
       })
       .where({
-        id: In(objectRecords.generatedMaps.map((record) => record.id)),
+        id: In(orderedIds),
       })
       .withDeleted()
       .take(QUERY_MAX_RECORDS)
       .getMany();
 
-    return upsertedRecords as ObjectRecord[];
+    // Preserve original input order by sorting results to match orderedIds
+    const recordsById = new Map<string, ObjectRecord>(
+      upsertedRecords.map((record) => [record.id, record as ObjectRecord]),
+    );
+
+    return orderedIds.map((id) => {
+      const record = recordsById.get(id);
+
+      if (!record) {
+        throw new Error(`Record with id ${id} not found after upsert`);
+      }
+
+      return record;
+    });
   }
 
   async processQueryResult(
--- a/packages/twenty-server/src/engine/api/common/common-query-runners/common-find-duplicates-query-runner.service.ts
+++ b/packages/twenty-server/src/engine/api/common/common-query-runners/common-find-duplicates-query-runner.service.ts
@@ -71,12 +71,21 @@ export class CommonFindDuplicatesQueryRunnerService extends CommonBaseQueryRunne
     });
 
     if (isDefined(args.ids) && args.ids.length > 0) {
-      objectRecords = (await existingRecordsQueryBuilder
+      const fetchedRecords = (await existingRecordsQueryBuilder
         .where({ id: In(args.ids) })
         .setFindOptions({
           select: columnsToSelect,
         })
         .getMany()) as ObjectRecord[];
+
+      // Preserve original input order
+      const recordsById = new Map(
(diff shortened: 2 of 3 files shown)


packages/twenty-server/src/engine/api/common/common-query-runners/common-create-many-query-runner/common-create-many-query-runner.service.ts
```diff
@@ -419,18 +419,33 @@ export class CommonCreateManyQueryRunnerService extends CommonBaseQueryRunnerSer
...
-        id: In(objectRecords.generatedMaps.map((record) => record.id)),
+        id: In(orderedIds),
       })
       .withDeleted()
       .take(QUERY_MAX_RECORDS)
       .getMany();
 
-    return upsertedRecords as ObjectRecord[];
+    // Preserve original input order by sorting results to match orderedIds
+    const recordsById = new Map<string, ObjectRecord>(
```
charlesBochet: why not use sort method?

packages/twenty-server/src/engine/api/common/common-query-runners/common-create-many-query-runner/common-create-many-query-runner.service.ts
```diff
@@ -419,18 +419,33 @@ export class CommonCreateManyQueryRunnerService extends CommonBaseQueryRunnerSer
       flatFieldMetadataMaps,
     });
 
+    const orderedIds = objectRecords.generatedMaps.map((record) => record.id);
```
charlesBochet, line 422: i'm confused on how this will preserver the order in the input: objectRecords does not seem to be the input. I wonder if the sort method should be done in a parent method that stil has access to the input

packages/twenty-server/src/engine/api/common/common-query-runners/common-find-duplicates-query-runner.service.ts
```diff
@@ -71,12 +71,21 @@ export class CommonFindDuplicatesQueryRunnerService extends CommonBaseQueryRunne
...
         })
         .getMany()) as ObjectRecord[];
+
+      // Preserve original input order
+      const recordsById = new Map(
+        fetchedRecords.map((record) => [record.id, record]),
+      );
+
+      objectRecords = args.ids
+        .map((id) => recordsById.get(id))
```
charlesBochet: here it makes sense

packages/twenty-server/src/engine/api/common/common-query-runners/common-find-duplicates-query-runner.service.ts
```diff
@@ -71,12 +71,21 @@ export class CommonFindDuplicatesQueryRunnerService extends CommonBaseQueryRunne
...
     if (isDefined(args.ids) && args.ids.length > 0) {
-      objectRecords = (await existingRecordsQueryBuilder
+      const fetchedRecords = (await existingRecordsQueryBuilder
         .where({ id: In(args.ids) })
         .setFindOptions({
           select: columnsToSelect,
         })
         .getMany()) as ObjectRecord[];
+
+      // Preserve original input order
```
charlesBochet: not useful :)

charlesBochet (approved):
left comments :)

packages/twenty-server/src/engine/api/common/common-query-runners/common-create-many-query-runner/common-create-many-query-runner.service.ts
FelixMalfait: For simple inserts, generatedMaps does preserve input order (TypeORM guarantee), so the current fix works.  For upserts, you're right - the order is scrambled because updates are processed before inserts. Fixing that properly would require a bigger change, I don't think it's worth it. I will look into it but will likely merge like that if I find out it's too much code changes just for this upsert edge-case

PR #17416, opened 2026-01-24
[Fix] Bug with one to many update in table #16340

### LINKED ISSUE #16340: Bug with one to many update in table
https://github.com/user-attachments/assets/65a58bd5-fe47-4a9c-9193-06305fd79021

carbonFibreCode:
fixes #16340 

we are updating the cache partially to prevent the flow of the corrupted fields into the cache 

the fields get corrupted during the mutation process potentially overriding the recoil cache as we are passing the `newRecordCache` directly in `upsertRecordsInStore`, the newRecordCache is thin and hence the recoil wipes out the fields that are undefined or empty

we prevent this by extracting the partial data ( only the data which is being updated in the field ) and passing  it to `upsertRecordsInStore`, as it only touched the specific fields and updates the data, leaving the other fields untouched.

https://github.com/user-attachments/assets/5256bef7-70c3-47b3-b2ce-dd02ee1a2de8

### DIFF
--- a/packages/twenty-front/src/modules/apollo/optimistic-effect/utils/triggerAttachRelationOptimisticEffect.ts
+++ b/packages/twenty-front/src/modules/apollo/optimistic-effect/utils/triggerAttachRelationOptimisticEffect.ts
@@ -123,10 +123,12 @@ export const triggerAttachRelationOptimisticEffect = ({
     return;
   }
 
+  const fieldNameOnTargetRecordValue = newCachedRecord[fieldNameOnTargetRecord];
+
   const partialRecordForStore = {
     id: newCachedRecord.id,
     __typename: newCachedRecord.__typename,
-    [fieldNameOnTargetRecord]: newCachedRecord[fieldNameOnTargetRecord],
+    [fieldNameOnTargetRecord]: fieldNameOnTargetRecordValue,
   } as ObjectRecord;
 
   upsertRecordsInStore({ partialRecords: [partialRecordForStore] });
--- a/packages/twenty-front/src/modules/apollo/optimistic-effect/utils/triggerAttachRelationOptimisticEffect.ts
+++ b/packages/twenty-front/src/modules/apollo/optimistic-effect/utils/triggerAttachRelationOptimisticEffect.ts
@@ -123,5 +123,11 @@ export const triggerAttachRelationOptimisticEffect = ({
     return;
   }
 
-  upsertRecordsInStore({ partialRecords: [newCachedRecord] });
+  const partialRecordForStore = {
+    id: newCachedRecord.id,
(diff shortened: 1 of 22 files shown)


lucasbordeau:
Hi @carbonFibreCode, I've been testing your solution but it felt like it missed the point because it did not fix the root cause but only avoided the side effect of the root cause, it could be enough for a fix, but in this part of the app, it is better to fix the root cause because the side effects could arise later and elsewhere as it is a central part.

The root cause in this case was related to the special treatment we have for activities, which have a junction activityTarget table, ie `noteTargets` or `taskTargets`. 

We already fixed a similar issue related to the computation of the gql fields of a note or task for regular find requests, in order to optimize our queries. 

Here we have the opposite problem, we miss the label identifier fields of a note or task, so it overwrites the note or task in the record store with the id field only.

So the root cause here is that we had to fix the generation of the record gql fields, inside `getRecordFromCache`, in order to return the required fields for computing the chip.

Here are both the Apollo cache read fragment before and after this fix, inside `getRecordFromCache` : 

```gql 
fragment PersonFragment on Person {
  __typename
  id
  noteTargets {
    edges {
      node {
        __typename
        id
      }
    }
  }
  taskTargets {
    edges {
      node {
        __typename
        id
      }
    }
  }
  // Other fields ...
}
```

```gql
fragment PersonFragment on Person {
  __typename
  id
  noteTargets {
    edges {
      node {
        __typename
        id
        note {
          __typename
          id
          title
        }
      }
    }
  }
  taskTargets {
    edges {
      node {
        __typename
        id
        task {
          __typename
          id
          title
        }
      }
    }
  }
  // Other fields ...
}
```

lucasbordeau:
Also I wanted to keep your fix as it seems like a good idea to only update what's relevant, but it introduced regression while doing a QA on relation modification, notably detaching a relation from a card, so I think it's safer to let what's in the cache overwrite the record store and only add the relevant missing fields to not risk introducing regression.

carbonFibreCode:
> Also I wanted to keep your fix as it seems like a good idea to only update what's relevant, but it introduced regression while doing a QA on relation modification, notably detaching a relation from a card, so I think it's safer to let what's in the cache overwrite the record store and only add the relevant missing fields to not risk introducing regression.

Hey @lucasbordeau, thanks a lot for digging into this and fixing the root cause.
I was initially leaning towards the same approach of fetching the data along with activityTargetGqlFields as mentioned in the issue discussion, but the optimization felt better at that moment and I completely overlooked the potential regressions, especially given this is such a central part of the app.

Your explanation and the updated fragments make a lot of sense . I appreciate you taking the time to walk through the reasoning.

lucasbordeau (approved):

PR #17427, opened 2026-01-25
Improve email onboarding step

BOHEUS:
Related to https://github.com/twentyhq/core-team-issues/issues/1477

### DIFF
--- a/packages/twenty-front/src/pages/onboarding/SyncEmails.tsx
+++ b/packages/twenty-front/src/pages/onboarding/SyncEmails.tsx
@@ -9,6 +9,7 @@ import { Title } from '@/auth/components/Title';
 import { OnboardingSyncEmailsSettingsCard } from '@/onboarding/components/OnboardingSyncEmailsSettingsCard';
 import { useSetNextOnboardingStatus } from '@/onboarding/hooks/useSetNextOnboardingStatus';
 
+import { currentWorkspaceState } from '@/auth/states/currentWorkspaceState';
 import { isGoogleCalendarEnabledState } from '@/client-config/states/isGoogleCalendarEnabledState';
 import { isGoogleMessagingEnabledState } from '@/client-config/states/isGoogleMessagingEnabledState';
 import { isMicrosoftCalendarEnabledState } from '@/client-config/states/isMicrosoftCalendarEnabledState';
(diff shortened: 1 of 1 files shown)


packages/twenty-front/src/pages/onboarding/SyncEmails.tsx
```diff
@@ -78,6 +79,12 @@ export const SyncEmails = () => {
     setNextOnboardingStatus();
   };
 
+  const currentWorkspace = useRecoilValue(currentWorkspaceState);
+
+  const isSSOEnabled =
```
charlesBochet: @BOHEUS I'm not sure this is the desired behavior, I think what @Bonapara wants is to leverage the sign up method used by the user. However, we don't track this on backend yet and we don't want to do it. I think we are storing something (lastmethod used) in the localStorage (or in recoil). Let's try to leverage this instead

charlesBochet (changes requested):
Left a comment

packages/twenty-front/src/pages/onboarding/SyncEmails.tsx
BOHEUS: I'll wait for response from product to adjust the code if needed

Bonapara: We should rely on the information available on the front end if we have it, but otherwise display both options

BOHEUS: I'm not entirely sure if relying on lastAuthenticatedMethodState is correct solution as it relies on local storage from what I understand and most probably, when new user signs up, local storage is empty as they're using their own device and not shared one

Also, lastAuthenticatedMethodState is set when user signs out which in case of new user signing up will be always empty (or undefined)

Bonapara: isn't the local storage updated as soon as he clicks on a sign-up button?

BOHEUS: I focused solely on set(lastAuthenticatedMethodState, value) in useAuth hook completely forgetting that it's Recoil state :sweat_smile: 

I'll change it then

BOHEUS: Done

charlesBochet (approved):

PR #17431, opened 2026-01-25
Patch the postgres db url while using an external resource enabling a…

larron:
…n existing secret and password key


charlesBochet:
As this is community maintained, I'll trust you on this one @larron

charlesBochet (approved):
TY :)

PR #17438, opened 2026-01-26
`[OBJECT_CACHE_FLUSH_REQUIRED_WHEN_RELEASED]` Remove `FlatObjectMetadata` custom `fieldMetadataIds` fk aggregator property

prastoin:
# Introduction
Currently refactoring `flatEntity` typing, encountering some tsc errors due to this fk aggregator custom override
It shall now follow generic pattern leading to be named `fieldIds`

Needs to flush object cache when released


packages/twenty-shared/src/database-events/database-event-payload.type.ts
```diff
@@ -11,7 +11,7 @@ type SimplifiedFlatObjectMetadata = {
   universalIdentifier: string;
   applicationId: string | null;
 } & {
-  fieldMetadataIds: string[];
+  fieldIds: string[];
```
prastoin, line 14: cc @martmull

Weiko (approved):
LGTM

PR #17439, opened 2026-01-26
fix: prevent record title reset on focus for notes and tasks

Richman018:
Fixes issue where focusing the record title input would reset the title to empty. This ensures the draft value is properly initialized from the field value if it's undefined or empty when the input is focused.

Fixes #17437


lucasbordeau (approved):

lucasbordeau:
Hi @Richman018, the root cause was that object metadata item drawn from context store was wrong in some cases. So instead of forcing the draft value to recompute, I forced the caller of openRecordTitleCell to give a fieldMetadataItemId, that way we cannot end up with a correct field name with the wrong object metadata item, which was the case here.

Richman018:
> Hi @Richman018, the root cause was that object metadata item drawn from context store was wrong in some cases. So instead of forcing the draft value to recompute, I forced the caller of openRecordTitleCell to give a fieldMetadataItemId, that way we cannot end up with a correct field name with the wrong object metadata item, which was the case here.

Thank you for your review

PR #17444, opened 2026-01-26
fix: use universalIdentifier to identify the field in migrate-attachment-to-morph-relations

mabdullahabaid:
Made changes to the command based on suggestions [here](https://github.com/twentyhq/twenty/pull/17381).

### DIFF
--- a/packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
+++ b/packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
@@ -5,7 +5,7 @@ import {
   FieldMetadataType,
   type FieldMetadataSettings,
 } from 'twenty-shared/types';
-import { capitalize, isDefined } from 'twenty-shared/utils';
+import { capitalize } from 'twenty-shared/utils';
 import { DataSource, Repository } from 'typeorm';
 
 import { ActiveOrSuspendedWorkspacesMigrationCommandRunner } from 'src/database/commands/command-runners/active-or-suspended-workspaces-migration.command-runner';
@@ -17,9 +17,7 @@ import { FeatureFlagService } from 'src/engine/core-modules/feature-flag/service
 import { WorkspaceEntity } from 'src/engine/core-modules/workspace/workspace.entity';
 import { DataSourceService } from 'src/engine/metadata-modules/data-source/data-source.service';
 import { computeMorphOrRelationFieldJoinColumnName } from 'src/engine/metadata-modules/field-metadata/utils/compute-morph-or-relation-field-join-column-name.util';
-import { findFlatEntityByIdInFlatEntityMaps } from 'src/engine/metadata-modules/flat-entity/utils/find-flat-entity-by-id-in-flat-entity-maps.util';
 import { findFlatEntityByUniversalIdentifier } from 'src/engine/metadata-modules/flat-entity/utils/find-flat-entity-by-universal-identifier.util';
-import { type FlatEntityMaps } from 'src/engine/metadata-modules/flat-entity/types/flat-entity-maps.type';
 import { getMetadataFlatEntityMapsKey } from 'src/engine/metadata-modules/flat-entity/utils/get-metadata-flat-entity-maps-key.util';
 import { getMetadataRelatedMetadataNames } from 'src/engine/metadata-modules/flat-entity/utils/get-metadata-related-metadata-names.util';
 import { isMorphOrRelationFlatFieldMetadata } from 'src/engine/metadata-modules/flat-field-metadata/utils/is-morph-or-relation-flat-field-metadata.util';
@@ -135,17 +133,7 @@ export class MigrateAttachmentToMorphRelationsCommand extends ActiveOrSuspendedW
       const attachmentTargetRelationFields = attachmentFieldMetadatas
         .filter(isMorphOrRelationFlatFieldMetadata)
         .filter((field) => field.type === FieldMetadataType.RELATION)
-        .flatMap((field) => {
-          const relatedObject = this.getRelatedObjectMetadata({
-            field,
-            flatObjectMetadataMaps,
-            workspaceId,
-          });
-
-          if (!isDefined(relatedObject)) {
-            return [];
-          }
-
+        .filter((field) => {
           const isStandardAppField = this.isTwentyStandardApplicationField({
             field,
             flatApplicationMaps,
@@ -160,11 +148,7 @@ export class MigrateAttachmentToMorphRelationsCommand extends ActiveOrSuspendedW
             !isStandardAppField &&
             field.standardId === ATTACHMENT_STANDARD_FIELD_IDS.targetCustom;
 
-          if (!isStandardTarget && !isCustomTarget) {
-            return [];
-          }
-
-          return [field];
+          return isStandardTarget || isCustomTarget;
         });
 
       const fieldMigrations = attachmentTargetRelationFields.map((field) => {
@@ -312,18 +296,10 @@ export class MigrateAttachmentToMorphRelationsCommand extends ActiveOrSuspendedW
     flatApplicationMaps,
     workspaceId,
   }: {
-    field: { id: string; name: string; applicationId: string | null };
+    field: { id: string; name: string; applicationId: string };
     flatApplicationMaps: FlatApplicationCacheMaps;
     workspaceId: string;
   }): boolean {
-    if (!field.applicationId) {
-      this.logger.error(
-        `🟥 Missing applicationId for field "${field.name}" (${field.id}) in workspace ${workspaceId}`,
-      );
-
-      return false;
-    }
-
     const application = flatApplicationMaps.byId[field.applicationId];
 
     if (!application) {
@@ -338,40 +314,5 @@ export class MigrateAttachmentToMorphRelationsCommand extends ActiveOrSuspendedW
       application.universalIdentifier ===
       TWENTY_STANDARD_APPLICATION.universalIdentifier
     );
-  }
-
-  private getRelatedObjectMetadata({
-    field,
-    flatObjectMetadataMaps,
-    workspaceId,
-  }: {
-    field: {
-      id: string;
-      name: string;
-      relationTargetObjectMetadataId?: string | null;
-    };
-    flatObjectMetadataMaps: FlatEntityMaps<FlatObjectMetadata>;
-    workspaceId: string;
-  }): FlatObjectMetadata | undefined {
-    if (!field.relationTargetObjectMetadataId) {
-      this.logger.error(
-        `🟥 Missing relationTargetObjectMetadataId for field "${field.name}" (${field.id}) in workspace ${workspaceId}`,
-      );
-
-      return;
-    }
-
-    const relatedObject = findFlatEntityByIdInFlatEntityMaps({
-      flatEntityMaps: flatObjectMetadataMaps,
-      flatEntityId: field.relationTargetObjectMetadataId,
-    });
-
-    if (!relatedObject) {
-      this.logger.error(
-        `🟥 Relation target object metadata not found for field "${field.name}" (${field.id}) in workspace ${workspaceId} (relationTargetObjectMetadataId: ${field.relationTargetObjectMetadataId})`,
-      );
-    }
-
-    return relatedObject;
   }
 }
--- a/packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
+++ b/packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
@@ -1,9 +1,11 @@
 import { InjectDataSource, InjectRepository } from '@nestjs/typeorm';
 
 import { Command } from 'nest-commander';
-import { STANDARD_OBJECT_IDS } from 'twenty-shared/metadata';
-import { type FieldMetadataRelationSettings } from 'twenty-shared/types';
-import { capitalize } from 'twenty-shared/utils';
+import {
+  FieldMetadataType,
+  type FieldMetadataRelationSettings,
+} from 'twenty-shared/types';
+import { capitalize, isDefined } from 'twenty-shared/utils';
 import { DataSource, Repository } from 'typeorm';
 
 import { ActiveOrSuspendedWorkspacesMigrationCommandRunner } from 'src/database/commands/command-runners/active-or-suspended-workspaces-migration.command-runner';
@@ -13,15 +15,18 @@ import { FeatureFlagService } from 'src/engine/core-modules/feature-flag/service
 import { WorkspaceEntity } from 'src/engine/core-modules/workspace/workspace.entity';
 import { DataSourceService } from 'src/engine/metadata-modules/data-source/data-source.service';
 import { computeMorphOrRelationFieldJoinColumnName } from 'src/engine/metadata-modules/field-metadata/utils/compute-morph-or-relation-field-join-column-name.util';
+import { findFlatEntityByUniversalIdentifier } from 'src/engine/metadata-modules/flat-entity/utils/find-flat-entity-by-universal-identifier.util';
 import { getMetadataFlatEntityMapsKey } from 'src/engine/metadata-modules/flat-entity/utils/get-metadata-flat-entity-maps-key.util';
 import { getMetadataRelatedMetadataNames } from 'src/engine/metadata-modules/flat-entity/utils/get-metadata-related-metadata-names.util';
-import { ObjectMetadataEntity } from 'src/engine/metadata-modules/object-metadata/object-metadata.entity';
+import { isMorphOrRelationFlatFieldMetadata } from 'src/engine/metadata-modules/flat-field-metadata/utils/is-morph-or-relation-flat-field-metadata.util';
 import { WorkspaceMetadataVersionService } from 'src/engine/metadata-modules/workspace-metadata-version/services/workspace-metadata-version.service';
 import { GlobalWorkspaceOrmManager } from 'src/engine/twenty-orm/global-workspace-datasource/global-workspace-orm.manager';
 import { WorkspaceCacheStorageService } from 'src/engine/workspace-cache-storage/workspace-cache-storage.service';
 import { WorkspaceCacheService } from 'src/engine/workspace-cache/services/workspace-cache.service';
 import { type WorkspaceCacheKeyName } from 'src/engine/workspace-cache/types/workspace-cache-key.type';
 import { getWorkspaceSchemaName } from 'src/engine/workspace-datasource/utils/get-workspace-schema-name.util';
+import { STANDARD_OBJECTS } from 'src/engine/workspace-manager/twenty-standard-application/constants/standard-object.constant';
+import { TWENTY_STANDARD_APPLICATION } from 'src/engine/workspace-manager/twenty-standard-application/constants/twenty-standard-applications';
 import { ATTACHMENT_STANDARD_FIELD_IDS } from 'src/engine/workspace-manager/workspace-migration/constant/standard-field-ids';
 
 @Command({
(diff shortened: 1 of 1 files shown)


packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
```diff
@@ -76,42 +81,30 @@ export class MigrateAttachmentToMorphRelationsCommand extends ActiveOrSuspendedW
...
-      ];
-
-      const fieldMigrations = fieldNameMigrations.map(
-        ({ old: oldFieldName, new: newFieldName }) => ({
-          old: `${oldFieldName}Id`,
-          new: `${newFieldName}Id`,
-        }),
-      );
+      const attachmentTargetFieldUniversalIdentifiers = new Set<string>([
+        ATTACHMENT_STANDARD_FIELD_IDS.targetTask,
```
prastoin: ***Remark:*** Please always use `STANDARD_OBJECTS` even though until we've totally deprecate standardId it might for some of them be a bridge to below variables

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
```diff
@@ -76,42 +81,30 @@ export class MigrateAttachmentToMorphRelationsCommand extends ActiveOrSuspendedW
       const schemaName = getWorkspaceSchemaName(workspaceId);
       const tableName = 'attachment';
 
-      const customObjectMetadata = await this.objectMetadataRepository.find({
+      const relatedObjectMetadata = await this.objectMetadataRepository.find({
```
prastoin: Lets use `WorkspaceCacheService` here and `flatObjectMetadataMaps` instead

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
```diff
@@ -76,42 +81,30 @@ export class MigrateAttachmentToMorphRelationsCommand extends ActiveOrSuspendedW
...
+        ATTACHMENT_STANDARD_FIELD_IDS.targetNote,
+        ATTACHMENT_STANDARD_FIELD_IDS.targetPerson,
+        ATTACHMENT_STANDARD_FIELD_IDS.targetCompany,
+        ATTACHMENT_STANDARD_FIELD_IDS.targetOpportunity,
+        ATTACHMENT_STANDARD_FIELD_IDS.targetDashboard,
+        ATTACHMENT_STANDARD_FIELD_IDS.targetWorkflow,
+      ]);
 
       const attachmentObjectMetadata =
         await this.objectMetadataRepository.findOne({
```
prastoin: Same use cache with `findFlatEntityFromUniversalIdentifier`

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
```diff
@@ -124,18 +117,85 @@ export class MigrateAttachmentToMorphRelationsCommand extends ActiveOrSuspendedW
...
+
+          const relatedObject = relatedObjectMetadataMap.get(
+            field.relationTargetObjectMetadataId,
+          );
+
+          if (!relatedObject) {
+            return false;
+          }
+
+          const isStandardTarget =
```
prastoin, line 142: ***Suggestion:*** To check if an metadata entity is own by the twenty-standard app I would rather check field `applicationId` and check if related `FlatApplication` universalIdentifier from `FlatApplicationMaps` is equal to `TWENTY_STANDARD_APPLICATION.universalIdentifier`

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
```diff
@@ -124,18 +117,85 @@ export class MigrateAttachmentToMorphRelationsCommand extends ActiveOrSuspendedW
         return;
       }
 
-      for (const { new: newField, old: oldField } of fieldMigrations) {
+      const attachmentTargetRelationFields = attachmentObjectMetadata.fields
+        .filter((field) => field.type === 'RELATION')
+        .filter((field) => {
```
prastoin: ***Remark:*** If this filter aims to filter out any field entry that do not have an `applicationId`nor `universalIdentifier` it's redundant as now not nullable

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
```diff
@@ -124,18 +117,85 @@ export class MigrateAttachmentToMorphRelationsCommand extends ActiveOrSuspendedW
...
+              field.universalIdentifier,
+            );
+          const isCustomTarget = relatedObject.isCustom;
+
+          return isStandardTarget || isCustomTarget;
+        });
+
+      const fieldMigrations = attachmentTargetRelationFields.flatMap(
+        (field) => {
+          if (!field.relationTargetObjectMetadataId) {
```
prastoin: ***Suggestion:*** `isMorphOrRelationFlatFieldMetadata`

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
```diff
@@ -227,11 +252,24 @@ export class MigrateAttachmentToMorphRelationsCommand extends ActiveOrSuspendedW
         workspaceId,
       );
 
-      await this.workspaceCacheService.invalidateAndRecompute(workspaceId, [
+      const relatedMetadataNames =
+        getMetadataRelatedMetadataNames('fieldMetadata');
+      const relatedCacheKeysToInvalidate: WorkspaceCacheKeyName[] =
+        relatedMetadataNames.map(getMetadataFlatEntityMapsKey);
+      const cacheKeysToInvalidate: WorkspaceCacheKeyName[] = [
```
prastoin, line 258: cool !

prastoin (commented):
Hey there, good direction lets mostly use cache and flat entity in the command
When are you updating the field type to be a Morph too ?

prastoin:
Double checked we also need to update the field `type` and `morphId`

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
```diff
@@ -124,18 +124,96 @@ export class MigrateAttachmentToMorphRelationsCommand extends ActiveOrSuspendedW
...
+        flatFieldMetadataMaps,
+      );
+
+      const isTwentyStandardApplicationField = (field: {
+        applicationId: string;
+      }) => {
+        const application = flatApplicationMaps.byId[field.applicationId];
+
+        return (
+          application?.universalIdentifier ===
```
prastoin: ***Polish:*** Would make a util or method but avoid declaring the util inside the run
***Polish:*** It should never be undefined, lets log an error if it's

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
```diff
@@ -124,18 +124,96 @@ export class MigrateAttachmentToMorphRelationsCommand extends ActiveOrSuspendedW
...
+          TWENTY_STANDARD_APPLICATION.universalIdentifier
+        );
+      };
+
+      const getRelatedObjectMetadata = (field: {
```
prastoin: ***Subjective polish:*** bit redundant as could use the `findFlatEntityInMapsFromId` directly

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
```diff
@@ -124,18 +124,96 @@ export class MigrateAttachmentToMorphRelationsCommand extends ActiveOrSuspendedW
...
+      };
+
+      const attachmentTargetRelationFields = attachmentFieldMetadatas
+        .filter(isMorphOrRelationFlatFieldMetadata)
+        .filter((field) => field.type === FieldMetadataType.RELATION)
+        .flatMap((field) => {
+          const relatedObject = getRelatedObjectMetadata(field);
+
+          if (!isDefined(relatedObject)) {
+            return [];
```
prastoin: ***Polish:*** log error as should never occur

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
```diff
@@ -124,18 +124,96 @@ export class MigrateAttachmentToMorphRelationsCommand extends ActiveOrSuspendedW
...
+
+          const isStandardAppField = isTwentyStandardApplicationField(field);
+          const isStandardTarget =
+            isStandardAppField &&
+            attachmentTargetFieldUniversalIdentifiers.has(
+              field.universalIdentifier,
+            );
+          const isCustomTarget = relatedObject.isCustom && !isStandardAppField;
+
+          if (!isStandardTarget && !isCustomTarget) {
```
prastoin: ***Question:*** I'm not sure to understand how this would be possible ? What's the goal of this function overall ?

If it's retrieving all attachement relation field, either custom or standard ( which by definition any field row is one of the two ) then `.filter(isFlatFieldMetadataOfType(RELATION))

Would be enough

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
```diff
@@ -145,60 +223,30 @@ export class MigrateAttachmentToMorphRelationsCommand extends ActiveOrSuspendedW
...
-          joinColumnName: computeMorphOrRelationFieldJoinColumnName({
-            name: newFieldName,
-          }),
+          joinColumnName: newJoinColumnName,
         };
 
         try {
           const result = await queryRunner.query(
             `UPDATE core."fieldMetadata"
-           SET name = $1, type = 'MORPH_RELATION', "morphId" = $3, settings = $4
```
prastoin, line 199: ***Remark:*** Ah my bad it was there !

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
```diff
@@ -124,18 +124,96 @@ export class MigrateAttachmentToMorphRelationsCommand extends ActiveOrSuspendedW
...
+            return [];
+          }
+
+          const isStandardAppField = isTwentyStandardApplicationField(field);
+          const isStandardTarget =
+            isStandardAppField &&
+            attachmentTargetFieldUniversalIdentifiers.has(
+              field.universalIdentifier,
+            );
+          const isCustomTarget = relatedObject.isCustom && !isStandardAppField;
```
prastoin: ***Question:*** Why checking related object isCustom here ?

prastoin (commented):
Nice job with the caches ! Please let me know

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
mabdullahabaid: We only migrate attachment relation fields that are either known standard targets (by universalIdentifier) or custom targets pointing to custom objects. The `relatedObject.isCustom` guard is there to avoid touching custom attachment fields that target standard objects, since those aren’t part of the known target set and could be unrelated. 

If the intent is to migrate all attachment relations, we should drop that guard.

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
mabdullahabaid: Replied on the other comment.

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
prastoin: To determine if an entity is custom or standard we need to verify by what application it's being own, if it's twenty-standard than it needs to be migrated
In order to isolate relation field resulting from an object creation side effect ( default relation to attachement creation ) we can still use the about to be deprecated though `standardId`
```ts
  targetCustom: '20202020-302d-43b3-9aea-aa4f89282a9f',
```

If a relation has such standard id it needs to be migrated to the be part of the morph but still be own by the custom app

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
mabdullahabaid: `getRelatedObjectMetadata` now logs an error when the relation target is missing or can’t be resolved, so the case is already covered without extra logging.

mabdullahabaid:
@prastoin done resolving the comments.

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
```diff
@@ -255,4 +306,72 @@ export class MigrateAttachmentToMorphRelationsCommand extends ActiveOrSuspendedW
...
+  private isTwentyStandardApplicationField({
+    field,
+    flatApplicationMaps,
+    workspaceId,
+  }: {
+    field: { id: string; name: string; applicationId: string | null };
+    flatApplicationMaps: FlatApplicationCacheMaps;
+    workspaceId: string;
+  }): boolean {
+    if (!field.applicationId) {
```
prastoin: redundant, always defined

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
```diff
@@ -255,4 +306,72 @@ export class MigrateAttachmentToMorphRelationsCommand extends ActiveOrSuspendedW
...
+      return;
+    }
+
+    const relatedObject = findFlatEntityByIdInFlatEntityMaps({
+      flatEntityMaps: flatObjectMetadataMaps,
+      flatEntityId: field.relationTargetObjectMetadataId,
+    });
+
+    if (!relatedObject) {
+      this.logger.error(
```
prastoin: might wanna throw

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
```diff
@@ -124,18 +127,83 @@ export class MigrateAttachmentToMorphRelationsCommand extends ActiveOrSuspendedW
...
+      const attachmentFieldMetadatas = getFlatFieldsFromFlatObjectMetadata(
+        attachmentObjectMetadata,
+        flatFieldMetadataMaps,
+      );
+
+      const attachmentTargetRelationFields = attachmentFieldMetadatas
+        .filter(isMorphOrRelationFlatFieldMetadata)
+        .filter((field) => field.type === FieldMetadataType.RELATION)
+        .flatMap((field) => {
+          const relatedObject = this.getRelatedObjectMetadata({
```
prastoin: Unless I'm mistaken useless related object retrieval here
```
          const relatedObject = this.getRelatedObjectMetadata({
            field,
            flatObjectMetadataMaps,
            workspaceId,
          });

          if (!isDefined(relatedObject)) {
            return [];
          }
```

What's the motivation ?

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
```diff
@@ -124,18 +127,83 @@ export class MigrateAttachmentToMorphRelationsCommand extends ActiveOrSuspendedW
...
+          const isStandardTarget =
+            isStandardAppField &&
+            attachmentTargetFieldUniversalIdentifiers.has(
+              field.universalIdentifier,
+            );
+          const isCustomTarget =
+            !isStandardAppField &&
+            field.standardId === ATTACHMENT_STANDARD_FIELD_IDS.targetCustom;
+
+          if (!isStandardTarget && !isCustomTarget) {
```
prastoin: Quite though to read IMO, if I understand correctly could be replaced by
```ts
          if (isStandardTarget || isCustomTarget) {
return [field]
}
```

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
```diff
@@ -124,18 +127,83 @@ export class MigrateAttachmentToMorphRelationsCommand extends ActiveOrSuspendedW
...
-      for (const { new: newField, old: oldField } of fieldMigrations) {
+      const attachmentFieldMetadatas = getFlatFieldsFromFlatObjectMetadata(
+        attachmentObjectMetadata,
+        flatFieldMetadataMaps,
+      );
+
+      const attachmentTargetRelationFields = attachmentFieldMetadatas
+        .filter(isMorphOrRelationFlatFieldMetadata)
+        .filter((field) => field.type === FieldMetadataType.RELATION)
+        .flatMap((field) => {
```
prastoin: Unless I'm mistaken flatmap could be replaced by a filter

prastoin (approved):
Hey there, as discussed I would highly recommend this to be tested on a prod extract before proceeding the other entities implementation
Lets few comments about polish implem aspects
I haven't tested the command myself

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
mabdullahabaid: We check the related object as a guard, not because the migration needs it. It skips relation fields with missing/dangling targets and surfaces the issue during upgrade. If targets are always valid, the check can be removed.

prastoin: There's a constraint at pg level, a relation field has a fk on this col and has to be defined

prastoin: So not really useful to double check that

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-attachment-to-morph-relations.command.ts
```diff
@@ -124,18 +125,69 @@ export class MigrateAttachmentToMorphRelationsCommand extends ActiveOrSuspendedW
...
+          const isStandardTarget =
+            isStandardAppField &&
+            attachmentTargetFieldUniversalIdentifiers.has(
+              field.universalIdentifier,
+            );
+          const isCustomTarget =
+            !isStandardAppField &&
+            field.standardId === ATTACHMENT_STANDARD_FIELD_IDS.targetCustom;
+
+          return isStandardTarget || isCustomTarget;
```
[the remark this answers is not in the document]
mabdullahabaid: Fixed.

mabdullahabaid:
Tested on prod extract of four workspace schemas and it works as expected — columns get renamed to have a target-prefix (both standard and custom), type in fieldMetadata table gets updated to `MORPH_RELATION` (with morphId set), and `settings.joinColumnName` stays aligned to the new column name while relationType remains MANY_TO_ONE.

PR #17448, opened 2026-01-26
fix: allow reusing Person email after deletion

GiantDragon9090:
## Problem

When a Person record is deleted, the email address cannot be reused when creating a new person. The system throws an error: "This record already exists. Please check your data and try again." even though the deleted person is no longer visible in the UI.

## Solution

Added `indexWhereClause: "\"deletedAt\" IS NULL"` to the `emailsUniqueIndex` definition. This creates a partial unique index that only enforces uniqueness for non-deleted records, allowing email addresses to be reused after deletion.

## Changes

- Updated `compute-person-standard-flat-index-metadata.util.ts` to include `indexWhereClause` in the `emailsUniqueIndex` configuration
- This follows the same pattern used in other unique indexes (e.g., `messageChannelIdMessageIdUniqueIndex`)

## Testing

This fix ensures that:
- Email uniqueness is still enforced for active (non-deleted) records
- Deleted person records no longer block email reuse
- The behavior matches user expectations

Fixes #17425

### DIFF
--- a/packages/twenty-server/src/engine/workspace-manager/twenty-standard-application/utils/index/compute-person-standard-flat-index-metadata.util.ts
+++ b/packages/twenty-server/src/engine/workspace-manager/twenty-standard-application/utils/index/compute-person-standard-flat-index-metadata.util.ts
@@ -36,6 +36,7 @@ export const buildPersonStandardFlatIndexMetadatas = ({
       indexName: 'emailsUniqueIndex',
       relatedFieldNames: ['emails'],
       isUnique: true,
+      indexWhereClause: '"deletedAt" IS NULL',
     },
     standardObjectMetadataRelatedEntityIds,
     dependencyFlatEntityMaps,


packages/twenty-server/src/engine/workspace-manager/twenty-standard-application/utils/index/compute-person-standard-flat-index-metadata.util.ts
```diff
@@ -36,6 +36,7 @@ export const buildPersonStandardFlatIndexMetadatas = ({
       indexName: 'emailsUniqueIndex',
       relatedFieldNames: ['emails'],
       isUnique: true,
+      indexWhereClause: '"deletedAt" IS NULL',
```
etiennejouan, line 39: Thanks for your PR @GiantDragon9090, but this behaviour is intended. 

If you need to re-use the email, you need to destroy the person record first (or wait for the 14-day retention period for soft-deleted records)."

PR #17450, opened 2026-01-26
Use pipeline to count event stream

thomtrp:
As title. So we only send one TCP call per batch

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/cache-storage/services/cache-storage.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/cache-storage/services/cache-storage.service.ts
@@ -233,9 +233,14 @@ export class CacheStorageService {
           pipeline.sCard(key);
         }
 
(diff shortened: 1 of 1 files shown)


packages/twenty-server/src/engine/core-modules/cache-storage/services/cache-storage.service.ts
```diff
@@ -227,11 +227,20 @@ export class CacheStorageService {
       const keys = result.keys;
 
       if (keys.length > 0) {
-        const counts = await Promise.all(
-          keys.map((key) => redisClient.sCard(key)),
-        );
+        const pipeline = redisClient.multi();
```
Weiko, line 230: Actually my bad, multi is blocking here. I'm not sure this is a good idea after all and we probably prefer the multiple round trips 🤔

Weiko (approved):
To monitor! @thomtrp

PR #17451, opened 2026-01-26
docs: add example output after creating postgres role

TomLarson20:
Added example output showing what the roles table should look like after creating the postgres role


charlesBochet (approved):

charlesBochet:
Thank you!

PR #17470, opened 2026-01-27
Fixed plain object in field value for workflow

lucasbordeau:
This PR fixes a bug that arises in `formatResult` following up the recent refactor for DATE_TIME : https://github.com/twentyhq/twenty/pull/17407

In the case of workflows, we pass a plain object to `formatResult` : 

```ts
{
  before: null, 
  after: '2026-01-27T10:59:15.525Z'
}
```

So a case has been added to handle this. 

@Weiko are we ok with this more restrictive else-if part in this util ? We could also just put a `continue` for unknown shapes.


Weiko (approved):
LGTM

thomtrp (approved):

PR #17475, opened 2026-01-27
Emit the restore event on restore

thomtrp:
Restore current triggers a delete event

### DIFF
--- a/packages/twenty-server/src/modules/workflow/workflow-trigger/automated-trigger/listeners/workflow-database-event-trigger.listener.ts
+++ b/packages/twenty-server/src/modules/workflow/workflow-trigger/automated-trigger/listeners/workflow-database-event-trigger.listener.ts
@@ -5,6 +5,7 @@ import {
   type ObjectRecordCreateEvent,
   type ObjectRecordDeleteEvent,
   type ObjectRecordDestroyEvent,
+  type ObjectRecordRestoreEvent,
   type ObjectRecordUpdateEvent,
   type ObjectRecordUpsertEvent,
 } from 'twenty-shared/database-events';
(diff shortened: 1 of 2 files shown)


packages/twenty-server/src/modules/workflow/workflow-trigger/automated-trigger/listeners/workflow-database-event-trigger.listener.ts
```diff
@@ -137,6 +138,13 @@ export class WorkflowDatabaseEventTriggerListener {
     });
   }
 
+  @OnDatabaseBatchEvent('*', DatabaseEventAction.RESTORED)
+  async handleObjectRecordRestoreEvent(
+    _payload: WorkspaceEventBatch<ObjectRecordRestoreEvent>,
+  ) {
+    return;
```
Weiko, line 145: Is it expected to be empty here? Previous restore won't go through the DELETE handler anymore so we probably want to implement the same code here, right? @thomtrp

Weiko (commented):
Small comment otherwise LGTM

packages/twenty-server/src/modules/workflow/workflow-trigger/automated-trigger/listeners/workflow-database-event-trigger.listener.ts
thomtrp: We do not provide a way to listen at restore events yet on frontend side. I can check with @Bonapara if we want it or if we do not care

Weiko: What about webhooks or DB event trigger for serverless though @thomtrp

thomtrp:
closing temporarily to decide what to do

PR #17476, opened 2026-01-27
Migrate noteTarget and taskTarget to Morph.

mabdullahabaid:
This PR migrates `noteTarget` and `taskTarget` to morph relations behind separate feature flags, following the Attachment/TimelineActivity pattern.

It introduces the `IS_NOTE_TARGET_MIGRATED` and `IS_TASK_TARGET_MIGRATED` flags, updates standard field metadata and indexes to use morph relations, and adds two **1.17 workspace migrations** that:
- rename `noteTarget.*Id` / `taskTarget.*Id` columns to `target*Id`
- convert the corresponding field metadata to `MORPH_RELATION` with a shared `morphId`

On the frontend, note/task target read and write paths switch to `target*Id` when the respective flag is enabled. Deleted targets are filtered on reload to prevent reappearing relations.


mabdullahabaid:
Ran the commands on a production extract of four workspaces and it works as expected.

- Columns are renamed to use a `target*` prefix (both standard and custom fields)
- `fieldMetadata.type` is updated to `MORPH_RELATION` with `morphId` set
- `settings.joinColumnName` is kept in sync with the new column name
- `relationType` remains `MANY_TO_ONE`

`MANY_TO_MANY` relations were not handled in this PR, as instructed.

charlesBochet:
@FelixMalfait could you take over this one?

FelixMalfait (approved):

packages/twenty-server/src/engine/workspace-manager/workspace-migration/constant/standard-field-ids.ts
```diff
@@ -0,0 +1,591 @@
+/**
+ * /!\ DO NOT EDIT THE IDS OF THIS FILE /!\
+ * This file contains static ids for standard objects.
+ * These ids are used to identify standard objects in the database and compare them even when renamed.
+ * For readability keys can be edited but the values should not be changed.
+ */
+
+// TODO: check if this can be deleted
+export const ACTIVITY_TARGET_STANDARD_FIELD_IDS = {
```
prastoin, line 9: Hey there, standards ids don't exist and shouldn't be inserted anymore
Please use STANDARD_OBJECTS and universal identifiers

PR #17477, opened 2026-01-27
Migrate favorites to navigation menu items (1.17 upgrade)

abdulrahmancodes:
**What this PR does:**
- Migrates existing `Favorite` and `FavoriteFolder` entities to the new `NavigationMenuItem` structure
- Preserves user-level vs workspace-level ownership
- Preserves folder structure, positions, and relationships
- Handles both view-based favorites (linked to views) and record-based favorites (linked to records)
- Soft-deletes original favorites and folders after successful migration
- Enables the `IS_NAVIGATION_MENU_ITEM_ENABLED` feature flag post-migration
- Skips migration if the feature flag is already enabled
- Idempotent: checks for existing navigation menu items to prevent duplicates

**Migration flow:**
1. Migrate favorite folders first (creates folder mapping)
2. Migrate favorites
3. Soft-delete migrated favorites and folders
4. Enable feature flag

**What's next:**
- After all workspaces are migrated and the navigation menu item feature is fully rolled out, we can:
  - Remove the old `Favorite` and `FavoriteFolder` entities and related code
  - Remove the feature flag check and make navigation menu items the default
  - Clean up any deprecated favorites-related code paths in the frontend


packages/twenty-server/src/database/commands/upgrade-version-command/upgrade.command.ts
```diff
@@ -187,13 +191,18 @@ export class UpgradeCommand extends UpgradeCommandRunner {
       this.migrateAttachmentToMorphRelationsCommand,
     ];
 
+    const commands_1180: VersionCommands = [
+      this.migrateFavoritesToNavigationMenuItemsCommand,
```
etiennejouan, line 65: Unless it's done on purpose, we could add it to 1.17 release instead

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,500 @@
...
+      );
+
+      const newId = uuidv4();
+      const now = new Date().toISOString();
+
+      folderIdMapping.set(favoriteFolder.id, newId);
+
+      flatNavigationMenuItemsToCreate.push({
+        id: newId,
+        universalIdentifier: newId,
```
etiennejouan: @prastoin let you tell us if there is no issue having id = universalIdentifier

etiennejouan (commented):
Thanks for your PR ! 🙏 

I'll let @prastoin tell us what he thinks of it, but I see some improvements to your migration command:

- Command should be idempotent: running the command once or twice should give the same result. In your case, the folder navigation menu items would be re-created, and same for the user-workspace navigation menu items.

- Concerning workspace-level favorites with defined viewId: they should be associated with the Twenty application ID + the universal identifier ID you defined in the previous PR (for views related to objects created in the Twenty app; for others, it should be the custom workspace app ID).

- Concerning favorites with defined recordId: we need to check all RELATION fields on favorite (programmatically via field metadata) because we should support favorites on custom objects.

- It would be great to toggle on the feature flag once the migration succeeds: this helps with idempotence, f we migrate only workspaces with the deactivated feature flag.

- Will the favorite and favoriteFolder tables be dropped in another command? What about soft-deleting favorites when converted into items? (Also good for idempotence)

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,500 @@
...
+import { FlatNavigationMenuItem } from 'src/engine/metadata-modules/flat-navigation-menu-item/types/flat-navigation-menu-item.type';
+import { FlatObjectMetadata } from 'src/engine/metadata-modules/flat-object-metadata/types/flat-object-metadata.type';
+import { GlobalWorkspaceOrmManager } from 'src/engine/twenty-orm/global-workspace-datasource/global-workspace-orm.manager';
+import { buildSystemAuthContext } from 'src/engine/twenty-orm/utils/build-system-auth-context.util';
+import { WorkspaceCacheService } from 'src/engine/workspace-cache/services/workspace-cache.service';
+import { WorkspaceMigrationValidateBuildAndRunService } from 'src/engine/workspace-manager/workspace-migration/services/workspace-migration-validate-build-and-run-service';
+import { FavoriteFolderWorkspaceEntity } from 'src/modules/favorite-folder/standard-objects/favorite-folder.workspace-entity';
+import { FavoriteWorkspaceEntity } from 'src/modules/favorite/standard-objects/favorite.workspace-entity';
+import { WorkspaceMemberWorkspaceEntity } from 'src/modules/workspace-member/standard-objects/workspace-member.workspace-entity';
+
```
FelixMalfait, line 39: Why 1.18 if 1.17 hasn't been released yet?

FelixMalfait: It has to be in a different version for this to work?

prastoin: It should be 1.17 it might be a mistake

prastoin:
Sorry was under water the whole day ! Will for sure check tomorrow morning

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
prastoin: that's ok for custom entities

packages/twenty-server/src/database/commands/upgrade-version-command/1-18/1-18-migrate-favorites-to-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,520 @@
...
+      );
+
+      return null;
+    }
+  }
+
+  private async getTargetRecordFromFavorite(
+    favorite: FavoriteWorkspaceEntity,
+    workspaceId: string,
+    authContext: ReturnType<typeof buildSystemAuthContext>,
```
prastoin: ***Remark:*** both `workspaceId` and `authContext` are unused

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,520 @@
...
+  override async runOnWorkspace({
+    workspaceId,
+    options,
+  }: RunOnWorkspaceArgs): Promise<void> {
+    this.logger.log(
+      `Starting migration of favorites to navigation menu items for workspace ${workspaceId}`,
+    );
+
+    const authContext = buildSystemAuthContext(workspaceId);
```
prastoin: ***Polish:*** Could be computed on at `getUserWorkspaceIdFromFavoriteFolder` level and could remove from all method args

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,520 @@
...
+    const flatNavigationMenuItemsToCreate: FlatNavigationMenuItem[] = [];
+
+    for (const favoriteFolder of favoriteFolders) {
+      const userWorkspaceId = await this.getUserWorkspaceIdFromFavoriteFolder(
+        favoriteFolder.id,
+        workspaceId,
+        authContext,
+      );
+
+      const existingFolder = Object.values(
```
prastoin: ***Question:*** Are users still able through the UI the create the old Favorite entities with the feature flag disabled ? => We should enforce not being able to create any  of the old entities after this command has been within a workspace, or this would have to be re-run ? ( or maybe are you populating both entities ) please let me know

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,520 @@
...
+
+    if (options.dryRun) {
+      this.logger.log(
+        `[DRY RUN] Would migrate favorites to navigation menu items for workspace ${workspaceId}`,
+      );
+
+      return;
+    }
+
+    const { workspaceCustomFlatApplication } =
```
prastoin: ***Question:*** We don't expect any entities to be migrated to twenty-standard application scope ?

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,520 @@
...
+        this.logger.log(
+          `Skipping favorite folder ${favoriteFolder.id} - navigation menu item folder already exists with name "${favoriteFolder.name}"`,
+        );
+
+        folderIdMapping.set(favoriteFolder.id, existingFolder.id);
+
+        continue;
+      }
+
+      const newId = uuidv4();
```
prastoin: ***Suggestion:*** Could we keep the same id so checking if a previous favoriteFolder ? so checking if a favoriteFolder has been migrated would be easier ( instead of all above properties assertions )

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,520 @@
...
+        },
+      );
+
+    if (isDefined(validateAndBuildResult)) {
+      throw new Error(
+        `Failed to create navigation menu items: ${JSON.stringify(validateAndBuildResult, null, 2)}`,
+      );
+    }
+
+    await this.workspaceCacheService.invalidateAndRecompute(workspaceId, [
```
prastoin: ***Polish:*** Redundant cache invalidation as the `validateBuildAndRunWorkspaceMigration` will handle it itself

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,520 @@
...
+
+    return { targetRecordId: null, targetObjectMetadataId: null };
+  }
+
+  private async createNavigationMenuItems(
+    workspaceId: string,
+    flatNavigationMenuItemsToCreate: FlatNavigationMenuItem[],
+    _existingFlatNavigationMenuItemMaps: FlatNavigationMenuItemMaps,
+  ): Promise<void> {
+    const validateAndBuildResult =
```
prastoin, line 682: ***Praise:*** Cool !

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,520 @@
...
+    workspaceId,
+    authContext,
+    applicationId,
+  }: {
+    workspaceId: string;
+    authContext: ReturnType<typeof buildSystemAuthContext>;
+    applicationId: string;
+  }): Promise<Map<string, string>> {
+    const favoriteFolderRepository =
+      await this.twentyORMGlobalManager.getRepository<FavoriteFolderWorkspaceEntity>(
```
prastoin, line 206: ***Remark:*** might technically contain soft deleted rows, as the new entity is not soft-delete able just wanna double check that's intended ( applicable to all repo interaction within the command )

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,520 @@
...
+      await this.workspaceCacheService.getOrRecompute(workspaceId, [
+        'flatNavigationMenuItemMaps',
+      ]);
+
+    const flatNavigationMenuItemsToCreate: FlatNavigationMenuItem[] = [];
+
+    for (const favorite of favorites) {
+      const userWorkspaceId = !isDefined(favorite.forWorkspaceMemberId)
+        ? null
+        : await this.getUserWorkspaceIdFromWorkspaceMemberId(
```
prastoin, line 428: ***Question:*** Don't we wanna throw or continue to next iteration if `favorite.forWorkspaceMemberId` is defined but not matching any ?

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,520 @@
...
+
+        if (isDefined(existingItem)) {
+          this.logger.log(
+            `Skipping favorite ${favorite.id} - navigation menu item already exists for view ${favorite.viewId}`,
+          );
+
+          continue;
+        }
+
+        const newId = uuidv4();
```
prastoin: ***Remark:*** Same

packages/twenty-server/src/database/commands/upgrade-version-command/1-18/1-18-migrate-favorites-to-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,520 @@
...
+      { field: 'dashboardId', objectName: 'dashboard' },
+    ];
+
+    for (const { field, objectName } of relationFields) {
+      const recordId = (
+        favorite as unknown as Record<string, string | undefined>
+      )[field];
+
+      if (isDefined(recordId)) {
+        const objectMetadata = Object.values(flatObjectMetadataMaps.byId).find(
```
prastoin: ***Request:*** Please standard relation field universalIdentifier instead

prastoin: Unless I'm mistaken from my understanding resulting migrated entities should be twenty-standard owned entities

prastoin (commented):
Hello there ! That's overall pretty good
I'm pretty concerned about not seeing any entities be associated to the workspace twenty-standard application whereas we are declaring some in the twenty-standard definition ( on a workspace creation some navigationItemMenu will be populated )
I would expect both custom and standard application to be consumed here

Lets few comments and recommendations please let me know

ps1: Please provide minial PR description that could ease identifiy what you understood of what needed to be done :pray:

ps2: 
```ts
  private async 
TwentyStandardApplicationService
// ...
createManyNavigationMenuItem({
    flatViews,
    workspaceId,
  }: {
```
Seems to be unused, unless I'm mistaken we're not declaring enough `standardNavigationMenuItem` either in the `twenty-standard-application` folder

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
abdulrahmancodes: Sorry, I thought 1.17 has been already released

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
abdulrahmancodes: we only want to migrate non–soft-deleted rows. I’ve added an explicit `where: { deletedAt: IsNull() }` to both `find()` calls. The fact that `NavigationMenuItem` isn’t soft-deletable is intentional

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
abdulrahmancodes: Yeah, should continue to next iteration

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
abdulrahmancodes: After `1-17` runs we set `IS_NAVIGATION_MENU_ITEM_ENABLED = true` for the workspace. So we do enforce ‘no new Favorites from the main UI path’ after the command has run.

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
prastoin: Ok sounds good, no need as typeorm by default sets `withDeleted` to false

packages/twenty-server/src/engine/metadata-modules/object-metadata/object-metadata.service.ts
```diff
@@ -517,6 +536,77 @@ export class ObjectMetadataService extends TypeOrmQueryService<ObjectMetadataEnt
...
+    );
+    const nextPosition =
+      workspaceLevelItems.length > 0
+        ? Math.max(...workspaceLevelItems.map((item) => item.position)) + 1
+        : 0;
+
+    const newId = uuidv4();
+    const now = new Date().toISOString();
+
+    const flatNavigationMenuItemToCreate: FlatNavigationMenuItem = {
```
etiennejouan: To validate with @prastoin but I think it should be add to the `validateBuildAndRunWorkspaceMigration` operation done on l-387 (same file)

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,578 @@
...
+              flatEntityToUpdate: [],
+            },
+          },
+          workspaceId,
+          isSystemBuild: true,
+        },
+      );
+
+    if (isDefined(validateAndBuildResult)) {
+      throw new Error(
```
etiennejouan, line 699: Throwing an error here would stop upgrade ? (@prastoin ?) If yes, it should be better to logger.error instead

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,578 @@
...
+      return [];
+    }
+
+    const { flatNavigationMenuItemMaps: existingFlatNavigationMenuItemMaps } =
+      await this.workspaceCacheService.getOrRecompute(workspaceId, [
+        'flatNavigationMenuItemMaps',
+      ]);
+
+    const favoriteObjectMetadata = await this.objectMetadataRepository.findOne({
+      where: {
```
etiennejouan: Favorite object and fields metadata are not available in cache ?

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,578 @@
...
+
+          continue;
+        }
+
+        const newId = uuidv4();
+        const now = new Date().toISOString();
+
+        flatNavigationMenuItemsToCreate.push({
+          id: newId,
+          universalIdentifier: newId,
```
etiennejouan: i think some navigation menu item should have specific universal identifier (from this constant `STANDARD_NAVIGATION_MENU_ITEMS`)

etiennejouan: You can check if the associated view has a matching universal identifier (in same constant)

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,578 @@
...
+          userWorkspaceId,
+          targetRecordId: null,
+          targetObjectMetadataId: null,
+          viewId: favorite.viewId,
+          folderId,
+          name: null,
+          position: favorite.position,
+          workspaceId,
+          applicationId: twentyStandardApplicationId,
```
etiennejouan: Then if standard navigation menu item, applicationId is twentyStandardApplicationId. If not, it is the custom one

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,578 @@
...
+        universalIdentifier: favorite.id,
+        userWorkspaceId,
+        targetRecordId,
+        targetObjectMetadataId,
+        viewId: null,
+        folderId,
+        name: null,
+        position: favorite.position,
+        workspaceId,
+        applicationId: twentyStandardApplicationId,
```
etiennejouan: custom application id, i would think here

etiennejouan (commented):
Well done ! Not an easy one. It looks very solid 

It may remain some issue with standard navigation menu items and applicationId value. I let @prastoin review and validate (or not) my comment.

packages/twenty-server/src/engine/metadata-modules/object-metadata/object-metadata.service.ts
prastoin: Yep you're right, should be included the same workspace migration as the object's creation one

packages/twenty-server/src/engine/metadata-modules/object-metadata/object-metadata.service.ts
```diff
@@ -454,10 +459,24 @@ export class ObjectMetadataService extends TypeOrmQueryService<ObjectMetadataEnt
       );
     }
 
-    await this.createWorkspaceFavoriteForNewObjectDefaultView({
-      view: flatDefaultViewToCreate,
-      workspaceId,
-    });
+    const isNavigationMenuItemEnabled =
```
prastoin: Should be within the above workpsace migration
```ts
// Dynamic assignation
            navigationMenuItem: {
              flatEntityToCreate: flatIndexMetadataToCreate,
              flatEntityToDelete: [],
              flatEntityToUpdate: [],
            }
```

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,578 @@
...
+      );
+
+    const { flatObjectMetadataMaps } =
+      await this.workspaceCacheService.getOrRecompute(workspaceId, [
+        'flatObjectMetadataMaps',
+      ]);
+
+    const folderIdMapping = await this.migrateFavoriteFolders({
+      workspaceId,
+      applicationId: twentyStandardFlatApplication.id,
```
prastoin: ***Remark:*** Unless I'm mistaken this is wrong, all existing favorite folder aren't to be owned by the twenty standard application
We should first compare expected standard favorites within the existing, assign them the expect universal identifier and set their applicationId to the twenty standard application one.
Using `STANDARD_NAVIGATION_MENU_ITEMS` and `STANDARD_OBJECTS`.
The goal of this command is to migrate all existing favorite folder while identifying them either be a twenty-standard owned or workspace-custom owned entity.
Risk here is not being able to track in future `twenty-standard` migration the expected entities

prastoin: You can find related such logic at `packages/twenty-server/src/database/commands/upgrade-version-command/1-16/1-16-identify-view-field-metadata.command.ts`
without the migration complexity though which specific to this command
we could have two seperated command, but it would require making the columns nullable etc

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,624 @@
...
+    const migratedFavoriteIds = await this.migrateFavorites({
+      workspaceId,
+      authContext,
+      workspaceCustomApplicationId: workspaceCustomFlatApplication.id,
+      flatObjectMetadataMaps,
+      flatViewMaps,
+      folderIdMapping,
+    });
+
+    await this.softDeleteMigratedFavorites({
```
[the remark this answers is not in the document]
prastoin: Quite redundant as we either check the feature flag + use identical primary key cross table to assert

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,578 @@
...
+      flatObjectMetadataMaps,
+      folderIdMapping,
+    });
+
+    await this.softDeleteMigratedFavorites({
+      workspaceId,
+      favoriteIds: migratedFavoriteIds,
+    });
+
+    await this.softDeleteMigratedFavoriteFolders({
```
prastoin: same

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,578 @@
...
+      ]);
+
+    const folderIdMapping = await this.migrateFavoriteFolders({
+      workspaceId,
+      applicationId: twentyStandardFlatApplication.id,
+    });
+
+    const migratedFavoriteIds = await this.migrateFavorites({
+      workspaceId,
+      twentyStandardApplicationId: twentyStandardFlatApplication.id,
```
prastoin: same comment than for `migrateFavoriteFolders`
Please let me know

prastoin (commented):
Hey there, unless I'm mistaken we're now incorrectly assigning every migrated entities within the standard whereas some of them could have been created by the user itself post workspace creation
Feel free to reach out if you have any question !

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
prastoin: any thrown exception will stop the whole upgrade yes

prastoin: but that's a good thing to throw here

etiennejouan:
- Add some updates (metadata fetching from cache, remove soft delete and rely on id for idempotence)
- Spot an issue during testing : favoriteFolder does not have workspace/workspaceMember-scope (and should have had). Then when a user creates a folder, it's available for all his workspaceMembers BUT each workspaceMember can insert its own favorite in it. Fixed in the migration.
- Tested : Folder with record fav and view fav, view fav only, record fav only + Idempotency


@abdulrahmancodes @prastoin do not hesitate to take a look ! Thanks

packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-favorites-to-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,704 @@
...
+    existingFlatNavigationMenuItemMaps,
+  }: {
+    workspaceId: string;
+    workspaceCustomApplicationId: string;
+    existingFlatNavigationMenuItemMaps: FlatNavigationMenuItemMaps;
+  }): Promise<{
+    folderIdMapping: Map<string, Map<string, string>>;
+    flatNavigationMenuItemsToCreate: FlatNavigationMenuItem[];
+  }> {
+    const existingFlatNavigationMenuItemIds = Object.values(
```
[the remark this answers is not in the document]
etiennejouan: And it works 🤔 fixing it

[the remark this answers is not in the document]
etiennejouan: Fortunately, universalIdentifier and id are set equals here

etiennejouan (approved):

PR #17480, opened 2026-01-27
Replace manual triggers with command menu items

abdulrahmancodes:


abdulrahmancodes:
Pausing this PR to focus on CommandMenuItem + FrontComponent first.
This PR covers the first part of the tech debt fix: decoupling the manual triggers from WorkflowVersions by introducing CommandMenuItem. When the feature flag is on, manual-trigger actions are shown from CommandMenuItem instead of workflow versions.
The remaining work (run permissions, workflow role, migration) will be done after the FrontComponent integration. I’ll reopen this PR once we’re ready to continue.

PR #17484, opened 2026-01-27
[Fix] display current object label in morph relation picker after rename 2/2

ijreilly:
Following https://github.com/twentyhq/twenty/pull/17209

The goal is to display the updated label. For SingleRecordPickerMenuItemsWithSearch, we did it in two parts to avoid the breaking change by calling `objectLabelSingular` after its addition has been deployed to prod
image

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/search/services/search.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/search/services/search.service.ts
@@ -453,7 +453,9 @@ export class SearchService {
           return {
             recordId: record.id,
             objectNameSingular: objectMetadataItem.nameSingular,
-            objectLabelSingular: objectMetadataItem.labelSingular,
+            objectLabelSingular:
+              objectMetadataItem.standardOverrides?.labelSingular ??
+              objectMetadataItem.labelSingular,
             label: this.getLabelIdentifierValue(
               record,
               objectMetadataItem,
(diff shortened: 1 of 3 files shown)


packages/twenty-server/src/engine/core-modules/search/services/search.service.ts
```diff
@@ -453,7 +453,9 @@ export class SearchService {
           return {
             recordId: record.id,
             objectNameSingular: objectMetadataItem.nameSingular,
-            objectLabelSingular: objectMetadataItem.labelSingular,
+            objectLabelSingular:
+              objectMetadataItem.standardOverrides?.labelSingular ??
```
ijreilly, line 457: @prastoin thats what made me think the update was not taken into account
I find it not intuitive that we need to think to read values from `objectMetadataItem.standardOverrides`
maybe i missed something but i had flushed the cache etc.

martmull (approved):

Weiko (approved):
LGTM

packages/twenty-server/src/engine/core-modules/search/services/search.service.ts
prastoin: Hey there, ok nice
For the moment that's the standard when overriding non owned entity properties, since v1 too

PR #17487, opened 2026-01-27
Add WorkspaceAuthContextMiddleware

Weiko:
## Context
Introduces a middleware that automatically sets the workspace auth context in AsyncLocalStorage for HTTP requests, making it available throughout the request lifecycle without explicit parameter passing.

The motivation behind this change is to reduce boilerplate and simplify the developer experience when working with workspace data in HTTP request handlers.

The Problem (Before)
Every HTTP request handler that needed to access workspace data had to:
- Extract auth-related info from decorators (@AuthWorkspace(), @AuthUserWorkspaceId(), etc.) in controller/resolver and pass down to services
- Build or pass the authContext explicitly (sometimes with type assertion which was flaky)
Then call executeInWorkspaceContext(authContext, async () => { ... })

## Changes
- Add WorkspaceAuthContextMiddleware that extracts auth context from the request and stores it in AsyncLocalStorage
- Register middleware for GraphQL, metadata, and REST routes (runs after hydration middlewares)
- Simplify executeInWorkspaceContext signature: fn is now the first parameter, authContext is optional second
- If authContext is not provided, it's automatically retrieved from the storage (set by middleware)
- Update all callers (~120 files) to use the new parameter order

- Fixes a bug in search where system auth context was used, bypassing RLS feature.

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/auth/middlewares/workspace-auth-context.middleware.ts
+++ b/packages/twenty-server/src/engine/core-modules/auth/middlewares/workspace-auth-context.middleware.ts
@@ -1,0 +1,33 @@
+import { Injectable, type NestMiddleware } from '@nestjs/common';
+
+import { type NextFunction, type Request, type Response } from 'express';
+import { isDefined } from 'twenty-shared/utils';
+
+import { type WorkspaceAuthContext } from 'src/engine/api/common/interfaces/workspace-auth-context.interface';
+
+import { withWorkspaceAuthContext } from 'src/engine/core-modules/auth/storage/workspace-auth-context.storage';
+
+@Injectable()
+export class WorkspaceAuthContextMiddleware implements NestMiddleware {
+  use(req: Request, _res: Response, next: NextFunction) {
+    if (!isDefined(req.workspace)) {
+      next();
+
+      return;
+    }
+
+    const authContext: WorkspaceAuthContext = {
+      user: req.user,
+      workspace: req.workspace,
+      workspaceMemberId: req.workspaceMemberId,
+      workspaceMember: req.workspaceMember,
+      userWorkspaceId: req.userWorkspaceId,
+      apiKey: req.apiKey,
+      application: req.application,
+    } as WorkspaceAuthContext;
+
+    withWorkspaceAuthContext(authContext, () => {
+      next();
+    });
+  }
+}
--- a/packages/twenty-server/src/engine/core-modules/auth/storage/workspace-auth-context.storage.ts
+++ b/packages/twenty-server/src/engine/core-modules/auth/storage/workspace-auth-context.storage.ts
@@ -1,0 +1,25 @@
+import { AsyncLocalStorage } from 'async_hooks';
+
+import { type WorkspaceAuthContext } from 'src/engine/api/common/interfaces/workspace-auth-context.interface';
+
+export const workspaceAuthContextStorage =
+  new AsyncLocalStorage<WorkspaceAuthContext>();
+
+export const getWorkspaceAuthContext = (): WorkspaceAuthContext => {
+  const context = workspaceAuthContextStorage.getStore();
+
+  if (!context) {
+    throw new Error(
+      'Workspace auth context not set. Operations must be wrapped with withWorkspaceAuthContext()',
+    );
+  }
+
+  return context;
+};
+
+export const withWorkspaceAuthContext = <T>(
+  context: WorkspaceAuthContext,
+  fn: () => T | Promise<T>,
+): T | Promise<T> => {
+  return workspaceAuthContextStorage.run(context, fn);
+};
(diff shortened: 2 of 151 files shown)


packages/twenty-server/src/engine/core-modules/auth/middlewares/workspace-auth-context.middleware.ts
```diff
@@ -0,0 +1,33 @@
...
+@Injectable()
+export class WorkspaceAuthContextMiddleware implements NestMiddleware {
+  use(req: Request, _res: Response, next: NextFunction) {
+    if (!isDefined(req.workspace)) {
+      next();
+
+      return;
+    }
+
+    const authContext: WorkspaceAuthContext = {
```
prastoin, line 19: ***Suggestion:*** might the good timing to refactor the `WorkspaceAuthContext` typing
To either or to use a discriminated union narrowed type generic
```ts
import { type AuthContext } from 'src/engine/core-modules/auth/types/auth-context.type';

interface AuthContextWithDefinedWorkspaceProperties {
  user: AuthContext['user'];
  workspace: NonNullable<AuthContext['workspace']>;
  workspaceMetadataVersion?: string;
  workspaceMemberId: AuthContext['workspaceMemberId'];
  workspaceMember: AuthContext['workspaceMember'];
}

interface ApiKeyAuthContext extends Request {
  apiKey: NonNullable<AuthContext['apiKey']>;
  application: null;
  userWorkspaceId: null;
}

interface UserWorkspaceAuthContext extends Request {
  userWorkspaceId: NonNullable<AuthContext['userWorkspaceId']>;
  application: null;
  apiKey: null;
}

interface ApplicationAuthContext extends Request {
  application: NonNullable<AuthContext['application']>;
  userWorkspaceId: null;
  apiKey: null;
}

export type WorkspaceAuthContext =
  | (AuthContextWithDefinedWorkspaceProperties & ApiKeyAuthContext)
  | (AuthContextWithDefinedWorkspaceProperties & UserWorkspaceAuthContext)
  | (AuthContextWithDefinedWorkspaceProperties & ApplicationAuthContext);

```

packages/twenty-server/src/engine/core-modules/auth/storage/workspace-auth-context.storage.ts
```diff
@@ -0,0 +1,25 @@
...
+    throw new Error(
+      'Workspace auth context not set. Operations must be wrapped with withWorkspaceAuthContext()',
+    );
+  }
+
+  return context;
+};
+
+export const withWorkspaceAuthContext = <T>(
+  context: WorkspaceAuthContext,
```
prastoin, line 21: ***Remark:*** Unless I'm mistaken we should be only expecting a subset of the authContext here, as at realtime we won't have everything in the end

packages/twenty-server/src/engine/core-modules/auth/middlewares/workspace-auth-context.middleware.ts
```diff
@@ -0,0 +1,33 @@
...
+      user: req.user,
+      workspace: req.workspace,
+      workspaceMemberId: req.workspaceMemberId,
+      workspaceMember: req.workspaceMember,
+      userWorkspaceId: req.userWorkspaceId,
+      apiKey: req.apiKey,
+      application: req.application,
+    } as WorkspaceAuthContext;
+
+    withWorkspaceAuthContext(authContext, () => {
```
prastoin, line 29: here below typing expect the whole req
```ts
interface UserWorkspaceAuthContext extends Request {
```

packages/twenty-server/src/engine/core-modules/auth/middlewares/workspace-auth-context.middleware.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    const authContext: WorkspaceAuthContext = {
+      user: req.user,
+      workspace: req.workspace,
+      workspaceMemberId: req.workspaceMemberId,
+      workspaceMember: req.workspaceMember,
+      userWorkspaceId: req.userWorkspaceId,
+      apiKey: req.apiKey,
+      application: req.application,
+    } as WorkspaceAuthContext;
```
prastoin, line 27: ***Suggestion:*** In the best of the world we should not have to force cast here

prastoin, line 27: ***Question:*** Would it make sense to throw an error if none of the expected authenticated ids are present ?

packages/twenty-server/src/modules/workspace-member/query-hooks/workspace-member-delete-one.post-query.hook.ts
```diff
@@ -78,6 +77,7 @@ export class WorkspaceMemberDeleteOnePostQueryHook
             withDeleted: true,
           });
         },
+        authContext as WorkspaceAuthContext,
```
prastoin, line 80: forced cast ( should be fixed if we remove Request extension )

prastoin (approved):
As discussed approving already as we're going to create a new more concise PR in order to refactor AuthContext typing

PR #17489, opened 2026-01-27
Add delete and restore event handling for table and board

lucasbordeau:
This PR adds what is required to handle soft-delete and restore SSE events in virtualized table and board.

Restore is not handled in board for now as it requires respecting sorts when inserting record ids.

Since virtualized table is refetching small chunks, we just refetch for now.

The long term goal is to handle event handling without refetching in all main components, and also handle SSE events that have the same origin that the current tab. But for now we implement what is easily doable.

# QA

Delete between table and board (delete only) :

https://github.com/user-attachments/assets/715dd44a-007a-44ab-bf49-5ef039cd57c3

Delete and restore between table and table : 

https://github.com/user-attachments/assets/f2122519-e969-491f-b71c-018d0f85bd86


thomtrp (approved):
LGTM

PR #17523, opened 2026-01-28
Handle SSE event stream reconnection

### LINKED ISSUE #2045: Handle server restarting while an event stream is open
We want to have a mechanism that checks if the server has ended the SSE from its side, because of a pod crash for example.

lucasbordeau:
This PR handles SSE event stream edge cases. 

- When a pod restarts, the front clients have to reconnect to SSE
- When the dev server restarts or is hot reloaded, the front client has to reconnect to SSE
- When redis server restarts or the redis key is cleared for any reason, the server has to recreate the event stream in redis, this can happen when navigating for example.
- Log in / log out flow

With this PR we avoid error messages in the front end due to TTL or pod crash, we implement a resilient way of reconnecting silently.

To avoid DDoSing our servers if pods crash or a full restart of the cluster is made, we evenly space retry attempts to reconnect from all the clients, to avoid n clients reconnection at the same time, we use a random wait time between 0 and a constant max wait time (set to 2 mins for now). This is the cheapest and most effective solution, clients who want to force reconnect have to refresh or navigate to another page.

Fixes https://github.com/twentyhq/core-team-issues/issues/2045


packages/twenty-front/src/modules/sse-db-event/hooks/useSseClient.util.ts
```diff
@@ -1,21 +1,54 @@
...
+import { getSnapshotValue } from '@/ui/utilities/state/utils/getSnapshotValue';
 import { createClient } from 'graphql-sse';
 import { useMemo } from 'react';
+import { useRecoilCallback } from 'recoil';
+import { isNonEmptyArray } from 'twenty-shared/utils';
 import { REACT_APP_SERVER_BASE_URL } from '~/config';
+import { sleep } from '~/utils/sleep';
 
 export const useSseClient = () => {
   const tokenPair = getTokenPair();
```
Devessier, line 7: I'm not 100% sure how the global structure of the application works. I assume the token would change when the user signs out and the signs in with another account.

I think we should kill the SSE connection in this case and recreates one if necessary.

Is there already a strategy regarding this concern?

(Not to mention that `getTokenPair` isn't a reactive function.)

lucasbordeau: Yes indeed, I think we'll have to find a way to store the sse client in a recoil state and plug its lifecycle into sign in / sign out instead of a context.

packages/twenty-front/src/modules/sse-db-event/components/SSEEventStreamEffect.tsx
```diff
@@ -1,22 +1,77 @@
...
+      currentUser.onboardingStatus === OnboardingStatus.COMPLETED &&
+      !shouldDestroyEventStream &&
+      !isNonEmptyString(sseEventStreamId);
+
+    const willDestroyEventStream =
+      isDefined(sseClient) &&
+      !isCreatingSseEventStream &&
+      !isDestroyingEventStream &&
+      isNonEmptyString(sseEventStreamId) &&
+      shouldDestroyEventStream;
```
thomtrp, line 60: would be cleaner with separated booleans. Like `isSseClientAvailable = isDefined(sseClient) && !isCreatingSseEventStream && !isDestroyingEventStream`

packages/twenty-server/src/engine/subscriptions/event-stream.service.ts
```diff
@@ -262,12 +267,38 @@ export class EventStreamService implements OnModuleInit {
...
+  async getStreamData(
     workspaceId: string,
     eventStreamChannelId: string,
   ): Promise<EventStreamData | undefined> {
     const key = this.getEventStreamKey(workspaceId, eventStreamChannelId);
 
     return this.cacheStorageService.get<EventStreamData>(key);
   }
+
+  private async cleanUpDanglingExpiredEventStreams(workspaceId: string) {
```
thomtrp: don't think we need this. We already have:
- ttl on event stream itself
- when a new event arrives,  active streams that are not linked to any streams are cleaned

packages/twenty-server/src/engine/subscriptions/event-stream.service.ts
```diff
@@ -50,6 +50,20 @@ export class EventStreamService implements OnModuleInit {
     );
   }
 
+  async checkIfEventStreamExists({
```
thomtrp: not sure it really brings more than `getStreamData` + `isDefined`

packages/twenty-server/src/engine/core-modules/cache-storage/services/cache-storage.service.ts
```diff
@@ -247,6 +247,33 @@ export class CacheStorageService {
     return totalCount;
   }
 
+  async getAllKeysForPattern(scanPattern: string): Promise<string[]> {
```
thomtrp: can we remove that function?

thomtrp (approved):

PR #17529, opened 2026-01-28
Allow DATE_TIME IS operand to filter on a whole day

### LINKED ISSUE #2027: DATE_TIME IS operand should filter on plain date for user timezone, like DATE.
There is not much need to filter a DATE_TIME on a precise minute, users would like to filter on a given day period instead, in the timezone of the user.

We can take the same logic as DATE type for IS operand.

carbonFibreCode:
Fixes:  https://github.com/twentyhq/core-team-issues/issues/2027

We've replaced the DATE_TIME picker with DATE picker, and changed the logic to filter for complete day period.

https://github.com/user-attachments/assets/ba7e1078-bab3-4c62-a803-d6a851f14b7d


lucasbordeau:
The approach is good. 

But instead of having two code paths to handle both a plain date and a date time value, we want to only handle plain date values for DATE_TIME + IS operand.

We'll create a migration command to migrate any date time string to a plain date string value in core.viewFilter table.

Let's take for granted in your code that viewFilter.value contains a plain date string.

Also during your QA, be careful with operand change, because sometimes changing an operand and keeping the value associated to a previous operand can provoke a crash, we have to take care of cleaning up the value for any operand change.

lucasbordeau:
If you are willing to work on this simple migration command you can do it in another PR, otherwise I'll do it no problem.

carbonFibreCode:
Ohk sure I'll do it in another PR.

packages/twenty-server/src/engine/metadata-modules/view/services/view-query-params.service.ts
```diff
@@ -60,6 +63,30 @@ export class ViewQueryParamsService {
       flatEntityMaps: flatObjectMetadataMaps,
     });
 
+    let timeZone = 'UTC';
+
+    if (currentWorkspaceMemberId) {
```
lucasbordeau: Could you put this into a util or separate function in this service ? `getWorkspaceMemberTimezoneIfAvailable` for example

lucasbordeau (approved):
Very nice PR

carbonFibreCode:
@lucasbordeau  I’ve addressed the CI failure by fixing the failing test; could you please re-enable auto-merge on this pull request when you have a moment?

lucasbordeau:
@carbonFibreCode I want to be sure of the strategy to roll this out before, because there is a data migration involved.

lucasbordeau (changes requested):
Waiting for deploy strategy

carbonFibreCode:
> @carbonFibreCode I want to be sure of the strategy to roll this out before, because there is a data migration involved.

@lucasbordeau ohh yes, we have two PRs for the [issue](https://github.com/twentyhq/core-team-issues/issues/2027)  , and both are dependent on each other.

lucasbordeau:
We will use a feature flag.

So here are the requirements for both PR. 

This PR : 
- Add a feature flag to enable this feature, set it to false by default (but true in the dev seeds)

The data migration PR : 
- The feature flag should be set to true right after the migration, in the same function.
- Be careful to flush the cache in the data migration also

carbonFibreCode:
@lucasbordeau done with feature flags, Please let me know further updates regarding the deployment strategy for this PR

lucasbordeau (approved):
Ready for migration

PR #17540, opened 2026-01-29
Remove versions from logicFunction

charlesBochet:
## Summary

- Remove `latestVersion` and `publishedVersions` columns from `LogicFunction` entity
- Remove version parameter from logic function execution, build, and source code retrieval
- Update all related services, DTOs, utilities, and tests
- Add database migration to drop the version columns


packages/twenty-server/src/modules/workflow/workflow-builder/workflow-version-step/workflow-version-step-operations.workspace-service.ts
```diff
@@ -883,33 +873,10 @@ export class WorkflowVersionStepOperationsWorkspaceService {
...
-              logicFunctionVersion: 'draft',
-            },
-          },
-        };
-      }
-      default: {
-        return step;
-      }
-    }
+    return step;
```
martmull, line 925: you need to create new serverlesses from the published version ones

packages/twenty-server/src/modules/workflow/workflow-builder/workflow-version-step/workflow-version-step-operations.workspace-service.ts
```diff
@@ -618,7 +610,6 @@ export class WorkflowVersionStepOperationsWorkspaceService {
         const newLogicFunction =
           await this.logicFunctionService.duplicateLogicFunction({
```
martmull, line 656: you need to double check this mecanism

packages/twenty-server/src/modules/workflow/workflow-status/jobs/workflow-statuses-update.job.ts
```diff
@@ -179,25 +175,10 @@ export class WorkflowStatusesUpdateJob {
...
-              `Fail to publish logic function ${logicFunction.id}. Latest version is null`,
-              WorkflowVersionStepExceptionCode.CODE_STEP_FAILURE,
-            );
-          }
-
-          newStepSettings.input.logicFunctionVersion =
-            logicFunction.latestVersion;
-
-          newStep.settings = newStepSettings;
+          await this.logicFunctionService.publishOneLogicFunctionOrFail(
```
martmull, line 178: publish does not mean anything anymore without logic function versionning, we should rename

packages/twenty-server/src/modules/workflow/workflow-status/jobs/workflow-statuses-update.job.ts
martmull, line 153: should be renamed (publish has no sense now)

martmull (changes requested):

packages/twenty-server/src/modules/workflow/workflow-builder/workflow-version-step/workflow-version-step-operations.workspace-service.ts
charlesBochet: indeed, will do in next PR

packages/twenty-server/src/modules/workflow/workflow-builder/workflow-version-step/workflow-version-step-operations.workspace-service.ts
charlesBochet: ok, good point

packages/twenty-server/src/modules/workflow/workflow-status/jobs/workflow-statuses-update.job.ts
charlesBochet: agree

PR #17541, opened 2026-01-29
File storage refactor - Switch to applicationUniversalIdentifier

etiennejouan:


packages/twenty-server/src/engine/core-modules/file/file-upload/services/file-upload.service.ts
charlesBochet, line 225: we could pass applicationUniversalId here too to avoid fetching internally and convert twice?

charlesBochet (approved):

packages/twenty-server/src/engine/core-modules/file/file-upload/services/file-upload.service.ts
etiennejouan: `uploadFilesFieldFile` is only called by the file upload resolver, which doesn't know the `applicationUniversalId`. So it should fetch the `applicationUniversalId` itself. I thought it was more or less the same.

PR #17556, opened 2026-01-29
Fix standard object translations not saving for non-English locales

### LINKED ISSUE #17382: Label from standard objects cannot be edited when the language is Finnish
A customer reported

"H! I am trying to update field and object names into Finnish, but after saving, they revert back to the English names."

I confirm that I can edit the label of Companies (for ex) when the workspace language is English but it is indeed not saved when the language is Finnish. I reproduced on another workspace than the customer's one.
I did not test for 
- other languages
- labels from custom objects
- label from custom and standard fields

Thanks!

Front logo [Front conversations](https://app.frontapp.com/open/top_3zx6vq)

neo773:
https://github.com/user-attachments/assets/f3d54bc9-e780-44a3-ae9d-015300d949e7

/closes https://github.com/twentyhq/twenty/issues/17382


packages/twenty-server/src/engine/metadata-modules/flat-object-metadata/utils/sanitize-raw-update-object-input.ts
```diff
@@ -57,52 +62,44 @@ export const sanitizeRawUpdateObjectInput = ({
...
 
-  const standardOverrides =
-    OBJECT_METADATA_STANDARD_OVERRIDES_PROPERTIES.reduce(
-      (standardOverrides, property) => {
-        const propertyValue = updatedEditableObjectProperties[property];
+  const safeLocale = locale ?? SOURCE_LOCALE;
 
-        const isPropertyUpdated =
-          updatedEditableObjectProperties[property] !== undefined;
+  let standardOverrides = existingFlatObjectMetadata.standardOverrides;
```
prastoin, line 67: ***Request:*** Please avoid any mutation that could have a side effect caller side
Not strong opinion on keeping the reduce but we should avoid any mutation

packages/twenty-server/src/engine/metadata-modules/flat-object-metadata/utils/sanitize-raw-update-object-input.ts
```diff
@@ -57,52 +62,44 @@ export const sanitizeRawUpdateObjectInput = ({
...
-        if (!isPropertyUpdated) {
-          return standardOverrides;
-        }
-        delete updatedEditableObjectProperties[property];
+  for (const property of OBJECT_METADATA_STANDARD_OVERRIDES_PROPERTIES) {
+    const propertyValue = updatedEditableObjectProperties[property];
+    const isPropertyUpdated = propertyValue !== undefined;
 
-        if (propertyValue === existingFlatObjectMetadata[property]) {
-          if (
```
prastoin, line 74: ***Praise:*** way better naming

packages/twenty-server/src/engine/metadata-modules/object-metadata/constants/object-metadata-translatable-properties.constant.ts
```diff
@@ -0,0 +1,8 @@
+export const OBJECT_METADATA_TRANSLATABLE_PROPERTIES = [
+  'labelSingular',
+  'labelPlural',
+  'description',
+] as const;
```
prastoin: ***Request:*** `as const satisfies (keyof ObjectMetadataEntity)[]`

packages/twenty-server/src/engine/metadata-modules/object-metadata/object-metadata.service.ts
```diff
@@ -56,9 +57,11 @@ export class ObjectMetadataService extends TypeOrmQueryService<ObjectMetadataEnt
   async updateOneObject({
     updateObjectInput,
     workspaceId,
+    locale,
   }: {
     workspaceId: string;
     updateObjectInput: UpdateOneObjectInput;
+    locale?: keyof typeof APP_LOCALES;
```
prastoin: ***Polish:*** use type `AppLocale`

prastoin (commented):
Hey there, thanks again for taking this !
Could you please explain the issue you've identified in the PR desc please and the error returned in the first place ?
Not sure to understand why translation is involved here
From my understangin they should be mapped after data extraction not insertion

packages/twenty-server/src/engine/metadata-modules/object-metadata/utils/remove-standard-override.util.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export const removeStandardOverride = ({
+  overrides,
+  property,
+  locale,
+}: {
+  overrides: ObjectStandardOverridesDTO | null;
+  property: ObjectMetadataStandardOverridesProperties;
+  locale: keyof typeof APP_LOCALES;
+}): ObjectStandardOverridesDTO | null => {
```
prastoin, line 16: ***Remark:*** We should not use DTO in inputs here
Also there's atm no `translations` property in any production `standardOverrides`

prastoin (commented):
Hey there, we should not store translations in database
Unless I'm mistaken the fix should be at `resolveFieldMetadataStandardOverride` and `resolveObjectMetadataStandardOverride` lvl
Or maybe at insertion lvl but rather a dynamic removal of the override property

charlesBochet:
I'm not sure about this one either. We haven't planed to translate overrides yet, like we don't translate customObjects names either.

If we want to do so, we should start offering a translation mecanism in the UX.

What we want:
- whatever the user language is, whatever he types in the label is what should be displayed.

neo773:
Closing this to reduce noise as I cannot find the bandwidth and this is more complicated than I initially thought.
Maybe @prastoin can take this one?

PR #17573, opened 2026-01-30
[GroupBy] Fix order by date granularity

ijreilly:
https://discord.com/channels/1130383047699738754/1466472357496623165/1466472357496623165

before
image

after
image

### DIFF
--- a/packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/resolvers/utils/get-group-by-order-expression.util.ts
+++ b/packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/resolvers/utils/get-group-by-order-expression.util.ts
@@ -1,0 +1,41 @@
+import { ObjectRecordGroupByDateGranularity } from 'twenty-shared/types';
+import { isDefined } from 'twenty-shared/utils';
+
+import { type GroupByField } from 'src/engine/api/graphql/graphql-query-runner/group-by/resolvers/types/group-by-field.types';
+import { getGroupByExpression } from 'src/engine/api/graphql/graphql-query-runner/group-by/resolvers/utils/get-group-by-expression.util';
+import { isGroupByDateField } from 'src/engine/api/graphql/graphql-query-runner/group-by/resolvers/utils/is-group-by-date-field.util';
+import { isGroupByRelationField } from 'src/engine/api/graphql/graphql-query-runner/group-by/resolvers/utils/is-group-by-relation-field.util';
+
+export const getGroupByOrderExpression = ({
+  groupByField,
+  columnNameWithQuotes,
+}: {
+  groupByField: GroupByField;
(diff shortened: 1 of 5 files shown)


packages/twenty-server/src/modules/dashboard/chart-data/utils/compare-dimension-values.util.ts
```diff
@@ -49,10 +49,11 @@ export const compareDimensionValues = ({
     direction === 'ASC' ? comparison : -comparison;
 
   if (isDefined(fieldType)) {
-    if (
-      isFieldMetadataDateKind(fieldType) &&
-      !isCyclicalDateGranularity(dateGranularity)
-    ) {
+    if (isFieldMetadataDateKind(fieldType)) {
+      if (isCyclicalDateGranularity(dateGranularity)) {
+        return 0;
```
ijreilly, line 54: @bosiraphael not sure what should be done here, I ll let you update that

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/resolvers/utils/get-group-by-order-expression.util.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  const groupByExpression = getGroupByExpression({
+    groupByField,
+    columnNameWithQuotes,
+  });
+
+  switch (dateGranularity) {
+    case ObjectRecordGroupByDateGranularity.DAY_OF_THE_WEEK:
+      return `CASE ${groupByExpression} WHEN 'Monday' THEN 1 WHEN 'Tuesday' THEN 2 WHEN 'Wednesday' THEN 3 WHEN 'Thursday' THEN 4 WHEN 'Friday' THEN 5 WHEN 'Saturday' THEN 6 WHEN 'Sunday' THEN 7 END`;
```
bosiraphael: We should take the firstDayOfTheWeek defined by the user here

packages/twenty-server/src/modules/dashboard/chart-data/utils/compare-dimension-values.util.ts
bosiraphael: It's good :) Values are already sorted by the group by

ijreilly: Yes indeed, but is that not always the case? why do we need that util at all ? except for potential localeCompare for currency

bosiraphael: Because some type of sorts are not done by the group by: select position sort, sor by aggregate sum etc.

bosiraphael (approved):
LGTM

PR #17582, opened 2026-01-30
Fix SSE for workflow show page 

thomtrp:
- Add SSE to workflow show page. Listens at workflow versions
- Add a tool for creating trigger
- Ensure step id is not used in params
- Make AI adding step linked to the last node by default

https://github.com/user-attachments/assets/41ee1835-9716-450e-82e8-54e1b63bd1ef

### DIFF
--- a/packages/twenty-front/src/modules/apollo/optimistic-effect/utils/triggerUpdateRelationsOptimisticEffect.ts
+++ b/packages/twenty-front/src/modules/apollo/optimistic-effect/utils/triggerUpdateRelationsOptimisticEffect.ts
@@ -37,7 +37,6 @@ type TriggerUpdateRelationsOptimisticEffectArgs = {
     ObjectPermissions & { objectMetadataId: string }
   >;
   upsertRecordsInStore: (props: { partialRecords: ObjectRecord[] }) => void;
-  updatedFieldNames?: string[];
 };
 
 export const triggerUpdateRelationsOptimisticEffect = ({
@@ -48,7 +47,6 @@ export const triggerUpdateRelationsOptimisticEffect = ({
   objectMetadataItems,
   objectPermissionsByObjectMetadataId,
   upsertRecordsInStore,
-  updatedFieldNames,
 }: TriggerUpdateRelationsOptimisticEffectArgs) => {
   const isDeletion =
     isDefined(updatedSourceRecord) &&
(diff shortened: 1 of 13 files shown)


packages/twenty-front/src/modules/apollo/optimistic-effect/utils/triggerUpdateRelationsOptimisticEffect.ts
```diff
@@ -60,6 +62,13 @@ export const triggerUpdateRelationsOptimisticEffect = ({
       return;
     }
 
+    if (
+      isDefined(updatedFieldNames) &&
+      !updatedFieldNames.includes(fieldMetadataItemOnSourceRecord.name)
+    ) {
+      return;
```
charlesBochet: why this change?

thomtrp: you are right, it needed to be double check by you or @lucasbordeau

In the case of relations, we do not return the full relation objet. Issue was that workflow version was returned without the workflow after an update. And apollo updated the workflow accordingly.

This line means that we should care about relations only if the fields was meant to be updated. If not, we assume we simply did not fetch the relation

packages/twenty-front/src/modules/sse-db-event/hooks/useTriggerOptimisticEffectFromSseUpdateEvents.ts
charlesBochet: check

charlesBochet, line 80: if true, use computeOptimisticRecordFromInput

packages/twenty-front/src/modules/apollo/optimistic-effect/utils/triggerUpdateRelationsOptimisticEffect.ts
```diff
@@ -60,6 +62,13 @@ export const triggerUpdateRelationsOptimisticEffect = ({
       return;
     }
 
+    if (
```
charlesBochet: remove

charlesBochet (approved):

PR #17589, opened 2026-01-30
feat: add event logs UI and backend query support

FelixMalfait:
## Summary
Add a new Event Logs feature to the Settings > Security section that allows querying and viewing event logs from ClickHouse.

### Frontend
- Add `SettingsEventLogs` page with table selector, filters, and results table
- Add `EventLogFilters` component with date range, event type, and user filters
- Add `EventLogResultsTable` with pagination support
- Add `EventLogTableSelector` for switching between PAGEVIEW, OBJECT_EVENT, and WORKSPACE_EVENT tables
- Update `SettingsRoutes` with new EventLogs path
- Update `SettingsSecurity` page with link to Event Logs

### Backend
- Add `EventLogsModule` with resolver and service
- Add GraphQL types for event log queries (`EventLogQueryInput`, `EventLogQueryResult`, `EventLogRecord`)
- Add ClickHouse query runners for event log data
- Add database migration for ClickHouse data source type enum

## Testing
- Navigate to Settings > Security > Event Logs
- Select different tables (Pageview, Object Event, Workspace Event)
- Apply filters and verify pagination works correctly

Co-Authored-By: Warp <agent@warp.dev>

### DIFF
--- a/packages/twenty-front/src/generated-metadata/graphql.ts
+++ b/packages/twenty-front/src/generated-metadata/graphql.ts
@@ -1368,6 +1368,90 @@ export type EmailsConfiguration = {
   configurationType: WidgetConfigurationType;
 };
 
+export type EventLogDateRangeInput = {
+  end?: InputMaybe<Scalars['DateTime']>;
+  start?: InputMaybe<Scalars['DateTime']>;
+};
+
+export type EventLogFiltersInput = {
+  /** Filter by timestamp range */
+  dateRange?: InputMaybe<EventLogDateRangeInput>;
+  /** Filter by event type */
+  eventType?: InputMaybe<Scalars['String']>;
+  /** Filter by object metadata ID (objectEvent only) */
+  objectMetadataId?: InputMaybe<Scalars['String']>;
+  /** Filter by record ID (objectEvent only) */
+  recordId?: InputMaybe<Scalars['String']>;
+  /** Filter by user ID */
+  userId?: InputMaybe<Scalars['String']>;
+};
+
+/** Order direction for event logs */
+export enum EventLogOrderByDirection {
+  ASC = 'ASC',
+  DESC = 'DESC'
+}
+
+/** Fields available for ordering event logs */
+export enum EventLogOrderByField {
+  EVENT = 'EVENT',
+  TIMESTAMP = 'TIMESTAMP',
+  USER_ID = 'USER_ID'
+}
+
+export type EventLogOrderByInput = {
+  direction?: EventLogOrderByDirection;
+  field?: EventLogOrderByField;
+};
+
+export type EventLogQueryInput = {
+  /** Filters to apply to the query */
+  filters?: InputMaybe<EventLogFiltersInput>;
+  /** Maximum number of records to return */
+  limit?: InputMaybe<Scalars['Int']>;
+  /** Number of records to skip */
+  offset?: InputMaybe<Scalars['Int']>;
+  /** Order by configuration */
+  orderBy?: InputMaybe<EventLogOrderByInput>;
+  /** The table to query */
+  table: EventLogTable;
+};
+
+export type EventLogQueryResult = {
+  __typename?: 'EventLogQueryResult';
+  /** Whether there are more records to fetch */
+  hasNextPage: Scalars['Boolean'];
+  /** The event log records */
+  records: Array<EventLogRecord>;
+  /** Total count of records matching the query */
+  totalCount: Scalars['Int'];
+};
+
+export type EventLogRecord = {
+  __typename?: 'EventLogRecord';
+  /** Event type or page name */
+  event: Scalars['String'];
+  /** Is custom object (objectEvent only) */
+  isCustom?: Maybe<Scalars['Boolean']>;
+  /** Object metadata ID (objectEvent only) */
+  objectMetadataId?: Maybe<Scalars['String']>;
+  /** Additional event properties */
+  properties?: Maybe<Scalars['JSON']>;
+  /** Record ID (objectEvent only) */
+  recordId?: Maybe<Scalars['String']>;
+  /** Timestamp of the event */
+  timestamp: Scalars['DateTime'];
+  /** User ID who triggered the event */
+  userId?: Maybe<Scalars['String']>;
+};
+
+/** Available event log tables in ClickHouse */
+export enum EventLogTable {
+  OBJECT_EVENT = 'OBJECT_EVENT',
+  PAGEVIEW = 'PAGEVIEW',
+  WORKSPACE_EVENT = 'WORKSPACE_EVENT'
+}
+
 export type EventSubscription = {
   __typename?: 'EventSubscription';
   eventStreamId: Scalars['String'];
@@ -3563,6 +3647,9 @@ export type Query = {
   getCoreViews: Array<CoreView>;
   getDatabaseConfigVariable: ConfigVariable;
   getEmailingDomains: Array<EmailingDomain>;
+  /** Get available event log tables */
+  getEventLogTables: Array<EventLogTable>;
+  getFrontComponentCode: FrontComponentCode;
   getIndicatorHealthStatus: AdminPanelHealthServiceData;
   getLogicFunctionSourceCode?: Maybe<Scalars['JSON']>;
   getMeteredProductsUsage: Array<BillingMeteredProductUsageOutput>;
@@ -3595,6 +3682,8 @@ export type Query = {
   object: Object;
   objects: ObjectConnection;
   pieChartData: PieChartDataOutput;
+  /** Query event logs from ClickHouse */
+  queryEventLogs: EventLogQueryResult;
   search: SearchResultConnection;
   skill?: Maybe<Skill>;
   skills: Array<Skill>;
@@ -3911,6 +4000,11 @@ export type QueryObjectsArgs = {
 
 export type QueryPieChartDataArgs = {
   input: PieChartDataInput;
+};
+
+
+export type QueryQueryEventLogsArgs = {
+  input: EventLogQueryInput;
 };
 
 
--- a/packages/twenty-front/src/generated/graphql.ts
+++ b/packages/twenty-front/src/generated/graphql.ts
@@ -1335,6 +1335,90 @@ export type EmailsConfiguration = {
   configurationType: WidgetConfigurationType;
 };
 
+export type EventLogDateRangeInput = {
+  end?: InputMaybe<Scalars['DateTime']>;
+  start?: InputMaybe<Scalars['DateTime']>;
+};
+
+export type EventLogFiltersInput = {
+  /** Filter by timestamp range */
+  dateRange?: InputMaybe<EventLogDateRangeInput>;
+  /** Filter by event type */
+  eventType?: InputMaybe<Scalars['String']>;
+  /** Filter by object metadata ID (objectEvent only) */
+  objectMetadataId?: InputMaybe<Scalars['String']>;
+  /** Filter by record ID (objectEvent only) */
+  recordId?: InputMaybe<Scalars['String']>;
+  /** Filter by user ID */
+  userId?: InputMaybe<Scalars['String']>;
+};
+
+/** Order direction for event logs */
+export enum EventLogOrderByDirection {
+  ASC = 'ASC',
+  DESC = 'DESC'
+}
+
+/** Fields available for ordering event logs */
+export enum EventLogOrderByField {
+  EVENT = 'EVENT',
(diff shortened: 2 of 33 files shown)


packages/twenty-front/src/generated-metadata/graphql.ts
```diff
@@ -1371,6 +1368,90 @@ export type EmailsConfiguration = {
   configurationType: WidgetConfigurationType;
 };
 
+export type EventLogDateRangeInput = {
+  end?: InputMaybe<Scalars['DateTime']>;
+  start?: InputMaybe<Scalars['DateTime']>;
+};
+
+export type EventLogFiltersInput = {
```
charlesBochet, line 1376: EventLogFilterInput

packages/twenty-front/src/generated-metadata/graphql.ts
```diff
@@ -1371,6 +1368,90 @@ export type EmailsConfiguration = {
...
+}
+
+export type EventLogOrderByInput = {
+  direction?: EventLogOrderByDirection;
+  field?: EventLogOrderByField;
+};
+
+export type EventLogQueryInput = {
+  /** Filters to apply to the query */
+  filters?: InputMaybe<EventLogFiltersInput>;
```
charlesBochet, line 1409: filter

packages/twenty-front/src/pages/settings/security/event-logs/components/EventLogDatePickerInput.tsx
```diff
@@ -0,0 +1,185 @@
...
+import { OverlayContainer } from '@/ui/layout/overlay/components/OverlayContainer';
+import { useListenClickOutside } from '@/ui/utilities/pointer-event/hooks/useListenClickOutside';
+import { isDefined } from 'twenty-shared/utils';
+import { IconCalendar } from 'twenty-ui/display';
+
+const StyledInputContainer = styled.div`
+  position: relative;
+`;
+
+const StyledInput = styled.div<{ hasValue: boolean }>`
```
charlesBochet, line 21: I'm surprised we have to create a new div with much style, we should be able to reuse components from twenty ui here

packages/twenty-front/src/generated-metadata/graphql.ts
```diff
@@ -1371,6 +1368,90 @@ export type EmailsConfiguration = {
...
+  /** Maximum number of records to return */
+  limit?: InputMaybe<Scalars['Int']>;
+  /** Number of records to skip */
+  offset?: InputMaybe<Scalars['Int']>;
```
FelixMalfait, line 1413: How hard would it be to adopt cursor-page pagination? Would it behave the same in ClickHouse in terms or performance or is there a downside to switching to cursor-based? I think Cursor-based would be nicer because new events can come in quite frequently so pagination would break easily (showing duplicates or incorrect order)

packages/twenty-front/src/generated-metadata/graphql.ts
```diff
@@ -3566,6 +3647,9 @@ export type Query = {
   getCoreViews: Array<CoreView>;
   getDatabaseConfigVariable: ConfigVariable;
   getEmailingDomains: Array<EmailingDomain>;
+  /** Get available event log tables */
+  getEventLogTables: Array<EventLogTable>;
+  getFrontComponentCode: FrontComponentCode;
```
FelixMalfait, line 3652: Not sure why getFrontComponentCode shows up in this PR?

packages/twenty-front/src/generated/graphql.ts
```diff
@@ -3467,6 +3523,8 @@ export type Query = {
   getCoreViews: Array<CoreView>;
   getDatabaseConfigVariable: ConfigVariable;
   getEmailingDomains: Array<EmailingDomain>;
+  /** Get available event log tables */
```
FelixMalfait, line 3526: Let's remove the 'description' field we added since it doesn't seem to be used elsewhere

packages/twenty-front/src/generated/graphql.ts
```diff
@@ -3497,6 +3555,8 @@ export type Query = {
   object: Object;
   objects: ObjectConnection;
   pieChartData: PieChartDataOutput;
+  /** Query event logs from ClickHouse */
+  queryEventLogs: EventLogQueryResult;
```
FelixMalfait, line 3559: Same let's remove description field

packages/twenty-server/src/engine/core-modules/event-logs/event-logs.service.ts
```diff
@@ -0,0 +1,276 @@
...
+        properties: record.properties,
+        recordId: record.recordId,
+        objectMetadataId: record.objectMetadataId,
+        isCustom: record.isCustom,
+      };
+    });
+  }
+
+  getAvailableTables(): EventLogTable[] {
+    return Object.values(EventLogTable);
```
FelixMalfait, line 220: Why is this different from allowed tables?  do we need both?

packages/twenty-server/src/engine/core-modules/event-logs/index.ts
```diff
@@ -0,0 +1,3 @@
+export { EventLogsModule } from './event-logs.module';
+export { EventLogsService } from './event-logs.service';
```
FelixMalfait, line 2: We don't use barrel export like that in twenty server. Make sure we always have one export per file and that we import directly from it

packages/twenty-server/src/engine/core-modules/event-logs/event-logs.service.ts
```diff
@@ -0,0 +1,276 @@
...
+  }
+
+  private async resolveWorkspaceMemberFilter(
```
FelixMalfait, line 223: Check we don't have userId or at least userWorkspaceId on the workspaceMember object on the frontend side? It would be easier if we did, then we wouldn't need that (or we would need a simple version)

I'm unclear whether we should push userId, userWorkspaceId, or WorkspaceMemberId on the clickhouse side. From this code it looks like it's userId so we might as well stay with this, but if we were to start from scratch today I would have used userWorkspaceId

packages/twenty-server/src/engine/core-modules/event-logs/dtos/event-log-table.enum.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export enum EventLogTable {
+  WORKSPACE_EVENT = 'workspaceEvent',
+  PAGEVIEW = 'pageview',
+  OBJECT_EVENT = 'objectEvent',
+}
+
+registerEnumType(EventLogTable, {
+  name: 'EventLogTable',
+  description: 'Available event log tables in ClickHouse',
```
FelixMalfait, line 11: Consistency? Should we add it everywhere in another PR? Maybe not worth it... TBC

packages/twenty-server/src/engine/core-modules/event-logs/dtos/event-log-query.input.ts
```diff
@@ -0,0 +1,37 @@
...
+import { EventLogTable } from './event-log-table.enum';
+
+@InputType()
+export class EventLogQueryInput {
+  @Field(() => EventLogTable, { description: 'The table to query' })
+  table: EventLogTable;
+
+  @Field(() => EventLogFiltersInput, {
+    nullable: true,
+    description: 'Filters to apply to the query',
```
FelixMalfait, line 14: Same question as everywhere else for the description (it does look nice but should we consider consistency?)

packages/twenty-server/src/engine/core-modules/event-logs/dtos/event-log-order-by.input.ts
```diff
@@ -0,0 +1,31 @@
+import { Field, InputType, registerEnumType } from '@nestjs/graphql';
+
+export enum EventLogOrderByField {
+  TIMESTAMP = 'timestamp',
```
FelixMalfait, line 4: Check if it shouldn't be CAPITAL LETTERS by convention, I feel its usually the case for enums

packages/twenty-server/src/engine/core-modules/event-logs/dtos/event-log-order-by.input.ts
```diff
@@ -0,0 +1,31 @@
...
+  description: 'Order direction for event logs',
+});
+
+@InputType()
+export class EventLogOrderByInput {
+  @Field(() => EventLogOrderByField, { defaultValue: EventLogOrderByField.TIMESTAMP })
+  field: EventLogOrderByField;
+
+  @Field(() => EventLogOrderByDirection, { defaultValue: EventLogOrderByDirection.DESC })
+  direction: EventLogOrderByDirection;
```
FelixMalfait, line 30: Is this used/exposed in the UI? What kind of query would be cheap vs costly in Clickhouse? (look at index/structure of table to understand what we really want to allow)

packages/twenty-server/src/engine/core-modules/event-logs/event-logs.service.ts
```diff
@@ -0,0 +1,276 @@
...
+    const { query, params } = buildClickHouseSelectQuery({
+      tableName: input.table,
+      filter,
+      orderBy,
+      limit: limit + 1,
+      offset,
+      workspaceId,
+    });
+
+    this.logger.debug(`Executing ClickHouse query: ${query}`);
```
FelixMalfait, line 79: Remove ALL debug logs added in this branch

packages/twenty-server/src/engine/api/clickhouse-query-runners/clickhouse-query-runners.module.ts
```diff
@@ -0,0 +1,22 @@
...
+    ClickHouseFindOneQueryRunnerService,
+    ClickHouseGroupByQueryRunnerService,
+  ],
+  exports: [
+    ClickHouseFindManyQueryRunnerService,
+    ClickHouseFindOneQueryRunnerService,
+    ClickHouseGroupByQueryRunnerService,
+  ],
+})
+export class ClickHouseQueryRunnersModule {}
```
FelixMalfait, line 22: It doesn't feel like it deserves to be here in the api folder at this point ; let's move it to be in "core-modules" maybe?

Also should this replace "clickHouse.service.ts"? There was already a wrapper for select/insert and maybe more there. We should not keep both in parallel!

packages/twenty-server/src/engine/api/clickhouse-query-runners/utils/clickhouse-query-builder.util.ts
```diff
@@ -0,0 +1,207 @@
...
+    filter,
+    orderBy,
+    limit = 100,
+    offset = 0,
+    workspaceId,
+  } = options;
+
+  const params: Record<string, unknown> = {};
+
+  // Build SELECT clause
```
FelixMalfait, line 58: Remove ALL useless comments

FelixMalfait:
@claude can you please address all comments made on this PR and push changes to this branch. It seems this PR is quite dirty with un-necessary code, I want you to take a step back and remove all un-necessary code to make things as clean as possible.

Also, this PR introduces a new table style in the frontend. While it's nice that columns are resizable (a must for logs) you should have re-used the header style from other settings table and implemented auto-load pagination (not next/previous) as this is the standard pattern. We try to re-use patterns and existing components as much as possible, rather than introducing new code.

FelixMalfait:
@claude you couldn't execute the previous plan because you were limited to 50 steps. I increased to 200. Please try again

FelixMalfait:
@claude continue on the branch you created claude/pr-17589-20260202-1704 and add cursor-page pagination

Don't show "loading ..." but use proper shimmers like other places in the app.

Also your record count and refresh button will never be visible if they are at the bottom while you implemented autoload on scroll. 

Also I thought we used getEventLogTables to get tables from the backend but it seems we don't? Cleanup and remove useless code then

In EventLogDatePickerInput do we need the UseEffect? It's usually forbidden in this codebase (and useRef is not recommended either). If we really do need, try to find similar patterns, for example isolating the effect in a dedicated component

Why the changes to dataSource model / migration? Remove them. 

EventLogTable: all caps.

queryEventLogs: this doesn't seem like the standard naming pattern. Maybe getEventLogs is more common? Or just eventLogs would be even better?

PUSH ON THE SAME BRANCH /!\

FelixMalfait:
https://www.highlight.io/blog/how-we-built-logging-with-clickhouse -> Cursor based pagination. We can use the timestamp and duplicate on frontend for now probably it's good enough if the timestamp has ms precision (maybe not if it's second-precision). If we wanted to go further we'd add uuid to the table like in the attached article

PR #17607, opened 2026-02-01
feat(imap): Initial IMAP Service Construction

Loewwe:
Lays the groundwork for the IMAP integration bounty ($2,500). Implements basic NestJS service structure for mail synchronization.


FelixMalfait:
IMAP is already in the codebase

PR #17614, opened 2026-02-02
Fix: Favorite records not showing in sidebar

abdulrahmancodes:
## Problem
Favorite records (`NavigationMenuItems`) were not showing in the sidebar under Favorites as the BE was returning `null` for `targetRecordIdentifier`.
## Root cause
The `targetRecordIdentifier` resolver used `context.req`, which does not have the typed `WorkspaceAuthContext` shape. 
## Fix
Use `getWorkspaceAuthContext()` instead of `context.req` so the resolver receives the typed auth context.


etiennejouan (approved):
LGTM

PR #17619, opened 2026-02-02
Revert "fix: prevent default object re-selection in relation field fo…

ijreilly:
I suggest to revert [this PR](https://github.com/twentyhq/twenty/pull/17313) as it had already been fixed [in this PR](https://github.com/twentyhq/twenty/pull/17209/changes) (which fixes more than it says in its title). Issue was tracked by [this ticket](https://github.com/twentyhq/core-team-issues/issues/2049) not very explicit - sorry, my fault.
Code added in the PR does not add value


etiennejouan (approved):
Tested and it works 👍 Thanks for catching this
I should have rebase before testing the other PR, or tested on main to be sure it needed to be fixed.

PR #17623, opened 2026-02-02
[Dashboards] Auto focus effect on standalone widget

ehconitin:
closes https://discord.com/channels/1130383047699738754/1467809255771078719

before - 

https://github.com/user-attachments/assets/3af6dbfd-358c-44fd-9419-338791fd5cfc

after - 

https://github.com/user-attachments/assets/2e8f89f8-5989-4e62-9bc0-6ddd47c0d421


ehconitin:
~~ah noticed a small issue -- when the widget has no command menu page -- it resets to root -- and that jump is clearly visible now - I am thinking of adding a new state that helps us 'skip' this reset~~

seems expensive for such short gain - let me know if its an issue

Devessier (approved):
Good for me! Thanks!

PR #17628, opened 2026-02-02
fix: Claude workflow — prevent bot self-cancellation and add CI permissions

FelixMalfait:
## Summary
- **Set `cancel-in-progress: false`** — Claude's own comments (e.g. error reports) were triggering new workflow runs via `issue_comment`, which cancelled the in-progress Claude run before the new run's `if` condition could skip it. Disabling cancel-in-progress prevents this.
- **Filter out Bot users in `if` conditions** — Added `github.event.comment.user.type != 'Bot'` so Claude's own comments don't start job evaluation at all.
- **Add back `additional_permissions: actions: read`** — Lets Claude read CI failure logs to help debug broken builds.
- **Remove `discussion_comment` trigger and job** — `claude-code-action` doesn't support this event type yet (throws `Unsupported event type: discussion_comment`). Listed as planned in their security.md.

## Context
Claude runs were consistently failing with `The operation was canceled` because:
1. Claude posts an error/status comment back to the issue
2. That comment triggers a new `issue_comment` workflow run on the same issue number
3. The concurrency group cancels the in-progress run before the new run evaluates its `if` and skips

## Test plan
- [ ] Tag `@claude` on an issue — should run to completion without being cancelled
- [ ] Verify Claude's own reply comments don't trigger new workflow runs

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -9,21 +9,19 @@ on:
     types: [submitted]
   issues:
     types: [opened, assigned]
-  discussion_comment:
-    types: [created]
   repository_dispatch:
     types: [claude-core-team-issues]
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.issue.number || github.event.pull_request.number || github.event.client_payload.issue_number || github.event.discussion.number }}
-  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ github.event.issue.number || github.event.pull_request.number || github.event.client_payload.issue_number }}
+  cancel-in-progress: false
 
 jobs:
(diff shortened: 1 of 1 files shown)


FelixMalfait:
@claude can you please address all comments made on this PR and push changes to this branch. It seems this PR is quite dirty with un-necessary code, I want you to take a step back and remove all un-necessary code to make things as clean as possible.

Also, this PR introduces a new table style in the frontend. While it's nice that columns are resizable (a must for logs) you should have re-used the header style from other settings table and implemented auto-load pagination (not next/previous) as this is the standard pattern. We try to re-use patterns and existing components as much as possible.

PR #17629, opened 2026-02-02
Improve qualification of rest errors

ijreilly:
Will help close [sentry](https://twenty-v7.sentry.io/issues/6827612895/?environment=prod&environment=prod-eu&project=4507072499810304&query=is%3Aunresolved%20%21issue.type%3A%5Bperformance_consecutive_db_queries%2Cperformance_consecutive_http%2Cperformance_file_io_main_thread%2Cperformance_db_main_thread%2Cperformance_n_plus_one_db_queries%2Cperformance_n_plus_one_api_calls%2Cperformance_p95_endpoint_regression%2Cperformance_slow_db_query%2Cperformance_render_blocking_asset_span%2Cperformance_uncompressed_assets%2Cperformance_http_overhead%2Cperformance_large_http_payload%5D%20timesSeen%3A%3E10&referrer=issue-stream&sort=date): a 400 error (attempt to filter on a `null` id) qualified as a 500 error from rest api.

I don't know what was our long term plan on validation for rest api, I suppose it's not a priority, making it acceptable to interceipt some postgres errors based on their messages. Ideally we would have a validation earlier, at args parsing level.

before
Capture d’écran 2026-02-02 à 14 08 53

after
Capture d’écran 2026-02-02 à 14 09 01

### DIFF
--- a/packages/twenty-server/src/engine/twenty-orm/error-handling/compute-twenty-orm-exception.ts
+++ b/packages/twenty-server/src/engine/twenty-orm/error-handling/compute-twenty-orm-exception.ts
@@ -35,10 +35,10 @@ export const computeTwentyORMException = async (
       );
     }
 
+    const errorCode = (error as QueryFailedErrorWithCode).code;
+
     if (
-      error.message.includes(
-        'duplicate key value violates unique constraint',
-      ) &&
+      errorCode === POSTGRESQL_ERROR_CODES['23505'] &&
       isDefined(objectMetadata) &&
       isDefined(entityManager) &&
       isDefined(internalContext)
@@ -51,38 +51,16 @@ export const computeTwentyORMException = async (
       );
     }
 
-    if (error.message.includes('invalid input value for')) {
+    // invalid input syntax or value
+    if (errorCode === '22P02') {
       return new TwentyORMException(
-        error.message,
+        error.message, // safe and useful
         TwentyORMExceptionCode.INVALID_INPUT,
       );
     }
-
-    if (error.message.includes('invalid input syntax for type')) {
-      return new TwentyORMException(
-        error.message,
-        TwentyORMExceptionCode.INVALID_INPUT,
-      );
-    }
-
-    if (error.message.includes('malformed array literal')) {
-      return new TwentyORMException(
-        error.message,
-        TwentyORMExceptionCode.INVALID_INPUT,
-      );
-    }
-
-    if (error.message.match(/column .+ does not exist/)) {
-      return new TwentyORMException(
-        error.message,
-        TwentyORMExceptionCode.INVALID_INPUT,
-      );
-    }
-
-    const errorCode = (error as QueryFailedErrorWithCode).code;
 
     if (isDefined(errorCode) && POSTGRESQL_ERROR_CODES.includes(errorCode)) {
-      throw new PostgresException(error.message, errorCode);
+      throw new PostgresException('Data validation error.', errorCode);
     }
     throw error;
   }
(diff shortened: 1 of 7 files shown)


packages/twenty-server/src/engine/twenty-orm/error-handling/compute-twenty-orm-exception.ts
```diff
@@ -58,6 +58,27 @@ export const computeTwentyORMException = async (
...
+    }
+
+    if (error.message.includes('malformed array literal')) {
+      return new TwentyORMException(
+        error.message,
+        TwentyORMExceptionCode.INVALID_INPUT,
+      );
+    }
+
+    if (error.message.match(/column .+ does not exist/)) {
```
Weiko: Is it really an INVALID_INPUT? It seems to be this is usually more on our end than the user (or at least it should be caught separately)

packages/twenty-server/src/engine/twenty-orm/error-handling/compute-twenty-orm-exception.ts
```diff
@@ -58,6 +58,27 @@ export const computeTwentyORMException = async (
       );
     }
 
+    if (error.message.includes('invalid input syntax for type')) {
+      return new TwentyORMException(
+        error.message,
```
Weiko: Not sure if we want to "leak" postgres errors to the user, should we use userFriendlyMessage for that?

packages/twenty-server/src/engine/twenty-orm/error-handling/compute-twenty-orm-exception.ts
```diff
@@ -58,6 +58,27 @@ export const computeTwentyORMException = async (
...
+        TwentyORMExceptionCode.INVALID_INPUT,
+      );
+    }
+
+    if (error.message.includes('malformed array literal')) {
```
Weiko: I believe typeorm returns SQL error codes, if we can use those these seems more robust

packages/twenty-server/src/engine/twenty-orm/error-handling/compute-twenty-orm-exception.ts
ijreilly: removing it indeed

packages/twenty-server/src/engine/twenty-orm/error-handling/compute-twenty-orm-exception.ts
ijreilly: userFriendlyMessage is of no use for rest api since it is supposed to be used in the FE.
I get your point, I think we can safely transfer the error message for that type of error because it will be usefl and it wont leak info other than the one the user is supposed to have (the expected type)

But I will remove error.message which is sent by default when throwing new `PostgresException` - agreed we should not send it there, we dont know what it holds

packages/twenty-server/src/engine/twenty-orm/error-handling/compute-twenty-orm-exception.ts
```diff
@@ -51,17 +51,16 @@ export const computeTwentyORMException = async (
       );
     }
 
-    if (error.message.includes('invalid input value for')) {
+    // invalid input syntax or value
```
ijreilly: gathers `invalid input value for ...` and `invalid input syntax for type ...`

Weiko: Maybe we want to have a file with const errors like
`INVALID_TEXT_REPRESENTATION` (cf: https://www.postgresql.org/docs/8.2/errcodes-appendix.html) would be easier to read. What do you think @ijreilly ?

Weiko (approved):
LGTM @ijreilly left one last comment!

PR #17572, opened 2026-01-30
`ObjectMetadata` and `FieldMetadata` agnostic workspace migration runner

prastoin:
# Introduction
Important note: This PR officially deprecates the `standardId`, about to drop col and entity property after this has been merged

Important note2: Haven't updated the optimistic tool to also update the universal identifier aggregators only the ids one, they should not be consumed in the runner context -> need to improve typing or either the optimistic tooling

In this PR we're introducing all the devxp allowing future metadata incremental universal migration -> this has an impact on all existing metadata actions handler ( explaining its size )
This PR also introduce workspace agnostic create update actions runner for both field and object metadata in order to battle test the described above devxp

Noting that these two metadata are the most complex to handle

Notes:
- A workspace migration is now highly bind to a `applicationUniversalIdentifier`. Though we don't strictly validate application scope for the moment

## Next
Migrate both object and field builder to universal comparison

### Concept

The migration system uses a two-phase action model:

1. **Universal Actions** - Actions defined using `universalIdentifier` (stable, portable identifiers like `standardId` + `applicationId`)
2. **Flat Actions** - Actions defined using database `entityId` (UUIDs specific to a workspace)

### Why This Separation?

- **Universal actions are portable**: They can be serialized, stored, and replayed across different workspaces
- **Flat actions are executable**: They contain the actual database IDs needed to perform operations
- **Decoupling**: The builder produces universal actions; the runner transpiles them to flat actions at execution time

### Transpiler Pattern

Each action handler must implement `transpileUniversalActionToFlatAction()`:

```typescript
@Injectable()
export class CreateFieldActionHandlerService extends WorkspaceMigrationRunnerActionHandler(
  'create',
  'fieldMetadata',
) {
  override async transpileUniversalActionToFlatAction(
    context: WorkspaceMigrationActionRunnerArgs<UniversalCreateFieldAction>,
  ): Promise<FlatCreateFieldAction> {
    // Resolve universal identifiers to database IDs
    const flatObjectMetadata = findFlatEntityByUniversalIdentifierOrThrow({
      flatEntityMaps: allFlatEntityMaps.flatObjectMetadataMaps,
      universalIdentifier: action.objectMetadataUniversalIdentifier,
    });
    
    return {
      type: action.type,
      metadataName: action.metadataName,
      objectMetadataId: flatObjectMetadata.id, // Resolved ID
      flatFieldMetadatas: /* ... transpiled entities ... */,
    };
  }
}
```

### Action Handler Base Class

[...]

PR #17632, opened 2026-02-02
`[FIXES_MAIN]` Remove `objectMetadata` `standardId`

prastoin:
# Introduction
In this PR we're deprecating the object metadata standard id and replacing it to the universalIdentifier usage
As we've totally removed its insertion for both new field and object in https://github.com/twentyhq/twenty/pull/17572

## Note
- Removed upgrade commands before `1.17`


Weiko (approved):
LGTM

PR #17633, opened 2026-02-02
Fix creation of objects with acronym names (e.g. "O&J")

### LINKED ISSUE #17544: Name should be in camelCase - Cannot create new object
## Bug Description
https://github.com/twentyhq/private-issues/issues/414

ijreilly:
Fixes https://github.com/twentyhq/twenty/issues/17544

**Problem**
When users create custom objects with short acronym names like "O&J", the system generates an object name oJ. When creating relation fields, the morph field name was built using string concatenation:
const morphFieldName = `target${capitalize("oJ")}`; // → "targetOJ"
This produced "targetOJ", which failed validation because the camelCase check performed in `validateFlatFieldMetadataName` (camelCase(name) === name) returns "targetOj" for "targetOJ". The issue comes from consecutive camelCase() operations.

**Solution**
Actually, the `camelCase(name) === name` check is questionnable. 
What we want to check is that a name is in camelCase format, not that it corresponds to the camelCase version of a given string, while that's we are doing here. lodash does not provide camelCase validator, only camelCase convertor, so we used it as a way to validate the format of the name.
We may feel like `camelCase(name) === name` checks whether a name is camel-cased, but in addition to that it is also checking for a camel case "idempotency" we don't necessarily have and do not need: for instance if an object's name is "iOS" (which could be inferred from a label "I O S"), it won't pass the check: camelCase("iOS") is "ios" and "ios" !== "iOS".

The existing check with `STARTS_WITH_LOWER_CASE_AND_CONTAINS_ONLY_CAPS_AND_LOWER_LETTERS_AND_NUMBER_STRING_REGEX` acts as a camel case validator, so we don't need that camelCase() check.


packages/twenty-server/src/engine/metadata-modules/object-metadata/utils/build-default-relation-flat-field-metadatas-for-custom-object.util.ts
```diff
@@ -138,7 +139,9 @@ export const buildDefaultRelationFlatFieldMetadatasForCustomObject = ({
             targetFlatObjectMetadata.nameSingular as keyof typeof STANDARD_OBJECT_ICONS
           ] || 'IconBuildingSkyscraper';
 
-        const morphFieldName = `target${capitalize(sourceFlatObjectMetadata.nameSingular)}`;
+        const morphFieldName = camelCase(
+          `target_${capitalize(sourceFlatObjectMetadata.nameSingular)}`,
```
prastoin: ***Question:*** Is the underscore mandatory here ? it would be a default relation custom object API metadata specific pattern ( meaning this won't be applied for twenty-app sync endpoint )
Also unless I'm mistaken we don't allow special char is this exempted ?

packages/twenty-server/src/engine/metadata-modules/object-metadata/utils/__tests__/build-default-relation-flat-field-metadatas-for-custom-object.util.spec.ts
```diff
@@ -0,0 +1,138 @@
+import camelCase from 'lodash.camelcase';
```
prastoin: ***Suggestion:*** Would it be possible to create an integration test instead of such unit test here ?

packages/twenty-server/src/engine/metadata-modules/object-metadata/utils/build-default-relation-flat-field-metadatas-for-custom-object.util.ts
Weiko: Not clear to me why we have this "_", won't it be removed by camelCase util anyway? And then the next word is already capitalised it seems. Maybe I'm missing something?

prastoin (commented):
Hey there, so the issue wasn't the name directly but rather the morph generated name interesting
I'm a bit concerned about the fix here as it's API metadata exclusive

> Another solution could have been to remove the check camelCase(name) === name which I think is useless since we already have the STARTS_WITH_LOWER_CASE_AND_CONTAINS_ONLY_CAPS_AND_LOWER_LETTERS_AND_NUMBER_STRING_REGEX check. I think this is actually the best solution but did not gather agreement on this

Hey there if you consider that removing camel case checking is the most appropriate solution I'm nobody to tell you it's not and would agree
I don't have the whole metadata entities validation pattern in my mind's RAM :P

packages/twenty-server/src/engine/metadata-modules/object-metadata/utils/build-default-relation-flat-field-metadatas-for-custom-object.util.ts
Weiko: Ok my bad, read too quickly about the multiple call to camelCase!

Weiko:
Not sure if the camelCase is relevant indeed @ijreilly @prastoin. Historically it's there to provide strong validation for postgres names and graphql type names. However none of them specify camelCase as their standards.
In fact, nestjs graphql depends from graphql-js https://github.com/graphql/graphql-js/blob/16.x.x/src/type/assertName.ts method (which is less rigid). As for the column/table/enum in pg, it's even more permissive.

prastoin (approved):
Sounds good to me !

PR #17635, opened 2026-02-02
fix: prevent anonymous users from bypassing workspace creation restriction

### LINKED ISSUE #17631: Feature Request: Add ability to disable anonymous user signup
### Discussed in https://github.com/twentyhq/twenty/discussions/17453

<div type='discussions-op-text'>

<sup>Originally posted by **ColourfulField** December  3, 2025</sup>
## Scope & Context

I am looking to create a setup with multiple workspaces, each assigned to a person I know personally. 

I want to give access to the CRM manually and disable signup for anonymous users.

## Current behavior

Currently when setting IS_MULTIWORKSPACE_ENABLED, all anonymous users can freely create new accounts and new workspaces.

Also, when setting an env variable IS_WORKSPACE_CREATION_LIMITED_TO_WORKSPACE_ADMINS=true, non-admin users who are already registered can no longer create workspaces. However, it does not prevent new account signups and workspace creation for those accounts.

More info here: #13460

### Option 1
IS_WORKSPACE_CREATION_LIMITED_TO_WORKSPACE_ADMINS=true disables workspace creation for both registered and anonymous users. 

For me this is a preferrable option due to how the variable is named. Since it was only introduced 2 months ago and wasn't added to the documentation or environment variable list inside admin panel, I assume that the impact of changing this behavior is low.

### Option 2
reintroduce IS_SIGN_UP_DISABLED variable. Setting this varialbe to true disables signup for anonymous users when IS_MULTIWORKSPACE_ENABLED=true.

## Technical inputs

The following code could be relevant in sign-in-up.service.ts. What I'm looking for is a way to disable signup when `this.isFirstWorkspaceForUser(currentUser.id) === true`
```
 async checkWorkspaceCreationIsAllowedOrThrow(
    currentUser: UserEntity,
  ): Promise<void> {
    if (!this.isWorkspaceCreationLimitedToServerAdmins()) return;

    if (await this.isFirstWorkspaceForUser(currentUser.id)) return;

    if (!currentUser.canAccessFullAdminPanel) {
      throw new AuthException(
        'Workspace creation is restricted to admins',
        AuthExceptionCode.FORBIDDEN_EXCEPTION,
        {
          userFriendlyMessage: msg`Workspace creation is restricted to admins`,
        },
      );
    }
  }
```

## Extra info

It might make sense to add the existing env variable IS_WORKSPACE_CREATION_LIMITED_TO_WORKSPACE_ADMINS to the list of config variables visible in the admin panel.
</div>

FelixMalfait:
When `IS_WORKSPACE_CREATION_LIMITED_TO_SERVER_ADMINS` is true, anonymous users could still create workspaces because `checkWorkspaceCreationIsAllowedOrThrow` checked per-user workspace count (always 0 for new users) instead of system-wide workspace count.

The bootstrap bypass now only applies when no workspaces exist in the entire system. Also adds the same check in `signUpOnNewWorkspace` to guard the `signInUp` code path.

Fixes #17631

Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/auth/services/sign-in-up.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/auth/services/sign-in-up.service.ts
@@ -403,8 +403,8 @@ export class SignInUpService {
     );
   }
 
-  private async isFirstWorkspaceForUser(userId: string): Promise<boolean> {
-    const count = await this.userWorkspaceService.countUserWorkspaces(userId);
+  private async isFirstWorkspaceInSystem(): Promise<boolean> {
+    const count = await this.workspaceRepository.count();
 
     return count === 0;
   }
@@ -414,7 +414,8 @@ export class SignInUpService {
   ): Promise<void> {
     if (!this.isWorkspaceCreationLimitedToServerAdmins()) return;
 
-    if (await this.isFirstWorkspaceForUser(currentUser.id)) return;
+    // Only allow bypass during initial system bootstrap (no workspaces exist yet)
+    if (await this.isFirstWorkspaceInSystem()) return;
 
     if (!currentUser.canAccessFullAdminPanel) {
       throw new AuthException(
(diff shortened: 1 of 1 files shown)


Weiko (approved):
LGTM

PR #17645, opened 2026-02-02
[FRONT COMPONENTS] Expose twenty UI in twenty sdk

bosiraphael:
- An app will declare its `twenty-sdk` version in the manifest.
- `twenty-sdk` will be served by a CDN to be imported in front component host.
- When we render the host we will load twenty-ui through the correct version of the sdk

We will proceed this way because twenty-ui is not mature enough yet to be deployed as its own package. So instead, since the sdk is already deployed as its own package, we will serve the ui with it. It allows us to have versioning to handle breaking changes in twenty-ui.


Weiko (approved):
LGTM

packages/twenty-sdk/project.json
```diff
@@ -4,8 +4,7 @@
   "sourceRoot": "packages/twenty-sdk/src",
   "projectType": "library",
   "tags": [
-    "scope:sdk",
-    "scope:shared"
+    "scope:frontend"
```
Weiko: Same as the other PR, you shouldn't need this change if you change scope to shared in twenty-ui. To fix before merging @bosiraphael

PR #17647, opened 2026-02-02
fix: prevent metadata schema leakage into /graphql and disable auto-discovery

### LINKED ISSUE #17638: Prevent metadata schema leakage into /graphql; disable GraphQL auto-discovery & speed up NestJS boot
## Summary
The `/graphql` endpoint currently exposes metadata resolvers/types that should only appear under `/metadata`. This is caused by GraphQL auto‑discovery scanning a broad module tree, which also contributes to slower NestJS boot times.

We should **disable auto‑discovery** for core GraphQL and move to **explicit schema construction**, while keeping resolvers co‑located with services. For metadata, we should also create an explicit schema from a central resolver list.

## Background / Evidence
- Core GraphQL is configured with `autoSchemaFile: true` and `include: [CoreEngineModule]`, which causes Nest/Yoga to scan the entire module graph for resolvers. That module graph includes metadata modules, so metadata resolvers get pulled into `/graphql`. (packages/twenty-server/src/engine/api/graphql/graphql-config/graphql-config.service.ts; packages/twenty-server/src/engine/core-modules/core-engine.module.ts)
- The `/metadata` endpoint uses the same auto‑discovery pattern via `autoSchemaFile` and `include`, pointing at `MetadataGraphQLApiModule`, which itself imports `MetadataEngineModule` (thereby scanning metadata resolvers). (packages/twenty-server/src/engine/api/graphql/metadata.module-factory.ts; packages/twenty-server/src/engine/api/graphql/metadata-graphql-api.module.ts)
- We already build the core workspace schema explicitly using `WorkspaceSchemaFactory` + `makeExecutableSchema`, so core GraphQL can work without discovery if we stop enabling it in config. (packages/twenty-server/src/engine/api/graphql/workspace-schema.factory.ts)

## Goals
1. Stop `/graphql` from exposing metadata schema.
2. Remove GraphQL auto‑discovery for the core endpoint and rely on explicit schema building.
3. Keep resolvers close to services (no per‑resolver modules).
4. Improve startup time by avoiding deep module graph scans.

### ✅ Core `/graphql` (explicit only)
- Remove `autoSchemaFile: true` and `include: [CoreEngineModule]` from core GraphQL config.
- Ensure the schema comes only from the existing `conditionalSchema` path that uses `WorkspaceSchemaFactory`.
- This eliminates Nest’s resolver scanning for core GraphQL entirely, preventing metadata bleed‑through and speeding startup.

Files involved:
- packages/twenty-server/src/engine/api/graphql/graphql-config/graphql-config.service.ts (remove auto‑discovery config, rely on conditional schema).
- packages/twenty-server/src/engine/api/graphql/workspace-schema.factory.ts (already builds schema explicitly; no changes expected unless needed).

### ✅ Metadata `/metadata` (explicit resolver list, no auto‑discovery)
- Keep resolvers where they are, but create a single centralized resolver list and build the metadata schema explicitly.
- Replace `autoSchemaFile` + `include` with an explicit schema build (e.g., buildSchema or Nest GraphQLSchemaFactory, depending on existing stack).
- Import service modules as needed for DI, but do not scan the module graph for resolvers.

Files involved:
- packages/twenty-server/src/engine/api/graphql/metadata.module-factory.ts (remove autoSchemaFile & include, provide explicit schema).
- packages/twenty-server/src/engine/api/graphql/metadata-graphql-api.module.ts (keep module wiring; update to use explicit schema instead of auto‑discovery).

## Acceptance Criteria
- `/graphql` introspection shows no metadata types or resolvers.
- `/metadata` continues to expose all metadata types/resolvers as before.
- Boot time improves or at least no longer worsens due to GraphQL scanning.
- No per‑resolver modules introduced; resolvers remain close to services.

## Optional Tests / Checks
- Add an integration test that introspects `/graphql` and asserts metadata types (e.g., ObjectMetadata, FieldMetadata) are absent.
- Add a complementary test for `/metadata` asserting those types are present.

## Notes
This change doesn’t require reorganizing resolvers into per‑resolver modules. The main shift is from auto‑discovery to explicit schema composition, which also reduces Nest’s module scanning overhead and makes schema boundaries explicit.

FelixMalfait:
Closes #17638

## Summary
- Remove `autoSchemaFile` and `include` from core `/graphql` config so schema comes exclusively from `conditionalSchema` via `WorkspaceSchemaFactory`
- Create centralized metadata resolver list (`metadata-resolvers.ts`)
- Build metadata `/metadata` schema explicitly using `GraphQLSchemaFactory` instead of auto-discovery scanning
- Add unit tests for config isolation and resolver list completeness
- Add integration test verifying `/graphql` does not expose metadata types

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/twenty-server/src/engine/api/graphql/graphql-config/graphql-config.service.ts
+++ b/packages/twenty-server/src/engine/api/graphql/graphql-config/graphql-config.service.ts
@@ -19,6 +19,18 @@ import { isDefined } from 'twenty-shared/utils';
 import { NodeEnvironment } from 'src/engine/core-modules/twenty-config/interfaces/node-environment.interface';
 
 import { WorkspaceSchemaFactory } from 'src/engine/api/graphql/workspace-schema.factory';
+import {
+  ApiConfig,
+  Billing,
+  Captcha,
+  ClientAIModelConfig,
+  NativeModelCapabilities,
+  PublicFeatureFlag,
+  PublicFeatureFlagMetadata,
+  Sentry as SentryConfig,
+  Support,
+} from 'src/engine/core-modules/client-config/client-config.entity';
+import { CoreEngineModule } from 'src/engine/core-modules/core-engine.module';
 import { ExceptionHandlerService } from 'src/engine/core-modules/exception-handler/exception-handler.service';
 import { useSentryTracing } from 'src/engine/core-modules/exception-handler/hooks/use-sentry-tracing';
 import { useDisableIntrospectionAndSuggestionsForUnauthenticatedUsers } from 'src/engine/core-modules/graphql/hooks/use-disable-introspection-and-suggestions-for-unauthenticated-users.hook';
@@ -79,6 +91,21 @@ export class GraphQLConfigService
     }
 
     const config: YogaDriverConfig = {
+      autoSchemaFile: true,
+      include: [CoreEngineModule],
+      buildSchemaOptions: {
+        orphanedTypes: [
+          ApiConfig,
+          Billing,
+          Captcha,
+          ClientAIModelConfig,
+          NativeModelCapabilities,
+          PublicFeatureFlag,
+          PublicFeatureFlagMetadata,
+          SentryConfig,
+          Support,
+        ],
+      },
       conditionalSchema: async (context) => {
         const { workspace, user } = context.req;
 
--- a/packages/twenty-server/src/engine/api/graphql/graphql-config/graphql-config.service.ts
+++ b/packages/twenty-server/src/engine/api/graphql/graphql-config/graphql-config.service.ts
@@ -19,18 +19,6 @@ import { isDefined } from 'twenty-shared/utils';
 import { NodeEnvironment } from 'src/engine/core-modules/twenty-config/interfaces/node-environment.interface';
 
 import { WorkspaceSchemaFactory } from 'src/engine/api/graphql/workspace-schema.factory';
-import {
-  ApiConfig,
-  Billing,
-  Captcha,
-  ClientAIModelConfig,
-  NativeModelCapabilities,
-  PublicFeatureFlag,
-  PublicFeatureFlagMetadata,
-  Sentry as SentryConfig,
-  Support,
-} from 'src/engine/core-modules/client-config/client-config.entity';
-import { CoreEngineModule } from 'src/engine/core-modules/core-engine.module';
 import { ExceptionHandlerService } from 'src/engine/core-modules/exception-handler/exception-handler.service';
 import { useSentryTracing } from 'src/engine/core-modules/exception-handler/hooks/use-sentry-tracing';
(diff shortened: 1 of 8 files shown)


FelixMalfait:
Might not be a good idea. Just testing

packages/twenty-server/src/engine/api/graphql/graphql-config/graphql-config.service.ts
```diff
@@ -91,21 +79,6 @@ export class GraphQLConfigService
     }
 
     const config: YogaDriverConfig = {
-      autoSchemaFile: true,
-      include: [CoreEngineModule],
```
FelixMalfait, line 95: What about the additional methods that should belong here like file upload? @claude which other mutation/queries do you think would fit in /graphql endpoint? Does it make sense to have them in both? (e.g. GetCurrentUser, etc.). Let's look at all the graphql methods (codegen file maybe) and post an analysis of what would make the most sense according to you

FelixMalfait:
@claude fix CI 
 nx run twenty-server:lint:ci
  
  Linting "twenty-server"...
  (node:4788) [DEP0180] DeprecationWarning: fs.Stats constructor is deprecated.
  (Use `node --trace-deprecation ...` to show where the warning was created)
  
  /home/runner/work/twenty/twenty/packages/twenty-server/src/engine/api/graphql/__tests__/metadata.module-factory.spec.ts
    26:11  warning  'config' is assigned a value but never used. Allowed unused vars must match /^_/u  unused-imports/no-unused-vars
  
  Also look at the comment above (graphql diff), we lost many types/queries/mutation in the /metadata which were used (billing, admin panel, api key, applications, etc.). Make sure we don't break anything in the app

packages/twenty-server/src/engine/api/graphql/graphql-config/graphql-config.service.ts
FelixMalfait: @claude I do not want to go with more than 2 paths. Please suggest many pairs if we go with 2 paths. 
Also between the 3 options you proposed what is your preferred option? Why? 
Which option is closest to Salesforce / what would Salesforce do?

FelixMalfait: /graphql/data ; /graphql/core ? fits core schema, core-modules folder etc

charlesBochet:
@FelixMalfait looks slower to iterate than IDE!

FelixMalfait:
@charlesBochet I actually like it! I can do emails or other things and come back. Not perfect for everything I agree! I really like it for reviews though, I think it's super valuable to see the chain of thoughts.

In this case I think it would be too hard to finish this with Claude only because it's too subtle (many breaking changes and things too check locally). But the exploration was useful

charlesBochet:
Closing this one @FelixMalfait, let's re-open if you have bandwidth, would be great to fix this indeed. I also think we can fix it easily by making a patch

PR #17652, opened 2026-02-02
Remote dom twenty UI POC

bosiraphael:
Create a proof of concept which allows us to use twenty-ui in the remote-dom.
For now only the button is allowed.

https://github.com/user-attachments/assets/e5441d2c-eb63-4b99-931b-86ee14621393

Known limitation: The icons are not yet available


packages/twenty-sdk/project.json
```diff
@@ -4,8 +4,7 @@
   "sourceRoot": "packages/twenty-sdk/src",
   "projectType": "library",
   "tags": [
-    "scope:sdk",
-    "scope:shared"
+    "scope:frontend"
```
Weiko: Prefer to change twenty-ui scope to shared instead imho @bosiraphael (then you don't have to do changes in this file)

Weiko (commented):
Left a comment otherwise LGTM @bosiraphael

packages/twenty-sdk/project.json
bosiraphael: Okay

Weiko (approved):
LGTM

PR #17659, opened 2026-02-03
Allow twenty standard app to access messaging items

ijreilly:
This is a quick fix to unblock a user who wants to create workflows around messaging. 

We do need to think of what should be the usage for other applications - I suppose other applications will want to access messaging. Maybe we should add an **Access to messages and calendar events** setting permission key that could be granted (or not) to applications?


charlesBochet (approved):

Weiko (approved):
LGTM

PR #17661, opened 2026-02-03
Files v2 - File table migration + data migration command

etiennejouan:
To add also in 1.16 patch


packages/twenty-server/src/database/typeorm/core/migrations/utils/1768572831179-updateFileTable.util.ts
```diff
@@ -0,0 +1,25 @@
+import { type QueryRunner } from 'typeorm';
+
+export const updateFileTableQueries = async (
+  queryRunner: QueryRunner,
+): Promise<void> => {
+  await queryRunner.query(`ALTER TABLE "core"."file" DROP COLUMN "name"`);
```
prastoin, line 6: No more name ?

prastoin (approved):
C'toi que je vais delete mon gars
image

PR #17674, opened 2026-02-03
[fix] Some fixes

ijreilly:
- fixes [sentry](https://twenty-v7.sentry.io/issues/7239112455/?environment=prod&environment=prod-eu&project=4507072563183616&query=anything%20else&referrer=issue-stream&sort=date)
- adapt useClearField logic to morph relations
- adapt useClearField logic to one-to-many relations (early return)
- fix creation of objects without "name" field from specific parts of the product (which involved a "name" field)

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-field/ui/hooks/useClearField.ts
+++ b/packages/twenty-front/src/modules/object-record/record-field/ui/hooks/useClearField.ts
@@ -43,12 +43,12 @@ export const useClearField = () => {
           foundFieldMetadataItem.type === FieldMetadataType.RELATION ||
           foundFieldMetadataItem.type === FieldMetadataType.MORPH_RELATION;
 
-        const isOneToManyRelation =
+        const shouldSkipClearingBecauseInvolvesMultipleRecords =
           isRelation &&
           foundFieldMetadataItem.settings?.relationType ===
             RelationType.ONE_TO_MANY;
 
-        if (isOneToManyRelation) {
+        if (shouldSkipClearingBecauseInvolvesMultipleRecords) {
           return;
         }
 
@@ -71,6 +71,13 @@ export const useClearField = () => {
         const updateFieldName = isManyToOneRelation
           ? getForeignKeyNameFromRelationFieldName(fieldName)
           : fieldName;
+
+        if (isManyToOneRelation) {
+          set(
+            recordStoreFamilySelector({ recordId, fieldName: updateFieldName }),
+            emptyFieldValue,
+          );
+        }
 
         updateRecord?.({
           variables: {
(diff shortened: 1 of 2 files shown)


packages/twenty-front/src/modules/object-record/record-field/ui/hooks/useClearField.ts
```diff
@@ -37,6 +39,19 @@ export const useClearField = () => {
...
+          foundFieldMetadataItem.type === FieldMetadataType.RELATION ||
+          foundFieldMetadataItem.type === FieldMetadataType.MORPH_RELATION;
+
+        const isOneToManyRelation =
+          isRelation &&
+          foundFieldMetadataItem.settings?.relationType ===
+            RelationType.ONE_TO_MANY;
+
+        if (isOneToManyRelation) {
+          return;
```
ijreilly, line 52: 1) I don't think we want to detach all the relations upon clear field action (which is triggered by a hot key)
2) it would need to trigger multiple records at once

packages/twenty-front/src/modules/object-record/utils/buildRecordLabelPayload.ts
```diff
@@ -16,17 +17,27 @@ export const buildRecordLabelPayload = ({
   const labelIdentifierField =
     getLabelIdentifierFieldMetadataItem(objectMetadataItem);
 
-  if (labelIdentifierField?.type === FieldMetadataType.FULL_NAME) {
+  if (
```
ijreilly, line 20: this because we no longer necessarily have a name field on objects so it as breaking (example: "Add new" from a relation picker)

packages/twenty-front/src/modules/object-record/record-field/ui/hooks/useClearField.ts
```diff
@@ -48,11 +63,20 @@ export const useClearField = () => {
...
         );
 
+        const isManyToOneRelation =
+          isRelation &&
+          foundFieldMetadataItem.settings?.relationType ===
+            RelationType.MANY_TO_ONE;
+
+        const updateFieldName = isManyToOneRelation
+          ? getForeignKeyNameFromRelationFieldName(fieldName)
+          : fieldName;
```
ijreilly, line 73: this fixes the sentry - we were trying to update `people.company` instead of `people.companyId`

packages/twenty-front/src/modules/object-record/record-field/ui/hooks/useClearField.ts
Weiko, line 62: should we use updateFieldName instead of fieldName here? Is it expected that local state is not using the same key than the mutation?

Weiko: Ok I think this is expected actually, my bad!

packages/twenty-front/src/modules/object-record/record-field/ui/hooks/useClearField.ts
Weiko: I see, maybe we could make it clearer 🤔 
maybe store this in a `clearingWouldDetachMultipleRelatedRecords` or something (probably better, I don't like that one actually 🙂)

Weiko (approved):
LGTM @ijreilly

packages/twenty-front/src/modules/object-record/record-field/ui/hooks/useClearField.ts
ijreilly: You have a point though, both are present so we should nullify both. Ive updated the logic

PR #17675, opened 2026-02-03
[FRONT COMPONENT] Stories in twenty-sdk for front component generation and upon render interactivity

### LINKED ISSUE #2179: [FRONT COMPONENTS] Create stories in twenty-sdk to test the output of the generated file
These stories have to import the FrontComponentRenderer from twenty-shared.

- Create some mocked react components
- Run the generate script
- The stories will try to render the generated files

In the CI: Run the generate script -> Run the stories

ehconitin:
closes https://github.com/twentyhq/core-team-issues/issues/2179


packages/twenty-sdk/src/cli/utilities/build/common/front-component-build/utils/replace-html-tags-with-remote-components.ts
```diff
@@ -1,6 +1,6 @@
 import { isDefined } from 'twenty-shared/utils';
 
-import { HTML_TAG_TO_REMOTE_COMPONENT } from '@/front-component-constants';
+import { HTML_TAG_TO_REMOTE_COMPONENT } from '../../../../../../front-component-constants';
```
bosiraphael, line 3: Fix import

packages/twenty-sdk/src/front-component/__stories__/utils/buildMockComponents.ts
```diff
@@ -0,0 +1,58 @@
+import * as esbuild from 'esbuild';
```
bosiraphael, line 1: We should use the same util used to build the front components (maybe with parameters for the output and the input), these stories are a way to test that the build works.

ehconitin: I think your comment was about the build config, not the esbuild import 🤔. I refactored so both the CLI front‑component build and the Storybook mock build use the same shared build helper (same plugins/externals/tsconfig). Only the entry points + output dir differ, which is expected.
Using the full watcher pipeline directly doesn’t fit Storybook because it’s tied to app paths/output and watch behavior. This keeps the build identical without forcing Storybook into the app build layout.

bosiraphael (approved):
LGTM

PR #17683, opened 2026-02-03
Added relations to SSE events

### LINKED ISSUE #2192: Handle relation with SSE
Relation fields are not handled with SSE events, we want it to be handled.

lucasbordeau:
Fixes https://github.com/twentyhq/core-team-issues/issues/2192

This PR implements what is necessary to re-create the query that we build on the frontend to obtain the returned object record from a mutation, but on the backend, which was only partially implemented for REST API.

Usually we want to have relations with only their id and label identifier field to have lighter payloads.

In the event we only had depth 0 fields, with this PR we have all events with depth 1 relations.

We have depth 2 for many-to-many cases, like updateOne or updateMany result :
- Junction tables
- Activity target tables


packages/twenty-server/src/engine/api/rest/core/rest-to-common-args-handlers/utils/check-if-field-is-image-identifier.util.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const checkIfFieldIsImageIdentifier = (
+  flatField: FlatFieldMetadata,
+  flatObjectMetadata: FlatObjectMetadata,
+): boolean => {
+  if (
+    flatObjectMetadata.nameSingular === 'company' &&
+    flatField.name === 'domainName'
+  ) {
+    return true;
```
thomtrp: I am surprised we never need it before. Looks a bit hacky TBH. Don't we have a better way to retrieve the label identifier?

packages/twenty-server/src/engine/metadata-modules/role-target/types/user-workspace-role-map.ts
```diff
@@ -0,0 +1,14 @@
...
+ * ```
+ * {
+ *    'userWorkspaceId1': 'roleId1',
+ *    'userWorkspaceId2': 'roleId1',
+ *    'userWorkspaceId3': 'roleId2',
+ *    ...
+ * }
+ * ```
+ */
+export type UserWorkspaceRoleMap = Record<string, string>;
```
thomtrp, line 1: Nit: I think we can simplify that comm

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.service.ts
```diff
@@ -222,6 +242,112 @@ export class WorkspaceEventEmitterService {
...
+      return;
+    }
+
+    const userWorkspaceId = streamData.authContext.userWorkspaceId;
+
+    if (!isDefined(userWorkspaceId)) {
+      throw new InternalServerErrorException(
+        'User workspace ID is undefined in stream auth context',
+      );
+    }
```
thomtrp: we are already checking it above. Maybe we can send it as input

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.service.ts
```diff
@@ -222,6 +242,112 @@ export class WorkspaceEventEmitterService {
...
+
+    const roleId = permissionsContext.userWorkspaceRoleMap[userWorkspaceId];
+
+    if (!isDefined(roleId)) {
+      throw new InternalServerErrorException(
+        'Role ID is undefined for user workspace ID',
+      );
```
thomtrp: same

thomtrp (commented):
LGTM but I think we need @Weiko feedbacks here

packages/twenty-server/src/engine/api/rest/core/rest-to-common-args-handlers/utils/check-if-field-is-image-identifier.util.ts
lucasbordeau: Indeed we already have it

Weiko: We have isImageIdentifierField in the frontend indeed. This could live in twenty-shared maybe

packages/twenty-server/src/engine/api/rest/core/rest-to-common-args-handlers/utils/check-if-field-is-label-identifier.util.ts
```diff
@@ -0,0 +1,9 @@
+import { type FlatFieldMetadata } from 'src/engine/metadata-modules/flat-field-metadata/types/flat-field-metadata.type';
+import { type FlatObjectMetadata } from 'src/engine/metadata-modules/flat-object-metadata/types/flat-object-metadata.type';
+
+export const checkIfFieldIsLabelIdentifier = (
```
Weiko: Could you check in the server where we do
`flatObjectMetadata.labelIdentifierFieldMetadataId ===` and use that new util? 🙂

Weiko: could be in shared as well

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.service.ts
```diff
@@ -222,6 +242,112 @@ export class WorkspaceEventEmitterService {
...
+      .map((event) => {
+        if ('before' in event.properties) {
+          return event.properties.before as Nullable<ObjectRecord>;
+        }
+
+        return undefined;
+      })
+      .filter(isDefined);
+
+    const recordsAfter = workspaceEventBatch.events
```
Weiko: nit: can be done in a single loop

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.service.ts
```diff
@@ -222,6 +242,112 @@ export class WorkspaceEventEmitterService {
...
+
+    const rolePermissionConfig: RolePermissionConfig = {
+      intersectionOf: [roleId],
+    };
+
+    const globalWorkspaceDataSource =
+      await this.globalWorkspaceOrmManager.getGlobalWorkspaceDataSourceReplica();
+
+    const selectedFields =
+      this.restToCommonSelectedFieldsHandler.computeFromDepth({
```
Weiko: I know why you are doing this but it feels a bit wrong to call rest services and graphql parser in the same path. I'll try to improve that and make those a bit more API agnostic!

packages/twenty-server/src/engine/metadata-modules/role-target/types/user-workspace-role-map.ts
Weiko: The comment is not that useful imho

packages/twenty-server/src/engine/api/rest/core/rest-to-common-args-handlers/selected-fields-handler.ts
```diff
@@ -107,12 +150,44 @@ export class RestToCommonSelectedFieldsHandler {
...
             objectsPermissions,
             depth: 1,
+            onlyUseLabelIdentifierFieldsInRelations,
           });
 
         relationsSelectFields[flatField.name] = {
           ...relationFieldSelectFields,
           ...depth2RelationsSelectFields,
         };
+      } else if (flatFieldIsActivityTarget) {
```
Weiko: Note: What about future many-to-many relations? I believe we are trying to achieve that very soon and the code is almost ready.
taskTarget/noteTarget are probably not migrated yet though but ideally this should be metadata driven instead of hardcoded

packages/twenty-server/src/engine/api/rest/core/rest-to-common-args-handlers/selected-fields-handler.ts
```diff
@@ -107,12 +150,44 @@ export class RestToCommonSelectedFieldsHandler {
...
+        const activityTargetRelationsSelectFields =
```
Weiko: all those else-if are hard to read and trying to achieve the same thing it seems
I suggest doing this
```typescript
const shouldRecurseIntoRelation =
  (depth === MAX_DEPTH && isDefined(flatField.relationTargetObjectMetadataId)) ||
  flatFieldIsActivityTarget ||
  flatFieldIsJoinColumn;

if (shouldRecurseIntoRelation) {
  const nestedRelationsSelectFields =
    this.getRelationsAndRelationsSelectFields({
      flatObjectMetadataMaps,
      flatFieldMetadataMaps,
      flatObjectMetadata: relationTargetObjectMetadata,
      objectsPermissions,
      depth: 1,
      onlyUseLabelIdentifierFieldsInRelations,
      isJunctionTable: flatFieldIsJoinColumn,
    });

  relationsSelectFields[flatField.name] = {
    ...relationFieldSelectFields,
    ...nestedRelationsSelectFields,
  };
} else {
  relationsSelectFields[flatField.name] = relationFieldSelectFields;
}
```
What do you think?

Weiko (commented):
Left some comments @lucasbordeau !

packages/twenty-server/src/engine/api/rest/core/rest-to-common-args-handlers/selected-fields-handler.ts
lucasbordeau: Ok I will abstract this into a util for now, as it will probably be easier to refactor once we are there.

packages/twenty-front/src/modules/object-record/utils/computeOptimisticRecordFromInput.ts
```diff
@@ -140,12 +140,6 @@ export const computeOptimisticRecordFromInput = ({
         continue;
       }
 
-      if (!isUndefined(recordInputFieldValue)) {
```
lucasbordeau, line 143: @charlesBochet I removed this because it makes no sense to check that for SSE events. I saw no problem removing it because the optimistic logic shouldn't crash if there is "too much" data. 

Also this was developed while taking for granted that we have all corresponding relations in the front cache, which is not the case for SSE events.

packages/twenty-front/src/modules/sse-db-event/hooks/useTriggerEventStreamCreation.ts
```diff
@@ -69,46 +69,60 @@ export const useTriggerEventStreamCreation = () => {
...
             message: ({ data, event }) => {
-              if (event === 'next') {
-                if (isDefined(data?.errors)) {
-                  const subCode = data.errors[0]?.extensions?.subCode;
-
-                  switch (subCode) {
-                    case 'EVENT_STREAM_ALREADY_EXISTS': {
-                      set(shouldDestroyEventStreamState, true);
-                      break;
+              const result = data as ExecutionResult<{
```
lucasbordeau, line 82: @thomtrp I put everything into the `message` handler since this is the top-level handler that get called first.

- Errors can be in the `data` normal flow, so we have to handle them here.
- If an error happens during processing the data, that is not caught here, the SSE client silently crashed and don't recover, it took me some time to figure this out, so I put a top-level try/catch here.

packages/twenty-server/src/engine/api/common/common-select-fields/common-select-fields-helper.ts
```diff
@@ -18,19 +20,23 @@ type SelectFields = {
 };
 
 @Injectable()
-export class RestToCommonSelectedFieldsHandler {
+export class CommonSelectFieldsHelper {
```
lucasbordeau, line 23: @Weiko For now I put it in api/common, but we could move it elsewhere (maybe in another PR though)

packages/twenty-shared/src/metadata/check-if-field-is-image-identifier.util.ts
```diff
@@ -0,0 +1,23 @@
+import { type Nullable } from '@/types';
+
+export const checkIfFieldIsImageIdentifier = (
```
lucasbordeau, line 3: I would replace this in twenty-front/ but in another PR to not bloat this one.

charlesBochet:
@thomtrp could you take over this one? :)

thomtrp:
Closing until @lucasbordeau's back from pto

thomtrp (approved):

PR #17685, opened 2026-02-03
Trigger refetch workflow on version creation

thomtrp:
https://github.com/user-attachments/assets/ec156aba-3a67-4eef-addf-86158c3f1837


packages/twenty-front/src/modules/workflow/states/workflowRefetchRequestedFamilyState.ts
```diff
@@ -0,0 +1,9 @@
+import { createFamilyState } from '@/ui/utilities/state/utils/createFamilyState';
+
+export const workflowRefetchRequestedFamilyState = createFamilyState<
+  boolean,
+  string
+>({
+  key: 'workflowRefetchRequestedFamilyState',
```
lucasbordeau: shouldWorkflowRefetchRequest would be easier to understand and aligned with other states

lucasbordeau (approved):

PR #17686, opened 2026-02-03
fix(workspace-cache): memory leak in deleteFromLocalCache

FelixMalfait:
## Problem

The `deleteFromLocalCache` method was only setting `lastHashCheckedAt=0` instead of actually deleting the cache entry. This caused old versions to accumulate in the local cache when `invalidateAndRecompute` was called.

### Flow that causes the leak:

1. `invalidateAndRecompute()` is called
2. `flush()` → `deleteFromLocalCache()` — **only sets `lastHashCheckedAt=0`, keeps old data**
3. `recomputeDataFromProvider()` → `setInLocalCache()` — **adds new version with new hash**
4. `cleanupStaleVersions()` — **never called** (only triggered from `getFromLocalCache` path)

Result: Each `invalidateAndRecompute` call adds a new version to `entry.versions` without removing the old one.

### Impact

For migration commands (like `1-17-migrate-attachment-to-morph-relations`) that process thousands of workspaces, this caused significant memory growth:
- Each workspace calls `getOrRecompute` (version 1)
- Then calls `invalidateAndRecompute` (version 2 added, version 1 stays)
- Memory accumulates as the command processes more workspaces

## Solution

Actually delete the local cache entry in `deleteFromLocalCache`, so `recomputeDataFromProvider` starts fresh.

```typescript
// Before (memory leak)
if (isDefined(entry)) {
  entry.lastHashCheckedAt = 0;
}

// After (proper cleanup)
this.localCache.delete(localKey);
```


packages/twenty-server/src/engine/workspace-cache/services/workspace-cache.service.ts
```diff
@@ -353,11 +353,8 @@ export class WorkspaceCacheService implements OnModuleInit {
   ): void {
     for (const keyName of cacheKeyNames) {
       const localKey = this.buildCacheKey(workspaceId, keyName);
-      const entry = this.localCache.get(localKey);
 
-      if (isDefined(entry)) {
-        entry.lastHashCheckedAt = 0;
-      }
+      this.localCache.delete(localKey);
```
Weiko: By doing this, we are losing all the versions stored in the cache, this was intended to avoid race conditions were references were still being used by concurrent requests.
To avoid accumulation I remember setting a MAX_LOCAL_STALE_VERSIONS (+ a TTL that should be quite short I believe), maybe we want to update this? Or maybe there is a bug and the stack is never cleaned and exceeds the max for some reason 🤔

charlesBochet: +1

charlesBochet: I would start by logging the size of the cache before trying to modify its logic

FelixMalfait:
@claude can you start from main and create a command to validate the scenario and our fix? For example in that command you could run a loop on workspaces to create pressure on the memory and log the  size of local cache to see if it confirms our hypothesis. Post script and detailed result in a comment here when you're finished. This fix is worthless if we can't prove it really was the problem in the first place

FelixMalfait:
@claude but is that really significant? The command failed after approximately 100 workspaces, on a node that had 3gb of RAM allocated. Do you think the OOM could really come from this? Try to run the number to see if it's really the same order of magnitude or if it comes from something else

FelixMalfait:
@claude I've read online that TypeORM was a know source for memory issues. For example eager loading relations (with join and not the mode with separate select queries) could consume a huge amount of RAM. Please Look at the 1.17 attachement command and keep writing scripts to try to replicate the memory leak. Don't base your analysis on guesses but rather comment with real observations/measurements that would confirm or negate our hypothesis.

FelixMalfait:
@claude can you please perform a detailed analysis focused on TypeORM EntityMetadata now? Same I want you to write concrete examples with a methodical analysis based on observation / test scripts you have created and measured, not guesses. If the issue is confirm, provide recommandations.

FelixMalfait:
@claude please take a step back and investigate other paths. I have a feeling it's linked to TypeORM. Try to reproduce through small scale experiment or isolation. The memory leak is particularly obvious in the attachment command since it crashes but overall it affects all of the app, we're also seeing restart pattern on http node ms and on worker nodes, but it can last hours / it's much slower to run out of memory than the command which runs out after ~100 workspace processed

Weiko (approved):
Ok with the new strategy 👍

PR #17702, opened 2026-02-04
Add IS filter for multi-select fields `[Low priority - close issue & PR if not valid]`

Bonapara:
## Summary
Add a new "IS" filter operand for MULTI_SELECT fields that filters records containing ALL selected values (AND logic), in contrast to the existing "Contains" filter which matches ANY selected value (OR logic).

Closes #17680

## Changes
- Add `containsEvery` field to MultiSelectFilterType GraphQL schema
- Add `containsEvery` to MultiSelectFilter TypeScript type
- Add IS operand to MULTI_SELECT filter operands list
- Implement IS operand handling using containsEvery in filter conversion
- Add containsEvery SQL operator using PostgreSQL @> (contains) operator
- Add containsEvery case to local matching function

Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/packages/twenty-server/src/engine/api/graphql/workspace-schema-builder/graphql-types/input/multi-select-filter.input-type.ts
+++ b/packages/twenty-server/src/engine/api/graphql/workspace-schema-builder/graphql-types/input/multi-select-filter.input-type.ts
@@ -11,6 +11,7 @@ export const MultiSelectFilterType = new GraphQLInputObjectType({
   name: 'MultiSelectFilter',
   fields: {
     containsAny: { type: new GraphQLList(GraphQLString) },
+    containsEvery: { type: new GraphQLList(GraphQLString) },
     is: { type: FilterIs },
     isEmptyArray: { type: GraphQLBoolean },
   },
--- a/packages/twenty-shared/src/types/RecordGqlOperationFilter.ts
+++ b/packages/twenty-shared/src/types/RecordGqlOperationFilter.ts
@@ -127,6 +127,7 @@ export type MultiSelectFilter = {
   is?: IsFilter;
   isEmptyArray?: boolean;
   containsAny?: string[];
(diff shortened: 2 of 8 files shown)


packages/twenty-shared/src/types/RecordGqlOperationFilter.ts
```diff
@@ -127,6 +127,7 @@ export type MultiSelectFilter = {
   is?: IsFilter;
   isEmptyArray?: boolean;
   containsAny?: string[];
+  containsEvery?: string[];
```
FelixMalfait, line 130: Contains Any / Contains Every is more clear than "is vs contain" ; I would expect IS to be a strict equality if we do want it (not sure we do I think it adds confusion? but could be useful so no strong opinion). I would rather align to that instead. Great if frontend filters are closer to the backend api.

packages/twenty-server/src/engine/api/graphql/workspace-schema-builder/graphql-types/input/multi-select-filter.input-type.ts
```diff
@@ -11,6 +11,7 @@ export const MultiSelectFilterType = new GraphQLInputObjectType({
   name: 'MultiSelectFilter',
```
etiennejouan, line 11: Need to also update createGqlEnumFilterType file

adding this : 
      containsEvery: { type: new GraphQLList(enumType) }

etiennejouan (changes requested):

etiennejouan:
@claude take the comments into accounrt

etiennejouan:
Seen with @Bonapara, advanced filter seems to be a good workaround 


Claude bot would become very interesting if it could test itself (or If testing could easily be done) cc @FelixMalfait

PR #17713, opened 2026-02-04
Display "Not shared" indicator for RLS-restricted relation fields

Weiko:
## Context
Previously, if for example a Person had a companyId but the company relation was null due to RLS, the company column appeared empty. Now it displays ForbiddenFieldDisplay to indicate the relation exists but is inaccessible.

When RLS restricts access to a related record, the frontend now shows a "Not shared" indicator (with lock icon) instead of an empty field.

Screenshot 2026-02-04 at 15 37 53

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/hooks/useRelationToOneFieldDisplay.ts
+++ b/packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/hooks/useRelationToOneFieldDisplay.ts
@@ -11,6 +11,7 @@ import { generateDefaultRecordChipData } from '@/object-metadata/utils/generateD
 import { FieldContext } from '@/object-record/record-field/ui/contexts/FieldContext';
 import { assertFieldMetadata } from '@/object-record/record-field/ui/types/guards/assertFieldMetadata';
 import { isFieldRelation } from '@/object-record/record-field/ui/types/guards/isFieldRelation';
+import { getJoinColumnName } from '@/object-record/record-field/ui/utils/junction/getJoinColumnName';
 import { useRecordFieldValue } from '@/object-record/record-store/hooks/useRecordFieldValue';
 import { isDefined } from 'twenty-shared/utils';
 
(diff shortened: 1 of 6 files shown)


packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/hooks/useRelationToOneFieldDisplay.ts
```diff
@@ -41,6 +42,15 @@ export const useRelationToOneFieldDisplay = () => {
     fieldDefinition,
   );
 
+  const joinColumnName =
```
ijreilly: do we have a util for that ? are we 100% sure about the logic here (fallback to fieldName + id) ?

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/helpers/process-nested-relations-v2.helper.ts
```diff
@@ -417,9 +417,6 @@ export class ProcessNestedRelationsV2Helper {
           (rel) => rel[joinField] === item.id,
         );
       } else {
-        if (relationResults.length === 0) {
```
ijreilly, line 420: why do we remove that ? was it unnecessary ?

Weiko: @ijreilly I don't remember why I did that code tbh. The idea was to obfuscate the foreign key if the relation was not present but there is no strong reason to do that imho (there was no permission at the time so it was not even on the table). To me, having access to the id only is not a real security issue @ijreilly (this could be one in combination with other things but the id itself does not reveal much, and we are talking about workspace specific access, not cross workspace, we usually favor the product in that case)

packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/hooks/useRelationToOneFieldDisplay.ts
Weiko: Tbh this settings should always be there on a relation MANY_TO_ONE. Maybe I can set foreignKeyFieldValue to undefined if that's the case (we probably don't want to throw and break the whole table view for example) and fallback to the previous behavior (not display the field if not accessible)
What do you think? @ijreilly

ijreilly: I see
I understand you would say we dont want to break the view for so little / for something that doesnt impeed usage of the view but I also feel it will make it harder to catch. What do we think about silently sending an error to sentry (for us to be warned) and then fallback to the previous behaviour (or else)? Maybe thats a bad pattern. 
Other than that my point was more about all these '`${...}id` we have through the code, not sure how reliable they are. But Im not strongly against adding this, up to you !

ijreilly (approved):

PR #17714, opened 2026-02-04
Google OAuth check real permissions before creating channels

neo773:


charlesBochet (approved):
LGTM, ty, good addition

PR #17716, opened 2026-02-04
Front components communication between host and remote

### LINKED ISSUE #2176: [FRONT COMPONENTS] Communication between host and remote
- Create a rendering context that can be passed to the component with some information about the app: theme, locale, tokens ...
- Allow the remote to do some actions on the host: navigate for example

bosiraphael:


packages/twenty-sdk/vite.config.ts
```diff
@@ -20,18 +20,15 @@ const entryFileNames = (chunk: any, extension: 'cjs' | 'mjs') => {
...
-  const moduleDirectory = splitFaceModuleId[splitFaceModuleId?.length - 2];
-  if (moduleDirectory === 'src') {
+  // Find which entry this chunk corresponds to
+  const entry = entries.find((e) => chunk.facadeModuleId?.endsWith(e));
+  if (!entry || entry === 'src/index.ts' || entry === 'src/cli/cli.ts') {
     return `${chunk.name}.${extension}`;
   }
-  return `${moduleDirectory}/index.${extension}`;
+
+  // Remove 'src/' prefix and '/index.ts' suffix to get the module path
```
charlesBochet, line 31: this whole dynamic barreling is way too complex for twenty-sdk that is not aimed to scaled in multiple barrels.

We should get rid of generateBarrels script, hard-code barrels in package.json and hard code them in vite.config.ts instead

charlesBochet:
Closing, let's re-open when ready

packages/twenty-sdk/vite.config.ts
bosiraphael: Indeed

charlesBochet (approved):

PR #17721, opened 2026-02-04
fix: use TwentyConfigService instead of ConfigService for ENTERPRISE_KEY

FelixMalfait:
## Summary

Replace NestJS `ConfigService` with `TwentyConfigService` for `ENTERPRISE_KEY` access in row-level permission services.

## Changes

- `row-level-permission-predicate.service.ts`: Updated to use `TwentyConfigService`
- `row-level-permission-predicate-group.service.ts`: Updated to use `TwentyConfigService`

## Why

`TwentyConfigService` also pulls config values from the database, not just environment variables. This ensures consistent config access across the codebase.


Weiko (approved):
LGTM!

PR #17736, opened 2026-02-05
fix: set isActive=true in morph migration & add system fields toggle

FelixMalfait:
## Summary

Three fixes in this PR:

### 1. Migration commands now set `isActive = true` and use generic 'Target' label

When converting relation fields to `MORPH_RELATION` type, the migration commands now:
- Set `isActive = true` to prevent inactive fields from being selected as the representative morph field
- Update all field labels to generic **'Target'** instead of keeping individual labels like 'Company', 'Person'

This ensures the UI shows a coherent label ('Target') alongside 'X Objects' for the type.

**Files changed:**
- `1-17-migrate-note-target-to-morph-relations.command.ts`
- `1-17-migrate-task-target-to-morph-relations.command.ts`

### 2. Added system fields/relations toggles in Settings

The fields and relations tables in Settings > Data Model were filtering out system fields with no way to view them. Added a "System fields" / "System relations" toggle (visible in advanced mode) to allow viewing these fields.

**Files changed:**
- `SettingsObjectFieldTable.tsx`
- `SettingsObjectRelationsTable.tsx`

This matches the existing behavior on the Objects table which already has a "System objects" toggle.


charlesBochet (approved):

charlesBochet:
stamping to unblock

packages/twenty-server/src/engine/metadata-modules/flat-object-metadata/utils/get-morph-label-from-morph-field-metadata.util.ts
```diff
@@ -0,0 +1,38 @@
...
+    namePlural,
+  });
+
+  const isManyToOneRelationType =
+    morphRelationFlatFieldMetadata.settings.relationType ===
+    RelationType.MANY_TO_ONE;
+
+  // Return capitalized morph name as label (e.g., "target" -> "Target", "targets" -> "Targets")
+  // For many-to-one, use singular form; for one-to-many, the name is already plural
+  if (isManyToOneRelationType) {
```
Weiko: I don't understand, this seems to be doing the same without if?

FelixMalfait: I'll go for a simpler method scoped to the migration command for now and ping @prastoin about the underlying issue

PR #17744, opened 2026-02-05
Fix twenty sdk build 5

martmull:
use prepublish instead of prepack


ijreilly (approved):
loving this absolute final version !

PR #17746, opened 2026-02-05
fix: ensure side panel content doesn't shrink

Devessier:
Ensure side panel's content doesn't change during the animation. I set a fixed width to the content and rely on the parent to hide overflowing content.

## Before

https://github.com/user-attachments/assets/c5086155-f59e-4c6a-8430-2945996db891

## After

https://github.com/user-attachments/assets/f5c0ac35-be17-4c80-a9ce-7a3899785c84


Devessier:
This introduces a small regression when resizing the side panel. I'm not fully satisfied. We should find a better solution.

PR #17748, opened 2026-02-05
Add sync front component

Weiko:
## Context
Allow twenty apps to sync front components


Weiko:
@greptileai trigger

packages/twenty-server/src/engine/core-modules/application/services/application-sync.service.ts
```diff
@@ -948,6 +960,109 @@ export class ApplicationSyncService {
...
+          `Failed to find frontComponent to sync with universalIdentifier ${frontComponentToUpdate.universalIdentifier}`,
+          ApplicationExceptionCode.FRONT_COMPONENT_NOT_FOUND,
+        );
+      }
+
+      await this.frontComponentService.updateOne({
+        id: frontComponentToUpdate.id,
+        update: {
+          ...frontComponentToSync,
+          checksum: frontComponentToSync.builtComponentChecksum ?? undefined,
```
Weiko: Maybe we want to name it builtComponentChecksum in the DB as well btw

packages/twenty-server/src/engine/metadata-modules/flat-front-component/utils/from-create-front-component-input-to-flat-front-component-to-create.util.ts
```diff
@@ -4,6 +4,10 @@ import { v4 } from 'uuid';
 import { type FlatFrontComponent } from 'src/engine/metadata-modules/flat-front-component/types/flat-front-component.type';
 import { type CreateFrontComponentInput } from 'src/engine/metadata-modules/front-component/dtos/create-front-component.input';
 
+const DEFAULT_SOURCE_COMPONENT_PATH = 'src/front-components/index.tsx';
```
Weiko: Not sure about this but I believe those should be required in the DB at least

packages/twenty-server/src/engine/workspace-manager/workspace-migration/workspace-migration-runner/action-handlers/front-component/services/create-front-component-action-handler.service.ts
```diff
@@ -27,18 +35,55 @@ export class CreateFrontComponentActionHandlerService extends WorkspaceMigration
...
       );
 
     await frontComponentRepository.insert({
-      ...flatEntity,
+      ...frontComponent,
+      workspaceId,
+    });
+  }
+
+  private async verifySourceAndBuiltFilesExist({
```
Weiko, line 62: Reimplemented this from logic-function handler, however I'm a bit confused because it seems source files are not uploaded (and this function always fail due to that), in the meantime I only check built path until it's clarified

packages/twenty-server/src/engine/core-modules/application/services/application-sync.service.ts
martmull: we can also call it checksum in the manifest

packages/twenty-server/src/engine/metadata-modules/flat-front-component/utils/from-create-front-component-input-to-flat-front-component-to-create.util.ts
martmull: we will probably need a seed base front-component mecanism like we have for logic functions to handle that. Those default file paths must point to real files stored in the file storage

packages/twenty-server/src/engine/core-modules/application/services/application-sync.service.ts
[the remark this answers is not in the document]
Weiko: Should never be null, will need to change that in the SDK in an upcoming PR

packages/twenty-server/src/engine/metadata-modules/front-component/dtos/create-front-component.input.ts
```diff
@@ -11,8 +11,37 @@ export class CreateFrontComponentInput {
...
+  builtComponentPath: string;
+
+  @IsString()
+  @Field()
+  componentName: string;
+
+  @IsString()
+  @IsOptional()
+  @HideField()
+  checksum?: string;
```
martmull: checksum shoudl be required (if we have builtComponentPath we can have checksum)

packages/twenty-server/src/database/typeorm/core/migrations/common/1770309316193-add-front-component-columns.ts
```diff
@@ -0,0 +1,43 @@
...
+      `ALTER TABLE "core"."frontComponent" ADD "sourceComponentPath" character varying NOT NULL`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "core"."frontComponent" ADD "builtComponentPath" character varying NOT NULL`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "core"."frontComponent" ADD "componentName" character varying NOT NULL`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "core"."frontComponent" ADD "checksum" text`,
```
martmull: not null like builtComponentPath

charlesBochet (approved):

PR #17752, opened 2026-02-05
File - Migrate avatarUrl > avatarFile on person (data migration + logic) + Attachment data migration

etiennejouan:
- Migration command
    - Check IS_FILES_FIELD_MIGRATED:false
    - Check or create avatarFile field
    - Fetch all people with avatarUrl
           - Move (Copy/move) file in storage
           - Create core.file record
           - Update person record
           
    - bonus : attachment migration  : fullPath > file (same logic)
   
- BE logic
    - Add avatarFile field on person

- FE logic 
   - Adapt logic to upload on/display avatarFile data

The whole imageIdentifier logic will be done later


packages/twenty-server/src/database/commands/upgrade-version-command/1-18/1-18-migrate-person-avatar-files.command.ts
```diff
@@ -0,0 +1,291 @@
...
+              filename: newResourcePath,
+            },
+          });
+
+          const fileEntity = fileRepository.create({
+            id: fileId,
+            path: newResourcePath,
+            workspaceId,
+            applicationId: twentyStandardApplication.id,
+            size: 0,
```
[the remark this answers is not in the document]
etiennejouan: can't be nullable and don't want to upload it to get the size

packages/twenty-server/src/database/commands/upgrade-version-command/1-18/1-18-migrate-person-avatar-files.command.ts
```diff
@@ -0,0 +1,291 @@
...
+          );
+        }
+
+        this.logger.log(`Migrated avatar for person ${person.id}`);
+      } catch (error) {
+        this.logger.error(
+          `Failed to migrate avatar for person ${person.id} in workspace ${workspaceId}: ${error.message}`,
+        );
+        throw error;
+      }
```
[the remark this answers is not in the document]
etiennejouan: throwing an error here should stop migration no ?

packages/twenty-server/src/database/commands/upgrade-version-command/1-18/1-18-migrate-person-avatar-files.command.ts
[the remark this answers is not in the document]
etiennejouan: -1 if you prefer

packages/twenty-front/src/modules/object-record/record-show/hooks/useRecordShowContainerActions.ts
```diff
@@ -32,35 +25,7 @@ export const useRecordShowContainerActions = ({
     return [updateEntity, { loading: false }];
   };
 
-  const onUploadPicture = async (file: File) => {
```
prastoin, line 35: ***Praise:*** Damn that's satisfying

packages/twenty-front/src/modules/object-record/record-show/hooks/usePersonAvatarUpload.ts
```diff
@@ -0,0 +1,90 @@
...
+    FeatureFlagKey.IS_FILES_FIELD_MIGRATED,
+  );
+
+  const { objectMetadataItem: personMetadata } = useObjectMetadataItem({
+    objectNameSingular: CoreObjectNameSingular.Person,
+  });
+
+  const avatarFileFieldMetadataId = personMetadata.fields.find(
+    (field) =>
+      field.type === FieldMetadataType.FILES && field.name === 'avatarFile',
```
prastoin, line 33: ***Remark:*** Might make sense to use universal identifier at some point 🤔

packages/twenty-front/src/modules/object-record/record-show/hooks/usePersonAvatarUpload.ts
```diff
@@ -0,0 +1,90 @@
...
+        variables: { file, fieldMetadataId: avatarFileFieldMetadataId },
+      });
+
+      const uploadedFile = result?.data?.uploadFilesFieldFile;
+
+      if (!isDefined(uploadedFile)) {
+        return;
+      }
+
+      await updateOneRecord({
```
prastoin, line 53: 🚀

packages/twenty-server/src/engine/core-modules/file/files-field/files-field.service.ts
```diff
@@ -63,7 +64,7 @@ export class FilesFieldService {
     const sanitizedFile = sanitizeFile({ file, ext, mimeType });
 
     const fileId = v4();
-    const name = `${fileId}${ext ? `.${ext}` : ''}`;
+    const name = `${fileId}${isNonEmptyString(ext) ? `.${ext}` : ''}`;
```
prastoin, line 64: ***Question:*** can we have a file without an extension ?

prastoin (commented):
Hey there first step review here
```ts

export const extractFileInfo = async ({
  file,
  declaredMimeType,
  filename,
}: {
  file: Buffer;
  declaredMimeType: string | undefined;
  filename: string;
}) => {
  const { ext: declaredExt } = buildFileInfo(filename);

  const detectedFileType = await FileType.fromBuffer(file);

  const mimeType = detectedFileType?.mime ?? declaredMimeType;

  const ext = detectedFileType?.ext ?? declaredExt;

  return {
    mimeType,
    ext,
  };
};
```

I'm not sure about this here, I wouldn't never trust the declared client data on this. And at the same time I don't know how it would behave if you upload a markdown mime type file but as `txt`
Unless I'm mistaken there's a possibility to insert a file with a mimetype and file ext fully provided by the client
Maybe never trust the mime type but do for the extension as several can match a single mime type ?
We might wanna dig that a little bit further on what's done usually

EDIT claude response
Great question — this is a nuanced topic. Here's my take:

## The core issue

`FileType.fromBuffer` (magic bytes detection) is the gold standard for binary files (images, PDFs, archives, executables, etc.), but it returns `undefined` for **all text-based formats** — `.txt`, `.md`, `.csv`, `.json`, `.html`, `.css`, `.yaml`, etc. — because these don't have distinctive magic byte signatures.

So the current fallback logic means:

- **Binary files**: Detected type wins — good, secure.
- **Text files**: Falls back entirely to client-declared data — that's the risky part.

## The dangerous scenario

Someone uploads a text file but declares `Content-Type: application/pdf` or even `application/javascript`. Since `fromBuffer` returns `undefined`, the code blindly trusts the declared MIME type. Depending on how the file is later served (e.g., `Content-Type` header on download), this could lead to MIME confusion or XSS if served inline.

## The markdown-as-txt case

Not really a security concern. Both are text formats, `fromBuffer` returns `undefined` for both, so you'd get the declared values. The worst that happens is a `.md` file being served as `text/plain` — harmless.

## Industry best practice

The common approach is a **three-tier strategy**:

1. **Prefer magic byte detection** when available (binary files).
2. **When detection fails, derive MIME type from the extension** using a trusted server-side mapping (e.g., the `mime-types` npm package), rather than trusting the client's declared MIME type.
3. **Fall back to the declared extension** — this is more acceptable because extensions are less security-sensitive than MIME types (MIME types control how browsers interpret content).

Something like:

```typescript
import mime from 'mime-types';

export const extractFileInfo = async ({
  file,
  declaredMimeType,
  filename,
}: {
  file: Buffer;
  declaredMimeType: string | undefined;
  filename: string;
}) => {
  const { ext: declaredExt } = buildFileInfo(filename);

  const detectedFileType = await FileType.fromBuffer(file);

  // For extension: prefer detected, fall back to declared (acceptable)
  const ext = detectedFileType?.ext ?? declaredExt;

  // For MIME type: prefer detected, then derive from extension,
  // only trust declared as last resort
  const mimeType =
    detectedFileType?.mime ?? mime.lookup(ext) || declaredMimeType;

  return { mimeType, ext };
};
```

## TL;DR — The reviewer's instinct is right

- **Never trust the declared MIME type** as a primary source. Derive it from the extension (via a trusted mapping) when magic bytes fail.
- **The declared extension is okay as a fallback** — it's less security-critical, and as you noted, multiple extensions can map to one MIME type so the client's choice is reasonable.
- An additional layer of defense would be a **whitelist of allowed MIME types** depending on the upload context, but that's a separate concern.

packages/twenty-server/src/engine/workspace-manager/workspace-migration/constant/standard-field-ids.ts
```diff
@@ -333,6 +333,7 @@ export const PERSON_STANDARD_FIELD_IDS = {
   phones: '20202020-0638-448e-8825-439134618022',
   city: '20202020-5243-4ffb-afc5-2c675da41346',
   avatarUrl: '20202020-b8a6-40df-961c-373dc5d2ec21',
+  avatarFile: '20202020-a7c9-4e3d-8f1b-2d5a6b7c8e9f',
```
prastoin: ***Remark*** shouldn't exist anymore

packages/twenty-server/src/database/commands/upgrade-version-command/1-18/1-18-migrate-person-avatar-files.command.ts
```diff
@@ -0,0 +1,291 @@
...
+
+    if (!isDefined(personObjectMetadata)) {
+      this.logger.warn(
+        `Person object metadata not found for workspace ${workspaceId}, skipping`,
+      );
+
+      return;
+    }
+
+    const twentyStandardApplication = Object.values(
```
prastoin: ***Info:*** Could use `applicationService.findWorkspaceTwentyStandardAndCustomApplicationOrThrow`

packages/twenty-server/src/database/commands/upgrade-version-command/1-18/1-18-migrate-person-avatar-files.command.ts
```diff
@@ -0,0 +1,291 @@
...
+        `Twenty standard application not found for workspace ${workspaceId}`,
+      );
+
+      return;
+    }
+
+    const avatarFileUniversalIdentifier =
+      STANDARD_OBJECTS.person.fields.avatarFile.universalIdentifier;
+
+    let avatarFileFieldMetadata = getFlatFieldsFromFlatObjectMetadata(
```
prastoin: ***Polish:*** Could be facto to
```ts
      const flatFieldMetadata = findFlatEntityByUniversalIdentifier({
        flatEntityMaps: flatFieldMetadataMaps,
        universalIdentifier: STANDARD_OBJECTS.person.fields.avatarFile.universalIdentifier
      })
```

packages/twenty-server/src/database/commands/upgrade-version-command/1-18/1-18-migrate-person-avatar-files.command.ts
```diff
@@ -0,0 +1,291 @@
...
+        field.universalIdentifier === avatarFileUniversalIdentifier,
+    ) as FlatFieldMetadata | undefined;
+
+    if (!isDefined(avatarFileFieldMetadata)) {
+      this.logger.log(
+        `Avatar file field metadata not found for workspace ${workspaceId}, creating it`,
+      );
+
+      if (!isDryRun) {
+        await this.fieldMetadataService.createOneField({
```
prastoin: ***Bug:*** We should handle the edge case where an already existing custom field collides with the one we wanna create here

packages/twenty-server/src/database/commands/upgrade-version-command/1-18/1-18-migrate-person-avatar-files.command.ts
```diff
@@ -0,0 +1,291 @@
...
+          workspaceId,
+          ownerFlatApplication: twentyStandardApplication,
+        });
+
+        const { flatFieldMetadataMaps: updatedFlatFieldMetadataMaps } =
+          await this.workspaceCacheService.getOrRecompute(workspaceId, [
+            'flatFieldMetadataMaps',
+          ]);
+
+        avatarFileFieldMetadata = getFlatFieldsFromFlatObjectMetadata(
```
prastoin: same

packages/twenty-server/src/database/commands/upgrade-version-command/1-18/1-18-migrate-person-avatar-files.command.ts
```diff
@@ -0,0 +1,291 @@
...
+            field.type === FieldMetadataType.FILES &&
+            field.universalIdentifier === avatarFileUniversalIdentifier,
+        ) as FlatFieldMetadata | undefined;
+
+        if (!isDefined(avatarFileFieldMetadata)) {
+          this.logger.error(
+            `Failed to create avatarFile field metadata for workspace ${workspaceId}`,
+          );
+
+          return;
```
prastoin: Should throw so it fails the workspace upgrade process, also I would rather consume the `await this.fieldMetadataService.createOneField` flat field metadata return type instead of pulling it

packages/twenty-server/src/database/commands/upgrade-version-command/1-18/1-18-migrate-person-avatar-files.command.ts
prastoin: We should try catch this operation and print meaningful logs

packages/twenty-server/src/database/commands/upgrade-version-command/1-18/1-18-migrate-person-avatar-files.command.ts
```diff
@@ -0,0 +1,291 @@
...
+        }
+      }
+      this.logger.log(
+        `Created avatarFile field metadata for workspace ${workspaceId}`,
+      );
+    }
+
+    if (!isDefined(avatarFileFieldMetadata)) {
```
prastoin, line 195: throw and maybe avoid such mutation by having a dedicated method

packages/twenty-server/src/database/commands/upgrade-version-command/1-18/1-18-migrate-person-avatar-files.command.ts
```diff
@@ -0,0 +1,291 @@
...
+      `Found ${personsWithAvatars.length} person(s) with avatarUrl containing 'people' folder in workspace ${workspaceId}`,
+    );
+
+    const fileRepository = this.coreDataSource.getRepository(FileEntity);
+
+    for (const person of personsWithAvatars) {
+      assertIsDefinedOrThrow(person.avatarUrl);
+
+      try {
+        const { type: fileExtension } = extractFolderPathFilenameAndTypeOrThrow(
```
prastoin, line 236: ***Question:*** Have you checked all production occurrences in order to search for any some edge cases ? ( such as no extension etc )

packages/twenty-server/src/database/commands/upgrade-version-command/1-18/1-18-migrate-attachment-files.command.ts
```diff
@@ -0,0 +1,287 @@
...
+        `File field metadata not found for attachments in workspace ${workspaceId}, creating it`,
+      );
+
+      if (!isDryRun) {
+        await this.fieldMetadataService.createOneField({
+          createFieldInput: {
+            objectMetadataId: attachmentObjectMetadata.id,
+            type: FieldMetadataType.FILES,
+            name: 'file',
+            label: 'File',
```
prastoin: same here double check no other custom file collides

prastoin (commented):
Nice job ! Solid
Left few comments on the upgrade commands ! Please let me know

packages/twenty-server/src/engine/core-modules/file/files-field/files-field.service.ts
etiennejouan: yes, as discussed for a text file without declared extension

packages/twenty-server/src/engine/core-modules/file/utils/extract-file-info.utils.ts
```diff
@@ -1,23 +1,60 @@
...
+        FileStorageExceptionCode.INVALID_EXTENSION,
+        {
+          userFriendlyMessage: msg`The file extension doesn't match the file content. Please check that your file is not corrupted and has the correct extension.`,
+        },
+      );
+    }
 
-  const ext = detectedFileType?.ext ?? declaredExt;
+    mimeType = mimeTypeFromExtension ?? 'application/octet-stream';
+  }
```
prastoin, line 54: you're the king

packages/twenty-server/src/engine/core-modules/file/entities/file.entity.ts
```diff
@@ -56,4 +56,7 @@ export class FileEntity extends WorkspaceRelatedEntity {
 
   @Column({ nullable: true, type: 'jsonb' })
   settings: FileSettings | null;
+
+  @Column({ nullable: true, type: 'varchar' })
+  mimeType: string | null;
```
prastoin: Is it really nullable ?

prastoin: Would type as enum from lib too

prastoin (approved):
Looks good to me ! 
Ultimate nitpick about the mimeType typing and nullability

PR #17755, opened 2026-02-06
Fixes security vulnerabilities associated to "tar" transient dependency.

mabdullahabaid:
The only places that import "tar" version 6.x are zapier-platform-cli and mintlify, both not associated to the server/front.

This PR fixes dependabot alerts associated to the use of tar 6.x by upgrading parent packages. 

[Dependabot Alert 424](https://github.com/twentyhq/twenty/security/dependabot/424), [Dependabot Alert 425](https://github.com/twentyhq/twenty/security/dependabot/425) and [Dependabot Alert 426](https://github.com/twentyhq/twenty/security/dependabot/426).

Not sure if it also fixes [Dependabot Alert 400](https://github.com/twentyhq/twenty/security/dependabot/400) since root yarn.lock holds reference to tar 6.x, but in case it does not, I will silence the alert since `zapier-platform-cli` and `mintlify` are used in `twenty-zapier` and `twenty-docs` respectively, not `twenty-server` or `twenty-front`.


mabdullahabaid:
I have verified the following.
- Bcrypt upgrade with login and signup using password - hashed password is stored, I am able to login with the new password, and I can login with the seeded `tim@apple.dev` identity.
- Sharp upgrade by uploading picture for object record, logged in member and workspace member.
- Node-gyp upgrade by triggering a background job - created a new record and the webhook was triggered with worker showing logs for the job.

I do not see any issues caused by the bcrypt upgrade since it is backward compatible, but might be good if someone tests before merge.

mabdullahabaid:
Closing this PR and creating a more targeted one for remaining alerts since another PR getting merged resolved a few alerts.

PR #17758, opened 2026-02-06
Fix record page layout BE

Weiko:
## Context
- Add missing conditional display
- Set default tab as null for most of the page layout, letting the FE handle that
- Fix duplicate "Note" widget in both task and note pages
- Simplify typing, removing redundant layoutName


Devessier (approved):
I have nothing to say. It looks like a true masterpiece. I'm excited to connect your work to the frontend!

PR #17759, opened 2026-02-06
Store record page layouts in a global state

Devessier:


packages/twenty-front/src/modules/users/components/MetadataProviderEffect.tsx
```diff
@@ -42,6 +46,8 @@ export const MetadataProviderEffect = () => {
   const [localIsCurrentUserLoaded, setLocalIsCurrentUserLoaded] =
     useState(false);
   const [localAreViewsLoaded, setLocalAreViewsLoaded] = useState(false);
+  const [localAreRecordPageLayoutsLoaded, setLocalAreRecordPageLayoutsLoaded] =
```
Weiko: I wouldn't add this to MetadataProvider and wouldn't make it blocking. 
Page layouts will load in the background and populate the cache when ready. Pages that need them will simply return undefined from useBasePageLayout until the data arrives.
What do you think? @Devessier

Weiko (approved):
LGTM

PR #17762, opened 2026-02-06
[FRONT COMPONENTS] Navigate from the remote

bosiraphael:
## PR Description

This PR:
- Introduces a `FrontComponentHostCommunicationApi`, which allows us to pass functions to be executed from the worker
- Exposes a navigate function from the host to the front component remote workers, enabling SDK components to trigger in-app navigation
- Gets rid of `useSyncExternalStore` and makes the execution context reactive without it
- Refactors and improves the `esbuild` plugin system

## Video

https://github.com/user-attachments/assets/7b26a1c2-f85f-4898-a71d-f60c70e61711


Weiko (approved):
LGTM

PR #17764, opened 2026-02-06
Add workflow prefix to code step source path

thomtrp:
Code step functions are not built, neither on activation, neither when testing workflow. Because `isWorkflowCodeStepLogicFunction` returns false. Source path needs the prefix workflow.

Also update code step duplication to copy the code but recreate a new function


thomtrp:
@greptile-ai

PR #17769, opened 2026-02-06
feat: improve AI chat - system prompt, tool output, context window display

FelixMalfait:
⚠️ **AI-generated PR — not ready for review** ⚠️

cc @FelixMalfait

---

### System prompt improvements
- Explicit skill-before-tools workflow to prevent the model from calling tools without loading the matching skill first
- Data efficiency guidance (default small limits, use filters)
- Pluralized `load_skill` → `load_skills` for consistency with `load_tools`

### Token usage reduction
- Output serialization layer: strips null/undefined/empty values from tool results
- Lowered default `find_*` limit from 100 → 10, max from 1000 → 100

### System object tool generation
- System objects (calendar events, messages, etc.) now generate AI tools
- Only workflow-related and favorite-related objects are excluded

### Context window display fix
- **Bug**: UI compared cumulative tokens (sum of all turns) against single-request context window → showed 100% after a few turns
- **Fix**: Track `conversationSize` (last step's `inputTokens`) which represents the actual conversation history size sent to the model
- New `conversationSize` column on thread entity with migration

### Workspace AI instructions
- Support for custom workspace-level AI instructions


packages/twenty-front/src/generated-metadata/graphql.ts
```diff
@@ -76,12 +76,13 @@ export type Agent = {
 export type AgentChatThread = {
   __typename?: 'AgentChatThread';
   contextWindowTokens?: Maybe<Scalars['Int']>;
+  conversationSize: Scalars['Int'];
```
FelixMalfait, line 92: Is this the sum of input token + output token? I wonder if we really need it.

FelixMalfait:
Let's also try Cerebras GLM 4.7. And also add Google / Mistral as options (doesn't mean we add it in prod)

FelixMalfait:
Let's improve the model selector, maybe something like Relevance does. Show cost, speed, etc

FelixMalfait:
Let's add a way to cancel an ongoing conversation ; also can I queue a message? Let's try to have a Cursor like behavior

FelixMalfait:
When switching to settings it automatically canceled the ongoing conversation

FelixMalfait:
Let's try tabs for last conversation history

FelixMalfait:
Let's fix MCP to also expose learn and execute tool

FelixMalfait:
Let's upgrade to AI SDK 6

packages/twenty-server/src/engine/metadata-modules/ai/ai-models/constants/groq-models.const.ts
```diff
@@ -0,0 +1,16 @@
+import { type AIModelConfig, ModelProvider } from './ai-models-types.const';
+
+export const GROQ_MODELS: AIModelConfig[] = [
+  {
+    modelId: 'openai/gpt-oss-120b',
```
FelixMalfait, line 5: Let's add an author enum, so we can differentiate author from provider (here author is OpenAI not groq)

Maybe we should also make it more robust and prefix by the provider or something similar? (e.g. if Bedrock and Groq or OpenRouter have the same model name)

packages/twenty-server/src/engine/core-modules/tool-provider/services/tool-registry.service.ts
```diff
@@ -52,15 +57,78 @@ export type ToolContext = {
...
 
+  private async getCachedGeneration(
+    context: ToolProviderContext,
+  ): Promise<CachedToolGeneration> {
+    const metadataVersion =
+      (await this.workspaceCacheStorageService.getMetadataVersion(
+        context.workspaceId,
+      )) ?? 0;
+
+    const cacheKey =
```
FelixMalfait: Avoid "as const" if possible?(look for all as X and try to get rid of them, if possible)

FelixMalfait: This seems like the legacy way of caching. Did we try with the new way? (has RAM caching on top of REDIS cache / seem smarter)

packages/twenty-server/src/engine/core-modules/tool-provider/services/tool-registry.service.ts
```diff
@@ -52,15 +57,78 @@ export type ToolContext = {
...
+          if (await provider.isAvailable(context)) {
+            const providerTools = await provider.generateTools(context);
+
+            Object.assign(tools, providerTools);
+            index.push(
+              ...this.toolSetToIndex(providerTools, provider.category),
+            );
+          }
+        }
+
```
FelixMalfait, line 104: When we add a new tool (= a serverless function) it doesn't change the metadata version, so cache wouldn't be updated / it wouldn't be visible?

packages/twenty-server/src/engine/core-modules/tool-provider/services/tool-registry.service.ts
```diff
@@ -52,15 +57,78 @@ export type ToolContext = {
...
 };
 
+type CachedToolGeneration = {
+  tools: ToolSet;
+  index: ToolIndexEntry[];
+};
+
+const MEMOIZER_TTL_MS = 60_000;
+
+const PROVIDER_TO_INDEX_CATEGORY: Record<
```
FelixMalfait: I feel like Provider = Category. Maybe we should just keep category? For me provider = model provider = groq, aws, openai, etc

FelixMalfait:
What happens if we get to token limit? Let's research online best practises for prompt compression and implement it (be proactive and compress before the error is received, since we know the limit for each model)

FelixMalfait:
@claude please run lint --fix/typecheck on front/server/shared ; then fix everything and push to this branch once it's all green

FelixMalfait:
@claude please run lint --fix/typecheck on front/server/shared ; then fix everything and push to this branch once it's all green

packages/twenty-front/src/modules/ai/utils/getWebSearchToolDisplayMessage.ts
```diff
@@ -41,16 +41,60 @@ const extractLoadingMessage = (input: ToolInput): string => {
   return 'Processing...';
 };
 
+// For execute_tool, unwrap the inner arguments so display logic sees the real tool's input
```
FelixMalfait: It looks like we have this in ToolStepRenderer ; should it be shared? Also web_search souldn't really be called through execute_tool (although I understand why we did it since the model can make mistake, I wonder if the code footprint of that adjustment could be smaller? Otherwise it might not be worth it)

packages/twenty-front/src/modules/ai/utils/getToolDisplayMessage.ts
```diff
@@ -41,16 +41,60 @@ const extractLoadingMessage = (input: ToolInput): string => {
...
+    return {
+      resolvedInput: input.arguments as ToolInput,
+      resolvedToolName: String(input.toolName),
+    };
+  }
+
+  return { resolvedInput: input, resolvedToolName: toolName };
+};
+
+const extractLearnToolNames = (input: ToolInput): string => {
```
FelixMalfait, line 64: I don't understand what this is doing in a file related to webSearch?

packages/twenty-front/src/pages/settings/ai/SettingsAIPrompts.tsx
```diff
@@ -0,0 +1,269 @@
...
+      ]}
+    >
+      <SettingsPageContainer>
+        {promptSections.map((section) => (
+          <Section key={section.title}>
+            <H2Title
+              title={section.title}
+              description={
+                section === promptSections[0]
+                  ? totalTokensDescription
```
FelixMalfait: Showing the total token description in the first section here is confusing since it does not relate to that section, let's display it in the title (even if it's not pretty)

packages/twenty-server/src/engine/core-modules/record-crud/tool-factory/direct-record-tools.factory.ts
```diff
@@ -106,10 +109,47 @@ export const createDirectRecordToolsFactory = (deps: DirectRecordToolsDeps) => {
...
+
+      tools[createManyKey] = {
+        description: `Create multiple ${objectMetadata.labelPlural} records in a single call. Provide an array of records, each containing the required fields. Maximum 20 records per call. Returns the results for each record creation.`,
+        inputSchema: generateCreateManyRecordInputSchema(
+          objectMetadata,
+          restrictedFields,
+        ),
+        execute: async (parameters) => {
+          const results = await Promise.all(
+            parameters.records.map((record: Record<string, unknown>) =>
```
FelixMalfait: This will do many queries in the db, how costly would it be to do a really createMany? same question for update many? I think they have been implemented in the common api so it shouldn't be too costly? If it does add a lot of code/complexity let's at least add a TODO for now

packages/twenty-server/src/engine/core-modules/record-crud/tool-factory/direct-record-tools.factory.ts
```diff
@@ -133,10 +173,55 @@ export const createDirectRecordToolsFactory = (deps: DirectRecordToolsDeps) => {
           });
         },
       };
+
+      const updateManyKey = `update_many_${camelToSnakeCase(objectMetadata.namePlural)}`;
+
+      tools[updateManyKey] = {
```
FelixMalfait: See comment on create many ; do one db query

packages/twenty-server/src/engine/core-modules/record-crud/tool-factory/direct-record-tools.factory.ts
```diff
@@ -133,10 +173,55 @@ export const createDirectRecordToolsFactory = (deps: DirectRecordToolsDeps) => {
...
+              (result) => result.recordReferences ?? [],
+            ),
+          };
+        },
+      };
     }
 
     if (canDelete) {
-      tools[`soft_delete_${objectMetadata.nameSingular}`] = {
+      tools[`soft_delete_${camelToSnakeCase(objectMetadata.nameSingular)}`] = {
```
FelixMalfait: Let's drop the "soft_" prefix

packages/twenty-front/src/pages/settings/ai/SettingsAIPrompts.tsx
```diff
@@ -0,0 +1,269 @@
...
+  );
+
+  const handleWorkspaceInstructionsInit = () => {
+    if (currentWorkspace?.aiAdditionalInstructions !== undefined) {
+      setWorkspaceInstructions(currentWorkspace.aiAdditionalInstructions ?? '');
+      setOriginalInstructions(currentWorkspace.aiAdditionalInstructions ?? '');
+    }
+  };
+
+  // Initialize workspace instructions when workspace data loads
```
FelixMalfait: Remove all useless comments like this (but keep the useful one that bring real value)

FelixMalfait (commented):
Partial review

packages/twenty-server/src/engine/core-modules/record-crud/tool-factory/direct-record-tools.factory.ts
FelixMalfait: Addressed: Renamed `soft_delete_` to `delete_` and renamed the schema file from `soft-delete-tool.zod-schema.ts` to `delete-tool.zod-schema.ts`, updating all exports.

packages/twenty-front/src/pages/settings/ai/SettingsAIPrompts.tsx
FelixMalfait: Addressed: Removed the useless comments ("Initialize workspace instructions when workspace data loads" and "Separate prompt sections from user-editable ones").

packages/twenty-front/src/pages/settings/ai/SettingsAIPrompts.tsx
FelixMalfait: Addressed: Moved the total token count to the page title (e.g. "System Prompt (~2.3k tokens)") instead of showing it in the first section's description.

packages/twenty-front/src/modules/ai/utils/getToolDisplayMessage.ts
FelixMalfait: Addressed: Renamed the file from `getWebSearchToolDisplayMessage.ts` to `getToolDisplayMessage.ts` since it contains generic tool display logic (not just web search). Updated all imports.

packages/twenty-server/src/engine/core-modules/tool-provider/services/tool-registry.service.ts
FelixMalfait: Addressed: Renamed `PROVIDER_TO_INDEX_CATEGORY` to `TOOL_CATEGORY_TO_INDEX_CATEGORY` to avoid confusion with model providers (Groq, OpenAI, etc.).

packages/twenty-server/src/engine/core-modules/tool-provider/services/tool-registry.service.ts
FelixMalfait: Addressed: Removed the `as const` from the cache key template literal in tool-registry.service.ts, and also from LEARN_TOOLS_TOOL_NAME and EXECUTE_TOOL_TOOL_NAME constants. The remaining `as` casts in `extractJsonSchema` / `stripInternalFieldsFromSchema` are necessary since they narrow from `unknown` after runtime type checks — TypeScript doesn't support narrowing `unknown` through `typeof`/`in` checks alone.

packages/twenty-server/src/engine/core-modules/record-crud/tool-factory/direct-record-tools.factory.ts
FelixMalfait: Addressed: Added TODO comments on both createMany and updateMany to replace the Promise.all loop with real batch DB operations (createMany/updateMany from the common API) to reduce round-trips.

packages/twenty-server/src/engine/core-modules/record-crud/tool-factory/direct-record-tools.factory.ts
FelixMalfait: Addressed: Added TODO (same as createMany above).

packages/twenty-front/src/modules/ai/utils/getWebSearchToolDisplayMessage.ts
FelixMalfait: Addressed: Exported `resolveToolInput` from `getToolDisplayMessage.ts` (renamed from `getWebSearchToolDisplayMessage.ts`) and refactored `ToolStepRenderer.tsx` to use it instead of duplicating the execute_tool unwrapping logic. This also fixes an inconsistency where ToolStepRenderer was checking for `'input'` property while the schema uses `'arguments'`.

packages/twenty-server/src/engine/core-modules/tool-provider/services/tool-registry.service.ts
FelixMalfait: Investigated: The codebase has `CacheStorageService` (RAM + Redis via NestJS cache-manager) and the more advanced `WorkspaceCacheService` (multi-tier with hash validation). The current `PromiseMemoizer` is process-local only with no Redis persistence.

Recommended approach: hybrid — keep PromiseMemoizer for short-lived request deduplication (10s TTL), add `CacheStorageService` for persistent Redis-backed caching. This matches the pattern used by `WorkspaceCacheService`. Will implement in a follow-up to keep this PR focused.

packages/twenty-server/src/engine/core-modules/tool-provider/services/tool-registry.service.ts
FelixMalfait: Good catch: when a new serverless function is deployed, the metadata version doesn't change, so the tool cache won't be invalidated. This means newly deployed functions won't appear until the 60s memoizer TTL expires.

Options:
1. Bump metadata version when serverless functions are deployed
2. Add a separate cache version for tools that gets incremented on function deployments
3. Listen for function deployment events and clear the tool cache explicitly

Will address in a follow-up — this is an existing issue independent of the caching approach.

packages/twenty-front/src/generated-metadata/graphql.ts
FelixMalfait: The `conversationSize` field stores the cumulative token count (input + output) for the conversation. It's used by the frontend `AIChatContextUsageButton` to display context window usage against the model's max context.

If we don't need the UI indicator, we can drop this field. Otherwise it's needed to show the user how close they are to the context limit. Happy to remove it if you think the UI indicator isn't useful enough to justify the extra field.

packages/twenty-server/src/engine/metadata-modules/ai/ai-models/constants/groq-models.const.ts
FelixMalfait: Acknowledged. Will add a `ModelAuthor` enum (e.g. OPENAI, META, etc.) to differentiate the model creator from the hosting provider. For the model ID prefix concern — we could prefix with provider like `groq/openai/gpt-oss-120b` or `groq:openai/gpt-oss-120b` to avoid collisions when the same model is available from multiple providers. Will implement in a follow-up.

FelixMalfait:
"Ask AI" title of command menu should be the conversation title instead

FelixMalfait:
Will continue in a followup PR

PR #17773, opened 2026-02-07
Fix row level permission predicate group creation order

camilo-agudelo-uma:
Split the upsertRowLevelPermissionPredicates call into two sequential operations:
1. First create/update predicate groups
2. Then create/update predicates that reference those groups

This prevents the error "Could not find rowLevelPermissionPredicateGroup for given rowLevelPermissionPredicateGroupId" when creating predicates that reference newly created groups in the same request.

### DIFF
--- a/packages/twenty-front/src/modules/settings/roles/role/hooks/useSaveDraftRoleToDB.ts
+++ b/packages/twenty-front/src/modules/settings/roles/role/hooks/useSaveDraftRoleToDB.ts
@@ -317,10 +317,10 @@ export const useSaveDraftRoleToDB = ({
         objectUsedGroupIds.has(group.id),
       );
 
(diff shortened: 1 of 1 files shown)


charlesBochet:
@Weiko could you take a look? If possible, we should avoid making several calls from the frontend, this might lead to flaky behaviors if network connection is unstable

Weiko:
@charlesBochet a fix has already been merged this week, addressing this issue in the backend already
@camilo-agudelo Hi, as mentioned above, we preferred to patch the backend (a regression was introduced by a big refactoring recently but didn't justify the necessity to query the backend with 2 separate calls). I'm closing this PR but feel free to reopen an issue if you still face it. Thanks again for your contribution

PR #17774, opened 2026-02-08
Fix importing people with mixed-case domain URL failing to match company

bugisthegod:
Fix #17711 
### Reproduce steps
1. import this xlsx file [test-import.xlsx](https://github.com/user-attachments/files/25156642/test-import.xlsx)

2. import company worksheet at first and then import people worksheet
3. you will see an error  Snipaste_2026-02-07_19-22-04

And I found neither the frontend nor the backend applies lowercase normalization to the query value. Although the standard UI input always displays company links in lowercase, user might mixed the case in their xlsx/csv bulk import. So I did a simple check in frontend.

### Additional findings:
1. Email has the same case sensitivity issue when opportunities import. You can test same as in test-import.xlsx file (I created a opportunities worksheet. It will have same issue: email error

2. API also has: I also tested via the GraphQL API Playground and confirmed that createOnePerson with a mixed-case URL fails to find an existing company.  Snipaste_2026-02-07_23-33-56

### Approach:
1. Only change it in frontend, and I will add email check later; (my current commit)
2. Or backend: Normalize values in computeUniqueConstraintCondition in twenty-server/src/engine/twenty-orm/utils/compute-relation-connect-query-configs.util.ts (which handles connect.where queries, and this will cover xlsx import and api import: createone, createmany, updatemany, updateone. And much better for future extension.

(Personally, I'd prefer the backend approach. But I'd love to hear your thoughts on which approach you'd prefer, and whether my analysis is on the right track.

### DIFF
--- a/packages/twenty-front/src/modules/object-record/spreadsheet-import/utils/buildRecordFromImportedStructuredRow.ts
+++ b/packages/twenty-front/src/modules/object-record/spreadsheet-import/utils/buildRecordFromImportedStructuredRow.ts
@@ -52,6 +52,9 @@ const buildRelationConnectFieldRecord = (
   fieldMetadataItem: FieldMetadataItem,
   importedStructuredRow: ImportedStructuredRow,
   spreadsheetImportFields: SpreadsheetImportFields,
+  compositeFieldTransformConfigs: Partial<
+    Record<FieldMetadataType, Record<string, ((value: any) => any) | undefined>>
+  >,
 ) => {
   if (fieldMetadataItem.relation?.type !== RelationType.MANY_TO_ONE)
     return undefined;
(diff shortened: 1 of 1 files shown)


etiennejouan (approved):
Hi @bugisthegod Thank you for your contribution 🙏 

Ideally we should also fix this issue back end side for all API users. In the data-arg.processor, we should transform relation connect values as you did it here. You can give a try if you want. Feel free to reach me if not clear.

bugisthegod:
Hi, @etiennejouan  I'm implementing the backend normalization for this issue. Where should I apply the URL lowercase logic?

**Option 1:** In `DataArgProcessor.processField()` (early validation layer)

**Option 2:** In `computeUniqueConstraintCondition()` at `compute-relation-connect-query-configs.util.ts:297`

Which approach fits better with Twenty's architecture?
（By the way, should I apply email lowercase logic when people import opportunities with email connect field?

Thanks!

PR #17775, opened 2026-02-08
Lowercase email

### LINKED ISSUE #120: Email field should always be lowercase
## Scope & Context

Currently, users are able to create the same email but with different cases.
Emails should always be stored in lowercase.

## Technical inputs

For the `emailsPrimaryEmail` and `emailsSecondaryEmails`:
- Add frontend validation
- Add backend validation (there is no backend validation for the emails at the moment, we don't even check that it's a valid email) and return an error if the email is not in lowercase
- Create a script to set all the existing emails to lowercase

BOHEUS:
Fixes https://github.com/twentyhq/core-team-issues/issues/120 #16976
Partially related to #17711

Frontend check surprisingly was one-liner covering both email input and import files

Migration script will be done in next commit


packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/validator-utils/validate-additional-email-array-or-throw.util.ts
```diff
@@ -0,0 +1,36 @@
...
+import { isNull } from '@sniptt/guards';
+import { msg } from '@lingui/core/macro';
+
+import {
+  CommonQueryRunnerException,
+  CommonQueryRunnerExceptionCode,
+} from 'src/engine/api/common/common-query-runners/errors/common-query-runner.exception';
+import { validateEmailValueOrThrow } from 'src/engine/api/common/common-args-processors/data-arg-processor/validator-utils/validate-email-value-or-throw.util';
+
+export const validateAdditionalEmailArrayOrThrow = (
```
charlesBochet: shouldn't we rather convert to lowercase instead of throwing @BOHEUS ?

BOHEUS: I'm not entirely sure about the approach whether it's better to throw an error to user or simply convert it, I just went with the former as I think it's better to inform user about an error and let them fix on their own rather than converting it to lowercase and having user surprised why email is changed

Speaking of converting to lowercase, there's already a convertion in transform-emails-value.util.ts

Let me know which approach is better and I'll make according changes

charlesBochet: I'm lost, so is the lowercasing already in place? It seems so as you mentioned the transform-emails-value.util.ts

I don't think we should block honestly, we are normalizing the input which makes sense

BOHEUS: Basically:
- lowercasing is already in place in transform-emails-value.util.ts (__I missed it earlier focusing mainly on the issue and validation__)
- there's no validation whether email is lowercase or not (which can be redundant if there's already lowercasing in place as I'm thinking now but I still think it's better to inform user that they should change it on their own and remove lowercasing)
- there's no validation whether email is really a email meaning if someone creates/updates a record via API with some kind of string instead of email, it'll be accepted instead of rejected (front has a validation so only backend is affected and it must be fixed)
- lowercasing email has been implemented almost 3 months ago from what I can see (#15491) so I'm not sure if earlier emails were lowercased by default or not, I'd assume they weren't

Given everything, should I remove lowercase validation (allowing emails to be lowercased by default) or remove default lowercasing? Email format validation must be added as well as migration command to run to ensure that all emails are really lowercase, please let me know which changes should I implement

charlesBochet: I think I remember that a command to lowercase was actually run

So I would say, we can have:
- keep the lowercasing we already have in place
- no need for the command
- let's use validation to block non valid emails => here I wonder if it should happen at a different / same place than the lowercasing. It feels weird to validate, then lowercase, I feel it's the same process @etiennejouan what do you think?

etiennejouan: I'm not sure we want to reject uppercase email. Data arg processor validate and transform each fields. Concerning EMAILS, it check the emails object format then it transforms with `transformEmailsValue` (lowercasing all emails - then no need for this PR's migration).


@BOHEUS when checking the issues you link to this PR, I think [this one](https://github.com/twentyhq/twenty/issues/16976) is relevant (the others should have been closed). We should have the same transformation on filter value to be sure condition matches. I'll include it in this [sprint](https://github.com/twentyhq/core-team-issues/issues/1627)

BOHEUS: Thanks @etiennejouan, should I apply changes proposed by Charles or do something more than that?

etiennejouan: I think that changes proposed by Charles are already live. I'll fix filtering issues on email. Then, I'm sorry but I think we should close this PR. Do not hesitate to reach me it seems unclear.

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/validator-utils/validate-additional-email-array-or-throw.util.ts
```diff
@@ -1,27 +1,30 @@
...
 
+import { validateEmailsPrimaryEmailSubfieldOrThrow } from 'src/engine/api/common/common-args-processors/data-arg-processor/validator-utils/validate-emails-primary-email-subfield-or-throw.util';
 import {
   CommonQueryRunnerException,
   CommonQueryRunnerExceptionCode,
 } from 'src/engine/api/common/common-query-runners/errors/common-query-runner.exception';
-import { validateEmailValueOrThrow } from 'src/engine/api/common/common-args-processors/data-arg-processor/validator-utils/validate-email-value-or-throw.util';
 
-export const validateAdditionalEmailArrayOrThrow = (
+export const validateEmailsAdditionalEmailsSubfieldOrThrow = (
```
etiennejouan: nitpick : renamed

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/validator-utils/validate-emails-primary-email-subfield-or-throw.util.ts
```diff
@@ -25,9 +25,7 @@ export const validateEmailValueOrThrow = (
     );
   }
 
-  if (
-    !z.email().trim().safeParse(value).success
-  ) {
+  if (!z.email().trim().safeParse(value).success && isNonEmptyString(value)) {
```
etiennejouan: this one important - when testing on UI and trying to remove an email, it sends empty string to the backend and then throws an error

packages/twenty-server/test/integration/graphql/suites/inputs-validation/create-validation/__snapshots__/emails-field-create-input-validation.integration-spec.ts.snap
```diff
@@ -2,4 +2,28 @@
 
 exports[`Create input validation - EMAILS Gql create input - failure EMAILS - should fail with : {"emailsField":"not-an-email"} 1`] = `"Expected type "EmailsCreateInput" to be an object."`;
 
+exports[`Create input validation - EMAILS Gql create input - failure EMAILS - should fail with : {"emailsField":{"additionalEmails":"not-an-email"}} 1`] = `"Invalid string value 'not-an-email' for email field "emailsField.additionalEmails""`;
```
etiennejouan, line 5: when updating integration test, you can launch this command to be sure test is ok + to update snapshot if needed

Image

etiennejouan (approved):
Thanks @BOHEUS nice PR and good catch ! 🙏 

I comment the change I made, feel free to reach me if not clear.

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/validator-utils/validate-additional-email-array-or-throw.util.ts
BOHEUS: While renaming, you probably duplicated a file by accident, I deleted it

packages/twenty-server/test/integration/graphql/suites/inputs-validation/create-validation/__snapshots__/emails-field-create-input-validation.integration-spec.ts.snap
BOHEUS: Could you please write down a command I can run to check if test is OK? I don't have such dropdown in my IDE

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/validator-utils/validate-emails-primary-email-subfield-or-throw.util.ts
BOHEUS: Why do I feel like it's somewhat related? https://github.com/twentyhq/core-team-issues/issues/183

etiennejouan: it's an intended behaviour. We accept empty string but we write null in db. Issue closed

packages/twenty-server/test/integration/graphql/suites/inputs-validation/create-validation/__snapshots__/emails-field-create-input-validation.integration-spec.ts.snap
etiennejouan: `nx run twenty-server:jest --config ./jest-integration.config.ts packages/twenty-server/test/integration/graphql/suites/inputs-validation/create-validation/emails-field-create-input-validation.integration-spec.ts --silent=false --updateSnapshot`

BOHEUS: Thanks!

PR #17776, opened 2026-02-08
New onboarding plan

BOHEUS:
Fixes github.com/twentyhq/core-team-issues/issues/637


packages/twenty-front/src/modules/app/hooks/useInitializeQueryParamState.ts
```diff
@@ -45,6 +46,34 @@ export const useInitializeQueryParamState = () => {
               );
             }
           },
+          plan: (value: string) => {
```
charlesBochet: I'm a bit lost here, why do we need a dedicated entry for plan? I would expect it to be part of billingCheckoutSession handled above?

packages/twenty-front/src/pages/onboarding/internal/ChooseYourPlanContent.tsx
```diff
@@ -260,17 +256,25 @@ export const ChooseYourPlanContent = ({ billing }: { billing: Billing }) => {
...
           href={calendarBookingPageId ? AppPath.BookCall : CAL_LINK}
           target={calendarBookingPageId ? '_self' : '_blank'}
           rel={calendarBookingPageId ? '' : 'noreferrer'}
         >
           <Trans>Book a Call</Trans>
         </ClickToActionLink>
+        <span />
+        {currentPlanKey === BillingPlanKey.PRO ? (
+          <ClickToActionLink
+            href={AppPath.PlanRequired.concat('?plan=enterprise')}
```
charlesBochet: this concatenation does not seem to be enough, we should create a link with a full billingSessionState

Image

charlesBochet (changes requested):
Thanks @BOHEUS, I've left comments

packages/twenty-front/src/modules/app/hooks/useInitializeQueryParamState.ts
BOHEUS: According to the issue, new `plan` parameter should be introduced, I assumed it should be only within planRequired page when user wants to change the plan and billingCheckoutSession should be left as is in pricing page

packages/twenty-front/src/pages/onboarding/internal/ChooseYourPlanContent.tsx
BOHEUS: @charlesBochet just to be clear, said `plan` parameter shouldn't be used and instead should we use billingCheckoutSession everywhere? If so, I'll quickly add changes

charlesBochet: yes, I think so, that's the purpose of this existing query param, let's use it

BOHEUS: On it

charlesBochet (approved):

charlesBochet:
@BOHEUS could you fix the conflits and we are good to go :)

BOHEUS:
Done @charlesBochet

PR #17777, opened 2026-02-08
Validate timezone input in group-by date queries

FelixMalfait:
## Summary

- Validates that the timezone parameter in group-by date expressions is a recognized IANA timezone 
- Adds SQL string literal escaping as a defense-in-depth measure for the timezone value interpolated into SQL expressions
- Moves the `IANA_TIME_ZONES` constant to `twenty-shared` so it can be reused across frontend and server packages
- Adds `INVALID_TIMEZONE` error code mapped to 400 Bad Request in both GraphQL and REST API exception handlers

## Test plan

- [x] Unit tests for `validateIanaTimeZone` (valid IANA zones, fixed offsets, rejects invalid strings)
- [x] Unit tests for `escapeSqlStringLiteral`
- [x] Integration tests for `getGroupByExpression` covering timezone validation and granularity handling


packages/twenty-front/src/modules/localization/constants/IanaTimeZones.ts
```diff
@@ -1,600 +1 @@
...
-  'US/Michigan',
-  'US/Mountain',
-  'US/Pacific',
-  'US/Samoa',
-  'UTC',
-  'W-SU',
-  'WET',
-  'Zulu',
-];
+export { IANA_TIME_ZONES } from 'twenty-shared/constants';
```
charlesBochet: what are we gaining from re-exporting?

charlesBochet (approved):

packages/twenty-front/src/modules/localization/constants/IanaTimeZones.ts
FelixMalfait: @claude can you please address Charles' comment and push the fix to this branch directly. Charles is right, we should directly import from twenty-shared and we don't need this intermediate re-export from twenty-front. Make sure to lint/typecheck before you push

PR #17778, opened 2026-02-08
Enforce SecureHttpClientService for outbound HTTP requests

FelixMalfait:
## Summary

- Adds an ESLint rule (`@typescript-eslint/no-restricted-imports`) to flag direct `axios` and `@nestjs/axios` `HttpService` imports, guiding developers toward `SecureHttpClientService` for outbound HTTP requests
- Renames `getSecureAdapter` to `getSecureAxiosAdapter` for clarity
- Migrates `SearchHelpCenterTool` to use `SecureHttpClientService` instead of raw `axios.post()`
- Adds targeted `eslint-disable` comments for legitimate uses (e.g. `axios.isAxiosError()` type guards, `HttpService` in code-interpreter where per-request adapter control is needed)

## Test plan

- [x] Existing unit tests for `getSecureAxiosAdapter` updated with the rename
- [ ] Verify lint rule triggers on new files importing axios directly
- [ ] Verify `SearchHelpCenterTool` still works with secure client

### DIFF
--- a/packages/twenty-server/eslint.config.mjs
+++ b/packages/twenty-server/eslint.config.mjs
@@ -251,6 +251,42 @@ export default [
     },
   },
 
+  // Security: Restrict unsafe HTTP clients
+  {
+    files: ['**/*.ts'],
+    // Exclude the files that implement the security mechanism
+    ignores: [
+      '**/secure-http-client.service.ts',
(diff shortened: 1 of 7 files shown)


FelixMalfait:
Superseded by a cleaner approach that migrates all callers instead of adding an ESLint rule with disable comments.

PR #17780, opened 2026-02-08
Fix Cloudflare webhook guard validation logic

FelixMalfait:
## Summary

- Restructures the `CloudflareSecretMatchGuard` to properly validate the `cf-webhook-auth` header before performing the comparison — checks for missing, empty, or mismatched-length values first
- Removes `@ts-expect-error` workarounds with explicit typed header access
- Expands test coverage with additional edge cases (missing header, wrong value, empty string, length mismatch)

## Test plan

- [x] Unit tests pass (6 tests covering matching, missing config, missing header, wrong value, empty string, length mismatch)

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/cloudflare/guards/cloudflare-secret.guard.ts
+++ b/packages/twenty-server/src/engine/core-modules/cloudflare/guards/cloudflare-secret.guard.ts
@@ -4,6 +4,7 @@ import {
   type CanActivate,
   type ExecutionContext,
   Injectable,
+  InternalServerErrorException,
 } from '@nestjs/common';
 
 import { timingSafeEqual } from 'crypto';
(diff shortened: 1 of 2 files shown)


FelixMalfait:
@claude fix the typecheck

 nx run twenty-server:typecheck
  
  > tsgo -p tsconfig.json
  
  Error: src/engine/core-modules/cloudflare/guards/cloudflare-secret.guard.ts(30,28): error TS2352: Conversion of type 'Headers' to type 'Record<string, unknown>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
    Index signature for type 'string' is missing in type 'Headers'.
  > tsgo -p tsconfig.json
  
  Error: src/engine/core-modules/cloudflare/guards/cloudflare-secret.guard.ts(30,28): error TS2352: Conversion of type 'Headers' to type 'Record<string, unknown>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  
  
Also the default behavior if Cloudflare secret is not configured should be to throw a server error not to accept it

Push directly to this branch/pr ; don't forget to lint

FelixMalfait:
@martmull Can you please check/take-over? Thanks!

martmull (approved):

PR #17781, opened 2026-02-08
Add billing health checkpoint

### LINKED ISSUE #80: Create a healthcheck endpoint for billing
We had a billing issue which took ~20 hours to detect.

I think it would be helpful to create a healthcheck endpoint for billing services which return 200 or 500 if any of the checks failed.

We can then add this endpoint to Cloudflare Healthchecks to receive an email / Discord notifcation.

One first check that could be added to this health check: for the last workspace created, but created more than 5 minutes ago, I should always find a subscription.

BOHEUS:
Fixes https://github.com/twentyhq/core-team-issues/issues/80

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/health/guards/healthz-secret.guard.ts
+++ b/packages/twenty-server/src/engine/core-modules/health/guards/healthz-secret.guard.ts
@@ -1,0 +1,45 @@
+import {
+  type CanActivate,
+  type ExecutionContext,
+  ForbiddenException,
+  Injectable,
+  InternalServerErrorException,
+} from '@nestjs/common';
+
+import { timingSafeEqual } from 'crypto';
+
+import { TwentyConfigService } from 'src/engine/core-modules/twenty-config/twenty-config.service';
+
+@Injectable()
+export class HealthzSecretGuard implements CanActivate {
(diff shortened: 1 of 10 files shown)


packages/twenty-server/src/engine/core-modules/health/guards/healthz-secret.guard.ts
```diff
@@ -0,0 +1,45 @@
...
+    const headerSecret = request.headers['x-healthz-secret'] as
+      | string
+      | undefined;
+
+    if (!headerSecret) {
+      throw new ForbiddenException('Missing X-Healthz-Secret header');
+    }
+
+    const isValid =
+      headerSecret.length === healthzSecret.length &&
```
Weiko, line 36: timing attack is still possible with this check

charlesBochet:
@BOHEUS Thanks for opening this PR,

I made a first commit to enhance it and add a secret on the endpoint but after discussing it with Felix and Thomas, we think that we should actually get rid of all these /healthz/:indicator endpoints as they should be served through prometheus.

i'm making another PR to implement this in a more standard way and closing this one :)

BOHEUS:
@charlesBochet thanks for the info :+1:

charlesBochet:
https://github.com/twentyhq/twenty/pull/17810

PR #17785, opened 2026-02-08
Remove unused GraphQL throttler plugin and graphql-rate-limit dependency

FelixMalfait:
## Summary

- Deletes `use-throttler.ts` — an envelop plugin that was never registered in the GraphQL config plugin chain (dead code since introduction)
- Removes the `graphql-rate-limit` dependency from `package.json` (only consumer was the deleted file)

API rate limiting continues to work via `ThrottlerService` in the query runner layer (for API key auth). Global coverage should be handled at the infrastructure level (Cloudflare).

## Test plan

- [x] Confirmed no imports of `useThrottler`, `ThrottlerPluginOptions`, or `UnauthenticatedError` from this file exist anywhere in the codebase
- [x] `graphql-rate-limit` has no other consumers


FelixMalfait:
This should be followed by an implementation at the twenty-infra level: https://github.com/twentyhq/twenty-infra/pull/391

PR #17791, opened 2026-02-09
[Fix] fix relations in apps

ijreilly:
When syncing an app with relation multiple times, it would fail because the relation fields could not be identified due to universalIdentifier being overwritten at field cretion

### DIFF
--- a/packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/validators/utils/validate-relation-creation-payload.util.ts
+++ b/packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/validators/utils/validate-relation-creation-payload.util.ts
@@ -43,7 +43,7 @@ export const validateRelationCreationPayload = async ({
         errors: [
           {
             code: FieldMetadataExceptionCode.FIELD_METADATA_RELATION_MALFORMED,
-            message: `Relation creation payload is invalid`,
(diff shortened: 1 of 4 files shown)


packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/validators/utils/validate-relation-creation-payload.util.ts
```diff
@@ -43,7 +43,7 @@ export const validateRelationCreationPayload = async ({
         errors: [
           {
             code: FieldMetadataExceptionCode.FIELD_METADATA_RELATION_MALFORMED,
-            message: `Relation creation payload is invalid`,
+            message: error.message ?? `Relation creation payload is invalid`,
```
ijreilly, line 46: unrelated, but needed to get helpful error message at app sync

(eg: `target fieldLabel is undefined` instead of `Relation creation payload is invalid`)

ijreilly: @prastoin in case there was a reason not to keep error.message

martmull (approved):

charlesBochet:
@ijreilly it looks that you just need to update snapshots

PR #17794, opened 2026-02-09
Use backend types in Record Page Layouts' frontend

Devessier:


packages/twenty-front/src/modules/page-layout/types/PageLayoutWidget.ts
```diff
@@ -12,6 +11,5 @@ export type PageLayoutWidget = Omit<
   'objectMetadataId' | 'configuration'
 > & {
   objectMetadataId?: Nullable<string>;
-  conditionalDisplay?: RulesLogic;
```
charlesBochet, line 15: why are we removing it?

charlesBochet (approved):
LGTM, added a quick question

packages/twenty-front/src/modules/page-layout/types/PageLayoutWidget.ts
Devessier: This is now defined by the backend. It's no longer needed to override the back-end type to add this property.

PR #17798, opened 2026-02-09
Serve frontend components

Weiko:
## Context
Ability to serve frontend component

See:
```typescript
curl -i 'http://localhost:3000/rest/front-components/35063b3f-bc4c-4358-8966-7762677802a3' \
--header 'Authorization: Bearer eyJhb...'
HTTP/1.1 200 OK
X-Powered-By: Express
Access-Control-Allow-Origin: *
Content-Type: application/javascript
Date: Mon, 09 Feb 2026 10:28:17 GMT
Connection: keep-alive
Keep-Alive: timeout=5
Transfer-Encoding: chunked

// react-globals:react/jsx-runtime
var jsx = globalThis.jsx;
var jsxs = globalThis.jsxs;
var Fragment = globalThis.React.Fragment;

// src/front-components/test.tsx
var RemoteComponents = globalThis.RemoteComponents;
var Component = () => {
  return /* @__PURE__ */ jsxs(RemoteComponents.HtmlDiv, { style: { padding: "20px", fontFamily: "sans-serif" }, children: [
    /* @__PURE__ */ jsx(RemoteComponents.HtmlH1, { children: "My new component!" }),
    /* @__PURE__ */ jsx(RemoteComponents.HtmlP, { children: "This is your front component: test" })
  ] });
};
var test_default = globalThis.jsx(Component, {});
export {
  test_default as default
};
//# sourceMappingURL=test.mjs.map
```

readFile_v2 returns a Node.js Stream object (Readable). Here we are using  stream pipeline which connects the readable stream (file) to the writable stream (HTTP response) which efficiently streams the file content directly to the HTTP response without loading the entire file into memory. (in chunks, handling backpressure and closing the connection when the file is fully sent)


martmull (approved):

packages/twenty-server/src/engine/metadata-modules/front-component/front-component.service.ts
```diff
@@ -282,4 +287,31 @@ export class FrontComponentService {
 
     return frontComponent;
   }
+
+  async getBuiltComponentStream({
```
charlesBochet, line 291: Note, not related to this PR:
to me we should really separate:

1) core-modules/front-component.service.ts (low level fc logic)
2) metadata-modules/front-component-api.service (high level service to serve the API)

like we have for logic function
IMO metadata-modules should become our highl level api service layer and everything else should be moved to core-modules

charlesBochet (approved):
LGTM, i've left a comment for later.

Would be also nice to add integration test on this endpoint (but can be done spearately)

packages/twenty-server/src/engine/metadata-modules/front-component/front-component.service.ts
Weiko: I agree! @charlesBochet

PR #17783, opened 2026-02-08
Harden local file storage driver path resolution

FelixMalfait:
## Summary

- Normalize all file paths with `path.resolve` instead of `join` to properly handle `..` segments in file path inputs
- Add `assertPathIsWithinStorage` guard on all write, delete, move, copy, and existence-check operations
- Introduce `ACCESS_DENIED` exception code with i18n-ready user-friendly message
- Read path already had realpath-based validation; updated its error code to `ACCESS_DENIED` for consistency

## Test plan

- [x] Typecheck passes
- [x] Lint passes
- [x] Manual: verify file upload/download still works with valid paths
- [x] Manual: verify `../` in file paths is rejected with ACCESS_DENIED

Made with [Cursor](https://cursor.com)

PR #17799, opened 2026-02-09
Validate safe storage path

Weiko:


packages/twenty-server/src/engine/workspace-manager/workspace-migration/workspace-migration-builder/validators/utils/is-storage-path-safe.util.ts
```diff
@@ -0,0 +1,19 @@
+import path from 'path';
+
+export const isStoragePathSafe = (storagePath: string): boolean => {
```
charlesBochet, line 3: @Weiko I think this should leave in file storage module and we should also add this to file-storage.service.ts

charlesBochet:
Very good idea, left comments!

Weiko:
Actually duplicate with https://github.com/twentyhq/twenty/issues/17783
Closing

PR #17800, opened 2026-02-09
Flush cache before and after upgrade for self-hosts

ijreilly:
as per title


packages/twenty-server/src/database/commands/command-runners/workspaces-migration.command-runner.ts
```diff
@@ -135,7 +135,7 @@ export abstract class WorkspacesMigrationCommandRunner<
 
     for (const [index, workspaceId] of workspaceIdsToProcess.entries()) {
       this.logger.log(
-        `Running command on workspace ${workspaceId} ${index + 1}/${workspaceIdsToProcess.length}`,
+        `Upgrading workspace ${workspaceId} ${index + 1}/${workspaceIdsToProcess.length}`,
```
ijreilly, line 138: i find this is clearer but can revert if someone disagrees

charlesBochet (approved):

PR #17813, opened 2026-02-09
[FRONT COMPONENTS] Retrieve the front components from the backend

### LINKED ISSUE #2207: [FRONT COMPONENTS] Retrieve the front component from the backend
Use the REST endpoint to retrieve the component from the backend instead of the mocked file.

bosiraphael:
- Pass the auth token to worker
- Fetch the file from the rest API with the auth token
- Create the blob url
- Update the stories to pass a fake token which will be ignored


bosiraphael:
@greptileai

charlesBochet (approved):

PR #17814, opened 2026-02-09
[Fix] fix timelineActivities on notes and tasks

ijreilly:
[Fixes sentry](https://twenty-v7.sentry.io/issues/7238708876/?environment=prod&environment=prod-eu&project=4507072499810304&query=Field%20metadata%20for%20field&referrer=issue-stream&sort=date): _Field metadata for field "targetTargetPersonId" is missing in object metadata timelineActivity_

Regression caused by migration of note and task targets - fields were renamed from personId to targetPersonId, impacting the event names, while we still expected them in their previous shape.


packages/twenty-server/src/modules/timeline/services/timeline-activity.service.ts
```diff
@@ -361,9 +354,12 @@ export class TimelineActivityService {
           targetColumnName
         ];
 
+        const objectSingularName =
+          extractObjectSingularNameFromTargetColumnName(targetColumnName);
+
         return {
           name: `linked-${activityType}.${action}`,
-          overrideObjectSingularName: targetColumnName.replace(/Id$/, ''),
+          overrideObjectSingularName: objectSingularName,
```
FelixMalfait: @claude it feels like overrideObjectSingularName since we don't override anything anymore, can you rename everywhere to just objectSingularName, then lint/typecheck and push to this branch

FelixMalfait (approved):

PR #17815, opened 2026-02-09
Fix RLS creation logic

Weiko:
Fix after resolveEntityRelationUniversalIdentifiers introduction
```
FlatEntityMapsException [Error]: Could not find rowLevelPermissionPredicateGroup for given rowLevelPermissionPredicateGroupId
        at resolveEntityRelationUniversalIdentifiers
```

- Creating util for RLS flat entity creation to be aligned with other entities
- Progressively update the flat entity maps as each new group is built, following the same optimistic pattern used by computeUniversalFlatEntityMapsFromTo in the validate build and run. The updated maps are now passed to computePredicateOperations so predicates can also resolve references to the newly created groups.
- Adding more coverage


prastoin (approved):

packages/twenty-server/src/engine/metadata-modules/flat-row-level-permission-predicate/utils/from-update-row-level-permission-predicate-input-to-flat-row-level-permission-predicate.util.ts
```diff
@@ -0,0 +1,56 @@
...
+        flatRowLevelPermissionPredicateGroupMaps,
+      },
+    });
+
+    return {
+      ...existingPredicate,
+      fieldMetadataId: input.fieldMetadataId,
+      fieldMetadataUniversalIdentifier,
+      operand: input.operand,
+      value: input.value ?? null,
```
ijreilly, line 43: is input.value always present ? No case where input.value is undefined because not present because we dont want to change its value?

ijreilly (approved):

PR #17824, opened 2026-02-09
Query cache instead of database for event listener webhook, logicFunction, triggers

charlesBochet:
## Fix
Replaced direct database queries with existing flat entity map caches for the two that already have cache infrastructure:
- **CallDatabaseEventTriggerJobsJob** - now uses flatLogicFunctionMaps cache via WorkspaceCacheService.getOrRecompute(), filtering in memory for non-deleted logic functions with databaseEventTriggerSettings.
- **CallWebhookJobsJob** - now uses flatWebhookMaps cache via WorkspaceCacheService.getOrRecompute(), filtering in memory for webhooks matching the event's operations.
- Note: **WorkflowDatabaseEventTriggerListener** - left as-is since WorkflowAutomatedTriggerWorkspaceEntity extends BaseWorkspaceEntity (not SyncableEntity) and has no flat entity map cache infrastructure yet.


charlesBochet:
@thomtrp some IA (but tested and reviewed!) gift

thomtrp (approved):

Weiko (approved):
Nice 👍 LGTM

PR #17827, opened 2026-02-10
fix: show user-friendly error message when duplicate invite is sent

### LINKED ISSUE #17822: No userFriendlyMessage when an invite has already been sent to that user
When inviting the same member twice it shows "an error occured" instead of showing a proper message

Image

This is because there is no user friendly message passed in the api response. Why? 

@claude please investigate and if you're sure you've found the issue then raise a PR directly

FelixMalfait:
Fixes #17822

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/workspace-invitation/services/workspace-invitation.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/workspace-invitation/services/workspace-invitation.service.ts
@@ -36,6 +36,7 @@ import {
 } from 'src/engine/core-modules/workspace-invitation/workspace-invitation.exception';
 import { WorkspaceEntity } from 'src/engine/core-modules/workspace/workspace.entity';
 import { WorkspaceMemberWorkspaceEntity } from 'src/modules/workspace-member/standard-objects/workspace-member.workspace-entity';
+import { CustomException } from 'src/utils/custom-exception';
 
 @Injectable()
 export class WorkspaceInvitationService {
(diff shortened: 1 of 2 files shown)


packages/twenty-server/src/engine/core-modules/workspace-invitation/services/workspace-invitation.service.ts
```diff
@@ -369,7 +372,13 @@ export class WorkspaceInvitationService {
     }>(
       (acc, invitation) => {
         if (invitation.status === 'rejected') {
-          acc.errors.push(invitation.reason?.message ?? 'Unknown error');
+          const reason = invitation.reason;
+
+          if (reason instanceof CustomException && reason.userFriendlyMessage) {
+            acc.errors.push(i18n._(reason.userFriendlyMessage));
```
charlesBochet, line 378: a bit non standard here, don't we usually throw exceptions with using t` or msg` and then the backend auto translates it?

I'm not clear why it's different here

FelixMalfait: Yes I didn't dig but my intuition was that it's because there are several invitations being sent (comma separated) = several errors = a catch with an aggregation

FelixMalfait: @claude can you please study the codebase and look if there would be anything smarter/cleaner to do? (more in-line with codebase practises). Maybe there isn't, post your findings.

FelixMalfait: I think we should merge like this

charlesBochet: ok, LGTM

charlesBochet (approved):

PR #17834, opened 2026-02-10
Fix blocknote.map crash with generic field-level RICH_TEXT_V2 handler

FelixMalfait:
## Summary

Fixes #17667

- **Root cause**: `ActivityQueryResultGetterHandler` called `JSON.parse()` on the `blocknote` field and assumed the result was always an array. When the stored value was valid JSON but not an array (e.g., `"{}"`), `blocknote.map()` crashed with `blocknote.map is not a function`, breaking the entire notes page.
- **Fix**: Replaced the object-level `ActivityQueryResultGetterHandler` (hardcoded for `note`/`task` only) with a generic field-level `RichTextV2FieldQueryResultGetterHandler` that safely parses blocknote JSON with `Array.isArray` validation and gracefully skips malformed values instead of crashing.
- **Bonus**: The new handler works for **all** objects with `RICH_TEXT_V2` fields (not just `note`/`task`), following the same pattern as the existing `FilesFieldQueryResultGetterHandler`.

## Changes

| File | Change |
|------|--------|
| `rich-text-v2-field-query-result-getter.handler.ts` | New field-level handler with safe blocknote parsing |
| `common-result-getters.service.ts` | Register new handler, remove `note`/`task` object handlers |
| `activity-query-result-getter.handler.ts` | Deleted (replaced by field-level handler) |
| `rich-text-v2-field-query-result-getter.handler.spec.ts` | 9 tests covering all edge cases |

## Test plan

- [x] Unit tests pass (9 tests covering: null blocknote, non-string blocknote, invalid JSON, non-array JSON like `"{}"`, no images, external URLs, internal URLs, multiple fields)
- [x] Lint passes (`lint:diff-with-main`)
- [x] Typecheck passes


packages/twenty-server/src/engine/api/common/common-result-getters/handlers/field-handlers/rich-text-v2-field-query-result-getter.handler.ts
```diff
@@ -0,0 +1,120 @@
...
+      ...block,
+      props: {
+        ...block.props,
+        url: `${process.env.SERVER_URL}/files/${signedPath}`,
+      },
+    };
+  });
+};
+
+const parseBlocknoteJsonSafely = (
```
FelixMalfait, line 58: Added this

packages/twenty-server/src/engine/api/common/common-result-getters/handlers/field-handlers/rich-text-v2-field-query-result-getter.handler.ts
```diff
@@ -0,0 +1,120 @@
...
+      return block;
+    }
+
+    const fileName = pathname.match(/files\/attachment\/(?:.+)\/(.+)$/)?.[1];
+
+    if (!isDefined(fileName)) {
+      return block;
+    }
+
+    const signedPath = fileService.signFileUrl({
```
FelixMalfait, line 43: As discussed with Etienne this will be refactored in the future

etiennejouan (approved):
LGTM !

PR #17837, opened 2026-02-10
SSO - Check if assertion is signed

thomtrp:
As title

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/auth/strategies/saml.auth.strategy.ts
+++ b/packages/twenty-server/src/engine/core-modules/auth/strategies/saml.auth.strategy.ts
@@ -47,14 +47,17 @@ export class SamlAuthStrategy extends PassportStrategy(
                 identityProvider &&
                 this.sSOService.isSAMLIdentityProvider(identityProvider)
               ) {
+                // IdP metadata XML typically has whitespace-formatted certificates
+                const sanitizedCertificate =
+                  identityProvider.certificate.replace(/\s/g, '');
+
                 const config: SamlConfig = {
                   entryPoint: identityProvider.ssoURL,
                   issuer: this.sSOService.buildIssuerURL(identityProvider),
                   callbackUrl:
                     this.sSOService.buildCallbackUrl(identityProvider),
-                  idpCert: identityProvider.certificate,
-                  // TODO: Improve the feature by sign the response
-                  wantAssertionsSigned: false,
+                  idpCert: sanitizedCertificate,
+                  wantAssertionsSigned: true,
                   wantAuthnResponseSigned: false,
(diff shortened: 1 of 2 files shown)


packages/twenty-server/src/engine/core-modules/auth/strategies/saml.auth.strategy.ts
```diff
@@ -47,14 +47,17 @@ export class SamlAuthStrategy extends PassportStrategy(
...
                   entryPoint: identityProvider.ssoURL,
                   issuer: this.sSOService.buildIssuerURL(identityProvider),
                   callbackUrl:
                     this.sSOService.buildCallbackUrl(identityProvider),
-                  idpCert: identityProvider.certificate,
-                  // TODO: Improve the feature by sign the response
-                  wantAssertionsSigned: false,
+                  idpCert: sanitizedCertificate,
+                  wantAssertionsSigned: true,
                   wantAuthnResponseSigned: false,
```
FelixMalfait, line 61: wantAuthnResponseSigned not relevant?

packages/twenty-server/src/engine/core-modules/auth/strategies/saml.auth.strategy.ts
```diff
@@ -47,14 +47,17 @@ export class SamlAuthStrategy extends PassportStrategy(
...
                 const config: SamlConfig = {
                   entryPoint: identityProvider.ssoURL,
                   issuer: this.sSOService.buildIssuerURL(identityProvider),
                   callbackUrl:
                     this.sSOService.buildCallbackUrl(identityProvider),
-                  idpCert: identityProvider.certificate,
-                  // TODO: Improve the feature by sign the response
-                  wantAssertionsSigned: false,
+                  idpCert: sanitizedCertificate,
+                  wantAssertionsSigned: true,
```
FelixMalfait, line 60: would it make sense to consider this at a connection level or not at all?

packages/twenty-server/src/engine/core-modules/auth/strategies/saml.auth.strategy.ts
thomtrp: it could be. The issue is while testing, I saw some idp that were not proposing signed responses. Seems less standard than assertions. So I thought assertions could be enough, at least as a first step. Wdyt?

FelixMalfait (approved):

packages/twenty-server/src/engine/core-modules/auth/strategies/saml.auth.strategy.ts
FelixMalfait: @thomtrp you please check https://sonarly.com/issue/36537?

PR #17843, opened 2026-02-10
Fix phone validation performance by using Set/Map instead of Array lookups

FelixMalfait:
## Summary
- **`isValidCountryCode`** was using `Array.includes()` on ~250 country codes (O(n) per call). Replaced with a `Set.has()` lookup (O(1)).
- **`getCountryCodesForCallingCode`** was iterating all ~250 countries and calling `getCountryCallingCode()` on each one **every invocation**. Replaced with a precomputed `Map<callingCode, CountryCode[]>` built once at module load (O(1) per call).

Both functions are called from `transformPhonesValue` on every phone field mutation, causing cumulative overhead visible in profiling (p95 self-time ~20-35ms).

## Test plan
- [ ] Verify phone field creation/update still works correctly (country code validation, calling code resolution)
- [ ] Verify spreadsheet import with phone fields still validates properly
- [ ] Confirm no regression in `isValidCountryCode` or `getCountryCodesForCallingCode` behavior

### DIFF
--- a/packages/twenty-shared/src/utils/validation/phones-value/getCountryCodesForCallingCode.ts
+++ b/packages/twenty-shared/src/utils/validation/phones-value/getCountryCodesForCallingCode.ts
@@ -1,15 +1,30 @@
-import { getCountries, getCountryCallingCode } from 'libphonenumber-js';
+import {
+  type CountryCode,
+  getCountries,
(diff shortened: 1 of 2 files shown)


Weiko (approved):
LGTM

PR #17845, opened 2026-02-10
Prevent leak between /metadata and /graphql GQL schemas

charlesBochet:
### Summary
- Patch `@nestjs/graphql` to support a `resolverSchemaScope` option that filters resolvers at both schema generation and runtime, preventing cross-endpoint leaking
- Introduce `@CoreResolver()` and `@MetadataResolver()` decorators to explicitly scope each resolver to its endpoint
- Move most resolvers (auth, billing, workspace, user, etc.) to the metadata schema where the frontend expects them; only workflow and timeline calendar/messaging resolvers remain on `/graphql`
- Fix frontend `SSEQuerySubscribeEffect` to use the default (metadata) Apollo client instead of the core client

### Problem
NestJS GraphQL's module-based resolver discovery traverses transitive imports, causing resolvers from `/metadata` modules to leak into the `/graphql` schema and vice versa. This made the schemas unpredictable and tightly coupled to module import order.

### Approach
- Added `resolverSchemaScope` to `GqlModuleOptions` via a patch on `@nestjs/graphql`, filtering in both `filterResolvers()` (runtime binding) and `getAllCtors()` (schema generation)
- Each resolver is explicitly decorated with `@CoreResolver()` or `@MetadataResolver()`
- Organized decorator, constant, and type files under `graphql-config/` following project conventions

Core GQL Schema: (see: no more fields!)
image

Metadata GQL Schema (see no more getTimelineCalendarEventsFromCompany)
image


Weiko (approved):
LGTM

FelixMalfait (approved):
Great, but a bit worried about the breaking change for end-users. + cc @etiennejouan we might want to ship this together with any file upload breaking change you do, so that at least we can communicate once

FelixMalfait:
Removing from queue because I just want to confirm @charlesBochet @Weiko - are we ok with the BC? Need a couple days notice before we deploy to prod then

PR #17850, opened 2026-02-11
Respect Gmail retry-after in messaging throttle

neo773:
Gmail 429/403 rate-limit responses include an explicit retry-after timestamp, usually ~15 minutes out. 

The exponential backoff starts at 1 minute, so the channel burns through all 5 retry attempts before the window actually closes and gets marked as permanently failed. 

Adds throttleRetryAfter to the message channel and uses max(backoff, retryAfter) in isThrottled().


packages/twenty-server/src/modules/messaging/message-import-manager/drivers/gmail/utils/parse-gmail-api-error.util.ts
```diff
@@ -57,6 +58,9 @@ export const parseGmailApiError = (
       return new MessageImportDriverException(
         gmailApiError.message,
         MessageImportDriverExceptionCode.TEMPORARY_ERROR,
+        {
+          throttleRetryAfter: parseGmailErrorRetryAfter(gmailApiError.message),
```
charlesBochet, line 62: naming should be improve: computeGmailErrorRetryAfter

packages/twenty-server/src/modules/messaging/message-import-manager/services/messaging-import-exception-handler.service.ts
```diff
@@ -166,6 +168,20 @@ export class MessageImportExceptionHandlerService {
         undefined,
         ['throttleFailureCount', 'id'],
       );
+
+      const throttleRetryAfter =
```
charlesBochet, line 172: not a fan here, it feels that we are using the exception to pass the information. Exception should not be used as a DTO (or it would be a first time). I can't really see a better way though. 
LGTM

charlesBochet (approved):

charlesBochet:
We are missing an upgrade command for existing workspaces

charlesBochet (approved):

PR #17856, opened 2026-02-11
Fix code step and logic function step in workflows

thomtrp:
- AI still often forgets to update the code step after creating it. Adding a next step 
- Starting by loading logic functions, so it avoids creating code steps when a function exists
- Fix create complete workflow logic. Should not create code steps directly

### DIFF
--- a/packages/twenty-server/src/modules/workflow/workflow-tools/tools/create-workflow-version-step.tool.ts
+++ b/packages/twenty-server/src/modules/workflow/workflow-tools/tools/create-workflow-version-step.tool.ts
@@ -10,13 +10,10 @@ import {
   type WorkflowToolDependencies,
 } from 'src/modules/workflow/workflow-tools/types/workflow-tool-dependencies.type';
 
-const createWorkflowVersionStepSchema = z.object({
+const baseStepFields = {
   workflowVersionId: z
     .string()
     .describe('The ID of the workflow version to add the step to'),
-  stepType: z
-    .enum(Object.values(WorkflowActionType) as [string, ...string[]])
-    .describe('The type of step to create'),
   parentStepId: z
     .string()
     .optional()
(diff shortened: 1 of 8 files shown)


packages/twenty-server/src/modules/workflow/workflow-tools/tools/create-workflow-version-step.tool.ts
```diff
@@ -40,8 +41,33 @@ const createWorkflowVersionStepSchema = z.object({
...
+    ),
 });
 
+const enrichResultWithNextStep = ({
+  result,
+  stepType,
+}: {
+  result: WorkflowVersionStepChangesDTO;
+  stepType: WorkflowActionType;
+}) => {
```
Weiko, line 81: missing nextStep in typing?

packages/twenty-server/src/modules/workflow/workflow-tools/tools/create-workflow-version-step.tool.ts
```diff
@@ -40,8 +41,33 @@ const createWorkflowVersionStepSchema = z.object({
     })
     .optional()
     .describe('Optional position coordinates for the step'),
+  defaultSettings: z
+    .record(z.string(), z.any())
+    .optional()
+    .describe(
+      'Optional default settings for the step. Required for LOGIC_FUNCTION steps: pass { input: { logicFunctionId: "<id>" } }. Use list_logic_function_tools to discover available logic function IDs.',
```
Weiko: nit: can we enforce that with zod?

Weiko (approved):
Left some comments @thomtrp

packages/twenty-server/src/modules/workflow/workflow-tools/tools/create-workflow-version-step.tool.ts
thomtrp: nextStep is only in the output. And there is no output type defined

Weiko: Oops read too quickly!

PR #17857, opened 2026-02-11
Speed up twenty-emails build by replacing vite-plugin-dts with tsgo

FelixMalfait:
## Summary

- Removed `vite-plugin-dts` (which used `tsc` internally) from the Vite build and replaced DTS generation with `tsgo` as a sequential post-build step — **~0.7s vs 1-10s**.
- Disabled `reportCompressedSize` to skip gzip computation for 64 output files.
- Converted the build target to an explicit `nx:run-commands` executor with sequential `vite build` → `tsgo` commands.

The `twenty-emails:build` step goes from ~22s to ~7s under load. 

## Test plan

- [x] `nx build twenty-emails` produces both JS (64 files) and DTS (74 files) correctly
- [x] `dist/index.d.ts` exports match the source `src/index.ts`
- [x] Full `nx build twenty-server` succeeds end-to-end
- [ ] CI build passes


FelixMalfait:
## Performance measurements

### Per-package build times (vite build + DTS generation)

| Package | Before (vite-plugin-dts) | After (vite + tsgo) | Speedup |
|---------|-------------------------|---------------------|---------|
| twenty-emails | ~22s | ~10s | **55%** |
| twenty-shared | 7.08s | 3.60s | **49%** |
| twenty-sdk | 30.18s | 21.66s | **28%** |

### Full pipeline: `nx build twenty-server` (cold cache)

| | Time |
|---|---|
| Before | ~128s |
| After | **~55s** |

Note: twenty-sdk is not in the twenty-server dependency chain, so its improvement applies to SDK-dependent builds separately.

### What changed
- Removed `vite-plugin-dts` from all three packages' Vite configs
- DTS generation now handled by `tsgo` (native Go binary) as a separate sequential step
- Added `--noEmit false` flag to override inherited `noEmit: true` from parent tsconfigs
- Set `reportCompressedSize: false` in twenty-emails

packages/twenty-emails/project.json
```diff
@@ -5,9 +5,17 @@
...
+      "executor": "nx:run-commands",
+      "cache": true,
+      "inputs": ["production", "^production"],
+      "outputs": ["{projectRoot}/dist"],
       "options": {
-        "outputPath": "{projectRoot}/dist"
+        "cwd": "{projectRoot}",
+        "commands": [
+          "npx vite build",
+          "tsgo -p tsconfig.lib.json --declaration --emitDeclarationOnly --outDir dist --rootDir src --composite false && npx tsc-alias -p tsconfig.lib.json --outDir dist"
```
prastoin, line 16: ***Remark:*** I feel like all these flags should be defaulted in the `tsconfig.lib.json` ( except the composite )
```
--declaration --emitDeclarationOnly --outDir dist --rootDir src --composite false
```

prastoin (approved):
Way faster ! we should check if a vite tsgo based plugin is released at some point

packages/twenty-emails/project.json
FelixMalfait: The tsconfig.lib.json currently serves double duty (Nx typecheck with outDir: ../../.cache/tsc + composite: true, vs. DTS generation needing outDir: dist + composite: false), so most of these flags genuinely need to differ between the two use cases. We could create a separate tsconfig.dts.json per package, but for 3 packages I think the CLI flags are a reasonable trade-off to avoid the extra files (for now! we can revisit)

PR #17858, opened 2026-02-11
Fix spurious logouts by deduplicating concurrent token renewals

FelixMalfait:
## Summary

- When returning to the app after idle (or after a deploy), the expired access token causes multiple simultaneous GraphQL queries to fail with `UNAUTHENTICATED`. Previously, each failure independently triggered its own `renewToken` call with the same refresh token. If **any single** renewal failed (e.g. server briefly slow after a deploy), the `catch` handler would nuke the session and redirect to sign-in — even if another concurrent renewal had already succeeded and written valid tokens.
- This adds a shared `renewalPromise` so that only the first `UNAUTHENTICATED` error triggers a server-side renewal. All concurrent callers await the same promise and replay their operations once it resolves. This eliminates redundant refresh token rotation on the server and removes the race condition where a straggling failure could log out an already-renewed session.

## Test plan

- [ ] Log in, wait >30 minutes (or manually expire the access token), then interact with the app — should silently renew without redirect to sign-in
- [ ] Open browser DevTools Network tab, trigger the above scenario, and verify only **one** `renewToken` mutation is sent (instead of N)
- [ ] With server temporarily stopped, verify that a genuine renewal failure still correctly redirects to sign-in (single `onUnauthenticatedError` call)
- [ ] Open multiple browser tabs, let access tokens expire, interact in one tab — other tabs should also recover gracefully on their next request

### DIFF
--- a/packages/twenty-front/src/modules/apollo/services/apollo.factory.ts
+++ b/packages/twenty-front/src/modules/apollo/services/apollo.factory.ts
@@ -22,6 +22,7 @@ import { type AuthTokenPair } from '~/generated/graphql';
 import { logDebug } from '~/utils/logDebug';
 
 import { REST_API_BASE_URL } from '@/apollo/constant/rest-api-base-url';
+import { REACT_APP_SERVER_BASE_URL } from '~/config';
 import { getTokenPair } from '@/apollo/utils/getTokenPair';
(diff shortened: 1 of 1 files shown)


Weiko (approved):
LGTM

charlesBochet (approved):

PR #17855, opened 2026-02-11
[FRONT COMPONENT] Add Front component token generation

ehconitin:
closes https://github.com/twentyhq/core-team-issues/issues/2180

https://github.com/user-attachments/assets/a898455d-eb1c-4d22-b585-785e98fc38a7

PR #17860, opened 2026-02-11
Generate GQL schema based on applicationId

charlesBochet:
## Add application-scoped GraphQL schema generation

When an application token is used to authenticate, the `/graphql` schema is now dynamically filtered to only include entities belonging to that application (plus the Twenty Standard Application). This enables third-party applications and the SDK to introspect a schema that is relevant to their scope, rather than seeing the full workspace schema with all custom objects.

### Changes

- **New `generateApplicationToken` mutation** on the `/metadata` endpoint, allowing callers to exchange an API key for an application-scoped JWT token
- **Schema filtering by application** in `WorkspaceSchemaFactory` — when `request.application` is present (from an application token), flat entity maps are filtered by `[appId, standardAppId]` before schema generation
- **Per-app caching** — both the Yoga in-memory cache and Redis cache now include the `appId` in their keys to avoid serving wrong schemas
- **Consolidated `getSubFlatEntityMapsByApplicationIdsOrThrow`** — unified the single-ID and multi-ID filtering utilities into one
- **Integration tests** covering token generation (admin + API key auth) and schema introspection filtering (standard app token excludes custom objects)

Schema generated on seeds with applicationToken (see that pets is missing)
image


packages/twenty-server/patches/@graphql-yoga+nestjs+2.1.0.patch
```diff
@@ -34,8 +34,9 @@ index 1684394..dd0773d 100644
 +                const workspaceId = request.req.workspace?.id ?? 'anonymous'
 +                const workspaceCacheVersion = request.req.workspaceMetadataVersion ?? '0'
 +                const url = request.req.baseUrl
++                const applicationId = request.req.application?.id ?? 'all'
 +
-+                const cacheKey = `${workspaceId}-${url}-${workspaceCacheVersion}`
++                const cacheKey = `${workspaceId}-${url}-${workspaceCacheVersion}-${applicationId}`
```
Weiko, line 39: There is no cache size limit nor TTL afaik, this could lead to memory leak

packages/twenty-server/src/engine/api/graphql/workspace-schema.factory.ts
```diff
@@ -111,11 +156,13 @@ export class WorkspaceSchemaFactory {
         workspace.id,
         metadataVersion,
         typeDefs,
+        applicationId,
       );
       await this.workspaceCacheStorageService.setGraphQLUsedScalarNames(
```
Weiko, line 161: I'm wondering if we could use mset now? 🤔 Not super important!

Weiko: Actually mset is not using pipeline internally so it's not super useful here (it parallelizes the network calls though)

packages/twenty-server/patches/@graphql-yoga+nestjs+2.1.0.patch
```diff
@@ -86,8 +86,9 @@ index 1684394..dd0773d 100644
...
 +
 +                if(this.schemaCache.has(cacheKey)) {
```
Weiko, line 93: Could use promizeMemoization pattern at some point

packages/twenty-server/src/engine/core-modules/application/resolvers/application.resolver.ts
```diff
@@ -96,6 +104,33 @@ export class ApplicationResolver {
...
+      nodeEnv !== NodeEnvironment.DEVELOPMENT &&
+      nodeEnv !== NodeEnvironment.TEST
+    ) {
+      throw new ApplicationException(
+        'This endpoint is only available in development mode',
+        ApplicationExceptionCode.FORBIDDEN,
+      );
+    }
+
+    const APPLICATION_TOKEN_EXPIRY_SECONDS = 30 * 24 * 60 * 60;
```
martmull, line 125: 30days is pretty big to me, why such a big number?

packages/twenty-server/src/engine/api/graphql/workspace-schema.factory.ts
```diff
@@ -42,32 +49,68 @@ export class WorkspaceSchemaFactory {
...
+    let flatFieldMetadataMaps = allFlatFieldMetadataMaps;
+    let flatIndexMaps = allFlatIndexMaps;
+
+    if (isDefined(applicationId)) {
+      const twentyStandardApplicationId =
+        flatApplicationMaps?.idByUniversalIdentifier[
+          TWENTY_STANDARD_APPLICATION.universalIdentifier
+        ];
+
+      const applicationIds = isDefined(twentyStandardApplicationId)
```
Weiko, line 93: Shouldn't we throw if standard app is not there?

packages/twenty-server/src/engine/core-modules/application/resolvers/application.resolver.ts
```diff
@@ -96,6 +100,21 @@ export class ApplicationResolver {
     });
   }
 
+  @Mutation(() => AuthToken)
+  @RequireFeatureFlag(FeatureFlagKey.IS_APPLICATION_ENABLED)
+  async generateApplicationToken(
+    @Args() { applicationId }: GenerateApplicationTokenInput,
+    @AuthWorkspace() { id: workspaceId }: WorkspaceEntity,
+  ): Promise<AuthToken> {
+    const APPLICATION_TOKEN_EXPIRY_SECONDS = 30 * 24 * 60 * 60; // 30 days
```
Weiko: Should we use ENV variable?

Weiko (commented):
Left some comments

packages/twenty-server/patches/@graphql-yoga+nestjs+2.1.0.patch
charlesBochet: that's not new :p this is likely our memory leak

packages/twenty-server/src/engine/core-modules/application/resolvers/application.resolver.ts
charlesBochet: since this is dev endpoint, i wold say not

martmull (approved):

Weiko (approved):
I will address those comments myself. LGTM!

packages/twenty-server/src/engine/core-modules/application/resolvers/application.resolver.ts
charlesBochet: this endpoint is only activate in NODE_ENV=development, no need to have short life token IMO

charlesBochet: I think all the endpoints for SDK should only be activated in development mode and not in prod

ehconitin: I think Martin is taking this context into account
 https://github.com/twentyhq/twenty/pull/17855

Where we would need a token for queries through front components

PR #17866, opened 2026-02-11
[FRONT COMPONENTS] Twenty UI elements generation

### LINKED ISSUE #2175: [FRONT COMPONENTS] Update remote-dom component generation script for twenty-ui
Currently, we generate components and elements only for standard html elements.

We want to do this for every component in twenty-ui

bosiraphael:
## PR description

This PR:
- Creates a TypeScript-based extractor that discovers all exported twenty-ui components by scanning barrel files, extracting props/slots/events via ts-morph type analysis, and generating the remote DOM bindings automatically.

- Adds a new ESLint rule which enforces all *Props types in twenty-ui components to be exported, which is required for the extractor to discover component prop types. Existing twenty-ui components are updated to comply with this rule.

- Extends the remote DOM generation to support slots, per-component events, forwardRef wrappers, and richer property types (array, object, function)

## Edge cases to fix in another PR

- Icons cannot be rendered inside buttons
- IconButtons throw an error when mounted
- MenuItems are not displayed correctly
- MenuItemNavigate throws on click

## Video Demo

https://github.com/user-attachments/assets/c2ed67cf-6a15-4896-9fec-e83fac0e862b


bosiraphael:
@greptileai

bosiraphael:
@cubic-dev-ai please review

charlesBochet:
@bosiraphael @ehconitin Changing my mind (sorry!) on this one after a discussion with Felix, I actually think that we should treat twenty ui as another library. It should be present in the sdk but usage should be taken in the built remote component as any other library. This means they should not be any work to do on host side or to prepare extraction anymore :(

charlesBochet:
I'm going to close it for now, let's discuss it tomorrow morning

charlesBochet:
Merging it but likelty to ve revisited

PR #17870, opened 2026-02-11
Handle 413 with user friendly message

etiennejouan:
Add friendly message for 413 errors. Currently, 413 errors originate from the nginx server.


FelixMalfait (approved):
Great!

PR #17871, opened 2026-02-11
Fix page layout seeding for record page layouts

Weiko:
## Context
Fix broken record page layout seeding. This was not detected by the CI because it doesn't have the env variable yet.

Following the same mechanism as labelIdentifier in object for circular dependency resolution


Devessier (approved):
Great! Thanks.

PR #17875, opened 2026-02-11
File v2 - Backfill mimeType and size - command

etiennejouan:


packages/twenty-server/src/database/commands/upgrade-version-command/1-18/1-18-backfill-file-size-and-mime-type.command.ts
```diff
@@ -0,0 +1,176 @@
...
+          size?: number;
+          mimeType?: string;
+        } = {};
+
+        if (fileEntity.size < 0) {
+          updateData.size = fileBuffer.length;
+        }
+
+        if (fileEntity.mimeType === 'application/octet-stream') {
+          const { mimeType } = await extractFileInfo({
```
prastoin: ***Remark:*** There's a world where we could get such exception here
```ts
      throw new FileStorageException(
        `File content does not match its extension. The file has extension '${ext}' (expected mime type: ${mimeTypeFromExtension}), but the file content could not be detected as this type. The file may be corrupted, have the wrong extension, or be a security risk.`,
        FileStorageExceptionCode.INVALID_EXTENSION,
        {
          userFriendlyMessage: msg`The file extension doesn't match the file content. Please check that your file is not corrupted and has the correct extension.`,
        },
      );
```

prastoin (approved):
Looks solid to me !
Greptile left a noticeable comment

PR #17876, opened 2026-02-11
Improve AI agent chat, tool display, and workflow agent management

FelixMalfait:
## Summary

- **Fix token renewal endpoint**: Use `/metadata` instead of `/graphql` for token renewal in agent chat, fixing auth issues
- **Improve tool display**: Add `load_skills` support, show formatted tool names (underscores → spaces) with finish/loading states, display tool icons during loading, and support custom loading messages from tool input
- **Refactor workflow agent management**: Replace direct `AgentRepository` access with `AgentService` for create/delete/find operations in workflow steps, improving encapsulation and consistency
- **Simplify Apollo client usage**: Remove explicit Apollo client override in `useGetToolIndex`, add `AgentChatProvider` to `AppRouterProviders`
- **Fix load-skill tool**: Change parameter type from `string` to `json` for proper schema parsing
- **Update agent-chat-streaming**: Use `AgentService` for agent resolution and tool registration instead of direct repository queries

## Test plan

- [ ] Verify AI agent chat works end-to-end (send message, receive response)
- [ ] Verify tool steps display correctly with icons and proper messages during loading and after completion
- [ ] Verify workflow AI agent step creation and deletion works correctly
- [ ] Verify workflow version cloning preserves agent configuration
- [ ] Verify token renewal works when tokens expire during agent chat


packages/twenty-front/src/modules/ai/utils/getToolDisplayMessage.ts
```diff
@@ -95,5 +112,21 @@ export const getToolDisplayMessage = (
...
+  }
+
+  const customMessage = extractCustomLoadingMessage(resolvedInput);
+
+  if (isDefined(customMessage)) {
+    return customMessage;
+  }
+
+  const formattedName = formatToolName(resolvedToolName);
+  const action = isFinished ? 'Ran' : 'Running';
```
FelixMalfait: All these strings should be translated with Lingui (either msg or t)

packages/twenty-server/src/engine/metadata-modules/ai/ai-chat/services/agent-chat-streaming.service.ts
```diff
@@ -65,6 +65,19 @@ export class AgentChatStreamingService {
       );
     }
 
+    // Save user message eagerly so it persists even if the stream is aborted
+    const lastUserText =
+      messages[messages.length - 1]?.parts.find((part) => part.type === 'text')
+        ?.text ?? '';
+
+    const userMessage = await this.agentChatService.addMessage({
```
FelixMalfait: Do we need the await? We probably do but I just want to check since anything that slows down streaming/time to first letter is bad

packages/twenty-server/src/engine/metadata-modules/ai/ai-chat/services/agent-chat-streaming.service.ts
```diff
@@ -174,54 +187,39 @@ export class AgentChatStreamingService {
...
-                    totalOutputTokens: () =>
-                      `"totalOutputTokens" + ${streamUsage.outputTokens}`,
-                    totalInputCredits: () =>
-                      `"totalInputCredits" + ${streamUsage.inputCredits}`,
-                    totalOutputCredits: () =>
-                      `"totalOutputCredits" + ${streamUsage.outputCredits}`,
-                    contextWindowTokens: modelConfig.contextWindowTokens,
-                    conversationSize: lastStepConversationSize,
-                  });
+                  // Skip usage update on abort since streamUsage may be incomplete
```
FelixMalfait: Not sure we need to skip. Ideally we would close the session cleanly with the SDK provider and retrieve the right cost/credits etc? We don't want to lose money because of interrupted requests to LLM providers not even being lost (or we don't want the billing event to be sent but not visible in thread metrics)

packages/twenty-front/src/modules/ai/utils/getToolDisplayMessage.ts
```diff
@@ -74,6 +74,23 @@ const extractLearnToolNames = (input: ToolInput): string => {
   return '';
 };
 
+const extractSkillNames = (input: ToolInput): string => {
```
FelixMalfait, line 69: This is ugly but how it's done in most places in the AI chat today since everything is typed at runtime. If we were to refactor all similar code in an upcoming PR would there be a more elegant way?

FelixMalfait:
Re: extractSkillNames refactoring question — Yes, we could define a Zod schema per tool name and use `.safeParse()` to validate at runtime. Something like:

```ts
const LoadSkillsInput = z.object({ skillNames: z.array(z.string()) });
const LearnToolsInput = z.object({ toolNames: z.array(z.string()) });
const WebSearchInput = z.union([
  z.object({ query: z.string() }),
  z.object({ action: z.object({ query: z.string() }) }),
]);

const TOOL_INPUT_SCHEMAS = {
  load_skills: LoadSkillsInput,
  learn_tools: LearnToolsInput,
  web_search: WebSearchInput,
} as const;
```

Then each extractor becomes a one-liner — `TOOL_INPUT_SCHEMAS[toolName].safeParse(input)` — and we get proper typing on the result. Would clean up all the nested `isDefined && typeof === 'object' && 'key' in input` chains. Happy to tackle in a follow-up PR.

packages/twenty-front/src/modules/ai/utils/getToolDisplayMessage.ts
```diff
@@ -83,17 +85,43 @@ export const getToolDisplayMessage = (
...
+    return isFinished ? t`Searched the web` : t`Searching the web`;
   }
 
   if (resolvedToolName === 'learn_tools') {
     const names = extractLearnToolNames(resolvedInput);
-    const action = isFinished ? 'Learned' : 'Learning';
 
-    return names ? `${action} ${names}` : `${action} tools...`;
+    if (isNonEmptyString(names)) {
+      return isFinished ? t`Learned ${names}` : t`Learning ${names}`;
```
charlesBochet: nitpick a lot of ternary is Finished, wonder if we could wrote this more elegantly

packages/twenty-front/src/modules/ai/schemas/toolOutputMessageSchema.ts
```diff
@@ -0,0 +1,3 @@
+import { z } from 'zod';
+
+export const ToolOutputMessageSchema = z.object({ message: z.string() });
```
charlesBochet, line 3: we usually put those in a schemas folder

packages/twenty-server/src/engine/core-modules/tool-provider/services/tool-registry.service.ts
```diff
@@ -435,6 +443,32 @@ export class ToolRegistryService {
...
+      return;
+    }
+
+    const sortedEntries = [...this.ramCache.entries()].sort(
+      ([, entryA], [, entryB]) => entryA.cachedAt - entryB.cachedAt,
+    );
+
+    const toEvict = sortedEntries.slice(0, MIN_EVICT_ENTRIES);
+
+    for (const [key] of toEvict) {
```
charlesBochet: file is veryyyyyy long

charlesBochet (approved):
LGTM codewise, left a few comments but non blocking

packages/twenty-server/src/engine/core-modules/tool-provider/services/tool-registry.service.ts
FelixMalfait: yes good point, will split it

PR #17886, opened 2026-02-12
Workflow bug fixes 

### LINKED ISSUE #17611: Workflow : Adding an If/Else node inside an Iterator breaks the flow
## Bug Description

If you have an iterator in a workflow, it will have loop, return and completed branches. 

Image

If you then add an If/Else node to the loop branch, it will remove the return and completed branches, so there are no longer the necessary branches to allow the iterator to complete. 

Image

Notice two of the earlier branches are now gone. 

I have tried manually add them back, but obviously not successfully as my loop never completes one round. 

As a result, the Workflow Run remains Runnng forever. If I stop it, it moves to Stopping but never actually completes

This is on paid/hosted twentycrm, so the latest version I assume.

thomtrp:
Fixes https://github.com/twentyhq/private-issues/issues/418 
Currency filtering was not properly managed. Since this is a select, we were doing 'USD' == [USD]. Replacing by contains as for other select
Capture d’écran 2026-02-12 à 11 08 42

Fixes https://github.com/twentyhq/twenty/issues/17611
If-else branches not executed has to be marked as skipped. Otherwise, the iterator will never start the next iteration. It will wait for some not started nodes.
Capture d’écran 2026-02-12 à 10 56 45

### DIFF
--- a/packages/twenty-server/src/modules/workflow/workflow-executor/workspace-services/workflow-executor.workspace-service.ts
+++ b/packages/twenty-server/src/modules/workflow/workflow-executor/workspace-services/workflow-executor.workspace-service.ts
@@ -169,10 +169,21 @@ export class WorkflowExecutorWorkspaceService {
       return;
     }
 
-    const nextStepIdsToExecute = await this.getNextStepIdsToExecute({
-      executedStep: stepToExecute,
-      executedStepOutput: actionOutput,
-    });
+    const { nextStepIdsToExecute, nextStepIdsToSkip } =
+      await this.getNextStepIdsToExecute({
+        executedStep: stepToExecute,
+        executedStepOutput: actionOutput,
+      });
+
+    if (isDefined(nextStepIdsToSkip) && nextStepIdsToSkip.length > 0) {
+      await this.skipStepsAndContinue({
+        stepIdsToSkip: nextStepIdsToSkip,
+        steps,
+        workflowRunId,
+        workspaceId,
+        executedStepsCount: (executedStepsCount ?? 0) + 1,
+      });
+    }
 
     if (isDefined(nextStepIdsToExecute) && nextStepIdsToExecute.length > 0) {
       await this.executeFromSteps({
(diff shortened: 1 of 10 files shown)


packages/twenty-server/src/modules/workflow/workflow-executor/workspace-services/workflow-executor.workspace-service.ts
```diff
@@ -415,6 +444,44 @@ export class WorkflowExecutorWorkspaceService {
...
+      stepIdsToSkip.map(async (stepId) => {
+        await this.workflowRunWorkspaceService.updateWorkflowRunStepInfo({
+          stepId,
+          stepInfo: { status: StepStatus.SKIPPED },
+          workflowRunId,
+          workspaceId,
+        });
+
+        const skippedStep = steps.find((step) => step.id === stepId);
+        const skippedStepNextStepIds = skippedStep?.nextStepIds ?? [];
```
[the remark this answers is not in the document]
thomtrp: this is already what is done. Because we mark the first step as skipped, then we continue the execution. And when we arrive to a step having its previous steps marked as skipped, it is marked as skipped and continue the execution

packages/twenty-server/src/modules/workflow/workflow-executor/workspace-services/workflow-executor.workspace-service.ts
```diff
@@ -169,10 +169,21 @@ export class WorkflowExecutorWorkspaceService {
       return;
     }
 
-    const nextStepIdsToExecute = await this.getNextStepIdsToExecute({
-      executedStep: stepToExecute,
-      executedStepOutput: actionOutput,
-    });
+    const { nextStepIdsToExecute, nextStepIdsToSkip } =
```
martmull, line 172: ah la la les bons souvenirs 🤣

packages/twenty-server/src/modules/workflow/workflow-executor/workspace-services/workflow-executor.workspace-service.ts
```diff
@@ -191,7 +202,10 @@ export class WorkflowExecutorWorkspaceService {
   }: {
     executedStep: WorkflowAction;
     executedStepOutput: WorkflowActionOutput;
-  }): Promise<string[] | undefined> {
+  }): Promise<{
+    nextStepIdsToExecute?: string[];
+    nextStepIdsToSkip?: string[];
+  }> {
```
martmull, line 208: this need to be tested `getNextStepIdsToExecute`: i have not got on it since 2 or 3 months and basically forgot how it works, tests will document that pretty nicely i guess

martmull (approved):

thomtrp:
@greptile-ai

PR #17888, opened 2026-02-12
[Apps] Content + permission tabs for marketplace and installed apps

ijreilly:
In this PR
- Content tab for marketplace apps and installed apps: complies with [figma](https://www.figma.com/design/xt8O9mFeLl46C5InWwoMrN/Twenty?node-id=34845-126856); addition of field section showing fields added to standard objects; initiative: arrow opens a sub-table of fields rows. (suggested because 1/ for marketplace apps we cannot redirect to the actual object page in settings since the object does not exist yet in the workspace 2/ since we dont have an quick "go back to application page" option, it can be annoying to be redirected to a different setting page when we just want to look at the content of the app objects)
- Permission tab for marketplace apps and installed apps
- left to do - settings tab (in another PR)

There are breaking changes but it's behind a feature flag not exposed (access to marketplace) so not problematic

marketplace apps
https://github.com/user-attachments/assets/4c660101-50fc-47ce-b90a-8d6f17db5e74

installed apps
https://github.com/user-attachments/assets/c9229ee1-e75f-4cad-8766-758b2c5b37b4


packages/twenty-front/src/modules/marketplace/graphql/queries/findManyMarketplaceApps.ts
```diff
@@ -16,6 +16,65 @@ export const FIND_MANY_MARKETPLACE_APPS = gql`
       providers
       websiteUrl
       termsUrl
+      objects {
```
martmull: can't this can be extracted from `APPLICATION_FRAGMENT`?

{
...marketplaceRelatedFields
...ApplicationFields
}

packages/twenty-front/src/pages/settings/applications/components/SettingsApplicationDataTable.tsx
```diff
@@ -0,0 +1,251 @@
...
+  if (objectRows.length === 0 && fieldGroupRows.length === 0) {
+    return null;
+  }
+
+  const renderRow = (row: ApplicationDataTableRow) => {
+    const Icon = getIcon(row.icon);
+    const isExpanded = expandedRows.has(row.key);
+    const hasFields = isDefined(row.fields) && row.fields.length > 0;
+
+    return (
```
martmull: can we move that in a dedicated component so it reduces the size of this component?

packages/twenty-server/src/engine/core-modules/application/dtos/marketplace-app.dto.ts
```diff
@@ -1,13 +1,202 @@
...
+  @Field(() => [MarketplaceAppRoleFieldPermissionDTO])
+  fieldPermissions: MarketplaceAppRoleFieldPermissionDTO[];
+
+  @IsArray()
+  @Field(() => [String])
+  permissionFlags: string[];
+}
+
 @ObjectType('MarketplaceApp')
 export class MarketplaceAppDTO {
```
martmull, line 201: i guess you might extend a lot from ApplicationDTO isn't it?

ijreilly: I wouldn't say extend as there ApplicationDTO stuff that marketplaceApplicationDTO don't have: yarnlock and packageJson files and checksums, etc.
Their objects and fields also are different; marketplaceApplicationDTO don't have ids. 
I think it's safe to keep two different types

packages/twenty-front/src/pages/settings/applications/components/SettingsApplicationDataTableRow.tsx
```diff
@@ -0,0 +1,117 @@
...
+  onToggle: () => void;
+}) => {
+  const theme = useTheme();
+  const { getIcon } = useIcons();
+
+  const Icon = getIcon(row.icon);
+  const hasFields = isDefined(row.fields) && row.fields.length > 0;
+
+  return (
+    <div>
```
martmull: why do you need div here? If you absolutelly need it, you need to use an empty <StyledDiv />

ijreilly: needed because there is another element at the same level as TableRow

martmull: in that case you might just need <> </>

martmull (approved):

PR #17893, opened 2026-02-12
Start Jotai Migration

charlesBochet:
### Benchmark

In the beginning, there was no hope:
image

Then the hope was reborn
image

### Approach

We introduce a **V2 state management layer** backed by Jotai that mirrors the existing Recoil API, enabling component-by-component migration without a big-bang rewrite.

#### V2 API (Jotai-backed, Recoil-ergonomic)

- `createStateV2` / `createFamilyStateV2` — drop-in replacements for `createState` / `createFamilyState`, returning wrapper types over Jotai atoms
- `useRecoilValueV2`, `useRecoilStateV2`, `useFamilyRecoilValueV2`, etc. — thin wrappers around Jotai's `useAtomValue` / `useAtom` / `useSetAtom`
- A shared `jotaiStore` (via `createStore()`) passed to a `<JotaiProvider>` wrapping `<RecoilRoot>`, also accessible imperatively for dual-writes

#### Dual-write bridge for progressive migration

For state shared between migrated and non-migrated components, we use **dual-write**: writers update both the Recoil atom and the Jotai V2 atom (via `jotaiStore.set()`). This avoids sync components or extra subscriptions.

Write sites updated: `useUpsertRecordsInStore`, `useSetRecordTableData`, `ListenRecordUpdatesEffect`, `RecordShowEffect`, `useLoadRecordIndexStates`, `useUpdateObjectViewOptions`.

#### First migration: ChipFieldDisplay render path

- `useChipFieldDisplay` → reads `recordStoreFamilyStateV2` via `useFamilyRecoilValueV2` (was `useRecoilValue(recordStoreFamilyState)`)
- `RecordChip` → reads `recordIndexOpenRecordInStateV2` via `useRecoilValueV2` (was `useRecoilValue(recordIndexOpenRecordInState)`)
- `Avatar` (twenty-ui) and event handlers (`useOpenRecordInCommandMenu`) left on Recoil — not on the render path / in a different package

#### Pattern for migrating additional state

1. Create V2 atom: `createStateV2` or `createFamilyStateV2`
2. Add `jotaiStore.set(v2Atom, value)` at each write site
3. Switch readers to `useRecoilValueV2(v2Atom)`
4. Once all readers are migrated, remove the Recoil atom and dual-writes

#### Why not jotai-recoil-adapter?

Evaluated [jotai-recoil-adapter](https://github.com/clockelliptic/jotai-recoil-adapter) — not production-ready (21 open issues, no React 19, forces providerless mode, missing types). We built a purpose-built thin layer instead.


thomtrp (approved):
🔥

packages/twenty-front/src/modules/object-record/record-store/hooks/useUpsertRecordsInStore.ts
```diff
@@ -29,11 +31,16 @@ export const useUpsertRecordsInStore = () => {
...
               id: partialRecord.id,
               __typename: partialRecord.__typename,
               ...filteredPartialRecord,
-            });
+            };
+            set(recordStoreFamilyState(partialRecord.id), newRecord);
+            jotaiStore.set(
+              recordStoreFamilyStateV2.atomFamily(partialRecord.id),
+              newRecord,
+            );
```
Devessier, line 43: If you use `jotaiStore.set()`, does it mean that there is no transaction mechanism in Jotai? Contrary to `set`, `snapshot` and `reset` from `useRecoilCallback`

charlesBochet: That's a good question, I'm unsure how we replace recoilcallback yet

charlesBochet:
Before:

image

charlesBochet:
After, no significative improvement a priori
image

packages/twenty-front/src/modules/object-record/record-store/hooks/useUpsertRecordsInStore.ts
charlesBochet: actually, you can just query jotaiStore.get()

If you want to do transaciton, you can do that with a derived atom, we can leave this question for a bit later but should not be blocking

PR #17896, opened 2026-02-12
Fix index field comparison

prastoin:
# Introduction
Related https://github.com/twentyhq/core-team-issues/issues/2227

On a field name update side effect leading to an index field mutation it wouldn't get caught by the builder leading to an index field desync

We should land on a standard pattern regarding the field index either jsonb or syncableEntity so this would not occur anymore as it would have been strictly typed


Weiko (approved):
LGTM

etiennejouan (approved):
👍

PR #17897, opened 2026-02-12
Prefil fields widgets to standard app

Weiko:
## Context
Prefill the FIELDS widget configuration in standard page layouts during workspace creation, linking each widget to a dedicated view with positioned fields organized into sections (via view field groups)

We wanted something very declarative (by manually setting position and visibility of each field per standard object). 
In this PR I've generated all the compute- utils via AI (😨) for position/visibility, we'll probably want to confirm with the product which ordering/visibility we want for each standard object but I feel like this can be merged as it is since it's behind a feature flag and this will unblock the work on the frontend


packages/twenty-server/src/engine/metadata-modules/page-layout-widget/dtos/fields-configuration.dto.ts
```diff
@@ -13,4 +13,9 @@ export class FieldsConfigurationDTO
   @IsIn([WidgetConfigurationType.FIELDS])
   @IsNotEmpty()
   configurationType: WidgetConfigurationType.FIELDS;
+
+  @Field(() => String, { nullable: true })
+  @IsOptional()
+  @IsUUID()
+  viewId: string | null;
```
Devessier, line 20: Why is it optional?

If the user wants to create a new Fields widget, we first create the widget in the database, and then create and link a view? I would rather see the opposite pattern.

packages/twenty-server/src/engine/workspace-manager/twenty-standard-application/utils/__tests__/__snapshots__/get-standard-object-metadata-related-entity-ids.util.spec.ts.snap
```diff
@@ -375,10 +376,11 @@ exports[`getStandardObjectMetadataRelatedEntityIds should return standard object
         "id": "00000000-0000-0000-0000-000000000102",
       },
     },
-    "id": "00000000-0000-0000-0000-000000000126",
```
charlesBochet, line 378: doesn't look stable but OK :) (I think @prastoin introduced some example)

charlesBochet (approved):

packages/twenty-server/src/engine/metadata-modules/page-layout-widget/dtos/fields-configuration.dto.ts
Weiko: Yes it's probably the opposite indeed, I don't see a good reason to do it the other way around 👍

PR #17902, opened 2026-02-12
Fix twenty sdk build

martmull:
as title

### DIFF
--- a/packages/create-twenty-app/src/utils/__tests__/app-template.spec.ts
+++ b/packages/create-twenty-app/src/utils/__tests__/app-template.spec.ts
@@ -70,7 +70,7 @@ describe('copyBaseApplicationProject', () => {
     const packageJson = await fs.readJson(packageJsonPath);
     expect(packageJson.name).toBe('my-test-app');
     expect(packageJson.version).toBe('0.1.0');
-    expect(packageJson.dependencies['twenty-sdk']).toBe('0.5.1');
+    expect(packageJson.dependencies['twenty-sdk']).toBe('0.5.2');
     expect(packageJson.scripts['app:dev']).toBe('twenty app:dev');
   });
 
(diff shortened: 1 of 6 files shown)


packages/create-twenty-app/vite.config.ts
```diff
@@ -70,13 +71,25 @@ export default defineConfig(() => {
...
       lib: { entry: entries, name: 'create-twenty-app' },
       rollupOptions: {
-        external: [
-          ...Object.keys((packageJson as any).dependencies || {}),
-          'path',
-          'fs',
-          'child_process',
-          'util',
-        ],
+        external: (id: string) => {
```
charlesBochet, line 74: could you explain this change?

packages/create-twenty-app/src/utils/__tests__/app-template.spec.ts
```diff
@@ -70,7 +70,7 @@ describe('copyBaseApplicationProject', () => {
     const packageJson = await fs.readJson(packageJsonPath);
     expect(packageJson.name).toBe('my-test-app');
     expect(packageJson.version).toBe('0.1.0');
-    expect(packageJson.dependencies['twenty-sdk']).toBe('0.5.1');
+    expect(packageJson.dependencies['twenty-sdk']).toBe('0.5.2');
```
Weiko, line 73: is it supposed to be in a spec? should we mock it instead?

thomtrp (approved):

charlesBochet:
Can you explain the change? why now?

packages/create-twenty-app/src/utils/__tests__/app-template.spec.ts
martmull: i would like to make sure the dependency equals the create-twenty-app version

martmull:
> Can you explain the change? why now?

Might be related to https://github.com/twentyhq/twenty/commit/cb7d0b83a816fbe9fa15ea8ee1e03c4b05a0be81#diff-18009ad85fc221870beff82dcd9a0224f4b926f276a933ca3eecbbc159dfd338 but not very sure

My hypothesis is that switching from a list of external to a util (source) => boolean does not match the previous behavior. postpublish and prepublish script might have an effect too

@charlesBochet @bosiraphael 

image

PR #17147, opened 2026-01-14
chore: cleanup repository root structure

FelixMalfait:
## Summary

This PR reduces clutter at the repository root to improve navigation on GitHub. The README is now visible much sooner when browsing the repo.

### Deleted from root
- `nx` wrapper script → use `npx nx` instead
- `render.yaml` → no longer used
- `jest.preset.js` → inlined `@nx/jest/preset` directly in each package's jest.config
- `.prettierrc` → moved config to `package.json`
- `.prettierignore` → patterns already covered by `.gitignore`

### Moved/Consolidated
| From | To |
|------|-----|
| `Makefile` | `packages/twenty-docker/Makefile` (merged) |
| `crowdin-app.yml` | `.github/crowdin-app.yml` |
| `crowdin-docs.yml` | `.github/crowdin-docs.yml` |
| `.vale.ini` | `.github/vale.ini` |
| `tools/eslint-rules/` | `packages/twenty-eslint-rules/` |
| `eslint.config.react.mjs` | `packages/twenty-front/eslint.config.react.mjs` |

## Result

Root items reduced from ~32 to ~22 (folders + files).

## Files updated

- GitHub workflow files updated to reference new crowdin config paths
- Jest configs updated to use `@nx/jest/preset` directly
- ESLint configs updated with new import paths
- `nx.json` updated with new paths
- `package.json` now includes prettier config and updated workspace paths
- Dockerfile updated with new eslint-rules path

PR #17916, opened 2026-02-13
chore(nx): remove leftover Nx wrapper artifacts

ehconitin:
#17147 removed the root ./nx script, but wrapper-mode artifacts were still present (installation in [nx.json](https://github.com/twentyhq/twenty/blob/main/nx.json) and tracked [nxw.js](https://github.com/twentyhq/twenty/blob/main/.nx/nxw.js), leaving an inconsistent setup.

This PR completes that cleanup by:

- removing installation from nx.json
- deleting tracked nxw.js
- ignoring nxw.js to prevent accidental re-introduction

Validated that Nx still works via yarn nx / npx nx.


Devessier (approved):
LGTM!
Thanks for making the VS Code extension work again!

PR #17918, opened 2026-02-13
`[OBJECT_MANIFEST_BREAKING_CHANGE]` Sync returns workspace migration

prastoin:
# Introduction
In this PR we start returning a workspace migration post sync so it can committed and provided within the tarball

## Universal aggregators utils
Created two utils
### deleteUniversalFlatEntityForeignKeyAggregators
Used when building a universal create action, a newly created actions should not contain any aggregated foreign key so they won't be codegen in the workspace migration but also they are overriden at uninversal to flat transpilation anw

### resetUniversalFlatEntityForeignKeyAggregators
Used before validating a new flat entity creation, some validator will consume the fk aggregator in order to validate integrity, but of optimstically provided it can result to errors. To avoid caller responsability we override them here

## create-field-action refactor
Refactored the universal and flat field create action to be following the base actions in order to ease typing
Also it was tailored to handle unlimited amount of flat field metadata in the same actions whereas in the reality we were always only sending at max 2 ( for relation fields )

Note: relation field has to be provided at the same as if not optimistic would fail to retrieve circular universal identifiers

## ObjectManifest
Now always expect a `labelIdentifierFieldMetadataUniversalIdentifier`

## Integration test
Created an integration test that creates an app,  sync a first manifest and a second implying update workspace migration action generation


packages/twenty-server/src/engine/core-modules/application/resolvers/application.resolver.ts
```diff
@@ -144,18 +145,23 @@ export class ApplicationResolver {
     });
   }
 
-  @Mutation(() => Boolean)
+  @Mutation(() => WorkspaceMigrationDTO)
```
Weiko: I actually would make it via REST here tbh. (that's probably the case for all the application stuff actually)
This is because graphql benefit is clear when you need partial records but this is not the case here.
A migration plan is a complete, self-contained object. The client will want all-or-nothing response.
This can be easily versioned `/v1/sync` if we later want to handle multiple versions of manifests etc
Rest is often better for large payload because graphql parsing can be heavy (and we plan to return large payloads with heavy nesting)
FYI @prastoin @charlesBochet

prastoin: Agreed

prastoin: Will do in a following PR

prastoin:
Fixing reset

packages/twenty-server/test/integration/metadata/suites/application/__snapshots__/successful-sync-application-workspace-migration.integration-spec.ts.snap
```diff
@@ -0,0 +1,138 @@
...
+          "canReadAllObjectRecords": false,
+          "canSoftDeleteAllObjectRecords": false,
+          "canUpdateAllObjectRecords": false,
+          "canUpdateAllSettings": false,
+          "createdAt": Any<String>,
+          "description": "A test role",
+          "icon": null,
+          "isEditable": true,
+          "label": "Test Role",
+          "rowLevelPermissionPredicateGroupUniversalIdentifiers": [],
```
prastoin, line 82: ***TODO in a next PR:*** @prastoin refactor typeorm entity types derivation to always be up to date ( ALL_METADATA_RELATION ) `rowLevelPermissionPredicateGroupUniversalIdentifiers` this should be deleted by metadata centralized tooling

packages/twenty-sdk/src/sdk/objects/__tests__/define-object.spec.ts
```diff
@@ -111,17 +113,6 @@ describe('defineObject', () => {
...
-  it('should accept missing fields', () => {
-    const config = {
-      ...validConfig,
-      fields: undefined,
-    };
-
-    const result = defineObject(config as any);
-
-    expect(result.config.fields).toBeUndefined();
-  });
```
martmull, line 123: why deleting this test?

martmull: if behavior change, we should update the test so it check the new behavior

packages/twenty-shared/src/application/objectManifestType.ts
```diff
@@ -9,4 +9,5 @@ export type ObjectManifest = SyncableEntityOptions & {
   description?: string;
   icon?: string;
   fields: ObjectFieldManifest[];
+  labelIdentifierFieldMetadataUniversalIdentifier: string;
```
martmull, line 12: Do we really need to force the user to define this `labelIdentifierFieldMetadataUniversalIdentifier` when defining an object in an app? It looks really technical to me

martmull: as discussed we need to remove it from objectManifestType and add it in to a new type object-config.ts that is used to define the manifest. We then need a util to inject it before calling sync endpoint

packages/twenty-sdk/src/sdk/objects/__tests__/define-object.spec.ts
prastoin: @martmull an object cannot be created without a `labelIdentifierFieldMetadata` which mean it should at least always have one field
When refactoring the `labelIdentifierFieldMetadataUniversalIdentifier` to become optional this test will make sense again as no custom fields will be needed to be provided as we will be adding a default `id` that will serve this purpsose

packages/twenty-shared/src/application/objectManifestType.ts
prastoin: Agreed lets do that in another PR

martmull (approved):

packages/twenty-server/test/integration/metadata/suites/application/__snapshots__/successful-sync-application-workspace-migration.integration-spec.ts.snap
prastoin: https://github.com/twentyhq/twenty/pull/17978

PR #17923, opened 2026-02-13
Create record page layout after custom object creation

Weiko:
## Context
Creating a new object should now also create its record page layout, tabs and widgets, including fields widget with its associated views/view fields.
Custom objects record page layout fields widgets don't have section per default

Note: I had to enable some widget creation through the custom API but we should now implement proper validation (which should be minimal since there is usually only the configuration type in the configuration (except for FIELDS widget which contains a viewId)

Next step: Create view field should also create a viewField for the FIELDS_WIDGET view (we should also add in the FIELDS widget configuration a newFieldDefaults which will contain default visibility and position to apply to the new view field)

The feature is still gated behind an env variable (this was necessary for workspace creation, not so much here in this case but I prefer to keep the same path for consistence)


charlesBochet:
Add integration test :)

charlesBochet (approved):

Devessier (commented):
It's interesting to see all the logic being moved to the backend. Looks correct to me. I let someone else with more backend knowledge approve the PR.

PR #17926, opened 2026-02-13
Support define is tool logic function

martmull:
- supports isTool and timeout settings in defineLogicFunction in apps and in setting tabs definition
- compute for all toolInputSchema for logic funciton, in settings and in code steps

image


packages/twenty-front/src/modules/logic-functions/hooks/useLogicFunctionUpdateFormState.ts
```diff
@@ -33,7 +33,9 @@ export const useLogicFunctionUpdateFormState = ({
   const [formValues, setFormValues] = useState<LogicFunctionFormValues>({
     name: '',
     description: '',
+    isTool: false,
     code: '',
+    timeoutSeconds: 10,
   });
```
Weiko, line 41: should we init toolInputSchema here too?

packages/twenty-front/src/modules/settings/logic-functions/components/tabs/SettingsLogicFunctionCodeEditorTab.tsx
```diff
@@ -24,7 +24,7 @@ export const SettingsLogicFunctionCodeEditorTab = ({
 }: {
   files: File[];
   handleExecute: () => void;
-  onChange: (filePath: string, value: string) => void;
+  onChange: (value: string) => void;
```
Weiko, line 27: Filepath wasn't used?

Weiko (approved):
LGTM, not sure why getFunctionInputFromSourceCode has been deleted, it seems we still use something similar or am I missing something?

martmull:
We just moved it in shared to be able to use it in twenty sdk

packages/twenty-front/src/modules/settings/logic-functions/components/tabs/SettingsLogicFunctionCodeEditorTab.tsx
martmull: Not anymore, now, logicFunction editable from ui -> one unique file

PR #17928, opened 2026-02-13
Improve security settings card grouping and description overflow

Bonapara:
# After

- Added a separtor between the two audit logs cards
- Rename the audit log card to avoid repetition
- Grouped "Invite by link" and "2 factor auth" in one group
- Changed the card component description to always be one line max with truncation & tooltips

CleanShot 2026-02-13 at 17 02 36

b49f2eb043b6712d013618bb0a4ef7f011cf2316e1163fbdee4c293bed036ac9


packages/twenty-front/src/modules/settings/components/SettingsOptions/SettingsOptionCardDescription.tsx
```diff
@@ -0,0 +1,64 @@
+import { StyledSettingsCardDescription } from '@/settings/components/SettingsOptions/SettingsCardContentBase';
+import { isValidElement, type ReactNode } from 'react';
+import { OverflowingTextWithTooltip } from 'twenty-ui/display';
+
+const getTextContentFromNode = (node: ReactNode): string => {
+  if (node === null || node === undefined || typeof node === 'boolean') {
```
charlesBochet: BEEEEURK :D
(this whole function)

charlesBochet (changes requested):

packages/twenty-front/src/modules/settings/components/SettingsOptions/SettingsOptionCardDescription.tsx
Bonapara: 🤮🤮🤮

charlesBochet (approved):
LGTM, thanks for taking the feedbacks

charlesBochet:
@claude fix the front typecheck, push in this PR

charlesBochet:
@Bonapara please recheck, there is a confusion:
- the SettingsCard should always take a string and not a reactNode, but in this case this means the <OverflowingTextWithTooltip text={description} /> should be in called internally, then it means that you are probably adding overflow hidden on parent for no reason

Bonapara:
Addressed the remaining feedback in commit 9eb7ac2b18:

- SettingsOptionCardContentSelect and SettingsOptionCardContentToggle now enforce description?: string again.
- Both components render descriptions through OverflowingTextWithTooltip internally (no ReactNode branch).
- Converted the two JSX-based descriptions to plain strings (date display format + junction relation toggle) to keep the card API consistent.
- Removed the now-unused getDisplayFormatSelectDescription helper.
- Reverted extra text-overflow/white-space styling on StyledSettingsCardDescription so truncation logic stays in OverflowingTextWithTooltip.

Bonapara:
Addressed the remaining feedback in :\n\n-  and  now enforce  again.\n- Both components render descriptions through  internally (no  branch).\n- Converted the two JSX-based descriptions to plain strings (date display format + junction relation toggle) to keep the card API consistent.\n- Removed the now-unused  helper.\n- Reverted extra  styling on  so truncation logic stays in .\n\nValidation:  passes.

Bonapara:
@greptileai

PR #17931, opened 2026-02-13
Refactor create input transpiler to return `universalFlatEntity`

prastoin:
# Introduction
As now the builder only deals with universal flat entity and that the application sync does not call the metadata service hacking their input anymore we can refactor the create input transpiler to be to universal flat instead of flat entitty


prastoin:
Not prio

PR #17944, opened 2026-02-14
Internet accounts SSRF

neo773:


packages/twenty-server/src/engine/core-modules/imap-smtp-caldav-connection/services/imap-smtp-caldav-connection-validator.service.ts
```diff
@@ -46,4 +60,47 @@ export class ImapSmtpCaldavValidatorService {
...
+    if (isPrivateIp(resolvedIp)) {
+      throw new UserInputError(
+        'Connection to private or internal network addresses is not allowed.',
+        {
+          userFriendlyMessage: msg`The server address you provided points to a private or internal network and cannot be used.`,
+        },
+      );
+    }
+  }
+
```
FelixMalfait, line 82: Could we share more code with the existing secure http driver / getSecureAdapter? For example it uses `OUTBOUND_HTTP_SAFE_MODE_ENABLED` (very rare but maybe people have a valid use case to connect to an internal private imap server, could be for testing locally for example)

Also consider two use-cases:
- What if IMAP is saved with a hostname pointing to a valid IP but then changes after it's validated
- What if the hostname is substituted milliseconds after the check (we had this issue for the HTTP node)

I don't think this SSRF is something that can be exploited because there isn't much control over the payload (unlike the HTTP node), but it's best to best extra-safe we never know 😄

neo773:
super seeded by https://github.com/twentyhq/twenty/pull/17973

PR #17953, opened 2026-02-14
Emitting the correct event based on whether the record is being soft deleted or restored.

Lakshayyy-m:
fix for #17262 
This change ensures that when the record is restored, instead of emitting a database event of "DELETED", a database event of "RESTORED" will be emitted, as it is happening in the inner working of TypeORM.

This ensures that the DELETED event are not fired on RESTORED, for example, triggering a workflow.

The screen recording shows that restoring the soft deleted records does not trigger the workflow set to to run on "Record Deleted"

https://github.com/user-attachments/assets/90e0184f-2e08-466c-a40d-1592b60e64ff


thomtrp (approved):
LGTM, thank you @Lakshayyy-m

PR #17954, opened 2026-02-14
Refactor barrel generator

neo773:
Refactored into a dedicated package CLI with proper norms and conventions for easier maintenance compared to old monolith script.

Removes massive code duplication between `packages/twenty-ui/scripts/generateBarrels.ts` and `packages/twenty-shared/scripts/generateBarrels.ts`

No regressions, can be verified by running
```
grep -rl '_   _|_      _____ _ __ | |_ _   _' packages/twenty-shared/src/ packages/twenty-ui/src/ --include='index.ts' | grep -v 'src/index.ts$' | xargs rm
npx tsx packages/twenty-build-tools/src/generate-barrels.ts --package-path packages/twenty-shared
npx tsx packages/twenty-build-tools/src/generate-barrels.ts --package-path packages/twenty-ui --with-style-css-export
git diff -- 'packages/twenty-ui/src/' 'packages/twenty-shared/src/' 'packages/twenty-ui/package.json' 'packages/twenty-shared/package.json'
```


FelixMalfait:
Hey thanks, what was the motivation for this? Was it only removing duplicated code or was there something else? I'm not sure it's worth creating a package (adds noise), maybe put it in a folder in utils package then?

neo773:
@FelixMalfait 
For motivation I saw the comment some time ago and saw it again last night and decided to address it for fun
For the dedicated package, the reasoning was the TODO comment by Paul
image

FelixMalfait:
Ok thanks! Having a large number of package creates noise and makes it harder to navigate for new users, let's move to utils

neo773:
Copy paste as is to utils folder?
Would be duplicated twice then, just confirming it before I do it

FelixMalfait:
Cut/paste rather than copy/paste :). What I mean is you should adapt the existing utils package rather than create a new package.

neo773:
Discussed with Charles, decided to not proceed for now, Closing it.

PR #17957, opened 2026-02-15
Update Dockerfile to expose the default port

d3473r:
Export the default port for auto-detection e.g. traefik: https://doc.traefik.io/traefik/reference/install-configuration/providers/docker/#port-detection


FelixMalfait:
Thanks for the contribution! Not convinced this is a good default, I think it's safer to leave it out so I'll close this for now

PR #17959, opened 2026-02-15
Fix domain issues

BOHEUS:
Fixes #17941


BOHEUS:
Validation schemas were moved directly to settings as they never returned a proper message but empty string (error messages aren't translated too as they don't appear in generated translations)

packages/twenty-front/src/modules/settings/domains/components/SettingPublicDomain.tsx
```diff
@@ -93,7 +93,23 @@ export const SettingPublicDomain = () => {
     });
   };
 
-  const validationSchema = getDomainValidationSchema(t);
+  const validationSchema = z
```
charlesBochet: @BOHEUS why inlining this? Let's move this to twenty-shared, this can be useful for other areas

packages/twenty-front/src/pages/settings/domains/SettingsDomain.tsx
```diff
@@ -40,8 +38,30 @@ export const SettingsDomain = () => {
 
   const validationSchema = z
     .object({
-      subdomain: getSubdomainValidationSchema(t),
-      customDomain: getDomainValidationSchema(t),
+      subdomain: z
```
charlesBochet: why this change?

packages/twenty-front/src/pages/settings/domains/SettingsDomain.tsx
```diff
@@ -172,6 +192,12 @@ export const SettingsDomain = () => {
   };
 
   const handleSave = async () => {
+    const isValid = await form.trigger();
```
charlesBochet, line 175: that's the real fix?

charlesBochet (changes requested):
Thanks @BOHEUS, left comments!

packages/twenty-front/src/modules/settings/domains/components/SettingPublicDomain.tsx
BOHEUS: @charlesBochet as per my earlier comment: "Validation schemas were moved directly to settings as they never returned a proper message but empty string (error messages aren't translated too as they don't appear in generated translations)"

Whenever there was a validation error, "Subdomain already taken" error was thrown instead of one specified by validation schema and submit button was disabled even if user fixes their error (see attached video), also it allowed for no subdomain at all leading to .{domain} and users not able to log in to workspace at all (in video I already did that with another workspace that's why I got error). 

https://github.com/user-attachments/assets/45dfdd24-8561-48d6-884b-89231e0f5f1a

Question, why should this be moved to twenty-shared when twenty-shared doesn't have any zod schemas in first place?

packages/twenty-front/src/pages/settings/domains/SettingsDomain.tsx
BOHEUS: Apparently? That's the part of the fix as without moved schema directly to SettingsDomain, it will just fail silently and show no error to user but they can submit (see attached video)

https://github.com/user-attachments/assets/b02ddc36-2075-4ff2-bc62-858fb04b396e

packages/twenty-front/src/pages/settings/domains/SettingsDomain.tsx
BOHEUS: Same as above, maybe there's a better way to get those moved out to separate files than last implementation as it didn't work perfectly (question is how to pass properly lingui's macro and get translated message as currently these messages aren't translated at all?)

packages/twenty-front/src/modules/settings/domains/components/SettingPublicDomain.tsx
charlesBochet: I think we could add zod schemas to twenty shared it's not an issue.
We should move it there so everybody can re-use it

BOHEUS: I still don't think that moving translations to twenty-shared is a good idea but at least the solution now works and schemas are extracted like earlier

BOHEUS:
Closed in favor of merged https://github.com/twentyhq/twenty/pull/18812

PR #17964, opened 2026-02-16
Fix Data Model instances counter showing 0

Bonapara:
## Summary
- make combined object count queries resilient to partial GraphQL failures by accepting partial data
- ignore invalid/null object connections when building totalCountByObjectMetadataItemNamePlural
- scope Data Model Settings count queries to the same active/system object subset currently visible in the table, avoiding root-resolver overfetch
- add regression tests for useCombinedGetTotalCount covering partial failure and full success

## Validation
- npx jest --config packages/twenty-front/jest.config.mjs --runTestsByPath packages/twenty-front/src/modules/object-record/multiple-objects/hooks/__tests__/useCombinedGetTotalCount.test.tsx
- npx eslint packages/twenty-front/src/pages/settings/data-model/SettingsObjectTable.tsx packages/twenty-front/src/modules/object-record/multiple-objects/hooks/useCombinedGetTotalCount.ts packages/twenty-front/src/modules/object-record/multiple-objects/hooks/__tests__/useCombinedGetTotalCount.test.tsx

### DIFF
--- a/packages/twenty-front/src/modules/object-record/multiple-objects/hooks/__tests__/useCombinedGetTotalCount.test.tsx
+++ b/packages/twenty-front/src/modules/object-record/multiple-objects/hooks/__tests__/useCombinedGetTotalCount.test.tsx
@@ -1,0 +1,137 @@
+import { gql } from '@apollo/client';
+import { renderHook, waitFor } from '@testing-library/react';
+import { GraphQLError } from 'graphql';
+
+import { useCombinedGetTotalCount } from '@/object-record/multiple-objects/hooks/useCombinedGetTotalCount';
+import { useGenerateCombinedFindManyRecordsQuery } from '@/object-record/multiple-objects/hooks/useGenerateCombinedFindManyRecordsQuery';
+import { generatedMockObjectMetadataItems } from '~/testing/utils/generatedMockObjectMetadataItems';
+import { getJestMetadataAndApolloMocksWrapper } from '~/testing/jest/getJestMetadataAndApolloMocksWrapper';
+
+jest.mock(
+  '@/object-record/multiple-objects/hooks/useGenerateCombinedFindManyRecordsQuery',
+  () => ({
+    useGenerateCombinedFindManyRecordsQuery: jest.fn(),
+  }),
+);
+
+const mockQuery = gql`
+  query CombinedFindManyRecords {
+    companies {
+      totalCount
+    }
+    apiKeys {
+      totalCount
+    }
+  }
+`;
+
+describe('useCombinedGetTotalCount', () => {
+  const companyObjectMetadataItem = generatedMockObjectMetadataItems.find(
+    ({ nameSingular }) => nameSingular === 'company',
+  );
+  const apiKeyObjectMetadataItem = generatedMockObjectMetadataItems.find(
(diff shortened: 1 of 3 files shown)


Bonapara:
Made with Codex
Saw it created a test too. Is it good?

FelixMalfait:
Nice try but it doesn't really identify what was the issue. It tries to patch malformed responses but we should rather look into why the response was malformed in the first place?

packages/twenty-front/src/modules/object-record/multiple-objects/hooks/__tests__/useCombinedGetTotalCount.test.tsx
```diff
@@ -0,0 +1,137 @@
+import { gql } from '@apollo/client';
+import { renderHook, waitFor } from '@testing-library/react';
+import { GraphQLError } from 'graphql';
+
+import { useCombinedGetTotalCount } from '@/object-record/multiple-objects/hooks/useCombinedGetTotalCount';
+import { useGenerateCombinedFindManyRecordsQuery } from '@/object-record/multiple-objects/hooks/useGenerateCombinedFindManyRecordsQuery';
+import { generatedMockObjectMetadataItems } from '~/testing/utils/generatedMockObjectMetadataItems';
+import { getJestMetadataAndApolloMocksWrapper } from '~/testing/jest/getJestMetadataAndApolloMocksWrapper';
+
+jest.mock(
```
charlesBochet, line 10: remove this tests, we tend to only test functions where we don't need to mock things (mock usually make the test unefficient as they offen need to be changed, and then the test is red without a real regression)

General rule of thumb, test utils, avoid testing hooks unless they don't have any dependency)

packages/twenty-front/src/modules/object-record/multiple-objects/hooks/useCombinedGetTotalCount.ts
```diff
@@ -35,14 +35,24 @@ export const useCombinedGetTotalCount = ({
...
-      namePlural,
-      objectRecordConnection.totalCount,
-    ]),
+  const totalCountByObjectMetadataItemNamePlural = Object.entries(
+    data ?? {},
+  ).reduce<Record<string, number>>(
+    (acc, [namePlural, objectRecordConnection]) => {
+      const totalCount = objectRecordConnection?.totalCount;
+
+      if (typeof totalCount === 'number') {
```
charlesBochet, line 49: use type guard isNumber

I don't really see the difference with before in this code.
If objectRecordConnection?.totalCount was undefined, we owuld have put acc['companies']: undefined, which is the same as not putting it like in your code

packages/twenty-front/src/pages/settings/data-model/SettingsObjectTable.tsx
```diff
@@ -58,9 +58,19 @@ export const SettingsObjectTable = ({
 
   const { updateOneObjectMetadataItem } = useUpdateOneObjectMetadataItem();
 
+  const objectMetadataItemsForTotalCount = useMemo(
```
charlesBochet, line 61: no need to memoize here, it's not helping much

charlesBochet (commented):
Not sure to fully understand the changes here, left comments

PR #17968, opened 2026-02-16
Migrate more to Jotai

charlesBochet:
As per title


charlesBochet:
@claude fix conflicts and raise the front:test coverage by adding a few tests on frontend on non covered utils. Looks for easy to test util that's not tested. We only want to test pure functions. Do not write too many test case, let's be concise to ease maintenance

Push in this PR

charlesBochet:
@claude let's lower the code coverage a bit so it passes

PR #17973, opened 2026-02-16
Prevent SSRF via IMAP/SMTP/CalDAV

neo773:
Prevents leaking of internal services by filtering out private IPs, same way we do for webhooks


FelixMalfait:
Should the previous PR be closed?

neo773:
Yes, I closed it

charlesBochet:
@neo773 could you add a desc on this PR? 
Looking at the comment above, should this PR be closed?

packages/twenty-server/src/engine/core-modules/hostname-guard/hostname-guard.module.ts
```diff
@@ -0,0 +1,9 @@
+import { Module } from '@nestjs/common';
+
+import { HostnameGuardService } from './hostname-guard.service';
+
+@Module({
+  providers: [HostnameGuardService],
+  exports: [HostnameGuardService],
+})
+export class HostnameGuardModule {}
```
charlesBochet: I am not sure we should have a host-name guard module. We already have a secure client module, let's add it to there too
(also guard naming should be used only for nest guards)

packages/twenty-server/src/engine/core-modules/hostname-guard/hostname-guard.service.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    return resolveAndValidateHostname(hostnameOrUrl);
+  }
+
+  async getValidatedUrl(serverUrl: string): Promise<string> {
+    if (!this.isSafeModeEnabled()) {
+      return serverUrl;
+    }
+
+    const resolvedIp = await resolveAndValidateHostname(serverUrl);
```
charlesBochet: code style, too many line returns

packages/twenty-server/src/engine/core-modules/secure-http-client/utils/resolve-and-validate-hostname.util.ts
```diff
@@ -0,0 +1,28 @@
+import * as dns from 'dns/promises';
+
+import { isPrivateIp } from 'src/engine/core-modules/secure-http-client/utils/is-private-ip.util';
+
+export const resolveAndValidateHostname = async (
```
charlesBochet, line 5: looks duplicated with host name guard

charlesBochet (commented):
Left comments, thanks for checking this!

charlesBochet (approved):

PR #17977, opened 2026-02-16
Add default fields on object manifest

martmull:
as title

image


packages/twenty-sdk/src/cli/utilities/build/manifest/utils/generate-default-field-universal-identifier.ts
```diff
@@ -0,0 +1,16 @@
+import type { ObjectConfig } from '@/sdk/objects/object-config';
+import { v5 } from 'uuid';
```
charlesBochet, line 2: we are using v4 on the whole project

packages/twenty-sdk/src/cli/utilities/build/manifest/utils/inject-default-fields-in-object-fields.ts
```diff
@@ -0,0 +1,20 @@
...
+import { getDefaultObjectFields } from '@/cli/utilities/build/manifest/utils/get-default-object-fields';
+
+export const injectDefaultFieldsInObjectFields = (
+  objectConfig: ObjectConfig,
+) => {
+  const defaultObjectFields = getDefaultObjectFields(objectConfig);
+
+  const objectConfigFieldNames = objectConfig.fields.map((f) => f.name);
+
+  const result = [...objectConfig.fields];
```
charlesBochet: naming

packages/twenty-sdk/src/cli/utilities/build/manifest/utils/generate-default-field-universal-identifier.ts
martmull: we need v5 here (for deterministic uuid)

charlesBochet: ok!

PR #17998, opened 2026-02-17
Add SSE for metadata and plug front components

bosiraphael:
Create the necessary tooling to listen to metadata events and plug it to the front components. Now we have a hot reload like experience when we edit a component in an app.

## Backend

- Split `EventWithQueryIds` into `ObjectRecordEventWithQueryIds` and `MetadataEventWithQueryIds`
- Publish metadata event batches to active SSE streams in `MetadataEventsToDbListener`

## Frontend

- Create a metadata event dispatching pipeline: SSE metadata events are grouped by metadata name, transformed into `MetadataOperationBrowserEventDetail` objects, and dispatched as browser `CustomEvents`
- Add `useListenToMetadataOperationBrowserEvent` hook for consuming metadata operation events filtered by metadata name and operation type
- Rename `useListenToObjectRecordEventsForQuery` to `useListenToEventsForQuery`, now accepting both `RecordGqlOperationSignature` and `MetadataGqlOperationSignature`
- Implement `useOnFrontComponentUpdated` which subscribes to front component metadata events and updates the Apollo cache when the component is modified
- Add `builtComponentChecksum` to the front component query and appends it to the component URL for browser cache invalidation


packages/twenty-front/src/modules/browser-event/hooks/useListenToMetadataOperationBrowserEvent.ts
```diff
@@ -0,0 +1,53 @@
+import { METADATA_OPERATION_BROWSER_EVENT_NAME } from '@/object-metadata/constants/MetadataOperationBrowserEventName';
+import { type MetadataOperation } from '@/object-metadata/types/MetadataOperation';
+import { type MetadataOperationBrowserEventDetail } from '@/object-metadata/types/MetadataOperationBrowserEventDetail';
+import { useEffect } from 'react';
+import { type AllMetadataName } from 'twenty-shared/metadata';
+import { isDefined, isNonEmptyArray } from 'twenty-shared/utils';
+
+export const useListenToMetadataOperationBrowserEvent = <
```
thomtrp, line 8: I think this should live in a common module with object record browser event. This is the same structure and it should remind us that both need to be maintain. I don't think this is SSE module though.
@lucasbordeau wdyt?

packages/twenty-front/src/modules/browser-event/types/MetadataOperation.ts
```diff
@@ -0,0 +1,14 @@
+export type MetadataOperation<T extends Record<string, unknown>> =
```
thomtrp, line 1: same. Looks like a frontend event module

packages/twenty-server/src/engine/metadata-event-emitter/listeners/metadata-events-to-db.listener.ts
```diff
@@ -49,5 +59,88 @@ export class MetadataEventsToDbListener {
     >(CallWebhookJobsForMetadataJob.name, metadataEventBatch, {
       retryLimit: 3,
     });
+
+    if (metadataEventBatch.events.length > 0) {
+      await this.publishToEventStreams(metadataEventBatch);
+    }
+  }
+
+  private async publishToEventStreams(
```
thomtrp: nope, this should not be done there. That should be dispatch to the `WorkspaceEventEmitterService` as the recordEvent listener does. This is the service responsible for query matching logic + event stream + redis publishing

packages/twenty-server/src/engine/metadata-event-emitter/listeners/metadata-events-to-db.listener.ts
```diff
@@ -49,5 +59,88 @@ export class MetadataEventsToDbListener {
...
+      }
+
+      if (Object.keys(streamData.queries).length === 0) {
+        continue;
+      }
+
+      const matchedQueryIds = this.getMatchingQueryIds(
+        streamData.queries,
+        metadataEventBatch.metadataName,
+      );
```
thomtrp: do we take care of permissions at some point? Also not sure why we need a different logic from the existing one in WorkspaceEventEmitterService

packages/twenty-server/src/engine/subscriptions/dtos/metadata-event.dto.ts
```diff
@@ -0,0 +1,29 @@
+import { Field, ObjectType, registerEnumType } from '@nestjs/graphql';
+
+import { ObjectRecordEventPropertiesDTO } from 'src/engine/subscriptions/dtos/object-record-event-properties.dto';
+
+export enum MetadataEventAction {
```
thomtrp: this could probably live in a separated file

lucasbordeau:
Taking a look

packages/twenty-front/src/modules/browser-event/hooks/useListenToMetadataOperationBrowserEvent.ts
lucasbordeau: Yes browser-event module seems like a good place, since those events can be emitted by anyhting.

bosiraphael:
@greptileai

packages/twenty-server/src/engine/workspace-manager/workspace-migration/workspace-migration-runner/constants/metadata-event-to-emit.constant.ts
```diff
@@ -0,0 +1,27 @@
+import { type AllMetadataName } from 'twenty-shared/metadata';
+
+export const METADATA_EVENTS_TO_EMIT = {
```
prastoin, line 3: ***Question:*** Shouldn't we only set false for the webhook ?

bosiraphael:
@greptileai

packages/twenty-server/src/engine/workspace-event-emitter/utils/is-metadata-record-matching-filter.util.ts
```diff
@@ -0,0 +1,198 @@
...
+const isNotFilter = (filter: MetadataFilter): boolean =>
+  'not' in filter && isDefined(filter.not);
+
+const isImplicitAndFilter = (filter: MetadataFilter): boolean =>
+  Object.keys(filter).length > 1;
+
+const escapeRegExp = (str: string): string =>
+  str.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+
+export const isMetadataRecordMatchingFilter = ({
```
thomtrp, line 37: why do you need such logic? Record is `Record<string, unknown>` and filter too. Why is it different from a record filter?

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.service.ts
```diff
@@ -99,12 +125,9 @@ export class WorkspaceEventEmitterService {
...
 
-    const permissionsContext = await this.fetchPermissionsContext(workspaceId);
-
-    const { flatWorkspaceMemberMaps } =
-      await this.workspaceCacheService.getOrRecompute(workspaceId, [
-        'flatWorkspaceMemberMaps',
-      ]);
+    const objectRecordContext = isMetadata
+      ? undefined
+      : await this.fetchObjectRecordStreamContext(workspaceId);
```
thomtrp: this should go directly into `processObjectRecordStreamEvents` to avoid that check

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.service.ts
```diff
@@ -436,13 +567,7 @@ export class WorkspaceEventEmitterService {
...
 
   private getMatchingQueryIds(
-    queries: Record<
-      string,
-      {
-        objectNameSingular: string;
-        variables?: { filter?: RecordGqlOperationFilter };
-      }
-    >,
+    queries: Record<string, RecordOrMetadataGqlOperationSignature>,
```
thomtrp, line 576: Since we are processing metadata in a separated path, why are we still handling both types there? Should be only records and function should be renamed

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.service.ts
```diff
@@ -468,15 +593,16 @@ export class WorkspaceEventEmitterService {
...
-    },
+    operationSignature: RecordOrMetadataGqlOperationSignature,
     event: ObjectRecordSubscriptionEvent,
     subscriberRLSFilter: RecordGqlOperationFilter | null,
     objectMetadata: FlatObjectMetadata,
     flatFieldMetadataMaps: FlatEntityMaps<FlatFieldMetadata>,
   ): boolean {
+    if (!isRecordGqlOperationSignature(operationSignature)) {
+      return false;
+    }
```
thomtrp: should not be useful since there is a function for metadata matching

packages/twenty-server/src/engine/workspace-event-emitter/utils/is-metadata-record-matching-filter.util.ts
thomtrp: Checked with you 👍

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.service.ts
bosiraphael: This allows us to check outside of the loop

thomtrp (approved):

PR #18000, opened 2026-02-17
fix: handle missing cached related record

AnkittChauhan:
### What this PR do ?

Fixes #17669 
Avoids a crash when the selected relation isn’t in cache yet


charlesBochet (commented):
I'm not a fan of this approach, could you explain why the record is missing in the first place?

charlesBochet:
@AnkittChauhan I'm closing the PR as I don't understand it but feel free to discuss by comment below, happy to re-open if it makes sense

AnkittChauhan:
@charlesBochet : Good catch this happens because the picker uses Recoil `SearchRecords` while attach expects a normalized Apollo record

The full record fetch is async so quick selection can happen before Apollo is hydrated and `getRecordFromCache `returns null

This patch prevents the crash and lets the mutation response reconcile the cache

Proper fix would be syncing the sources like hydrate Apollo from search, await fetch, or remove cache

PR #18005, opened 2026-02-17
Fixed group by query order by inside group

### LINKED ISSUE #2229: Fix board dev seeds for drag & drop
Avoid having bugs with drag & drop with board dev seeds

lucasbordeau:
Fixes https://github.com/twentyhq/core-team-issues/issues/2229

This PR fixes a bug on board, that we thought was due to dev seeds, but that was in fact a conflict of `ORDER BY` clauses at the SQL level in group by queries.

The problem was that an ORDER BY was applied on top of a sub-query ORDER BY, thus breaking the initial ordering of each group.

# Before

Capture d’écran 2026-02-17 à 16 32 56

# After 

Capture d’écran 2026-02-17 à 16 32 27


Weiko (approved):
Interesting, LGTM

PR #18011, opened 2026-02-17
fix: laggy edition of FormNumberFieldInput

AnkittChauhan:
### What this PR do ?

Stops the delay fields from updating the workflow on every keystroke by keeping values locally and saving them on blur, which fixes the cached relation error

Fixed #15709


charlesBochet:
@lucasbordeau could you take a look at the issue? I think this kind of problem should already have been solved in the past.
It's about controlled / uncontrolled components

lucasbordeau (approved):
LGTM, we could have also used a debounce but this works.

PR #18015, opened 2026-02-17
fixes: loosen up front validation, add resolveEntityRelationUniversalIdentifiers to update and restore 

ehconitin:
closes https://github.com/twentyhq/private-issues/issues/419


packages/twenty-front/src/modules/command-menu/pages/page-layout/types/PieChartConvertibleFields.ts
```diff
@@ -4,6 +4,7 @@ import {
 } from '~/generated-metadata/graphql';
 
 export type PieChartConvertibleFields = {
+  aggregateFieldMetadataId?: string;
```
prastoin, line 7: ***Question:*** were they updated server side but not in the frontend ?

packages/twenty-front/src/modules/page-layout/widgets/graph/utils/hasMinimalRequiredConfigForGraph.ts
```diff
@@ -0,0 +1,30 @@
+import { type FieldConfiguration } from '@/page-layout/types/FieldConfiguration';
+import { type FieldsConfiguration } from '@/page-layout/types/FieldsConfiguration';
+import { isDefined } from 'twenty-shared/utils';
+import { type WidgetConfiguration } from '~/generated-metadata/graphql';
+
+export const hasMinimalRequiredConfigForGraph = (
+  configuration: WidgetConfiguration | FieldsConfiguration | FieldConfiguration,
+): boolean => {
+  switch (configuration.__typename) {
```
prastoin: ***Question:*** Shouldn't we be checking the `configuration.configurationType` instead ?

packages/twenty-server/src/engine/metadata-modules/page-layout/services/page-layout-update.service.ts
```diff
@@ -505,12 +502,26 @@ export class PageLayoutUpdateService {
 
         const updatedConfiguration = widgetInput.configuration ?? null;
 
+        const {
```
prastoin, line 505: ***Remark:*** Nice catch my bad, we should also verify that this is reproductible from within the page layout widget creation through the page layout creation endpoint

prastoin: The `pageLayoutTabId` is not configured as editable though

prastoin:
```ts
import { type MetadataEntityPropertyName } from 'src/engine/metadata-modules/flat-entity/constant/all-entity-properties-configuration-by-metadata-name.constant';

export const FLAT_PAGE_LAYOUT_WIDGET_EDITABLE_PROPERTIES = [
  'title',
  'type',
  'objectMetadataId',
  'gridPosition',
  'position',
  'configuration',
] as const satisfies MetadataEntityPropertyName<'pageLayoutWidget'>[];
```

It seems like we're not providing same strictness depending on the widget creation location ( atomic endpoint, tab including endpoint, page layout including endpoint )

prastoin: But anw this will result in noop at builder lvl as
```ts
    pageLayoutTabId: {
      toCompare: false,
      toStringify: false,
      universalProperty: 'pageLayoutTabUniversalIdentifier',
    },
```

prastoin (commented):
Hey there, thanks for catching this
Lets create an integration test that cover such case as there's three endpoints to create widget from it can be tedious to keep consistency
We should also consider centralizing this logic between handlers

packages/twenty-front/src/modules/page-layout/widgets/graph/utils/hasMinimalRequiredConfigForGraph.ts
ehconitin: yes indeed!

prastoin (approved):
Approving as discussed lets create a followup PR

packages/twenty-front/src/modules/command-menu/pages/page-layout/types/PieChartConvertibleFields.ts
ehconitin: it was always a required field on PieChartConfiguration server side -- this util never added it since its inception

PR #18018, opened 2026-02-17
Harden server-side input validation and auth defaults

FelixMalfait:
## Summary

- **File storage (LocalDriver):** Add realpath resolution and symlink rejection to `writeFile`, `downloadFile`, and `downloadFolder` — brings them in line with the existing `readFile` protections. Includes unit tests.
- **JWT:** Pin signing/verification to HS256 explicitly.
- **Auth:** Revoke active refresh tokens when a user changes their password.
- **Logic functions:** Validate `handlerName` as a safe JS identifier at both DTO and runtime level, preventing injection into the generated runner script.
- **User entity:** Remove `passwordHash` from the GraphQL schema (`@Field` decorator removed, column stays).
- **Query params:** Use `crypto.randomBytes` instead of `Math.random` for SQL parameter name generation.
- **Exception filter:** Mirror the request `Origin` header instead of sending `Access-Control-Allow-Origin: *`.

## Test plan

- [x] `local.driver.spec.ts` — writeFile rejects symlinks, downloadFile rejects paths outside storage
- [ ] Verify JWT auth flow still works (login, token refresh)
- [ ] Verify password change invalidates existing sessions
- [ ] Verify logic function creation with valid/invalid handler names
- [ ] Verify file upload/download in dev environment

### DIFF
--- a/packages/twenty-server/src/engine/api/graphql/graphql-query-runner/utils/compute-where-condition-parts.ts
+++ b/packages/twenty-server/src/engine/api/graphql/graphql-query-runner/utils/compute-where-condition-parts.ts
@@ -35,9 +35,9 @@ export const computeWhereConditionParts = ({
   fieldMetadataType: FieldMetadataType;
   useDirectTableReference?: boolean;
 }): WhereConditionParts => {
-  const uuid = randomBytes(5).toString('hex');
+  const paramSuffix = randomBytes(5).toString('hex');
 
-  const secondUuid = randomBytes(5).toString('hex');
+  const secondParamSuffix = randomBytes(5).toString('hex');
 
   const fieldReference = useDirectTableReference
     ? `"${key}"`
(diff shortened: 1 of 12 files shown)


FelixMalfait:
@claude  please update the graphql schema and push to this branch (remove passwordHash as per below)

===================================================
diff --git a/packages/twenty-front/src/generated-metadata/graphql.ts b/packages/twenty-front/src/generated-metadata/graphql.ts
index 13467de80b..a606bf9ad0 100644
--- a/packages/twenty-front/src/generated-metadata/graphql.ts
+++ b/packages/twenty-front/src/generated-metadata/graphql.ts
@@ -4844,7 +4844,6 @@ export type User = {
   lastName: Scalars['String'];
   locale: Scalars['String'];
   onboardingStatus?: Maybe<OnboardingStatus>;
-  passwordHash?: Maybe<Scalars['String']>;
   supportUserHash?: Maybe<Scalars['String']>;
   updatedAt: Scalars['DateTime'];
   userVars?: Maybe<Scalars['JSONObject']>;
===================================================

packages/twenty-server/src/engine/core-modules/file-storage/drivers/local.driver.ts
```diff
@@ -74,17 +74,49 @@ export class LocalDriver implements StorageDriver {
...
+          FileStorageExceptionCode.ACCESS_DENIED,
+        );
+      }
+    } catch (error) {
+      if (error instanceof FileStorageException) {
+        throw error;
+      }
+    }
+
+    await fs.writeFile(realFilePath, params.sourceFile);
```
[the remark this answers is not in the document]
FelixMalfait: Not worth it I think

packages/twenty-server/src/engine/core-modules/file-storage/drivers/local.driver.ts
```diff
@@ -74,17 +74,49 @@ export class LocalDriver implements StorageDriver {
...
     await this.createFolder(folderPath);
 
-    await fs.writeFile(filePath, params.sourceFile);
+    const realFolderPath = realpathSync(folderPath);
+    const realFilePath = path.join(realFolderPath, path.basename(filePath));
+
+    this.assertRealPathIsWithinStorage(realFilePath);
+
+    try {
+      const stats = await fs.lstat(realFilePath);
```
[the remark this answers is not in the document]
FelixMalfait: Not worth it probably? Codex did that (not Opus) but I felt it made the code very low-level for a risk that probably doesn't exist? TBC

packages/twenty-server/src/engine/core-modules/file-storage/drivers/local.driver.ts
```diff
@@ -74,17 +74,49 @@ export class LocalDriver implements StorageDriver {
...
+        throw new FileStorageException(
+          'Access denied',
+          FileStorageExceptionCode.ACCESS_DENIED,
+        );
+      }
+    } catch (error) {
+      if (error instanceof FileStorageException) {
+        throw error;
+      }
+    }
```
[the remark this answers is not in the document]
Weiko: We probably don't want to swallow here indeed?

packages/twenty-server/src/engine/core-modules/file-storage/drivers/local.driver.ts
```diff
@@ -97,26 +129,57 @@ export class LocalDriver implements StorageDriver {
...
+    try {
+      rootFolderPath = realpathSync(resolvedPath);
+    } catch {
+      throw new FileStorageException(
+        'File not found',
+        FileStorageExceptionCode.FILE_NOT_FOUND,
+      );
+    }
+
+    this.assertRealPathIsWithinStorage(rootFolderPath);
```
Weiko, line 147: just realised that assertRealPathIsWithinStorage is a bit fragile (but not on you).
It's a simple call to "startWith" so both `/data/storage` and `/data/storage-hack` would pass the assertion

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/utils/compute-where-condition-parts.ts
```diff
@@ -33,9 +35,9 @@ export const computeWhereConditionParts = ({
   fieldMetadataType: FieldMetadataType;
   useDirectTableReference?: boolean;
 }): WhereConditionParts => {
-  const uuid = Math.random().toString(36).slice(2, 7);
+  const uuid = randomBytes(5).toString('hex');
```
Weiko: Why this change? (I don't think this is still a valid uuid after that btw) 
Actually I don't see why we actually want uuid for unique parameter names, maybe the variable name is just wrong

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/local.driver.ts
```diff
@@ -215,6 +215,12 @@ export class LocalDriver implements LogicFunctionDriver {
     builtFileAbsPath: string;
     handlerName: string;
   }) {
+    if (!/^[a-zA-Z_$][a-zA-Z0-9_$]*$/.test(handlerName)) {
```
Weiko, line 218: Do we want "$" in the handler name?

Weiko: Actually this is valid with js so why not!

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
```diff
@@ -618,6 +618,18 @@ export class AuthService {
       passwordHash: newPasswordHash,
     });
 
+    // Invalidate all existing refresh tokens for this user across all workspaces
+    await this.appTokenRepository.update(
```
Weiko, line 622: note: do we have indexes on that table btw?

packages/twenty-server/src/filters/unhandled-exception.filter.ts
```diff
@@ -9,28 +9,33 @@ import { type Response } from 'express';
...
-      'GET,HEAD,PUT,PATCH,POST,DELETE',
-    );
-    response.header(
-      'Access-Control-Allow-Headers',
-      'Origin, X-Requested-With, Content-Type, Accept',
-    );
+    const origin = request?.headers?.origin;
+
+    if (origin) {
+      response.header('Access-Control-Allow-Origin', origin);
```
Weiko: Not sure to understand this change 🤔

FelixMalfait: Good point, the model convinced me to do this change, but I can't find a good rationale, I'll revert

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
FelixMalfait: Not sure but this line wouldn't happen often, it's just when you reset a password so maybe a few times per time only

packages/twenty-server/src/engine/core-modules/file-storage/drivers/local.driver.ts
FelixMalfait: I checked and it seems to be covered already

It appends path.sep (i.e. /) to the storage root before the startsWith check. So if storageRoot is /data/storage, it checks against /data/storage/ -- meaning /data/storage-hack/foo would not pass because it doesn't start with /data/storage/.

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/utils/compute-where-condition-parts.ts
FelixMalfait: Keeps coming up in scanners so good to clean ; agree the variable should be renamed I'll do it, good point!

Weiko (approved):
LGTM

PR #18021, opened 2026-02-18
fix: Search does not get expected result

mykh-hailo:
## Summary

This PR refactors the search functionality in `search.service.ts` to improve search functionality. 

Closes: #12962 

## Changes
Parse `searchInput` parameter in `buildSearchQueryAndGetRecords` function
Update the function to filter and return matching records using `ILIKE` based on the search inputs


FelixMalfait:
I don't think this is the right fix. We should understand why search vector doesn't work well with chinese characters and apply the right transformations in search vector. Thanks

mykh-hailo:
@FelixMalfait Thanks for your comment. As far as I checked the project, it does not happen only for chinese characters but also other letters.
You can test with `Test-Ree-Question` and `TestReeQuestion` and same thing happens to them also.

FelixMalfait:
Thanks @mykh-hailo I'll open a PR to show you what I think would be the right fix (to confirm / you can test and give me your feedback if you have time!)

FelixMalfait:
https://github.com/twentyhq/twenty/pull/18027 ok trying something here (not fully tested yet, still WIP)

PR #18022, opened 2026-02-18
New field creates fields widget field

Weiko:
## Context
Introducing "NewFieldDefaultConfiguration" to FIELDS widget configurations
```typescript
{
  isVisible: boolean;
  viewFieldGroupId: string | null;
}
```

This configuration will define where a new field should be added (which section) and its default visibility inside FIELDS widget views.
The new field position should always be at the end (meaning the last position for the view fields OR the last position of a viewFieldGroup)

See "New fields" on this screenshot
Layout V1

### DIFF
--- a/packages/twenty-server/src/engine/metadata-modules/object-metadata/object-metadata.service.ts
+++ b/packages/twenty-server/src/engine/metadata-modules/object-metadata/object-metadata.service.ts
@@ -2,14 +2,14 @@ import { Injectable } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
 
 import { TypeOrmQueryService } from '@ptc-org/nestjs-query-typeorm';
-import { fromArrayToUniqueKeyRecord, isDefined } from 'twenty-shared/utils';
-import { FindManyOptions, FindOneOptions, Repository } from 'typeorm';
-import { v4 as uuidv4, v4 } from 'uuid';
 import {
   ViewOpenRecordIn,
   ViewType,
   ViewVisibility,
 } from 'twenty-shared/types';
+import { fromArrayToUniqueKeyRecord, isDefined } from 'twenty-shared/utils';
+import { FindManyOptions, FindOneOptions, Repository } from 'typeorm';
+import { v4 as uuidv4, v4 } from 'uuid';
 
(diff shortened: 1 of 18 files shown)


packages/twenty-server/src/engine/metadata-modules/object-metadata/object-metadata.service.ts
```diff
@@ -443,7 +443,10 @@ export class ObjectMetadataService extends TypeOrmQueryService<ObjectMetadataEnt
     };
 
     if (
```
Devessier, line 445: What would be the process if I wanted to seed standard record page layouts on a twenty-main record? The seed  happens on workspace creation but I can't set a feature flag before a workspace exists?

packages/twenty-server/src/engine/workspace-manager/dev-seeder/core/utils/seed-feature-flags.util.ts
```diff
@@ -87,17 +87,17 @@ export const seedFeatureFlags = async ({
...
         workspaceId: workspaceId,
         value: true,
       },
       {
-        key: FeatureFlagKey.IS_MARKETPLACE_ENABLED,
+        key: FeatureFlagKey.IS_COMMAND_MENU_ITEM_ENABLED,
         workspaceId: workspaceId,
         value: true,
       },
       {
```
Devessier, line 99: 💪

Devessier (approved):
LGTM

packages/twenty-server/src/engine/metadata-modules/object-metadata/object-metadata.service.ts
Weiko: AFAIK this service is not called during seeding, it's only for custom objects

Weiko: AFAIK this service is not called during seeding, it's only for custom objects

Weiko: AFAIK this service is not called during seeding, it's only for custom objects

PR #18023, opened 2026-02-18
Improve API Client usage and add Typescript check

charlesBochet:
## Summary

https://github.com/user-attachments/assets/1e75cc9d-d9d2-4ef2-99f9-34450f5d8de7

Add background incremental type checking (`tsc --watch`) to the SDK dev mode, so type regressions are caught when the generated API client changes — without requiring a full rebuild of source files.

Previously, removing a field from the data model would regenerate the API client, but existing front components/logic functions referencing the removed field wouldn't surface type errors (since their source didn't change, esbuild wouldn't rebuild them).

## What changed

- **Background `tsc --watch`**: a long-lived TypeScript watcher runs alongside esbuild watchers, incrementally re-checking all files when the generated client changes. Only logs on state transitions (errors appear / errors clear) to stay quiet.
- **Atomic client generation**: API client is now generated into a temp directory and swapped in atomically, avoiding a race condition where `tsc --watch` could see an empty `generated/` directory mid-regeneration.
- **Step decoupling**: orchestrator steps no longer receive `uploadFilesStep` directly. Instead, they use callbacks (`onFileBuilt`, `onApiClientGenerated`), and each step manages its own `builtFileInfos` state.
- **`apiClientChecksum` omitted from `ApplicationConfig`**: it's a build-time computed value, same as `packageJsonChecksum`. 
image

image

### DIFF
--- a/packages/twenty-sdk/src/cli/utilities/dev/orchestrator/steps/generate-api-client-orchestrator-step.ts
+++ b/packages/twenty-sdk/src/cli/utilities/dev/orchestrator/steps/generate-api-client-orchestrator-step.ts
@@ -1,46 +1,28 @@
 import { type ClientService } from '@/cli/utilities/client/client-service';
 import { type ConfigService } from '@/cli/utilities/config/config-service';
 import { type OrchestratorState } from '@/cli/utilities/dev/orchestrator/dev-mode-orchestrator-state';
-import crypto from 'crypto';
-import * as fs from 'fs-extra';
-import { join } from 'path';
-import { OUTPUT_DIR } from 'twenty-shared/application';
-import { FileFolder } from 'twenty-shared/types';
-
-export type ApiClientGeneratedFile = {
-  fileFolder: FileFolder;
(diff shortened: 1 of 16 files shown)


packages/twenty-sdk/src/cli/utilities/dev/orchestrator/steps/generate-api-client-orchestrator-step.ts
```diff
@@ -52,4 +75,56 @@ export class GenerateApiClientOrchestratorStep {
...
+      this.state.steps.uploadFiles.output.builtFileInfos.set(file.builtPath, {
+        checksum: file.checksum,
+        builtPath: file.builtPath,
+        sourcePath: file.sourcePath,
+        fileFolder: file.fileFolder,
+      });
+    }
+  }
+
+  private async copyApiClientFiles(
```
charlesBochet: should be in upload file step

martmull (approved):

PR #18025, opened 2026-02-18
fix: show "Not shared" for junction relation fields when target or intermediate object is not readable

mabdullahabaid:
When a user's role lacks read permission on the target object (e.g., Company) or the intermediate junction object (e.g., EmploymentHistory), junction relation fields like "Previous Companies" displayed as blank instead of showing "Not shared."

- In RecordFieldList, junction fields now check the junction object's read permission and set isForbidden on the field context so FieldDisplay renders "Not shared" instead of an empty field.
- In RelationFromManyFieldDisplay, if junction records exist but all nested target records are null (permission-denied by the API), the component renders "Not shared" instead of an empty list.


FelixMalfait (approved):

Weiko (approved):
LGTM

Issue #17999, opened 2026-02-17
Can't remove the first character after adding a tag in rich text


FelixMalfait:
Can't remove the letter d:

https://github.com/user-attachments/assets/fc217ed6-4a50-4031-9a8e-6b4558e585a2

Harshit-RV:
I'd like to fix this @FelixMalfait

dataCenter430:
Hi, @FelixMalfait  May I handle this issue? Could you please assign it to me? 
thank you

Lakshayyy-m:
Hey @FelixMalfait , what version is this, I just pulled the latest one and it works all fine in mine. I am able to delete the letters before the tag. Is it maybe fixed in the newer version?

https://github.com/user-attachments/assets/28f6a4e6-daca-4a2a-9870-7cf63121a730

BOHEUS:
@FelixMalfait is it still valid? Can't confirm on latest

https://github.com/user-attachments/assets/6bf8f62d-09c3-4c01-a3c2-d78b67dd44ee

FelixMalfait:
Good @BOHEUS let's close then! Thanks!

PR #18036, opened 2026-02-18
fix: unable to delete text character before mention chip with Backspace

speedcoder430:
When a text character immediately precedes an atom inline node (e.g. a mention chip) in the note editor, pressing Backspace fails to delete it. This happens because the browser's contenteditable behavior cannot correctly handle removing text adjacent to atom inline nodes.

Added a ProseMirror handleKeyDown handler that intercepts Backspace in this specific case and explicitly dispatches a delete transaction.

This PR addresses issue #17999


FelixMalfait:
Is this a documented issue? It feels kind of odd that we have to patch this way. Would be nice to understand the root cause first rather than patching. Thanks!

speedcoder430:
> Is this a documented issue? It feels kind of odd that we have to patch this way. Would be nice to understand the root cause first rather than patching. Thanks!

Thanks for the review. So the root cause is actually a well-known quirk in how browsers handle `contenteditable` around inline nodes that can't be edited (like our mention chips). When you delete the last character right before a mention, the browser removes it from the DOM but ProseMirror gets confused by the result and silently undoes the change — so it looks like Backspace just stops working.

This has been reported in the ProseMirror community before.

[https://github.com/ProseMirror/prosemirror/issues/1263]
[https://discuss.prosemirror.net/t/cursor-and-backspace-problems-with-inline-nodes/3469]

The standard way to deal with these contenteditable edge cases in ProseMirror-based editors is exactly what we're doing here — catch the specific key event and handle the deletion ourselves instead of letting the browser mess it up. It's a small, targeted handler that only kicks in for this one scenario.

Devessier:
Hey! Thanks for your contribution.

I’ve identified two regressions that this PR introduces: you can no longer delete the whole preceding word or the entire line before a chip. I don’t think fixing the issue is worth introducing additional bugs. Let’s try to find a more standard approach.

If this is a known browser issue, libraries like ProseMirror have surely found a standard way to work around it.

https://github.com/user-attachments/assets/6d012d5f-8fa5-4f75-9ef4-c369370e0d43

Devessier:
Closing for now while you find time to re-work on this PR. Thanks!

speedcoder430:
> Closing for now while you find time to re-work on this PR. Thanks!

Thanks for your detailed feedback. Let me find another approach to address this issue. Thanks again.

PR #18039, opened 2026-02-18
Add missing objects to rich app integration tests

thomtrp:
Updating rich app so it also create:
- a many to many relation
- views
- navigation items

The app was built successfully.

Will still be missing front component examples

Capture d’écran 2026-02-18 à 17 47 25


thomtrp:
@greptile-ai

packages/twenty-sdk/src/cli/__tests__/apps/rich-app/src/fields/post-card-on-post-card-recipient.field.ts
```diff
@@ -0,0 +1,24 @@
...
+  relationTargetObjectMetadataUniversalIdentifier:
+    POST_CARD_UNIVERSAL_IDENTIFIER,
+  relationTargetFieldMetadataUniversalIdentifier:
+    POST_CARD_RECIPIENTS_ON_POST_CARD_ID,
+  universalSettings: {
+    relationType: RelationType.MANY_TO_ONE,
+    onDelete: OnDeleteAction.CASCADE,
+    joinColumnName: 'postCardId',
+  },
+});
```
martmull, line 24: can't we gather forward and reverse field configs in case of relation? I think that defining 2 config for a relation field is impacting a lot the dev experience @charlesBochet @prastoin

packages/twenty-sdk/src/cli/__tests__/apps/rich-app/src/logic-functions/lookup-recipient.function.ts
```diff
@@ -0,0 +1,24 @@
...
+  toolInputSchema: {
+    type: 'object',
+    properties: {
+      recipientName: {
+        type: 'string',
+        description: 'The name of the recipient to look up',
+      },
+    },
+    required: ['recipientName'],
+  },
```
martmull: you don"t need to define that, most of the case it won't be defined so we should remove it for a rich-app testing

martmull (approved):

packages/twenty-sdk/src/cli/__tests__/apps/rich-app/src/fields/post-card-on-post-card-recipient.field.ts
prastoin: We should still have to declare the field through a "portal" from the forward relation

prastoin: I feel like we might wanna find a better devxp shortcut in the future if it's really painful for both human and llms to configure

PR #18045, opened 2026-02-18
Add currency hover tooltip in CurrencyDisplay

### LINKED ISSUE #17756: Currency display in Views is ambigious with multiple, same-named Currencies
When you have have multiple dollar-named currencies (eg. USD, AUD, SGD, NZD) displayed in a view, the icons used to identify the currency are ambigious. eg. AUD will show with a A$ previx, SGD with S$, but NZD, USD and FJD (and possibly others) all just show as $.

You actually have to edit the field value before you can see the currency code. 

Two Amounts:

Image

What are they? USD? NZD? Something else?

Image
Image

This obviously leads to confusion. 

## Expected behavior

Given AUD shows as A$ and SGD shows as S$, I would expect all currencies to have similar disambiguating icons. Clearest would be AU$, NZ$, US$, SG$, etc. Unless a currency has a unique symbol (eg. Euro), I'd expect the icon to be unambiguous. 

If someone in the US, Australia, etc is only dealing with a single currency, then perhaps there should be a setting for them to just use $, but otherwise the icons should all be unique.

Bonapara:
https://github.com/user-attachments/assets/af768ecb-b5e9-4e8d-a9f9-fee1a08ea9a0

Fixes https://github.com/twentyhq/twenty/issues/17756

### DIFF
--- a/packages/twenty-front/src/modules/ui/field/display/components/CurrencyDisplay.tsx
+++ b/packages/twenty-front/src/modules/ui/field/display/components/CurrencyDisplay.tsx
@@ -1,5 +1,5 @@
 import { useTheme } from '@emotion/react';
-import { useId } from 'react';
+import { useId, useState } from 'react';
 import { createPortal } from 'react-dom';
 import { AppTooltip, TooltipDelay, TooltipPosition } from 'twenty-ui/display';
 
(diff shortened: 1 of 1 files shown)


Bonapara:
@greptileai

charlesBochet (approved):
This will slightly degrade performances but acceptable :)

PR #18047, opened 2026-02-18
[FRONT COMPONENTS] Declare command menu items in front components

bosiraphael:
## Description

- Adds support for declaring command menu items directly within `defineFrontComponent` via an optional command config property
- Introduces a new `CommandMenuItemManifest` type in twenty-shared and wires it through the manifest build pipeline

## Example Of usage

```tsx
import { defineFrontComponent } from "twenty-sdk";

const TestAction = () => {
  return <div>Test Action</div>;
};

export default defineFrontComponent({
  universalIdentifier: "6c289461-0007-4a62-a99f-69e5c11a4ce7",
  name: "test-action",
  description: "Test Action",
  component: TestAction,
  command: {
    universalIdentifier: "c07df864-495f-46f3-9f5b-9d3ce2589e9b",
    label: "Run My Action",
    icon: "IconBolt",
    isPinned: false,
  },
});

```

## Video QA

https://github.com/user-attachments/assets/f910fc6a-44a9-45d1-87c5-f0ce64bb3878


packages/twenty-sdk/src/cli/utilities/build/common/front-component-build/utils/unwrap-define-front-component-to-direct-export.ts
```diff
@@ -1,43 +1,160 @@
...
+
+    defineFrontComponentSpecifier.remove();
+
+    if (importDeclaration.getNamedImports().length === 0) {
+      importDeclaration.remove();
+    }
+  }
+};
+
+const extractComponentNameFromDefaultExport = (
```
charlesBochet: why are we doing that? the component could be in another file
I'm not sure about these changes and the value of it

packages/twenty-sdk/src/cli/utilities/build/manifest/manifest-build.ts
```diff
@@ -216,6 +218,14 @@ export const buildManifest = async (
 
         frontComponents.push(config);
         frontComponentsFilePaths.push(relativePath);
+
+        if (command) {
+          commandMenuItems.push({
```
charlesBochet: I don't think we should put them in the manifest under their own key. I would keep it under frontComponent (as part of the FrontComponentManifest)

packages/twenty-shared/src/application/commandMenuItemManifestType.ts
```diff
@@ -0,0 +1,10 @@
+import { type SyncableEntityOptions } from '@/application/syncableEntityOptionsType';
+
+export type CommandMenuItemManifest = SyncableEntityOptions & {
```
charlesBochet: I don't think it's worth it to add this to the manifest

charlesBochet (commented):
Not a fan of the changes on the unwrap-define-front-component-to-direct-export, not sure why we actually need them

Also, I think it does not make sense to have CommandMenuItemsManifest as they are always tied to Front components

bosiraphael:
@greptileai

charlesBochet (approved):

PR #18054, opened 2026-02-18
Add objectRecordCounts query to /metadata endpoint

FelixMalfait:
## Summary

- Adds an `objectRecordCounts` query on the `/metadata` GraphQL endpoint that returns approximate record counts for all objects in the workspace
- Uses PostgreSQL's `pg_class.reltuples` catalog stats — a single instant query instead of N `COUNT(*)` table scans
- Replaces the previous `CombinedFindManyRecords` approach which hit the server's 20 root resolver limit and silently showed 0 for all counts on the settings Data Model page

### Server
- `ObjectRecordCountDTO` — GraphQL type with `objectNamePlural` and `totalCount`
- `ObjectRecordCountService` — reads `pg_class` catalog for the workspace schema
- Query added to `ObjectMetadataResolver` with `@MetadataResolver()` + `NoPermissionGuard`

### Frontend
- `OBJECT_RECORD_COUNTS` query added to `object-metadata/graphql/queries.ts`
- `useCombinedGetTotalCount` simplified to a zero-argument hook using the new query
- `SettingsObjectTable` simplified to a single hook call


packages/twenty-server/src/engine/metadata-modules/object-metadata/object-record-count.service.ts
```diff
@@ -0,0 +1,68 @@
...
+    ).filter(isDefined);
+
+    const schemaName = getWorkspaceSchemaName(workspaceId);
+
+    const dataSource =
+      await this.globalWorkspaceOrmManager.getGlobalWorkspaceDataSource();
+
+    const rows: { relname: string; approximate_count: number }[] =
+      await dataSource.query(
+        `SELECT relname, reltuples::bigint AS approximate_count
```
FelixMalfait, line 38: Image

FelixMalfait:
@Bonapara fyi this fixes the instances count bug you reported

etiennejouan (approved):
Nice !

PR #18073, opened 2026-02-19
FILES field - Attachment name display fix

etiennejouan:
with new 'file' FILES field on attachment, UI should display attachment name from file field value.
Issue with API users updating only 'file' FILES field (and not name field anymore)

### DIFF
--- a/packages/twenty-front/src/modules/activities/files/components/AttachmentRow.tsx
+++ b/packages/twenty-front/src/modules/activities/files/components/AttachmentRow.tsx
@@ -22,8 +22,8 @@ import { useHasPermissionFlag } from '@/settings/roles/hooks/useHasPermissionFla
 import { IconCalendar, OverflowingTextWithTooltip } from 'twenty-ui/display';
 import { isNavigationModifierPressed } from 'twenty-ui/utilities';
 import {
-  PermissionFlagType,
   FeatureFlagKey,
+  PermissionFlagType,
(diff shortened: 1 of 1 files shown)


packages/twenty-front/src/modules/activities/files/components/AttachmentRow.tsx
```diff
@@ -96,7 +96,11 @@ export const AttachmentRow = ({
   );
 
   const { name: originalFileName, extension: attachmentFileExtension } =
-    getFileNameAndExtension(attachment.name);
+    getFileNameAndExtension(
+      isFilesFieldMigrated
+        ? (attachment.file?.[0]?.label as string)
```
thomtrp, line 101: maybe ensure this is not null before trying to cast? do you actually really need that casting?

thomtrp (approved):

packages/twenty-front/src/modules/activities/files/components/AttachmentRow.tsx
etiennejouan: Absolutely ! I'll improve Attachment type when removing feature flag

I added a todo few lines after
https://github.com/twentyhq/twenty/pull/18073/changes/BASE..31111faeb907e352befc8b47fc4763b9ed93f05f#diff-7f7fa74b775e635071c3fcfbb5e65f9554b941d26b5d99621cb8fef0bc6c4a32R113

PR #18075, opened 2026-02-19
Billing for self-hosts

ijreilly:
## Summary

Implements enterprise licensing and per-seat billing for self-hosted environments, with Stripe as the single source of truth for subscription data.

### Components

- **twenty-website** hosts the private key to sign `ENTERPRISE_KEY` and `ENTERPRISE_VALIDITY_TOKEN`. It communicates with Stripe to emit the daily `ENTERPRISE_VALIDITY_TOKEN` if the subscription is active, based on the user's Stripe subscription ID stored in `ENTERPRISE_KEY`.
- **Stripe** is the single source of truth for subscription data (status, seats, billing).
- **The client** (twenty-server + DB + workers) saves `ENTERPRISE_KEY` in the `keyValuePair` table (or `.env` if `IS_CONFIG_VARIABLES_IN_DB_ENABLED` is false) and the daily-renewed `ENTERPRISE_VALIDITY_TOKEN` in the `appToken` table. `ENTERPRISE_VALIDITY_TOKEN` is verified client-side using a public key to grant access to enterprise features (RLS, SSO, audit logs, etc.).

### Flow

1. When requesting an upgrade to an enterprise plan (from **Enterprise** in settings), the user is shown a modal to choose monthly/yearly billing, then redirected to Stripe to enter payment details. After checkout, they land on twenty-website where they are exposed to their `ENTERPRISE_KEY`, which they paste in the UI. It is saved in the `keyValuePair` table. On activation, a first `ENTERPRISE_VALIDITY_TOKEN` with 30-day validity is stored in the `appToken` table.

2. **Every day**, a cron job runs and does two things:
   - **Refreshes the validity token**: communicates with twenty-website to get a new `ENTERPRISE_VALIDITY_TOKEN` with 30-day validity if the Stripe subscription is still active. If the subscription is in cancellation, the emitted token has a validity equal to the cancellation date. If it's no longer valid, the token is not replaced. The cron only needs to run every 30 days in practice, but runs daily so it's resilient to occasional failures.
   - **Reports seat count**: counts active (non-soft-deleted) `UserWorkspace` entries and sends the count to twenty-website, which updates the Stripe subscription quantity with proration. Seats are also reported on first activation. If the subscription is canceled or scheduled for cancellation, the seat update is skipped.

3. `ENTERPRISE_VALIDITY_TOKEN` is verified server-side via a public key to grant access to enterprise features.

### Key concepts

Three distinct checks are exposed as GraphQL fields on `Workspace`:

| Field | Meaning |
|---|---|
| `hasValidEnterpriseKey` | Has any valid enterprise key (signed JWT **or** legacy plain string) |
| `hasValidSignedEnterpriseKey` | `ENTERPRISE_KEY` is a properly signed JWT (billing portal makes sense) |
| `hasValidEnterpriseValidityToken` | `ENTERPRISE_VALIDITY_TOKEN` is present and not expired (expiration depends on signed token payload, not on "expiresAt" on appToken table which is only indicative) |

Feature access is gated by `isValid()` = `hasValidEnterpriseValidityToken || hasValidEnterpriseKey` (to support both new and legacy keys during transition). After transition isValid() = hasValidEnterpriseValidityToken

### Frontend states

The Enterprise settings page handles multiple states:
- **No key**: show "Get Enterprise" with checkout modal
- **Orphaned validity token** (token valid but no signed key): prompt user to set a valid enterprise key
- **Active/trialing but no validity token**: show subscription status with a "Reload validity token" action 
- **Active/trialing**: show full subscription info, billing portal access, cancel option
- **Cancellation scheduled**: show cancellation date, billing portal
- **Canceled**: show billing history link and option to start a new subscription
- **Past due / Incomplete**: prompt to update payment or restart

### Temporary retro-compatibility: legacy plain-text keys

Previously, enterprise features were gated by a simple check: any non-empty string in `ENTERPRISE_KEY` granted access. With this PR, we transition to a controlled system relying on signed JWTs.

To avoid breaking existing self-hosted users:
- **Legacy plain-text keys still grant access** to enterprise features. `hasValidEnterpriseKey` returns `true` for both signed JWTs and plain strings, and `isValid()` checks `hasValidEnterpriseKey` as a fallback when no validity token is present.
- **A deprecation banner** is shown at the top of the app when `hasValidEnterpriseKey` is `true` but `hasValidSignedEnterpriseKey` is `false`, informing the user that their key format is deprecated and they should activate a new signed key.
- **No billing portal or subscription management** is available for legacy keys since there is no Stripe subscription to manage.

This retro-compatibility will be removed in a future version. At that point, `isValid()` will only check `hasValidEnterpriseValidityToken`.

### Edge cases

- **Air-gapped / production environments**: for self-hosted clients that block external traffic (or for our own production), provide a long-lived `ENTERPRISE_VALIDITY_TOKEN` (e.g. 99 years) directly in the `appToken` table, with no `ENTERPRISE_KEY`. The daily cron will skip the refresh (no enterprise key to authenticate with), but the pre-seeded validity token will be used to grant feature access. No billing or seat reporting occurs in this mode.
- **`IS_CONFIG_VARIABLES_IN_DB_ENABLED` is false**: if the user tries to activate an enterprise key but DB config writes are disabled, the backend returns a clear error asking them to add `ENTERPRISE_KEY` to their `.env` file manually.
- **Canceled subscriptions**: the `/seats` endpoint skips Stripe updates for canceled or cancellation-scheduled subscriptions to avoid Stripe API errors.

### How to test
- launch twenty-website on a different url (eg localhost:1002)
- add ENTERPRISE_API_URL=http://localhost:3002/api/enterprise (or else) in your server .env
- ask me for twenty-website's .env file content (STRIPE_SECRET_KEY; STRIPE_ENTERPRISE_MONTHLY_PRICE_ID;STRIPE_ENTERPRISE_YEARLY_PRICE_ID; ENTERPRISE_JWT_PRIVATE_KEY; ENTERPRISE_JWT_PUBLIC_KEY; NEXT_PUBLIC_WEBSITE_URL)
- visit Admin panel / enterprise


packages/twenty-server/src/engine/core-modules/workspace/workspace.resolver.ts
```diff
@@ -480,4 +466,67 @@ export class WorkspaceResolver {
...
+  @UseGuards(
+    WorkspaceAuthGuard,
+    SelfHostGuard,
+    SettingsPermissionGuard(PermissionFlagType.WORKSPACE),
+  )
+  async setEnterpriseKey(
+    @Args('enterpriseKey') enterpriseKey: string,
+  ): Promise<EnterpriseLicenseInfoDTO> {
+    try {
+      await this.twentyConfigService.set('ENTERPRISE_KEY', enterpriseKey);
```
FelixMalfait: Does this also works in VARS IN DB not enabled?

packages/twenty-server/src/engine/core-modules/workspace/workspace.resolver.ts
```diff
@@ -480,4 +466,67 @@ export class WorkspaceResolver {
...
+    SelfHostGuard,
+    SettingsPermissionGuard(PermissionFlagType.WORKSPACE),
+  )
+  async enterprisePortalSession(
+    @Args('returnUrlPath', { nullable: true }) returnUrlPath?: string,
+    @AuthUser() user?: UserEntity,
+  ): Promise<string | null> {
+    return this.enterpriseKeyService.getPortalUrl(
+      returnUrlPath ?? undefined,
+      user?.email,
```
ijreilly: I need to check that, we should probably not use user email as multiple users could do
We should extract the stripe id from ENTERPRISE_KEY, that's the identifier

packages/twenty-website/src/shared/enterprise/enterprise-jwt.ts
```diff
@@ -0,0 +1,190 @@
...
+  const key = process.env.ENTERPRISE_JWT_PRIVATE_KEY;
+
+  if (!key) {
+    throw new Error('ENTERPRISE_JWT_PRIVATE_KEY is not configured');
+  }
+
+  return key.replace(/\\n/g, '\n');
+};
+
+// Base64url encode
```
FelixMalfait: You can probably ask AI to remove the comments that don't bring much value like this one

packages/twenty-front/src/pages/settings/enterprise/SettingsEnterprise.tsx
```diff
@@ -0,0 +1,330 @@
...
+      }
+    } catch {
+      enqueueErrorSnackBar({
+        message: t`Error opening billing portal`,
+      });
+    }
+  }, [fetchPortalSession, enqueueErrorSnackBar, t, returnUrlPath]);
+
+  // Portal first (reactivate or manage), then checkout (new subscription)
+  const openPortalOrCheckout = useCallback(async () => {
```
ijreilly: I wonder if this logic shouldn't lie on the back end instead

if it finds an enterprise key then it gives a portal url
if it does not then it gives a checkout url

packages/twenty-server/src/engine/core-modules/twenty-config/twenty-config.service.ts
```diff
@@ -201,6 +201,10 @@ export class TwentyConfigService {
     return this.get('TYPEORM_LOGGING');
   }
 
+  isSelfHost(): boolean {
```
FelixMalfait: I don't think we should have any notion of "isSelfHost", any instance should be considered the same way, no backdoor. We can forge a token for developer instances or the cloud instance

packages/twenty-server/src/engine/core-modules/twenty-config/config-variables.ts
```diff
@@ -1397,6 +1397,24 @@ export class ConfigVariables {
...
 
+  @ConfigVariablesMetadata({
+    group: ConfigVariablesGroup.SERVER_CONFIG,
+    isSensitive: true,
+    description:
+      'Ephemeral validity token for Enterprise license, refreshed daily',
+    type: ConfigVariableType.STRING,
+  })
+  @IsOptional()
+  ENTERPRISE_VALIDITY_TOKEN: string;
```
FelixMalfait, line 1547: TBC if ENV is the right storage for this because people can set IS_CONFIG_VARIABLES_IN_DB_ENABLED to false today

FelixMalfait: (for example on the cloud we have set it to false for security reasons)

FelixMalfait: I think I would rather use the appTokens table (but maybe it's workspace scoped and doesn't work perfectly today? Maybe adapt it?)

packages/twenty-server/src/engine/core-modules/workspace/workspace.resolver.ts
```diff
@@ -480,4 +466,63 @@ export class WorkspaceResolver {
       domainValidRecords,
     );
   }
+
```
FelixMalfait: As discussed, this shouldn't really live at the workspace level but more at the server level with the admin panel guard

packages/twenty-server/src/engine/core-modules/twenty-config/config-variables.ts
ijreilly: If the only issue is to have something working if IS_CONFIG_VARIABLES_IN_DB_ENABLED is falsy then we can do like existing UserVarService which still writes in keyValuePair table but without going through twentyConfigService. 
appToken feels a bit off for ENTERPRISE_KEY but makes more sense for ENTERPRISE_VALIDITY_TOKEN (workspaceId is nullable). We could try to store ENTERPRISE_KEY in keyValuePair and fallback to an error message saying "IS_CONFIG_VARIABLES_IN_DB_ENABLED is falsy on your server so you need to add this var to your .env" or something; and when reading it we could get it either from keyValuePair or from .env ? and ENTERPRISE_VALIDITY_TOKEN would be stored in appToken always (i think it makes sense as it stores other kinds of access token, even if they are workspace scope)

@FelixMalfait wdyt ? in the meantime i kept the config-variable implementation before we make a choice

packages/twenty-server/src/engine/core-modules/twenty-config/twenty-config.service.ts
ijreilly: yes agreed - cc.ing this from our discussion (with @Weiko )

_For self-hosted enterprise clients, I think we're forced to provide them with an ENTERPRISE_VALIDITY_TOKEN with a 99-year expiration -- they may block external traffic so the call to twenty-website to refresh their token won't work. And actually we can do the same for production. When checking isValid() for enterprise features, we only check the validity of the ENTERPRISE_VALIDITY_TOKEN.
At refresh time, we use ENTERPRISE_KEY, but if it's absent we skip.
So for these self-hosted enterprise clients / production, I recommend forging an always-valid ENTERPRISE_VALIDITY_TOKEN and no ENTERPRISE_KEY._

packages/twenty-server/src/engine/core-modules/workspace/workspace.resolver.ts
ijreilly: Updated the logic - now both are supported ! (see description)

packages/twenty-front/src/modules/app/components/SettingsRoutes.tsx
```diff
@@ -250,6 +251,12 @@ const SettingsBilling = lazy(() =>
   })),
 );
 
+const SettingsEnterprise = lazy(() =>
```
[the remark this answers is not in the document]
ijreilly: fixed

packages/twenty-website/src/app/api/enterprise/portal/route.ts
```diff
@@ -0,0 +1,59 @@
...
+} from '@/shared/enterprise/stripe-client';
+
+export const dynamic = 'force-dynamic';
+
+const ACTIVE_SUBSCRIPTION_STATUSES = ['active', 'trialing'];
+
+export async function POST(request: Request) {
+  try {
+    const body = await request.json();
+    const { enterpriseKey, returnUrl, billingInterval } = body;
```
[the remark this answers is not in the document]
ijreilly: fixed

packages/twenty-website/src/app/api/enterprise/portal/route.ts
```diff
@@ -0,0 +1,59 @@
+import { verifyEnterpriseKey } from '@/shared/enterprise/enterprise-jwt';
+import {
+  getStripeClient
+} from '@/shared/enterprise/stripe-client';
+
+export const dynamic = 'force-dynamic';
+
+const ACTIVE_SUBSCRIPTION_STATUSES = ['active', 'trialing'];
```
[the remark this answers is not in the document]
ijreilly: fixed

packages/twenty-front/src/modules/app/components/SettingsRoutes.tsx
```diff
@@ -242,6 +243,12 @@ const SettingsBilling = lazy(() =>
...
```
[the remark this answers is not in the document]
ijreilly: fixed

packages/twenty-website/src/shared/enterprise/enterprise-jwt.ts
```diff
@@ -0,0 +1,190 @@
...
+const base64UrlEncode = (data: string): string => {
+  return Buffer.from(data)
+    .toString('base64')
+    .replace(/\+/g, '-')
+    .replace(/\//g, '_')
+    .replace(/=+$/, '');
```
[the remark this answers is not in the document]
ijreilly: This is a false positive — the data parameter in base64UrlEncode comes from Buffer.from(data).toString('base64'), which always produces a well-formed base64 string (only A-Za-z0-9+/= characters). The /=+$/ regex is a standard pattern for stripping trailing = padding from base64.

packages/twenty-front/src/modules/settings/enterprise/components/EnterprisePlanModal.tsx
```diff
@@ -0,0 +1,178 @@
...
+  const { closeModal } = useModal();
+  const { enqueueErrorSnackBar } = useSnackBar();
+  const [selectedInterval, setSelectedInterval] =
+    useState<BillingInterval>('monthly');
+  const [isLoading, setIsLoading] = useState(false);
+
+  const [fetchCheckoutSession] = useLazyQuery(ENTERPRISE_CHECKOUT_SESSION);
+
+  const benefits = [
+    t`SSO (SAML / OIDC)`,
```
ijreilly, line 90: maybe we shouldn't _t_ this one

packages/twenty-front/src/modules/settings/enterprise/components/EnterprisePlanModal.tsx
```diff
@@ -0,0 +1,178 @@
...
+import { themeCssVariables } from 'twenty-ui/theme-constants';
+
+export const ENTERPRISE_PLAN_MODAL_ID = 'enterprise-plan-modal';
+
+type BillingInterval = 'monthly' | 'yearly';
+
+const MONTHLY_PRICE = 25;
+const YEARLY_PRICE = 19;
+
+const StyledSubscriptionContainer = styled.div`
```
ijreilly, line 24: TO DO - reuse existing components

packages/twenty-server/src/database/commands/cron-register-all.command.ts
```diff
@@ -150,6 +154,13 @@ export class CronRegisterAllCommand extends CommandRunner {
       },
     ];
 
+    if (!this.twentyConfigService.isBillingEnabled()) {
```
ijreilly: Maybe instead we should only run it if the enterprise key is present

packages/twenty-server/src/engine/core-modules/enterprise/constants/enterprise-public-key.constant.ts
```diff
@@ -0,0 +1,11 @@
+// RS256 public key for verifying enterprise license JWTs signed by twenty.com
+// The corresponding private key is held exclusively by twenty.com
+export const ENTERPRISE_PUBLIC_KEY = `-----BEGIN PUBLIC KEY-----
```
ijreilly: TO DO - this should be a configuration variable instead (doesnt have to be a secret)

packages/twenty-server/src/engine/core-modules/enterprise/constants/enterprise-validity-token-default-expiration-days.constant.ts
```diff
@@ -0,0 +1,5 @@
+/* @license Enterprise */
+
+const DAYS_TO_MS = (days: number) => days * 24 * 60 * 60 * 1000;
+
+export const ENTERPRISE_VALIDITY_TOKEN_DEFAULT_EXPIRATION_MS = DAYS_TO_MS(30);
```
ijreilly, line 5: 30 could be a configuration variable

packages/twenty-server/src/engine/core-modules/enterprise/services/enterprise-plan.service.ts
```diff
@@ -0,0 +1,556 @@
...
+    );
+
+    const payload = this.verifyJwt<EnterpriseValidityPayload>(token);
+
+    await this.appTokenRepository.save({
+      type: AppTokenType.EnterpriseValidityToken,
+      value: token,
+      userId: null,
+      workspaceId: null,
+      expiresAt: payload?.exp
```
ijreilly: this is only indicative

packages/twenty-website/src/shared/enterprise/enterprise-jwt.ts
```diff
@@ -0,0 +1,188 @@
...
+
+export type EnterpriseValidityPayload = {
+  sub: string;
+  status: 'valid';
+  iat: number;
+  exp: number;
+};
+
+const ALGORITHM = 'RS256';
+const DEFAULT_VALIDITY_TOKEN_DURATION_DAYS= 30;
```
ijreilly, line 17: should be a configuration variable maybe

packages/twenty-server/src/engine/core-modules/enterprise/services/enterprise-plan.service.ts
```diff
@@ -0,0 +1,550 @@
...
+      {
+        type: AppTokenType.EnterpriseValidityToken,
+        userId: IsNull(),
+        workspaceId: IsNull(),
+        revokedAt: IsNull(),
+      },
+      { revokedAt: new Date() },
+    );
+
+    const payload = this.verifyJwt<EnterpriseValidityPayload>(token);
```
Weiko, line 100: do we still want to save even if verify returns null?

packages/twenty-website/src/shared/enterprise/enterprise-jwt.ts
```diff
@@ -0,0 +1,188 @@
...
+export const signValidityToken = (
+  subscriptionId: string,
+  options?: SignValidityTokenOptions,
+): string => {
+  const now = Math.floor(Date.now() / 1000);
+  const durationDays = getValidityTokenDurationDays();
+  const subscriptionCancelAt = options?.subscriptionCancelAt ?? null;
+  const exp = computeValidityExp(now, durationDays, subscriptionCancelAt);
+
+  const payload: EnterpriseValidityPayload = {
```
Weiko, line 162: Seems we don't validate if exp cancelation date is in the past

packages/twenty-server/src/engine/core-modules/enterprise/services/enterprise-plan.service.ts
```diff
@@ -0,0 +1,550 @@
...
+
+    await this.appTokenRepository.save({
```
Weiko: We probably want the update and save in a same transaction? If for some reason the save fails, old tokens will be revoked with no new tokens, could it be an issue here?

packages/twenty-server/src/engine/core-modules/enterprise/services/enterprise-plan.service.ts
```diff
@@ -0,0 +1,550 @@
...
+    if (this.twentyConfigService.isBillingEnabled()) {
+      return true;
+    }
+
+    return isDefined(this.cachedKeyPayload);
+  }
+
+  hasValidEnterpriseValidityToken(): boolean {
+    if (this.twentyConfigService.isBillingEnabled()) {
+      return true;
```
Weiko: Can't we completely bypass everything with a simple env variable or am I missing something? What happens if a self-host enables billing? (or even what happens with the cloud? Won't this bypass the validation? I thought we wanted to replicate the flow for cloud as well)

packages/twenty-website/src/app/api/enterprise/checkout/route.ts
```diff
@@ -0,0 +1,47 @@
...
+    const stripe = getStripeClient();
+    const body = await request.json();
+    const billingInterval = body.billingInterval === 'yearly' ? 'yearly' : 'monthly';
+    const priceId = getEnterprisePriceId(billingInterval);
+    const successUrl =
+      body.successUrl ??
+      `${process.env.NEXT_PUBLIC_WEBSITE_URL}/enterprise/activate?session_id={CHECKOUT_SESSION_ID}`;
+
+
+    const session = await stripe.checkout.sessions.create({
```
Weiko, line 19: Since this route is not protected, users can generate unlimited checkout links. This is really a nitpick as this has no cost for us and this is not a security issue. I don't think we can avoid those public routes with the current design

packages/twenty-website/src/app/api/enterprise/seats/route.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    if (typeof seatCount !== 'number' || seatCount < 1) {
+      return new Response(
+        JSON.stringify({ error: 'Invalid seatCount' }),
+        { status: 400, headers: { 'Content-Type': 'application/json' } },
+      );
+    }
+
+    const payload = verifyEnterpriseKey(enterpriseKey);
```
Weiko, line 25: Similar to the other comment but a bit more dangerous imho since people can abuse this endpoint to down their seat count manually (cc @FelixMalfait)

packages/twenty-website/src/app/(public)/enterprise/activate/page.tsx
```diff
@@ -0,0 +1,172 @@
...
+      setError('No session ID provided. Please complete the checkout first.');
+      setLoading(false);
+
+      return;
+    }
+
+    const activate = async () => {
+      try {
+        const response = await fetch(
+          `/api/enterprise/activate?session_id=${sessionId}`,
```
Weiko, line 33: Same again, a bit more nitpick though but technically you can retrieve entreprise key from any sessionId (if you guess it somehow)

packages/twenty-server/src/engine/core-modules/enterprise/types/enterprise-key-payload.type.ts
```diff
@@ -0,0 +1,19 @@
+export type EnterpriseKeyPayload = {
+  sub: string;
+  licensee: string;
+  iat: number;
```
Weiko, line 4: there is no exp for Enterprise key, this is intended right?

packages/twenty-server/src/engine/core-modules/enterprise/constants/enterprise-public-key.constant.ts
Weiko: Hardcoding this key here will make key rotation harder in the future.
If for some reason the private key has been leaked, anyone can generate enterprise key for anyone meaning we will want to rotate the key but this wouldn't have effect for all self-host instances that are not up to date (and we will need to also manage the 2 keys in parallel)
(cc @FelixMalfait if you have some thoughts)
The other solution would be to fetch that key from our server but that would defy the whole flow... I think this is acceptable and that's probably a common trade-off with offline validation unless I'm missing something

Weiko (commented):
Left some comments @ijreilly

packages/twenty-server/src/engine/core-modules/enterprise/services/enterprise-plan.service.ts
ijreilly: No you are right, it was harmless but there is no point in saving an empty token

packages/twenty-website/src/shared/enterprise/enterprise-jwt.ts
ijreilly: Do you mean exp could be < now because exp will potentially be equal to subscriptionCancelAt which could be in the past ? 
This code here (signValidityToken) is only called if the subscription status is active or trialing; so subscriptionCancelAt should either be null or in the future (if the subscription is active but user has planned cancellation already)
Also, in enterprisePlanService's hasValidEnterpriseValidityToken we check that exp is > now to deem the token valid

Do you think I should add a check to early return in signValidityToken if exp is < now ? I wonder if it's its reponsibility or not - maybe it should just sign a validity token with an expiration date, be it in the past or in the future

Weiko: > Also, in enterprisePlanService's hasValidEnterpriseValidityToken we check that exp is > now to deem the token valid
Do you think I should add a check to early return in signValidityToken if exp is < now ? I wonder if it's its reponsibility or not - maybe it should just sign a validity token with an expiration date, be it in the past or in the future

That's a valid point, alright, LGTM!

packages/twenty-server/src/engine/core-modules/enterprise/services/enterprise-plan.service.ts
ijreilly: You are right I forgot about that. 
I removed the isBillingEnabled() bypass patterns

packages/twenty-server/src/engine/core-modules/enterprise/services/enterprise-plan.service.ts
Weiko: Something like that
```typescript
await this.appTokenRepository.manager.transaction(async (manager) => {
  await manager.update(AppTokenEntity, 
    {
      type: AppTokenType.EnterpriseValidityToken,
      userId: IsNull(),
      workspaceId: IsNull(),
      revokedAt: IsNull(),
    },
    { revokedAt: new Date() },
  );

 const payload = this.verifyJwt<EnterpriseValidityPayload>(token);

  await manager.save(AppTokenEntity, {
    type: AppTokenType.EnterpriseValidityToken,
    value: token,
    userId: null,
    workspaceId: null,
    expiresAt: expiresAt,
  });
});
```
What do you think?

packages/twenty-website/src/app/api/enterprise/checkout/route.ts
ijreilly: should we have rate-limiting on twenty-website and whitelist twenty-server ?

packages/twenty-server/src/engine/core-modules/enterprise/types/enterprise-key-payload.type.ts
ijreilly: yes in the current design

packages/twenty-server/src/engine/core-modules/enterprise/constants/enterprise-public-key.constant.ts
ijreilly: Could we have this key as a configuration variable that is set by default for self-hosts; and if we need to update we do and they will have to reach out to get a new one - but the issue is that they will only know when their access to enterprise feature breaks
Or they could somehow get the enterprise public key from twenty-website ?

packages/twenty-server/src/engine/core-modules/enterprise/enterprise.resolver.ts
```diff
@@ -0,0 +1,150 @@
...
+    private readonly userWorkspaceRepository: Repository<UserWorkspaceEntity>,
+  ) {}
+
+  private async getActiveUserWorkspaceCount(): Promise<number> {
+    const count = await this.userWorkspaceRepository.count({
+      where: { deletedAt: IsNull() },
+    });
+
+    return Math.max(1, count);
+  }
```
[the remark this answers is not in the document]
ijreilly: this is intended

Weiko (approved):
LGTM

PR #18076, opened 2026-02-19
fix: remove the error message for test failure in ci-front

mabdullahabaid:
Introduced an error message on twenty-front CI earlier to try and inform the user that test failure could be a coverage issue if no individual test was failing. However, it led to the assumption that it must be coverage failure in all cases even when it was test failure leading to the CI being red.

This PR reverts the change.


charlesBochet:
@mabdullahabaid I think we should push the logs into an artifact or something so we can download it

PR #18080, opened 2026-02-19
Factorize and add public-assets/*path endpoint

martmull:
as title


Weiko (approved):
LGTM. FYI @etiennejouan

PR #18085, opened 2026-02-19
feat: introduce role selector when inviting members to a workspace

mabdullahabaid:
This PR adds an explicit role selector to the "Invite by email" flow, requires a role choice before sending, and stores the selected role with each invitation. The backend now accepts and persists `roleId` on invitations and applies it when the invite is accepted, while keeping it optional to avoid breaking existing clients and legacy invites.

---

### Frontend

- **Settings → Members → Invite by email**
  - New **Role** dropdown (same `Select` pattern as member/API key role selectors) between the email input and Invite button.
  - Roles are loaded via `SettingsRolesQueryEffect` and `settingsAllRolesSelector`; only roles with `canBeAssignedToUsers` are shown.
  - Role is **required**: form validates `roleId` (e.g. `z.string().min(1)`) and the Invite button is disabled until a role is selected and emails are valid.
  - `WorkspaceInviteTeam` receives `roles` as a prop from the parent; layout is responsive (e.g. stacked on small viewports).
- **Pending invitations table**
  - New **Role** column showing the invitation’s role label (or "Unknown role" for legacy invites without `roleId`), using the same roles source for lookup.
- **Onboarding invite step**
  - When sending invites during onboarding, the workspace **default role** is used when available (`currentWorkspace?.defaultRole?.id`), so no role selector is added there.
- **GraphQL**
  - `sendInvitations` mutation accepts optional `roleId`; `findWorkspaceInvitations` and resend mutation responses include `roleId` on `WorkspaceInvitation`. Frontend types (e.g. `WorkspaceInvitation`, hook variables) updated accordingly.

---

### Backend

- **API**
  - `SendInvitationsInput` has an **optional** `roleId` (UUID, nullable). The resolver normalises `null` to `undefined` so existing callers and legacy flows are not broken.
- **Validation (when `roleId` is provided)**
  - Role checks are centralised in **RoleValidationService** (`RoleValidationModule`, in `metadata-modules/role-validation/`). It validates that the role exists in the workspace and has `canBeAssignedToUsers`, and throws a permissions-style error otherwise. This avoids circular dependencies (e.g. `RoleModule` imports `UserWorkspaceModule`, so invite/accept flows cannot depend on `RoleModule`).
  - **Send flow:** `WorkspaceInvitationResolver` and `WorkspaceInvitationService.sendInvitations` both call `RoleValidationService.validateRoleAssignableToUsersOrThrow` when `roleId` is present (resolver before calling the service; service again before creating tokens so that **resend** also validates the stored role and fails fast if the role was deleted or made unassignable).
  - **Accept flow:** `UserWorkspaceService.addUserToWorkspaceIfUserNotInWorkspace` uses the same service in `resolveRoleIdForNewMember` when an invitation provides a `roleId`, then falls back to `workspace.defaultRoleId` when not. Role/default is resolved and validated before any user/workspace/member creation.
- **Persistence**
  - Invitation app tokens store `roleId` in `context` next to `email` (`context: { email, roleId? }`). `generateInvitationToken` and `createWorkspaceInvitation` accept an optional `roleId` and only add it to `context` when defined.
- **Resend**
  - Resend passes the existing invitation’s `context.roleId` into `sendInvitations`. The service validates that role (when present) before creating the new token, so if the role was deleted or made unassignable, resend fails with a clear error instead of sending a broken link.
- **Response shape**
  - `SendInvitationsOutput.result` remains `WorkspaceInvitation[]`. When `usePersonalInvitation` is false we only push full invitation records (from `castAppTokenToWorkspaceInvitationUtil`), so the result always matches the GraphQL type (`id`, `email`, `roleId`, `expiresAt`).
- **Modules**
  - `WorkspaceInvitationModule` and `UserWorkspaceModule` import **RoleValidationModule** (not `RoleModule`) and inject **RoleValidationService** for validation. `RoleModule` imports `RoleValidationModule` and `RoleService` delegates to `RoleValidationService` for the same validation where the module graph allows.

---

### Backward compatibility

- **Optional `roleId`**: Clients that don’t send `roleId` (or send `null`) are unchanged; invitations are created without a role and the accept flow uses the workspace default role.
- **Legacy invitations**: App tokens with only `context.email` still work; `context.roleId` is optional and the UI can show e.g. "Unknown role" for those in the pending-invitations table.


packages/twenty-server/src/engine/core-modules/user-workspace/user-workspace.service.ts
```diff
@@ -163,9 +164,11 @@ export class UserWorkspaceService extends TypeOrmQueryService<UserWorkspaceEntit
 
       await this.createWorkspaceMember(workspace.id, user);
 
-      const defaultRoleId = workspace.defaultRoleId;
+      const resolvedRoleId = isDefined(roleId)
+        ? roleId
+        : workspace.defaultRoleId;
```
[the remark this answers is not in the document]
mabdullahabaid: Fixed. We now validate roleId when adding a user to a workspace: role must exist, belong to the workspace, and be assignable to users.

packages/twenty-server/src/engine/core-modules/workspace-invitation/workspace-invitation.resolver.ts
```diff
@@ -118,10 +129,34 @@ export class WorkspaceInvitationResolver {
...
+          PermissionsExceptionCode.ROLE_CANNOT_BE_ASSIGNED_TO_USERS,
+        );
+      }
+    }
+
     return await this.workspaceInvitationService.sendInvitations(
       sendInviteLinkInput.emails,
       workspace,
       workspaceMember,
+      sendInviteLinkInput.roleId ?? undefined,
```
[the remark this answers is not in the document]
mabdullahabaid: Fixed. roleId: null is now rejected in the resolver, so it can’t bypass validation. roleId remains optional (undefined) for backward compatibility with legacy clients.

packages/twenty-front/src/modules/workspace/components/WorkspaceInviteTeam.tsx
```diff
@@ -59,33 +87,61 @@ const validationSchema = z
...
 type FormInput = {
   emails: string;
+  roleId: string;
 };
 
-export const WorkspaceInviteTeam = () => {
+type WorkspaceInviteTeamProps = {
+  roles: RoleWithPartialMembers[];
+};
+
```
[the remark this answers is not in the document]
mabdullahabaid: Intentional. We keep a safe fallback (IconUser) when a role icon isn’t registered to avoid breaking the UI. No logging added to avoid noisy warnings in production.

FelixMalfait:
Screenshot 2026-02-20 at 14 45 20
Screenshot 2026-02-20 at 14 46 01

packages/twenty-front/src/pages/settings/members/SettingsWorkspaceMembers.tsx
```diff
@@ -243,211 +252,224 @@ export const SettingsWorkspaceMembers = () => {
...
+                          />
+                        </StyledIconWrapper>
+                        <StyledTextContainerWithEllipsis>
+                          {workspaceInvitation.email}
+                        </StyledTextContainerWithEllipsis>
+                      </TableCell>
+                      <TableCell>
+                        <StyledTextContainerWithEllipsis>
+                          {rolesById.get(workspaceInvitation.roleId ?? '')
+                            ?.label ?? t`Unknown role`}
```
FelixMalfait: Default role then?

FelixMalfait (commented):
Let's move the default role section to the Security tab since it becomes less important. Explain it's for workspace invite or if a role is deleted after an invitation has been sent

packages/twenty-front/src/generated-metadata/graphql.ts
```diff
@@ -2906,6 +2906,7 @@ export type MutationSaveImapSmtpCaldavAccountArgs = {
 
 export type MutationSendInvitationsArgs = {
   emails: Array<Scalars['String']>;
+  roleId?: InputMaybe<Scalars['UUID']>;
```
FelixMalfait, line 3038: Making it optional doesn't really reflect the UI. I think I like it but then we should allow leaving it to "Role" on the frontend and clicking on Send. It that case it just picks the default role

packages/twenty-front/src/pages/onboarding/InviteTeam.tsx
```diff
@@ -129,7 +129,17 @@ export const InviteTeam = () => {
...
         ),
       );
-      const result = await sendInvitation({ emails });
+
+      if (emails.length === 0) {
+        setNextOnboardingStatus();
+        return;
+      }
+
+      const roleId = currentWorkspace?.defaultRole?.id;
```
FelixMalfait: Maybe let the backend handle this since you made the roleId optional on the backend signature? The backend can retrieve the default role

packages/twenty-front/src/pages/settings/members/SettingsWorkspaceMembers.tsx
```diff
@@ -243,211 +252,224 @@ export const SettingsWorkspaceMembers = () => {
...
+            )}
+          <Section>
+            <H2Title
+              title={t`Invite by email`}
+              description={t`Send an invite email to your team`}
+            />
+            <WorkspaceInviteTeam roles={roles} />
+            {isNonEmptyArray(workspaceInvitations) && (
+              <StyledTable>
+                <TableRow
```
FelixMalfait, line 299: Broken, see image in PR

packages/twenty-server/src/engine/core-modules/workspace-invitation/services/workspace-invitation.service.ts
```diff
@@ -380,11 +400,11 @@ export class WorkspaceInvitationService {
...
           }
         } else {
-          acc.result.push(
-            invitation.value.isPersonalInvitation
-              ? castAppTokenToWorkspaceInvitationUtil(invitation.value.appToken)
-              : { email: invitation.value.email },
-          );
+          if (invitation.value.isPersonalInvitation) {
+            acc.result.push(
+              castAppTokenToWorkspaceInvitationUtil(invitation.value.appToken),
```
FelixMalfait: Curious why we had the ternary before, I don't really understand

mabdullahabaid: The ternary existed because sendInvitations supports a usePersonalInvitation parameter with two code paths: personal (creates a unique token per email) and non-personal (generic invite link, no token). Only personal invitations produce app tokens, so the ternary branched on which result to push. In practice though, usePersonalInvitation is always true (no caller ever passes false) so the non-personal branch is dead code.

Should I remove or is there a reason to keep the non-personal branch logic? I think it's redundant since we already have the public join URL functionality implemented.

packages/twenty-front/src/generated-metadata/graphql.ts
mabdullahabaid: Renamed "Role" to "Default Role" in the UI to make this information explicit and enabled the button without a role selection so backend can fallback to default role in this case.

packages/twenty-front/src/pages/settings/members/SettingsWorkspaceMembers.tsx
mabdullahabaid: Yes, fixed.

packages/twenty-front/src/pages/onboarding/InviteTeam.tsx
mabdullahabaid: Done.

packages/twenty-front/src/pages/settings/members/SettingsWorkspaceMembers.tsx
mabdullahabaid: Fixed.

packages/twenty-front/src/modules/settings/roles/components/SettingsRolesContainer.tsx
```diff
@@ -16,10 +14,9 @@ import { H3Title } from 'twenty-ui/display';
 export const SettingsRolesContainer = () => {
   const { t } = useLingui();
 
-  const settingsAllRoles = useSettingsAllRoles();
   const settingsRolesIsLoading = useAtomStateValue(settingsRolesIsLoadingState);
 
-  if (settingsRolesIsLoading && !settingsAllRoles) {
+  if (settingsRolesIsLoading) {
```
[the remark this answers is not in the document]
mabdullahabaid: `settingsAllRoles` was used by SettingsRoleDefaultRole, which has now moved to security tab instead. Therefore, this is a false positive.

packages/twenty-front/src/pages/onboarding/InviteTeam.tsx
```diff
@@ -129,6 +129,12 @@ export const InviteTeam = () => {
             .filter((email) => email.length > 0),
         ),
       );
+
+      if (emails.length === 0) {
+        setNextOnboardingStatus();
+        return;
+      }
+
       const result = await sendInvitation({ emails });
```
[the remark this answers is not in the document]
mabdullahabaid: The backend's `resolveRoleIdForNewMember` automatically assigns `workspace.defaultRoleId` when no roleId is provided. The invitation table correctly shows "Default role" (not "Unknown role") for invitations without a roleId.

packages/twenty-server/src/engine/core-modules/workspace-invitation/services/workspace-invitation.service.ts
mabdullahabaid: Removed the redundant logic.

mabdullahabaid:
This is how the table looks now. If the email is too long to fit in the cell, we truncate it and display the complete email on hover.

image

FelixMalfait (approved):
Great work!

PR #18092, opened 2026-02-19
Support `Skill` in manifest

prastoin:
# Introduction
Support skill in manifest, pre-requisite for the twenty standard app migration


packages/twenty-sdk/src/cli/commands/entity/entity-add.ts
```diff
@@ -116,6 +117,16 @@ export class EntityAddCommand {
         return { name, file };
       }
 
+      case SyncableEntity.Skill: {
+        const name = await this.getEntityName(entity);
+
+        const file = getSkillBaseFile({
```
martmull, line 123: you need to add a template in create-twenty-app also

packages/twenty-sdk/src/sdk/skills/define-skill.ts
```diff
@@ -0,0 +1,26 @@
...
+const skillManifestSchema = z.object({
+  universalIdentifier: z
+    .string()
+    .min(1, 'Skill must have a universalIdentifier'),
+  name: z.string().min(1, 'Skill must have a name'),
+  label: z.string().min(1, 'Skill must have a label'),
+  content: z.string().min(1, 'Skill must have content'),
+  icon: z.string().optional(),
+  description: z.string().optional(),
+});
```
martmull: we don't use zod for other define entity function please lets unify the behavior

martmull (commented):
we need an update of create-twenty-app to seed new skill entity
We also need to update documentation packages/twenty-docs/developers/extend/capabilities/apps.mdx and Readme.md of twenty-sdk and create-twenty-app

packages/twenty-sdk/src/sdk/skills/define-skill.ts
prastoin: Ok reverting to existing pattern, in my opinion we should add stronger data integrity validation. Checking uuid integrity on expect fields etc to avoid back and fort server validation

packages/twenty-sdk/src/cli/commands/entity/entity-add.ts
prastoin: Interesting thanks !

packages/twenty-server/test/integration/metadata/suites/application/successful-sync-application-workspace-migration.integration-spec.ts
```diff
@@ -35,7 +36,7 @@ describe('syncApplication', () => {
     appCreated = true;
   }, 60000);
 
-  afterAll(async () => {
+  afterEach(async () => {
```
martmull, line 39: Why?

martmull (approved):
Question but overall lgtm

packages/twenty-server/test/integration/metadata/suites/application/successful-sync-application-workspace-migration.integration-spec.ts
prastoin: If you don't you'll be syncing a single skill on the previous test app instance which will result in its field deletion that will make noise in the snapshot

I want every tests to start from a clean app instance

=> We wanna preserve the black box testing pattern here, the test order should not have an impact on the snapshot

PR #17793, opened 2026-02-09
feat: add draft email workflow action

neo773:

PR #18093, opened 2026-02-19
Fix google compose scope not gated by feature flag

Weiko:
## Context
New google api scope has been introduced in https://github.com/twentyhq/twenty/pull/17793 without being gated behind a feature flag

Microsoft is using pre-existing Mail.ReadWrite scope there is nothing to gate

## Before
Screenshot 2026-02-19 at 14 57 58

## After
Screenshot 2026-02-19 at 14 58 44


bosiraphael (approved):
LGTM

charlesBochet (approved):

PR #18095, opened 2026-02-19
Add page layout backfill command

Weiko:
## Context
Command to backfill record page layouts and related entities for legacy workspaces.

## Test
Set SHOULD_SEED_STANDARD_RECORD_PAGE_LAYOUTS=false, reset DB then run the command and compare with Set SHOULD_SEED_STANDARD_RECORD_PAGE_LAYOUTS=true on a different workspace


packages/twenty-server/src/database/commands/upgrade-version-command/upgrade-version-command.module.ts
```diff
@@ -14,6 +15,7 @@ import { DataSourceModule } from 'src/engine/metadata-modules/data-source/data-s
     V1_17_UpgradeVersionCommandModule,
     V1_18_UpgradeVersionCommandModule,
     V1_19_UpgradeVersionCommandModule,
+    V1_20_UpgradeVersionCommandModule,
```
prastoin: ***Remark:*** Should be 1.19

prastoin (approved):
LGTM !

PR #18098, opened 2026-02-19
Update docs - add an info for self-hosted Organization license key

BOHEUS:
As per title


FelixMalfait:
Thanks for this @BOHEUS! We are actually going to rollout the billing portal for self-serve subscriptions very soon, most likely next week. The PR has been opened by Marie already

BOHEUS:
@FelixMalfait thanks for the info!

PR #18102, opened 2026-02-19
[Feat] : Add ability to filter on all fields of an ACTOR field type

carbonFibreCode:
Fixes the issue #18099 

We have created the separate UI feature of tabs for the sub fields which are then used in the objectFilterDropDownFilterInput

Tried to implement easy and intuitive UI & UX

Let me know if the UI needs any change

Screenshot 2026-02-20 at 12 52 59 AM


FelixMalfait:
Hey, think the issue requirements weren't understood, I think it was just a search issue (text not being searched in the right place), not something that requires inventing a new UX (that would require a validation from our design lead first). Thanks for the attempt though!

PR #18142, opened 2026-02-20
Add server health check every 2 secs on app:dev

thomtrp:
https://github.com/user-attachments/assets/ff10503f-ebc0-417b-b0f3-8aa40dee74e2


martmull (approved):
excellent

PR #18143, opened 2026-02-20
Introduce metadata client in twenty sdk

Weiko:
## Context
Since graphql client leaks have been fixed, we now have clear separation between what the introspection returns for /graphql and /metadata. To reflect that we need to expose 2 separate clients.

Also fixed hello-world app with the latest sdk version


Weiko:
Blocked by regression on twenty-server not creating relations for new objects

packages/twenty-apps/hello-world/src/actions/create-new-post-card.ts
```diff
@@ -1,67 +1,61 @@
...
-        __args: {
-          data: {
-            name,
-          },
-        },
-        name: true,
-        id: true,
-      },
-    });
+    // TODO: restore after first sync creates the PostCard object on the server
```
Weiko: temporary due to circular dependency

PR #18149, opened 2026-02-20
Added record filter hidden fields in query

lucasbordeau:
Fixes https://github.com/twentyhq/twenty/issues/17506

Hidden fields are now queried when they are in record filters, to avoid optimistic and filtering bugs with hidden fields.


Weiko (approved):
LGTM

bosiraphael (approved):
LGTM

PR #18157, opened 2026-02-22
Improve helm chart // Fix linting issues & introduce Redis externalSecret for redis password // Add additional ENVs // Improve migrations

HuppertzL:
This pull request enhances the Helm chart for the Twenty application by improving how environment variables and Redis credentials are handled for both server and worker deployments. The main changes include support for injecting additional environment variables, improved Redis password management (including external secrets), and a more robust database migration workflow.

**Environment Variable Injection:**
- Added support for specifying additional environment variables for both the server and worker deployments via the `additionalEnv` field in `values.yaml`. These variables are automatically injected into the respective pods. [[1]](diffhunk://#diff-b5d958eae48fd1919e5623bcf0144aac7abb323ae8743e6f31367e383c63c296R55) [[2]](diffhunk://#diff-b5d958eae48fd1919e5623bcf0144aac7abb323ae8743e6f31367e383c63c296R109-R110) [[3]](diffhunk://#diff-20bb91909627a12b50b3c165a2a027b663479c0104ed8dbf91d2b9ad8ea8a931R74-R77) [[4]](diffhunk://#diff-20bb91909627a12b50b3c165a2a027b663479c0104ed8dbf91d2b9ad8ea8a931R157-R172) [[5]](diffhunk://#diff-20bb91909627a12b50b3c165a2a027b663479c0104ed8dbf91d2b9ad8ea8a931R225-R229) [[6]](diffhunk://#diff-fb612a3b7a13156aaa607b27d23025e2c6831f111b6a582fd313fad26d2fdb5bR89-R92)

**Redis Credential Management:**
- Introduced support for using external secrets for Redis passwords by adding `secretName` and `passwordKey` fields under `redis.external` in `values.yaml`, and logic to inject `REDIS_PASSWORD` from a Kubernetes secret if configured. [[1]](diffhunk://#diff-b5d958eae48fd1919e5623bcf0144aac7abb323ae8743e6f31367e383c63c296R180-R182) [[2]](diffhunk://#diff-5c4fa358b10abd7581188995feb9b4d6be0bc4f06a95bf27bb31b5595d6693d8R92-R100) [[3]](diffhunk://#diff-20bb91909627a12b50b3c165a2a027b663479c0104ed8dbf91d2b9ad8ea8a931R157-R172) [[4]](diffhunk://#diff-20bb91909627a12b50b3c165a2a027b663479c0104ed8dbf91d2b9ad8ea8a931R196-R205) [[5]](diffhunk://#diff-fb612a3b7a13156aaa607b27d23025e2c6831f111b6a582fd313fad26d2fdb5bR70-R79)
- Updated the logic for constructing the `REDIS_URL` to include authentication information if a password is set or an external secret is used.

**Database Migration Workflow:**
- Improved the startup command for the server deployment to optionally skip database migrations (using `DISABLE_DB_MIGRATIONS`), check for an existing schema before running migrations, and ensure setup scripts are only run on empty databases.

These changes make the chart more flexible and secure, especially for production deployments requiring externalized secrets and custom environment configurations.


HuppertzL:
@dnplkndll @larron FYI
AFAICS you were involved in adding/improving the helm chart.

charlesBochet:
Hey @HuppertzL, thanks for the PR! We've pushed a commit on top to simplify the scope. Here's a summary of what we changed and why:

### Kept (great additions)
- **Redis external secret support** (`password`, `secretName`, `passwordKey` in `values.yaml` + `twenty.redis.useExternalSecret` helper + Redis URL auth with `:$(REDIS_PASSWORD)@`)
- **Redis PVC template** (`pvc-redis-internal.yaml`) — fixed the guard to also check `persistence.enabled`
- **YAML lint/indentation fixes** (EOSQL block normalization, Redis `volumes:` indentation under pod spec)

### Changed
- **`additionalEnv` → `extraEnv`**: Replaced the key-value map with a standard Kubernetes `EnvVar` array (`extraEnv: []`) on both server and worker. This supports both plain `value` and `valueFrom` (secretKeyRef, configMapKeyRef) in a single mechanism, which is the standard pattern used by most Helm charts (Bitnami, etc.).

### Removed
- **Migration init container rewrite**: The Docker entrypoint (`packages/twenty-docker/twenty/entrypoint.sh`) already handles the full migration lifecycle on server boot: schema check → `setup-db.ts` → `database:migrate:prod` → `cache:flush` → `upgrade` → `cache:flush`. Since the entrypoint blocks before `exec "$@"`, migrations complete before the server process starts — same guarantee as an init container. Duplicating this in the init container was redundant and fragile (the init container version had a P0 regression where `yarn database:migrate:prod` was inside a conditional that skipped it on upgrades).
- **`DISABLE_DB_MIGRATIONS: "true"` on server container**: The original PR added this to disable the entrypoint's migrations (since it moved that logic to the init container). Since we reverted to letting the entrypoint handle migrations, we removed this flag — the server runs migrations on boot as intended, just like in the docker-compose setup. Note: the worker doesn't need this flag either, because in Kubernetes `command:` overrides the Docker `ENTRYPOINT`, so `entrypoint.sh` doesn't execute for the worker at all.
- The original `run-migrations` init container (from `main`) that ran `npx -y typeorm migration:run` was also removed — it was redundant with the entrypoint.

### Result
The diff vs main is now focused: Redis auth support, Redis PVC, extraEnv, and lint fixes. `helm lint` and `helm template` pass cleanly with various configurations (internal Redis, external Redis with password, external Redis with secretKeyRef, extraEnv injection).

charlesBochet (approved):

charlesBochet:
Thank you @HuppertzL

PR #18168, opened 2026-02-23
Fix CI

charlesBochet:
Fix 2 issues on CI:
- website still using ubuntu-latest
- linter that cannot use more than 2GB


bosiraphael (approved):
LGTM

PR #18199, opened 2026-02-24
Fix app env var fetching

Devessier:


martmull (approved):

ehconitin (approved):
may be use the generated hook directly?

import { useUpdateOneApplicationVariableMutation } from '~/generated-metadata/graphql'; 

also same fix here, could you review it :)

https://github.com/twentyhq/twenty/pull/18193

Weiko (approved):

PR #18202, opened 2026-02-24
fix: enable scroll in Fields dropdown when many fields are present

njlazzar-su:
Fixes #18198

## Problem
The Fields dropdown (Options → Fields → Hidden Fields) did not scroll when there were many fields, preventing access to fields below the fold.

## Root Cause
`DropdownMenuItemsContainer` in `ViewFieldsVisibleDropdownSection` and `ViewFieldsHiddenDropdownSection` was not passing the `hasMaxHeight` prop. This meant the max-height constraint (168px) was not applied, causing the dropdown to grow beyond the viewport boundary without a scrollbar.

## Fix
Added `hasMaxHeight` prop to `DropdownMenuItemsContainer` in:

- `ViewFieldsVisibleDropdownSection.tsx`
- `ViewFieldsHiddenDropdownSection.tsx`

This ensures the dropdown respects the max-height limit and renders a scrollbar when content overflows, allowing users to access all fields.

### DIFF
--- a/packages/twenty-front/src/modules/views/components/ViewFieldsHiddenDropdownSection.tsx
+++ b/packages/twenty-front/src/modules/views/components/ViewFieldsHiddenDropdownSection.tsx
@@ -52,7 +52,7 @@ export const ViewFieldsHiddenDropdownSection = () => {
 
   return (
     <>
-      <DropdownMenuItemsContainer>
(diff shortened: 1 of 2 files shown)


ehconitin:
sorry @njlazzar-su this is not the right fix!

if you want ot have a look here it is  - https://github.com/twentyhq/twenty/pull/18239

PR #18213, opened 2026-02-25
followup 18044

ehconitin:


packages/twenty-sdk/src/cli/utilities/client/__tests__/clientServiceGeneratedClientAuth.test.ts
```diff
@@ -235,6 +235,7 @@ describe('ClientService generated Twenty auth behavior', () => {
 
   it('refreshes and retries once after auth error when refresh callback is available', async () => {
     process.env.TWENTY_APP_ACCESS_TOKEN = 'stale-token';
+    process.env.TWENTY_API_KEY = 'legacy-api-key-token';
```
charlesBochet, line 247: let's completely remove it

packages/twenty-sdk/src/cli/utilities/client/twenty-client-template.ts
```diff
@@ -0,0 +1,410 @@
+import {
+  DEFAULT_API_KEY_NAME,
+  DEFAULT_API_URL_NAME,
+  DEFAULT_APP_ACCESS_TOKEN_NAME,
+} from 'twenty-shared/application';
+
+export const TWENTY_CLIENT_CONTENT = `
```
charlesBochet: @ehconitin the point is to try to have the client as not a string litteral and to inject it :)

charlesBochet: so we can have regular devXP (typescript + lint)

PR #18221, opened 2026-02-25
Resolve involved application ids from API metadata

prastoin:
# Introduction
Resolving each create|updated|deleted entities related entities application through their universal foreingKey ( silently failing if not found leaving the validator handling that )
In order to compute all the required application in the dependency flat entity maps

## Tested
Creating a field on a custom local twenty-apps installed on the workspace + view field

## Out of scope
- updated snapshot
- update graphql generated front


thomtrp (approved):
Ty !

charlesBochet (approved):

PR #18227, opened 2026-02-25
Fix bugs tied to jotai migration

charlesBochet:
Fixing a few bugs:
- CommandMenu not reactive
- Filtering on index view infinite loop


packages/twenty-front/src/modules/command-menu/hooks/useCommandMenuCloseAnimationCompleteCleanup.ts
```diff
@@ -42,24 +42,28 @@ export const useCommandMenuCloseAnimationCompleteCleanup = () => {
   const commandMenuCloseAnimationCompleteCleanup = useCallback(() => {
     closeDropdown(COMMAND_MENU_CONTEXT_CHIP_GROUPS_DROPDOWN_ID);
 
+    // Snapshot values before any mutations (Jotai store.get is live, unlike
```
Weiko, line 45: Do we need this comment?

Weiko (approved):
Interesting, LGTM!

packages/twenty-front/src/modules/command-menu/hooks/useCommandMenuCloseAnimationCompleteCleanup.ts
charlesBochet: no :)

PR #18229, opened 2026-02-25
Migrate pagelayout position frontend

Weiko:
## Context
Part 1 of migrating gridPosition in favor of typed position
FE should now always send both values to the BE and use both.

Next steps: 
- Update the backend to enforce and validate the new position field + DB migrations gridPositon -> position (type: GRID)
- Cleanup frontend usage
- Cleanup backend


packages/twenty-front/src/modules/page-layout/utils/convertPageLayoutDraftToUpdateInput.ts
```diff
@@ -27,6 +30,13 @@ export const convertPageLayoutDraftToUpdateInput = (
             rowSpan: widget.gridPosition.rowSpan,
             columnSpan: widget.gridPosition.columnSpan,
           },
+          position: {
```
Devessier, line 33: This function is called when updating dashboards and record page layouts. I think we should add a guard not to send layoutMode=GRID for record page layouts.

packages/twenty-front/src/modules/page-layout/utils/createDefaultFrontComponentWidget.ts
```diff
@@ -24,6 +25,14 @@ export const createDefaultFrontComponentWidget = (
...
     },
     gridPosition,
+    position: {
+      __typename: 'PageLayoutWidgetGridPosition',
+      layoutMode: PageLayoutTabLayoutMode.GRID,
+      row: gridPosition.row,
+      column: gridPosition.column,
+      rowSpan: gridPosition.rowSpan,
+      columnSpan: gridPosition.columnSpan,
+    },
```
Devessier, line 35: I feel we should instruct this function about the layout mode it should use. Indeed, we could create front component in record page layouts as well.

packages/twenty-front/src/modules/page-layout/utils/createDefaultGraphWidget.ts
```diff
@@ -74,6 +75,14 @@ export const createDefaultGraphWidget = ({
     type: WidgetType.GRAPH,
     configuration,
     gridPosition,
+    position: {
```
Devessier, line 78: Same thing here about layout mode

packages/twenty-front/src/modules/page-layout/utils/createDefaultIframeWidget.ts
```diff
@@ -24,6 +25,14 @@ export const createDefaultIframeWidget = (
       url,
     },
     gridPosition,
+    position: {
```
Devessier, line 28: Same thing here about layout mode

packages/twenty-front/src/modules/page-layout/utils/createDefaultStandaloneRichTextWidget.ts
```diff
@@ -24,6 +25,14 @@ export const createDefaultStandaloneRichTextWidget = (
       body,
     },
     gridPosition,
+    position: {
```
Devessier, line 28: Same thing here about layout mode

Devessier (commented):
Left a few comments. I would like to discuss the best practice to create default widgets whose position makes sense with the layout mode they are used within.

FelixMalfait:

Your preview environment is available at: http://bore.pub:15344

PR #18248, opened 2026-02-25
Add define post isntall logic function

martmull:
As title


packages/twenty-sdk/src/cli/utilities/build/manifest/manifest-build.ts
```diff
@@ -304,6 +324,34 @@ export const buildManifest = async (
...
+      ...application,
+      preInstallLogicFunctionUniversalIdentifier:
+        preInstallLogicFunctionUniversalIdentifiers[0],
+    };
+  }
+
+  if (application && postInstallLogicFunctionUniversalIdentifiers.length >= 1) {
+    application = {
+      ...application,
+      postInstallLogicFunctionUniversalIdentifier:
```
prastoin, line 350: Unless I'm mistaken this will break if the pre install or post install universalIdentifier comes to mutate unless we have an application update logic in the service we should double check that

prastoin (commented):
LGTM

prastoin (approved):

charlesBochet (approved):

PR #18257, opened 2026-02-26
Add events and properties to video, audio and iFrame

bosiraphael:
- Add media-specific events
- Extend `iFrame` with missing properties
- Enrich `SerializedEventData` with media-related target fields so serialized events carry the media element state.
- Refactor the `remote-elements` code generator to support per-element custom events


thomtrp (approved):

charlesBochet (changes requested):
I don't understand the code :)

packages/twenty-sdk/scripts/remote-dom/generators/remote-elements.generator.ts
```diff
@@ -162,12 +162,15 @@ const generateElementDefinition = (
   sourceFile: SourceFile,
   component: ComponentSchema,
   specificProperties: Record<string, PropertySchema>,
-  useSharedEvents: boolean,
-  useSharedPropertiesConfig: boolean,
+  commonEventNames: Set<string>,
+  shouldUseSharedPropertiesConfig: boolean,
```
charlesBochet: use common wording

packages/twenty-sdk/src/sdk/front-component-api/constants/SerializedEventData.ts
```diff
@@ -23,4 +23,11 @@ export type SerializedEventData = {
   deltaY?: number;
   deltaZ?: number;
   deltaMode?: number;
+  currentTime?: number;
```
charlesBochet, line 26: so we can modify it? or is it just for initialize?

bosiraphael: No, it's to get the currentTime from the events

PR #18266, opened 2026-02-26
OAuth Client - Add OAuth propagator

etiennejouan:
For OAuth Server without wildcard redirect URL

https://www.benanderson.co.uk/2023/07/28/dynamic-redirect-uris-oauth/


packages/twenty-server/src/engine/core-modules/auth/controllers/oauth-propagator.controller.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    if (
+      isDefined(twentyBaseDomain) &&
+      isDefined(redirectBaseDomain) &&
+      twentyBaseDomain === redirectBaseDomain
+    ) {
+      return true;
+    }
+
+    const workspace =
```
charlesBochet, line 79: we should use getWorkspaceUrls for this whole logic

packages/twenty-server/src/engine/core-modules/domain/utils/__test__/extract-base-domain.util.spec.ts
```diff
@@ -0,0 +1,41 @@
+import { extractBaseDomain } from 'src/engine/core-modules/domain/utils/extract-base-domain.util';
+
+describe('extractBaseDomain', () => {
```
charlesBochet: no need

packages/twenty-server/src/engine/core-modules/domain/utils/extract-base-domain.util.ts
```diff
@@ -0,0 +1,35 @@
+import psl from 'psl';
+
+import { isParsedDomain } from 'src/modules/contact-creation-manager/types/is-psl-parsed-domain.type';
+
+const isLocalhost = (hostname: string): boolean => {
```
charlesBochet: no need, this is not needed and if useful already exsits in the codebase

charlesBochet (changes requested):

charlesBochet (approved):

PR #18260, opened 2026-02-26
feat(server): add AppRegistration entity and server-level secrets

FelixMalfait:
## Summary
- Introduces a server-level `AppRegistration` entity for managing application identity, OAuth client credentials (clientId, hashed clientSecret), redirect URIs, and scopes
- Adds `AppRegistrationVariable` entity with AES-256-GCM encryption for server-scoped secrets
- Full CRUD GraphQL resolver secured with `SettingsPermissionGuard`
- Universal identifier guard service for collision detection across applications
- Variable resolution service for runtime secret access
- Database migration creating both tables with proper indexes and foreign keys
- `ApplicationEntity` gains `appRegistrationId` FK (`ON DELETE SET NULL`)
- Shared types updated: `ServerVariables` type, `ApplicationManifest` flattened (removes `marketplaceData` nesting)
- Unit tests for encryption service, identifier guard, and service logic

## Test plan
- [ ] Run `npx nx test twenty-server` — new unit tests pass
- [ ] Run `npx nx run twenty-server:database:reset` — migration applies cleanly
- [ ] Verify `appRegistration` and `appRegistrationVariable` tables are created in `core` schema
- [ ] Verify existing `Application` entities are unaffected (nullable FK)

**Part 1 of 3** — Foundation for OAuth server (PR 2) and frontend UI (PR 3)

Made with [Cursor](https://cursor.com)

PR #18262, opened 2026-02-26
feat: app registration UI, OAuth consent, and CLI integration

FelixMalfait:
### Frontend
- **SettingsAppRegistrationDetails** page with sections for:
  - Install stats (active installs, version distribution) — hidden when empty
  - General info (name, description, universal ID) with copy-on-click
  - OAuth credentials (client ID, secret rotation with confirmation modal)
  - Redirect URIs with URL validation (uses `isValidUrl` from twenty-shared)
  - Server variables (schema-driven, values editable by admin)
  - Danger zone (delete disabled when active installs exist)
- **Developer tab** replacing the old "Create" tab on Settings > Applications
  - "My Apps" section listing app registrations (hidden when empty)
- **OAuth consent screen** (`Authorize.tsx`) updated to work with AppRegistration
- GraphQL queries, mutations, and fragments for app registration CRUD
- `SettingsAdminTableCard` enhanced with copy-on-click support
- New route and navigation item for `/settings/applications/registrations/:id`

### CLI / SDK
- `RegisterAppOrchestratorStep` for automatic server-side app registration during `app:dev`
- API service methods for `findAppRegistrationByUniversalIdentifier` and `createAppRegistration`
- Config service persists `clientId` and `clientSecret` locally
- Dev mode orchestrator wires registration step into the dev flow

## Test plan
- [ ] Navigate to Settings > Applications > Developer tab
- [ ] Run `yarn app:dev` — verify app appears in "My Apps" list
- [ ] Click an app — verify details page loads with all sections
- [ ] Test redirect URI add/remove with validation (reject invalid URLs)
- [ ] Test client secret rotation with confirmation modal
- [ ] Test OAuth consent flow via browser
- [ ] Verify delete button is disabled when app has active installs

**Part 3 of 3** — Depends on PR 1 (entity) and PR 2 (OAuth server).

Made with [Cursor](https://cursor.com)

PR #18267, opened 2026-02-26
OAuth Client — Unified ApplicationRegistration, OAuth server, and frontend

FelixMalfait:
## Summary

Consolidates three separate PRs (#18260, #18261, #18262) into a single unified branch with all review feedback addressed:

### New features
- **ApplicationRegistration entity** — server-level registration for OAuth apps with encrypted server variables
- **OAuth 2.0 server** — authorization code, client credentials, refresh token grants with PKCE support
- **OAuth discovery endpoint** — `.well-known/oauth-authorization-server` metadata
- **Frontend UI** — app registration details page with credential management, redirect URI editing, and server variable configuration
- **CLI integration** — `twenty dev` auto-registers apps and stores OAuth credentials locally
- **Authorize consent screen** — OAuth consent page at `/authorize` showing requested scopes

### Review feedback addressed

**Renames (PR #18260):**
- `appRegistration` → `applicationRegistration` (entity, tables, files, imports, GraphQL types)
- `appRegistrationVariable` → `applicationRegistrationVariable`
- `clientId` → `oAuthClientId`, `clientSecretHash` → `oAuthClientSecretHash`, `redirectUris` → `oAuthRedirectUris`, `scopes` → `oAuthScopes`

**Security fixes (PR #18261):**
- Fixed redirect URI validation bypass when `oAuthRedirectUris` is an empty array
- Fixed workspace isolation in `clientCredentialsGrant` — now uses `find()` with explicit handling for multiple installations
- Added error logging in refresh token `catch` block instead of silently swallowing

**Code quality (PR #18262):**
- Split `VersionDistributionEntry` into its own file (one export per file)
- Split GraphQL queries and mutations into individual files with a shared fragment
- Removed unused `OAuth` entry from `AuthProviderEnum`
- Added loading state to `handleRotateSecret`
- Removed 27 narration-style comments from test files
- Added proper guards (`PublicEndpointGuard`, `NoPermissionGuard`) to controllers and resolvers

## Test plan

- [ ] Verify `twenty dev` registers an app and stores OAuth credentials
- [ ] Test OAuth authorization code flow end-to-end (authorize → token → API call)
- [ ] Test client credentials grant
- [ ] Verify redirect URI validation rejects requests when no URIs are registered
- [ ] Verify app registration detail page renders correctly
- [ ] Test secret rotation with loading state
- [ ] Verify server variable editing and saving
- [ ] Run `npx nx database:reset twenty-server` to validate migration

Closes #18260, #18261, #18262


packages/twenty-server/src/engine/core-modules/application-registration/__tests__/application-registration.service.spec.ts
```diff
@@ -0,0 +1,317 @@
...
+    });
+
+    it('should reject invalid scopes', async () => {
+      applicationRegistrationRepository.findOne.mockResolvedValue(null);
+
+      await expect(
+        service.create(
+          {
+            name: 'Bad Scopes App',
+            scopes: ['api', 'invalid:scope'],
```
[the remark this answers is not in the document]
FelixMalfait: Already addressed — the test uses `oAuthScopes` (not `scopes`) which matches the DTO field name.

packages/twenty-server/src/engine/core-modules/application-registration/__tests__/application-registration.service.spec.ts
```diff
@@ -0,0 +1,317 @@
...
+        },
+        {
+          provide: ApplicationRegistrationEncryptionService,
+          useValue: {
+            encrypt: jest.fn((value: string) => `enc_${value}`),
+            decrypt: jest.fn((value: string) => value.replace('enc_', '')),
+          },
+        },
+      ],
+    }).compile();
```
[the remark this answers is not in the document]
FelixMalfait: Not applicable — the `ApplicationEntity` mock is already present in the test module providers (lines 53-58).

packages/twenty-server/src/engine/core-modules/auth/services/oauth.service.ts
```diff
@@ -1,157 +1,374 @@
...
+      if (
+        applicationRegistration.oAuthRedirectUris.length === 0 ||
+        !applicationRegistration.oAuthRedirectUris.includes(redirectUri)
+      ) {
+        return this.errorResponse(
+          'invalid_grant',
+          'Redirect URI does not match any registered redirect URI',
+        );
+      }
+    }
```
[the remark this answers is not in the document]
FelixMalfait: Already handled — when `redirect_uri` is omitted, it defaults to empty string. The code checks `if (!redirectUri)` (falsy for empty string) and returns an `invalid_request` error when redirect URIs are configured on the registration. Empty string never bypasses validation.

packages/twenty-server/src/engine/core-modules/application-registration/controllers/oauth-token.controller.ts
```diff
@@ -0,0 +1,86 @@
...
+          clientId: body.client_id ?? '',
+          clientSecret: body.client_secret,
+        });
+
+      default:
+        return {
+          error: 'unsupported_grant_type',
+          error_description: `Grant type '${body.grant_type}' is not supported`,
+        };
+    }
```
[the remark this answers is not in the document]
FelixMalfait: Already fixed — the error message is now a static string that does not interpolate user input: `"The provided grant_type is not supported. Supported values: authorization_code, client_credentials, refresh_token"`.

packages/twenty-front/src/pages/auth/Authorize.tsx
```diff
@@ -94,14 +118,19 @@ export const Authorize = () => {
...
           redirectUrl,
         },
-        onCompleted: (data) => {
-          redirect(data.authorizeApp.redirectUrl);
+        onCompleted: (responseData) => {
+          redirect(responseData.authorizeApp.redirectUrl);
         },
       });
     }
   };
```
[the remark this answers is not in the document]
FelixMalfait: The server already validates `redirectUrl` against registered URIs in `auth.service.ts` (lines 517-526) before issuing the authorization code. Client-side re-validation is unnecessary defense-in-depth for this flow. Regarding PKCE: `codeChallenge` is now optional — the `handleAuthorize` function only requires `clientId` and `redirectUrl`.

packages/twenty-sdk/src/cli/utilities/dev/orchestrator/steps/register-app-orchestrator-step.ts
```diff
@@ -0,0 +1,120 @@
...
+      {
+        message: `App registration created: ${input.manifest.application.displayName}`,
+        status: 'success',
+      },
+      {
+        message: `Client ID: ${createResult.data.applicationRegistration.oAuthClientId}`,
+        status: 'info',
+      },
+      {
+        message: `Client Secret: ${createResult.data.clientSecret}`,
```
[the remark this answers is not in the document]
FelixMalfait: Intentional behavior for the CLI dev workflow. This is a one-time display during `twenty dev` setup (analogous to Stripe CLI showing the secret once). The message explicitly warns: "The secret will not be shown again." The secret is also persisted in the local config store. Masking it would make the dev onboarding experience worse.

packages/twenty-server/src/database/typeorm/core/migrations/common/1772267875868-create-application-registration.ts
```diff
@@ -0,0 +1,113 @@
...
+        "oAuthClientId" text NOT NULL,
+        "oAuthClientSecretHash" text,
+        "oAuthRedirectUris" text[] NOT NULL DEFAULT '{}',
+        "oAuthScopes" text[] NOT NULL DEFAULT '{}',
+        "createdByUserId" uuid,
+        "websiteUrl" text,
+        "termsUrl" text,
+        "createdAt" TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
+        "updatedAt" TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
+        "deletedAt" TIMESTAMP WITH TIME ZONE,
```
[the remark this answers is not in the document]
FelixMalfait: Intentional design decision. Soft-delete is appropriate for `applicationRegistration` because permanently deleting an OAuth client config would instantly break all applications relying on it. The `deletedAt` column enables a recovery path. The partial unique indexes on `universalIdentifier` and `oAuthClientId` ensure uniqueness among active registrations while allowing the same identifiers to be reclaimed after soft-deletion.

packages/twenty-server/src/engine/core-modules/application-registration/application-registration.resolver.ts
```diff
@@ -0,0 +1,179 @@
...
+
+  @UseGuards(PublicEndpointGuard, NoPermissionGuard)
+  @Query(() => ApplicationRegistrationEntity, { nullable: true })
+  async findApplicationRegistrationByClientId(
+    @Args('clientId') clientId: string,
+  ): Promise<ApplicationRegistrationEntity | null> {
+    return this.applicationRegistrationService.findOneByClientId(clientId);
+  }
+
+  @UseGuards(WorkspaceAuthGuard, NoPermissionGuard)
```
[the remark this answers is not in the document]
FelixMalfait: Intentional for CLI auto-registration. The `twenty dev` CLI needs to create application registrations without requiring `API_KEYS_AND_WEBHOOKS` permission. The `findApplicationRegistrationByUniversalIdentifier` query uses the same guard for the same reason. The CLI authenticates with a workspace token, and the created registration is global (not workspace-scoped), so the settings permission scope does not apply to this operation.

packages/twenty-sdk/src/cli/utilities/config/config-service.ts
```diff
@@ -8,6 +8,8 @@ export type TwentyConfig = {
   apiKey?: string;
   applicationAccessToken?: string;
   applicationRefreshToken?: string;
+  oauthClientId?: string;
+  oauthClientSecret?: string;
```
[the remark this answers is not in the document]
FelixMalfait: Acknowledged — `getConfigForWorkspace` does not return the new `oauthClientId`/`oauthClientSecret` fields. In practice this does not affect functionality: the CLI orchestrator keeps credentials in memory for the current session, and on subsequent runs it discovers existing registrations via the GraphQL `findApplicationRegistrationByUniversalIdentifier` query. Will address in a follow-up if needed.

packages/twenty-server/src/engine/core-modules/application-registration/application-registration-identifier-validation.service.ts
```diff
@@ -0,0 +1,55 @@
...
+  // Grandfather policy: applications without applicationRegistrationId are not checked.
+  async validateUniversalIdentifierOwnership(params: {
+    universalIdentifier: string;
+    applicationId: string;
+  }): Promise<{ isValid: boolean; errorMessage?: string }> {
+    const application = await this.applicationRepository.findOne({
+      where: { id: params.applicationId },
+    });
+
+    if (!isDefined(application?.applicationRegistrationId)) {
```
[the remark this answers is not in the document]
FelixMalfait: Not applicable. The method implements a grandfather clause: applications without an `applicationRegistrationId` (pre-existing apps created before the registration system) are exempt from ownership checks. If the application does not exist (`findOne` returns null), `application?.applicationRegistrationId` is `undefined`, so `isDefined(undefined)` returns `false`, and the grandfather clause correctly lets it pass. This method is only called during workspace metadata sync where `applicationId` is already validated as an existing application.

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
```diff
@@ -495,46 +495,58 @@ export class AuthService {
...
+        `redirectUrl not provided for '${clientId}'`,
         AuthExceptionCode.FORBIDDEN_EXCEPTION,
       );
     }
 
-    if (client.redirectUrl !== authorizeAppInput.redirectUrl) {
+    if (
+      !applicationRegistration.oAuthRedirectUris.includes(
+        authorizeAppInput.redirectUrl,
+      )
```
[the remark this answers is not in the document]
FelixMalfait: Not applicable. Both `oAuthRedirectUris` and `oAuthScopes` are defined with `NOT NULL DEFAULT '{}'` in the database migration and `{ type: 'text', array: true, default: '{}' }` on the entity. TypeORM hydrates these as empty arrays, never `null`. The entity type annotations are `string[]` (not `string[] | null`). The database constraint prevents null values at the storage layer.

packages/twenty-front/src/modules/settings/admin-panel/components/SettingsAdminTableCard.tsx
```diff
@@ -82,7 +83,11 @@ export const SettingsAdminTableCard = ({
                 {item.Icon && <item.Icon size={theme.icon.size.md} />}
                 <span>{item.label}</span>
               </StyledTableCellLabel>
-              <StyledTableCellValue align={valueAlign}>
+              <StyledTableCellValue
+                align={valueAlign}
+                onClick={item.onClick}
+                style={item.onClick ? { cursor: 'pointer' } : undefined}
```
Devessier: `cursor: pointer` could be moved to the styled component

packages/twenty-front/src/pages/auth/Authorize.tsx
```diff
@@ -55,53 +57,93 @@ const StyledButtonContainer = styled.div`
...
+
+    return null;
+  }
+
+  if (!loading && !isDefined(applicationRegistration)) {
+    navigate(AppPath.NotFound);
+
+    return null;
+  }
+
```
Devessier: Do we usually navigate directly from the component rendering function? I would use a useEffect here but Claude might be able to tell us if we can just do what you did.

packages/twenty-front/src/pages/auth/Authorize.tsx
```diff
@@ -55,53 +57,93 @@ const StyledButtonContainer = styled.div`
...
-      logo: 'images/integrations/chrome-icon.svg',
-    },
-  ]);
-  const [app, setApp] = useState<App>();
+
+  const oauthScopeLabels: Record<string, string> = {
+    api: t`Access workspace data`,
+    profile: t`Read your profile`,
+  };
+
```
Devessier: Could it be typed strictly?

packages/twenty-sdk/src/cli/utilities/api/api-service.ts
```diff
@@ -260,8 +260,126 @@ export class ApiService {
...
+        query FindApplicationRegistrationByUniversalIdentifier($universalIdentifier: String!) {
+          findApplicationRegistrationByUniversalIdentifier(universalIdentifier: $universalIdentifier) {
+            id
+            universalIdentifier
+            name
+            oAuthClientId
+          }
+        }
+      `;
+
```
Devessier, line 284: I thought the SDK was auto-generated but it seems that the "auto-generation" involves Claude tokens from our side 🥸

packages/twenty-server/src/engine/core-modules/application-registration/application-registration-encryption.service.ts
```diff
@@ -0,0 +1,71 @@
...
+
+import { TwentyConfigService } from 'src/engine/core-modules/twenty-config/twenty-config.service';
+
+const ALGORITHM = 'aes-256-gcm';
+const IV_LENGTH = 16;
+const AUTH_TAG_LENGTH = 16;
+const HKDF_CONTEXT = 'application-registration-variable';
+
+@Injectable()
+export class ApplicationRegistrationEncryptionService {
```
Devessier: I thought we already had encryption service like SimpleSecretEncryptionUtil. I think it would be great to unify.

packages/twenty-server/src/engine/core-modules/application-registration/application-registration-encryption.service.ts
```diff
@@ -0,0 +1,71 @@
...
+  private encryptionKey: Buffer;
+
+  constructor(private readonly twentyConfigService: TwentyConfigService) {
+    const appSecret = this.twentyConfigService.get('APP_SECRET');
+
+    this.encryptionKey = Buffer.from(
+      crypto.hkdfSync('sha256', appSecret, '', HKDF_CONTEXT, 32),
+    );
+  }
```
Devessier: I'm unsure if we should use crypto sync methods. Cryptography is known to be slow and CPU intensive. Using raw level sync methods will block the whole event loop and make the app globally slower.

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
```diff
@@ -495,46 +495,58 @@ export class AuthService {
...
+    const isLocalhost =
+      parsedRedirectUrl.hostname === 'localhost' ||
+      parsedRedirectUrl.hostname === '127.0.0.1';
+
+    if (parsedRedirectUrl.protocol !== 'https:' && !isLocalhost) {
+      throw new AuthException(
+        'Redirect URL must use HTTPS',
+        AuthExceptionCode.FORBIDDEN_EXCEPTION,
+      );
+    }
```
Devessier: There is similar logic elsewhere in the PR, maybe should we unify.

packages/twenty-server/src/engine/core-modules/auth/services/oauth.service.ts
```diff
@@ -1,157 +1,386 @@
...
+  ): Promise<OAuthErrorResponse | null> {
+    const codeChallenge = crypto
+      .createHash('sha256')
+      .update(codeVerifier)
+      .digest()
+      .toString('base64')
+      .replace(/\+/g, '-')
+      .replace(/\//g, '_')
+      .replace(/=/g, '');
+
```
Devessier: Is it standard to replace these characters? If so, I would create a util properly named like `escapeBase64ForOauthCodeChallenge`

packages/twenty-server/src/engine/core-modules/application-registration/oauth.service.ts
```diff
@@ -1,157 +1,386 @@
...
+    }
+
+    await this.appTokenRepository.update(challengeToken.id, {
+      revokedAt: new Date(),
+    });
+
+    return null;
+  }
+
+  private async findOrInstallApplication(
```
Devessier, line 349: findOrInstallApplication contains "install". Here, it seems that we only create a record in the database. Should we rename the function?

packages/twenty-server/src/engine/core-modules/application-registration/oauth.service.ts
```diff
@@ -1,157 +1,386 @@
...
+
+    const userWorkspace = await this.userWorkspaceRepository.findOne({
+      where: {
+        userId: authCodeToken.userId,
+        workspaceId: authCodeToken.workspaceId,
+      },
+    });
+
+    const { applicationAccessToken, applicationRefreshToken } =
+      await this.applicationTokenService.generateApplicationTokenPair({
```
Devessier, line 159: In `this.applicationTokenService.generateApplicationTokenPair`, I'm unsure why we use `Promise.all` as both provided values aren't promises.

packages/twenty-server/src/engine/core-modules/application-registration/oauth.service.ts
```diff
@@ -1,157 +1,386 @@
...
+        workspaceId: authCodeToken.workspaceId,
+        applicationId: application.id,
+        userId: authCodeToken.userId,
+        userWorkspaceId: userWorkspace?.id,
+      });
+
```
Devessier, line 165: I don't understand the link between `OAUTH_ACCESS_TOKEN_EXPIRES_IN` and the validity duration of `applicationAccessToken` and `applicationRefreshToken`. They seem unrelated. I feel we should make it more explicit in the code.

packages/twenty-server/src/engine/core-modules/application-registration/oauth.service.ts
```diff
@@ -1,157 +1,386 @@
...
+        applicationId: application.id,
+      });
+
+    return {
+      access_token: applicationAccessToken.token,
+      token_type: 'Bearer',
+      expires_in: OAUTH_ACCESS_TOKEN_EXPIRES_IN,
+      scope: applicationRegistration.oAuthScopes.join(' '),
+    };
+  }
```
Devessier, line 230: Same thing here. It seems that there is no strong link between the validity duration of the token generated by `generateApplicationAccessToken` and `expires_in: OAUTH_ACCESS_TOKEN_EXPIRES_IN`.

packages/twenty-server/src/engine/core-modules/application-registration/oauth.service.ts
```diff
@@ -1,157 +1,386 @@
...
+        await this.applicationTokenService.renewApplicationTokens(payload);
+
+      return {
+        access_token: applicationAccessToken.token,
+        token_type: 'Bearer',
+        expires_in: OAUTH_ACCESS_TOKEN_EXPIRES_IN,
+        refresh_token: applicationRefreshToken.token,
+        scope: applicationRegistration.oAuthScopes.join(' '),
+      };
+    } catch (error) {
```
Devessier, line 274: Same question here about the expiration

Devessier (approved):
The strategy looks good to me. I left a few technical comments.

packages/twenty-server/src/engine/core-modules/application-registration/application-registration-encryption.service.ts
FelixMalfait: This should only be called once during the application startup

packages/twenty-server/src/engine/core-modules/application-registration/application-registration-encryption.service.ts
FelixMalfait: Good point, will check

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
FelixMalfait: Thanks, will do

packages/twenty-server/src/engine/core-modules/auth/services/oauth.service.ts
FelixMalfait: Yes! Thanks

packages/twenty-server/src/engine/core-modules/application-registration/oauth.service.ts
FelixMalfait: It was to avoid having a PR too big but the next step will be to do a real install soon / in a follow-up PR

packages/twenty-front/src/modules/settings/admin-panel/components/SettingsAdminTableCard.tsx
FelixMalfait: Done — moved to the styled component.

packages/twenty-front/src/pages/auth/Authorize.tsx
FelixMalfait: Good call — moved navigation to a `useEffect` so we're not triggering side effects during render.

packages/twenty-front/src/pages/auth/Authorize.tsx
FelixMalfait: Typed the keys explicitly as `{ [scope: string]: string | undefined }` — the scope strings come from the server dynamically so we need to handle unknown scopes gracefully with the `?? scope` fallback.

packages/twenty-sdk/src/cli/utilities/api/api-service.ts
FelixMalfait: Yeah the SDK API service methods are hand-written — just structured to follow the REST endpoints closely. Not auto-generated from OpenAPI or similar.

packages/twenty-server/src/engine/core-modules/application-registration/application-registration-encryption.service.ts
FelixMalfait: Good point. `SimpleSecretEncryptionUtil` uses `aes-256-cbc` (no authentication tag) while this one uses `aes-256-gcm` (authenticated encryption, stronger). They also derive keys differently — the TOTP service uses `JwtWrapperService` while this uses HKDF from `APP_SECRET`. Unifying would break existing encrypted 2FA secrets since the format/algorithm differs. I think this is worth a follow-up to converge on GCM for both, with a migration for existing 2FA data.

packages/twenty-server/src/engine/core-modules/application-registration/application-registration-encryption.service.ts
FelixMalfait: Agree — moved the `hkdfSync` key derivation to an async `onModuleInit` lifecycle hook. The remaining `createCipheriv`/`createDecipheriv` calls don't have async counterparts in Node.js crypto (you'd need Web Crypto `subtle` which is a very different API), but for small secrets the sync cipher operations are microseconds and won't meaningfully block the event loop.

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
FelixMalfait: Done — extracted a shared `validateRedirectUri` util that both `auth.service.ts` and `application-registration.service.ts` now use. Validates URL parsing, HTTPS requirement (except localhost), and no fragments per RFC 6749.

packages/twenty-server/src/engine/core-modules/auth/services/oauth.service.ts
FelixMalfait: Yes, this is RFC 4648 base64url encoding — standard in OAuth/OIDC for PKCE code challenges. Created a `base64UrlEncode` util and used it in both `oauth.service.ts` and the integration test.

packages/twenty-server/src/engine/core-modules/application-registration/oauth.service.ts
FelixMalfait: Good catch — `signApplicationToken` is synchronous (returns `AuthToken`, not `Promise<AuthToken>`), so the `Promise.all` was unnecessary. Removed it.

packages/twenty-server/src/engine/core-modules/application-registration/oauth.service.ts
FelixMalfait: You're right — `OAUTH_ACCESS_TOKEN_EXPIRES_IN` was a duplicate of the value used inside `ApplicationTokenService`. They need to be the same since the OAuth response `expires_in` must match the actual token lifetime. Removed the duplicate and moved both application token durations (`APPLICATION_ACCESS_TOKEN_EXPIRES_IN` and `APPLICATION_REFRESH_TOKEN_EXPIRES_IN`) to `config-variables.ts` alongside all other token durations. The `OAuthService` now reads the config and converts to seconds for the RFC 6749 response.

packages/twenty-server/src/engine/core-modules/application-registration/oauth.service.ts
FelixMalfait: Fixed — same as above, now using the shared config variable.

packages/twenty-server/src/engine/core-modules/application-registration/oauth.service.ts
FelixMalfait: Fixed — same as above.

packages/twenty-server/src/engine/core-modules/application-registration/application-registration-encryption.service.ts
FelixMalfait: OK AI disagreed apparently :p see above

packages/twenty-front/src/pages/auth/Authorize.tsx
```diff
@@ -55,53 +57,90 @@ const StyledButtonContainer = styled.div`
   flex-direction: row;
   gap: 10px;
   width: 100%;
+
+  > * {
```
charlesBochet: polish: we avoid this usually, is this useful here?

packages/twenty-front/src/pages/settings/applications/SettingsApplicationRegistrationDetails.tsx
```diff
@@ -0,0 +1,622 @@
...
+            will stop working. Please type {`"${confirmationValue}"`} to
+            confirm.
+          </Trans>
+        }
+        onConfirmClick={handleRotateSecret}
+        confirmButtonText={t`Rotate secret`}
+        loading={isLoading}
+      />
+
+      <ConfirmationModal
```
charlesBochet, line 608: polish: component is a bit long but acceptable

packages/twenty-sdk/src/cli/__tests__/integration/utils/run-app-dev-in-process.util.ts
```diff
@@ -31,7 +32,20 @@ export const runAppDevInProcess = async (options: {
...
+  const stepStatuses = state
+    ? Object.fromEntries(
+        Object.entries(state.steps).map(([key, step]) => [key, step.status]),
+      )
+    : undefined;
+
   await command.close();
 
-  return { success: false };
+  return { success: false, events, stepStatuses };
```
charlesBochet, line 50: good idea

packages/twenty-server/src/database/typeorm/core/migrations/common/1772000000000-create-application-registration.ts
```diff
@@ -0,0 +1,143 @@
...
+    await queryRunner.query(`
+      CREATE TABLE "core"."applicationRegistration" (
+        "id" uuid NOT NULL DEFAULT uuid_generate_v4(),
+        "universalIdentifier" uuid NOT NULL,
+        "name" text NOT NULL,
+        "description" text,
+        "logoUrl" text,
+        "author" text,
+        "oAuthClientId" text NOT NULL,
+        "oAuthClientSecretHash" text,
```
charlesBochet, line 18: NOT NULL?

packages/twenty-server/src/database/typeorm/core/migrations/common/1772000000000-create-application-registration.ts
```diff
@@ -0,0 +1,143 @@
+import { type MigrationInterface, type QueryRunner } from 'typeorm';
+
+export class CreateApplicationRegistration1772000000000
```
charlesBochet: this looks not to be generated by generate migration command (checking timestamp)

This is containing non standard column types, I think we are usually using character varying and not text

packages/twenty-server/src/engine/core-modules/application-registration/__tests__/application-registration-encryption.service.spec.ts
```diff
@@ -0,0 +1,58 @@
+import { Test, type TestingModule } from '@nestjs/testing';
+
+import { ApplicationRegistrationEncryptionService } from 'src/engine/core-modules/application-registration/application-registration-encryption.service';
+import { TwentyConfigService } from 'src/engine/core-modules/twenty-config/twenty-config.service';
+
+describe('ApplicationRegistrationEncryptionService', () => {
```
charlesBochet: no need to generate test on top of services anymore, they don't bring value, we only test utils / zod schemas (anything containing business logic with low dependencies) unitary

packages/twenty-server/src/engine/core-modules/application-registration/__tests__/application-registration-identifier-guard.service.spec.ts
```diff
@@ -0,0 +1,114 @@
+import { Test, type TestingModule } from '@nestjs/testing';
+import { getRepositoryToken } from '@nestjs/typeorm';
+
+import { type Repository } from 'typeorm';
+
+import { ApplicationRegistrationEntity } from 'src/engine/core-modules/application-registration/application-registration.entity';
+import { ApplicationRegistrationIdentifierGuardService } from 'src/engine/core-modules/application-registration/application-registration-identifier-guard.service';
+import { ApplicationEntity } from 'src/engine/core-modules/application/application.entity';
+
+describe('ApplicationRegistrationIdentifierGuardService', () => {
```
charlesBochet: same

packages/twenty-server/src/engine/core-modules/application-registration/__tests__/application-registration.service.spec.ts
```diff
@@ -0,0 +1,327 @@
...
+import { type Repository } from 'typeorm';
+
+import { ApplicationRegistrationVariableEntity } from 'src/engine/core-modules/application-registration/application-registration-variable.entity';
+import { ApplicationRegistrationEncryptionService } from 'src/engine/core-modules/application-registration/application-registration-encryption.service';
+import { ApplicationRegistrationEntity } from 'src/engine/core-modules/application-registration/application-registration.entity';
+import { ApplicationRegistrationExceptionCode } from 'src/engine/core-modules/application-registration/application-registration.exception';
+import { ApplicationRegistrationService } from 'src/engine/core-modules/application-registration/application-registration.service';
+import { ApplicationEntity } from 'src/engine/core-modules/application/application.entity';
+
+describe('ApplicationRegistrationService', () => {
```
charlesBochet: same

packages/twenty-server/src/engine/core-modules/application-registration/__tests__/oauth-scopes.spec.ts
```diff
@@ -0,0 +1,25 @@
+import {
+  ALL_OAUTH_SCOPES,
+  OAUTH_SCOPE_DESCRIPTIONS,
+  OAUTH_SCOPES,
+} from 'src/engine/core-modules/application-registration/constants/oauth-scopes';
+
+describe('OAuth Scopes', () => {
```
charlesBochet, line 7: here it makes sense

packages/twenty-server/src/engine/core-modules/application-registration/dtos/application-registration-stats.output.ts
```diff
@@ -0,0 +1,15 @@
+import { Field, Int, ObjectType } from '@nestjs/graphql';
+
+import { VersionDistributionEntry } from 'src/engine/core-modules/application-registration/dtos/version-distribution-entry.output';
+
+@ObjectType()
+export class ApplicationRegistrationStatsOutput {
```
charlesBochet: I don't remember, do we use Output or DTO naming for output?

packages/twenty-server/src/engine/core-modules/application-registration/dtos/update-application-registration-variable.input.ts
```diff
@@ -0,0 +1,22 @@
...
+export class UpdateApplicationRegistrationVariableInput {
+  @Field()
+  @IsUUID()
+  id: string;
+
+  @Field({ nullable: true })
+  @IsString()
+  @MaxLength(10000)
+  @IsOptional()
+  value?: string;
```
charlesBochet, line 19: for update, we follow the pattern:
```
@Type(() => UpdateObjectPayload)
  @ValidateNested()
  @Field(() => UpdateObjectPayload)
  update: UpdateObjectPayload;

  @IsNotEmpty()
  @Field(() => UUIDScalarType, {
    description: 'The id of the object to update',
  })
  @IsUUID()
  id!: string;
```

packages/twenty-server/src/engine/core-modules/application-registration/dtos/update-application-registration.input.ts
```diff
@@ -0,0 +1,69 @@
...
+  IsString,
+  IsUUID,
+  MaxLength,
+} from 'class-validator';
+
+@InputType()
+export class UpdateApplicationRegistrationInput {
+  @Field()
+  @IsUUID()
+  id: string;
```
charlesBochet: same

packages/twenty-server/src/engine/core-modules/application-registration/application-registration-encryption.service.ts
```diff
@@ -0,0 +1,78 @@
...
+import crypto from 'crypto';
+
+import { TwentyConfigService } from 'src/engine/core-modules/twenty-config/twenty-config.service';
+
+const ALGORITHM = 'aes-256-gcm';
+const IV_LENGTH = 16;
+const AUTH_TAG_LENGTH = 16;
+const HKDF_CONTEXT = 'application-registration-variable';
+
+const hkdf = promisify(crypto.hkdf);
```
charlesBochet: do we need that, I'm surprised, first time seeing it

charlesBochet: overall, I'm surprised that we are doing this low level logic in the code, is that the right way?

packages/twenty-server/src/engine/core-modules/application-registration/application-registration-identifier-guard.service.ts
```diff
@@ -0,0 +1,55 @@
...
+import { InjectRepository } from '@nestjs/typeorm';
+
+import { isDefined } from 'twenty-shared/utils';
+import { Repository } from 'typeorm';
+
+import { ApplicationRegistrationEntity } from 'src/engine/core-modules/application-registration/application-registration.entity';
+import { ApplicationEntity } from 'src/engine/core-modules/application/application.entity';
+
+@Injectable()
+export class ApplicationRegistrationIdentifierGuardService {
```
charlesBochet: is that standard naming guardService guard.service?
we usually keep guard naming for nest guards

packages/twenty-server/src/engine/core-modules/application-registration/application-registration-identifier-validation.service.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    if (!isDefined(ownerRegistration)) {
+      return { isValid: true };
+    }
+
+    if (ownerRegistration.id !== application.applicationRegistrationId) {
+      return {
+        isValid: false,
+        errorMessage:
+          `Universal identifier '${params.universalIdentifier}' is claimed by ` +
```
charlesBochet: is it translated by lingui higher in the stack?

packages/twenty-server/src/engine/core-modules/application-registration/application-registration-variable-resolution.service.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    if (isDefined(workspaceVariable)) {
+      if (workspaceVariable.isSecret) {
+        return this.secretEncryptionService.decrypt(workspaceVariable.value);
+      }
+
+      return workspaceVariable.value;
+    }
+
+    const application = await this.applicationRepository.findOne({
```
charlesBochet: nitpick: I believe we have caches for that. Not sure we have one for application variable but we might want. This is lowering the load on the database

i guess this resolveVariable will have high QPS

packages/twenty-server/src/engine/core-modules/application-registration/application-registration-variable.entity.ts
```diff
@@ -0,0 +1,74 @@
...
+  @Field()
+  @Column({ nullable: false, type: 'boolean', default: true })
+  isSecret: boolean;
+
+  @Field()
+  @Column({ nullable: false, type: 'boolean', default: false })
+  isRequired: boolean;
+
+  @Field()
+  get isFilled(): boolean {
```
charlesBochet, line 52: non standard but interesting!

packages/twenty-server/src/engine/core-modules/application-registration/application-registration.entity.ts
```diff
@@ -0,0 +1,113 @@
...
+  @Field()
+  @CreateDateColumn({ type: 'timestamptz' })
+  createdAt: Date;
+
+  @Field()
+  @UpdateDateColumn({ type: 'timestamptz' })
+  updatedAt: Date;
+
+  @DeleteDateColumn({ type: 'timestamptz' })
+  deletedAt: Date | null;
```
charlesBochet, line 112: needed?

packages/twenty-server/src/engine/core-modules/application-registration/application-registration.resolver.ts
```diff
@@ -0,0 +1,179 @@
...
+    @Args('id') id: string,
+  ): Promise<ApplicationRegistrationStatsOutput> {
+    return this.applicationRegistrationService.getStats(id);
+  }
+
+  @UseGuards(WorkspaceAuthGuard, NoPermissionGuard)
+  @Mutation(() => CreateApplicationRegistrationOutput)
+  async createApplicationRegistration(
+    @Args('input') input: CreateApplicationRegistrationInput,
+    @AuthUser({ allowUndefined: true }) user: UserEntity | undefined,
```
charlesBochet, line 95: do we allow undefined?

packages/twenty-server/src/engine/core-modules/application-registration/application-registration.service.ts
```diff
@@ -0,0 +1,391 @@
...
+
+    if (isDefined(input.oAuthRedirectUris)) {
+      this.validateRedirectUris(input.oAuthRedirectUris);
+    }
+
+    if (isDefined(input.oAuthScopes)) {
+      this.validateScopes(input.oAuthScopes);
+    }
+
+    const updateData: QueryDeepPartialEntity<ApplicationRegistrationEntity> =
```
charlesBochet: nitpick: this code is non great as we have side effects on const, could lead to bad patterns. Here it's very local so we should be fine but opening the door to bad coding habits

packages/twenty-server/src/engine/core-modules/application-registration/application-registration.service.ts
```diff
@@ -0,0 +1,391 @@
...
+    clientSecret: string,
+  ): Promise<boolean> {
+    if (!registration.oAuthClientSecretHash) {
+      return false;
+    }
+
+    return bcrypt.compare(clientSecret, registration.oAuthClientSecretHash);
+  }
+
+  // Variable operations
```
charlesBochet: make another service (this one is long)

packages/twenty-server/src/engine/core-modules/application-registration/application-registration.service.ts
```diff
@@ -0,0 +1,391 @@
...
+      await this.variableRepository.delete({ applicationRegistrationId });
+    }
+  }
+
+  async getStats(
+    applicationRegistrationId: string,
+  ): Promise<ApplicationRegistrationStatsOutput> {
+    await this.findOneById(applicationRegistrationId);
+
+    const installs = await this.applicationRepository.find({
```
charlesBochet: we should leverage cache here

packages/twenty-server/src/engine/core-modules/application/services/application-sync.service.ts
```diff
@@ -118,13 +120,61 @@ export class ApplicationSyncService {
       },
     );
 
+    let applicationRegistrationId = application.applicationRegistrationId;
```
charlesBochet: why let?

packages/twenty-server/src/engine/core-modules/application/application.entity.ts
```diff
@@ -94,6 +96,16 @@ export class ApplicationEntity extends WorkspaceRelatedEntity {
   @Column({ nullable: false, type: 'boolean', default: true })
   canBeUninstalled: boolean;
 
+  @Column({ nullable: true, type: 'uuid' })
+  applicationRegistrationId: string | null;
+
+  @ManyToOne(() => ApplicationRegistrationEntity, {
+    onDelete: 'SET NULL',
```
charlesBochet, line 103: don't we want to restrict here?

packages/twenty-server/src/engine/core-modules/application-registration/controllers/oauth-discovery.controller.ts
```diff
@@ -0,0 +1,32 @@
+import { Controller, Get, UseGuards } from '@nestjs/common';
+
+import { ALL_OAUTH_SCOPES } from 'src/engine/core-modules/application-registration/constants/oauth-scopes';
+import { TwentyConfigService } from 'src/engine/core-modules/twenty-config/twenty-config.service';
+import { NoPermissionGuard } from 'src/engine/guards/no-permission.guard';
+import { PublicEndpointGuard } from 'src/engine/guards/public-endpoint.guard';
+
+@Controller('.well-known')
```
charlesBochet, line 8: is that standard?

packages/twenty-server/src/engine/core-modules/auth/controllers/oauth-token.controller.ts
```diff
@@ -0,0 +1,111 @@
...
+import { IsOptional, IsString, MaxLength } from 'class-validator';
+
+import { AuthRestApiExceptionFilter } from 'src/engine/core-modules/auth/filters/auth-rest-api-exception.filter';
+import { OAuthService } from 'src/engine/core-modules/auth/services/oauth.service';
+import { OAuthErrorResponse } from 'src/engine/core-modules/auth/types/oauth-error-response.type';
+import { OAuthTokenResponse } from 'src/engine/core-modules/auth/types/oauth-token-response.type';
+import { NoPermissionGuard } from 'src/engine/guards/no-permission.guard';
+import { PublicEndpointGuard } from 'src/engine/guards/public-endpoint.guard';
+
+export class OAuthTokenRequestDto {
```
charlesBochet: looks non standard + should be in its own file

packages/twenty-server/src/engine/core-modules/application-registration/oauth.service.ts
```diff
@@ -1,157 +1,392 @@
...
+    });
+
+    if (!authCodeToken.userId || !authCodeToken.workspaceId) {
+      return this.errorResponse(
+        'server_error',
+        'Authorization code is missing user or workspace context',
+      );
+    }
+
+    const application = await this.findOrInstallApplication(
```
charlesBochet, line 146: I'm confused by this, we auto install on oAuth?

packages/twenty-server/src/engine/core-modules/application-registration/oauth.service.ts
```diff
@@ -1,157 +1,392 @@
...
+      description: applicationRegistration.description,
+      version: '0.0.0',
+      sourcePath: 'oauth-install',
+      applicationRegistrationId: applicationRegistration.id,
+      workspaceId,
+    });
+  }
+
+  // OAuth RFC 6749 requires expires_in as seconds
+  private getAccessTokenExpiresInSeconds(): number {
```
charlesBochet, line 381: service is a bit long but acceptable. Looks to be doing a lot. Also I wonder if this should leave under applications rather than auth

packages/twenty-server/src/engine/core-modules/auth/strategies/jwt.auth.strategy.ts
```diff
@@ -359,12 +359,36 @@ export class JwtAuthStrategy extends PassportStrategy(Strategy, 'jwt') {
...
-    return {
-      application,
-      workspace,
-    };
+    const context: AuthContext = { application, workspace };
+
+    // When the token carries user context (e.g. OAuth authorization code flow),
+    // populate user fields so downstream resolvers can identify the acting user.
+    if (payload.userId) {
+      const user = await this.userRepository.findOne({
```
charlesBochet: change: this is bad, this will impact performances. I believe we already have an hydratation logic, i don't remember where exactly but we should uniformize

We should not hydrate twice and it should be in a single place in the code

packages/twenty-server/src/engine/core-modules/auth/strategies/jwt.auth.strategy.ts
```diff
@@ -359,12 +359,36 @@ export class JwtAuthStrategy extends PassportStrategy(Strategy, 'jwt') {
...
+        where: { id: payload.userId },
+      });
+
+      if (isDefined(user)) {
+        context.user = user;
+      }
+    }
+
+    if (payload.userWorkspaceId) {
+      const userWorkspace = await this.userWorkspaceRepository.findOne({
```
charlesBochet: same

packages/twenty-server/src/engine/core-modules/auth/token/services/application-token.service.ts
```diff
@@ -82,27 +84,31 @@ export class ApplicationTokenService {
...
-        this.signApplicationToken({
-          workspaceId,
-          applicationId,
-          userWorkspaceId,
-          userId,
-          tokenType: JwtTokenTypeEnum.APPLICATION_REFRESH,
-          expiresInSeconds: APPLICATION_REFRESH_TOKEN_EXPIRY_SECONDS,
-        }),
-      ],
+    const accessTokenExpiresIn = this.twentyConfigService.get(
```
charlesBochet, line 87: I disagree with this:
- for logicFunction we will provide a token that is long lived enough for the logicFunction to execute
- for frontComponent we should respect APPLICATION_ACCESS_TOKEN_EXPIRES_IN

packages/twenty-shared/src/utils/base64UrlEncode.ts
```diff
@@ -0,0 +1,7 @@
+// RFC 4648 base64url encoding (URL-safe, no padding)
+export const base64UrlEncode = (buffer: Buffer): string =>
```
charlesBochet, line 2: first time we need it? likely should be moved to shared

packages/twenty-server/src/engine/core-modules/auth/utils/validate-redirect-uri.util.ts
```diff
@@ -0,0 +1,31 @@
+// RFC 6749 redirect URI validation: must be absolute, HTTPS (except localhost), no fragments
+export const validateRedirectUri = (
```
charlesBochet, line 2: not the first time i see this, likely duplicated. Very surprising this is missing at this stage of the project

charlesBochet (approved):
Overall, very high quality :)
Left comments. 

We would also benefit from integration tests on registration and variables

packages/twenty-server/src/database/typeorm/core/migrations/common/1772000000000-create-application-registration.ts
FelixMalfait: Intentionally nullable — public OAuth clients (PKCE-only flow) don't have a client secret, so `oAuthClientSecretHash` must be nullable for that use case.

packages/twenty-server/src/engine/core-modules/application-registration/__tests__/oauth-scopes.spec.ts
FelixMalfait: Agreed — kept this one since it tests the scope validation logic directly. Removed the other three service-level tests.

packages/twenty-server/src/engine/core-modules/application-registration/application-registration.resolver.ts
FelixMalfait: CLI = Api key = no user ; there was also a path where it was created by UI so this is more legacy from this (also maybe the createdByUserId isn't so relevant anymore?). Maybe if we add OAuth login to the CLI in the future?

packages/twenty-server/src/database/typeorm/core/migrations/common/1772000000000-create-application-registration.ts
FelixMalfait: Thanks will address

packages/twenty-server/src/engine/core-modules/application-registration/dtos/application-registration-stats.output.ts
FelixMalfait: I cleaned up everything again (~100 misnamed input/output/dto files, with different type of inconsistencies). Looks like AI prefers output to DTO but we had picked DTO (what NestJS uses).

packages/twenty-server/src/engine/core-modules/application-registration/application-registration-encryption.service.ts
FelixMalfait: You're right we should reuse SecretEncryptionService which we already have, I'll try to merge the two

packages/twenty-server/src/engine/core-modules/application-registration/controllers/oauth-discovery.controller.ts
FelixMalfait: Yes!

packages/twenty-server/src/engine/core-modules/application-registration/oauth.service.ts
FelixMalfait: We will in a followup PR yes. That's what Shopify does I believe. It makes thing more clear. 
For Salesforce authorizing a ConnectedApp != installing a ManagedPackage, but for end-users I think it's best to couple the two (but developers can always create a ConnectedApp with an empty ManagedPackage or a ManagedPackage that is never used to access Twenty data from outside)

packages/twenty-server/src/engine/core-modules/application/application.entity.ts
FelixMalfait: I'd say yes in the future but probably too early, would need backfilling / to put more thoughts in this

packages/twenty-front/src/pages/auth/Authorize.tsx
FelixMalfait: Fixed

packages/twenty-server/src/engine/core-modules/application-registration/application-registration.entity.ts
FelixMalfait: Not a strong conviction but we have it on some other core entities and since it's sensitive I though it's good to keep an archive/soft delete

packages/twenty-server/src/engine/core-modules/application-registration/application-registration-identifier-guard.service.ts
FelixMalfait: Fixed

FelixMalfait:
@claude fix both typecheck and lint issues in this branch (both on twenty-server and front affected). Push to this branch directly

packages/twenty-server/src/engine/core-modules/application-registration/oauth.service.ts
FelixMalfait: Moved to applications! I agree it's better suited

packages/twenty-server/src/engine/core-modules/auth/strategies/jwt.auth.strategy.ts
FelixMalfait: Indeeed 🤦‍♂️

packages/twenty-server/src/engine/core-modules/auth/strategies/jwt.auth.strategy.ts
FelixMalfait: Indeeed 🤦‍♂️

packages/twenty-server/src/engine/core-modules/auth/token/services/application-token.service.ts
FelixMalfait: I did not change this because I feel the values currently are ok but I agree it does not need to be correlated in the future, we can always introduce a different variable if we need to

PR #18272, opened 2026-02-26
Builder runs delete update and then create

prastoin:
# Introduction
We need to build and validate the flat entity operation in the following order delete update and create
For example if not, if a created field has the same name than a deleted one than it will fail whereas it should not


packages/twenty-server/test/integration/metadata/suites/application/successful-sync-application-workspace-migration.integration-spec.ts
```diff
@@ -177,6 +177,67 @@ describe('syncApplication', () => {
     });
   }, 60000);
 
+  it.only('should delete old field and create equivalent one when field universalIdentifier changes', async () => {
```
Weiko: probably a left-over?

Weiko (approved):
LGTM (need to "fix" the integration test first though) 👍

martmull (approved):
Dis you check the test you have added is red without the change in the validator? If yes lgtm

prastoin:
@martmull yes integration test 8 has been red for few days

packages/twenty-server/test/integration/metadata/suites/application/successful-sync-application-workspace-migration.integration-spec.ts
prastoin: thanks

prastoin: I'll add an eslint rule against this

packages/twenty-server/src/engine/workspace-manager/workspace-migration/workspace-migration-builder/services/workspace-entity-migration-builder.service.ts
```diff
@@ -310,6 +238,80 @@ export abstract class WorkspaceEntityMigrationBuilderService<
       `EntityBuilder ${this.metadataName}`,
       'update validation',
     );
+
+    const remainingFlatEntityMapsToCreate = structuredClone(
```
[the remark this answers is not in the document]
prastoin: That's right

PR #18275, opened 2026-02-26
Fix ai agent node prompt and variables

thomtrp:
- prompt stored on workflow lvl so input variables can be resolved and it can evolves with versions
- make ai agent node output available as variables


bosiraphael (approved):
LGTM

charlesBochet (approved):

PR #18292, opened 2026-02-27
Fix self host application

martmull:
- Fixes self host application
- add new telemetry information
- add serverId to identify a server instance
- remove .twenty from git tracking
- tree-shake "twenty-sdk" usage in built logic functions and front components
- fix "twenty-sdk" version usage
- fix twenty-zapier cli


Weiko (commented):
@martmull we shouldn't push .output folder imho or am I missing something?

.github/workflows/ci-zapier.yaml
```diff
@@ -23,14 +24,81 @@ jobs:
...
         !packages/twenty-zapier/CHANGELOG.md
-  zapier-test:
+  server-setup:
     needs: changed-files-check
     if: needs.changed-files-check.outputs.any_changed == 'true'
     timeout-minutes: 30
+    runs-on: depot-ubuntu-24.04-8
+    services:
+      postgres:
+        image: twentycrm/twenty-postgres-spilo
```
prastoin, line 34: ***Polish:*** We should not be using spilo anymore

packages/twenty-apps/internal/self-hosting/tsconfig.json
```diff
@@ -5,13 +5,14 @@
...
     "rootDir": ".",
+    "jsx": "react-jsx",
     "moduleResolution": "node",
     "allowSyntheticDefaultImports": true,
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "importHelpers": true,
     "allowUnreachableCode": false,
-    "strictNullChecks": true,
+    "strict": true,
```
prastoin, line 15: Image

packages/twenty-shared/src/database-events/database-event-payload.type.ts
prastoin, line 3: ***Question:*** Now looks like a complete `FlatObjectMetadata` is it expected ?

prastoin (approved):
Looks good to me !

packages/twenty-shared/src/database-events/database-event-payload.type.ts
martmull: checking

.github/workflows/ci-zapier.yaml
martmull: what should we use instead?

packages/twenty-shared/src/database-events/database-event-payload.type.ts
martmull: FlatObjectMetadata is defined in twenty-server, that is a big move that deserves its own pr i think

.github/workflows/ci-zapier.yaml
martmull: all ci uses twentycrm/twenty-postgres-spilo so i guess this is not relevant to change that in this PR

prastoin: Should be postgres 16 about to change that for server in a following pr

packages/twenty-shared/src/database-events/database-event-payload.type.ts
prastoin: Indeed my bad it's in twenty shared

prastoin: Do we need the ids aggregator outside of the server though ?

Weiko (approved):

PR #18297, opened 2026-02-27
fix: [Note] Title not filled by default #13838

### LINKED ISSUE #13838: [Note] Title not filled by default
## Bug Description

When creating a Note (using the Add note button from the tab - I am on the opportunity object) and I start typing, the text is entered in the body and not the title.
It feels weird as a user, I'd have assumed to start entering the title - and I'm pretty sure this was the behavior a few weeks ago.

bugisthegod:
Fixes #13838 
When creating a note from the command menu side panel (e.g. clicking "Add Note" in a related notes section on an Opportunity/company/people page), the title field was not auto-focused — focus point went to body instead.

## Root Cause

When a record opens in the side panel, there is no page navigation, so `PageChangeEffect` (which handles title auto-focus for full-page views) never runs. `openNewRecordTitleCell()` was simply never called for the side-panel path.

## Fix

`openRecordInCommandMenu` is the single entry point for all side-panel record opens, so title auto-focus is handled there once for all callers.
Previously, `useCreateNewIndexRecord` called `openRecordInCommandMenu` and then called `openNewRecordTitleCell` separately, which would have caused a double invocation after this fix. The redundant call has been removed.

## Before

https://github.com/user-attachments/assets/df0d9e4f-dc25-4a0d-a49e-898a14f9c0a0

## After

https://github.com/user-attachments/assets/1a5044f7-6bb7-4333-8934-c1081b935e97


etiennejouan (approved):
LGTM ! Thank you for your contribution !

Issue #18298, opened 2026-02-27
Bug: Unable to delete files once upload - AI chatbot


LTan-101104:
## Bug Description

In the upload functionality for AI chatbot, once a file is uploaded, it cannot be deleted even though trying to click 'X' icon.
Environment: local, self-hosted
See attached video for better description:

https://github.com/user-attachments/assets/55ef43a9-53b9-48ec-b5be-b8136c176b64

## Expected behavior

When click the 'X', files should be removed/ disappear from the chat instead of directing to its URL.

## Technical inputs

Very likely because of the nested component of this file preview UI component. The 'X' icon is nested within a ```StyledLink``` component and thus remove event cannot be triggered properly.

LTan-101104:
Pushed PR, ready for review.

PR #18299, opened 2026-02-27
Allow files uploaded in AI chat box to be removed

LTan-101104:
This aims to fix the bug mentioned in #18298

Additionally, after removing a file and attempting to re-upload the same file, the file preview does not appear. This is because the browser's <input type="file"> only fires onChange when the value changes — if the same file is selected again,  the value (C:\fakepath\<filename>) remains unchanged, so the event never fires.

##Summary##

Fixed at the twenty-ui Chip component level rather than in consumer code. The renderRightComponent function in Chip.tsx now wraps the right component in a container that calls event.stopPropagation(), preventing clicks from bubbling to parent handlers. This ensures any Chip consumer with an interactive rightComponent gets correct behavior automatically.

For the 2nd issue, reset event.target.value to '' in AgentChatFileUploadButton after processing the selected files. This clears the browser's stored fake path so that selecting the same file again is treated as a new change and triggers onChange.


charlesBochet:
@LTan-101104 Thanks for opening an issue and a PR
This needs to be adressed at twenty-ui level as having Chips with rightComponent (here x Icon) is a UI component we should have.

Investigating your fix, I figured out that we needed to refactor a bit our Chip system first: https://github.com/twentyhq/twenty/pull/18313

I'll merge it shortly. Then, please work at twenty-ui level (let's use twenty-ui storybook taht you can serve through npx nx run twenty-ui:storybook:serve:dev

Once you are happy with the behavior, just use it in the Chat file section

LTan-101104:
Hi @charlesBochet, thank you for your review. I will look into this and reimplement once your PR is merged.

charlesBochet:
@LTan-101104 PR has been merged, could you take another look?

LTan-101104:
@charlesBochet Sure, will ping you when I'm done.

LTan-101104:
@charlesBochet Done, modify behavior in UI level. Also I identify another small problem, if you remove the file and then reupload it won't show the file in the preview and I have made changes to fix it. Please let me know what you think.

charlesBochet:
@LTan-101104 Thanks for opening a PR, we have actually gone with the following fix: https://github.com/twentyhq/twenty/pull/18779

I'm going to close this PR but appreciate the help :)

PR #18301, opened 2026-02-28
Fix: AI chat file attachment X button navigates instead of removing file

RahulXDTT:
Fixes a bug where clicking the X (remove) button on a file attachment in the AI chat redirected to the file URL instead of removing the file.

Current behavior
When a file is uploaded and attached to the AI chatbot, clicking the X icon to remove it navigates the browser to the file's token URL (e.g., /files/agent-chat/eyJhbGci...) instead of removing the file from the chat input.

Root cause
In AgentChatFilePreview, when a file has a URL (i.e., it's been uploaded and is a FileUIPart), the component renders a LinkChip. LinkChip wraps all its children — including the X remove button — inside a <Link> element. The <Link> uses useMouseDownNavigation which triggers navigation on mousedown (not click), so the link navigation fires before the remove button's onClick ever reaches, making the X button non-functional.

Fix
When onRemove is provided and the file has a URL, render a non-navigating Chip instead of LinkChip. This avoids the nested link problem entirely.
Wrap the remove AvatarChip in a container that intercepts both mousedown and click with stopPropagation() + preventDefault() for safety.
When no onRemove is passed (e.g., files in message history that shouldn't be removable), LinkChip is still used so the file remains clickable.

File changed
AgentChatFilePreview.tsx


RahulXDTT:
@LTan-101104 , Just waiting for your review, I have also attached the video file which demonstrates the fix.

https://github.com/user-attachments/assets/13f63417-bacc-4823-bfff-f9d0921aea63

LTan-101104:
@RahulXDTT Thank you for your work, but for the issue I have already opened a PR to fix it and am waiting for the core team to review it.

RahulXDTT:
@LTan-101104  Oh Got it. No worries. I actually did not check your PR for the fix. Thanks for the response.

charlesBochet:
Closing this one, thanks for contributing @RahulXDTT

PR #18304, opened 2026-02-28
A few fixes on table performance

charlesBochet:
## RecordTable Performance Investigation & Optimization (WIP)

Investigates what makes the RecordTable slow (14 components, ~12 hooks per cell) and starts applying fixes.

### Key Findings (2,000 cells benchmark)

Cell Render Benchmark

State Access Benchmark

- **Jotai atoms are the dominant cost**: 10 atom reads/cell = +312%. Full sim with atoms = +476%.
- **Derived atoms are 3x cheaper** than individual reads (12 sources: +93% vs +294%).
- **Component depth is expensive**: 4-level nesting = +82%, 14 wrappers = +109%.
- **Styling engines are comparable**: Linaria vs Emotion is within noise.
- **Context reads and useState are nearly free** vs baseline.

### Optimizations Applied

1. **Static focus providers** — replaced per-cell `useState(false)` with static context. Eliminates 400 useState instances.
2. **Delegated onMouseMove** — single handler on table body instead of 400 per-cell handlers.
3. **Hoisted `useObjectMetadataItems()`** — moved from per-cell to table-level context. Eliminates 400 global atom reads.

### Tooling

- Perf page at `/__perf__/table`: 17 cell render + 13 state access benchmarks
- Render profiler: `window.__RECORD_TABLE_PROFILE = true`
- Full plan in `__perf__/PERFORMANCE_PLAN.md`

### Remaining Phases (not high priority to-be-honest)

| Phase | What | Status |
|-------|------|--------|
| 1 | Separate display from interaction | Partial |
| 2 | Flatten hierarchy (14 → ~5 components/cell) | TODO |
| 3 | Reduce atom reads per cell | Partial |
| 4 | CSS-only hover/focus | TODO |
| 5 | Event delegation, lazy Draggable | TODO |


packages/twenty-front/src/modules/app/hooks/useCreateAppRouter.tsx
```diff
@@ -34,12 +34,13 @@ export const useCreateAppRouter = (
 ) =>
   createBrowserRouter(
     createRoutesFromElements(
-      <Route
-        element={<AppRouterProviders />}
-        // To switch state to `loading` temporarily to enable us
-        // to set scroll position before the page is rendered
-        loader={async () => Promise.resolve(null)}
-      >
+      <>
```
charlesBochet: let's revert this

packages/twenty-front/src/modules/object-record/record-field/ui/contexts/FieldFocusContextProvider.tsx
```diff
@@ -2,6 +2,19 @@ import { useState } from 'react';
...
+  isFocused: false,
+  setIsFocused: NOOP,
+};
+
+const STATIC_FOCUSED_VALUE = {
+  isFocused: true,
+  setIsFocused: NOOP,
+};
+
+// Stateful provider for contexts that need dynamic focus (inline cells, etc.)
```
charlesBochet: remove comment

packages/twenty-front/src/modules/object-record/record-field/ui/contexts/FieldFocusContextProvider.tsx
```diff
@@ -2,6 +2,19 @@ import { useState } from 'react';
 
 import { FieldFocusContext } from '@/object-record/record-field/ui/contexts/FieldFocusContext';
 
+const NOOP = () => {};
```
charlesBochet: useful?

packages/twenty-front/src/modules/object-record/record-table/components/RecordTableContent.tsx
```diff
@@ -91,13 +92,48 @@ export const RecordTableContent = ({
     }
   }, [store, isSomeCellInEditMode, setRecordTableHoverPosition]);
 
+  const lastHoverPositionRef = useRef<TableCellPosition | null>(null);
```
charlesBochet: can we re-implement this without using useRef. Also, we should leverage html data-attributes rather than extracting from id

FelixMalfait (approved):

PR #18308, opened 2026-02-28
feat: remember original URL and redirect after login

FelixMalfait:
## Summary

- Implement a return-to-path mechanism that preserves the user's intended destination across authentication flows (login, magic link, cross-domain redirects)
- Uses layered persistence: Jotai atom (in-memory), sessionStorage with TTL (tab-switch resilience), URL query parameter (cross-domain propagation)
- Includes path validation to prevent open redirects, automatic cleanup after successful login, and comprehensive test coverage
- Replaces the unused `previousUrlState` with a robust `returnToPathState` system

## Test plan

- [ ] Visit a deep link (e.g. `/objects/tasks`) while logged out — should redirect to login, then back to `/objects/tasks` after logging in
- [ ] Visit an OAuth authorize link while logged out — should redirect to login, then to the authorize page
- [ ] Test magic link flow: click sign-in link that opens new tab — should still redirect to original destination
- [ ] Test cross-domain: visit `app.twenty.com/objects/tasks` — should preserve path through workspace domain redirect
- [ ] Verify auth/onboarding paths are excluded from being saved as return paths
- [ ] Verify return-to-path is cleared after successful navigation
- [ ] All 215 existing `usePageChangeEffectNavigateLocation` tests pass


packages/twenty-front/src/modules/auth/hooks/useReturnToPath.ts
```diff
@@ -0,0 +1,57 @@
...
+  const setReturnToPath = useSetAtomState(returnToPathState);
+
+  const saveReturnToPath = useCallback(
+    (path: string) => {
+      if (!isValidReturnToPath(path)) {
+        return;
+      }
+
+      setReturnToPath(path);
+      writeReturnToPathToSessionStorage(path);
```
charlesBochet: this should be handled by our atom api (we arelady write to localStorage and cookies, should be the same here). Btw, why not writing to localStorage here?

packages/twenty-front/src/modules/auth/utils/readReturnToPathFromSessionStorage.ts
```diff
@@ -0,0 +1,29 @@
+import { RETURN_TO_PATH_SESSION_KEY } from '@/auth/constants/ReturnToPathSessionKey';
+import { RETURN_TO_PATH_TTL_MS } from '@/auth/constants/ReturnToPathTtl';
+import { isReturnToPathSessionStorageEntry } from '@/auth/utils/isReturnToPathSessionStorageEntry';
+
+export const readReturnToPathFromSessionStorage = (): string | null => {
```
charlesBochet: I wonder if all this could be avoid with proper atom tooling

packages/twenty-front/src/modules/auth/utils/isValidReturnToPath.ts
```diff
@@ -0,0 +1,24 @@
...
+
+const extractPathPrefix = (appPath: string): string => appPath.split('/:')[0];
+
+const EXCLUDED_PATH_PREFIXES = [
+  ...ONGOING_USER_CREATION_PATHS,
+  ...ONBOARDING_PATHS,
+  AppPath.ResetPassword,
+].map(extractPathPrefix);
+
+export const isValidReturnToPath = (path: string): boolean => {
```
charlesBochet, line 14: looks a bit "duplicated" with [isReturnToPathSessionStorageEntry.ts‎](https://github.com/twentyhq/twenty/pull/18308/changes#diff-79594017360e6281ca5b322749e545cbde7472617d69d6f72631fc12eec7a3bf). Not checking the same but this one is more accurate

PR #18319, opened 2026-03-02
[FRONT COMPONENTS] Introduce conditionalAvailabilityExpression to command menu items

bosiraphael:
## PR Description

- Uses `expr-eval` to enable front components (SDK plugins) to define conditional availability as declarative expressions.
- Moves shared types and constants to `twenty-shared`
- Introduces a `conditionalAvailabilityExpression` field on `CommandMenuItemEntity`, allowing command menu items to store an `expr-eval` compatible expression string that is evaluated against a CommandMenuContext to determine if the item should be shown.
- Creates an esbuild transform plugin `conditional-availability-transform-plugin` in `twenty-sdk` that converts TypeScript conditional availability expressions into `expr-eval` compatible syntax at build time, so SDK developers can write natural TS expressions that get transformed to evaluable strings.
- Removes deprecated `forceRegisteredActionsByKey` state and its usage.
- Creates `useCommandMenuContext` hook that builds the full `CommandMenuContext` object from React state, which is then passed to `useCommandMenuItemFrontComponentActions` for evaluating conditional availability expressions.


bosiraphael:
@greptileai

packages/twenty-front/src/modules/action-menu/actions/types/ShouldBeRegisteredRulesLogic.ts
```diff
@@ -0,0 +1,13 @@
+import { type AdditionalOperation, type RulesLogic } from 'json-logic-js';
+
+type ShouldBeRegisteredCustomOperations = {
```
charlesBochet: I don't understand this naming?

packages/twenty-front/src/modules/action-menu/actions/utils/__tests__/evaluateShouldBeRegisteredRule.test.ts
```diff
@@ -0,0 +1,387 @@
...
+    canSoftDeleteObjectRecords: true,
+    canDestroyObjectRecords: true,
+    restrictedFields: {},
+    rowLevelPermissionPredicates: [],
+    rowLevelPermissionPredicateGroups: [],
+  },
+  getTargetObjectReadPermission: () => true,
+  getTargetObjectWritePermission: () => true,
+  isFeatureFlagEnabled: () => false,
+  forceRegisteredActionsByKey: {},
```
charlesBochet: what is this about?

packages/twenty-front/src/modules/action-menu/actions/utils/__tests__/evaluateShouldBeRegisteredRule.test.ts
```diff
@@ -0,0 +1,387 @@
...
+};
+
+// Params are passed as plain data to jsonLogic.apply, so partial nested
+// objects are fine at runtime. One cast here avoids dozens in tests.
+const withParams = (
```
charlesBochet: I don't get this pattern looks like HOC

packages/twenty-front/src/modules/action-menu/actions/utils/registerShouldBeRegisteredOperators.ts
```diff
@@ -0,0 +1,64 @@
+import jsonLogic from 'json-logic-js';
+
+import { type ShouldBeRegisteredFunctionParams } from '@/action-menu/actions/types/ShouldBeRegisteredFunctionParams';
+
+let registered = false;
+
+export const registerShouldBeRegisteredOperators = () => {
```
charlesBochet: why going with the singleton behavior? why not but we should have good reason? why not new JsonLogicJs?

packages/twenty-front/src/modules/action-menu/actions/utils/evaluateShouldBeRegisteredRule.ts
```diff
@@ -0,0 +1,22 @@
+import jsonLogic from 'json-logic-js';
+
+import { type ShouldBeRegisteredFunctionParams } from '@/action-menu/actions/types/ShouldBeRegisteredFunctionParams';
+import { type ShouldBeRegisteredRulesLogic } from '@/action-menu/actions/types/ShouldBeRegisteredRulesLogic';
+import { registerShouldBeRegisteredOperators } from '@/action-menu/actions/utils/registerShouldBeRegisteredOperators';
+
+registerShouldBeRegisteredOperators();
+
+export const evaluateShouldBeRegisteredRule = (
+  rule: ShouldBeRegisteredRulesLogic | null | undefined,
```
charlesBochet: use Nullable

packages/twenty-front/src/modules/action-menu/actions/utils/__tests__/evaluateShouldBeRegisteredRule.test.ts
bosiraphael: We introduced this at some point when we wanted to force an action to be registered (I think it was for the edit mode in the page layouts) I have to check if it can be deprecated or not yet.

packages/twenty-sdk/src/sdk/front-component-api/json-logic-converter/__tests__/convert-should-be-registered-to-json-logic.test.ts
```diff
@@ -0,0 +1,355 @@
...
+  const arrowFunctions = sourceFile.getDescendantsOfKind(
+    SyntaxKind.ArrowFunction,
+  );
+
+  return convertArrowFunctionToJsonLogic({ node: arrowFunctions[0] });
+};
+
+describe('convertShouldBeRegisteredToJsonLogic', () => {
+  describe('Tier 1 - Trivial patterns', () => {
+    it('converts () => true', () => {
```
charlesBochet: all this can be simplified, right?

packages/twenty-sdk/src/sdk/front-component-api/json-logic-converter/constants/allowed-constants-in-should-be-registered.ts
```diff
@@ -0,0 +1,21 @@
...
+
+type KnownConstantValue = string | number | Record<string, string | number>;
+
+export const ALLOWED_CONSTANTS_IN_SHOULD_BE_REGISTERED: Record<
+  string,
+  KnownConstantValue
+> = {
+  BACKEND_BATCH_REQUEST_MAX_COUNT,
+  MUTATION_MAX_MERGE_RECORDS,
+  CoreObjectNameSingular: {
```
charlesBochet: do we need that, I'm not sure we need to put constant here. regarding BACKEND_BATCH_REQUEST_MAX_COUNT they should be rather exporter in twenty-shared so available in apps?

charlesBochet:
Discussed with you, I feel we should be using https://www.npmjs.com/package/expr-eval or similar

Weiko:
> Discussed with you, I feel we should be using https://www.npmjs.com/package/expr-eval or similar

As discussed @Devessier we should align here. Let's find a common pattern for conditionalDisplay in page layouts and command menu items

bosiraphael:
@greptileai

bosiraphael:
@cubic-dev-ai please review

packages/twenty-front/src/modules/action-menu/contexts/ActionMenuContextProviderWorkflowObjects.tsx
```diff
@@ -50,8 +51,10 @@ const ActionMenuContextProviderWorkflowObjectsContent = ({
   const runWorkflowRecordAgnosticActions =
     useRunWorkflowRecordAgnosticActions();
 
+  const commandMenuContext = useCommandMenuContext();
```
charlesBochet: I think Context is a bad naming given that it means something in React

charlesBochet (approved):
Awesome work!

PR #18320, opened 2026-03-02
SDK Split command and cli logic

prastoin:
# Introduction

Allow a consumer call the commands programmatically instead of passing by the exec
To do so extract from the command definition all the core logic, created a new error api that allow keeping same error logs granularity than before

## Usage
```ts
import { authLogin, appUninstall, functionExecute } from 'twenty-sdk/cli';

const result = await authLogin({
  apiKey: 'my-key',
  apiUrl: 'https://my-twenty.com',
});

if (!result.success) {
  throw new Error(result.error);
}
```

## `app:build`
Introduced a new command that will allow building the whole project without any watch setup
- Build and validate manifest
- Get or create app
- Synchronize manifest with twenty-sdk stub and no typecheck
- generate client
- Run typecheck
- Synchronize manifest again


charlesBochet (approved):

packages/twenty-apps/hello-world/package.json
```diff
@@ -16,7 +16,7 @@
     "auth": "twenty auth login"
   },
```
martmull, line 17: commands are not up to date

packages/twenty-apps/hello-world/package.json
```diff
@@ -16,7 +16,7 @@
...
   "dependencies": {
-    "twenty-sdk": "0.2.4"
+    "twenty-sdk": "0.6.3"
```
martmull, line 19: should be a dev dependency (look at seeded package.json in create-twenty-app)

packages/twenty-apps/hello-world/package.json
```diff
@@ -16,7 +16,7 @@
...
   },
   "devDependencies": {
     "@types/node": "^24.7.2"
```
martmull, line 22: add seed files in create-twenty-app/src/constants/base-application

packages/twenty-sdk/src/cli/public-operations/app-build.ts
martmull, line 1: finally you did not add a `private-operations` folder. Why?

prastoin: No only public operations

packages/twenty-apps/hello-world/package.json
prastoin: Wdym ?

Weiko (approved):
LGTM

PR #18334, opened 2026-03-03
fix: validate input before formatting in MultiItemFieldInput

neo773:
Reorder validateInput to run before formatInput to prevent parsePhoneNumber from throwing INVALID_COUNTRY on bad input.

Fixes TWENTY-FRONT-5RQ
/closes https://github.com/twentyhq/twenty/issues/17670


FelixMalfait:

Your preview environment is available at: http://bore.pub:34809

charlesBochet (commented):
@neo773 the change looks good, could we extract this logic to a util (I feel we should already have something similar for other field types, mignt not be very consolidated but we always have this need to first try to cast (format, trim, cast), then validate) and add unit test on top of it

neo773:
pushed changes its much cleaner now

packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/input/components/MultiItemFieldInput.tsx
```diff
@@ -88,13 +91,15 @@ export const MultiItemFieldInput = <T,>({
...
       }
-      const { isValid } = validateInputAndComputeUpdatedItems();
+      const { isValid, updatedItems } = validateInputAndComputeUpdatedItems();
 
       if (!isValid && isInputDisplayed) {
         return;
       }
 
-      handleSubmitChanges();
+      if (isValid) {
```
charlesBochet: isn't this already guarded above?

packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/input/components/MultiItemFieldInput.tsx
```diff
@@ -223,62 +228,51 @@ export const MultiItemFieldInput = <T,>({
     setInputValue('');
   };
 
-  const handleSubmitChanges = () => {
-    const { isValid, updatedItems } = validateInputAndComputeUpdatedItems();
-    if (!isValid) {
-      return;
+  const handleItemDeletionDisplayState = (remainingItems: T[]) => {
```
charlesBochet: I don't get this naming

neo773:
addressed the new changes you requested

charlesBochet (approved):

PR #18335, opened 2026-03-03
fix: enforce the user to pass in property with id suffix for morph relations in Rest API

mabdullahabaid:
REST API allowed users to pass in targetOpportunity, targetPerson, targetCompany etc when trying to create a noteTarget or a taskTarget. The request went through, we got back a 201, the record was created, but the relationship was never established since the FK was empty in the database.

This PR enforces users to send in the property with the `Id` suffix for consistency. So, the user sends in targetOpportunityId, targetPersonId, targetCompanyId etc. 

<p align="center">
image
</p>

If the users try to send without the "Id" suffix, they get an error explaining what to do.

<p align="center">
image
</p>

Additionally, the documentation itself contains the correct property names.

<p align="center">
image
</p>

Finally, the filters also enforce this "Id" suffix convention in the GET request.

<p align="center">
image
</p>

Edit: Updated error messages after the screenshots were taken to make them a little generic. Secondly, this PR also fixes the issue of morph relation ids and objects not appearing in the response (when depth is 1).


packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/data-arg-processor.service.ts
```diff
@@ -244,6 +243,35 @@ export class DataArgProcessor {
 
         return value;
       }
+      case FieldMetadataType.MORPH_RELATION: {
```
mabdullahabaid, line 229: Separated out cases for RELATION and MORPH_RELATION. Some code duplication, but easier to read in comparison to using a complicated "if" block inside a single case block.

```
if (
    fieldMetadata.type === FieldMetadataType.MORPH_RELATION &&
    key === fieldMetadata.name &&
    isDefined(fieldMetadataRelationSettings.joinColumnName)
)
```

FelixMalfait:

Your preview environment is available at: http://bore.pub:4215

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-filter/graphql-query-filter-field.parser.ts
```diff
@@ -59,6 +64,20 @@ export class GraphqlQueryFilterFieldParser {
...
+        FieldMetadataType.MORPH_RELATION,
+      ) &&
+      key === fieldMetadata.name
+    ) {
+      throw new GraphqlQueryRunnerException(
+        `Use "${fieldMetadata.settings.joinColumnName}" instead of "${key}" to filter on this relation.`,
+        GraphqlQueryRunnerExceptionCode.INVALID_QUERY_INPUT,
+        { userFriendlyMessage: STANDARD_ERROR_MESSAGE },
+      );
+    }
```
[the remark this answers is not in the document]
mabdullahabaid: For any MANY_TO_ONE morph relation that reaches the filter parser guard, joinColumnName is guaranteed to be defined because the ONE_TO_MANY case is already rejected earlier in the code flow

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/utils/is-relation-nested-operation.util.ts
```diff
@@ -0,0 +1,14 @@
+import { RELATION_NESTED_QUERY_KEYWORDS } from 'twenty-shared/constants';
+import { isDefined } from 'twenty-shared/utils';
+
+export const isRelationNestedOperation = (value: unknown): boolean => {
+  if (!isDefined(value) || typeof value !== 'object') {
+    return false;
+  }
+
+  return Object.keys(value).some(
```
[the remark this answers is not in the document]
mabdullahabaid: The issue is not valid. The `isRelationNestedOperation` check is intentionally a lightweight gate — it only needs to distinguish "is this an object with connect/disconnect keys (from GraphQL)" from "is this a string UUID (from REST)". The actual structural validation of connect/disconnect payloads happens downstream in `extractNestedRelationFieldsByEntityIndex` with much stricter checks.

packages/twenty-server/src/engine/utils/convert-object-metadata-to-schema-properties.util.ts
```diff
@@ -115,13 +115,31 @@ export const convertObjectMetadataToSchemaProperties = ({
...
     if (
       isFieldMetadataEntityOfType(field, FieldMetadataType.RELATION) &&
       field.settings?.relationType === RelationType.ONE_TO_MANY
     ) {
       return node;
     }
 
+    if (isFieldMetadataEntityOfType(field, FieldMetadataType.MORPH_RELATION)) {
+      return node;
+    }
```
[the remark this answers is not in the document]
mabdullahabaid: This has been addressed. `getSchemaComponentsRelationProperties` in `components.utils.ts` now includes `MORPH_RELATION` fields, so the response schema documents expanded morph relation objects. Additionally, `get-all-selectable-fields.util.ts` and `common-select-fields-helper.ts` were updated to select `joinColumnName` and recurse into morph relations at depth 1, so the actual API response now includes them too.

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/data-arg.processor.ts
```diff
@@ -244,6 +244,38 @@ export class DataArgProcessor {
...
+          );
+        }
+
+        if (key === fieldMetadataMorphRelationSettings.joinColumnName) {
+          return validateUUIDFieldOrThrow(value, key);
+        }
+
+        if (
+          isDefined(fieldMetadataMorphRelationSettings.joinColumnName) &&
+          !isRelationNestedOperation(value)
```
mabdullahabaid: Not feeling good about having to use this utility here. Serves the purpose, but seems like a hack/bad practice.

packages/twenty-server/test/integration/graphql/suites/inputs-validation/filter-validation/constants/failing-filter-input-by-field-metadata-type.constant.ts
```diff
@@ -77,7 +77,7 @@ export const failingFilterInputByFieldMetadataType: {
       },
       gqlErrorMessage: 'is not defined by type',
       restFilterInput: `${[joinColumnNameForManyToOneMorphRelationField1.replace('Id', '')]}[eq]:"6dd71a46-68fe-4420-82b3-0d5b00ad2642"`,
-      restErrorMessage: `Data validation error`,
+      restErrorMessage: `${joinColumnNameForManyToOneMorphRelationField1.replace('Id', '')} to filter on this relation`,
```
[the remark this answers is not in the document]
mabdullahabaid: The test does `JSON.stringify(response.body.messages).toContain(errorMessage)`. The messages is an array of strings, so J`SON.stringify produces ["message here"]`. A `toContain` on that stringified output works as long as the expected substring doesn't contain " characters (since those get escaped to \").

Using a specific substring that avoids quotes but is still unique to the error solves our problem here, thus the weakened message.

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/data-arg.processor.ts
mabdullahabaid: Reason: The `isRelationNestedOperation` utility was needed because `data-arg.processor.ts` is shared by both REST and GraphQL code paths. When I added the guard to reject morph relation fields passed by their relation name (e.g., targetPerson instead of targetPersonId), it broke GraphQL mutations that use connect/disconnect operations. 

The utility distinguishes between a REST user incorrectly passing a string/UUID under the relation name (which should be rejected) and a GraphQL mutation passing a structured connect/disconnect object (which should be allowed through to the ORM layer for processing).

charlesBochet:
Taking it :)

charlesBochet:
@etiennejouan actually, i'm missing a bit the vision here

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/data-arg.processor.ts
charlesBochet: @mabdullahabaid I don't understand how it's different between relation and morphRelation (I understand that's it could be easier to have two different switch cases but RELATION and MORPH_RELATION should be super similar we should:
- if that's the relationId gqlField, validaite it's a valid UUID
- if that's the relation gqlField, validate it's a connect / disconnect payload

I'm not fully sure what's the vision on validation if it should be done here or not @etiennejouan

packages/twenty-server/src/engine/api/common/common-select-fields/utils/get-all-selectable-fields.util.ts
```diff
@@ -85,6 +85,12 @@ export const getAllSelectableFields = ({
       isDefined(flatField.settings.joinColumnName)
     ) {
       result[flatField.settings.joinColumnName] = true;
+    } else if (
+      isFlatFieldMetadataOfType(flatField, FieldMetadataType.MORPH_RELATION) &&
```
charlesBochet: looks super similar to relaiton

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-filter/graphql-query-filter-field.parser.ts
```diff
@@ -59,6 +64,20 @@ export class GraphqlQueryFilterFieldParser {
       throw new Error(`Field metadata not found for field: ${key}`);
     }
 
+    if (
+      isFlatFieldMetadataOfType(
```
charlesBochet: I don't understand why there is an edge case about morph here (but not relation) suddenly

packages/twenty-server/src/engine/core-modules/open-api/utils/components.utils.ts
```diff
@@ -99,43 +99,55 @@ const getSchemaComponentsRelationProperties = (
   >['flatObjectMetadataMaps'],
 ): Properties => {
   return flatFieldMetadatas.reduce((node, field) => {
-    if (field.type !== FieldMetadataType.RELATION) {
+    if (!isDefined(field.relationTargetObjectMetadataId)) {
```
charlesBochet: we should rather start by assessing if this is a relation field, and validate field relationTargetObjectMetadataId and relationType are complete (throw if this is not the case rather than silently swallow)

packages/twenty-server/src/engine/utils/convert-object-metadata-to-schema-properties.util.ts
```diff
@@ -115,13 +115,31 @@ export const convertObjectMetadataToSchemaProperties = ({
...
     }
 
+    if (
+      isFieldMetadataEntityOfType(field, FieldMetadataType.MORPH_RELATION) &&
+      field.settings.relationType === RelationType.MANY_TO_ONE &&
+      isDefined(field.settings.joinColumnName)
+    ) {
+      return {
+        ...node,
+        [field.settings.joinColumnName]: {
```
charlesBochet: joinColumnName is deprecated, actually use {field.name}{objectName}Id for morph (we should have an util for that already)

packages/twenty-server/src/engine/utils/convert-object-metadata-to-schema-properties.util.ts
```diff
@@ -115,13 +115,31 @@ export const convertObjectMetadataToSchemaProperties = ({
...
+    }
+
     if (
       isFieldMetadataEntityOfType(field, FieldMetadataType.RELATION) &&
       field.settings?.relationType === RelationType.ONE_TO_MANY
     ) {
       return node;
     }
 
+    if (isFieldMetadataEntityOfType(field, FieldMetadataType.MORPH_RELATION)) {
```
charlesBochet: looks symetric with relation

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/data-arg-processor.service.ts
```diff
@@ -244,6 +244,38 @@ export class DataArgProcessor {
...
```
etiennejouan, line 229: You can separate but logic should be similar (to edit also in RELATION switch)

- if ONE_TO_MANY we throw
- if joinColumnName, validate UUID
- if isRelationNestedOperation, we throw with "Relation "${key}" requires connect or disconnect operation"
- else return value

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-filter/graphql-query-filter-field.parser.ts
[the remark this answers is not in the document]
etiennejouan: Concerning filter, we will add a new filter arg processor (same as data arg processor) for validation. We can remove this logic

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/data-arg.processor.ts
etiennejouan: Agree with Charles, logic should be similar.
And .... I agree with M. Abdullah, nested operation validation here makes sense. Validation layer should catch all invalid data args input.
My [comment](https://github.com/twentyhq/twenty/pull/18335/changes#r2877930298) above should help.

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/utils/is-relation-nested-operation.util.ts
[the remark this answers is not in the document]
etiennejouan: I wonder if it wouldn't be more optimized to do a stricter check 🤔

packages/twenty-server/src/engine/api/common/common-select-fields/utils/get-all-selectable-fields.util.ts
```diff
@@ -85,6 +85,12 @@ export const getAllSelectableFields = ({
       isDefined(flatField.settings.joinColumnName)
     ) {
       result[flatField.settings.joinColumnName] = true;
+    } else if (
```
etiennejouan: I'm not sure adding an if statement here brings more readability (it's subjective, sure)

```
} else if (
  (isFlatFieldMetadataOfType(flatField, FieldMetadataType.RELATION) ||
   isFlatFieldMetadataOfType(flatField, FieldMetadataType.MORPH_RELATION)) &&
  flatField.settings.relationType === RelationType.MANY_TO_ONE &&
  isDefined(flatField.settings.joinColumnName)
) {
```

etiennejouan (commented):
Good fix, thanks ! 

Overall, I'll try to handle as much shared logic as possible between relation and morph relation.

packages/twenty-server/src/engine/api/common/common-select-fields/utils/get-all-selectable-fields.util.ts
mabdullahabaid: Fixed by consolidating.

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-filter/graphql-query-filter-field.parser.ts
mabdullahabaid: You're right, should have captured for relation too. Etienne suggested in the review that we'd introduce filter arg processor and we can remove the logic in this PR, so did that.

Currently, it throws 500 when morph relation name is provided and `Id` suffix is not added to it.

packages/twenty-server/src/engine/core-modules/open-api/utils/components.utils.ts
mabdullahabaid: Done.

packages/twenty-server/src/engine/utils/convert-object-metadata-to-schema-properties.util.ts
mabdullahabaid: `computeMorphOrRelationFieldJoinColumnName` is the util, replaced joinColumnName occurrences.

packages/twenty-server/src/engine/utils/convert-object-metadata-to-schema-properties.util.ts
mabdullahabaid: Consolidated.

packages/twenty-server/src/engine/api/common/common-select-fields/utils/get-all-selectable-fields.util.ts
mabdullahabaid: Done.

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/data-arg-processor.service.ts
mabdullahabaid: Done.

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-filter/graphql-query-filter-field.parser.ts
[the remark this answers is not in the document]
mabdullahabaid: Removed as suggested.

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/data-arg.processor.ts
mabdullahabaid: Fixed it. Made the check stricter too as suggested by the referenced comment.

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/data-arg-processor.service.ts
```diff
@@ -224,24 +227,39 @@ export class DataArgProcessor {
...
+            })
+          : relationSettings.joinColumnName;
+
+        if (key === joinColumnName) {
           return validateUUIDFieldOrThrow(value, key);
         }
 
+        if (isDefined(joinColumnName) && !isRelationNestedOperation(value)) {
+          throw new CommonQueryRunnerException(
+            `Relation "${key}" requires connect or disconnect operation`,
```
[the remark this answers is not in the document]
mabdullahabaid: This is intentional, followed suggestion.

mabdullahabaid:
The stricter validation introduced in the data-arg processor would break any existing saved workflow actions that store relation fields in the old `{ id: "uuid" }` format. The frontend components `WorkflowEditActionCreateRecord` and `WorkflowEditActionUpsertRecord` were updated to store many-to-one relation values directly under their join column name (e.g. `companyId: "uuid"` instead of `company: { id: "uuid" }`), so newly created or edited workflow actions are saved in the correct format going forward. For existing workflow actions already persisted in the database, the upgrade command `upgrade:1-19:migrate-workflow-relation-fields` scans all workflow versions and transforms `CREATE_RECORD` / `UPSERT_RECORD` steps from the old format to the new one, ensuring they remain compatible with the stricter validation at runtime.

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/data-arg-processor.service.ts
```diff
@@ -224,24 +227,39 @@ export class DataArgProcessor {
...
+            CommonQueryRunnerExceptionCode.INVALID_ARGS_DATA,
+            { userFriendlyMessage: STANDARD_ERROR_MESSAGE },
+          );
+        }
```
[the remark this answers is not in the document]
mabdullahabaid: This was intentional as per the recommendation in [this](https://github.com/twentyhq/twenty/pull/18335#discussion_r2877930298) comment.

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/data-arg-processor.service.ts
```diff
@@ -224,24 +227,39 @@ export class DataArgProcessorService {
...
 
-        if (key === fieldMetadataRelationSettings.joinColumnName) {
+        const joinColumnName = isFlatFieldMetadataOfType(
+          fieldMetadata,
+          FieldMetadataType.MORPH_RELATION,
+        )
+          ? computeMorphOrRelationFieldJoinColumnName({
+              name: fieldMetadata.name,
+            })
+          : relationSettings.joinColumnName;
```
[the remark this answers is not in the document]
mabdullahabaid: joinColumnName was deprecated and we need to use the util to compute name of the morph relation identifier field. Source: [this comment](https://github.com/twentyhq/twenty/pull/18335#discussion_r2877561321)

mabdullahabaid:
@charlesBochet @etiennejouan ready for re-review.

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/utils/is-relation-nested-operation.util.ts
```diff
@@ -0,0 +1,17 @@
+import { RELATION_NESTED_QUERY_KEYWORDS } from 'twenty-shared/constants';
+import { isDefined } from 'twenty-shared/utils';
+
+const { CONNECT, DISCONNECT } = RELATION_NESTED_QUERY_KEYWORDS;
+
+export const isRelationNestedOperation = (value: unknown): boolean => {
+  if (!isDefined(value) || typeof value !== 'object' || Array.isArray(value)) {
```
etiennejouan: **nitpick**  isObject  from '@sniptt/guards';

packages/twenty-server/src/engine/api/common/common-select-fields/utils/get-all-selectable-fields.util.ts
```diff
@@ -80,11 +81,25 @@ export const getAllSelectableFields = ({
...
+          FieldMetadataType.MORPH_RELATION,
+        )) &&
+      flatField.settings.relationType === RelationType.MANY_TO_ONE
     ) {
-      result[flatField.settings.joinColumnName] = true;
+      const joinColumnName = isFlatFieldMetadataOfType(
+        flatField,
+        FieldMetadataType.MORPH_RELATION,
+      )
+        ? computeMorphOrRelationFieldJoinColumnName({ name: flatField.name })
```
etiennejouan: const joinColumnName = computeMorphOrRelationFieldJoinColumnName({ name: flatField.name })

works for RELATION and MORPH_RELATION, no ?

packages/twenty-server/src/engine/api/common/common-select-fields/utils/get-all-selectable-fields.util.ts
```diff
@@ -80,11 +81,25 @@ export const getAllSelectableFields = ({
...
+        : flatField.settings.joinColumnName;
+
+      if (isDefined(joinColumnName)) {
+        result[joinColumnName] = true;
+      } else {
+        result[flatField.name] = true;
+      }
```
etiennejouan: not sure to understand this

packages/twenty-server/src/engine/utils/convert-object-metadata-to-schema-properties.util.ts
```diff
@@ -102,22 +103,32 @@ export const convertObjectMetadataToSchemaProperties = ({
...
+        isFieldMetadataEntityOfType(field, FieldMetadataType.MORPH_RELATION)) &&
+      field.settings?.relationType === RelationType.MANY_TO_ONE;
+
+    if (isRelationManyToOne) {
+      const key = isFieldMetadataEntityOfType(
+        field,
+        FieldMetadataType.MORPH_RELATION,
+      )
+        ? computeMorphOrRelationFieldJoinColumnName({ name: field.name })
+        : `${field.name}Id`;
```
etiennejouan: same as earlier, it should work for both RELATION and MORPH_RELATION

packages/twenty-server/src/engine/utils/convert-object-metadata-to-schema-properties.util.ts
```diff
@@ -102,22 +103,32 @@ export const convertObjectMetadataToSchemaProperties = ({
...
         },
       };
     }
 
     if (
-      isFieldMetadataEntityOfType(field, FieldMetadataType.RELATION) &&
-      field.settings?.relationType === RelationType.ONE_TO_MANY
+      (isFieldMetadataEntityOfType(field, FieldMetadataType.RELATION) &&
+        field.settings?.relationType === RelationType.ONE_TO_MANY) ||
+      isFieldMetadataEntityOfType(field, FieldMetadataType.MORPH_RELATION)
```
etiennejouan:
```
(isFieldMetadataEntityOfType(field, FieldMetadataType.RELATION) ||
        isFieldMetadataEntityOfType(field, FieldMetadataType.MORPH_RELATION)) &&
      field.settings?.relationType === RelationType.ONE_TO_MANY
``` 

instead ?

etiennejouan (commented):
Hi @mabdullahabaid, nice catch concerning workflows! I'm not sure relation update/create was working since we linked workflow actions to the common API 🤔 Then with the stricter validation you added, it showed up. Great!
I would not recommend the data migration approach, always sensitive. It's not a big deal if workflow action output doesn't match the common API data args. Instead, I would add formatting logic in the WorkflowAction services (check if the update step needs it too).

packages/twenty-server/src/engine/api/common/common-select-fields/utils/get-all-selectable-fields.util.ts
```diff
@@ -80,11 +80,18 @@ export const getAllSelectableFields = ({
...
+          FieldMetadataType.MORPH_RELATION,
+        )) &&
+      flatField.settings.relationType === RelationType.MANY_TO_ONE
     ) {
-      result[flatField.settings.joinColumnName] = true;
+      const joinColumnName = computeMorphOrRelationFieldJoinColumnName({
+        name: flatField.name,
+      });
+
+      result[joinColumnName] = true;
```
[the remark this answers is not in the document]
mabdullahabaid: `joinColumnName` in settings is deprecated — the codebase is migrating to computing it at runtime via `computeMorphOrRelationFieldJoinColumnName`. The computed value (`${name}Id`) is the canonical naming convention. There's no desynchronization risk since MANY_TO_ONE join columns always follow the `${fieldName}Id` convention.

packages/twenty-server/src/modules/workflow/workflow-executor/utils/format-workflow-record-relation-fields.util.ts
```diff
@@ -0,0 +1,73 @@
...
+import { type ObjectMetadataInfo } from 'src/modules/workflow/common/workspace-services/workflow-common.workspace-service';
+
+const extractLegacyRelationId = (value: unknown): string | undefined => {
+  if (!isObject(value)) {
+    return undefined;
+  }
+
+  const record = value as Record<string, unknown>;
+
+  if (Object.keys(record).length !== 1 || !isString(record.id)) {
```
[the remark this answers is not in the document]
mabdullahabaid: The strictness is intentional. This function targets a specific legacy format (`{ id: "uuid" }`) from older workflow actions created before the frontend was updated. The current frontend already saves relations in join column format (`companyId: "uuid"`). An object like `{ id, name }` was never a valid relation format in workflow settings; accepting it would silently discard fields and mask malformed input that should fail validation downstream.

mabdullahabaid:
@etiennejouan made the changes. CI appears to be stuck for some reason?

Edit: nvm, it's now running after a delay.

packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/components/WorkflowEditActionCreateRecord.tsx
```diff
@@ -4,6 +4,7 @@ import { useObjectMetadataItems } from '@/object-metadata/hooks/useObjectMetadat
 import { formatFieldMetadataItemAsFieldDefinition } from '@/object-metadata/utils/formatFieldMetadataItemAsFieldDefinition';
```
etiennejouan, line 4: We don't need this changes anymore if we apply change server side, no ? I think it can be misleading to have different data format. wdyt ?

packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/components/WorkflowEditActionUpsertRecord.tsx
```diff
@@ -4,6 +4,7 @@ import { formatFieldMetadataItemAsFieldDefinition } from '@/object-metadata/util
 import { FormFieldInput } from '@/object-record/record-field/ui/components/FormFieldInput';
```
etiennejouan, line 4: We don't need this changes anymore if we apply change server side, no ? I think it can be misleading to have different data format. wdyt ?

packages/twenty-server/src/modules/workflow/workflow-executor/utils/format-workflow-record-relation-fields.util.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  const manyToOneRelationFieldNames = new Set<string>();
+
+  for (const fieldId of flatObjectMetadata.fieldIds) {
+    const universalIdentifier =
+      flatFieldMetadataMaps.universalIdentifierById[fieldId];
+
+    const field = isDefined(universalIdentifier)
+      ? flatFieldMetadataMaps.byUniversalIdentifier[universalIdentifier]
+      : undefined;
```
etiennejouan: can you use getFlatFieldsFromFlatObjectMetadata ?
and isFlatFieldMetadataOfType should avoid type casting

etiennejouan (commented):
> I've not approved because I don't know how to test the workflow action. I've created a morph many-to-one relation field. But I can't fill (don't know how to do) this field through workflow UI on create, update or upsert. Reaching you on discord

Edit : Tested on relation field and it works fien

etiennejouan (approved):

packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/components/WorkflowEditActionCreateRecord.tsx
mabdullahabaid: Yes, I agree. Removing.

packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/components/WorkflowEditActionUpsertRecord.tsx
mabdullahabaid: Agreed, removing.

packages/twenty-server/src/modules/workflow/workflow-executor/utils/format-workflow-record-relation-fields.util.ts
mabdullahabaid: Fixed.

PR #18337, opened 2026-03-03
Fix workflow crash on trying to diff undefined state when cache is empty

neo773:
fixes TWENTY-FRONT-5FC

### DIFF
--- a/packages/twenty-front/src/modules/workflow/workflow-steps/hooks/useCreateWorkflowVersionEdge.ts
+++ b/packages/twenty-front/src/modules/workflow/workflow-steps/hooks/useCreateWorkflowVersionEdge.ts
@@ -1,4 +1,6 @@
 import { useApolloCoreClient } from '@/object-metadata/hooks/useApolloCoreClient';
+import { CoreObjectNameSingular } from '@/object-metadata/types/CoreObjectNameSingular';
+import { useFindOneRecordQuery } from '@/object-record/hooks/useFindOneRecordQuery';
 import { useMutation } from '@apollo/client';
 import {
   type CreateWorkflowVersionEdgeMutation,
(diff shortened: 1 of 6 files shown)


FelixMalfait:

Your preview environment is available at: http://bore.pub:56915

packages/twenty-front/src/modules/workflow/workflow-steps/hooks/useCreateWorkflowVersionEdge.ts
```diff
@@ -21,7 +28,15 @@ export const useCreateWorkflowVersionEdge = () => {
   const createWorkflowVersionEdge = async (
     input: CreateWorkflowVersionEdgeInput,
   ) => {
-    const result = await mutate({ variables: { input } });
+    const result = await mutate({
+      variables: { input },
+      refetchQueries: [
```
charlesBochet, line 33: we usually avoid refetchQueries are they are not super reliable (and obsolete since we have realtime), what's the issue behind it?

charlesBochet (changes requested):
Left comments :) could you explain more why refetch query is needed?

neo773:
Same pattern as delete work flow step, just ensures UI renders correctly if diff cache (its not always reliable) is wrong so user doesn't see a stale UI

charlesBochet:
I'm really not a fan of sending refetch queries for no reason. What are the cases that lead to stale cache?

neo773:
Honestly it can be N number of cases, I will try do a deep dive but nothing can be bullet proof here, it's the whole architecture here that leads to this because the server always returns diffs and client state can never be 100% reliable

charlesBochet:
@neo773 closing this one, I don't think this is the right fix and I'm not sure the issue is still there

PR #18339, opened 2026-03-03
Fix name singular updates in dev mode

martmull:
as title


packages/twenty-sdk/src/cli/utilities/build/manifest/utils/get-default-relation-object-fields.ts
```diff
@@ -153,7 +154,7 @@ export const getDefaultRelationObjectFields = (
       label: config.label,
       description: `${objectConfig.labelPlural} tied to the ${config.targetLabel(objectConfig)}`,
       icon: config.icon,
-      isNullable: false,
+      isNullable: true,
```
Weiko, line 157: I don't get this change. I'm also confused because it seems our FK are actually non-nullable already so it's probably ignored? Anyway, this is probably a good change but this seems unused so it could be confusing. Not on you though

Weiko (approved):
LGTM

prastoin (approved):

PR #18350, opened 2026-03-03
Only update value at creation

martmull:
as title


packages/twenty-server/src/engine/core-modules/applicationVariable/application-variable.service.ts
```diff
@@ -97,31 +97,33 @@ export class ApplicationVariableEntityService {
...
     )) {
       const isSecretValue = isSecret ?? false;
-      const encryptedValue = this.encryptSecretValue(
-        value ?? '',
-        isSecretValue,
-      );
 
-      if (
-        await this.applicationVariableRepository.findOne({
+      const existingVariable = await this.applicationVariableRepository.findOne(
```
charlesBochet: very inefficient here, let's batch

charlesBochet:
@martmull is this really what we want from a devXP perspective?

martmull:
> isSecret: isSecretValue,

yes we don't want the value to be reset after each updates

ijreilly (changes requested):
I don't understand why we wouldn't allow users to update application variables. I think the spotted bug has a different origin (maybe it's being encrypted multiple times or something)

ijreilly (commented):
@martmull Ok I understand now, this function is only called at application-sync
I do find it "dangerous" though because we could easily think this function actually performs an update on the value while it doesn't. 
Could we change the design a bit - maybe a new argument in upsertManyApplicationVariableEntities's signature like shouldUpdateValue ?

ijreilly (approved):

PR #18358, opened 2026-03-03
feat: add npm and tarball app distribution with upgrade mechanism

FelixMalfait:
## Summary

- **npm + tarball app distribution**: Apps can be installed from the npm registry (public or private) or uploaded as `.tar.gz` tarballs, with `AppRegistrationSourceType` tracking the origin
- **Upgrade mechanism**: `AppUpgradeService` checks for newer versions, supports rollback for npm-sourced apps, and a cron job runs every 6 hours to update `latestAvailableVersion` on registrations
- **Security hardening**: Tarball extraction uses path traversal protection, and `enableScripts: false` in `.yarnrc.yml` disables all lifecycle scripts during `yarn install` to prevent RCE
- **Frontend**: "Install from npm" and "Upload tarball" modals, upgrade button on app detail page, blue "Update" badge on installed apps table when a newer version is available
- **Marketplace catalog sync**: Hourly cron job syncs a hardcoded catalog index into `ApplicationRegistration` entities
- **Integration tests**: Coverage for install, upgrade, tarball upload, and catalog sync flows

## Backend changes

| Area | Files |
|------|-------|
| Entity & migration | `ApplicationRegistrationEntity` (sourceType, sourcePackage, latestAvailableVersion), `ApplicationEntity` (applicationRegistrationId), migration |
| Services | `AppPackageResolverService`, `ApplicationInstallService`, `AppUpgradeService`, `MarketplaceCatalogSyncService` |
| Cron jobs | `MarketplaceCatalogSyncCronJob` (hourly), `AppVersionCheckCronJob` (every 6h) |
| REST endpoint | `AppRegistrationUploadController` — tarball upload with secure extraction |
| Resolver | `MarketplaceResolver` — simplified `installMarketplaceApp` (removed redundant `sourcePackage` arg) |
| Security | `.yarnrc.yml` — `enableScripts: false` to block postinstall RCE |

## Frontend changes

| Area | Files |
|------|-------|
| Modals | `SettingsInstallNpmAppModal`, `SettingsUploadTarballModal`, `SettingsAppModalLayout` |
| Hooks | `useUploadAppTarball`, `useInstallMarketplaceApp` (cleaned up) |
| Upgrade UI | `SettingsApplicationVersionContainer`, `SettingsApplicationDetailAboutTab` |
| Badge | `SettingsApplicationTableRow` — blue "Update" tag, `SettingsApplicationsInstalledTab` — fetches registrations for version comparison |
| Styling | Migrated to Linaria (matching main) |

## Test plan

- [ ] Install an app from npm via the "Install from npm" modal
- [ ] Upload a `.tar.gz` tarball via the "Upload tarball" modal
- [ ] Verify upgrade badge appears when `latestAvailableVersion > version`
- [ ] Verify upgrade flow from app detail page
- [ ] Run integration tests: `app-distribution.integration-spec.ts`, `marketplace-catalog-sync.integration-spec.ts`
- [ ] Verify `enableScripts: false` blocks postinstall scripts during yarn install


FelixMalfait:
## 📊 API Changes Report

### GraphQL Metadata Schema Changes
# GraphQL Metadata Schema Changes

[log] 
Detected the following changes (17) between schemas:

[log] ✖  Field Application.description changed type from String! to String
[log] ✖  Field Application.version changed type from String! to String
[log] ✖  Argument universalIdentifier: String! added to field Mutation.installMarketplaceApp
[log] ⚠  Enum value AppTarball was added to enum FileFolder
[log] ⚠  Argument version: String added to field Mutation.installMarketplaceApp
[log] ✔  Type AppRegistrationSourceType was added
[log] ✔  Field applicationRegistration was added to object type Application
[log] ✔  Field applicationRegistrationId was added to object type Application
[log] ✔  Field isFeatured was added to object type ApplicationRegistration
[log] ✔  Field latestAvailableVersion was added to object type ApplicationRegistration
[log] ✔  Field sourcePackage was added to object type ApplicationRegistration
[log] ✔  Field sourceType was added to object type ApplicationRegistration
[log] ✔  Type ApplicationRegistrationSummary was added
[log] ✔  Field sourcePackage was added to object type MarketplaceApp
[log] ✔  Field installNpmApp was added to object type Mutation
[log] ✔  Field upgradeApplication was added to object type Mutation
[log] ✔  Field uploadAppTarball was added to object type Mutation
[error] Detected 3 breaking changes
⚠️ **Breaking changes or errors detected in GraphQL metadata schema**

```
[log] 
Detected the following changes (17) between schemas:

[log] ✖  Field Application.description changed type from String! to String
[log] ✖  Field Application.version changed type from String! to String
[log] ✖  Argument universalIdentifier: String! added to field Mutation.installMarketplaceApp
[log] ⚠  Enum value AppTarball was added to enum FileFolder
[log] ⚠  Argument version: String added to field Mutation.installMarketplaceApp
[log] ✔  Type AppRegistrationSourceType was added
[log] ✔  Field applicationRegistration was added to object type Application
[log] ✔  Field applicationRegistrationId was added to object type Application
[log] ✔  Field isFeatured was added to object type ApplicationRegistration
[log] ✔  Field latestAvailableVersion was added to object type ApplicationRegistration
[log] ✔  Field sourcePackage was added to object type ApplicationRegistration
[log] ✔  Field sourceType was added to object type ApplicationRegistration
[log] ✔  Type ApplicationRegistrationSummary was added
[log] ✔  Field sourcePackage was added to object type MarketplaceApp
[log] ✔  Field installNpmApp was added to object type Mutation
[log] ✔  Field upgradeApplication was added to object type Mutation
[log] ✔  Field uploadAppTarball was added to object type Mutation
[error] Detected 3 breaking changes
Error generating diff
```



⚠️ **Please review these API changes carefully before merging.**

FelixMalfait:

Your preview environment is available at: http://bore.pub:60734

packages/twenty-server/src/engine/core-modules/application/utils/extract-tarball-securely.util.ts
```diff
@@ -0,0 +1,44 @@
...
+): Promise<void> => {
+  let totalExtractedSize = 0;
+
+  await tar.extract({
+    file: tarballPath,
+    cwd: targetDir,
+    filter: (entryPath, entry) => {
+      if (entryPath.includes('..')) {
+        return false;
+      }
```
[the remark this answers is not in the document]
FelixMalfait: Fixed — now uses `resolve(targetDir, entryPath)` + `startsWith(resolvedTarget)` instead of `includes('..')` for robust path traversal protection.

packages/twenty-sdk/src/cli/commands/app/app-push.ts
```diff
@@ -0,0 +1,98 @@
...
+    const response = await fetch(
+      `${serverUrl}/api/application-registrations/upload-tarball`,
+      {
+        method: 'POST',
+        headers: {
+          Authorization: `Bearer ${token}`,
+        },
+        body: formData,
+      },
+    );
```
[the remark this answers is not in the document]
FelixMalfait: Fixed — the SDK now sends `Content-Type: application/json` with a `{ tarball: base64 }` body, matching the controller's `TarballUploadBody` type.

packages/twenty-server/src/engine/core-modules/application/application-marketplace/services/marketplace-catalog-sync.service.ts
```diff
@@ -0,0 +1,163 @@
...
+    sourcePackage: string;
+    logoUrl: string | null;
+    websiteUrl: string | null;
+    termsUrl: string | null;
+    latestAvailableVersion: string | null;
+    workspaceId: string;
+  }): Promise<void> {
+    const existing = await this.appRegistrationRepository.findOne({
+      where: { universalIdentifier: params.universalIdentifier },
+    });
```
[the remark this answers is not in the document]
FelixMalfait: Fixed — added `workspaceId` to the `findOne` query in `upsertRegistration`, so each workspace gets its own registration.

packages/twenty-sdk/src/cli/commands/app/app-push.ts
```diff
@@ -0,0 +1,98 @@
...
+    const fileBuffer = fs.readFileSync(tarballPath);
+
+    formData.append(
+      'file',
+      new Blob([fileBuffer], { type: 'application/gzip' }),
+      path.basename(tarballPath),
+    );
+
+    const response = await fetch(
+      `${serverUrl}/api/application-registrations/upload-tarball`,
```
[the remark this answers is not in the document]
FelixMalfait: Fixed — corrected the endpoint to `/api/app-registrations/upload-tarball` and switched to JSON body with base64-encoded tarball to match the controller's expected format.

packages/twenty-server/src/engine/core-modules/application-registration/controllers/app-registration-upload.controller.ts
```diff
@@ -0,0 +1,230 @@
...
+          sourceType: AppRegistrationSourceType.NONE,
+          oAuthClientId: v4(),
+          oAuthRedirectUris: [],
+          oAuthScopes: [],
+          workspaceId,
+        });
+
+        appRegistration =
+          await this.appRegistrationRepository.save(appRegistration);
+      }
```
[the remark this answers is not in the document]
FelixMalfait: Fixed — new registrations are now created with `sourceType: AppRegistrationSourceType.TARBALL` directly, eliminating the TOCTOU window where an orphaned NONE record could persist.

packages/twenty-server/src/engine/core-modules/application/resolvers/marketplace.resolver.ts
```diff
@@ -14,17 +26,164 @@ import { WorkspaceAuthGuard } from 'src/engine/guards/workspace-auth.guard';
 @MetadataResolver()
 @UseGuards(UserAuthGuard, WorkspaceAuthGuard, NoPermissionGuard)
 export class MarketplaceResolver {
-  constructor(private readonly marketplaceService: MarketplaceService) {}
+  private readonly logger = new Logger(MarketplaceResolver.name);
+  private hasSyncedOnce = false;
```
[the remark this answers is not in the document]
FelixMalfait: Moved the flag assignment to after successful sync. Re: multi-pod — the sync is an idempotent upsert, so duplicate syncs in a scale-out scenario are harmless (just redundant). Adding Redis/DB coordination would be overengineering for a lazy-init optimization that only runs once per cold start.

packages/twenty-server/src/engine/core-modules/application/services/application-install.service.ts
```diff
@@ -0,0 +1,267 @@
...
+    let hash = 0;
+
+    for (let i = 0; i < combined.length; i++) {
+      const char = combined.charCodeAt(i);
+
+      hash = (hash << 5) - hash + char;
+      hash |= 0;
+    }
+
+    return hash;
```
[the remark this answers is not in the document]
FelixMalfait: Acknowledged — the 32-bit key space is unlikely to cause real collisions in practice (we'd need many concurrent installs of *different* apps in the *same* workspace), but I'll track this as a follow-up improvement to use a proper 64-bit hash.

packages/twenty-front/src/pages/settings/applications/components/SettingsApplicationsTable.tsx
```diff
@@ -70,21 +72,33 @@ export const SettingsApplicationsTable = ({
...
-        ))}
+        {filteredApplications.map((application) => {
+          const latestVersion = application.applicationRegistrationId
+            ? registrationVersionMap?.get(application.applicationRegistrationId)
+            : undefined;
+
+          const hasUpdate =
+            latestVersion !== undefined &&
+            application.version !== undefined &&
+            latestVersion !== application.version;
```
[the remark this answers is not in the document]
FelixMalfait: Good catch — for the initial marketplace launch, the simple inequality is intentional since versions should only go up. A proper semver comparison with `semver.gt()` would be more correct; I'll track this as a follow-up.

packages/twenty-server/src/engine/core-modules/application/services/app-package-resolver.service.ts
```diff
@@ -0,0 +1,269 @@
...
+    };
+
+    await fs.writeFile(
+      join(workDir, 'package.json'),
+      JSON.stringify(packageJson, null, 2),
+    );
+  }
+
+  private async runYarnInstall(workDir: string): Promise<void> {
+    const localYarnPath = join(workDir, '.yarn/releases/yarn-4.9.2.cjs');
```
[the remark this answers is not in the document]
FelixMalfait: The yarn engine directory is a constant that ships bundled with the server — the `.cjs` filename and the constant are updated together. This won't silently break; a version bump to the engine would necessarily update both. Not a runtime concern.

packages/twenty-server/src/engine/core-modules/application/application-install/app-package-resolver.service.ts
```diff
@@ -0,0 +1,269 @@
...
+} from 'src/engine/core-modules/application/application.exception';
+import { YARN_ENGINE_DIRNAME } from 'src/engine/core-modules/application/constants/yarn-engine-dirname';
+import { extractTarballSecurely } from 'src/engine/core-modules/application/utils/extract-tarball-securely.util';
+import { FileStorageService } from 'src/engine/core-modules/file-storage/file-storage.service';
+import { TwentyConfigService } from 'src/engine/core-modules/twenty-config/twenty-config.service';
+import { streamToBuffer } from 'src/utils/stream-to-buffer';
+
+const execFilePromise = promisify(execFile);
+
+const APP_RESOLVER_TMPDIR = join(tmpdir(), 'twenty-app-resolver');
```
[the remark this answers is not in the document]
FelixMalfait: The parent `/tmp/twenty-app-resolver` directory is empty after each operation completes (all subdirectories are cleaned up). The OS cleans `/tmp` on reboot. Adding explicit cleanup of the parent would be a minor improvement but not a real issue.

packages/twenty-server/src/engine/core-modules/application-registration/controllers/app-registration-upload.controller.ts
```diff
@@ -0,0 +1,230 @@
...
+          name: manifest.application?.displayName ?? 'Unknown App',
+          sourceType: AppRegistrationSourceType.NONE,
+          oAuthClientId: v4(),
+          oAuthRedirectUris: [],
+          oAuthScopes: [],
+          workspaceId,
+        });
+
+        appRegistration =
+          await this.appRegistrationRepository.save(appRegistration);
```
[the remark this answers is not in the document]
FelixMalfait: Fixed — new registrations are now created with `sourceType: TARBALL` directly instead of NONE-then-update, and the JSON.parse of manifest.json is now wrapped in a try-catch. The TOCTOU window is eliminated.

packages/twenty-server/src/engine/core-modules/application/application-install/application-install.service.ts
```diff
@@ -0,0 +1,267 @@
...
+        params.workspaceId,
+      );
+
+      const sourceType = this.mapSourceType(appRegistration.sourceType);
+
+      await this.updateApplicationSourceType(
+        appRegistration.universalIdentifier,
+        params.workspaceId,
+        sourceType,
+      );
```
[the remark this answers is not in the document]
FelixMalfait: The `NONE` case in `mapSourceType` returns `'local'` intentionally (for OAuth-only apps), but the install flow already returns early for NONE-sourced apps at L68-L74 before reaching `mapSourceType`. Added a `default` fallback to the switch for extra safety.

packages/twenty-server/src/engine/core-modules/application/application-install/application-install.service.ts
```diff
@@ -0,0 +1,267 @@
...
+    const entries = await fs.readdir(dir, { withFileTypes: true });
+
+    for (const entry of entries) {
+      const fullPath = join(dir, entry.name);
+
+      if (entry.name === 'node_modules' || entry.name === '.yarn') {
+        continue;
+      }
+
+      if (entry.isDirectory()) {
```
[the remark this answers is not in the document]
FelixMalfait: Fixed — `collectFiles` now skips symlinks with an `entry.isSymbolicLink()` check.

packages/twenty-server/src/engine/core-modules/application/resolvers/marketplace.resolver.ts
```diff
@@ -14,17 +26,164 @@ import { WorkspaceAuthGuard } from 'src/engine/guards/workspace-auth.guard';
...
+    private readonly appRegistrationRepository: Repository<ApplicationRegistrationEntity>,
+    private readonly marketplaceCatalogSyncService: MarketplaceCatalogSyncService,
+    private readonly applicationInstallService: ApplicationInstallService,
+    private readonly appUpgradeService: AppUpgradeService,
+  ) {}
 
   @Query(() => [MarketplaceAppDTO])
   async findManyMarketplaceApps(): Promise<MarketplaceAppDTO[]> {
-    return this.marketplaceService.findAllMarketplaceApps();
+    let registrations = await this.appRegistrationRepository.find({
```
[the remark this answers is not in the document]
FelixMalfait: Fixed — `hasSyncedOnce` is now set *after* the sync succeeds, so a transient failure won't suppress retries.

packages/twenty-server/src/engine/core-modules/application-registration/controllers/app-registration-upload.controller.ts
```diff
@@ -0,0 +1,230 @@
...
+    await fs.mkdir(tempDir, { recursive: true });
+
+    try {
+      const tarballPath = join(tempDir, 'app.tar.gz');
+
+      await fs.writeFile(tarballPath, tarballBuffer);
+
+      const manifestPath = await this.extractAndValidate(tarballPath, tempDir);
+      const manifestContent = await fs.readFile(manifestPath, 'utf-8');
+      const manifest = JSON.parse(manifestContent);
```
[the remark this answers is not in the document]
FelixMalfait: Fixed — `JSON.parse(manifestContent)` is now wrapped in a try-catch that throws a clear `INVALID_INPUT` exception.

packages/twenty-front/src/pages/settings/applications/components/SettingsUploadTarballModal.tsx
```diff
@@ -0,0 +1,110 @@
...
+    const file = event.target.files?.[0];
+
+    if (!isDefined(file)) {
+      return;
+    }
+
+    const uploadResult = await upload(file);
+
+    if (uploadResult.success) {
+      await install({
```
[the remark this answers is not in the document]
FelixMalfait: Fixed — the modal now only closes when install returns truthy. If install fails, the modal stays open so the user can see the error.

packages/twenty-server/src/engine/core-modules/application/application-marketplace/utils/get-admin-workspace-id.util.ts
```diff
@@ -0,0 +1,28 @@
...
+// Finds the oldest active workspace that has at least one admin user
+// (canAccessFullAdminPanel or canImpersonate). Used to assign marketplace
+// catalog ApplicationRegistrations to a real workspace.
+export const getAdminWorkspaceId = async (
+  dataSource: DataSource,
+): Promise<string | null> => {
+  const result = await dataSource.query<Array<{ workspaceId: string }>>(
+    `SELECT uw."workspaceId"
+     FROM core."userWorkspace" uw
+     JOIN core."user" u ON u.id = uw."userId" AND u."deletedAt" IS NULL
```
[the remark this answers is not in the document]
FelixMalfait: Acknowledged — the admin workspace query is straightforward and correct for the current single-admin-workspace model. The `deletedAt IS NULL` filter is handled by TypeORM's soft-delete support. No change needed.

packages/twenty-server/src/engine/core-modules/application/application.exception.ts
```diff
@@ -35,6 +41,18 @@ const getApplicationExceptionUserFriendlyMessage = (
       return msg`You do not have permission to perform this action.`;
     case ApplicationExceptionCode.INVALID_INPUT:
       return msg`Invalid input provided.`;
+    case ApplicationExceptionCode.SOURCE_CHANNEL_MISMATCH:
+      return msg`Source channel mismatch.`;
+    case ApplicationExceptionCode.PACKAGE_RESOLUTION_FAILED:
+      return msg`Failed to resolve app package.`;
```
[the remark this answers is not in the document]
FelixMalfait: Fixed — improved the `PACKAGE_RESOLUTION_FAILED` message to be more user-friendly: 'Unable to retrieve the application package.'

packages/twenty-front/src/pages/settings/applications/hooks/useUploadAppTarball.ts
```diff
@@ -0,0 +1,88 @@
...
+export const useUploadAppTarball = () => {
+  const { enqueueErrorSnackBar } = useSnackBar();
+  const [isUploading, setIsUploading] = useState(false);
+
+  const upload = async (file: File): Promise<UploadResult> => {
+    setIsUploading(true);
+
+    try {
+      const arrayBuffer = await file.arrayBuffer();
+      const base64 = btoa(
```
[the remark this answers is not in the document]
FelixMalfait: The base64 encoding adds ~33% overhead, but the server enforces a 50MB limit on the decoded buffer. For the initial implementation this is acceptable — switching to multipart/form-data streaming would be a larger refactor that I'd prefer to do as a follow-up if large tarballs become common.

packages/twenty-server/src/engine/core-modules/application/application-rest-api-exception-filter.ts
```diff
@@ -0,0 +1,62 @@
...
+export class ApplicationRestApiExceptionFilter implements ExceptionFilter {
+  constructor(
+    private readonly httpExceptionHandlerService: HttpExceptionHandlerService,
+  ) {}
+
+  catch(exception: ApplicationException, host: ArgumentsHost) {
+    const ctx = host.switchToHttp();
+    const response = ctx.getResponse<Response>();
+
+    switch (exception.code) {
```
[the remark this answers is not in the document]
FelixMalfait: Fixed — added a `default` branch to the switch that returns 500, so any future exception codes won't silently fall through.

packages/twenty-server/src/engine/core-modules/application/services/app-upgrade.service.ts
```diff
@@ -0,0 +1,154 @@
...
+      return null;
+    }
+
+    const registryUrl =
+      appRegistration.registryUrl ??
+      this.twentyConfigService.get('APP_REGISTRY_URL');
+
+    try {
+      const response = await fetch(
+        `${registryUrl}/${appRegistration.sourcePackage}/latest`,
```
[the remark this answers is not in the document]
FelixMalfait: Fixed — scoped npm package names (e.g. `@twenty/hello-world`) are now URL-encoded with `encodeURIComponent()` before being used in the registry URL.

packages/twenty-server/src/engine/core-modules/application/services/app-package-resolver.service.ts
```diff
@@ -0,0 +1,269 @@
...
+  private async readManifest(dir: string): Promise<Manifest> {
+    const manifestPath = join(dir, 'manifest.json');
+
+    try {
+      const content = await fs.readFile(manifestPath, 'utf-8');
+
+      return JSON.parse(content) as Manifest;
+    } catch {
+      throw new ApplicationException(
+        `manifest.json not found at ${manifestPath}`,
```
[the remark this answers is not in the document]
FelixMalfait: Fixed — error messages in `readManifest` and `readPackageJson` no longer include the filesystem path. They now say 'manifest.json not found in resolved package' / 'package.json not found in resolved package'.

packages/twenty-server/src/engine/core-modules/application/application-install/application-install.service.ts
```diff
@@ -0,0 +1,267 @@
...
+        { targetVersion: params.version },
+      );
+
+      if (!resolvedPackage) {
+        await queryRunner.commitTransaction();
+
+        return true;
+      }
+
+      await this.writeFilesToStorage(
```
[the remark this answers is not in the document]
FelixMalfait: Added a `default` fallback to `mapSourceType`. The NONE case already returns early at L68-L74 before reaching this method, so the mapping is correct — but the default provides extra safety.

packages/twenty-server/src/engine/core-modules/application/resolvers/marketplace.resolver.ts
```diff
@@ -14,17 +26,164 @@ import { WorkspaceAuthGuard } from 'src/engine/guards/workspace-auth.guard';
...
+      where: { sourceType: AppRegistrationSourceType.NPM },
+    });
+
+    if (registrations.length === 0 && !this.hasSyncedOnce) {
+      this.logger.log(
+        'No marketplace registrations found, triggering initial catalog sync',
+      );
+      this.hasSyncedOnce = true;
+
+      await this.marketplaceCatalogSyncService.syncCatalog();
```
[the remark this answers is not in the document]
FelixMalfait: Fixed — `hasSyncedOnce` is now set after the sync call succeeds, not before.

packages/twenty-server/src/engine/core-modules/application/services/application-install.service.ts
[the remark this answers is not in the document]
FelixMalfait: Now fixed — `computeLockKeys` returns two separate INT4 hashes (one per input), and we use `pg_advisory_xact_lock(key1, key2)` for a 64-bit key space.

packages/twenty-front/src/pages/settings/applications/components/SettingsApplicationsTable.tsx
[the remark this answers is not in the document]
FelixMalfait: Now fixed — added `isNewerSemver()` utility that does proper major.minor.patch comparison. The update badge only shows when the latest version is strictly greater.

packages/twenty-server/src/engine/core-modules/application/resolvers/marketplace.resolver.ts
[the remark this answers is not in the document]
FelixMalfait: Further improved — replaced the boolean flag with a timestamp-based cooldown (5 minutes). This allows retries after transient failures while still avoiding spam. Each pod converges independently via idempotent upserts.

packages/twenty-server/src/engine/core-modules/application/services/app-package-resolver.service.ts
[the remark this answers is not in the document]
FelixMalfait: Now fixed — `runYarnInstall` reads `.yarnrc.yml` to resolve the `yarnPath` dynamically instead of hardcoding the version.

packages/twenty-server/src/engine/core-modules/application/application-install/app-package-resolver.service.ts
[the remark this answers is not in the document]
FelixMalfait: Now fixed — `AppPackageResolverService` implements `OnModuleInit` and wipes `APP_RESOLVER_TMPDIR` on startup, clearing any stale temp files from previous server runs.

packages/twenty-server/test/integration/metadata/suites/application/app-distribution.integration-spec.ts
```diff
@@ -0,0 +1,296 @@
...
+import { join } from 'path';
+import { tmpdir } from 'os';
+
+import request from 'supertest';
+import * as tar from 'tar';
+import { type DataSource } from 'typeorm';
+
+const TEST_WORKSPACE_ID = '20202020-1c25-4d02-bf25-6aeccf7ea419';
+
+const createTestTarball = async (
```
prastoin, line 13: ***Polish/Question:*** Could this be a constant test folder ?

packages/twenty-server/test/integration/metadata/suites/application/marketplace-catalog-sync.integration-spec.ts
```diff
@@ -0,0 +1,188 @@
+import crypto from 'crypto';
+
+import request from 'supertest';
+import { type DataSource } from 'typeorm';
+
+const TEST_WORKSPACE_ID = '20202020-1c25-4d02-bf25-6aeccf7ea419';
+
+const MARKETPLACE_QUERY = `
```
prastoin, line 8: ***Polish:*** Create dedicated testing integration util

packages/twenty-server/src/engine/core-modules/application/services/application-install.service.ts
```diff
@@ -0,0 +1,241 @@
...
+
+    const queryRunner = this.dataSource.createQueryRunner();
+
+    await queryRunner.connect();
+    await queryRunner.startTransaction();
+
+    let resolvedPackage: ResolvedPackage | null = null;
+
+    try {
+      await queryRunner.query(`SELECT pg_advisory_xact_lock($1, $2)`, [
```
prastoin: ***Question:*** Why not setting up the lock within redis directly ? feels like a bit late at pg lvl already
Would ask for Coco's opinion about this

Also unless I'm mistaken we don't gracefully handle nor return the exception to the end user

packages/twenty-server/src/engine/core-modules/application/application-install/application-install.service.ts
```diff
@@ -0,0 +1,241 @@
...
+  folder: FileFolder;
+}> = [
+  { pattern: /\.function\.mjs$/, folder: FileFolder.BuiltLogicFunction },
+  {
+    pattern: /\.front-component\.mjs$/,
+    folder: FileFolder.BuiltFrontComponent,
+  },
+  { pattern: /^public\//, folder: FileFolder.PublicAsset },
+];
+
```
prastoin, line 38: ***Self reminder:*** Challenge, looks like either duplicated or not reliable

packages/twenty-server/src/engine/core-modules/application/services/application-install.service.ts
```diff
@@ -0,0 +1,241 @@
...
+        lockKey2,
+      ]);
+
+      resolvedPackage = await this.appPackageResolverService.resolvePackage(
+        appRegistration,
+        { targetVersion: params.version },
+      );
+
+      if (!resolvedPackage) {
+        await queryRunner.commitTransaction();
```
prastoin: ***Remark:*** Unless I'm mistaken could be removed

packages/twenty-server/src/engine/core-modules/application/services/app-upgrade.service.ts
```diff
@@ -0,0 +1,163 @@
...
+    const registryUrl = this.twentyConfigService.get('APP_REGISTRY_URL');
+
+    if (!appRegistration.sourcePackage) {
+      return null;
+    }
+
+    try {
+      const encodedPackage = encodeURIComponent(appRegistration.sourcePackage);
+
+      const response = await fetch(`${registryUrl}/${encodedPackage}/latest`, {
```
prastoin: ***Polish:*** replace fetch by axios

packages/twenty-server/src/engine/core-modules/application/services/app-upgrade.service.ts
```diff
@@ -0,0 +1,163 @@
...
+          `Version check failed for ${appRegistration.sourcePackage}: ${response.status}`,
+        );
+
+        return null;
+      }
+
+      const metadata = (await response.json()) as Record<string, unknown>;
+      const latestVersion = metadata.version;
+
+      if (typeof latestVersion !== 'string') {
```
prastoin: ***Polish:*** zod validation

packages/twenty-server/src/engine/core-modules/marketplace/services/marketplace.service.ts
```diff
@@ -0,0 +1,90 @@
...
+  private readonly logger = new Logger(MarketplaceService.name);
+
+  constructor(private readonly twentyConfigService: TwentyConfigService) {}
+
+  async fetchAppsFromNpmRegistry(): Promise<MarketplaceAppDTO[]> {
+    const registryUrl = this.twentyConfigService.get('APP_REGISTRY_URL');
+    const apps: MarketplaceAppDTO[] = [];
+
+    try {
+      const response = await fetch(
```
prastoin: use axios

packages/twenty-server/src/engine/core-modules/marketplace/services/marketplace.service.ts
```diff
@@ -0,0 +1,90 @@
...
+
+      const results = (await response.json()) as Record<string, unknown>;
+
+      if (!Array.isArray(results.objects)) {
+        this.logger.warn('Unexpected npm search response shape');
+
+        return apps;
+      }
+
+      for (const result of results.objects as NpmSearchResult['objects']) {
```
prastoin: use zod to parse response data model

packages/twenty-server/src/engine/core-modules/application/application-marketplace/services/marketplace.service.ts
```diff
@@ -0,0 +1,90 @@
...
+  }>;
+};
+
+@Injectable()
+export class MarketplaceService {
+  private readonly logger = new Logger(MarketplaceService.name);
+
+  constructor(private readonly twentyConfigService: TwentyConfigService) {}
+
+  async fetchAppsFromNpmRegistry(): Promise<MarketplaceAppDTO[]> {
```
prastoin, line 31: ***Remark:*** Npm naming in code, shouldn't this be an agnostic registry driver naming ?

packages/twenty-server/src/engine/core-modules/application/application-marketplace/services/marketplace.service.ts
```diff
@@ -0,0 +1,90 @@
...
+          : name;
+
+        apps.push({
+          id: universalIdentifier,
+          name,
+          description: description ?? '',
+          icon: 'IconApps',
+          version,
+          author: author?.name ?? 'Unknown',
+          category: '',
```
prastoin: always empty string

packages/twenty-server/src/engine/core-modules/application/application-marketplace/dtos/marketplace-app.dto.ts
```diff
@@ -276,4 +276,9 @@ export class MarketplaceAppDTO {
   @IsOptional()
   @Field(() => MarketplaceAppDefaultRoleDTO, { nullable: true })
   defaultRole?: MarketplaceAppDefaultRoleDTO;
+
+  @IsOptional()
+  @IsString()
+  @Field({ nullable: true })
+  sourcePackage?: string;
 }
```
prastoin, line 288: Enabled class validator nested validation

packages/twenty-server/src/engine/core-modules/application/application-marketplace/resolvers/marketplace.resolver.ts
```diff
@@ -0,0 +1,88 @@
...
+    return this.applicationInstallService.installApplication({
+      appRegistrationId: registration.id,
+      version,
+      workspaceId: workspace.id,
+    });
+  }
+
+  @Mutation(() => Boolean)
+  @UseGuards(SettingsPermissionGuard(PermissionFlagType.MARKETPLACE_APPS))
+  async installNpmApp(
```
prastoin, line 81: Would rename npm to `fromRegistry`

prastoin: As discussed I feel like this should be done through the `installMarketplaceApp`, if we decide that all npm twenty applications are listed in the marketplace ( we would still to vent )

packages/twenty-server/src/engine/core-modules/application/application-marketplace/services/marketplace-query.service.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  constructor(
+    @InjectRepository(ApplicationRegistrationEntity)
+    private readonly appRegistrationRepository: Repository<ApplicationRegistrationEntity>,
+    @InjectMessageQueue(MessageQueue.cronQueue)
+    private readonly messageQueueService: MessageQueueService,
+  ) {}
+
+  async findManyMarketplaceApps(): Promise<MarketplaceAppDTO[]> {
+    const registrations = await this.appRegistrationRepository.find({
```
prastoin, line 42: ***Suggestion:*** We should store all this hourly hydrated data within redis

packages/twenty-server/src/engine/core-modules/application/application-registration/application-registration.entity.ts
```diff
@@ -39,7 +45,7 @@ import { WorkspaceEntity } from 'src/engine/core-modules/workspace/workspace.ent
   },
 )
 @Index('IDX_APPLICATION_REGISTRATION_CREATED_BY_USER_ID', ['createdByUserId'])
-@Index('IDX_APPLICATION_REGISTRATION_WORKSPACE_ID', ['workspaceId'])
+@Index('IDX_APPLICATION_REGISTRATION_WORKSPACE_ID', ['ownerWorkspaceId'])
 export class ApplicationRegistrationEntity {
```
prastoin, line 52: extends `WorkspaceRelatedEntity`

packages/twenty-server/src/engine/core-modules/application/application-registration/application-registration.entity.ts
```diff
@@ -87,13 +93,38 @@ export class ApplicationRegistrationEntity {
...
+    type: 'text',
+    default: AppRegistrationSourceType.LOCAL,
+  })
+  sourceType: AppRegistrationSourceType;
+
+  @Field(() => String, { nullable: true })
+  @Column({ nullable: true, type: 'text' })
+  sourcePackage: string | null;
+
+  @Column({ nullable: true, type: 'uuid' })
```
prastoin, line 120: ***Challenge:*** I feel like we should add a constraint that enforce the tarball to be defined for `sourceType` `tarball` ?

packages/twenty-server/src/engine/core-modules/application/application-registration/application-registration.entity.ts
```diff
@@ -87,13 +93,38 @@ export class ApplicationRegistrationEntity {
...
+  @Field(() => AppRegistrationSourceType)
+  @Column({
+    type: 'text',
+    default: AppRegistrationSourceType.LOCAL,
+  })
+  sourceType: AppRegistrationSourceType;
+
+  @Field(() => String, { nullable: true })
+  @Column({ nullable: true, type: 'text' })
+  sourcePackage: string | null;
```
prastoin, line 118: Would add dynamic constraint to be required for source type `npm`

packages/twenty-server/src/engine/core-modules/application-registration/application-registration.entity.ts
```diff
@@ -10,16 +10,22 @@ import {
...
 
 import { UUIDScalarType } from 'src/engine/api/graphql/workspace-schema-builder/graphql-types/scalars';
 import { ApplicationRegistrationVariableEntity } from 'src/engine/core-modules/application-registration/application-registration-variable.entity';
+import { AppRegistrationSourceType } from 'src/engine/core-modules/application-registration/enums/app-registration-source-type.enum';
+import { FileEntity } from 'src/engine/core-modules/file/entities/file.entity';
+import { type MarketplaceDisplayData } from 'src/engine/core-modules/marketplace/types/marketplace-display-data.type';
 import { UserEntity } from 'src/engine/core-modules/user/user.entity';
 import { WorkspaceEntity } from 'src/engine/core-modules/workspace/workspace.entity';
 
+export { AppRegistrationSourceType };
```
prastoin: rm

packages/twenty-server/src/engine/core-modules/application/application-registration/application-registration.entity.ts
```diff
@@ -87,13 +93,38 @@ export class ApplicationRegistrationEntity {
...
+  @Column({ nullable: true, type: 'uuid' })
+  tarballFileId: string | null;
+
+  @OneToOne(() => FileEntity, { onDelete: 'SET NULL', nullable: true })
+  @JoinColumn({ name: 'tarballFileId' })
+  tarballFile: Relation<FileEntity> | null;
+
+  @Field(() => String, { nullable: true })
+  @Column({ nullable: true, type: 'text' })
+  latestAvailableVersion: string | null;
```
prastoin, line 129: From my understanding this should only be nullable for source type `local`

packages/twenty-server/src/engine/core-modules/marketplace/services/marketplace-catalog-sync.service.ts
```diff
@@ -0,0 +1,168 @@
...
+    sourceType: AppRegistrationSourceType;
+    sourcePackage: string;
+    logoUrl: string | null;
+    websiteUrl: string | null;
+    termsUrl: string | null;
+    latestAvailableVersion: string | null;
+    isFeatured: boolean;
+    marketplaceDisplayData: MarketplaceDisplayData | null;
+    ownerWorkspaceId: string;
+  }): Promise<void> {
```
prastoin: ***Polish:*** Pick in existing type

packages/twenty-server/src/engine/core-modules/marketplace/services/marketplace-catalog-sync.service.ts
```diff
@@ -0,0 +1,168 @@
...
+      existing.sourceType = params.sourceType;
+      existing.sourcePackage = params.sourcePackage;
+      existing.logoUrl = params.logoUrl;
+      existing.websiteUrl = params.websiteUrl;
+      existing.termsUrl = params.termsUrl;
+      existing.latestAvailableVersion = params.latestAvailableVersion;
+      existing.isFeatured = params.isFeatured;
+      existing.marketplaceDisplayData = params.marketplaceDisplayData;
+
+      await this.appRegistrationRepository.save(existing);
```
prastoin: avoid mutation
```ts
      await this.appRegistrationRepository.save({
        ...existing,
        name: params.name,
        description: params.description,
        author: params.author,
        sourceType: params.sourceType,
        sourcePackage: params.sourcePackage,
        logoUrl: params.logoUrl,
        websiteUrl: params.websiteUrl,
        termsUrl: params.termsUrl,
        latestAvailableVersion: params.latestAvailableVersion,
        isFeatured: params.isFeatured,
        marketplaceDisplayData: params.marketplaceDisplayData,
      });

```

packages/twenty-server/src/engine/core-modules/application/application-marketplace/utils/get-admin-workspace-id.util.ts
```diff
@@ -0,0 +1,36 @@
...
+     LIMIT 1`,
+    [WorkspaceActivationStatus.ACTIVE],
+  );
+
+  if (result.length === 0) {
+    return null;
+  }
+
+  return result[0].workspaceId;
+};
```
prastoin, line 36: We should pursue this before merging, as said in comment this is too fragile
I feel like this should rather be a temporary backfill command behavior rather than a core engine one

Lets put in place a nullable col swap to incrementally release this, or hacky typeorm migration + upgrade command that applies the migration etc ( we can discuss that )

packages/twenty-server/src/engine/core-modules/application/application-marketplace/services/marketplace.service.ts
```diff
@@ -0,0 +1,90 @@
...
+          screenshots: [],
+          aboutDescription: description ?? '',
```
prastoin: should be nullable ?

packages/twenty-server/src/engine/core-modules/application/application-registration/application-registration.entity.ts
prastoin: I understand the need of a custom `foreignKey` naming 🤔

packages/twenty-server/src/engine/core-modules/application/application-marketplace/constants/marketplace-catalog-index.constant.ts
```diff
@@ -0,0 +1,179 @@
...
+      logicFunctions: [
+        {
+          name: 'create-new-post-card',
+          description: 'Creates a new postcard record',
+        },
+      ],
+      frontComponents: [],
+    },
+  },
+];
```
prastoin, line 179: As discussed in voc:
- Storing this in GitHub we might be loosing rich and interesting telemetry ( though we can use cloudfare )
- I'm wondering if we're okay with the fact not to store a copy of the installed tarball on a given workspace even from npm ? How do we do if the registry gets deleted

Are we aligned on ?:
In the future we might wanna have a dedicated microservice handling market place data source for both self hosted and cloud instances, that might also keep a copy of all related tarball ?

packages/twenty-sdk/src/cli/commands/app/app-push.ts
```diff
@@ -0,0 +1,68 @@
...
+    if (!buildResult.success) {
+      console.error(chalk.red(buildResult.error.message));
+      process.exit(1);
+    }
+
+    console.log(chalk.gray('Packing tarball...'));
+
+    const outputDir = path.join(appPath, '.twenty', 'output');
+
+    const packOutput = execSync('npm pack --pack-destination .', {
```
prastoin: ***Question:*** Why not using our own pack command here ? ( or programmatically calling the what could be public operation under app:pack )

packages/twenty-sdk/src/cli/utilities/api/api-service.ts
```diff
@@ -794,6 +800,85 @@ export class ApiService {
...
+    universalIdentifier?: string;
+  }): Promise<
+    ApiResponse<{
+      id: string;
+      universalIdentifier: string;
+      name: string;
+    }>
+  > {
+    try {
+      const mutation = `
```
prastoin, line 819: We should use the `MetadataClient` instead cc, dependent of https://github.com/twentyhq/core-team-issues/issues/2289

packages/twenty-sdk/src/cli/commands/app/app-push.ts
```diff
@@ -0,0 +1,68 @@
...
+    const result = await apiService.uploadAppTarball({ tarballBuffer });
+
+    if (!result.success) {
+      console.error(chalk.red(`Upload failed: ${result.error}`));
+      process.exit(1);
+    }
+
+    console.log(chalk.green('✓ Application pushed successfully'));
+  }
+}
```
prastoin, line 62: ***Remark:*** From my understanding this would never be consumed by a user that publish to npm ?
Maybe we should make this explicit in the cli naming

packages/twenty-server/src/engine/core-modules/application/dtos/application.dto.ts
```diff
@@ -64,6 +65,11 @@ export class ApplicationDTO {
   @Field(() => GraphQLJSON)
   availablePackages: Record<string, string>;
 
+  @IsOptional()
+  @IsUUID()
+  @Field(() => UUIDScalarType, { nullable: true })
+  applicationRegistrationId?: string;
```
prastoin, line 71: ***Challenge:*** Should this be nullable ? 
Maybe we should have a RESTRICT instead of set null for
```ts
  @ManyToOne(() => ApplicationRegistrationEntity, {
    onDelete: 'SET NULL',
    nullable: true,
  })
  @JoinColumn({ name: 'applicationRegistrationId' })
  applicationRegistration: Relation<ApplicationRegistrationEntity> | null;
```

I feel like an app should always have a registration
And a registration cannot be deleted until we've unistall all the related apps

packages/twenty-server/src/engine/core-modules/application/services/app-package-resolver.service.ts
```diff
@@ -0,0 +1,260 @@
...
+  ): Promise<void> {
+    const lines: string[] = [];
+    const registryHost = new URL(registryUrl).host;
+
+    if (packageName.startsWith('@')) {
+      const scope = packageName.split('/')[0];
+
+      lines.push(`${scope}:registry=${registryUrl}`);
+    } else if (registryUrl !== 'https://registry.npmjs.org') {
+      lines.push(`registry=${registryUrl}`);
```
prastoin: ***Question:*** Double check how this behave with already configured package.json registryConfigUrl, who takes precedence
```json
  "publishConfig": {
    "registry": "http://nexus.dsv.myhost/nexus/repository/npmjs"
  },
```

packages/twenty-server/src/engine/core-modules/application/services/app-package-resolver.service.ts
```diff
@@ -0,0 +1,260 @@
...
+        ApplicationExceptionCode.TARBALL_EXTRACTION_FAILED,
+      );
+    }
+  }
+
+  private async writeNpmrc(
+    workDir: string,
+    packageName: string,
+    registryUrl: string,
+    authToken?: string,
```
prastoin: ***polish:*** single record as arg

packages/twenty-server/src/engine/core-modules/application/application-install/app-package-fetcher.service.ts
```diff
@@ -0,0 +1,260 @@
...
+
+      const sourcePackage = appRegistration.sourcePackage;
+      const versionSpec = targetVersion ?? 'latest';
+
+      await this.writeNpmrc(workDir, sourcePackage, registryUrl, authToken);
+      await this.setupYarnEngine(workDir);
+      await this.writeMinimalPackageJson(workDir, sourcePackage, versionSpec);
+      await this.runYarnInstall(workDir);
+
+      const packageDir = join(workDir, 'node_modules', sourcePackage);
```
prastoin, line 115: ***Remark:*** Might wanna give a try to a direct api call maybe instead of this setup ? Feels quite hacky

packages/twenty-server/src/engine/core-modules/application/utils/read-json-file.util.ts
```diff
@@ -0,0 +1,25 @@
+import { promises as fs } from 'fs';
+import { join } from 'path';
+
+import {
+  ApplicationException,
+  ApplicationExceptionCode,
+} from 'src/engine/core-modules/application/application.exception';
+
+export const readJsonFile = async <T>(
```
prastoin, line 9: Naming `readJsonFileOrThrow`

packages/twenty-server/src/engine/core-modules/application/application-install/app-upgrade.service.ts
```diff
@@ -0,0 +1,163 @@
...
+      where: {
+        applicationRegistrationId: params.appRegistrationId,
+        workspaceId: params.workspaceId,
+      },
+    });
+
+    const previousVersion = application?.version ?? null;
+
+    try {
+      return await this.applicationInstallService.installApplication({
```
prastoin, line 113: Should handle the pre and post install logic function trigger here, can be done in a next PR

packages/twenty-server/src/engine/core-modules/application/services/app-upgrade.service.ts
```diff
@@ -0,0 +1,163 @@
...
+
+          this.logger.log(
+            `Rollback to ${previousVersion} succeeded for ${appName}`,
+          );
+        } catch (rollbackError) {
+          this.logger.error(
+            `Rollback also failed for ${appName}: ${rollbackError}`,
+          );
+        }
+      }
```
prastoin, line 126: There's no need to rollback as manifest sync runs in a transaction, if an exception if thrown this means nothing has been applied

Though this won't apply to the pre install and post install

packages/twenty-server/src/engine/core-modules/application/application-install/app-package-fetcher.service.ts
```diff
@@ -0,0 +1,260 @@
...
+      throw new ApplicationException(
+        'yarnPath not found in .yarnrc.yml',
+        ApplicationExceptionCode.PACKAGE_RESOLUTION_FAILED,
+      );
+    }
+
+    return join(workDir, match[1].trim());
+  }
+
+  private async runYarnInstall(workDir: string): Promise<void> {
```
prastoin, line 260: This feels off, even if a tmp dir this will have an impact on the shared yarn cache
This won't scale well

packages/twenty-server/src/engine/core-modules/application/services/app-package-resolver.service.ts
```diff
@@ -0,0 +1,260 @@
...
+      }
+
+      throw new ApplicationException(
+        `Failed to resolve tarball for app ${appRegistration.universalIdentifier}: ${error}`,
+        ApplicationExceptionCode.TARBALL_EXTRACTION_FAILED,
+      );
+    }
+  }
+
+  private async writeNpmrc(
```
prastoin: should be unit tested

packages/twenty-server/src/engine/core-modules/application-registration/services/app-tarball-upload.service.ts
```diff
@@ -0,0 +1,146 @@
...
+      let manifest: {
+        application?: {
+          universalIdentifier?: string;
+          displayName?: string;
+        };
+      };
+
+      try {
+        manifest = await readJsonFile(contentDir, 'manifest.json');
+      } catch {
```
prastoin: redundant catch, maybe should not throw natively

packages/twenty-server/src/engine/core-modules/application-registration/services/app-tarball-upload.service.ts
```diff
@@ -0,0 +1,146 @@
...
+        });
+
+        appRegistration =
+          await this.appRegistrationRepository.save(appRegistration);
+      }
+
+      const packageJson = await readJsonFile<{ version?: string }>(
+        contentDir,
+        'package.json',
+      ).catch(() => null);
```
prastoin: readJsonFile should have two variants, orThrow

prastoin (commented):
Did not reviwed the front !
A lot of job !

Few blockers to me:
- getAdminWorkspaceId
- data model nullable related comments
- npm tarball extract from yarn install

Others:
- npm code naming -> should this be registry agnostic ?

packages/twenty-server/src/engine/core-modules/application/application-registration/application-registration.entity.ts
FelixMalfait: I renamed to ownerWorkspaceId to avoid confusion because it's not workspaceScoped. The owner is this workspace but this entity is access from other workspaces, it's different from object, fields, etc which are really "workspace entities"

packages/twenty-server/src/engine/core-modules/application/application-marketplace/utils/get-admin-workspace-id.util.ts
FelixMalfait: I agree we should address this ; but I think we can still merge as long as there's a followup soon

FelixMalfait:
Recreating PR on a fresh branch to fix stuck GitHub Actions CI

PR #18365, opened 2026-03-03
fix google signup edge case

neo773:
Fixes an edge case when a user signs up with Google and the profile avatar network request times out, we crash instead of creating the user without an avatar.

Added `axios-retry` to retry max 2 times and if it still fails we gracefully skip avatar image instead of crashing

Fixes
Sentry TWENTY-SERVER-FDQ
Sonarly https://sonarly.com/issue/6564


FelixMalfait:

Your preview environment is available at: http://bore.pub:39676

packages/twenty-server/src/engine/core-modules/file/file-core-picture/services/file-core-picture.service.ts
```diff
@@ -187,6 +190,9 @@ export class FileCorePictureService {
   ): Promise<{ buffer: Buffer; extension: string } | undefined> {
     try {
       const httpClient = this.secureHttpClientService.getHttpClient();
+
+      axiosRetry(httpClient, { retries: 2, shouldResetTimeout: true });
```
charlesBochet: @neo773 I feel this should be part of the secureHttpClientService as it's a useful feature?

const httpClient = this.secureHttpClientService.getHttpClient({ retries: 0, shouldResetTimeout: true });

packages/twenty-server/src/engine/core-modules/file/file-upload/services/file-upload.service.ts
```diff
@@ -95,33 +98,49 @@ export class FileUploadService {
...
+
+      const type = await FileType.fromBuffer(buffer);
+
+      if (!type || !type.ext || !type.mime || !type.mime.startsWith('image/')) {
+        return undefined;
+      }
+
+      return { buffer, extension: type.ext, mimeType: type.mime };
+    } catch (error) {
+      this.logger.warn(
```
charlesBochet: I feel we should throw here not swallow. The caller can choose if it want to swallow or not

neo773:
@charlesBochet 
Pushed changes

packages/twenty-server/src/engine/core-modules/file/file-upload/services/file-upload.service.ts
```diff
@@ -95,31 +97,48 @@ export class FileUploadService {
...
-        'Unable to detect image type from buffer. The file may not be a valid image format.',
-      );
-    }
-
-    if (!type.mime.startsWith('image/')) {
-      throw new Error(
-        `Detected file type is not an image: ${type.mime}. Please provide a valid image URL.`,
-      );
+    if (!type || !type.ext || !type.mime || !type.mime.startsWith('image/')) {
+      return null;
```
charlesBochet: we should keep throwing here if there is something invalid (we should check if buffer is empty and in this case swallow but if not we should still check and throw based on type ext or mime)

charlesBochet (approved):
Approved with a last comment

PR #18367, opened 2026-03-03
Fixed scrollbar height issue in Kanban view and adjusted the calendar view to adjust with the new change

Lakshayyy-m:
fix for #18331 

## Issue 
The height of the container for the kanban board and calendar was set after calculating the offset from the top bar which contains the filters. The main issue was that it was calculated wrong. To fix this, I have added flex:1 to ensure that the board and calendar will grow into the empty space

## Proof of successful change
The video below shows that the scrollbar is accessible now and that the calendar view is also adjusted to not cause any errors because of the new change introduced.

https://github.com/user-attachments/assets/7f58342f-6cbf-4d30-878a-ec57f1e6666a


thomtrp (commented):
Hi @Lakshayyy-m, thank you for your contribution! I am a bit confused since the index page did not have the issue. I would have expected a fix in the RecordBoard. Why was it working in the index page then?

Lakshayyy-m:
HI @thomtrp , 
So if you have a look at the RecordIndexContainer.tsx, you will see that the index page is being rendered without the StyledContainerWithPadding component whereas the calendar and the kanban board are. That very component had the fixed height defined in it which I have now switched to dynamic. The fixed height was working alright when the board used to load because it accounted for the view bar at the top but broke down when the filter chips were added to the bar.

thomtrp (approved):
LGTM

Lakshayyy-m:
Thank you @thomtrp Just resolved the merge conflicts, should be good to merge now!

PR #18369, opened 2026-03-04
Increase size of input in test setting logic function tab

martmull:
## Before
image

## After

image


packages/twenty-front/src/modules/ai/components/CodeExecutionDisplay.tsx
```diff
@@ -275,6 +275,7 @@ export const CodeExecutionDisplay = ({
         >
           <StyledCodeEditorContainer>
             <CodeEditor
+              componentInstanceId={'code-execution-display'}
```
charlesBochet: component instance id should be a constant for re-usability

packages/twenty-front/src/modules/logic-functions/components/LogicFunctionExecutionResult.tsx
```diff
@@ -8,11 +8,11 @@ import { LogicFunctionExecutionStatus } from '~/generated-metadata/graphql';
 
 export const LogicFunctionExecutionResult = ({
   logicFunctionTestData,
-  maxHeight,
   isTesting = false,
+  componentInstanceId,
```
charlesBochet: do we really need to drill down here?

packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/input/components/RawJsonFieldInput.tsx
```diff
@@ -153,6 +153,7 @@ export const RawJsonFieldInput = () => {
       {isEditing ? (
         <StyledCodeEditorContainer>
           <CodeEditor
+            componentInstanceId={`code-editor-${instanceId}`}
```
charlesBochet: we should make an util here to

packages/twenty-front/src/modules/workflow/components/WorkflowStepExecutionResult.tsx
```diff
@@ -78,6 +77,7 @@ export const WorkflowStepExecutionResult = ({
   isTesting = false,
   loadingMessage = t`Processing...`,
   idleMessage = t`Output`,
+  componentInstanceId,
```
charlesBochet: we are drilling down to much and losing track of what this componentInstanceId is (here its a codeEditorComponentInstanceId for instance)

packages/twenty-ui/src/layout/resize-handle/components/ResizeHandle.tsx
```diff
@@ -0,0 +1,34 @@
+import { styled } from '@linaria/react';
+import { type PointerEvent } from 'react';
+import { themeCssVariables } from '@ui/theme-constants';
+
+type ResizeHandleProps = {
+  onPointerDown: (event: PointerEvent) => void;
+};
+
+const StyledResizeHandleArea = styled.div`
```
charlesBochet, line 12: I feel we are re-inventing the wheel here. We already have ReiszeHandle in the project, we should be re-using

packages/twenty-ui/src/input/code-editor/hooks/useResizableEditor.ts
```diff
@@ -0,0 +1,85 @@
...
+    return existing;
+  }
+
+  const heightAtom = atom(defaultHeight);
+  heightAtom.debugLabel = `resizableEditorHeight__${key}`;
+  heightAtomCache.set(key, heightAtom);
+  return heightAtom;
+};
+
+export const useResizableEditor = ({
```
charlesBochet: umbrella hook let's avoid

packages/twenty-ui/src/input/code-editor/hooks/useResizableEditor.ts
```diff
@@ -0,0 +1,85 @@
...
+    defaultHeight: initialHeight,
+  });
+
+  const [height, setHeight] = useAtom(heightAtom);
+
+  const [isResizing, setIsResizing] = useState(false);
+  const [resizeStartY, setResizeStartY] = useState(0);
+  const [resizeStartHeight, setResizeStartHeight] = useState(0);
+
+  useEffect(() => {
```
charlesBochet: we never put useEffect in hooks as this creates a nest fo re-renders.

Is the useEffect absolutely needed here? If yes, we need an Effect. It's not the first time we resize in the project, we should think of an approach for the whole project

packages/twenty-ui/src/input/code-editor/hooks/useResizableEditor.ts
```diff
@@ -0,0 +1,85 @@
+import { atom, useAtom, type PrimitiveAtom } from 'jotai';
+import { useEffect, useState } from 'react';
+import { isDefined } from 'twenty-shared/utils';
+
+const CODE_EDITOR_RESIZE_MIN_HEIGHT = 50;
+const CODE_EDITOR_RESIZE_MAX_HEIGHT = 500;
+
+const heightAtomCache = new Map<string, PrimitiveAtom<number>>();
+
+const getHeightAtom = ({
```
charlesBochet: why do we need cache here?

packages/twenty-ui/src/input/code-editor/hooks/useResizableEditor.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  const heightAtom = atom(defaultHeight);
+  heightAtom.debugLabel = `resizableEditorHeight__${key}`;
+  heightAtomCache.set(key, heightAtom);
+  return heightAtom;
+};
+
+export const useResizableEditor = ({
+  initialHeight,
+  componentInstanceId,
```
charlesBochet: if there is no plan to change the height from the outside we don't need an atom for that or at least we don't need to drill it down. We would usually do a ComponentInstanceContext

charlesBochet (commented):
@martmull Left comments :)
As you are off today, I'm taking it!

packages/twenty-ui/src/input/code-editor/components/CodeEditor.tsx
```diff
@@ -126,6 +130,18 @@ export const CodeEditor = ({
   >(undefined);
   const [isEditorFocused, setIsEditorFocused] = useState(false);
 
+  const numericHeight = typeof height === 'number' ? height : 450;
```
charlesBochet, line 133: use typeGuard isNumber

PR #18378, opened 2026-03-04
fix: use ForbiddenException in DevelopmentGuard to prevent Sentry noise

ehconitin:


etiennejouan (approved):
LGTM !

charlesBochet (approved):

packages/twenty-server/src/engine/guards/development.guard.ts
```diff
@@ -15,7 +19,7 @@ export class DevelopmentGuard implements CanActivate {
       nodeEnv !== NodeEnvironment.DEVELOPMENT &&
       nodeEnv !== NodeEnvironment.TEST
     ) {
-      throw new Error(
```
charlesBochet, line 18: is this the standard in guards to throw ForbiddenException (rest)? I'm asking because if they are used by resolver, this might not be properly converted to Graphql error:
- check other guards pattern
- check what we get in the http graphql response

ehconitin: - NoImpersonationGuard -- throws ForbiddenException
- SettingsPermissionGuard / ImpersonatePermissionGuard -- throw PermissionsException (custom)
- FeatureFlagGuard -- throws new Error() (has the same bug)
  
You were right to flag this -- ForbiddenException was ***NOT*** being properly converted!

fixed it

PR #18384, opened 2026-03-04
Replace align-center with padding

thomtrp:
Toggle using align-self prevents the use of align-items

Before
Capture d’écran 2026-03-04 à 16 11 22

After
Capture d’écran 2026-03-04 à 16 11 14


Weiko (approved):
LGTM

PR #18385, opened 2026-03-04
Fix invalid universal identifier format command cache flush

prastoin:
# Introduction
Invalidating command impacted metadata and related metadata caches entries


Devessier (approved):
Thanks

PR #18392, opened 2026-03-04
Stop catching all workflow errors

thomtrp:
Steps now throw WorkflowStepExecutorException. Then workflow executor decides if error should be catch or not.

Since tools are not only used in workflow and these do not throw, we may still miss errors here.

Workflow jobs now only catch errors to end the workflow run and throw.


packages/twenty-server/src/modules/workflow/workflow-executor/workflow-actions/ai-agent/ai-agent.workflow-action.ts
```diff
@@ -56,60 +52,47 @@ export class AiAgentWorkflowAction implements WorkflowAction {
...
-          agent,
-          userPrompt: resolveInput(prompt, context) as string,
-          actorContext: executionContext.isActingOnBehalfOfUser
-            ? executionContext.initiator
-            : undefined,
-          rolePermissionConfig: executionContext.rolePermissionConfig,
-          authContext: executionContext.authContext,
-        });
+    if (agentId) {
+      agent = await this.agentRepository.findOne({
```
Weiko, line 58: We could query the cache FYI.

charlesBochet (approved):

PR #18411, opened 2026-03-05
Set widget position's type according to parent tab

Devessier:
Fixes workspaces seeded a few weeks ago and containing position=NULL widgets


Weiko (approved):
LGTM

PR #18414, opened 2026-03-05
[Fix] : Dragged element is considered to be part of a dropdown in dashboard tab list

### LINKED ISSUE #15327: Dragged element is considered to be part of a dropdown in dashboard tab list
Follow-up for https://github.com/twentyhq/twenty/pull/14798

**This seems to only happen when dragging the last element of the tab list.**

https://github.com/user-attachments/assets/fd691971-0849-4349-afba-b9e7f35cd0c2

carbonFibreCode:
Fixes #15327

The issue occurred because the drag clone's visual state was previously tied strictly to hovering over the `VISIBLE_TABS` boundaries. When a tab was dragged outside this area (such as the last tab naturally crossing into the `MORE_BUTTON` hover zone), the drag clone incorrectly fell back to the dropdown menu item style.

We fixed this by making the `isHoveringTabList` logic more robust. Instead of enforcing the tab style only within the `VISIBLE_TABS` boundary, the dropdown style is now strictly restricted to the `OVERFLOW_TABS` boundary.

With this change:
- Visible tabs successfully maintain their appearance when dragged anywhere outside the dropdown.
- Dropdown tabs correctly transition to the normal tab style when dragged out of the dropdown area, improving UX.

https://github.com/user-attachments/assets/9474e4c1-26a8-46e3-b9ee-4c6dbd8a4ea6


ehconitin (approved):
thanks for the PR @carbonFibreCode !!

I resolved the conflicts and pushed a more simpler mental model which grep proposed

ie the drag clone should only render as a dropdown menu item when hovering over the overflow dropdown. Everywhere else (tab bar, more button, or outside all zones), it should look like a tab.

carbonFibreCode:
> ie the drag clone should only render as a dropdown menu item when hovering over the overflow dropdown. Everywhere else (tab bar, more button, or outside all zones), it should look like a tab.

@ehconitin  yes i checked it, following a complementary check is indeed more clean and straight forward.

lucasbordeau:
Nice work

PR #18423, opened 2026-03-05
Fix server logs leak

prastoin:
# Introduction

Previously the auth jwt stragegy would lod the whole user entity in the auth user context
On an exception it would completely get logged on the pods

## Security layer
- 0/ Updating the type system ( devxp only though )
- 1/ The jwt auth stragegy only load a specific sub set of the user entity
- 2/ Sanitizing at the exception log level directly in case of a user context
- 3/ Sanitizing at the console driver

The last two sanitization could sound a bit redundant though they're still good fallback to keep in case new path occurs in the cb


charlesBochet (approved):

prastoin:
Nice catch @greptileai fixing the relation load

Weiko (approved):
LGTM

PR #11984, opened 2025-05-12
Implemented LINKS and EMAILS sub-field fitering

lucasbordeau:
This PR introduces LINKS and EMAILS sub-field filtering. It's mainly about the implementation of secondaryLinks and additionalEmails sub-fields, which are treated like additionalPhones.

There's also a refactor on the computeViewRecordGqlOperationFilter, a big file that becomes very difficult to read and maintain. This PR breaks it down into multiple smaller utils. There's still work to be done to clean it as it is a central part of the record filter module, this PR lays the foundation.

PR #9604, opened 2025-01-14
Moving record filter related types, hooks and utils into record-filter module

lucasbordeau:
This PR is only moving and renaming types, hooks and utils to record-filter module folder.

- Moved and renamed types from object filter modules to record filter…- Moved and renamed types from object filter modules to record filter module
- Moved useApplyRecordFilter to record filter module
- Renamed util getOperandsForFilterDefinition to getRecordFilterOperandsForRecordFilterDefinition

PR #18426, opened 2026-03-05
feat: add IS/IS_NOT option in filters

### LINKED ISSUE #2314: Filter with exact search - Add "IS" option instead of "CONTAINS" in the Filters
## Scope & Context

Some users are using the Filter instead of the Search (this is not corner case). 
When looking for a specific firstName + lastName, the filter contains returns all the matches with the firstName at first. This means that the specific firstName + lastName can be returned (for ex) line 30. This is the expected behavior bc of the contains. 
From a user's perspective, it looks off to not have the exact match returned at the very top - think about the cases where the first name is a very common one, the issue is likely to keep happening. 

This could be solved by offering to filter with IS instead of CONTAINS, per my discussion with @charlesBochet

bugisthegod:
### Summary

Add IS/IS_NOT option for Full name, text, phone, Links , email. Fixes twentyhq/core-team-issues#2314 

In the advanced filter, only show is/is_not option in phone (primaryPhoneNumber), Links (primaryLinkUrl), email (primaryEmail).

https://github.com/user-attachments/assets/54f615d6-1cfb-4560-8f00-a513a74abdcf

https://github.com/user-attachments/assets/8797bd15-f6df-40f6-a2c8-399a7096fe3f

Note: test file getOperandsForFilterType.test.ts is misplaced  (should be under record-filter/utils/__tests__/) . It seems it was forgotten when moving — leftover from twentyhq/twenty#9604 .

### Question For the Maintainer

1. Should DOES_NOT_CONTAIN include records where the field is NULL? Currently it doesn't — only records with an existing value that doesn't match are returned. I'm making IS_NOT include NULLs, so wanted to confirm the intended behavior for consistency.

2. Debounce problem. When a user types quickly (e.g. "Google"), the filter fires on every keystroke, causing results to momentarily include values that should be excluded. Should we add debounce to filter inputs?

https://github.com/user-attachments/assets/d584f03e-d942-437e-8244-88bcfa50b0e6

3. Should we extract the PHONES, full_name, address filter logic into its own util (like computeGqlOperationFilterForPhones.ts, based on PR twentyhq/twenty#11984 which want more readable and maintainable),  consistent with how EMAILS and LINKS are structured? If yes, submit it in this pr or open a separate PR for that.


bugisthegod:
Hi @Bonapara , could you take a look at this PR when you get a chance? Let me know if anything needs to be changed. Thanks!

lucasbordeau:
I'll check if that is corresponding to the initial need, after reading the issue and testing it, I'm not sure.

lucasbordeau:
I'm closing this PR as currently the requirements are not clear and it seems to be a low priority issue, I'll put the relevant context in the issue so that we can take back the work later on.

PR #18434, opened 2026-03-05
Simplify z-index logic and enhance portal styling (spec-kit experiment)

lucasbordeau:
This PR as an experiment with spec-kit, it is a refactor with a complex understanding with small but precise modification to make.

The testing part was manual and handed to the developer, between each phase.

It successfuly removes the complex logic that was used to manage z-index switching to have the hovered cell portal correctly behave when its borders were overlaping cells ones. 

We now have the hovered portal inside a cell, thus removing the need for a z-index dynamic logic.

The code has been simplified in the parts where the logic was implemented and the constant that holds the all z indices for the tables is still needed but with less options.


lucasbordeau:
Closing for documenting the experiment.

PR #18441, opened 2026-03-05
Add default viewField when creating object

martmull:
as title


packages/twenty-sdk/src/cli/commands/entity/entity-add.ts
```diff
@@ -199,9 +201,20 @@ export class EntityAddCommand {
     const viewUniversalIdentifier = v4();
 
     const viewFile = getViewBaseFile({
-      name: `all-${kebabCase(objectName)}`,
+      name: `all-${kebabcase(objectName)}`,
       universalIdentifier: viewUniversalIdentifier,
       objectUniversalIdentifier: this.lastObjectUniversalIdentifier,
+      fields: this.lastNameFieldUniversalIdentifier
```
charlesBochet, line 209: I'm confused, why is lastName special? shouldn't we add the labelIdentifier?

packages/twenty-sdk/src/cli/utilities/entity/entity-view-template.ts
```diff
@@ -1,17 +1,57 @@
...
+      parts.push(`      isVisible: ${field.isVisible}`);
+    }
+    if (field.size !== undefined) {
+      parts.push(`      size: ${field.size}`);
+    }
+    return `    {\n${parts.join(',\n')},\n    }`;
+  })
+  .join(',\n')}
+  ],`
+      : `  // fields: [
```
charlesBochet: devXP is not great here

packages/twenty-sdk/src/cli/utilities/entity/entity-view-template.ts
```diff
@@ -1,17 +1,57 @@
...
+    const uid = field.universalIdentifier ?? v4();
+    const parts = [
+      `      universalIdentifier: '${uid}'`,
+      `      fieldMetadataUniversalIdentifier: '${field.fieldMetadataUniversalIdentifier}'`,
+      `      position: ${field.position}`,
+    ];
+    if (field.isVisible !== undefined) {
+      parts.push(`      isVisible: ${field.isVisible}`);
+    }
+    if (field.size !== undefined) {
```
charlesBochet: use isDefined

packages/twenty-sdk/src/cli/commands/entity/entity-add.ts
martmull: we always use name as labelIdentifier

packages/create-twenty-app/src/create-app.command.ts
```diff
@@ -183,16 +183,36 @@ export class CreateAppCommand {
...
+    const includeNavigationMenuItem =
+      selectedExamples.includes('navigationMenuItem');
     const includeExampleIntegrationTest =
       selectedExamples.includes('integrationTest');
+    const includeView =
+      selectedExamples.includes('view') || includeNavigationMenuItem;
     const includeObject =
-      selectedExamples.includes('object') || includeField || includeView;
+      selectedExamples.includes('object') ||
+      includeField ||
```
charlesBochet: Why do we want the create-app to be that complex, I would say it's the job en entity-add
create-twenty-app should make something simple but not so many templates

If we want to keep templates, let's be more straight forward (one instruction file per template), instead of mixing them all

Code quality: I don't see this scaling well, complex conditions I feel we should have a clearer pattern
entity:add object --with-view --with-navigation-item

and that it should be a nice switch with different files (create very straight forward

packages/twenty-sdk/src/cli/utilities/entity/entity-view-template.ts
```diff
@@ -1,17 +1,56 @@
...
+  const defaultFields = `  // fields: [
+  //   {
+  //     universalIdentifier: '...',
+  //     fieldMetadataUniversalIdentifier: '...',
+  //     position: 0,
+  //     isVisible: true,
+  //   },
+  // ],`;
+
+  const fieldsBlock =
```
charlesBochet, line 45: long terme we should find a clearer way with templating

packages/twenty-server/src/engine/core-modules/application/application-manifest/converters/from-view-manifest-to-universal-flat-view.util.ts
```diff
@@ -28,7 +29,7 @@ export const fromViewManifestToUniversalFlatView = ({
     isCustom: true,
     visibility: viewManifest.visibility ?? ViewVisibility.WORKSPACE,
     openRecordIn: viewManifest.openRecordIn ?? ViewOpenRecordIn.SIDE_PANEL,
-    key: null,
+    key: ViewKey.INDEX,
```
charlesBochet: I don't think we should default to Index, this should be in the manifest

charlesBochet (approved):
I don't think this approach scales well from a quality perspective, I've left comment. I'm approving to unblock but I would recommend to re-work

packages/twenty-server/src/engine/core-modules/application/application-manifest/converters/from-view-manifest-to-universal-flat-view.util.ts
charlesBochet: @martmull this is blocking to me

packages/create-twenty-app/src/create-app.command.ts
martmull: you already have minimal, interactive or exhausitve options

packages/twenty-server/src/engine/core-modules/application/application-manifest/converters/from-view-manifest-to-universal-flat-view.util.ts
martmull: i don't really know what this is. But this is the only option in ViewKey, and all other seeded key are seeded with this value. Setting null makes the navigator item ugly

charlesBochet: this should be properly set on he manifest build

charlesBochet (approved):

PR #18430, opened 2026-03-05
Remove all styled(Component) patterns in favor of parent wrappers and props

charlesBochet:
## Summary

Eliminates all ~350 `styled(Component)` usages across `twenty-front` and `twenty-ui` (212 files changed). Each was replaced following these rules:

- **Margin/layout CSS** (margin, padding, flex, align-self, width) → wrapped in a `styled.div`/`styled.span` parent container
- **Third-party components** (Link, TextareaAutosize, ReactPhoneNumberInput, Handle, etc.) → parent container with child CSS selectors (`> a`, `> textarea`, `> input`, etc.)
- **Intrinsic behavior via existing props** (TableRow `gridTemplateColumns`, TableCell `color`/`align`) → replaced `styled(TableRow)` / `styled(TableCell)` with direct prop usage
- **Other visual overrides on twenty-ui components** (Card, Section, TabList, Button, MenuItem, ScrollWrapper, etc.) → parent wrappers with `> div` / `> *` child selectors
- **Extending styled.div/span** → merged all CSS into a single `styled.div`/`styled.span`

Also adds `overflow: hidden` to parent containers wrapping `ScrollWrapper` so scroll activates correctly with the new wrapper structure.

### Migration patterns

| Before | After |
|--------|-------|
| `styled(Avatar)` with `margin-right` | `<StyledAvatarContainer><Avatar /></StyledAvatarContainer>` |
| `styled(Link)` with `text-decoration: none` | `<StyledLinkContainer><Link /></StyledLinkContainer>` with `> a { ... }` |
| `styled(TableRow)` with `grid-template-columns` | `<TableRow gridTemplateColumns="..." />` |
| `styled(TableCell)` with `color` / `align` | `<TableCell color={...} align="right" />` |
| `styled(Card)` with `margin-top` | `<StyledCardContainer><Card /></StyledCardContainer>` |
| `styled(TabList)` with `background` | `<StyledTabListContainer><TabList /></StyledTabListContainer>` with `> div { ... }` |
| `styled(StyledBase)` extending a `styled.div` | Single merged `styled.div` with all styles inlined |

PR #18448, opened 2026-03-05
fix: remove add record on workflow runs/versions

charlesBochet:
## Summary

- Disable manual record creation (add button, + header button, add new row) for **WorkflowRun** and **WorkflowVersion** objects since these are system-managed and should not be created manually
- Fix vertical centering of the record table empty state placeholder (regression from `styled(Component)` refactor in #18430 — the wrapper lost `height: 100%` / `width: 100%`)

## Test plan

- [ ] Navigate to Workflow Runs index page → empty state should show centered placeholder **without** "Add a Workflow Run" button
- [ ] Navigate to Workflow Versions index page → empty state should show centered placeholder **without** "Add a Workflow Version" button
- [ ] Navigate to any other object index page (e.g. People, Companies) → empty state should still show the "Add a ..." button and be centered
- [ ] Verify the + button in the record table header is hidden for workflow runs/versions
- [ ] Verify the "Add New" row at the bottom of the table is hidden for workflow runs/versions

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-table/utils/isRecordTableCreateDisabled.ts
+++ b/packages/twenty-front/src/modules/object-record/record-table/utils/isRecordTableCreateDisabled.ts
@@ -1,0 +1,12 @@
+import { CoreObjectNameSingular } from 'twenty-shared/types';
+
+const OBJECTS_WITHOUT_MANUAL_RECORD_CREATION: string[] = [
+  CoreObjectNameSingular.WorkflowRun,
+  CoreObjectNameSingular.WorkflowVersion,
+];
+
+export const isRecordTableCreateDisabled = (
+  objectNameSingular: string,
+): boolean => {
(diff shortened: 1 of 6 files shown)


packages/twenty-front/src/modules/object-record/record-table/utils/isRecordTableCreateDisabled.ts
```diff
@@ -0,0 +1,12 @@
+import { CoreObjectNameSingular } from 'twenty-shared/types';
+
+const OBJECTS_WITHOUT_MANUAL_RECORD_CREATION: string[] = [
+  CoreObjectNameSingular.WorkflowRun,
+  CoreObjectNameSingular.WorkflowVersion,
+];
+
+export const isRecordTableCreateDisabled = (
```
charlesBochet, line 6: this likely need to be revisited when we surface all objects, what's the product rule here

PR #18449, opened 2026-03-06
Improve build performance 2x

charlesBochet:
## Summary

Front Before:
image

Front After:
image

Server Before:
image

Server After:
image

### CI Server Pipeline Restructuring
- Split monolithic `server-setup` job into three parallel jobs: `server-build`, `server-lint-typecheck`, and `server-validation`
- `server-build` only handles build + Nx cache save (~1m vs old 3.5m), unblocking downstream jobs faster
- `server-lint-typecheck` runs in parallel with no DB dependency
- `server-validation` handles DB setup, migration checks, and GraphQL generation checks in parallel with tests
- Make `server-test` (unit tests) fully independent — no longer waits for server-setup, builds its own artifacts
- Increase integration test shards from 8 to 10 for better parallelism
- Expected critical path reduction: ~10m → ~7m (~30% faster)

### CI Front Pipeline Improvements
- Use artifact upload/download for storybook build instead of rebuilding in test shards
- Serve pre-built storybook via `http-server` in test jobs, with `STORYBOOK_URL` env var
- Update `vitest.config.ts` to use `storybookUrl` when `STORYBOOK_URL` is set
- Remove redundant `twenty-shared`, `twenty-ui`, `twenty-sdk` builds from storybook test shards

### Vite Build Optimizations
- Conditionally enable `rollup-plugin-visualizer` behind `ANALYZE=true` env var (not loaded by default)
- Broaden Istanbul coverage exclusions to skip test files, stories, mocks, and decorators
- Remove `@tabler/icons-react` alias from twenty-front and storybook configs
- Bundle `@tabler/icons-react` into twenty-ui instead of treating it as an external dependency
- Add lazy loading with `React.lazy` + `Suspense` for all page-level route components in `useCreateAppRouter`


packages/twenty-front/vite.config.ts
```diff
@@ -105,19 +110,24 @@ export default defineConfig(({ mode }) => {
...
-      visualizer({
-        open: true,
-        gzipSize: true,
-        brotliSize: true,
-        filename: 'dist/stats.html',
-      }) as PluginOption, // https://github.com/btd/rollup-plugin-visualizer/issues/162#issuecomment-1538265997,
+      ...(env.ANALYZE === 'true'
+        ? [
+            visualizer({
+              open: true,
```
Weiko: should probably be false for a CI? According to the doc here https://www.npmjs.com/package/rollup-plugin-visualizer it says it opens a browser

packages/twenty-front/src/modules/app/hooks/useCreateAppRouter.tsx
```diff
@@ -43,26 +114,111 @@ export const useCreateAppRouter = (
...
+            element={
+              <Suspense fallback={null}>
+                <SignInUp />
+              </Suspense>
+            }
+          />
+          <Route
+            path={AppPath.Invite}
+            element={
+              <Suspense fallback={null}>
```
Weiko: should we have a fallback there?

.github/workflows/ci-server.yaml
```diff
@@ -137,25 +175,21 @@ jobs:
             echo ""
             exit 1
           fi
-      - name: Save server setup
-        uses: ./.github/actions/save-cache
-        with:
-          key: ${{ steps.restore-server-setup-cache.outputs.cache-primary-key }}
+
   server-test:
+    needs: changed-files-check
```
Weiko: should server-test need server-build here?

packages/twenty-front/vite.config.ts
```diff
@@ -236,7 +246,6 @@ export default defineConfig(({ mode }) => {
     resolve: {
       alias: {
         path: 'rollup-plugin-node-polyfills/polyfills/path',
-        '@tabler/icons-react': '@tabler/icons-react/dist/esm/icons/index.mjs',
```
Weiko, line 239: do we have some eslint rule to enforce that we'll never import from there already?

Weiko (approved):

packages/twenty-front/src/modules/app/hooks/useCreateAppRouter.tsx
charlesBochet: Good idea!

.github/workflows/ci-server.yaml
```diff
@@ -58,18 +106,12 @@ jobs:
...
-        id: restore-server-setup-cache
+      - name: Restore server build cache
         uses: ./.github/actions/restore-cache
         with:
-          key: ${{ env.SERVER_SETUP_CACHE_KEY }}
+          key: ${{ env.SERVER_BUILD_CACHE_KEY }}
       - name: Build twenty-shared
         run: npx nx build twenty-shared
-      - name: Server / Run lint & typecheck
-        uses: ./.github/actions/nx-affected
```
prastoin, line 69: thanks for that

.github/workflows/ci-server.yaml
charlesBochet: indeed

prastoin (approved):
nice

.github/workflows/ci-server.yaml
```diff
@@ -27,11 +27,59 @@ jobs:
...
+        run: npx nx build twenty-shared
+      - name: Server / Write .env
+        run: npx nx reset:env twenty-server
+      - name: Server / Build
+        run: npx nx build twenty-server
+      - name: Save server build cache
+        uses: ./.github/actions/save-cache
+        with:
+          key: ${{ steps.restore-server-build-cache.outputs.cache-primary-key }}
+
```
prastoin, line 58: I don't think this will behave as you think it will

prastoin:
```
name: Save cache
inputs:
  key:
    required: true
    description: Primary key to the cache, should be retrieved from `cache-restore` composite action outputs.
  additional-paths:
    required: false

runs:
  using: "composite"
  steps:
    - name: Save cache
      uses: actions/cache/save@v4
      with:
        key: ${{ inputs.key }}
        path: |
          .cache
          .nx
          node_modules/.cache
          packages/*/node_modules/.cache
          ${{ inputs.additional-paths }}
```

prastoin: should we add the build/dist folder of the twenty-server in additional path here ?
Unless I'm mistaken we're not adding anything new to the cache ?

packages/twenty-front/vite.config.ts
charlesBochet: it's unrelated, it's imported behind the scene, I have made sure twenty-ui is properly bundliing it

packages/twenty-front/vite.config.ts
charlesBochet: good point

.github/workflows/ci-server.yaml
charlesBochet: taking a look

charlesBochet: I think it works because all nx build output are stored in .nx/cache
so when you try to rebuild, it immediately populate twenty-server/dist from .nx/cache

prastoin: What I was trying to say is that in the end it's iso with the previous implem

prastoin: But totally fine with it, just wanted to double check with you

PR #18456, opened 2026-03-06
Fix workflow nodes color in dark mode

bugisthegod:
## Why
  `ThemeProvider` already exposes `colorScheme` (`'light' | 'dark'`) via  `ThemeContext`, but `WorkflowDiagramCanvasBase` was only extracting`theme` and never passing `colorMode` to `ReactFlow`.
Fix: #18453 

## Before
image

## After

image


Weiko (approved):
LGTM, thanks!

PR #18457, opened 2026-03-06
Bug fixes batch

### LINKED ISSUE #16842: Can't stop running workflowRun
## Bug Description

I believe it used to be possible to stop a workflowRun in status in Running. Should be a straight forward fix unless this uncovers a specific race condition

thomtrp:
Fixes https://github.com/twentyhq/twenty/issues/18181

Fixes https://github.com/twentyhq/twenty/issues/16842
Iterators remain running, which prevent the stopping state to eventually become stopped

Fixes https://github.com/twentyhq/twenty/issues/18186

### DIFF
--- a/packages/twenty-server/src/modules/workflow/workflow-runner/workspace-services/workflow-runner.workspace-service.ts
+++ b/packages/twenty-server/src/modules/workflow/workflow-runner/workspace-services/workflow-runner.workspace-service.ts
@@ -16,7 +16,7 @@ import {
   WorkflowVersionStepExceptionCode,
 } from 'src/modules/workflow/common/exceptions/workflow-version-step.exception';
 import { WorkflowRunStatus } from 'src/modules/workflow/common/standard-objects/workflow-run.workspace-entity';
-import { buildIteratorToStopStepInfos } from 'src/modules/workflow/common/utils/build-iterator-to-stop-step-infos.util';
+import { setAllIteratorsStepInfosAsStopped } from 'src/modules/workflow/common/utils/set-all-iterators-step-infos-as-stopped.util';
 import { workflowHasRunningSteps } from 'src/modules/workflow/common/utils/workflow-has-running-steps.util';
 import { WorkflowCommonWorkspaceService } from 'src/modules/workflow/common/workspace-services/workflow-common.workspace-service';
 import { WorkflowVersionStepOperationsWorkspaceService } from 'src/modules/workflow/workflow-builder/workflow-version-step/workflow-version-step-operations.workspace-service';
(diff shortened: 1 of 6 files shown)


packages/twenty-server/src/modules/workflow/workflow-runner/workspace-services/workflow-runner.workspace-service.ts
```diff
@@ -218,17 +219,29 @@ export class WorkflowRunnerWorkspaceService {
...
+
       await this.workflowRunWorkspaceService.updateWorkflowRun({
         workflowRunId,
         workspaceId,
         partialUpdate: {
           status: WorkflowRunStatus.STOPPING,
+          state: {
+            ...workflowRun.state,
+            stepInfos: mergedStepInfos,
+          },
```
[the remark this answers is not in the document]
thomtrp: This is an acceptable behavior. Iterator will not update records so we prefer having it marked as stopped a bit earlier rather than keeping it running forever

packages/twenty-ui/src/components/chip/Chip.tsx
```diff
@@ -197,7 +199,7 @@ export const Chip = ({
       maxWidth={maxWidth}
     >
       {leftComponent}
-      {!isLabelHidden && label && label.trim() ? (
+      {!isLabelHidden && isDefined(label) && isNonEmptyString(label) ? (
```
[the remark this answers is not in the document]
thomtrp: that's intended

packages/twenty-server/src/modules/workflow/workflow-runner/workspace-services/workflow-runner.workspace-service.ts
```diff
@@ -218,17 +219,29 @@ export class WorkflowRunnerWorkspaceService {
...
-      workflowHasRunningSteps({
-        stepInfos: workflowRun.state.stepInfos,
-        steps: workflowRun.state.flow.steps,
-      })
-    ) {
+    const stepInfos = workflowRun.state.stepInfos;
+    const steps = workflowRun.state.flow.steps;
+
+    if (workflowHasRunningSteps({ stepInfos, steps })) {
+      const stoppedIteratorStepInfos = getStoppedIteratorStepInfos({
```
charlesBochet: naming improve: get means that we are not changing. Here we are mutating.
setAllIteratorsStepInfosAsStopped (is it really what we want btw?)

thomtrp: I renamed it but that's not a mutation yet, we are just building the infos.
That's what we want because current stoppage only works when there are no running steps left. But an iterator remain running until it has been called by its last iteration. Which will never happen if the iteration steps have been stopped.

charlesBochet: better still not happy with naming :p is the iterator the one stopping? I prefer: setAllIteratorsStepInfosAsStopped

charlesBochet: approving though, non blocking

charlesBochet (approved):

PR #18463, opened 2026-03-06
feat(twenty-ui): add Skeleton component system with tests

surfer19:
- Add Skeleton, SkeletonText, and SkeletonRow components to twenty-ui
- Pure CSS shimmer animation (zero extra dependencies vs react-loading-skeleton)
- Export from twenty-ui/feedback barrel
- Add Storybook stories for all three components
- Add unit tests (20 tests across 3 suites)
- Refactor DropdownMenuSkeletonItem and ShowPageSummaryCard to use new Skeleton


charlesBochet:
Hey @surfer19, thanks a lot for putting this together, and sorry for the delay.

A couple of things happened since you opened it. We reworked our UI packages: the old twenty-ui is now twenty-ui-deprecated and a new twenty-ui has landed, migrated token for token with no visual change. That's why this branch now conflicts.

Also, this swaps react-loading-skeleton for a custom component in only 2 of the ~20 places we use it, so we'd end up running two skeleton systems side by side, which we'd rather not do. We don't have anything tracking a move off react-loading-skeleton on our side right now either.

For those reasons I'm going to close this one. Really appreciate the effort though, and we'd be happy to have you back. If you want, pick up a good first issue and I'll gladly review.

PR #18466, opened 2026-03-06
Table refactor : removed z-index dynamic logic completely and flex-wrap

lucasbordeau:
This PR removes the leftovers from the z-index dynamic logic removal.

It also removes the flex-wrap mechanism that was used to have all the cells in the same div, and instead creates a container for each part of the table : header, body and footer, so that z-index management becomes straighforward.

We also fix some minor bugs.

## Demo 

https://github.com/user-attachments/assets/29dc4966-376d-4eb1-9e37-99769e77f4f4

https://github.com/user-attachments/assets/78218517-812a-4531-84c3-067700b46b59


bosiraphael (approved):
LGTM

PR #18394, opened 2026-03-04
Fix one-to-many relation attach crash on Apollo cache miss

neo773:
fixes sentry TWENTY-FRONT-4NZ

Fixes https://github.com/twentyhq/twenty/pull/18394

PR #18477, opened 2026-03-06
Enable update when record not in cache

thomtrp:
Fixes https://github.com/twentyhq/twenty/issues/17669

I may be wrong but I don't get why we absolutely need the cache to perform the update

### DIFF
--- a/packages/twenty-front/src/modules/object-record/hooks/useRecordOneToManyFieldAttachTargetRecord.ts
+++ b/packages/twenty-front/src/modules/object-record/hooks/useRecordOneToManyFieldAttachTargetRecord.ts
@@ -60,44 +60,43 @@ export const useRecordOneToManyFieldAttachTargetRecord = () => {
       objectPermissionsByObjectMetadataId,
     });
 
-    if (!cachedTargetRecord) {
-      throw new Error('Could not find cached related record');
-    }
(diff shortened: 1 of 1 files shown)


packages/twenty-front/src/modules/object-record/hooks/useRecordOneToManyFieldAttachTargetRecord.ts
```diff
@@ -60,44 +60,43 @@ export const useRecordOneToManyFieldAttachTargetRecord = () => {
...
-      updateRecordFromCache({
-        objectMetadataItems,
-        objectMetadataItem: targetObjectMetadataItem,
-        cache: apolloCoreClient.cache,
-        record: {
+    if (cachedTargetRecord) {
+      const previousRecordId =
+        cachedTargetRecord[`${targetGQLFieldName}Id`];
+
+      if (isDefined(previousRecordId)) {
```
charlesBochet, line 67: @thomtrp Huzef came with the same fix. I'm not fully against it but I don't understand what functional path make it happen. How do you reproduce it?

It's likely hiding something deeper

charlesBochet:
@thomtrp Closing this one as we already have: https://github.com/twentyhq/twenty/pull/18394

Let's work on the other one :)

PR #18478, opened 2026-03-06
[Fix] : NOT filter not supported for record groups

### LINKED ISSUE #14746: NOT filter not supported for record groups
## Bug Description

Image

In the code above, we implicitly convert "NOT" to "OR", while they obviously has different meaning.
Looking at turnRecordFilterGroupsIntoGqlOperationFilter, NOT is currently not supported, while it's a possible value in the db.

carbonFibreCode:
Fixes #14746 

Screenshot 2026-03-07 at 2 03 14 AM

### DIFF
--- a/packages/twenty-front/src/modules/views/utils/mapViewFilterGroupLogicalOperatorToRecordFilterGroupLogicalOperator.ts
+++ b/packages/twenty-front/src/modules/views/utils/mapViewFilterGroupLogicalOperatorToRecordFilterGroupLogicalOperator.ts
@@ -7,7 +7,14 @@ export const mapViewFilterGroupLogicalOperatorToRecordFilterGroupLogicalOperator
   }: {
     viewFilterGroupLogicalOperator: ViewFilterGroupLogicalOperator;
   }) => {
-    return viewFilterGroupLogicalOperator === ViewFilterGroupLogicalOperator.AND
-      ? RecordFilterGroupLogicalOperator.AND
-      : RecordFilterGroupLogicalOperator.OR;
+    return (
+      {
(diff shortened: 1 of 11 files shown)


packages/twenty-front/src/modules/views/utils/mapViewFilterGroupLogicalOperatorToRecordFilterGroupLogicalOperator.ts
```diff
@@ -7,7 +7,14 @@ export const mapViewFilterGroupLogicalOperatorToRecordFilterGroupLogicalOperator
...
-      : RecordFilterGroupLogicalOperator.OR;
+    return (
+      {
+        [ViewFilterGroupLogicalOperator.AND]:
+          RecordFilterGroupLogicalOperator.AND,
+        [ViewFilterGroupLogicalOperator.OR]:
+          RecordFilterGroupLogicalOperator.OR,
+        [ViewFilterGroupLogicalOperator.NOT]:
+          RecordFilterGroupLogicalOperator.NOT,
+      }[viewFilterGroupLogicalOperator] ?? RecordFilterGroupLogicalOperator.AND
```
lucasbordeau: Let's use a switch case with throw assertUnreachable instead

lucasbordeau:
@Bonapara The UI is confusing for this requirement, we want NOT in advanced filter, but how should we do it ? 

image

lucasbordeau:
@carbonFibreCode I'm closing this issue because it wasn't clearly defined beforehand, please make sure to only work on defined issues and wait for an approbation to avoid working with the risk of having your PR closed.

PR #18481, opened 2026-03-07
feat(helm): extraEnv with valueFrom, server.env pass-through, DISABLE_DB_MIGRATIONS, redis auth tests

dnplkndll:
## Relationship to PR #18157

This PR builds directly on the work in PR #18157 ([@HuppertzL](https://github.com/HuppertzL/twenty/tree/improve-helm-chart) — _Improve helm chart // Fix linting issues & introduce Redis externalSecret for redis password // Add additional ENVs // Improve migrations_), which is currently open and awaiting review.

All commits from that PR have been cherry-picked here verbatim. **If PR #18157 merges first, this PR can be cleanly rebased onto main** — the cherry-picked commits will drop out and only our additions will remain.

We're opening this separately so the additional improvements can be reviewed in parallel rather than blocking on #18157.

---

### Bug fix: Redis volumes indentation
`volumes:` in `deployment-redis-internal.yaml` was indented inside `containers:` instead of the pod spec — a structural error that would cause mount failures on any cluster.

### `server.env` pass-through
Arbitrary keys added to `server.env` are now injected into the server pod automatically, filtered against a reserved list of keys already rendered explicitly by the template (`SERVER_URL`, `PG_DATABASE_URL`, `REDIS_URL`, `SIGN_IN_PREFILLED`, `STORAGE_TYPE`, `ACCESS_TOKEN_EXPIRES_IN`, `LOGIN_TOKEN_EXPIRES_IN`, `APP_SECRET`). This is the most ergonomic way to add simple string env vars:

```yaml
server:
  env:
    EMAIL_DRIVER: smtp
    EMAIL_SMTP_HOST: mail.example.com
    AUTH_GOOGLE_ENABLED: "true"
    AUTH_GOOGLE_CLIENT_ID: my-client-id
```

### `extraEnv` array with `valueFrom` support (server + worker)
For env vars that must reference Kubernetes secrets or ConfigMaps, both server and worker now support an `extraEnv` array following the standard Kubernetes `EnvVar` spec:

```yaml
server:
  extraEnv:
    - name: SMTP_PASSWORD
      valueFrom:
        secretKeyRef:
          name: smtp-creds
          key: password
worker:
  extraEnv:
    - name: CUSTOM_SECRET
      valueFrom:
        secretKeyRef:
          name: my-secret
          key: custom
```

This supersedes `additionalEnv` (from PR #18157) for secret references. Both are supported for backward compatibility.

### `server.disableDbMigrations` flag
Boolean flag (default `false`) that injects `DISABLE_DB_MIGRATIONS=true` into the `run-migrations` init container, short-circuiting the migration step. Useful for read-only replicas or environments where migrations are run separately.

```yaml
server:
  disableDbMigrations: true
```

### JSON schema validation
All new fields (`extraEnv`, `disableDbMigrations`) are covered by `values.schema.json` with correct types and `oneOf: [value, valueFrom]` constraint on `extraEnv` items.

### Helm unit tests (63 tests across 11 suites)
- **`tests/env_mapping_test.yaml`** (new): covers `server.env` pass-through for Google auth vars, `server.extraEnv` with plain value, `worker.extraEnv` with `valueFrom`, `server.extraEnv` with `valueFrom` on server pod, `disableDbMigrations` injects `"true"` into init container, `disableDbMigrations` defaults to `"false"`, email SMTP vars injected via pass-through
- **`tests/redis_auth_test.yaml`** (new): covers `REDIS_PASSWORD` injection from external secret (server + worker), plaintext password injection (server + worker), and absence of `REDIS_PASSWORD` when using internal Redis — validates the redis external secret feature from PR #18157
- **`tests/schema_permissions_test.yaml`** (updated): migration assertion updated to match the new `yarn database:migrate:prod` command from PR #18157's improved init container

## Test results

```
Charts:      1 passed, 1 total
Test Suites: 11 passed, 11 total
Tests:       63 passed, 63 total
Time:        314ms
```


dnplkndll:
## Review findings addressed (66fcf46)

Thanks to greptile, copilot, and cubic for the thorough reviews. All actionable findings have been fixed:

### P0 — Migration regression (greptile + cubic)

> `yarn database:migrate:prod` was inside `if [ "$has_schema" = "f" ]`, so on every `helm upgrade` of an existing deployment the `core` schema already exists, the block is skipped, and **pending migrations are silently never applied**.

Fixed by moving `yarn database:migrate:prod` outside the conditional. Only the initial `tsx ./scripts/setup-db.ts` setup script remains gated on the schema-absence check (correct — it should only run on a fresh database).

```sh
# Before (broken on upgrade):
if [ "$has_schema" = "f" ]; then
    tsx ./scripts/setup-db.ts
    yarn database:migrate:prod   # ← skipped on all upgrades
fi

# After (correct):
if [ "$has_schema" = "f" ]; then
    tsx ./scripts/setup-db.ts    # only on fresh DB
fi
yarn database:migrate:prod       # always runs
```

### P2 — Cosmetic / test quality

- **Duplicate `tmp/` in `.gitignore`** — removed the second copy (artifact of stash conflict resolution)
- **`run-migrations` `$img` indentation** — fixed 16-space indent to 8
- **`storageEnv` whitespace** — changed `              {{ nindent }}` to `            {{- nindent }}` to avoid orphaned whitespace line in rendered manifests
- **Brittle `env[0]` index assertions** in `server_url_test.yaml` and `env_mapping_test.yaml` — replaced with JSONPath name filters (`env[?(@.name=="SERVER_URL")]`) so tests don't break if env ordering changes
- **Missing worker negative Redis test** — added assertion that `REDIS_PASSWORD` is NOT injected into the worker when using internal Redis (server had this test, worker did not)

### Not addressed (pre-existing or out of scope)

- External DB secret password substitution via `$(DB_PASSWORD)` shell expansion — this was pre-existing behavior from the original chart, not introduced by this PR
- `type: Opaque` missing from `secret-tokens.yaml` — pre-existing, out of scope
- `podSecurityContext`/`containerSecurityContext` schema stubs without implementation — pre-existing

**64/64 helm unittest pass.**

dnplkndll:
## Second round of inline comment fixes (aaea0de)

**`additionalEnv` in `ensure-database-exists` init container** (copilot, cubic) — removed. That container only needs `PGPASSWORD` and `APP_PASSWORD` to run `psql` commands. Injecting `server.additionalEnv` there was a copy-paste artifact from HuppertzL's original commit; it has no effect on a postgres container and allows user-supplied keys to shadow the secret-backed credentials (e.g. `PGPASSWORD` override breaking DB init). `additionalEnv` remains in the `run-migrations` init container and the server container where it's intentional.

**`${PG_DATABASE_URL}` unquoted** (copilot) — quoted as `"${PG_DATABASE_URL}"` in the `has_schema` psql invocation.

**Missing trailing newline** (copilot) — added to `deployment-server.yaml`.

**Test description** (copilot) — "yarn database migrate" → "yarn database:migrate:prod".

**`values.yaml` comment** (copilot) — "Set to `\"true\"`" → "Set to true" (it's a boolean, not a string).

**`extraEnv` schema permissiveness** (cubic) — the existing `oneOf: [required: value, required: valueFrom]` constraint already enforces exactly one of the two; `oneOf` rejects inputs where both are present. No change needed.

64/64 tests pass.

HuppertzL:
Hey @dnplkndll,
thanks for incorporating my suggestions into your PR.
Could you also consider including these changes?
https://github.com/twentyhq/twenty/pull/18157/changes/388be202c68f0e68c2f5549d8d5e3f156cbf8c09

I noticed that the PVC for the internal Redis instance was missing.

dnplkndll:
> Hey @dnplkndll, thanks for incorporating my suggestions into your PR. Could you also consider including these changes? [388be20](https://github.com/twentyhq/twenty/commit/388be202c68f0e68c2f5549d8d5e3f156cbf8c09)
> 
> I noticed that the PVC for the internal Redis instance was missing.

thanks. yes I did not test with that as I was just using ephemeral redis. seems like there are cases on a larger production you would not want to have to rebuild that cache. 


on the migrations. I notices the DISABLE_DB_MIGRATIONS and was thinking about how would the init containers interact with replica count > 1?
```
This is a problem. With replicaCount > 1, every replica runs its own init containers,
  so you'd get multiple pods racing to run migrations concurrently. There's no locking.

  Here's the current situation:

  What happens with replicas > 1:
  - All N pods start simultaneously
  - All N run-migrations init containers run in parallel
  - Multiple yarn database:migrate:prod commands hit the DB at the same time
  - TypeORM migrations aren't idempotent by design — concurrent runs can cause duplicate
  key errors, partial migrations, or deadlocks

  What we have:
  - server.replicaCount and worker.replicaCount — plain replica counts, no HPA
  - No migration locking mechanism (no Kubernetes Job, no advisory lock, no leader
  election)

  Options to fix:

  1. Kubernetes Job (cleanest) — run migrations as a separate Job with helm.sh/hook:
  pre-upgrade,pre-install instead of an init container. Only one pod runs, completes
  before the Deployment rolls out.
  2. Advisory lock in the script — wrap the migration in a Postgres advisory lock:
  psql "${PG_DATABASE_URL}" -c "SELECT pg_advisory_lock(12345)"
  yarn database:migrate:prod
  psql "${PG_DATABASE_URL}" -c "SELECT pg_advisory_unlock(12345)"
  2. All replicas still run the init container, but only one migrates at a time. Others
  wait then no-op (migrations already applied).
  3. Keep it simple — document that replicaCount: 1 is required during upgrades, scale up
   after.

  Option 1 (Helm hook Job) is the most robust.
  ```
  I hate making these overly complicated. But it does seem like a case for a job pre-install/upgrade

HuppertzL:
Yeah, 
i stepped over this migration part, too. But kept it as it is.
In my personal projects i use Helm hooks since it can easily executed at the beginning, end etc of an helm upgrade is not tied to a pod like initContainers are.
I prefer doing common stuff in Helm hooks and run only things in initContainers that setup the pod itself (eg app cache warmups).

charlesBochet:
Hey @dnplkndll, thanks for all the work on this.

We've merged #18157 with a simplified scope (Redis auth, Redis PVC, extraEnv array, lint fixes — no migration init container rewrite, no additionalEnv, no server.env pass-through). See the comment there for the full rationale.

Since #18157 is now on main, the cherry-picked commits from this PR are redundant. We merged main into a local copy of your branch (`helm-improvements` on twentyhq/twenty) to see what remains — here's what's left vs main:

**Remaining changes:**
- Worker `storageEnv` whitespace fix (`{{-` trim) — minor, good to have
- `values.schema.json` additions for `extraEnv` and `disableDbMigrations` — the `extraEnv` schema is useful, but `disableDbMigrations` no longer applies since we removed that feature
- Spilo tag bump (`3.3-p2` → `v0.43.5`) — is this intentional?
- Test files (`env_mapping_test.yaml`, `redis_auth_test.yaml`) — new test coverage
- Test updates (`schema_permissions_test.yaml`, `server_url_test.yaml`) — JSONPath fixes and migration assertion update

**About the tests:** Several of them test features that no longer exist on main (`disableDbMigrations`, `server.env` pass-through, `run-migrations` init container). They would need to be updated or removed. The Redis auth tests and extraEnv tests are valid and useful.

Could you rebase onto current main, drop the stale tests, and we can merge the remaining useful bits?

charlesBochet:
Opened https://github.com/twentyhq/twenty/pull/18836 with your changes (revised)
Thanks a lot @dnplkndll

dnplkndll:
thanks!

PR #18483, opened 2026-03-07
fix: handle info log level alias and filter invalid levels gracefully

### LINKED ISSUE #18356: Error when setting LOG_LEVELS to query,schema,error,warn,info,log,migration
## Bug Description
When we set the env LOG_LEVELS=query,schema,error,warn,info,log,migration in **version v1.17**
we are getting the following error:

```
/app/node_modules/@nestjs/common/services/utils/is-log-level-enabled.util.js:25
        .map(level => LOG_LEVEL_VALUES[level])
         ^
TypeError: logLevels.map is not a function
    at isLogLevelEnabled (/app/node_modules/@nestjs/common/services/utils/is-log-level-enabled.util.js:25:10)
    at ConsoleLogger.isLevelEnabled (/app/node_modules/@nestjs/common/services/console-logger.service.js:133:66)
    at ConsoleLogger.debug (/app/node_modules/@nestjs/common/services/console-logger.service.js:79:19)
    at LoggerService.debug (/app/packages/twenty-server/dist/engine/core-modules/logger/logger.service.js:54:28)
    at Logger.debug (/app/node_modules/@nestjs/common/services/logger.service.js:69:36)
    at /app/node_modules/@nestjs/common/services/logger.service.js:106:45
    at Array.forEach (<anonymous>)
    at Logger.flush (/app/node_modules/@nestjs/common/services/logger.service.js:106:24)
    at NestApplicationContext.flushLogs (/app/node_modules/@nestjs/core/nest-application-context.js:150:25)
    at NestApplicationContext.useLogger (/app/node_modules/@nestjs/core/nest-application-context.js:142:18)

```
## Expected behavior
The logging env variable should be processed correctly and all debug outputs should show up.

## Technical inputs
See above

Felipeness:
## Summary

Fixes #18356

Setting `LOG_LEVELS=debug,info,error,warn` crashes the application with `TypeError: logLevels.map is not a function` because `info` is not a valid NestJS log level (NestJS uses `log` instead of `info`).

The `CastToLogLevelArray` decorator previously rejected the **entire** value when any single level was invalid, returning `undefined`. This `undefined` propagated through `ConfigService` → `loggerModuleFactory` → `ConsoleLogger.setLogLevels()`, where NestJS called `.map()` on it and crashed.

## Changes

- **Alias mapping**: common aliases like `info` are mapped to their NestJS equivalents (`log`) before validation
- **Graceful filtering**: invalid levels are filtered out instead of rejecting the entire value — `debug,info,error,toto` now resolves to `['debug', 'log', 'error']` instead of `undefined`
- **Deduplication**: prevents duplicates when an alias maps to an already-present level (e.g. `log,info` → `['log']`)
- `undefined` is only returned when **no** valid levels remain

## Test plan

- [x] Added test: `"info"` → `["log"]` (alias mapping)
- [x] Added test: `"debug,info,error,warn"` → `["debug", "log", "error", "warn"]` (exact scenario from #18356)
- [x] Added test: `"log,info"` → `["log"]` (deduplication)
- [x] Updated test: `"verbose,error,toto"` → `["verbose", "error"]` (filter instead of reject)
- [x] Existing test: `"toto"` → `undefined` (all invalid = no override)
- [x] All existing happy-path tests unchanged and passing

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/twenty-config/decorators/__tests__/cast-to-log-level-array.decorator.spec.ts
+++ b/packages/twenty-server/src/engine/core-modules/twenty-config/decorators/__tests__/cast-to-log-level-array.decorator.spec.ts
@@ -56,10 +56,43 @@ describe('CastToLogLevelArray Decorator', () => {
     expect(transformedClass.logLevels).toBeUndefined();
   });
 
-  it('should cast "verbose,error,toto" to undefined', () => {
+  it('should cast "verbose,error,toto" to ["verbose", "error"] filtering invalid levels', () => {
     const transformedClass = plainToClass(TestClass, {
       logLevels: 'verbose,error,toto',
     });
+
+    expect(transformedClass.logLevels).toStrictEqual(['verbose', 'error']);
+  });
+
+  it('should cast "info" to ["log"] mapping info alias', () => {
+    const transformedClass = plainToClass(TestClass, { logLevels: 'info' });
+
+    expect(transformedClass.logLevels).toStrictEqual(['log']);
+  });
+
+  it('should cast "debug,info,error,warn" to ["debug", "log", "error", "warn"]', () => {
+    const transformedClass = plainToClass(TestClass, {
+      logLevels: 'debug,info,error,warn',
+    });
+
+    expect(transformedClass.logLevels).toStrictEqual([
+      'debug',
+      'log',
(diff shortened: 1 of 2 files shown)


charlesBochet:
Commented on the original issue, closing this one until what we want to do is clearer

PR #18487, opened 2026-03-08
fix: use shared isValidHostname for blocklist domain validation

WocaiStart:
## Summary
Fixes #18482

- Adding any email or domain to the blocklist via Settings → Emails → Blocklist returned GraphQL 400
- **Root cause**: Server used `isDomain()` from `engine/utils/is-domain.ts` which rejects valid multi-level subdomains (e.g., `sub.domain.example.com`), while the frontend uses `isValidHostname()` from `twenty-shared` which accepts them — validation mismatch
- **Fix**: Replace `isDomain` with `isValidHostname` from `twenty-shared` using `{ allowIp: false, allowLocalhost: false }`, matching the frontend

## Changes
- `blocklist-validation.service.ts`: Replaced `isDomain` import/call with `isValidHostname` from `twenty-shared/utils`
- `blocklist-validation.service.spec.ts`: New test file with 10 test cases

## Proof of mismatch
```
isDomain("sub.domain.example.com")      = false  ← server rejects
isValidHostname("sub.domain.example.com") = true   ← frontend accepts
```

## Test plan
- [x] Valid email `eddy@gmail.com` → passes
- [x] Valid domain `@apple.com` → passes
- [x] Multi-level subdomain `@sub.domain.example.com` → passes (was rejected)
- [x] IP `@192.168.1.1` → rejected (security preserved)
- [x] Localhost `@localhost` → rejected (security preserved)
- [x] 11/11 tests pass

### DIFF
--- a/packages/twenty-server/src/modules/blocklist/blocklist-validation-manager/services/__tests__/blocklist-validation.service.spec.ts
+++ b/packages/twenty-server/src/modules/blocklist/blocklist-validation-manager/services/__tests__/blocklist-validation.service.spec.ts
@@ -1,0 +1,130 @@
+import {
+  CommonQueryRunnerException,
+  CommonQueryRunnerExceptionCode,
+} from 'src/engine/api/common/common-query-runners/errors/common-query-runner.exception';
+
+import { BlocklistValidationService } from '../blocklist-validation.service';
+
+const buildBlocklistItem = (handle: string | null) =>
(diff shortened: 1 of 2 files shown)


charlesBochet:
I'm not reading it (I see "any" in the code, I'm stopping there)

See: https://github.com/twentyhq/twenty/pull/18488

PR #18499, opened 2026-03-09
Fix wrong type usage

martmull:
fix wrong type usage + add tests

### DIFF
--- a/packages/create-twenty-app/package.json
+++ b/packages/create-twenty-app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "create-twenty-app",
-  "version": "0.6.4",
+  "version": "0.7.0-canary.0",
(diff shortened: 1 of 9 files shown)


packages/create-twenty-app/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "create-twenty-app",
-  "version": "0.6.4",
```
Weiko, line 3: Why do we need to upgrade this version?

Weiko (approved):
LGTM

packages/create-twenty-app/package.json
martmull: because create-twenty-app and twenty-sdk must share the same version

PR #18510, opened 2026-03-09
`[SDK]` Client fallback stub

prastoin:


prastoin:
Not the good direction

PR #18523, opened 2026-03-10
Run vulnerable operation in isolated environment

martmull:
When driver = LAMBDA:
- run esbuild ts transpilation on dedicated lambda
- run yarn install on app dependencies on a dedicated lambda


packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/constants/builder/index.mjs
```diff
@@ -0,0 +1,102 @@
...
+
+  const { NODE_OPTIONS: _nodeOptions, ...cleanEnv } = process.env;
+
+  // Lambda runs as a sandboxed user whose $HOME doesn't exist.
+  // Yarn needs a writable HOME for its global cache/config.
+  cleanEnv.HOME = '/tmp';
+
+  await execFilePromise(
+    process.execPath,
+    [YARN_ENGINE_PATH, 'workspaces', 'focus', '--all', '--production'],
```
prastoin: ***Question:*** Do you use yarn workspaces in a lambda layer ?

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/constants/common-layer-dependencies/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "dependencies": {
+    "archiver": "^7.0.1",
+    "esbuild": "^0.25.0"
```
prastoin, line 4: ***Question:*** Why not having two distinct package.json one for build and one for the node_modules zip ?

packages/twenty-server/src/engine/metadata-modules/logic-function/services/logic-function-from-source.service.ts
```diff
@@ -302,14 +304,33 @@ export class LogicFunctionFromSourceService {
...
+    });
+
+    await this.logicFunctionResourceService.uploadBuiltFile({
+      workspaceId,
+      applicationUniversalIdentifier: ownerFlatApplication.universalIdentifier,
+      builtHandlerPath: flatLogicFunction.builtHandlerPath,
+      builtCode,
+    });
+
+    const checksum = crypto.createHash('md5').update(builtCode).digest('hex');
```
prastoin, line 333: Checksum could be returned by the lambda directly ?
The zip archive might already have one

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/lambda.driver.ts
```diff
@@ -138,67 +186,416 @@ export class LambdaDriver implements LogicFunctionDriver {
...
+        );
+      }
+
+      return result.LayerVersionArn;
+    } finally {
+      await temporaryDirManager.clean();
+    }
+  }
+
+  private async ensureBuilderLambdaExists(): Promise<void> {
```
prastoin: ***Suggestion:*** What would you think of creating a dedicated service for each layer ?
As I'm also about to add one new

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/constants/transpiler/index.mjs
```diff
@@ -0,0 +1,65 @@
...
+
+  if (!resolved.startsWith(dir + '/')) {
+    throw new Error(`Path traversal detected: ${filePath}`);
+  }
+
+  return resolved;
+};
+
+export const handler = async (event) => {
+  const { action, sourceCode, sourceFileName, builtFileName } = event;
```
prastoin: ***Polish:*** Not against the raw `mjs` files here, might wanna still add on top of the file a `// @ts-check` with some `JSDoc`

prastoin (commented):
Hey there nice job ! That's really nice
Main points are that I feel like we've swap the naming of both layers
The current `builder` does a yarn install and zip the node_modules ( has a dep on archiver )
The current transpiler does a project `build` using esbuild

We might also quite extract the layer specific logic in dedicated services out of the lambda driver !

Please let me know
Also highly related to https://github.com/twentyhq/twenty/pull/18544

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/constants/builder/index.mjs
martmull: this is the command to install production dependencies in yarn 4

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/constants/common-layer-dependencies/package.json
martmull: i though it was much simple to get only one layer for all core lambda like builder, transpiler or executor

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/lambda.driver.ts
martmull: I tough about it, maybe we will do that because the drivers will be too big. But we cannot really create an injectable service nestJs speaking as drivers are not injectable. I would keep it this way for now, and eventually split later

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/utils/copy-builder.ts
```diff
@@ -0,0 +1,17 @@
+import { promises as fs } from 'fs';
+import { resolve, join } from 'path';
+
+import { ASSET_PATH } from 'src/constants/assets-path';
+
+const BUILDER_FILE_PATH = resolve(
```
prastoin, line 6: We might wanna use nest js assets instead

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/lambda.driver.ts
prastoin: What would you think of having dedicated drivers then ?

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/constants/common-layer-dependencies/package.json
prastoin: I don't think it's really worth it in the end, both technically and functionally as it will install unused deps

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/constants/builder/index.mjs
prastoin: ok !

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/constants/yarn-install/index.mjs
```diff
@@ -99,9 +119,9 @@ export const handler = async (event) => {
     await runYarnInstall(nodejsDir);
     await createZip(buildDir, zipPath);
 
-    const zipBase64 = (await fs.readFile(zipPath)).toString('base64');
+    await uploadToPresignedUrl(zipPath, presignedUploadUrl);
 
-    return { zipBase64 };
+    return { success: true };
```
prastoin, line 124: ***Polish:*** return the s3 id/name

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/constants/yarn-install/index.mjs
```diff
@@ -77,8 +77,24 @@ const createZip = async (buildDir, zipPath) => {
...
+  if (!response.ok) {
+    throw new Error(
+      `Failed to upload zip to S3: ${response.status} ${response.statusText}`,
+    );
+  }
+};
+
 export const handler = async (event) => {
-  const { action, packageJson, yarnLock } = event;
+  const { action, packageJson, yarnLock, presignedUploadUrl } = event;
```
prastoin, line 97: ***Concern:*** What would be the impact if somehow the token leaks from the deps install
Anyone could upload a file on the s3 🤔 

The initial motivation was to avoid executing any risky operations on the server, though we still provide something to steal here

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/constants/yarn-install/index.mjs
```diff
@@ -0,0 +1,133 @@
...
+  const { NODE_OPTIONS: _nodeOptions, ...cleanEnv } = process.env;
+
+  // Lambda runs as a sandboxed user whose $HOME doesn't exist.
+  // Yarn needs a writable HOME for its global cache/config.
+  cleanEnv.HOME = '/tmp';
+
+  try {
+    await execFilePromise(
+      process.execPath,
+      [YARN_ENGINE_PATH, 'workspaces', 'focus', '--all', '--production'],
```
prastoin, line 38: ***remark:*** we might wanna explicitly disable postinstall and preinstall here ? also explicitly enable the hardened mode ?

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/constants/yarn-install/index.mjs
martmull: yes, indeed, i think it is ok tbh it is just an upload

prastoin (approved):
discussed on discord

PR #18529, opened 2026-03-10
Fix add more tab bottom separator

thomtrp:
Adding a full height prop

Before 
Capture d’écran 2026-03-10 à 14 06 13

After
Capture d’écran 2026-03-10 à 14 06 33


packages/twenty-front/src/modules/ui/layout/dropdown/components/Dropdown.tsx
```diff
@@ -39,14 +39,17 @@ const StyledDropdownFallbackAnchor = styled.div`
 
 const StyledClickableComponent = styled.div<{
   width?: Width;
+  fullHeight?: boolean;
 }>`
-  height: fit-content;
+  display: flex;
+  height: ${({ fullHeight }) => (fullHeight ? '100%' : 'fit-content')};
```
charlesBochet: rather add height on clickableComponent in the caller

charlesBochet (changes requested):

charlesBochet (approved):

PR #18533, opened 2026-03-10
Fix PDF Upload edge case

neo773:
we were using an older version of `file-type` which has limited support for PDF as it's a complex spec
Updated to latest version which includes support for plugins and added `@file-type/pdf` which has extensive spec compliant detection approach

fixes TWENTY-SERVER-FAN


FelixMalfait:

Your preview environment is available at: http://bore.pub:64615

charlesBochet:
@neo773 I'm unsure this is related to this PR but fileCategory seems to be completely broken
image


Could you investigate?

charlesBochet (dismissed):
see comment

neo773:
@charlesBochet 

seems to be deprecated

image

FelixMalfait (approved):

PR #18535, opened 2026-03-10
Fix permission settings page

thomtrp:
Capture d’écran 2026-03-10 à 15 59 37


charlesBochet (approved):

packages/twenty-front/src/modules/settings/roles/role-permissions/object-level-permissions/object-form/components/OverridableCheckbox.tsx
```diff
@@ -11,15 +11,15 @@ const StyledOverridableCheckboxContainer = styled.div`
   align-items: center;
   display: inline-flex;
   justify-content: flex-start;
-  width: calc(${themeCssVariables.icon.size.xl} * 2);
+  width: calc(${themeCssVariables.icon.size.xl} * 2 * 1px);
```
Weiko, line 14: To double check if we can have something cleaner

Weiko (approved):
LGTM

PR #18536, opened 2026-03-10
Add system view fallback

Weiko:
## Context
The goal is to add a "See records" button in all objects that would redirect to that view (this will be done in a later PR). See screenshot below.
Screenshot 2026-03-10 at 15 54 36

## Implementation
- If a view does not exist on an object, there is a **temporary** fallback where the frontend creates the missing view as a custom view when going over the object index page
- System objects are now surfaced but we don't want their records to be editable, they will be readonly (mostly, all fields will be non-editable except for their custom fields).
- We can't create a new record of a system object, some actions are also hidden.
- The backend now rejects if you are trying to delete the last view of an object

### DIFF
--- a/packages/twenty-front/src/modules/command-menu-item/record/constants/DefaultRecordCommandMenuItemsConfig.tsx
+++ b/packages/twenty-front/src/modules/command-menu-item/record/constants/DefaultRecordCommandMenuItemsConfig.tsx
@@ -116,8 +116,13 @@ export const DEFAULT_RECORD_COMMAND_MENU_ITEMS_CONFIG: Record<
     position: 2,
     isPinned: true,
     Icon: IconPlus,
-    shouldBeRegistered: ({ objectPermissions, hasAnySoftDeleteFilterOnView }) =>
-      (objectPermissions.canUpdateObjectRecords &&
+    shouldBeRegistered: ({
+      objectMetadataItem,
+      objectPermissions,
+      hasAnySoftDeleteFilterOnView,
+    }) =>
+      (!objectMetadataItem?.isSystem &&
+        objectPermissions.canUpdateObjectRecords &&
         !hasAnySoftDeleteFilterOnView) ??
       false,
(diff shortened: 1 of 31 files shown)


FelixMalfait:

Your preview environment is available at: http://bore.pub:12446

packages/twenty-front/src/modules/command-menu-item/record/constants/DefaultRecordCommandMenuItemsConfig.tsx
```diff
@@ -116,8 +116,13 @@ export const DEFAULT_RECORD_COMMAND_MENU_ITEMS_CONFIG: Record<
     position: 2,
     isPinned: true,
     Icon: IconPlus,
-    shouldBeRegistered: ({ objectPermissions, hasAnySoftDeleteFilterOnView }) =>
-      (objectPermissions.canUpdateObjectRecords &&
+    shouldBeRegistered: ({
+      objectMetadataItem,
```
charlesBochet, line 120: @bosiraphael FYI

packages/twenty-front/src/modules/object-record/record-index/components/RecordIndexLoadBaseOnContextStoreEffect.tsx
```diff
@@ -35,13 +41,16 @@ export const RecordIndexLoadBaseOnContextStoreEffect = () => {
     if (isDefined(view)) {
       loadRecordIndexStates(view, objectMetadataItem);
       setLoadedViewId(contextStoreCurrentViewId);
+    } else {
+      createDefaultViewForObject(objectMetadataItem);
```
charlesBochet, line 45: this does not seem the place to do this, this is an effect, we will likely have race conditions here. We can keep it if that's working for now (overall, the frontend creating views is not a long term vision as discussed together)

charlesBochet (approved):

PR #18538, opened 2026-03-10
Fix form field select + form field number

thomtrp:
Before
Capture d’écran 2026-03-10 à 16 23 19

After
Capture d’écran 2026-03-10 à 16 32 56

Addition:
- hover on forget password
- settings field width


packages/twenty-front/src/modules/object-record/record-field/ui/form-types/components/FormSingleRecordPicker.tsx
```diff
@@ -191,6 +196,7 @@ export const FormSingleRecordPicker = ({
             clickableComponent={
               <StyledFormSelectContainerWrapper>
                 <FormFieldInputInnerContainer
+                  className={`${formSelectInnerContainerClassName} ${formSelectInnerContainerHoverClassName}`}
```
charlesBochet: no, we should not do that :)

charlesBochet (approved):

PR #18544, opened 2026-03-10
Refactor twenty client sdk provisioning for logic function and front-component

prastoin:
## 1. The `twenty-client-sdk` Package (Source of Truth)

The monorepo package at `packages/twenty-client-sdk` ships with:
- A **pre-built metadata client** (static, generated from a fixed schema)
- A **stub core client** that throws at runtime (`CoreApiClient was not generated...`)
- Both ESM (`.mjs`) and CJS (`.cjs`) bundles in `dist/`
- A `package.json` with proper `exports` map for `twenty-client-sdk/core`, `twenty-client-sdk/metadata`, and `twenty-client-sdk/generate`

## 2. Generation & Upload (Server-Side, at Migration Time)

**When**: `WorkspaceMigrationRunnerService.run()` executes after a metadata schema change.

**What happens in `SdkClientGenerationService.generateAndStore()`**:
1. Copies the stub `twenty-client-sdk` package from the server's assets (resolved via `SDK_CLIENT_PACKAGE_DIRNAME` — from `dist/assets/twenty-client-sdk/` in production, or from `node_modules` in dev)
2. Filters out `node_modules/` and `src/` during copy — only `package.json` + `dist/` are kept (like an npm publish)
3. Calls `replaceCoreClient()` which uses `@genql/cli` to introspect the **application-scoped** GraphQL schema and generates a real `CoreApiClient`, then compiles it to ESM+CJS and overwrites `dist/core.mjs` and `dist/core.cjs`
4. Archives the **entire package** (with `package.json` + `dist/`) into `twenty-client-sdk.zip`
5. Uploads the single archive to S3 under `FileFolder.GeneratedSdkClient`
6. Sets `isSdkLayerStale = true` on the `ApplicationEntity` in the database

## 3. Invalidation Signal

The `isSdkLayerStale` boolean column on `ApplicationEntity` is the invalidation mechanism:
- **Set to `true`** by `generateAndStore()` after uploading a new client archive
- **Checked** by both logic function drivers before execution — if `true`, they rebuild their local layer
- **Set back to `false`** by `markSdkLayerFresh()` after the driver has successfully consumed the new archive

Default is `false` so existing applications without a generated client aren't affected.

## 4a. Logic Functions — Local Driver

**`ensureSdkLayer()`** is called before every execution:
1. Checks if the local SDK layer directory exists AND `isSdkLayerStale` is `false` → early return
2. Otherwise, cleans the local layer directory
3. Calls `downloadAndExtractToPackage()` which streams the zip from S3 directly to disk and extracts the full package into `<tmpdir>/sdk/<workspaceId>-<appId>/node_modules/twenty-client-sdk/`
4. Calls `markSdkLayerFresh()` to set `isSdkLayerStale = false`

**At execution time**, `assembleNodeModules()` symlinks everything from the deps layer's `node_modules/` **except** `twenty-client-sdk`, which is symlinked from the SDK layer instead. This ensures the logic function's `import ... from 'twenty-client-sdk/core'` resolves to the generated client.

## 4b. Logic Functions — Lambda Driver

**`ensureSdkLayer()`** is called during `build()`:
1. Checks if `isSdkLayerStale` is `false` and an existing Lambda layer ARN exists → early return
2. Otherwise, deletes all existing layer versions for this SDK layer name
3. Calls `downloadArchiveBuffer()` to get the raw zip from S3 (no disk extraction)
4. Calls `reprefixZipEntries()` which streams the zip entries into a **new zip** with the path prefix `nodejs/node_modules/twenty-client-sdk/` — this is the Lambda layer convention path. All done in memory, no disk round-trip
5. Publishes the re-prefixed zip as a new Lambda layer via `publishLayer()`
6. Calls `markSdkLayerFresh()`

**At function creation**, the Lambda is created with **two layers**: `[depsLayerArn, sdkLayerArn]`. The SDK layer is listed last so it overwrites the stub `twenty-client-sdk` from the deps layer (later layers take precedence in Lambda's `/opt` merge).

## 5. Front Components

Front components are built by `app:build` with `twenty-client-sdk/core` and `twenty-client-sdk/metadata` as **esbuild externals**. The stored `.mjs` in S3 has unresolved bare import specifiers like `import { CoreApiClient } from 'twenty-client-sdk/core'`.

SDK import resolution is split between the **frontend host** (fetching & caching SDK modules) and the **Web Worker** (rewriting imports):

**Server endpoints**:
- `GET /rest/front-components/:id` — `FrontComponentService.getBuiltComponentStream()` returns the **raw `.mjs`** directly from file storage. No bundling, no SDK injection.
- `GET /rest/sdk-client/:applicationId/:moduleName` — `SdkClientController` reads a single file (e.g. `dist/core.mjs`) from the generated SDK archive via `SdkClientGenerationService.readFileFromArchive()` and serves it as JavaScript.

**Frontend host** (`FrontComponentRenderer` in `twenty-front`):
1. Queries `FindOneFrontComponent` which returns `applicationId`, `builtComponentChecksum`, `usesSdkClient`, and `applicationTokenPair`
2. If `usesSdkClient` is `true`, renders `FrontComponentRendererWithSdkClient` which calls the `useApplicationSdkClient` hook
3. `useApplicationSdkClient({ applicationId, accessToken })` checks the Jotai atom family cache for existing blob URLs. On cache miss, fetches both SDK modules from `GET /rest/sdk-client/:applicationId/core` and `/metadata`, creates **blob URLs** for each, and stores them in the atom family
4. Once the blob URLs are cached, passes them as `sdkClientUrls` (already blob URLs, not server URLs) to `SharedFrontComponentRenderer` → `FrontComponentWorkerEffect` → worker's `render()` call via `HostToWorkerRenderContext`

**Worker** (`remote-worker.ts` in `twenty-sdk`):
1. Fetches the raw component `.mjs` source as text
2. If `sdkClientUrls` are provided and the source contains SDK import specifiers (`twenty-client-sdk/core`, `twenty-client-sdk/metadata`), **rewrites** the bare specifiers to the blob URLs received from the host (e.g. `'twenty-client-sdk/core'` → `'blob:...'`)
3. Creates a blob URL for the rewritten source and `import()`s it
4. Revokes only the component blob URL after the module is loaded — the SDK blob URLs are owned and managed by the host's Jotai cache

This approach eliminates server-side esbuild bundling on every request, caches SDK modules per application in the frontend, and keeps the worker's job to a simple string rewrite.

## Summary Diagram

```
app:build (SDK)
  └─ twenty-client-sdk stub (metadata=real, core=stub)
       │
       ▼
WorkspaceMigrationRunnerService.run()
  └─ SdkClientGenerationService.generateAndStore()
       ├─ Copy stub package (package.json + dist/)
       ├─ replaceCoreClient() → regenerate core.mjs/core.cjs
       ├─ Zip entire package → upload to S3
       └─ Set isSdkLayerStale = true
              │
     ┌────────┴────────────────────┐
     ▼                             ▼
Logic Functions               Front Components
     │                             │
     ├─ Local Driver               ├─ GET /rest/sdk-client/:appId/core
     │   └─ downloadAndExtract     │    → core.mjs from archive
     │      → symlink into         │
     │        node_modules         ├─ Host (useApplicationSdkClient)
     │                             │    ├─ Fetch SDK modules
     └─ Lambda Driver              │    ├─ Create blob URLs
         └─ downloadArchiveBuffer  │    └─ Cache in Jotai atom family
            → reprefixZipEntries   │
            → publish as Lambda    ├─ GET /rest/front-components/:id
              layer                │    → raw .mjs (no bundling)
                                   │
                                   └─ Worker (browser)
                                        ├─ Fetch component .mjs
                                        ├─ Rewrite imports → blob URLs
                                        └─ import() rewritten source
```

## Next PR
- Estimate perf improvement by implementing a redis caching for front component client storage ( we don't even cache front comp initially )
- Implem frontent blob invalidation sse event from server


packages/twenty-front/src/modules/front-components/utils/fetchSdkClientBlobUrls.ts
```diff
@@ -0,0 +1,36 @@
+import { type SdkClientBlobUrls } from '@/front-components/states/sdkClientBlobUrlsFamilyState';
+import { getSdkClientUrls } from '@/front-components/utils/getSdkClientUrls';
+
+const fetchAndCreateBlobUrl = async (
+  url: string,
+  token: string,
+): Promise<string> => {
+  const response = await fetch(url, {
```
prastoin, line 8: might wanna use axios here

prastoin:
## From there:
### Logic function
- local driver ✅ 
- lambda driver ❓ 

### Front component with client
- 💔 

### Apps devxp
- ✅

prastoin:
### Front component with client
- ✅

prastoin:
## Lambda driver ✅

charlesBochet:
Closing temporarily as this is staled, let's re-open tomorrow!

prastoin:
As we've finally decided to remove yoga this PR could be completely challenged

.github/actions/spawn-twenty-docker-image/action.yaml
```diff
@@ -30,20 +31,27 @@ outputs:
...
   steps:
-    - name: Validate version
+    - name: Resolve version
+      id: resolve
       shell: bash
       run: |
         VERSION="${{ inputs.twenty-version }}"
-        if ! echo "$VERSION" | grep -qE '^v[0-9]+\.[0-9]+\.[0-9]+$'; then
-          echo "::error::twenty-version must be a semver tag (e.g., v0.40.0). Got: '$VERSION'"
+        if [ "$VERSION" = "latest" ]; then
```
charlesBochet, line 39: for later: we can simplify this by using the joint image directly

packages/create-twenty-app/README.md
```diff
@@ -127,7 +128,7 @@ The server is pre-seeded with a workspace and user (`tim@apple.dev` / `tim@apple
 - Use `yarn twenty remote add --local` to authenticate with your Twenty workspace via OAuth.
 - Explore the generated project and add your first entity with `yarn twenty add` (logic functions, front components, objects, roles, views, navigation menu items, skills).
 - Use `yarn twenty dev` while you iterate — it watches, builds, and syncs changes to your workspace in real time.
-- `CoreApiClient` (for workspace data via `/graphql`) is auto-generated by `yarn twenty dev`. `MetadataApiClient` (for workspace configuration and file uploads via `/metadata`) ships pre-built with the SDK. Both are available via `import { CoreApiClient, MetadataApiClient } from 'twenty-sdk/clients'`.
+- `CoreApiClient` (for workspace data via `/graphql`) is auto-generated by `yarn twenty dev`. `MetadataApiClient` (for workspace configuration and file uploads via `/metadata`) ships pre-built with the SDK. Both are available via `import { CoreApiClient } from 'twenty-client-sdk/core'` and `import { MetadataApiClient } from 'twenty-client-sdk'`.
```
charlesBochet: twenty-client-sdk/metadata

packages/twenty-apps/hello-world/yarn.lock
```diff
@@ -5750,6 +5751,17 @@ __metadata:
   languageName: node
   linkType: hard
 
+"twenty-client-sdk@portal:../../twenty-client-sdk::locator=hello-world%40workspace%3A.":
```
charlesBochet, line 5754: this won't work at runtime

packages/twenty-sdk/src/front-component-renderer/remote/worker/remote-worker.ts
```diff
@@ -37,6 +37,49 @@ exposeGlobals({
   __HTML_TAG_TO_CUSTOM_ELEMENT_TAG__: HTML_TAG_TO_CUSTOM_ELEMENT_TAG,
 });
 
+const fetchTextOrThrow = async (
```
charlesBochet: don't like the naming, this is too broad

packages/twenty-server/src/database/typeorm/core/migrations/common/1773000000000-add-is-sdk-layer-stale-to-application.ts
```diff
@@ -0,0 +1,19 @@
+import { type MigrationInterface, type QueryRunner } from 'typeorm';
+
+export class AddIsSdkLayerStaleToApplication1773000000000
```
charlesBochet, line 3: AI generated, generate with typeorm

packages/twenty-server/src/database/typeorm/core/migrations/common/1773100000000-add-uses-sdk-client-to-front-component.ts
```diff
@@ -0,0 +1,19 @@
+import { type MigrationInterface, type QueryRunner } from 'typeorm';
+
+export class AddUsesSdkClientToFrontComponent1773100000000
+  implements MigrationInterface
+{
+  name = 'AddUsesSdkClientToFrontComponent1773100000000';
```
charlesBochet, line 6: same

packages/twenty-server/src/engine/core-modules/file/interfaces/file-folder.interface.ts
```diff
@@ -57,6 +57,9 @@ export const fileFolderConfigs: Record<FileFolder, FileFolderConfig> = {
   [FileFolder.AppTarball]: {
     ignoreExpirationToken: false,
   },
+  [FileFolder.GeneratedSdkClient]: {
+    ignoreExpirationToken: false,
```
charlesBochet, line 61: Not sure we need a new file folder (could go to dependencies) but LGTM

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/local.driver.ts
```diff
@@ -110,6 +172,59 @@ export class LocalDriver implements LogicFunctionDriver {
...
     });
+    await this.ensureSdkLayer({
+      flatApplication,
+      applicationUniversalIdentifier,
+    });
+  }
+
+  // Symlinks everything from the deps layer except twenty-client-sdk,
+  // which comes from the SDK layer (workspace-specific generated client).
+  private async assembleNodeModules({
```
charlesBochet, line 183: looks a bit too low level, I think this is because of the portal stuff, this should be handled by build tools. but not fully against it, just suprised we have to deal with it, looks like for wrong reasons

packages/twenty-server/src/engine/core-modules/sdk-client/controllers/sdk-client.controller.ts
```diff
@@ -0,0 +1,69 @@
...
+
+@Controller('rest/sdk-client')
+@UseGuards(WorkspaceAuthGuard)
+export class SdkClientController {
+  constructor(
+    private readonly workspaceCacheService: WorkspaceCacheService,
+    private readonly sdkClientGenerationService: SdkClientGenerationService,
+  ) {}
+
+  @Get(':applicationId/:moduleName')
```
charlesBochet, line 32: I'm not against it but it will create a bit of confusion in the code as the applicationId is often read from the JWT token. Current hypothesis is that, we don't add applicationId in endpoints, we read it from token. If present, we take it from there, if not we generate the schema / sdk for the whole app (custom app point of view)

packages/twenty-server/src/engine/core-modules/sdk-client/sdk-client-archive.service.ts
```diff
@@ -0,0 +1,272 @@
...
+    });
+
+    this.logger.log(
+      `Generated SDK client for application ${applicationUniversalIdentifier}`,
+    );
+  }
+
+  // Downloads the archive from file storage and extracts the full
+  // twenty-client-sdk package into targetPackagePath.
+  async downloadAndExtractToPackage({
```
charlesBochet, line 39: I feel the module sdk-client-generation containing downloadAndExtractToPackage is a bit weird naming wise. should the module be named sdk-client and have a sdk-client-generation service and sdk-client-serve service?

packages/twenty-server/src/engine/core-modules/sdk-client-generation/sdk-client-generation.service.ts
```diff
@@ -0,0 +1,272 @@
...
+      workspaceId,
+      applicationUniversalIdentifier,
+    });
+
+    return streamToBuffer(archiveStream);
+  }
+
+  // Reads a single file from the generated SDK archive (e.g. 'dist/core.mjs').
+  // Used to serve individual SDK modules to the frontend worker.
+  async readFileFromArchive({
```
charlesBochet: all these methods do not seems to be sdk specific (likely duplicated with existing logic), we should clean

yarn.lock
```diff
@@ -89,18 +89,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@ai-sdk/groq@npm:^3.0.24":
```
charlesBochet: already on main

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/local.driver.ts
prastoin: It seems like the most lightweigth and reliable way to merge both layers dependencies to me 🤔
Though we could directly remove the twenty-client-sdk out of the deps layer, but we would loose the default stub that throws an error

.github/actions/spawn-twenty-docker-image/action.yaml
prastoin: will simplify in a following PR

packages/twenty-apps/hello-world/yarn.lock
prastoin: That's expected

packages/twenty-server/src/engine/core-modules/sdk-client/controllers/sdk-client.controller.ts
prastoin: How do you do when the passed token is the user's one from the client ? It will be associated with several applicationId

prastoin: as discussed won't do for the moment

charlesBochet (approved):

PR #18545, opened 2026-03-10
Always display a cross button in side panel header when showing a sub page

Devessier:


Devessier:
Closing as the logic is incomplete

PR #18563, opened 2026-03-11
Display a single Add a Section button

Devessier:
- Display a single Add a Section button at the end of the list
- Move other buttons to the section's dropdown menu

https://github.com/user-attachments/assets/b51d8846-635a-477a-9205-bf3266cfcff4


Weiko (approved):
Should we keep "section" for user facing messages and keep group in our codebase?
LGTM

PR #18565, opened 2026-03-11
Fix dashboard creation + role permission page design

thomtrp:
1. **Creating a new dashboard crashes with "Tab not found"** and widgets can't be added after the crash is prevented.

**Root cause:** `initializePageLayout` wrapped both the persisted and draft state updates behind an `isDeeplyEqual` guard. After navigation, `resetPageLayoutEditMode` resets the draft atom to its default but leaves the persisted atom untouched. On re-initialization, `isDeeplyEqual` returns true (persisted unchanged), so the draft is never repopulated. But edit mode is still activated. 

**Fix**: Move the draft store.set outside the isDeeplyEqual guard so it's always set on initialization. Also add a defensive check in `PageLayoutRendererContent` to prevent the crash when activeTabId doesn't match available tabs.

https://github.com/user-attachments/assets/bcd69866-63eb-4e5e-a1bb-655e71ba6dc5

2. **Permission role page design broken**
Before
role-page-broken

After
Capture d’écran 2026-03-11 à 14 25 55

### DIFF
--- a/packages/twenty-front/src/modules/page-layout/components/PageLayoutRendererContent.tsx
+++ b/packages/twenty-front/src/modules/page-layout/components/PageLayoutRendererContent.tsx
@@ -151,9 +151,10 @@ export const PageLayoutRendererContent = () => {
             )}
             defaultEnableXScroll={false}
           >
-            {isDefined(activeTabId) && (
-              <PageLayoutMainContent tabId={activeTabId} />
-            )}
+            {isDefined(activeTabId) &&
+              currentPageLayout.tabs.some(
+                (tab) => tab.id === activeTabId,
+              ) && <PageLayoutMainContent tabId={activeTabId} />}
           </ScrollWrapper>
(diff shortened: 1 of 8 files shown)


packages/twenty-front/src/modules/page-layout/components/PageLayoutRendererContent.tsx
```diff
@@ -151,9 +151,10 @@ export const PageLayoutRendererContent = () => {
             )}
             defaultEnableXScroll={false}
           >
-            {isDefined(activeTabId) && (
-              <PageLayoutMainContent tabId={activeTabId} />
-            )}
+            {isDefined(activeTabId) &&
+              currentPageLayout.tabs.some(
+                (tab) => tab.id === activeTabId,
```
bosiraphael: We could add a boolean constant above to make it more readable

bosiraphael (approved):
LGTM

PR #18568, opened 2026-03-11
fix: replace > button with button selector in styled wrappers

bugisthegod:
fix CSS selector > button not reaching Button component's inner element
### Reproduce
- when you click Data model and create a new field, you will see this bug.
- When you import record and check the height of remove button in the final validation step.

### Root Reason
the Button component internally renders a wrapper div around the <button> element, so > button only reaches the intermediate div, not the actual button.

### Fix
- padding-right not applied → chevron overlapped with text
- ValidationStep: height: 24px not applied → Remove button was 32px instead of 24px

image

image


bugisthegod:
Hi, @lucasbordeau @charlesBochet. Could you help me review this pr? There are only three lines changes.

lucasbordeau (approved):
LGTM

PR #18577, opened 2026-03-12
Fix onboarding flow: workspace creation modal and invite team skip

charlesBochet:
## Summary

- **Fix create-profile modal not showing after workspace creation**: After activating a workspace, `CreateWorkspace.onSubmit` called `refreshObjectMetadataItems()` which only updated the `objectMetadataItemsState` atom but never marked the metadata store as ready (`metadataStoreState` stayed at `'empty'`). Since `MetadataGater` excludes `CreateWorkspace` but not `CreateProfile` from its loading check, navigating to `/create/profile` triggered the skeleton loader instead of the modal. The fix adds the full metadata pipeline after refresh — `updateDraft('objectMetadataItems')` + `applyChanges()` for objects, and `fetchAndLoadIndexViews()` for views — so `isAppMetadataReady` is `true` before navigation.

- **Fix invite-team "Skip" not persisting to server**: Clicking "Skip" on the invite-team page called `setNextOnboardingStatus()` which only updated the local Jotai atom. The early return for empty emails bypassed `sendInvitation`, so the server never cleared the `ONBOARDING_INVITE_TEAM_PENDING` user var. On page refresh, `GetCurrentUser` returned `INVITE_TEAM` and the user was stuck. The fix removes the early return so `sendInvitation({ emails: [] })` always runs — the server handles empty arrays fine and clears the pending flag.

### DIFF
--- a/packages/twenty-front/src/pages/onboarding/CreateWorkspace.tsx
+++ b/packages/twenty-front/src/pages/onboarding/CreateWorkspace.tsx
@@ -9,7 +9,10 @@ import { Logo } from '@/auth/components/Logo';
 import { SubTitle } from '@/auth/components/SubTitle';
 import { Title } from '@/auth/components/Title';
 import { currentWorkspaceState } from '@/auth/states/currentWorkspaceState';
+import { useFetchAndLoadIndexViews } from '@/metadata-store/hooks/useFetchAndLoadIndexViews';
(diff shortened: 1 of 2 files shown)


thomtrp (approved):

FelixMalfait:

Your preview environment is available at: http://bore.pub:14510

PR #18580, opened 2026-03-12
Fix flaky return-to-path e2e tests

charlesBochet:
## Summary

Fixes flaky `return-to-path` e2e tests that were failing intermittently in CI merge queue runs.

**Root cause:** In the multi-workspace environment used by CI (`IS_MULTIWORKSPACE_ENABLED=true`), navigating to `localhost:3001/settings/accounts` triggers a full page redirect to `app.localhost:3001/welcome` via `useRedirectToDefaultDomain`. This redirect is a hard navigation (not a React Router transition), which clears all in-memory Jotai state — including the `returnToPathState` atom that stores the path the user should be redirected to after login. After the redirect, the app has no memory of the intended destination and falls back to `/objects/companies`.

**Fix:** Before performing the cross-domain redirect in `useRedirectToDefaultDomain`, read the `returnToPath` from the Jotai store and pass it as a URL search parameter. On the new page load, `useInitializeQueryParamState` picks it up from the URL and re-hydrates the Jotai atom, preserving the return-to-path across the full page reload.

## Test plan

- [x] Verified locally against production build (`serve -s build`) with `IS_MULTIWORKSPACE_ENABLED=true` — 33/33 consecutive passes of `return-to-path.spec.ts`
- [x] Lint passes (`npx nx lint:diff-with-main twenty-front`)


FelixMalfait:

Your preview environment is available at: http://bore.pub:29929

thomtrp (approved):

PR #18584, opened 2026-03-12
Upgrade Apollo Client to v4 and refactor error handling

FelixMalfait:
## Summary
This PR upgrades Apollo Client from v3.10.0 to v4 and refactors error handling patterns across the codebase to use a new centralized `useSnackBarOnQueryError` hook.

## Key Changes

- **Dependency Update**: Upgraded `@apollo/client` from `^3.10.0` to `^3.11.0` in root package.json
- **New Hook**: Added `useSnackBarOnQueryError` hook for centralized Apollo query error handling with snack bar notifications
- **Error Handling Refactor**: Updated 100+ files to use the new error handling pattern:
  - Removed direct `ApolloError` imports where no longer needed
  - Replaced manual error handling logic with `useSnackBarOnQueryError` hook
  - Simplified error handling in hooks and components across multiple modules
- **GraphQL Codegen**: Updated codegen configuration files to work with Apollo Client v3.11.0
- **Type Definitions**: Added TypeScript declaration file for `apollo-upload-client` module
- **Test Updates**: Updated test files to reflect new error handling patterns

## Notable Implementation Details

- The new `useSnackBarOnQueryError` hook provides a consistent way to handle Apollo query errors with automatic snack bar notifications
- Changes span across multiple feature areas: auth, object records, settings, workflows, billing, and more
- All changes maintain backward compatibility while improving code maintainability and reducing duplication
- Jest configuration updated to work with the new Apollo Client version

https://claude.ai/code/session_019WGZ6Rd7sEHuBg9sTrXRqJ

PR #18585, opened 2026-03-12
Fix AI chat re-renders and refactored code

lucasbordeau:
This PR: 

- Breaks useAgentChatData into focused effect components (streaming, fetch, 
init, auto-scroll, diff sync)
- Splits message list into non-last (stable) + last (streaming/error) to    
prevent full re-renders on each stream chunk
- Adds scroll-to-bottom button and MutationObserver-based auto-scroll on
thread switch
- Lifts loading state from context to atoms
- Adds areEqual to selector
factories

We could improve further but this sets up a robust architecture for further refactoring.

## Messages flow

The flow of messages loading and streaming is now more solid. 

Everything goes out from `AgentChatAiSdkStreamEffect`, whether loaded from the DB or streaming directly, and every consumers is using only one atom `agentChatMessagesComponentFamilyState`

## Data sync effect with callbacks new hook 

See `packages/twenty-front/src/modules/apollo/hooks/useQueryWithCallbacks.ts` which allows to fix Apollo v4 migration leftovers and is an implementation of the pattern we talked about with @charlesBochet 

We could refine this pattern in another PR.

# Before

https://github.com/user-attachments/assets/84e7a96f-6790-405d-8a73-2dacbf783be5

# After

https://github.com/user-attachments/assets/4c692e3a-2413-4513-abcc-44d0da311203


packages/twenty-front/src/modules/ai/components/AgentChatAiSdkStreamEffect.tsx
```diff
@@ -0,0 +1,83 @@
...
+    useEnsureAgentChatThreadIdForSend(createChatThread);
+
+  useListenToBrowserEvent({
+    eventName: AGENT_CHAT_ENSURE_THREAD_FOR_DRAFT_EVENT_NAME,
+    onBrowserEvent: ensureThreadExistsForDraft,
+  });
+
+  const chatState = useAgentChat(agentChatUiMessages, ensureThreadIdForSend);
+
+  const [, setAgentChatMessages] = useAtomComponentFamilyState(
```
charlesBochet: reactive for no reason :)

packages/twenty-front/src/modules/ai/components/AIChatScrollToBottomOnMountEffect.tsx
```diff
@@ -0,0 +1,44 @@
...
+  useEffect(() => {
+    const scrollWrapperElement = document.getElementById(
+      `scroll-wrapper-${AI_CHAT_SCROLL_WRAPPER_ID}`,
+    );
+
+    if (!scrollWrapperElement) {
+      return;
+    }
+
+    let hasScrolled = false;
```
charlesBochet: non standard, is that the best way?

packages/twenty-front/src/modules/ai/hooks/useCreateAgentChatThread.ts
```diff
@@ -0,0 +1,135 @@
...
+} from '~/generated-metadata/graphql';
+
+export const useCreateAgentChatThread = () => {
+  const setCurrentAIChatThread = useSetAtomState(currentAIChatThreadState);
+  const setAgentChatInput = useSetAtomState(agentChatInputState);
+  const setAgentChatUsage = useSetAtomState(agentChatUsageState);
+  const setCurrentAIChatThreadTitle = useSetAtomState(
+    currentAIChatThreadTitleState,
+  );
+  const [, setIsCreatingChatThread] = useAtomState(isCreatingChatThreadState);
```
charlesBochet: same here

packages/twenty-front/src/modules/ai/hooks/useCreateAgentChatThread.ts
```diff
@@ -0,0 +1,135 @@
...
+  const setAgentChatDraftsByThreadId = useSetAtomState(
+    agentChatDraftsByThreadIdState,
+  );
+  const store = useStore();
+  const apolloClient = useApolloClient();
+
+  const [createChatThread] = useCreateChatThreadMutation({
```
charlesBochet: rework based on apollo4 upgrade (see #18584)

packages/twenty-front/src/modules/ai/hooks/useInitialAgentChatThreadQuery.ts
```diff
@@ -0,0 +1,75 @@
...
+export const useInitialAgentChatThreadQuery = (currentAIChatThread: string) => {
+  const setCurrentAIChatThread = useSetAtomState(currentAIChatThreadState);
+  const setAgentChatInput = useSetAtomState(agentChatInputState);
+  const setAgentChatUsage = useSetAtomState(agentChatUsageState);
+  const setCurrentAIChatThreadTitle = useSetAtomState(
+    currentAIChatThreadTitleState,
+  );
+  const store = useStore();
+
+  const { loading: threadsLoading } = useGetChatThreadsQuery({
```
charlesBochet: same

.mcp.json
```diff
@@ -2,8 +2,8 @@
   "mcpServers": {
     "postgres": {
       "type": "stdio",
-      "command": "npx",
-      "args": ["-y", "@modelcontextprotocol/server-postgres", "${PG_DATABASE_URL}"],
+      "command": "bash",
+      "args": ["-c", "source packages/twenty-server/.env && npx -y @modelcontextprotocol/server-postgres \"$PG_DATABASE_URL\""],
```
charlesBochet, line 6: why?

charlesBochet: ok

charlesBochet (commented):
LGTM, the only thing we should re-work is to migrate to the new apollo v4

packages/twenty-front/src/modules/ai/components/AIChatTabMessageList.tsx
```diff
@@ -15,28 +19,38 @@ const StyledScrollWrapperContainer = styled.div`
...
 
-  if (!hasMessages) {
+  if (!agentChatHasMessage) {
     return null;
   }
 
   return (
-    <StyledScrollWrapperContainer>
+    <StyledScrollWrapperContainer
+      style={{
```
charlesBochet, line 41: we should never pass styled like this. Either it should be part of the intrinsec API of the ScrollWrapper to handle this, either you can add a container)

packages/twenty-front/src/modules/ai/hooks/useCreateAgentChatThread.ts
```diff
@@ -0,0 +1,134 @@
...
+      setAgentChatInput(newDraft);
+      setCurrentAIChatThreadTitle(null);
+      setAgentChatUsage(null);
+
+      const newThread = data.createChatThread;
+      const threadListVariables = {
+        paging: { first: CHAT_THREADS_PAGE_SIZE },
+      };
+      const existingThreadsCache =
+        apolloClient.cache.readQuery<GetChatThreadsQuery>({
```
charlesBochet: is that mandatory to read in cache? don't we have an interdermediary state?

packages/twenty-front/src/modules/ai/hooks/useInitialAgentChatThreadQuery.ts
```diff
@@ -0,0 +1,89 @@
...
+import { currentAIChatThreadState } from '@/ai/states/currentAIChatThreadState';
+import { currentAIChatThreadTitleState } from '@/ai/states/currentAIChatThreadTitleState';
+import { hasTriggeredCreateForDraftState } from '@/ai/states/hasTriggeredCreateForDraftState';
+import { useSetAtomState } from '@/ui/utilities/state/jotai/hooks/useSetAtomState';
+
+import { useQuery } from '@apollo/client/react';
+import { useEffect } from 'react';
+import { GetChatThreadsDocument } from '~/generated-metadata/graphql';
+
+export const useInitialAgentChatThreadQuery = (currentAIChatThread: string) => {
```
charlesBochet: hook with effect init will likely lead to effect being added without the caller noticing it? can we use a ComponentEffect here?

lucasbordeau: Quite relevant indeed, this is another use case for our new query with callbacks hook.

charlesBochet (approved):

PR #18593, opened 2026-03-12
Fix app install file upload

martmull:
remove wrong file path based file selection


packages/create-twenty-app/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "create-twenty-app",
-  "version": "0.7.0-canary.0",
+  "version": "0.7.0-canary.1",
```
prastoin: are we pushing canary ? what if we publish in prod ?

martmull: don't really know what we should publish, but i would like to publish a new version

packages/twenty-server/src/engine/core-modules/application/application-install/application-install.service.ts
```diff
@@ -177,47 +179,38 @@ export class ApplicationInstallService {
...
 
-    if (isDefined(exact)) {
-      return exact;
-    }
+    files.push(
+      { relativePath: 'package.json', fileFolder: FileFolder.Dependencies },
+      { relativePath: 'manifest.json', fileFolder: FileFolder.Source },
+    );
 
-    for (const { pattern, folder } of FILE_FOLDER_PATTERN_MAPPING) {
```
prastoin, line 187: ***Question:*** What was the issue with the previous implem ?
 Is this only a refactor in this file ?

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-resource/logic-function-resource.service.ts
```diff
@@ -239,22 +240,52 @@ export class LogicFunctionResourceService {
...
-        workspaceId,
-        applicationUniversalIdentifier,
-        fileFolder: FileFolder.Dependencies,
-        resourcePath: 'yarn.lock',
-        localPath: join(inMemoryFolderPath, 'yarn.lock'),
-      }),
-    ]);
+    );
+
+    if (yarnLockExists) {
```
prastoin, line 262: ***Question:*** Shouldn't this always be the case ?

martmull: npm pack does not pack yarn.lock

prastoin: When does it exist then ?

packages/twenty-server/src/engine/core-modules/application/application-install/application-install.service.ts
martmull: FILE_FOLDER_PATTERN_MAPPING was based on filename to detect if we have a logic-function or a front-component, it is not the case anymore. The source of truth is on the manifest

prastoin: Ok great

prastoin (approved):

PR #18594, opened 2026-03-12
Add dataloader and read from cache for view entities

Weiko:
## Context
Improve view resolution using cache and dataloader

## Performance Comparison

|Run|Main (no DataLoaders/cache)|Feature Branch (DataLoaders + cache)|Speedup|
|---|---|---|---|
|1 (cold)|418ms|95ms|~4.4x faster|
|2|42ms|19ms|~2.2x faster|
|3|37ms|19ms|~1.9x faster|
|4|39ms|12ms|~3.2x faster|
|5|33ms|13ms|~2.5x faster|

The biggest improvement is to use dataloaders for the multiple relations associated with views. Cache is a bit less significant since there are other cache mechanism such as PostgreSQL buffer cache but it will probably be more meaningful with bigger workspaces


packages/twenty-server/src/engine/metadata-modules/view/services/view.service.ts
```diff
@@ -372,103 +381,269 @@ export class ViewService {
...
-      }
+    if (!isDefined(flatView) || flatView.deletedAt !== null) {
+      return null;
+    }
 
-      return false;
+    return fromFlatViewToViewDto(flatView);
+  }
+
+  async findManyWithRelationsFromCache(
```
Weiko, line 490: Not a big fan of this, this kind of defy the purpose of this PR but this is needed for REST

packages/twenty-server/src/engine/api/common/common-query-runners/common-group-by-query-runner.service.ts
```diff
@@ -263,7 +264,7 @@ export class CommonGroupByQueryRunnerService extends CommonBaseQueryRunnerServic
       },
     });
 
-    let view: ViewEntity | null = viewFilters[0]?.view;
+    let view: ViewEntity | ViewDTO | null = viewFilters[0]?.view;
```
charlesBochet: typing looks very wrong

charlesBochet (approved):

FelixMalfait:

Your preview environment is available at: http://bore.pub:55220

PR #18605, opened 2026-03-12
Fix relation connect where failing on mixed-case email and URL

bugisthegod:
Related to issue #17711 
Follow-up to PR #17774  which fixed the frontend link normalization only
## Summary
- Normalize `primaryEmail` (lowercase) and `primaryLinkUrl` in relation `connect.where` composite values
- Applied in both frontend spreadsheet import and backend `DataArgProcessorService` to cover all entry points (UI import, GraphQL API)


packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/data-arg-processor.service.ts
```diff
@@ -260,6 +261,24 @@ export class DataArgProcessorService {
           );
         }
 
+        const connectOperation = value as Record<
+          string,
+          Record<string, unknown>
+        >;
+        const connectWhere = connectOperation.connect?.where;
+
+        if (isDefined(connectWhere) && typeof connectWhere === 'object') {
```
etiennejouan: could be nice to create a isRelationConnectWithWhere guard here, it then enables to avoid casting at line 276. wdyt ?

etiennejouan (commented):
Hi @bugisthegod, thank you for your new contribution ! Sorry for the late review. 

I really appreciate your approach but wondering if we can take advantage of the already existing processField method (instead of the transform-relation-connect-... util).

The idea would be to call processField to process the where arguments. This way any normalisation added to processField (new composite types, new transforms, etc.) automatically applies to connect-where values too. No more hardcoded primaryLinkUrl / primaryEmail special-casing.

Wdyt ? If not clear, you can reach me on discord or I can also make the changes on your branch.

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/utils/is-defined-object.util.ts
```diff
@@ -0,0 +1,7 @@
+import { isDefined } from 'twenty-shared/utils';
+
+export const isDefinedObject = (
```
etiennejouan: isObject from @sniptt/guards is the util we use in the whole codebase

etiennejouan (approved):
Hi @bugisthegod, thank you for your contribution! 
It was a blind spot I had wanted to fill for a long time. Very nice job. 
I took the liberty of adding some update. I did not want to bother you with minor feedback.

bugisthegod:
> Hi @bugisthegod, thank you for your contribution! It was a blind spot I had wanted to fill for a long time. Very nice job. I took the liberty of adding some update. I did not want to bother you with minor feedback.

Thanks a lot for the review and for pushing the improvements yourself!🤗 I'm happy to contribute for twenty! Feel free to ping me anytime if you think there's something I can help with!

PR #18614, opened 2026-03-13
fix: split tsvector migration, add configurable DB timeout, reorder 1.19 commands

charlesBochet:
## Summary

- **Split tsvector migration into individual per-field transactions**: each tsvector field now runs in its own `workspaceMigrationRunnerService.run()` call (its own DB transaction). Since STORED generated columns trigger full table rewrites, a timeout on one large table (e.g. `timelineActivity`) no longer rolls back the others. Each field has its own idempotency check, so the migration is fully resumable.
- **Add configurable `DATABASE_STATEMENT_TIMEOUT_MS` env var** (default 15000ms): controls the `query_timeout` on the core datasource globally, allowing operators to raise it for long-running upgrade commands without code changes.
- **Reorder 1.19 upgrade commands**: move `fixRoleAndAgentUniversalIdentifiersCommand` first so that subsequent commands see corrected universal identifiers.

### DIFF
--- a/packages/twenty-server/src/database/typeorm/core/core.datasource.ts
+++ b/packages/twenty-server/src/database/typeorm/core/core.datasource.ts
@@ -72,7 +72,7 @@ export const typeORMCoreModuleOptions: TypeOrmModuleOptions = {
         }
       : undefined,
   extra: {
-    query_timeout: 15000,
+    query_timeout: Number(process.env.DATABASE_STATEMENT_TIMEOUT_MS ?? 15000),
   },
 };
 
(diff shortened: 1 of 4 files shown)


thomtrp (approved):

FelixMalfait:

Your preview environment is available at: http://bore.pub:45566

packages/twenty-server/src/database/typeorm/core/core.datasource.ts
```diff
@@ -72,7 +72,7 @@ export const typeORMCoreModuleOptions: TypeOrmModuleOptions = {
         }
       : undefined,
   extra: {
-    query_timeout: 15000,
+    query_timeout: Number(process.env.DATABASE_STATEMENT_TIMEOUT_MS ?? 15000),
```
Weiko, line 75: Shouldn't we use twenty config service?

PR #18615, opened 2026-03-13
fix: add missing React key props to ButtonGroup and FloatingButtonGro…

williamjusticedavis:
Fix missing React key props on ButtonGroup and FloatingButtonGroup story children                                                                                      
                  
JSX element arrays defined in Storybook args require explicit key props, otherwise React emits a "missing key" warning in development. This adds keys to the children arrays in ButtonGroup.stories.tsx and FloatingButtonGroup.stories.tsx.


charlesBochet (approved):
Thank you :)

PR #18616, opened 2026-03-13
fix: add placeholders for required fields when creating records via quick-create

FelixMalfait:
## Summary
Fixes the error `Field "title" of required type "String!" was not provided` when creating records via "Add New" with minimal input (e.g. Playbook with only `{ position: 'last' }`).

## Changes
- **New utility** `getRecordInputPlaceholdersForRequiredFields`: fills missing required non-relation fields with safe defaults before the create mutation
- **Label identifier**: uses `Untitled {objectLabelSingular}` (e.g. "Untitled Playbook")
- **Other TEXT fields**: uses "Untitled"
- **Other types**: UUID, BOOLEAN, NUMBER, DATE, SELECT, FULL_NAME, LINKS, ADDRESS, EMAILS, PHONES, RICH_TEXT_V2, CURRENCY — each gets an appropriate placeholder
- **Required relations** (MANY_TO_ONE): cannot be auto-filled — throws a clear error with field labels instead of opaque GraphQL error
- **Merge order**: `{ ...placeholders, ...recordInput }` so filters/RLS/explicit input override placeholders
- **Bug fix**: `setLoading(false)` in catch block when mutation fails
- **Tests**: unit tests for placeholder logic and error message formatting

## Testing
- `npx jest packages/twenty-front/src/modules/object-record/utils/__tests__/getRecordInputPlaceholdersForRequiredFields.test.ts` — all 6 tests pass

### DIFF
--- a/packages/twenty-front/src/modules/object-record/hooks/useCreateOneRecord.ts
+++ b/packages/twenty-front/src/modules/object-record/hooks/useCreateOneRecord.ts
@@ -1,6 +1,7 @@
 import { useState } from 'react';
 import { v4 } from 'uuid';
 
+import { type ApolloCache, type FetchResult } from '@apollo/client';
 import { triggerCreateRecordsOptimisticEffect } from '@/apollo/optimistic-effect/utils/triggerCreateRecordsOptimisticEffect';
 import { triggerDestroyRecordsOptimisticEffect } from '@/apollo/optimistic-effect/utils/triggerDestroyRecordsOptimisticEffect';
 import { currentWorkspaceMemberState } from '@/auth/states/currentWorkspaceMemberState';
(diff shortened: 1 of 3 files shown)


packages/twenty-front/src/modules/object-record/hooks/useCreateOneRecord.ts
```diff
@@ -86,10 +92,26 @@ export const useCreateOneRecord = <
 
     const idForCreation = recordInput.id ?? v4();
 
+    const { placeholders, missingRequiredRelationFields } =
+      getRecordInputPlaceholdersForRequiredFields(objectMetadataItem, recordInput);
```
charlesBochet: I feel this is duplicated with pre-existing logic: prefill-record, generateEmptyFieldValue

I'm very surprised we need this kind of util at this stage of the project

charlesBochet (changes requested):

packages/twenty-front/src/modules/object-record/hooks/useCreateOneRecord.ts
charlesBochet: why is title required in the first place? we have always avoided making anything required in the GQL schema (we have an edge case for relations)

FelixMalfait: It came through the AI chat, it's possible for apps / via the API

FelixMalfait: I was hesitant on disabling it there but I don't feel it's the right long-term path (valid to want something nullable on the backend)

FelixMalfait:

Your preview environment is available at: http://bore.pub:34043

FelixMalfait:
Closing in favor of a broader redesign. See issue #TBD for the creation forms proposal.

FelixMalfait:
Superseded by twentyhq/core-team-issues#2336

PR #18652, opened 2026-03-14
fix: respect standard object rename across all locales (Closes #18650)

Assad-Robles:
### Problem
When a user renames a standard object (for example, changing **"People" → "Contacts"**), the custom name is only respected when the UI language is set to **English**.

For any other locale, the resolver ignores the user-defined override and falls back to the i18n translation of the original default label.

As a result, the custom name defined by the user is not displayed when the UI language changes.

### Root Cause
`resolveObjectMetadataStandardOverride` only applied direct overrides when the locale matched `SOURCE_LOCALE` (English):

```ts
if (
  safeLocale === SOURCE_LOCALE &&
  isNonEmptyString(objectMetadata.standardOverrides?.[labelKey])
) {
  return objectMetadata.standardOverrides[labelKey] ?? '';
}


charlesBochet (approved):

charlesBochet:
Thank you @Assad-Robles, this looks to be the right fix

PR #18658, opened 2026-03-15
[Fix] : Standard View Names to update dynamically 

carbonFibreCode:
fixes #18607 

- in this PR we have added the useViewName and getDynamicView hook and util respectively with existing hardcoded strings to prevent db and code mismatch. 

Suggested migration strategy:
Using the feature flags false for default and the other PR containing the migration command along with relevant cache invalidation. 

Before :
Screenshot 2026-03-15 at 7 43 32 PM

After : 
Screenshot 2026-03-15 at 8 45 47 PM

Please let me know deployment strategy as this might involve migration commands, also internationalisation.


charlesBochet:
@carbonFibreCode I'm sorry but this does not seem to be the right way of doing it. I really appreciate the contribution and the time you took to make a reasonable PR

Our current approach is:
- views will be created by backend (with {objectLabelPlural} placeholder)
- the backend resolver/controller will compute the proper name, this should already be in place, the frontend should not do it
- we should actually have no fallback in front end

I'm not sure why it's working already actually, it should be and we should not add an extra layer

I'm going to close this PR and the follow up

Thanks again for your help

PR #18663, opened 2026-03-16
Fix custom object navigation menu items missing icon and color

### LINKED ISSUE #2299: Custom object navigation menu items created without icon and color
## Summary
When a custom object is created, the corresponding navigation menu item is created with `icon: null` and `color: null`. As a result, those items appear in the sidebar without an icon or color (e.g. Pet Care Agreements, Survey results, or any other custom object).

## Expected behavior
Navigation menu items for custom objects should get the object’s icon and a deterministic default color when they are created, so they match the behavior of standard objects and are easy to distinguish in the sidebar.

## Root cause
In `object-metadata.service.ts`, `computeFlatNavigationMenuItemToCreate` built the navigation menu item with `icon: null` and `color: null` and did not receive the new object’s metadata, so icon and color were never set for custom objects.

abdulrahmancodes:
Closes [#2299](https://github.com/twentyhq/core-team-issues/issues/2299)

## Fix

Custom object navigation menu items were created with `icon: null` and `color: null`, so entries like Pet Care Agreements and Survey results had no icon or color in the sidebar.

## Changes

- **Server:** When creating a nav item for a new object, pass the object into `computeFlatNavigationMenuItemToCreate` and set `icon` from the object and `color` via a shared default.
- **Shared (twenty-shared):** Add `MAIN_COLOR_NAMES` as the single source of truth for theme color names; add `getDefaultColorForCustomObject(nameSingular)` that picks a stable default color (hash over name, colors = `MAIN_COLOR_NAMES` minus `gray`). Twenty-ui now imports and uses this list and types `MAIN_COLORS_LIGHT` against it to avoid drift.
- **Frontend:** Use `getDefaultColorForCustomObject` from twenty-shared in `getStandardObjectIconColor` for custom objects so fallback matches server-assigned colors.
- **Twenty-ui:** Fix readonly `MAIN_COLOR_NAMES` usage (readonly casts, `ReadonlyArray` in `getNextThemeColor`, `[...MAIN_COLOR_NAMES]` for zod and stories).

New and existing custom objects now get a deterministic icon and color in the sidebar; server and frontend stay in sync.


packages/twenty-ui/src/display/status/components/Status.tsx
```diff
@@ -4,7 +4,7 @@ import { type ThemeColor, MAIN_COLOR_NAMES } from '@ui/theme';
 import { themeCssVariables } from '@ui/theme-constants';
 
 const parseThemeColor = (color: string): ThemeColor =>
-  (MAIN_COLOR_NAMES as string[]).includes(color)
+  (MAIN_COLOR_NAMES as readonly string[]).includes(color)
```
Devessier: Could we avoid all the type castings by changing the type of `MAIN_COLOR_NAMES`?

Currently, it's typed as a tuple (`[red, blue, etc.]`). I don't think there is any benefit to that. Instead, we could type it as a string array. It seems to be more practical.

Devessier (commented):
The `[...MAIN_COLOR_NAMES]` look a bit hacky to me. I would simplify the typing of the exported constant.

Devessier (approved):

charlesBochet:
@abdulrahmancodes I'm refactoring a bit NavigationMenuItems, let's wait for this PR to be merged: https://github.com/twentyhq/twenty/pull/18672

abdulrahmancodes:
It seems to have already been fixed by #18672

PR #18670, opened 2026-03-16
refactor: consolidate database query timeout config variables

charlesBochet:
## Summary

- Removes the redundant `DATABASE_STATEMENT_TIMEOUT_MS` config variable (default 15s) from `ConfigVariables`
- Updates the core TypeORM datasource to use `PG_DATABASE_PRIMARY_TIMEOUT_MS` (default 10s) for its `query_timeout`, aligning it with the workspace datasource which already uses this variable
- This consolidates two separate env vars that controlled the same concern (database query timeout) into a single one

### DIFF
--- a/packages/twenty-server/src/database/typeorm/core/core.datasource.ts
+++ b/packages/twenty-server/src/database/typeorm/core/core.datasource.ts
@@ -72,7 +72,7 @@ export const typeORMCoreModuleOptions: TypeOrmModuleOptions = {
         }
       : undefined,
   extra: {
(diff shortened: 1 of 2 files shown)


FelixMalfait:

Your preview environment is available at: http://bore.pub:45362

FelixMalfait (approved):
Nit: we should probably get rid of PG_DATABASE_REPLICA_TIMEOUT_MS too since you probably don't know if the repo will hit the replica or not so it doesn't make sense to have 2 different timeouts

PR #18671, opened 2026-03-16
fix: return method-specific MCP responses

### LINKED ISSUE #18524: MCP Server returns initialize-style response for all methods (breaks tools/list)
## Summary

Twenty's MCP server at `/mcp` returns the full initialization response envelope for **every** JSON-RPC method, not just `initialize`. This breaks standard MCP clients (e.g. rmcp/Rust) that expect method-specific response shapes.

## Steps to Reproduce

1. Enable MCP via `IS_AI_ENABLED` feature flag
2. Send a valid `initialize` request — works fine
3. Send a `tools/list` request:

```bash
curl -X POST http://localhost:3000/mcp \
  -H "Authorization: Bearer <API_KEY>" \
  -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}}'
```

## Expected Response

Per MCP spec, `tools/list` should return:
```json
{
  "id": 2,
  "jsonrpc": "2.0",
  "result": {
    "tools": [...]
  }
}
```

## Actual Response

Twenty returns the full init envelope for every method:
```json
{
  "id": 2,
  "jsonrpc": "2.0",
  "result": {
    "capabilities": {"tools": {"listChanged": false}, "resources": {...}, "prompts": {...}},
    "tools": [...],
    "resources": [],
    "prompts": [],
    "metadata": {...},
    "protocolVersion": "2024-11-05",
    "serverInfo": {"name": "Twenty MCP Server", "version": "0.0.1"}
  }
}
```

The extra fields (`capabilities`, `protocolVersion`, `serverInfo`, `metadata`, `resources`, `prompts`) cause strict MCP clients to fail deserialization.

## Additional Notes

- HTTP status is `201 Created` instead of `200 OK` — also non-standard for Streamable HTTP transport
- No `Mcp-Session-Id` header in responses (stateless mode)
- The tools themselves work correctly — `http_request`, `send_email`, `draft_email`, `search_help_center`, `code_interpreter` are all returned
- The `initialize` response itself is correct
- Tested with Twenty v1.18.1 and rmcp v1.1 (Rust MCP client used by [Spacebot](https://github.com/spacedriveapp/spacebot))

## Environment

- Twenty v1.18.1 (Docker: `twentycrm/twenty:v1.18.1`)
- MCP protocol version: `2024-11-05`
- Client: rmcp 1.1 (Rust), also verified with curl

channi23:
## Summary

Fixes #18524 

Fixes the MCP response contract for non-`initialize` methods.

Previously, `/mcp` returned initialize-style metadata for methods like `tools/list`, which caused strict MCP clients to reject the response shape. The endpoint also returned `201 Created` for RPC calls even though no resource was being created.

## Changes

- return only method-specific payloads for MCP list methods
  - `tools/list` -> `{ tools: [...] }`
  - `prompts/list` -> `{ prompts: [] }`
  - `resources/list` -> `{ resources: [] }`
- keep MCP server metadata only on `initialize`
- make `/mcp` return `200 OK` instead of `201 Created`
- add regression tests for:
  - `tools/list` response shape
  - `prompts/list` response shape
  - `resources/list` response shape

## Why

Strict MCP clients expect:
- standard RPC transport semantics over HTTP
- method-specific JSON-RPC result payloads

Returning initialize metadata for non-`initialize` methods breaks that expectation and can cause client deserialization or protocol validation failures.

## Verification

- reproduced the issue locally against `/mcp`
- verified `tools/list` was previously returning initialize-style fields
- verified `tools/list` now returns only `result.tools`
- verified `/mcp` now returns `200 OK`
- ran targeted Jest tests:

```bash
cd /Users/apple/MyProjects/OpenSource/twenty/packages/twenty-server
npx jest --runInBand src/engine/api/mcp/services/__tests__/mcp-protocol.service.spec.ts src/engine/api/mcp/services/__tests__/mcp-tool-executor.service.spec.ts


FelixMalfait (approved):
Thank you!

PR #18676, opened 2026-03-16
fix: add viewFieldGroupId to ViewField fragment and connect page layout selectors to live metadata store

ehconitin:


Devessier (approved):
Good for me!

ehconitin:
@charlesBochet merging as it is to unblock me!
let me know if its fine with you!

PR #18678, opened 2026-03-16
fix: allow identical singular and plural labels for objects

FelixMalfait:
## Summary

Closes #18673

Some languages (e.g., German "Unternehmen") and even English words (sheep, deer, aircraft, series) have identical singular and plural forms. Twenty previously blocked saving when labels matched, making it impossible to correctly name objects in these cases.

- **Labels** are purely display strings — removed the equality validation from both the frontend Zod schema and backend validator
- **API names** (nameSingular/namePlural) must stay different since they generate distinct GraphQL resolvers (`findOne` vs `findMany`, `createOne` vs `createMany`, etc.) and REST endpoints — this validation is preserved
- Added a shared `computeMetadataNamesFromLabels` util in `twenty-shared` that auto-appends `'s'` to the plural API name when both labels produce the same camelCase name (e.g., "Unternehmen" → `unternehmen` / `unternehmens`)
- Both the frontend form and backend sync-check use the same shared util — single source of truth, no duplicated logic

**No retroactive impact**: since the old code prevented identical labels from ever being saved, no existing workspace has `labelSingular === labelPlural`.

## Test plan

- [x] New unit tests for `computeMetadataNamesFromLabels` (7 tests: standard labels, Sheep, Unternehmen, Aircraft, empty labels, different labels, applyCustomSuffix)
- [x] Updated frontend schema validation tests (identical labels with different names now passes; identical names still fails)
- [x] Updated backend integration test cases (removed identical-label failing cases)
- [ ] Manual: create a new object with identical singular/plural labels (e.g. "Sheep" / "Sheep") — should save successfully with API names `sheep` / `sheeps`
- [ ] Manual: verify existing objects with different labels still work unchanged


FelixMalfait:

Your preview environment is available at: http://bore.pub:22034

packages/twenty-front/src/pages/settings/data-model/utils/computeMetadataNamesFromLabels.ts
```diff
@@ -0,0 +1,13 @@
+import { computeMetadataNamesFromLabels as computeMetadataNamesFromLabelsCore } from 'twenty-shared/metadata';
+
+// Frontend-specific wrapper that returns empty strings on error instead of throwing
+export const computeMetadataNamesFromLabels = (
+  labelSingular: string,
+  labelPlural: string,
+): { nameSingular: string; namePlural: string } => {
+  try {
+    return computeMetadataNamesFromLabelsCore({ labelSingular, labelPlural });
```
charlesBochet: by convention the util should be suffixed by orThrow to make clear that it can throw

charlesBochet (approved):
Left one comment, approving though LGTM if it works from a product perspective

PR #18679, opened 2026-03-16
[Chore] : Migration of Standard View Names 

carbonFibreCode:
Migration Command in response to PR #18658


charlesBochet:
See my comment on https://github.com/twentyhq/twenty/pull/18658

PR #18681, opened 2026-03-16
[Command menu items] Create engine commands

bosiraphael:
## Description

- Introduces a new engine command execution model that replaces the previous approach of mapping `EngineComponentKey` to React components. Instead, engine commands are now mounted headlessly via `HeadlessEngineCommandMountRoot`, with their execution context populated synchronously before mounting.
- Creates new headless command components
- Moves error handling from the SDK layer to the host app by wrapping all mounted commands with a new `CommandMenuItemErrorBoundary`

The new flow works as follows:
- When a command menu item with an `engineComponentKey` is clicked, `useCommandMenuItemFrontComponentCommands` calls `useMountEngineCommand`, which synchronously reads the current context store (object metadata, selected records, filters, view ID, etc.) and writes a `MountedEngineCommandContext` into `mountedEngineCommandsState`.
- The command is then mounted into `mountedEngineCommandsState`, which triggers `HeadlessEngineCommandMountRoot` to render the corresponding headless component from `ENGINE_COMPONENT_KEY_HEADLESS_COMPONENT_MAP`, wrapped in `CommandMenuItemErrorBoundary`, `ContextStoreComponentInstanceContext.Provider`, and `EngineCommandComponentInstanceContext.Provider`.
- Each command component reads its execution context and delegates to one of the 4 execution patterns: `HeadlessEngineCommandWrapperEffect` (simple actions), `HeadlessConfirmationModalEngineCommandEffect` (destructive actions needing confirmation), `HeadlessNavigateEngineCommand` (GO_TO_* commands), or `HeadlessOpenSidePanelPageEngineCommand` (SEARCH_RECORDS, ASK_AI, VIEW_PREVIOUS_AI_CHATS).
- After execution, the command self-unmounts via `useUnmountEngineCommand`, which removes the entry from `mountedEngineCommandsState` and stops rendering the component.


packages/twenty-front/src/modules/command-menu-item/engine-command/components/HeadlessEngineCommandWrapperEffect.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+    setIsInitialized(true);
+
+    const run = async () => {
+      await execute();
+
+      unmountEngineCommand(engineCommandId);
+    };
+
+    run();
```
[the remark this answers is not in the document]
bosiraphael: It is caught up above and unmounted in the error boundary

packages/twenty-front/src/modules/command-menu-item/confirmation-modal/components/CommandMenuConfirmationModalManager.tsx
```diff
@@ -23,7 +23,7 @@ export const CommandMenuConfirmationModalManager = () => {
     commandMenuItemConfirmationModalConfigState,
   );
 
-  const callerId = commandMenuItemConfirmationModalConfig?.frontComponentId;
+  const callerId = commandMenuItemConfirmationModalConfig?.callerId;
```
charlesBochet: callerId is too generic, i would prefer a union type:
caller: frontComponentId | engineKey or something like this

packages/twenty-front/src/modules/command-menu-item/engine-command/hooks/useIsHeadlessEngineCommandEffectInitialized.ts
```diff
@@ -0,0 +1,16 @@
+import { useRef } from 'react';
+
+export const useIsHeadlessEngineCommandEffectInitialized = () => {
+  // eslint-disable-next-line twenty/no-state-useref
+  const isInitializedRef = useRef(false);
+
+  const getIsInitialized = () => {
```
charlesBochet: why returning a callback here and not the ref direclty?

bosiraphael: We could indeed

packages/twenty-front/src/modules/command-menu-item/engine-command/components/HeadlessEngineCommandMountRoot.tsx
```diff
@@ -0,0 +1,41 @@
...
+  const mountedEngineCommands = useAtomStateValue(mountedEngineCommandsState);
+  const unmountEngineCommand = useUnmountEngineCommand();
+
+  return (
+    <>
+      {[...mountedEngineCommands.entries()].map(
+        ([engineCommandId, mountContext]) => (
+          <CommandMenuItemErrorBoundary
+            key={engineCommandId}
+            resetKeys={[engineCommandId]}
```
charlesBochet: why is that an array?

charlesBochet: not sure what it means too (from an external api point of view)

charlesBochet: or why we need it

packages/twenty-front/src/modules/command-menu-item/engine-command/components/HeadlessNavigateEngineCommand.tsx
```diff
@@ -0,0 +1,24 @@
+import { HeadlessEngineCommandWrapperEffect } from '@/command-menu-item/engine-command/components/HeadlessEngineCommandWrapperEffect';
+import { type PathParam, useNavigate } from 'react-router-dom';
+import { type AppPath } from 'twenty-shared/types';
+import { getAppPath } from 'twenty-shared/utils';
+
+export const HeadlessNavigateEngineCommand = <T extends AppPath>({
```
charlesBochet, line 6: looks duplicated with packages/twenty-front/src/modules/command-menu-item/engine-command/components/HeadlessEngineCommandWrapperEffect.tsx

packages/twenty-front/src/modules/command-menu-item/engine-command/components/HeadlessOpenSidePanelPageEngineCommand.tsx
```diff
@@ -0,0 +1,37 @@
+import { HeadlessEngineCommandWrapperEffect } from '@/command-menu-item/engine-command/components/HeadlessEngineCommandWrapperEffect';
+import { useNavigateSidePanel } from '@/side-panel/hooks/useNavigateSidePanel';
+import { sidePanelSearchState } from '@/side-panel/states/sidePanelSearchState';
+import { useSetAtomState } from '@/ui/utilities/state/jotai/hooks/useSetAtomState';
+import { type MessageDescriptor } from '@lingui/core';
+import { t } from '@lingui/core/macro';
+import { type SidePanelPages } from 'twenty-shared/types';
+import { type IconComponent } from 'twenty-ui/display';
+
+export const HeadlessOpenSidePanelPageEngineCommand = ({
```
charlesBochet, line 10: same here. not sure what's the value here to have different Command Component

Not strongly against it but I don't see what's special

packages/twenty-front/src/modules/command-menu-item/engine-command/types/MountedEngineCommandContext.ts
```diff
@@ -0,0 +1,19 @@
+import { type ContextStoreTargetedRecordsRule } from '@/context-store/states/contextStoreTargetedRecordsRuleComponentState';
```
charlesBochet, line 1: file name should not be .types

packages/twenty-front/src/modules/command-menu-item/engine-command/states/mountedEngineCommandsState.ts
```diff
@@ -0,0 +1,9 @@
+import { type MountedEngineCommandContext } from '@/command-menu-item/engine-command/types/MountedEngineCommandContext.types';
+import { createAtomState } from '@/ui/utilities/state/jotai/utils/createAtomState';
+
+export const mountedEngineCommandsState = createAtomState<
```
charlesBochet, line 4: not a fan of the naming: what does "mounted" brings?
why not a familyState rather than a map?

I would have done: 

engineCommandContextApiFamilyState or something like this

Also I wonder why it's different from FrontComponentCase here? Can't we align? Why would the Engine need a state for ContextApi and not the FCRenderer?

In my mind, FCRenderer is a type of EngineCommand

packages/twenty-front/src/modules/command-menu-item/engine-command/types/MountedEngineCommandContext.ts
```diff
@@ -0,0 +1,19 @@
...
+import { type ObjectMetadataItem } from '@/object-metadata/types/ObjectMetadataItem';
+import { type ObjectRecord } from '@/object-record/types/ObjectRecord';
+import {
+  type Nullable,
+  type RecordGqlOperationFilter,
+} from 'twenty-shared/types';
+import { type EngineComponentKey } from '~/generated-metadata/graphql';
+
+export type MountedEngineCommandContext = {
```
charlesBochet: .types file

packages/twenty-front/src/modules/front-components/components/HeadlessFrontComponentMountRoot.tsx
```diff
@@ -16,34 +18,37 @@ export const HeadlessFrontComponentMountRoot = () => {
...
-                      }
-                    : undefined,
-                layoutType: PageLayoutType.DASHBOARD,
-                isInSidePanel: false,
-              }}
-            >
-              <FrontComponentRenderer frontComponentId={frontComponentId} />
-            </LayoutRenderingProvider>
-          </Suspense>
+          <CommandMenuItemErrorBoundary
```
charlesBochet, line 27: I feel it's just another type of EngineCommand, we could unify

charlesBochet (commented):
I've left comments but I feel what's have been produced is maintainable and scalable so I'm not blocking it!

charlesBochet (approved):

PR #18700, opened 2026-03-17
[FIx]: using the dynamic {objectLabelPlural} instead of hardcoded name

carbonFibreCode:
Fixes #18607 

So previously i followed the frontend approach which led to architectural mismatch.

- We already have the correct things implemented in the `processViewNameWithTemplate` and in object metadata service. Found that the seeder files had the hardcoded names instead of {objectLabelPlural}. So changed all the hardcoded string to contain {objectLabelPlural} to seed the new workspaces accurately.

- it will have a follow up PR for typeORM migration to migrate the existing workspaces

https://github.com/user-attachments/assets/3b460f9f-c59d-49d1-8baa-17322d696ecc

I hope i am correct this time :)


charlesBochet (approved):

charlesBochet:
This seems the write call, thank you!

PR #18701, opened 2026-03-17
[Chore]: Migration to dynamic view names for standard views

carbonFibreCode:
Follow up for #18700


charlesBochet:
@carbonFibreCode Please follow upgrade command pattern, we never do data migration in type orm migrations (if it fails / it's slow, it could block another important migration)

charlesBochet (changes requested):
See comment

carbonFibreCode:
@charlesBochet please check the command pattern migration commands

charlesBochet (approved):

charlesBochet:
Thank you @carbonFibreCode :)

PR #18706, opened 2026-03-17
fix(auth): use dynamic SSE headers and add token renewal retry logic

FelixMalfait:
## Summary

- **Dynamic SSE headers**: The `graphql-sse` client was created with a static `Authorization` header captured at creation time. When the access token refreshed, the SSE client kept using the expired token on every reconnection attempt, causing up to 10 wasted retries before the client was disposed and recreated. Now `headers` is passed as a function that reads the latest token from the Jotai store on each connection attempt.

- **Token renewal retry with error classification**: `handleTokenRenewal` previously had zero retry tolerance — any failure during `renewToken` (including transient network errors, server 500s, or timeouts) triggered an immediate full logout via `onUnauthenticatedError()`. Now the renewal retries up to 3 times with linear backoff for transient errors. Only explicit server rejections (`CombinedGraphQLErrors`, e.g. expired/revoked refresh token) skip retries and proceed to logout immediately.

- **Preserved error types in `renewTokenMutation`**: The old code caught all errors and re-threw them as a generic `new Error('Something went wrong...')`, destroying the original error type. Callers couldn't distinguish a GraphQL auth rejection from a network failure. Now errors propagate with their original type.

- **Simplified SSE retry handler**: With dynamic headers handling token freshness automatically, the retry handler no longer needs the `initialTokenForSseClient` comparison to detect token mismatches. It now only resets the SSE client when the user has logged out (no token) or after 10+ consecutive failures.

## Test plan

- [ ] Log in, wait for access token to expire (~30 min or configure shorter expiry), verify no unexpected logout occurs
- [ ] Simulate transient network failure during token renewal (e.g. throttle network in devtools), verify the retry logic recovers without logging out
- [ ] Verify SSE real-time updates continue working after a token refresh
- [ ] Verify genuine logout still works when refresh token is actually invalid/expired
- [ ] Open multiple browser tabs, verify token refresh works correctly across all tabs without "suspicious activity" revocation

### DIFF
--- a/packages/twenty-front/src/modules/apollo/services/apollo.factory.ts
+++ b/packages/twenty-front/src/modules/apollo/services/apollo.factory.ts
@@ -43,6 +43,9 @@ const logger = loggerLink(() => 'Twenty');
 // deduplicate into a single renewal request.
 let renewalPromise: Promise<void> | null = null;
 
+const TOKEN_RENEWAL_MAX_RETRIES = 3;
+const TOKEN_RENEWAL_RETRY_DELAY_MS = 1000;
+
 export interface Options {
   uri: string;
   cache: ApolloClient.Options['cache'];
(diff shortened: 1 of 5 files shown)


FelixMalfait:

Your preview environment is available at: http://bore.pub:43556

lucasbordeau (commented):
LGTM for the SSE part, this is indeed cleaner than before.

packages/twenty-front/src/modules/apollo/services/apollo.factory.ts
```diff
@@ -156,27 +159,57 @@ export class ApolloFactory implements ApolloManager {
         },
       });
 
+      const attemptTokenRenewal = async (): Promise<void> => {
```
thomtrp, line 163: Logic looks fine but seems a bit hacky tbh. We should either use an existing util either find the right abstraction for this

packages/twenty-front/src/modules/auth/services/AuthService.ts
```diff
@@ -22,34 +22,30 @@ const renewTokenMutation = async (
...
-      },
-      fetchPolicy: 'network-only',
-    });
-    data = result.data;
-  } catch {
-    throw new Error('Something went wrong during token renewal');
-  }
+  // Let errors propagate with their original type so callers can
+  // distinguish GraphQL rejections (e.g. expired refresh token)
+  // from transient network failures.
```
thomtrp: Please remove the comment. It explains the change but not the code

packages/twenty-front/src/modules/sse-db-event/components/SSEClientEffect.tsx
```diff
@@ -33,19 +33,22 @@ export const SSEClientEffect = () => {
...
     if (hasAccessTokenPair && !isDefined(sseClient) && isDefined(tokenPair)) {
-      const token = tokenPair?.accessOrWorkspaceAgnosticToken?.token;
-
       const newSseClient = createClient({
         url: `${REACT_APP_SERVER_BASE_URL}/metadata`,
-        headers: {
-          Authorization: token ? `Bearer ${token}` : '',
+        headers: () => {
+          const currentTokenPair = store.get(tokenPairState.atom);
+          const token = currentTokenPair?.accessOrWorkspaceAgnosticToken?.token;
```
thomtrp, line 40: 👍

packages/twenty-front/src/modules/sse-db-event/hooks/useHandleSseClientConnectionRetry.ts
```diff
@@ -27,10 +27,11 @@ export const useHandleSseClientConnectionRetry = () => {
       const tokenPair = store.get(tokenPairState.atom);
       const currentAppToken = tokenPair?.accessOrWorkspaceAgnosticToken?.token;
 
+      // With dynamic headers, the SSE client automatically picks up the
+      // latest token on each reconnection attempt. We only need to reset
+      // when the user has logged out or after too many consecutive failures.
```
thomtrp: same, please remove

PR #18712, opened 2026-03-17
fix: add Windows platform compatibility for local development

Piyush0049:
## Summary
Fixes cross-platform compatibility issues that prevent Twenty from building and running on Windows environments.

Closes #18470

### 1. Fix asset path resolution on Windows ([assets-path.ts](cci:7://file:///d:/rocket_chat/twenty/packages/twenty-server/src/constants/assets-path.ts:0:0-0:0))
The `IS_BUILT_THROUGH_TESTING_MODULE` check only looked for `/dist/` (Unix paths). On Windows, `__dirname` uses backslashes (`D:\...\dist\...`), so the check always returned `true`, causing ENOENT errors during signup when the server couldn't find seed dependency files.

### 2. Fix SDK build externalization on Windows ([vite.config.sdk.ts](cci:7://file:///d:/rocket_chat/twenty/packages/twenty-sdk/vite.config.sdk.ts:0:0-0:0))
The [isExternal](cci:1://file:///d:/rocket_chat/twenty/packages/twenty-sdk/vite.config.sdk.ts:4:0-24:2) function didn't account for:
- Windows absolute paths (e.g., `D:\...`) being passed as module IDs by Rollup
- Workspace packages (`twenty-shared`, `twenty-ui`) needing to be bundled

This caused "Entry module cannot be external" build failures.

### 3. Use cross-env for NODE_ENV in server commands (`project.json`)
PowerShell/CMD doesn't support `VAR=value command` inline syntax. Using `cross-env` makes environment variable setting cross-platform.

## Testing
- [x] Backend builds and starts successfully on Windows
- [x] Frontend builds and starts successfully on Windows
- [x] User signup works without ENOENT errors
- [x] Health check returns 200 OK at localhost:3000/healthz
- [x] Frontend loads correctly at localhost:3001

### DIFF
--- a/package.json
+++ b/package.json
@@ -138,6 +138,7 @@
     "@yarnpkg/types": "^4.0.0",
     "chromatic": "^6.18.0",
     "concurrently": "^8.2.2",
+    "cross-env": "^10.1.0",
     "danger": "^13.0.4",
     "dotenv-cli": "^7.4.4",
     "esbuild": "^0.25.10",
--- a/packages/twenty-sdk/vite.config.sdk.ts
+++ b/packages/twenty-sdk/vite.config.sdk.ts
@@ -11,6 +11,16 @@ const isExternal = (id: string): boolean => {
     return false;
   }
 
+  // Workspace packages must be bundled, not externalized
+  if (id.startsWith('twenty-shared') || id.startsWith('twenty-ui')) {
+    return false;
+  }
+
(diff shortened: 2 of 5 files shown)


FelixMalfait:
Thank you! We are looking to stay focused and reduce bugs so we don't want to extend to more platforms for now, we prefer to stay focused :)

Piyush0049:
Understood, thanks for the quick response! Happy to contribute in other areas. 🙂

PR #18714, opened 2026-03-17
fix: preserve URL encoding in lowercaseUrlOriginAndRemoveTrailingSlash

Piyush0049:
Closes #18698

## Summary
Fixes a URL encoding bug that breaks Google Maps links and other URLs containing percent-encoded characters (e.g., `%2F`, `%20`, `%2520`).

## Root Cause
[lowercaseUrlOriginAndRemoveTrailingSlash](cci:1://file:///d:/rocket_chat/twenty/packages/twenty-shared/src/utils/url/lowercaseUrlOriginAndRemoveTrailingSlash.ts:3:0-29:2) was using [safeDecodeURIComponent()](cci:1://file:///d:/rocket_chat/twenty/packages/twenty-shared/src/utils/url/safeDecodeURIComponent.ts:0:0-6:2) on `url.pathname` and `url.search`, which decoded `%2F` → `/`, `%20` → ` `, etc. This changed the meaning of URLs like Google Maps links where encoding is significant.

## Fix
Extract the path, query, and hash directly from the raw URL string instead of using the URL API's parsed (and auto-decoded) components. This preserves all percent-encoding exactly as the user provided it.

## Testing
- All 14 existing + updated unit tests pass
- Updated test expectations to verify encoding is preserved, not decoded

### DIFF
--- a/packages/twenty-shared/src/utils/url/__tests__/lowercaseUrlOriginAndRemoveTrailingSlash.test.ts
+++ b/packages/twenty-shared/src/utils/url/__tests__/lowercaseUrlOriginAndRemoveTrailingSlash.test.ts
@@ -44,9 +44,9 @@ describe('lowercaseUrlOriginAndRemoveTrailingSlash', () => {
       expected: 'https://test.test/edouard-ménard-22219837',
     },
     {
-      title: 'should decode already encoded special characters in path',
+      title: 'should keep already encoded special characters in path',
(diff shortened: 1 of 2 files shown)


charlesBochet:
Hi @Piyush0049, thanks for opening a PR, unfortunately this one is not meeting our quality standards, I'm going to close it. The interest for the project is much appreciated tough :)

PR #18716, opened 2026-03-17
fix: make 1.19 upgrade commands resilient to pre-existing data

charlesBochet:
## Summary

- **BackfillMissingStandardViewsCommand**: When view validation fails (e.g. a viewField references a field metadata that doesn't exist in the workspace), log a warning and skip instead of throwing — so the workspace upgrade continues with the remaining commands.
- **AddMissingSystemFieldsToStandardObjectsCommand**: Wrap both the non-tsVector batch migration and each individual tsVector migration in try-catch. If a field already exists (e.g. duplicate key on `name + objectMetadataId + workspaceId`), the error is logged as a warning and the command moves on to the next field.

These errors were observed during the 1.18 → 1.19 production upgrade for workspaces with non-standard state (missing "owner" field metadata on Opportunity, or searchVector fields already present with a different universalIdentifier).

## Test plan

- [ ] Re-run upgrade on the affected production workspaces
- [ ] Verify upgrade completes successfully with warnings instead of failures
- [ ] Confirm that workspaces which were already upgrading cleanly are unaffected


FelixMalfait:

Your preview environment is available at: http://bore.pub:43416

PR #18717, opened 2026-03-17
fix: route object color through standardOverrides for standard objects

charlesBochet:
## Summary

- Fixes object `color` to use the `standardOverrides` mechanism for standard objects, matching how `label`, `description`, and `icon` already work
- Previously, color was written directly to the `objectMetadata.color` column for **both** standard and custom objects, which meant user color customizations on standard objects could be overwritten during metadata syncs
- Custom objects continue to have `color` updated directly on the entity (no change)

## Changes

| File | What changed |
|------|-------------|
| `object-metadata-standard-overrides-properties.constant.ts` | Added `'color'` to `OBJECT_METADATA_STANDARD_OVERRIDES_PROPERTIES` so `sanitizeRawUpdateObjectInput` routes color into `standardOverrides` for standard objects |
| `resolve-object-metadata-standard-override.util.ts` | Extended to support `'color'` as a key — handled like `icon` (no i18n/translation, just direct override check) |
| `object-metadata.resolver.ts` | Added `@ResolveField` for `color` that resolves through `resolveObjectMetadataStandardOverride`, matching the existing `labelSingular`/`labelPlural`/`description`/`icon` resolve fields |
| `flat-object-metadata-validator.service.ts` | Removed `'color'` from `allowedOverrideKeys` for system objects since it now flows through `standardOverrides` |
| `resolve-object-metadata-standard-override.util.spec.ts` | Added test cases for custom object color, standard object color override, and standard object color fallback |
| `successful-update-one-standard-object-metadata.integration-spec.ts` | Added `'when updating color'` test case, included `color` in GraphQL queries and `standardOverrides` fragment, reset color in `afterEach` cleanup |

## How it works now

| Object type | Color update flow |
|---|---|
| **Custom** | Written directly to `objectMetadata.color` column |
| **Standard** | Stored in `objectMetadata.standardOverrides.color`, resolved via `@ResolveField` at query time |

This is identical to how `label`, `description`, and `icon` have always worked.

## Test plan

- [x] Unit tests pass (`resolve-object-metadata-standard-override.util.spec.ts` — 21 tests)
- [x] Typecheck passes (`npx nx typecheck twenty-server`)
- [x] Lint passes (`npx nx lint:diff-with-main twenty-server`)
- [ ] Integration test snapshot regenerates correctly (`successful-update-one-standard-object-metadata`)
- [ ] Verify standard object color editing from sidebar persists via `standardOverrides`
- [ ] Verify custom object color editing from sidebar persists directly on entity


FelixMalfait:

Your preview environment is available at: http://bore.pub:56798

packages/twenty-server/src/engine/workspace-event-emitter/workspace-event-emitter.service.ts
```diff
@@ -308,6 +313,65 @@ export class WorkspaceEventEmitterService {
     return { ...metadataEventBatch, events: enrichedEvents };
   }
 
+  private resolveObjectMetadataStandardOverrides(
```
thomtrp, line 316: we should have a separated resolve service that also take care of relations, and label in future. Will do in a follow-up PR

thomtrp (approved):

PR #18726, opened 2026-03-18
Treat Microsoft Graph 400 with empty body as transient error

neo773:
Graph SDK occasionally returns a 400 with a null error message which is not a real bad request but a transient hiccup. 
Classify these as temporary errors so they get retried instead of flooding Sentry.

Fixes TWENTY-SERVER-D3X


charlesBochet (approved):

FelixMalfait:

Your preview environment is available at: http://bore.pub:45573

PR #18730, opened 2026-03-18
fix: navigation menu item type backfill and frontend loading

charlesBochet:
## Summary

- Move `BackfillNavigationMenuItemTypeCommand` from the 1-19 to the 1-20 upgrade path and split the DB transaction into two phases (data backfill, then schema changes) to avoid the PostgreSQL error "cannot ALTER TABLE because it has pending trigger events."
- Fix backfill logic to prefer `OBJECT` over `VIEW` for navigation menu items that have `targetObjectMetadataId`, and correct already mis-typed items. Tighten the `CHECK` constraint to enforce `viewId IS NULL` for `OBJECT` type items.
- On the frontend, force `navigationMenuItems` into `staleEntityKeys` when the server's `minimalMetadata` response omits the collection hash (happens when the Redis cache hasn't been warmed after an upgrade), ensuring the sidebar loads navigation items.

## Test plan

- [ ] Upgrade from 1.18 or 1.19 to 1.20 and verify the migration completes without errors
- [ ] Verify navigation menu items of type `OBJECT` do not have a `viewId` set in the database
- [ ] Sign out and sign in — confirm navigation menu items appear in the sidebar on first load
- [ ] Verify `VIEW`-typed items also appear correctly in the sidebar


FelixMalfait:

Your preview environment is available at: http://bore.pub:62011

etiennejouan (approved):
LGTM

PR #18732, opened 2026-03-18
Add progress tracking to command menu items

### LINKED ISSUE #2309: [COMMAND MENU ITEMS] Command progress api
Add a way for a command to communicate it's progress with the app

bosiraphael:
Introduces a progress indicator for command menu items (both engine commands and front components). When a command is running, the menu item now displays a percentage alongside the loader spinner instead of just a spinner.

- Added `commandMenuItemProgressFamilyState` to track per-item progress and `CommandListItemLoader` to render it
- Exposed `updateProgress` in the twenty-sdk public API so front components can report execution progress back to the host
- Wired progress reporting into `ExportMultipleRecordsCommand` as the first consumer, showing CSV export progress
- Progress state is cleaned up on unmount for both engine commands and headless front components
- Refactor


packages/twenty-front/src/modules/command-menu-item/engine-command/record/multiple-records/components/ExportMultipleRecordsCommand.tsx
```diff
@@ -1,24 +1,75 @@
...
+import { commandMenuItemProgressFamilyState } from '@/command-menu-item/states/commandMenuItemProgressFamilyState';
+import { type ObjectMetadataItem } from '@/object-metadata/types/ObjectMetadataItem';
 import { useRecordIndexExportRecords } from '@/object-record/record-index/export/hooks/useRecordIndexExportRecords';
+import { useAvailableComponentInstanceIdOrThrow } from '@/ui/utilities/state/component-state/hooks/useAvailableComponentInstanceIdOrThrow';
+import { useSetAtomFamilyState } from '@/ui/utilities/state/jotai/hooks/useSetAtomFamilyState';
+import { ViewComponentInstanceContext } from '@/views/states/contexts/ViewComponentInstanceContext';
+import { useEffect } from 'react';
 import { isDefined } from 'twenty-shared/utils';
 
+const ExportMultipleRecordsCommandContent = ({
```
Devessier, line 13: Usually, don't we declare only one component per file?

packages/twenty-front/src/modules/front-components/hooks/useFrontComponentExecutionContext.ts
```diff
@@ -133,6 +141,15 @@ export const useFrontComponentExecutionContext = ({
       closeSidePanelMenu();
     };
 
+  const updateProgress: FrontComponentHostCommunicationApi['updateProgress'] =
+    async (progress) => {
+      if (!isDefined(commandMenuItemId)) {
+        return;
+      }
+
+      setCommandMenuItemProgress(Math.max(0, Math.min(100, progress)));
```
Devessier, line 150: Why Math.min/max? Isn't it guaranteed that the progress will be between 0 and 100?

Devessier (approved):
Looks good to me from a technical POV!

I left two comments. Let's check them before merging.

packages/twenty-front/src/modules/command-menu-item/engine-command/record/multiple-records/components/ExportMultipleRecordsCommand.tsx
bosiraphael: We only export one, but here it is acceptable I think since it is not exported, it is just to be able to cut the tree before the hooks usage

Devessier:
❤️ 😍 

https://github.com/user-attachments/assets/654d158a-4d61-4041-a846-971351f2351a

packages/twenty-front/src/modules/front-components/hooks/useFrontComponentExecutionContext.ts
bosiraphael: The caller in the sdk can call it with whatever number they want, so we prevent this in the API we expose

packages/twenty-front/src/modules/command-menu-item/engine-command/record/multiple-records/components/ExportMultipleRecordsCommand.tsx
Devessier: I approve the practice, and I would've liked to do it myself a few dozens times 😄 LGTM!

PR #18734, opened 2026-03-18
Scaffold light twenty app dev container

martmull:
as title


packages/create-twenty-app/src/constants/base-application/README.md
```diff
@@ -2,7 +2,7 @@ This is a [Twenty](https://twenty.com) application project bootstrapped with [`c
 
 ## Getting Started
 
-Start development mode — it auto-connects to your local Twenty server at localhost:3000:
+Start development mode — it auto-connects to your local Twenty server at localhost:2020:
```
martmull: remove port

packages/create-twenty-app/src/utils/setup-local-instance.ts
```diff
@@ -48,147 +46,216 @@ const isTwentyServerRunning = async (): Promise<boolean> => {
...
-      { encoding: 'utf-8' },
-    );
+const streamDockerLogs = (): ChildProcess => {
+  const logProcess = spawn(
+    'docker',
+    ['logs', '-f', '--tail', '0', CONTAINER_NAME],
+    { stdio: ['ignore', 'pipe', 'pipe'] },
+  );
+
+  const formatLogLine = (line: string): string | null => {
```
FelixMalfait: A bit strange?

packages/create-twenty-app/src/utils/setup-local-instance.ts
```diff
@@ -48,147 +46,216 @@ const isTwentyServerRunning = async (): Promise<boolean> => {
...
+  console.log(
+    chalk.gray('Workspace ready — login with tim@apple.dev / tim@apple.dev'),
+  );
 
-    return { running: true };
-  }
+  const openCommand =
+    platform() === 'darwin'
+      ? 'open'
+      : platform() === 'win32'
```
FelixMalfait: Not sure we want to support windows tbh, I think it will fail in any case? I know we've already merged similar code

packages/create-twenty-app/src/utils/setup-local-instance.ts
```diff
@@ -48,147 +46,216 @@ const isTwentyServerRunning = async (): Promise<boolean> => {
...
     }
 
+    console.log(chalk.gray('Starting Twenty container...'));
+
     try {
-      execSync(`bash <(curl -sL ${INSTALL_SCRIPT_URL})`, {
-        stdio: 'inherit',
-        shell: '/bin/bash',
-      });
+      execSync(
```
FelixMalfait: I didn't expect to see code related to the docker image in create-twenty-app, do we really need it there? Or could we concentrate all the logic in one place / twenty-sdk?

martmull: yes, indeed, we could call twenty-sdk: server command

packages/create-twenty-app/src/utils/setup-local-instance.ts
```diff
@@ -48,147 +51,210 @@ const isTwentyServerRunning = async (): Promise<boolean> => {
   }
 };
```
FelixMalfait, line 25: Code duplication between this and server ==> imo it confirms intuition that this might not need to exist?

packages/twenty-server/src/engine/workspace-manager/workspace-manager.module.ts
```diff
@@ -22,6 +22,7 @@ import { TwentyStandardApplicationModule } from 'src/engine/workspace-manager/tw
 import { WorkspaceMigrationModule } from 'src/engine/workspace-manager/workspace-migration/workspace-migration.module';
 
 import { WorkspaceManagerService } from './workspace-manager.service';
+import { ApplicationRegistrationModule } from 'src/engine/core-modules/application/application-registration/application-registration.module';
```
FelixMalfait: I'm surprised this isn't catched by the linter? Maybe it is?

packages/twenty-sdk/src/cli/commands/server.ts
```diff
@@ -0,0 +1,216 @@
...
+      const healthy = running ? await isServerHealthy() : false;
+
+      const statusText = healthy
+        ? chalk.green('running (healthy)')
+        : running
+          ? chalk.yellow('running (starting...)')
+          : chalk.gray('stopped');
+
+      console.log(`  Status:  ${statusText}`);
+      console.log(`  URL:     http://localhost:2020`);
```
FelixMalfait: Could be another port set by user I guess?

packages/twenty-sdk/src/cli/commands/remote.ts
```diff
@@ -6,6 +6,33 @@ import chalk from 'chalk';
 import type { Command } from 'commander';
 import inquirer from 'inquirer';
 
+const LOCAL_PORTS = [2020, 3000];
+
+const detectLocalServer = async (): Promise<string | null> => {
```
FelixMalfait: checkHealthz / isServerHealthy ; redundancy?

packages/twenty-docker/twenty-app-dev/rootfs/etc/s6-overlay/scripts/init-db.sh
```diff
@@ -0,0 +1,40 @@
...
+  echo "Database appears to be empty, running initial setup..."
+  NODE_OPTIONS="--max-old-space-size=1500" tsx ./scripts/setup-db.ts
+fi
+
+# Always run migrations (idempotent — skips already-applied ones)
+yarn database:migrate:prod
+
+yarn command:prod cache:flush
+yarn command:prod upgrade
+yarn command:prod cache:flush
```
FelixMalfait, line 43: It the duplicated yarn command:prod cache:flush on purpose?

packages/twenty-docker/twenty-app-dev/Dockerfile
```diff
@@ -0,0 +1,121 @@
...
+
+# Tell s6-overlay to preserve container environment variables
+ENV S6_KEEP_ENV=1
+
+# Environment defaults
+ENV PG_DATABASE_URL=postgres://twenty:twenty@localhost:5432/default \
+    SERVER_URL=http://localhost:2020 \
+    REDIS_URL=redis://localhost:6379 \
+    STORAGE_TYPE=local \
+    APP_SECRET=change-me-in-production \
```
FelixMalfait: Generate automatically maybe?  We could argue it's not needed dev but I guess it doesn't cost much with AI.... In any case "change-me-in-production" feels odd since it's not made for prod

packages/twenty-server/src/engine/workspace-manager/dev-seeder/data/services/dev-seeder-data.service.ts
```diff
@@ -120,6 +120,15 @@ import { TimelineActivitySeederService } from 'src/engine/workspace-manager/dev-
 import { prefillWorkflows } from 'src/engine/workspace-manager/standard-objects-prefill-data/prefill-workflows';
 import { TWENTY_STANDARD_APPLICATION } from 'src/engine/workspace-manager/twenty-standard-application/constants/twenty-standard-applications';
 
+const LIGHT_MAX_RECORDS = 5;
+
+const LIGHT_EXCLUDED_TABLES = new Set([
```
FelixMalfait: Not sure I understand why there's an underscore here and not in LIGHT_EXCLUDED_OBJECTS ; should it be a shared const?

packages/create-twenty-app/src/utils/setup-local-instance.ts
FelixMalfait: Feels like a code smell / AI pattern we should get rid of

packages/twenty-docs/developers/extend/capabilities/apps.mdx
```diff
@@ -19,20 +19,52 @@ Apps let you build and manage Twenty customizations **as code**. Instead of conf
 
 ## Prerequisites
 
-- Node.js 24+ and Yarn 4
-- A Twenty workspace and an API key (create one at https://app.twenty.com/settings/api-webhooks)
+- Node.js 22+ and Yarn 4
```
FelixMalfait: Why not 24? let's align everything

packages/twenty-server/src/database/commands/data-seed-dev-workspace.command.ts
```diff
@@ -1,32 +1,61 @@
...
-    SEED_YCOMBINATOR_WORKSPACE_ID,
-  ] as const;
   private readonly logger = new Logger(DataSeedWorkspaceCommand.name);
 
   constructor(private readonly devSeederService: DevSeederService) {
     super();
   }
 
-  async run(): Promise<void> {
+  @Option({
```
FelixMalfait: These two flags feel either too granular or not granular enough :p

I think I would have prefered either 2 clear seed path or making seeds modular. But this feels like arbitrary modularity. I'm fine with that but maybe you can explore quickly with Claude if something a bit more elegant can be done easily?

martmull: I think i will just use light

packages/twenty-docker/twenty-app-dev/rootfs/etc/s6-overlay/scripts/init-db.sh
martmull: yes, done on regular image

packages/twenty-docker/twenty-app-dev/Dockerfile
```diff
@@ -0,0 +1,130 @@
+ARG APP_VERSION
```
charlesBochet, line 1: Here I would recommend actually taking dependency on a published version of the app to speed builds. Instead of starting from node:22-alpine we could start from twenty-crm/twenty@APP_VERSION

packages/twenty-sdk/src/cli/commands/server.ts
```diff
@@ -0,0 +1,243 @@
...
+    console.error(chalk.red('Invalid port number.'));
+    process.exit(1);
+  }
+
+  return port;
+};
+
+export const registerServerCommands = (program: Command): void => {
+  const server = program
+    .command('server')
```
charlesBochet, line 61: is server the right naming? I would expect instance or local-instance

packages/twenty-sdk/src/cli/utilities/server/detect-local-server.ts
```diff
@@ -0,0 +1,30 @@
...
+    return body.status === 'ok';
+  } catch {
+    return false;
+  } finally {
+    clearTimeout(timeoutId);
+  }
+};
+
+export const detectLocalServer = async (): Promise<string | null> => {
+  for (const port of LOCAL_PORTS) {
```
charlesBochet, line 23: I'm a bit lost, why 3000?

packages/twenty-server/src/engine/workspace-manager/dev-seeder/metadata/services/dev-seeder-metadata.service.ts
```diff
@@ -193,6 +191,20 @@ export class DevSeederMetadataService {
       );
     }
 
+    return light ? this.getLightConfig(config) : config;
```
charlesBochet, line 194: all this getConfig, getLightConfig, workspaceSeedConfig is convoluted for no reason :p

packages/twenty-server/src/engine/workspace-manager/dev-seeder/services/dev-seeder.service.ts
```diff
@@ -89,16 +93,19 @@ export class DevSeederService {
     await this.devSeederMetadataService.seed({
       dataSourceMetadata,
       workspaceId,
+      light,
```
charlesBochet, line 96: bad naming light, I don't know what it means from the external

Also, I don't get why we are not using the prefill here (like when we create a new workspace). I prefer maintaining only 2 (seed / prefill) data-seeds than 3

packages/twenty-sdk/src/cli/commands/server.ts
FelixMalfait: I don't love local-instance (commands needs to be short / no dash). And instance isn't very clear either (could be an object instance etc).

@claude could you please suggest names? it's basically a command like Laravel Sail that provides a local development server to work against. Currently its yarn twenty server up, yarn twenty server down etc

FelixMalfait: @claude yes I like dev too but we already use it for the script that does hot reload to sync an app with that local server, so it would conflict. How would you name it then? Can you make more suggestions? We could also move the current dev command to become something like twenty dev sync but that's hard to remember and probably confusing to put it under a different umbrella while it's a slightly different nature? Take inspirations from the best DX on this market and make recommandations

FelixMalfait: I like the recommandation to flatten things. We can still group them visually in the help command in the future like void zero does it

PR #18667, opened 2026-03-16
refactor: remove "core" naming from views and eliminate converter layer

charlesBochet:
## Summary

- **Remove all "core" prefixes** from the views system — the metadata-based storage migration is complete, so `CoreView`, `coreViewsSelector`, `getCoreViews`, etc. are now just `View`, `viewsSelector`, `getViews`
- **Eliminate the entire converter layer** (15 files, ~850 lines deleted) — `convertCoreViewToView` and all sub-converters were either no-ops or trivially adding `__typename` / mapping identical enum values. Local enums now re-export from generated GraphQL types directly (single source of truth)
- **Unify `View` and `ViewWithRelations`** into one type — `ViewWithRelations` is now a type alias for `View`, selectors return data directly without conversion

### Backend
- Rename `@ObjectType('CoreView')` → `@ObjectType('View')` (and all sub-entities)
- Rename resolver methods: `getCoreViews` → `getViews`, `createCoreView` → `createView`, etc.
- Rename `FIND_ALL_CORE_VIEWS_GRAPHQL_OPERATION` → `FIND_ALL_VIEWS_GRAPHQL_OPERATION`

### Frontend
- Delete 15 converter files (`convertGqlView*ToView*`, `convertView*ToGql`, `convertViewWithRelationsToView`)
- Re-export `ViewType`, `ViewKey`, `ViewFilterGroupLogicalOperator` from generated enums (no more duplicate enum definitions with different casing)
- Replace `ViewOpenRecordInType` with `ViewOpenRecordIn` from generated
- Remove `__typename` from all local view sub-types
- Remove unused `variant` from `ViewFilter`, make `displayValue` and `definition` optional
- Rename ~45 GraphQL query/mutation files and all selectors to drop "core" prefix
- Delete unused `viewsWithRelationsSelector`

PR #18745, opened 2026-03-18
fix: restore ViewFilter value stringification lost in converter removal

charlesBochet:
## Summary

- **Restores `convertViewFilterValueToString()` calls** that were lost when the converter layer was removed in #18667. The GraphQL `ViewFilter.value` is typed as `JSON` (can be a string, array, or object), but the frontend type system expects a `string`. Without stringification, SELECT/MULTI_SELECT filter values (e.g. `['LOST']`) reach `arrayOfStringsOrVariablesSchema` as raw arrays, causing a `ZodError: expected string, received array`.
- **Fixes applied at two data boundary points**: `splitViewWithRelated` (primary entry from metadata store) and `mapViewFiltersToFilters` (which also accepts `GqlViewFilter[]` directly).

Fixes a production regression introduced by #18667.

## Test plan

- [ ] Apply a SELECT filter (e.g. filter Opportunities by Stage = "Lost") — should no longer throw ZodError
- [ ] Apply a MULTI_SELECT filter — should work correctly
- [ ] Verify filters with multiple selected values work (e.g. Stage is "Lost" or "Won")
- [ ] Verify empty filters and "is not" operands still work
- [ ] Verify filters loaded from saved views still work after page refresh


FelixMalfait:

Your preview environment is available at: http://bore.pub:11987

PR #18747, opened 2026-03-18
Add record table widget to dashboards

lucasbordeau:
## Demo

https://github.com/user-attachments/assets/584de452-544a-41f8-ae9f-4be9e9d0cd9f

## Problem
- Dashboards only supported chart widgets — tabular record data had no inline widget type
- `RecordTable` was tightly coupled to the record index: HTML IDs, CSS variables, and hover portals were global strings with no per-instance scoping, so multiple tables on the same page would collide
- `updateRecordTableCSSVariable`, `RECORD_TABLE_HTML_ID`, and cell portal IDs were hardcoded — placing two tables caused hover portals and CSS column widths to bleed across instances
- Grid drag-select captured record UUIDs as cell IDs, producing `NaN` layout coordinates and a full-page freeze on second widget creation

## Fix
- `RECORD_TABLE` is now a valid widget type across the full stack — server DTOs, DB enum migration, universal config mapping, GraphQL codegen, shared types (`RecordTableConfigurationDto`, `WidgetType`, `addRecordTableWidgetType` migration)
- A record table widget can be placed on a dashboard and boots from a View ID with no record index dependency — `StandaloneRecordTableProvider` + `StandaloneRecordTableViewLoadEffect` (wraps existing `RecordTableWithWrappers` unchanged)
- Selecting a data source auto-creates a dedicated View with up to 6 initial fields; switching source or deleting the widget cleans up the View — `useCreateViewForRecordTableWidget` + `useDeleteViewForRecordTableWidget`
- The settings panel exposes source, field visibility/reorder, filter conditions, sort rules, and editable widget title — `SidePanelPageLayoutRecordTableSettings` + sub-pages, matching chart widget pattern
- Filters, sorts, and aggregate operations update the table in real time but only persist to the View on explicit dashboard save — `useSaveRecordTableWidgetsViewDataOnDashboardSave` (diff + flush on save)
- Headers are always non-interactive (no dropdown, no cursor pointer); columns are resizable only in edit mode; cells are non-editable in both modes — `isRecordTableColumnHeadersReadOnlyComponentState`, `isRecordTableColumnResizableComponentState`, `isRecordTableCellsNonEditableComponentState` (Jotai component states)
- Hover portals and CSS column widths no longer bleed between multiple table widgets — `getRecordTableHtmlId(tableId)`, `getRecordTableCellId(tableId, …)`, `updateRecordTableCSSVariable(tableId, …)` scope all DOM IDs and CSS variables per instance
- Clicking inside a widget's content area no longer opens the settings panel — `WidgetCardContent` stops click propagation when editable, limiting settings-open to the card header and chrome
- Second widget creation no longer freezes the page — `PageLayoutGridLayout` drag-select filters by `cell-` prefix to exclude record UUIDs from grid cell detection

## Follow-up fixes

**Widget save flow**
- Saving a dashboard silently dropped record table widget changes (column visibility, order, filters, sorts, aggregates) because widget data save was bundled inside the layout save and only ran when layout structure changed
- Widget data now persists independently via `useSavePageLayoutWidgetsData`, called in all save paths (dashboard save, record page save, layout customization save); saves are also skipped when nothing has changed

**Drag-and-drop / checkbox columns in widget**
- Record table widgets showed the drag handle column and checkbox selection column even though row reordering and multi-select are meaningless in a read-only widget
- Two new component states (`isRecordTableDragColumnHiddenComponentState`, `isRecordTableCheckboxColumnHiddenComponentState`) hide each column independently; widget tables now display only data columns

**Sticky column layout**
- Sticky positioning of the first three columns used `:nth-of-type` CSS selectors — when drag or checkbox columns were hidden, the selector targeted the wrong column and the first data column didn't stick
- Sticky CSS now targets semantic class names (`RECORD_TABLE_COLUMN_DRAG_AND_DROP_WIDTH_CLASS_NAME`, etc.) so sticky behavior is correct regardless of which columns are hidden

**Save/Cancel buttons during edit mode**
- Save and Cancel command-menu buttons were unpinned during dashboard edit mode because the pin logic excluded all items while `isPageInEditMode` was true
- Items whose availability expression contains `isPageInEditMode` are now exempted from the unpin rule; Save/Cancel stay pinned during editing

**Title input auto-focus**
- Selecting "Record Table" as widget type auto-focused the title input, interrupting the configuration flow
- `focusTitleInput` is now `false` when navigating to record table settings

**Morph relation field error**
- A field with missing `morphRelations` metadata crashed the page with a "refresh" error from `mapObjectMetadataToGraphQLQuery`
- Now returns an empty array and silently omits the field from the query instead of crashing

**`updateRecordMutation` prop removal**
- `RecordTableWithWrappers` required callers to pass an `updateRecordMutation` callback, duplicating `useUpdateOneRecord` at every usage site
- The mutation is now owned inside `RecordTableContextProvider` via `RecordTableUpdateContext`; the prop is gone

**Standalone → Widget module rename**
- `record-table-standalone` module renamed to `record-table-widget` — `StandaloneRecordTable` → `RecordTableWidget`, `StandaloneRecordTableViewLoadEffect` → `RecordTableWidgetViewLoadEffect`, etc.

**RecordTableRow cell extraction**
- Row rendering logic (`RecordTableCellDragAndDrop`, `RecordTableCellCheckbox`, `RecordTableFieldsCells`, hotkey/arrow-key effects) was duplicated between `RecordTableRow` and `RecordTableRowVirtualizedFullData`
- Extracted `RecordTableRowCells` (shared cell content) and `RecordTableStaticTr` (non-draggable `<tr>` wrapper); when drag column is hidden, rows render inside a static `<tr>` instead of the draggable wrapper

**View load effect metadata tracking**
- `RecordTableWidgetViewLoadEffect` now tracks `objectMetadataItem.updatedAt` alongside `viewId` to re-load states when metadata changes (e.g. field additions), preventing stale column data

**Data source dropdown deduplication**
- Extracted `filterReadableActiveObjectMetadataItems` util, shared by both chart and record table data source dropdowns — removes duplicated permission-filtering logic

**RECORD_TABLE view identifier mapping (server)**
- Added `RECORD_TABLE` case to `fromPageLayoutWidgetConfigurationToUniversalConfiguration` and `fromUniversalConfigurationToFlatPageLayoutWidgetConfiguration` so widget views are properly mapped during workspace import/export

**GraphQL error handler typing (server)**
- `formatError` parameter changed from `any` to `unknown`; `workspaceQueryRunnerGraphqlApiExceptionHandler` broadened from `QueryFailedErrorWithCode` to `Error | QueryFailedError` — removes unsafe type casts

**Save hook signature**
- `useSaveRecordTableWidgetsViewDataOnDashboardSave` no longer takes `pageLayoutId` in constructor; receives it as a callback parameter, eliminating the need for `useAtomComponentStateCallbackState`

**Customize Dashboard hidden during edit mode**
- The "Customize Dashboard" command was still visible while already editing — its `conditionalAvailabilityExpression` now includes `not isPageInEditMode`

**Fields dropdown split**
- `RecordTableFieldsDropdownContent` (300+ lines) split into `RecordTableFieldsDropdownVisibleFieldsContent` and `RecordTableFieldsDropdownHiddenFieldsContent`

**Checkbox placeholder cleanup**
- Removed unnecessary `StyledRecordTableTdContainer` wrapper from `RecordTableCellCheckboxPlaceholder`


packages/twenty-server/src/engine/api/graphql/direct-execution/direct-execution.service.ts
```diff
@@ -280,7 +280,14 @@ export class DirectExecutionService {
     );
   }
 
-  private formatError(error: any, req: Request): GraphQLFormattedError {
+  private formatError(error: unknown, req: Request): GraphQLFormattedError {
+    if (!(error instanceof Error)) {
```
lucasbordeau, line 284: @etiennejouan Are you ok with this fix ?

etiennejouan: Sorry for that. You will have conflict with nitin here https://github.com/twentyhq/twenty/pull/18874#discussion_r2979904256

This logic is gated behind feature flag, I'll fix this later.

packages/twenty-front/src/modules/object-metadata/utils/filterReadableActiveObjectMetadataItems.ts
```diff
@@ -0,0 +1,21 @@
+import { type EnrichedObjectMetadataItem } from '@/object-metadata/types/EnrichedObjectMetadataItem';
+import { type ObjectPermissions } from 'twenty-shared/types';
+import { isDefined } from 'twenty-shared/utils';
+
+export const filterReadableActiveObjectMetadataItems = (
```
charlesBochet, line 5: duplicate with @FelixMalfait PR

packages/twenty-front/src/modules/object-metadata/utils/mapObjectMetadataToGraphQLQuery.ts
```diff
@@ -65,9 +65,7 @@ export const mapObjectMetadataToGraphQLQuery = ({
       }
 
       if (!isDefined(fieldMetadata.morphRelations)) {
-        throw new Error(
-          `Field ${fieldMetadata.name} is missing, please refresh the page. If the problem persists, please contact support.`,
-        );
+        return [];
```
charlesBochet, line 68: revert this should not be needed

packages/twenty-front/src/modules/object-record/record-table-widget/states/lastLoadedRecordTableWidgetViewIdComponentState.ts
```diff
@@ -0,0 +1,12 @@
...
+import { createAtomComponentState } from '@/ui/utilities/state/jotai/utils/createAtomComponentState';
+
+export const lastLoadedRecordTableWidgetViewIdComponentState =
+  createAtomComponentState<{
+    viewId: string;
+    objectMetadataItemUpdatedAt: string;
+  } | null>({
+    key: 'lastLoadedRecordTableWidgetViewIdComponentState',
+    defaultValue: null,
+    componentInstanceContext: ContextStoreComponentInstanceContext,
```
charlesBochet, line 11: should be RecordTableWidgetInstanceContext

packages/twenty-front/src/modules/object-record/record-table/record-table-header/components/RecordTableHeaderEmptyLastColumn.tsx
```diff
@@ -0,0 +1,71 @@
...
+    shouldDisplayBorderBottom
+      ? `1px solid ${themeCssVariables.border.color.light}`
+      : 'none'};
+  height: ${RECORD_TABLE_ROW_HEIGHT}px;
+  max-height: ${RECORD_TABLE_ROW_HEIGHT}px;
+  width: ${RECORD_TABLE_COLUMN_ADD_COLUMN_BUTTON_WIDTH}px;
+  z-index: 1;
+`;
+
+export const RecordTableHeaderEmptyLastColumn = () => {
```
charlesBochet, line 30: is this a refactor?

packages/twenty-front/src/modules/object-record/record-table/record-table-row/components/RecordTableRowCells.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { isRecordTableRowFocusActiveComponentState } from '@/object-record/record-table/states/isRecordTableRowFocusActiveComponentState';
+import { isRecordTableRowFocusedComponentFamilyState } from '@/object-record/record-table/states/isRecordTableRowFocusedComponentFamilyState';
+import { useAtomComponentFamilyStateValue } from '@/ui/utilities/state/jotai/hooks/useAtomComponentFamilyStateValue';
+import { useAtomComponentStateValue } from '@/ui/utilities/state/jotai/hooks/useAtomComponentStateValue';
+
+type RecordTableRowCellsProps = {
+  rowIndexForFocus: number;
+};
+
+export const RecordTableRowCells = ({
```
charlesBochet, line 20: refactor?

packages/twenty-front/src/modules/page-layout/hooks/useCreatePageLayoutRecordTableWidget.ts
```diff
@@ -0,0 +1,120 @@
...
+        minH: minimumSize.h,
+      };
+
+      const updatedLayouts = getUpdatedTabLayouts(
+        allTabLayouts,
+        activeTabId,
+        newLayout,
+      );
+
+      store.set(pageLayoutCurrentLayoutsState, updatedLayouts);
```
charlesBochet, line 101: is this expected?

packages/twenty-front/src/modules/page-layout/widgets/record-table/hooks/useMapRecordFieldToViewFieldWithCurrentAggregateOperation.ts
```diff
@@ -0,0 +1,36 @@
+import { type RecordField } from '@/object-record/record-field/types/RecordField';
+import { viewFieldAggregateOperationState } from '@/object-record/record-table/record-table-footer/states/viewFieldAggregateOperationState';
+import { convertExtendedAggregateOperationToAggregateOperation } from '@/object-record/utils/convertExtendedAggregateOperationToAggregateOperation';
+import { type ViewField } from '@/views/types/ViewField';
+import { mapRecordFieldToViewField } from '@/views/utils/mapRecordFieldToViewField';
+import { useStore } from 'jotai';
+import { useCallback } from 'react';
+import { isDefined } from 'twenty-shared/utils';
+
+export const useMapRecordFieldToViewFieldWithCurrentAggregateOperation = () => {
```
charlesBochet, line 10: was it broken on the existing table?

packages/twenty-front/src/modules/page-layout/widgets/record-table/hooks/useMapViewFieldToRecordTableWidgetViewFieldItem.ts
```diff
@@ -0,0 +1,25 @@
+import { useGetFieldMetadataItemByIdOrThrow } from '@/object-metadata/hooks/useGetFieldMetadataItemById';
+import { type RecordTableWidgetViewFieldItem } from '@/page-layout/widgets/record-table/types/RecordTableWidgetViewFieldItem';
+import { type ViewField } from '@/views/types/ViewField';
+import { useCallback } from 'react';
+
+export const useMapViewFieldToRecordTableWidgetViewFieldItem = () => {
+  const { getFieldMetadataItemByIdOrThrow } =
+    useGetFieldMetadataItemByIdOrThrow();
+
+  const mapViewFieldToRecordTableWidgetViewFieldItem = useCallback(
```
charlesBochet, line 10: looks convoluted

charlesBochet: we should derive

packages/twenty-front/src/modules/page-layout/widgets/record-table/hooks/useRecordTableWidgetViewFieldItems.ts
```diff
@@ -0,0 +1,26 @@
+import { useMapViewFieldToRecordTableWidgetViewFieldItem } from '@/page-layout/widgets/record-table/hooks/useMapViewFieldToRecordTableWidgetViewFieldItem';
+import { type RecordTableWidgetViewFieldItem } from '@/page-layout/widgets/record-table/types/RecordTableWidgetViewFieldItem';
+import { useViewById } from '@/views/hooks/useViewById';
+import { useMemo } from 'react';
+import { isDefined } from 'twenty-shared/utils';
+import { sortByProperty } from '~/utils/array/sortByProperty';
+
+export const useRecordTableWidgetViewFieldItems = (viewId: string) => {
```
charlesBochet, line 8: same

packages/twenty-front/src/modules/page-layout/widgets/record-table/hooks/useSaveRecordTableWidgetsViewDataOnDashboardSave.ts
```diff
@@ -0,0 +1,306 @@
...
+import { getViewSortsToUpdate } from '@/views/utils/getViewSortsToUpdate';
+import { mapRecordFilterGroupToViewFilterGroup } from '@/views/utils/mapRecordFilterGroupToViewFilterGroup';
+import { mapRecordFilterToViewFilter } from '@/views/utils/mapRecordFilterToViewFilter';
+import { mapRecordSortToViewSort } from '@/views/utils/mapRecordSortToViewSort';
+import { useStore } from 'jotai';
+import { useCallback } from 'react';
+import { isDefined } from 'twenty-shared/utils';
+import { WidgetConfigurationType } from '~/generated-metadata/graphql';
+
+export const useSaveRecordTableWidgetsViewDataOnDashboardSave = () => {
```
charlesBochet, line 32: could we move all these hook to a settings-edit subfolder and make clear in their naming to avoid confusion

packages/twenty-front/src/modules/page-layout/widgets/record-table/types/RecordTableWidgetViewFieldItem.ts
```diff
@@ -0,0 +1,7 @@
+import { type FieldMetadataItem } from '@/object-metadata/types/FieldMetadataItem';
+import { type ViewField } from '@/views/types/ViewField';
+
+export type RecordTableWidgetViewFieldItem = {
```
charlesBochet, line 4: likely not useful

packages/twenty-front/src/modules/page-layout/widgets/record-table/utils/filterFieldsForRecordTableViewCreation.ts
```diff
@@ -0,0 +1,8 @@
+import { type FieldMetadataItem } from '@/object-metadata/types/FieldMetadataItem';
+import { isHiddenSystemField } from '@/object-metadata/utils/isHiddenSystemField';
+
+export const filterFieldsForRecordTableViewCreation = (
```
charlesBochet, line 4: I feel this should already exist in the codebase, nothing specific to record table

packages/twenty-front/src/modules/page-layout/widgets/record-table/utils/sortFieldsByRelevanceForRecordTableWidget.ts
```diff
@@ -0,0 +1,39 @@
+import { type FieldMetadataItem } from '@/object-metadata/types/FieldMetadataItem';
+import { FieldMetadataType, RelationType } from 'twenty-shared/types';
+
+export const sortFieldsByRelevanceForRecordTableWidget = (
```
charlesBochet, line 4: same here, I feel this should be unify even when we want to add a field on regular table (even on any index view

packages/twenty-server/src/engine/metadata-modules/page-layout-widget/dtos/record-table-configuration.dto.ts
```diff
@@ -0,0 +1,19 @@
...
+export class RecordTableConfigurationDTO implements RecordTableConfiguration {
+  @Field(() => WidgetConfigurationType)
+  @IsIn([WidgetConfigurationType.RECORD_TABLE])
+  @IsNotEmpty()
+  configurationType: WidgetConfigurationType.RECORD_TABLE;
+
+  @Field(() => String, { nullable: true })
+  @IsOptional()
+  @IsUUID()
+  viewId?: string;
```
charlesBochet, line 18: is this expected to be optional?

charlesBochet (approved):

PR #18748, opened 2026-03-18
[Fix]: Vertical bar charts with max data range do not show data up to the top if above max data range

carbonFibreCode:
fixes issue : #18606 

we have removed the irrelevant use of filters in the backend which removes the values that are out of max, min range hence modifying the rawValues.

- Modified relevant tests
- fixed regressions (Horizontal bar charts rightLabel cropping)
- removed dead utility filterByRange code

Before :
Screenshot 2026-03-18 at 10 24 14 PM

After:
Screenshot 2026-03-18 at 11 16 33 PM

Screenshot 2026-03-19 at 12 16 41 AM


ehconitin (approved):
lgtm! I cleaned up some code -- and pushed some random edge cases I saw while reviewing!

thanks alot for the help!

PR #18766, opened 2026-03-19
Fix: Hide object from Opened section when it already has a workspace nav item

abdulrahmancodes:
Previously, the Opened section was always hidden for all workflow objects. We now base the “in nav” check on the full set of object metadata IDs that have a workspace nav item, instead of only active non-system objects. So: if an object has a nav item (e.g. workflow runs), it no longer appears under Opened; if it doesn’t, it can appear there. The hook was simplified to only expose `objectMetadataIdsInWorkspaceNav`, and the unused return value was removed.


etiennejouan (approved):
LGTM ! But from a product point of view, not sure it's coherent 🤔 Think it needs some more explanations from bonapara. In any case, always good to remove dead code, thanks !

PR #18768, opened 2026-03-19
Improve workflow metrics with faillure reason

thomtrp:
- split workflow errors into system and user errors
- add workspace id to attributes for debugging


etiennejouan (approved):
LGTM !

PR #18770, opened 2026-03-19
fix - handle invoice.paid webhook to recover unpaid subscriptions

etiennejouan:
Fixes https://github.com/twentyhq/private-issues/issues/432

## Problem

When a user's invoice goes unpaid, Stripe moves their subscription to `unpaid` status, and Twenty suspends the workspace. But if the user pays that invoice while a new billing period has started, Stripe has already generated a new **draft** invoice for that period. Since the draft isn't finalized or paid, Stripe doesn't reactivate the subscription — the workspace stays suspended indefinitely.

## What was missing

- No handler for the `invoice.paid` Stripe webhook event
- No mechanism to finalize draft invoices that accumulated during the `unpaid` period
- No way to reset the workspace deletion countdown (`suspendedAt`) when a user shows payment intent

### 1. `StripeInvoiceService` — new Stripe SDK wrapper

- `listDraftInvoices(stripeSubscriptionId)` — lists all draft invoices for a subscription
- `finalizeInvoice(invoiceId)` — finalizes a draft with `auto_advance: true` so Stripe auto-charges it

### 2. `INVOICE_PAID` enum value

Added to `BillingWebhookEvent` so the controller can route it.

### 3. `processInvoicePaid()` in `BillingWebhookInvoiceService`

New private handler that:

- Fetches all draft invoices for the subscription
- Filters to only those whose `period_end` is in the past (already overdue)
- Finalizes each one (with error handling per invoice to avoid blocking the webhook)
- If the workspace is suspended, resets `suspendedAt` to now (buys time before deletion)

### 4. Controller routing

`INVOICE_FINALIZED` and `INVOICE_PAID` are now grouped in the same switch case, both calling `processStripeEvent(data, eventType)`, which forks internally in the service.

## Expected recovery flow

User pays overdue invoice
→ Stripe fires invoice.paid
→ Handler finalizes past-due draft invoices (auto_advance: true)
→ Stripe auto-charges them
→ Subscription becomes active
→ customer.subscription.updated fires
→ Existing logic unsuspends workspace
→ suspendedAt refreshed (resets deletion countdown while payments cascade)

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/billing-webhook/billing-webhook.controller.ts
+++ b/packages/twenty-server/src/engine/core-modules/billing-webhook/billing-webhook.controller.ts
@@ -128,8 +128,7 @@ export class BillingWebhookController {
       case BillingWebhookEvent.INVOICE_FINALIZED:
       case BillingWebhookEvent.INVOICE_PAID:
         return await this.billingWebhookInvoiceService.processStripeEvent(
-          event.data,
-          event.type,
+          event,
         );
 
       case BillingWebhookEvent.CUSTOMER_CREATED:
--- a/packages/twenty-server/src/engine/core-modules/billing-webhook/billing-webhook.controller.ts
+++ b/packages/twenty-server/src/engine/core-modules/billing-webhook/billing-webhook.controller.ts
@@ -126,8 +126,10 @@ export class BillingWebhookController {
         );
 
       case BillingWebhookEvent.INVOICE_FINALIZED:
+      case BillingWebhookEvent.INVOICE_PAID:
         return await this.billingWebhookInvoiceService.processStripeEvent(
           event.data,
+          event.type,
         );
 
       case BillingWebhookEvent.CUSTOMER_CREATED:
--- a/packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-invoice.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-invoice.service.ts
@@ -148,24 +148,29 @@ export class BillingWebhookInvoiceService {
 
   private async processInvoicePaid(data: Stripe.InvoicePaidEvent.Data) {
     const stripeSubscriptionId = getSubscriptionIdFromInvoice(data.object);
-
-    if (!isDefined(stripeSubscriptionId)) {
-      return;
-    }
-
     const stripeCustomerId = data.object.customer as string | undefined;
     const paidInvoicePeriodEnd = data.object.period_end;
 
-    if (isDefined(paidInvoicePeriodEnd)) {
-      await this.finalizePastDueDraftInvoicesAfterPaidInvoice(
-        stripeSubscriptionId,
-        paidInvoicePeriodEnd,
+    if (
+      !isDefined(stripeSubscriptionId) ||
+      !isDefined(stripeCustomerId) ||
+      !isDefined(paidInvoicePeriodEnd)
+    ) {
+      throw new BillingException(
+        'Invalid invoice paid event data',
+        BillingExceptionCode.BILLING_STRIPE_ERROR,
       );
     }
 
-    if (isDefined(stripeCustomerId)) {
-      await this.delaySuspendedWorkspaceCleanup(stripeCustomerId);
-    }
+    // Paying a past-due invoice won't reactivate the subscription if Stripe
+    // already generated a draft for the next period. Finalize it so Stripe
+    // can collect payment and resume the subscription.
+    await this.finalizePastDueDraftInvoicesAfterPaidInvoice(
+      stripeSubscriptionId,
+      paidInvoicePeriodEnd,
+    );
+
+    await this.delaySuspendedWorkspaceCleanup(stripeCustomerId);
 
     return { stripeSubscriptionId };
   }
--- a/packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-invoice.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-invoice.service.ts
@@ -9,6 +9,10 @@ import { type Repository } from 'typeorm';
 import type Stripe from 'stripe';
 
 import { getSubscriptionIdFromInvoice } from 'src/engine/core-modules/billing-webhook/utils/get-subscription-id-from-invoice.util';
+import {
+  BillingException,
+  BillingExceptionCode,
+} from 'src/engine/core-modules/billing/billing.exception';
 import { BillingCustomerEntity } from 'src/engine/core-modules/billing/entities/billing-customer.entity';
 import { BillingSubscriptionItemEntity } from 'src/engine/core-modules/billing/entities/billing-subscription-item.entity';
 import { BillingSubscriptionEntity } from 'src/engine/core-modules/billing/entities/billing-subscription.entity';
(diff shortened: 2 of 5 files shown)


packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-invoice.service.ts
```diff
@@ -1,33 +1,62 @@
...
+    eventType: string,
+  ) {
+    if (eventType === BillingWebhookEvent.INVOICE_PAID) {
+      return this.processInvoicePaid(data as Stripe.InvoicePaidEvent.Data);
+    }
+
+    if (eventType === BillingWebhookEvent.INVOICE_FINALIZED) {
+      return this.processInvoiceFinalized(
+        data as Stripe.InvoiceFinalizedEvent.Data,
+      );
```
thomtrp, line 58: I think that would be cleaner if we have something like:

`event: StripeEvent`
`StripeEvent: { type: BillingWebhookEvent.INVOICE_FINALIZED, data: Stripe.InvoiceFinalizedEvent.Data } | { type: BillingWebhookEvent.INVOICE_PAID ... }`

packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-invoice.service.ts
```diff
@@ -112,6 +141,74 @@ export class BillingWebhookInvoiceService {
...
+    return { stripeSubscriptionId };
+  }
+
+  private async finalizePastDueDraftInvoices(
+    stripeSubscriptionId: string,
+  ): Promise<void> {
+    const draftInvoices =
+      await this.stripeInvoiceService.listDraftInvoices(stripeSubscriptionId);
+
+    const now = Date.now() / 1000;
```
thomtrp: why do we need this / 1000 ?

packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-invoice.service.ts
```diff
@@ -112,6 +141,74 @@ export class BillingWebhookInvoiceService {
...
   }
 
+  private async processInvoicePaid(data: Stripe.InvoicePaidEvent.Data) {
+    const stripeSubscriptionId = getSubscriptionIdFromInvoice(data.object);
+
+    if (!isDefined(stripeSubscriptionId)) {
+      return;
+    }
+
+    const stripeCustomerId = data.object.customer as string | undefined;
```
thomtrp, line 151: nit: `stripeCustomerId: string | undefined = data.object.customer`

packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-invoice.service.ts
thomtrp: ok got it bellow. We should find a better naming that now then

packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-invoice.service.ts
```diff
@@ -112,6 +141,74 @@ export class BillingWebhookInvoiceService {
...
+      (invoice) => isDefined(invoice.period_end) && invoice.period_end < now,
+    );
+
+    for (const invoice of pastDueDraftInvoices) {
+      try {
+        await this.stripeInvoiceService.finalizeInvoice(invoice.id);
+      } catch (error) {
+        this.logger.error(
+          `Failed to finalize draft invoice ${invoice.id}: ${error instanceof Error ? error.message : String(error)}`,
+        );
```
thomtrp, line 201: will we see that error?

charlesBochet: agree

packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-invoice.service.ts
```diff
@@ -112,6 +141,74 @@ export class BillingWebhookInvoiceService {
...
+      }
+    }
+  }
+
+  private async refreshSuspendedAtIfNeeded(
```
charlesBochet: naming could be clearer

packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-invoice.service.ts
charlesBochet: nowInSeconds

packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-invoice.service.ts
charlesBochet: actually let's not lie on the types.

we should rather use isNonEmptyString line 155 (and you should actually throw if this is not a string but a Customer type (your code does not handle it)

packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-invoice.service.ts
charlesBochet: we should throw a CustomException here

packages/twenty-server/src/engine/core-modules/billing-webhook/billing-webhook.controller.ts
```diff
@@ -126,8 +126,10 @@ export class BillingWebhookController {
         );
 
       case BillingWebhookEvent.INVOICE_FINALIZED:
+      case BillingWebhookEvent.INVOICE_PAID:
         return await this.billingWebhookInvoiceService.processStripeEvent(
```
charlesBochet, line 130: I would pass the event here and make a Pick<> in billingWebhookInvoiceService.processStripeEvent

packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-invoice.service.ts
```diff
@@ -112,6 +141,74 @@ export class BillingWebhookInvoiceService {
...
+
+    const now = Date.now() / 1000;
+
+    const pastDueDraftInvoices = draftInvoices.filter(
+      (invoice) => isDefined(invoice.period_end) && invoice.period_end < now,
+    );
+
+    for (const invoice of pastDueDraftInvoices) {
+      try {
+        await this.stripeInvoiceService.finalizeInvoice(invoice.id);
```
charlesBochet, line 196: I don't know about this, is it our job to do that? I would leave it to Stripe?

etiennejouan: Unfortunately, that's the issue we had. The next cycle's invoice was not 'finalized' after previous invoice paid, so it wasn't payable (as seen in Stripe), and therefore the subscription could not be reactivated (it's not easy to be certain that paying the last invoice would reactivate an 'unpaid' subscription, because of incomplete stripe doc. but, it's the bet we're making).

packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-invoice.service.ts
```diff
@@ -112,6 +146,85 @@ export class BillingWebhookInvoiceService {
...
   }
 
+  private async processInvoicePaid(data: Stripe.InvoicePaidEvent.Data) {
+    const stripeSubscriptionId = getSubscriptionIdFromInvoice(data.object);
+
+    if (!isDefined(stripeSubscriptionId)) {
+      return;
+    }
+
+    const stripeCustomerId = data.object.customer as string | undefined;
```
charlesBochet, line 151: throw before
and this is nominal logic

packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-invoice.service.ts
```diff
@@ -112,6 +146,85 @@ export class BillingWebhookInvoiceService {
...
+    }
+
+    if (isDefined(stripeCustomerId)) {
+      await this.delaySuspendedWorkspaceCleanup(stripeCustomerId);
+    }
+
+    return { stripeSubscriptionId };
+  }
+
+  private async finalizePastDueDraftInvoicesAfterPaidInvoice(
```
charlesBochet, line 178: explain why we finalize invoices (edge case)

charlesBochet (approved):

charlesBochet:
Left final comments

PR #18771, opened 2026-03-19
fix: preserve percent-encoding in URL fields during import and record transform

victorjzq:
## Problem

Fixes #18698

When importing records with URLs containing percent-encoded characters (e.g., `%2F` in Google Maps `data=` params), Twenty CRM decodes them to their literal equivalents (`/`), breaking the URLs.

**Before:**
```
https://www.google.com/maps/place/...!16s%2Fg%2F1ptwh8096!...
```
**After import (broken):**
```
https://www.google.com/maps/place/...!16s/g/1ptwh8096!...
```

## Root Cause

`lowercaseUrlOriginAndRemoveTrailingSlash` was calling `safeDecodeURIComponent` on `url.pathname` and `url.search` when reconstructing the URL. This decoded `%2F` → `/` inside structured URL components like Google Maps' `data=` parameter.

## Fix

Remove the `safeDecodeURIComponent` calls. `URL.pathname` and `URL.search` already preserve the original percent-encoding — no decoding is needed when reconstructing the URL string.

```diff
-  const path =
-    safeDecodeURIComponent(url.pathname) +
-    safeDecodeURIComponent(url.search) +
-    url.hash;
+  const path = url.pathname + url.search + url.hash;
```

The origin is still lowercased (correct behavior). Trailing slash removal still works.


victorjzq:
Closing in favor of a cleaner branch. The URL encoding fix is in PR #18773.

PR #18773, opened 2026-03-19
fix: preserve percent-encoding in URL fields during import and record transform

victorjzq:
## Problem

Fixes #18698

When importing records with URLs containing percent-encoded characters like `%2F`, Twenty decodes them to literal characters (`/`), breaking structured URLs such as Google Maps links.

**Broken import result:**
```
https://www.google.com/maps/...!16s/g/1ptwh8096!...   ❌ (broken)
```
**Expected:**
```
https://www.google.com/maps/...!16s%2Fg%2F1ptwh8096!... ✅
```

## Root Cause

`lowercaseUrlOriginAndRemoveTrailingSlash` was calling `safeDecodeURIComponent` on `url.pathname` and `url.search`. The `URL` object already parses and preserves encoding correctly — decoding the path/search unnecessarily broke encoded characters.

## Fix

```diff
-  const path =
-    safeDecodeURIComponent(url.pathname) +
-    safeDecodeURIComponent(url.search) +
-    url.hash;
+  const path = url.pathname + url.search + url.hash;
```

The origin is still lowercased. Trailing slash removal still works. Only the unnecessary decode is removed.


charlesBochet:
Closing

PR #18774, opened 2026-03-19
fix: FIND_RECORDS RELATION IS filter should return 0 results when variable resolves to null

### LINKED ISSUE #18744: Bug: FIND_RECORDS silently drops filter conditions when variable resolves to null/empty
### Description

When a `FIND_RECORDS` workflow step uses a variable reference (e.g., `{{previousStep.first.id}}`) in a `RELATION` type filter, and that variable resolves to `null`, an empty string, or `undefined` at runtime (because the previous step returned no results), the **filter condition is silently removed** from the query instead of returning 0 results.

This causes the `FIND_RECORDS` step to match **all records** that satisfy the remaining filter conditions — a dangerous data leak / logic error.

### Reproduction

**Workflow structure:**
1. `Search Retreat` (FIND_RECORDS on `retreat`) — returns 0 results (empty)
2. `Find PAID Orders` (FIND_RECORDS on `order`) — filters:
   - `Retreat IS {{Search Retreat.first.id}}` → resolves to **null**
   - `Status IS "PAID"`
   - `Step Current IS "INTEGRATION_PROCESSING"`

**Expected behavior:**
`Find PAID Orders` should return **0 results** because no order can have `retreatId = null` (or alternatively, the step should fail with an error indicating the filter value is invalid).

**Actual behavior:**
`Find PAID Orders` returns **all orders** matching `Status = PAID AND stepCurrent = INTEGRATION_PROCESSING` across **ALL retreats**, because the RELATION filter is silently dropped.

**Evidence from workflow runs:**

| Run | Search Retreat totalCount | Retreat filter resolves to | Find PAID Orders result |
|-----|--------------------------|---------------------------|------------------------|
| #7  | 1 (retreat found)        | Valid UUID                 | 3 orders ✅ (correct)   |
| #8  | 0 (empty)                | null                       | 3 orders ❌ (BUG — returns orders from ALL retreats) |
| #11 | 0 (empty)                | null                       | 1 order ❌ (BUG)        |

### Root Cause

The issue is in **`turnRecordFilterIntoRecordGqlOperationFilter`** ([source](https://github.com/twentyhq/twenty/blob/main/packages/twenty-shared/src/utils/filter/turnRecordFilterIntoGqlOperationFilter.ts)), specifically the `RELATION` case:

```typescript
case 'RELATION': {
  const { isCurrentWorkspaceMemberSelected, selectedRecordIds } =
    jsonRelationFilterValueSchema
      .catch({
        isCurrentWorkspaceMemberSelected: false,
        selectedRecordIds: arrayOfUuidOrVariableSchema.parse(
          recordFilter.value,  // ← This is null/empty string at runtime
        ),
      })
      .parse(recordFilter.value);

  // ...

  if (!isDefined(recordIds) || recordIds.length === 0) return;  // ← Returns undefined!
```

When `recordFilter.value` is null/empty (because `resolveInput()` resolved the variable `{{step.first.id}}` to null), `arrayOfUuidOrVariableSchema.parse()` produces an empty array `[]`, and the function **returns `undefined`**.

Then in **`turnRecordFilterGroupIntoGqlOperationFilter`** ([source](https://github.com/twentyhq/twenty/blob/main/packages/twenty-shared/src/utils/filter/turnRecordFilterGroupIntoGqlOperationFilter.ts)):

```typescript
const groupRecordGqlOperationFilters = recordFiltersInGroup
  .map((recordFilter) => turnRecordFilterIntoRecordGqlOperationFilter(...))
  .filter(isDefined);  // ← Silently removes the undefined RELATION filter

return { and: [...groupRecordGqlOperationFilters, ...] };
// ← AND condition now has one fewer condition, broadening the query
```

The `undefined` result is filtered out by `.filter(isDefined)`, effectively removing that condition from the AND group. The query degrades from:

```sql
WHERE retreatId IN [<uuid>] AND status = 'PAID' AND stepCurrent = '...'
```

to:

```sql
WHERE status = 'PAID' AND stepCurrent = '...'
```

### Scope of Impact

This is **not limited to RELATION filters**. The same `return undefined` pattern exists for **all filter types** that validate their input:

| Filter Type | Code | Same Bug? |
|-------------|------|-----------|
| `RELATION`  | `if (recordIds.length === 0) return;` | ✅ Yes |
| `UUID`      | `if (recordIds.length === 0) return;` | ✅ Yes |
| `SELECT`    | `if (options.length === 0) return;` | ✅ Yes |
| `MULTI_SELECT` | `if (options.length === 0) return;` | ✅ Yes |
| `PHONES`    | `if (!isNonEmptyString(filterValue)) return;` | ✅ Yes |

Any workflow that chains `FIND_RECORDS` steps where a later step references a value from an earlier step that might be empty is vulnerable to this silent filter-dropping behavior.

### Suggested Fix

In `turnRecordFilterIntoRecordGqlOperationFilter`, when a filter value resolves to empty/null **in a workflow context**, the function should either:

**Option A (Recommended): Return an impossible filter instead of `undefined`**

```typescript
// Instead of:
if (!isDefined(recordIds) || recordIds.length === 0) return;

// Return a filter that matches nothing:
if (!isDefined(recordIds) || recordIds.length === 0) {
  return {
    [correspondingFieldMetadataItem.name + 'Id']: {
      is: 'NOT_NULL',
    },
    // Or alternatively, use a dedicated "match nothing" sentinel
  };
}
```

**Option B: Throw an error when a resolved variable produces an empty filter value**

This would make the workflow step fail explicitly rather than silently broadening the query.

**Option C: Add null/empty-value awareness to `computeRecordGqlOperationFilter`**

Track whether any filter returned `undefined` due to empty values (vs. being intentionally unset), and if so, either return an empty result set or throw.

### Context

- Related (but different issue): #16141 — fixed variable resolution in frontend, but the null-value-drops-filter behavior was not addressed
- Related PR: #16147 — moved filter computation to backend (correct), but the backend `turnRecordFilterIntoRecordGqlOperationFilter` still has the same silent-drop behavior
- The `resolveInput()` function in `variable-resolver.ts` correctly resolves missing variables to null/empty — the problem is downstream in filter construction

### Environment

- Twenty version: Self-hosted, latest as of March 2026
- Workflow type: CRON-triggered with chained FIND_RECORDS steps

### Severity

**Critical** — This is a silent data integrity issue. Workflows that appear to complete successfully are actually operating on the wrong dataset. There is no error, no warning, and no indication in the workflow run logs that a filter was dropped. The only way to detect the issue is by manually inspecting the result counts.

victorjzq:
## Problem

Fixes #18744

**Critical bug**: When a `FIND_RECORDS` workflow step uses a variable reference (e.g. `{{previousStep.first.id}}`) in a `RELATION IS` filter, and that variable resolves to `null` at runtime, the filter is **silently dropped** — causing the step to return ALL matching records instead of zero. This is a data leak.

| Run | Variable value | Expected | Actual (before fix) |
|-----|----------------|----------|---------------------|
| Previous step found a record | Valid UUID | Matching records | ✅ Correct |
| Previous step returned empty | `null` | 0 records | ❌ ALL records (filter dropped) |

## Root Cause

`arrayOfUuidOrVariableSchema.parse(null)` returns `[]` (via `.catch([])`). The code then hit:

```typescript
if (!isDefined(recordIds) || recordIds.length === 0) return; // ← drops filter
```

Returning `undefined` caused the caller to treat the filter as if it didn't exist.

## Fix

When `recordIds` is empty and operand is `IS`, return a match-nothing filter (`in: []`) instead of `undefined`:

```typescript
if (!isDefined(recordIds) || recordIds.length === 0) {
  if (recordFilter.operand === RecordFilterOperand.IS) {
    return {
      [correspondingFieldMetadataItem.name + 'Id']: { in: [] },
    };
  }
  return; // IS_NOT with no IDs: drop is correct (exclude nothing = all)
}
```

### DIFF
--- a/packages/twenty-shared/src/utils/filter/turnRecordFilterIntoGqlOperationFilter.ts
+++ b/packages/twenty-shared/src/utils/filter/turnRecordFilterIntoGqlOperationFilter.ts
@@ -532,7 +532,18 @@ export const turnRecordFilterIntoRecordGqlOperationFilter = ({
           ]
         : selectedRecordIds;
 
-      if (!isDefined(recordIds) || recordIds.length === 0) return;
+      if (!isDefined(recordIds) || recordIds.length === 0) {
+        // When the filter value is empty (e.g. a variable resolved to null),
+        // IS should match nothing rather than being silently dropped.
+        if (recordFilter.operand === RecordFilterOperand.IS) {
+          return {
+            [correspondingFieldMetadataItem.name + 'Id']: {
+              in: [],
+            } as RelationFilter,
+          };
+        }
+        return;
+      }
 
       switch (recordFilter.operand) {
         case RecordFilterOperand.IS:


charlesBochet:
Closing, see: https://github.com/twentyhq/twenty/pull/18776

PR #18775, opened 2026-03-19
fix: show tooltip when email edit is disabled due to multiple workspaces

### LINKED ISSUE #2335: Inform user when they can't change their email if they belong to 2 or more workspaces
Users can change their emails if:
- in Security settings email is selected as editable profile field
- user has a role with permission allowing to change their profile details (under Actions)
- user belongs to only one workspace (if they belong to two or more workspaces, they can't change their email)

I found out about the last point accidentally today and it's unclear why I couldn't change an email despite meeting 2 first requirements

In short, there should be a popup appearing after hover over the pen icon informing user that they can't change their email because user belongs to 2 or more workspaces.

https://github.com/user-attachments/assets/f9c6cb5b-fd02-404f-b98d-3ae846f40e09

victorjzq:
## Problem

Fixes twentyhq/core-team-issues#2335

When a user belongs to multiple workspaces, the email edit pencil button is disabled with no explanation. Users have no way to discover why.

## Fix

1. `useCanEditProfileField` now returns a `disabledReason` field (`'multipleWorkspaces' | undefined`)
2. `EmailField` shows an `AppTooltip` on hover when `disabledReason === 'multipleWorkspaces'`:

> *"Email cannot be changed when you belong to multiple workspaces"*

| Scenario | Before | After |
|----------|--------|-------|
| Single workspace | Edit works | Edit works (unchanged) |
| Multiple workspaces | Button disabled, no hint | Button disabled + tooltip |

### DIFF
--- a/packages/twenty-front/src/modules/settings/profile/components/EmailField.tsx
+++ b/packages/twenty-front/src/modules/settings/profile/components/EmailField.tsx
@@ -1,3 +1,4 @@
+import { t } from '@lingui/core/macro';
 import { styled } from '@linaria/react';
 import { useId, useState } from 'react';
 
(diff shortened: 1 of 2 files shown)


charlesBochet:
Closing: see https://github.com/twentyhq/twenty/pull/18776

PR #18776, opened 2026-03-19
fix: sync core user firstName/lastName when workspace member name is updated

victorjzq:
## Problem

Fixes #18610

When a user completes onboarding on a self-hosted instance, the `core.user` table has empty `firstName` and `lastName` fields, even though `workspaceMember` is correctly populated. This causes "Created by" to show "Untitled" on all records created by the first admin.

## Root Cause

The onboarding flow writes the user's name to `workspaceMember` but never syncs it back to `core.user`. The `WorkspaceMemberUpdatedListener` was not listening to the `updated` event on workspace members.

## Fix

Added `updateUserNames()` to `UserService` and a `WorkspaceMemberUpdatedListener` that syncs `firstName`/`lastName` to `core.user` whenever `workspaceMember.name` is updated.

- `user.service.ts` — new `updateUserNames(workspaceMemberId, firstName, lastName)` method
- `workspace-member-updated.listener.ts` — new event listener that calls `updateUserNames` on workspace member update events

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/user/services/user.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/user/services/user.service.ts
@@ -360,6 +360,26 @@ export class UserService extends TypeOrmQueryService<UserEntity> {
     return updatedUser;
   }
 
+  async updateUserNames(
+    userId: string,
+    {
+      firstName,
+      lastName,
+    }: { firstName?: string; lastName?: string },
+  ) {
+    const user = await this.findUserByIdOrThrow(userId);
+
+    if (isDefined(firstName)) {
+      user.firstName = firstName;
+    }
+
+    if (isDefined(lastName)) {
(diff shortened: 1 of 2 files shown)


charlesBochet:
@victorjzq Please stop opening PRs with IA, you are not helping :( 
This clearly has not been carefully thought, you are actually harming the community here. Some of your PRs actually makes sense but we could not afford this kind of behavior on the project or this will become a nightmare.

If you truly want to help, take the time to understand the product. Here: a user can have multiple workspaceMember / userWorkspace. When a given workspaceMember is updated, should it update the user?

charlesBochet:
Confirmed AI slop, here is a correct fix: https://github.com/twentyhq/twenty/pull/18799

@victorjzq please avoid making PRs on twenty in the future, they are not welcome unless you treat us with more respect

PR #18779, opened 2026-03-19
fix: stop event propagation when removing file in AI chat preview

### LINKED ISSUE #18298: Bug: Unable to delete files once upload - AI chatbot

victorjzq:
## Summary

When clicking the ✕ button on an uploaded file in the AI chatbot context preview, the click event bubbled up to the `StyledClickableContainer` parent, which triggered `handleClick` (opening the file preview modal) instead of — or in addition to — calling `onRemove`.

**Root cause:** `StyledClickableContainer` has `onClick={handleClick}` at the div level. The `AvatarOrIcon` (X button) `onClick={onRemove}` callback didn't stop propagation, so the event continued to bubble and opened the preview.

**Fix:** Wrap `onRemove` in a `handleRemove` callback that calls `e.stopPropagation()` before invoking the original handler.

Fixes #18298

### DIFF
--- a/packages/twenty-front/src/modules/ai/components/internal/AgentChatFilePreview.tsx
+++ b/packages/twenty-front/src/modules/ai/components/internal/AgentChatFilePreview.tsx
@@ -7,7 +7,7 @@ import { filePreviewState } from '@/ui/field/display/states/filePreviewState';
 import { useSetAtomState } from '@/ui/utilities/state/jotai/hooks/useSetAtomState';
 import { styled } from '@linaria/react';
 import { t } from '@lingui/core/macro';
-import { useCallback, useContext } from 'react';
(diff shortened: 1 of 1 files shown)


charlesBochet (approved):
Rewrote according to our project guidelines

PR #18074, opened 2026-02-19
Rework types for logic function

charlesBochet:
## Summary

- **Consolidate logic function services**: Remove `LogicFunctionMetadataService` and consolidate all logic function CRUD operations into `LogicFunctionFromSourceService`, with a new `LogicFunctionFromSourceHelperService` for shared validation/migration logic
- **Introduce typed conversion utils following the skill pattern**: Add `fromCreateLogicFunctionFromSourceInputToUniversalFlatLogicFunctionToCreate` and `fromUpdateLogicFunctionFromSourceInputToFlatLogicFunctionToUpdate` that convert DTO inputs directly to flat entities (`UniversalFlatLogicFunction` / `FlatLogicFunction`), replacing the previous intermediate `UpdateLogicFunctionMetadataParams` indirection
- **Simplify `CodeStepBuildService`**: Remove ~100 lines of manual duplication logic by delegating to `LogicFunctionFromSourceService.duplicateOneWithSource`
- **Remove completed 1-17 migration**: Delete `MigrateWorkflowCodeStepsCommand` and associated utils that migrated workflow code steps from serverless functions to logic functions

PR #18780, opened 2026-03-19
Restore missing workflow code step migration (#18296)

dharunashokkumar:
The upgrade command that migrates CODE step JSONB fields from serverlessFunctionId to logicFunctionId was deleted in #18074, breaking self-hosted instances upgrading from v1.16 to v1.18+.


charlesBochet:
Hi @dharunashokkumar, you should not upgrade from 1.16 to 1.18 right now, please upgrade from 1.17 (which contains the code you are restoring) first then 1.18

dharunashokkumar:
got it, thanks for the context! Makes sense that the migration lives in the 1.17 path. I'll close this out...... appreciate the quick response.

PR #18778, opened 2026-03-19
fix: restore overflow +N count for relation-to-many chips in table cells

victorjzq:
## Summary

The table performance optimization in #18304 (commit `159bb9d70a`) replaced `FieldFocusContextProvider` (dynamic `useState(false)`) with `FieldFocusStaticUnfocusedProvider` (hardcoded `isFocused: false`) in `RecordTableCell.tsx`. This caused `RelationFromManyFieldDisplay` and `MorphRelationOneToManyFieldDisplay` to always pass `isChipCountDisplayed={false}` to `ExpandableList`, permanently suppressing the `+N` overflow count badge in non-hovered cells.

The chips themselves are still sliced (hidden) but the `+N` indicator that shows users there are more is never rendered.

**Fix:** Remove the `isChipCountDisplayed` prop from `ExpandableList` in both components, letting `ExpandableList` use its own internal `onMouseEnter`/`onMouseLeave` hover logic to show the count.

## Changes

- `RelationFromManyFieldDisplay.tsx`: Remove `isChipCountDisplayed={isFocused}` from all `ExpandableList` usages; remove unused `useFieldFocus`, `StyledContainer`, `styled`, `themeCssVariables`
- `MorphRelationOneToManyFieldDisplay.tsx`: Same — remove `isChipCountDisplayed={isFocused}`, remove unused `useFieldFocus`

Fixes #18729

PR #18781, opened 2026-03-19
fix: restore overflow +N count for multi-select chips in table cells

### LINKED ISSUE #18729: Relation to many display field doesn't show all relation until hovered
https://github.com/user-attachments/assets/e97af074-e5ba-4897-a087-f30cb540d159

victorjzq:
## Problem

Fixes #18729 (multi-select aspect)

The table performance optimization (commit `159bb9d7`, [#18304](https://github.com/twentyhq/twenty/pull/18304)) replaced `FieldFocusContextProvider` with `FieldFocusStaticUnfocusedProvider`, which hardcodes `isFocused = false` in all table cells.

`MultiSelectFieldDisplay` used `isFocused` to conditionally render:
- **When focused (hovered):** `<ExpandableList isChipCountDisplayed={isFocused}>` — shows +N overflow badge  
- **When unfocused (default):** `<MultiSelectDisplay>` — `overflow: hidden` with no overflow indicator

Since `isFocused` is **always `false`** after the perf optimization, the `ExpandableList` was never rendered. Overflowing tags were silently hidden with no visual indicator of how many values are missing.

## Root Cause

Same as [#18778](https://github.com/twentyhq/twenty/pull/18778) for relation-to-many fields:
- `ExpandableList` has two modes: props-controlled (`isChipCountDisplayed` prop) and internal mouse-hover
- Passing `isChipCountDisplayed={false}` **disables** the internal hover fallback entirely
- Since `isFocused` is always false, the +N badge never appeared

## Fix

Always render `<ExpandableList>` without `isChipCountDisplayed` prop, letting it use its own internal mouse-hover state to show the overflow count badge.

Remove the now-unused `useFieldFocus` import and `MultiSelectDisplay` fallback (which was only needed for the unfocused case).

## Before / After

| State | Before | After |
|-------|--------|-------|
| Normal (unfocused) | Tags clipped with no count | Tags shown with +N badge on hover |
| Hovered | ExpandableList shown (never reached) | ExpandableList with +N badge |

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/display/components/MultiSelectFieldDisplay.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/display/components/MultiSelectFieldDisplay.tsx
@@ -1,14 +1,10 @@
-import { useFieldFocus } from '@/object-record/record-field/ui/hooks/useFieldFocus';
 import { useMultiSelectFieldDisplay } from '@/object-record/record-field/ui/meta-types/hooks/useMultiSelectFieldDisplay';
-import { MultiSelectDisplay } from '@/ui/field/display/components/MultiSelectDisplay';
 import { ExpandableList } from '@/ui/layout/expandable-list/components/ExpandableList';
 import { Tag } from 'twenty-ui/components';
 import { isDefined } from 'twenty-shared/utils';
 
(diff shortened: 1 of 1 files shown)


charlesBochet:
Closing this, we are going to revisit focus and active states in upcoming weeks

PR #18782, opened 2026-03-19
fix: emails and phones field display overflow and missing +N chip count

victorjzq:
## Problem

In table cells, `FieldFocusStaticUnfocusedProvider` (introduced in commit `159bb9d7` for performance optimization) hardcodes `isFocused = false`. Both `EmailsDisplay` and `PhonesDisplay` had a conditional pattern:

```tsx
return isFocused ? (
  <ExpandableList isChipCountDisplayed>...</ExpandableList>
) : (
  <StyledContainer> {/* overflow: hidden, no +N count */} </StyledContainer>
);
```

Since `isFocused` is always `false` in table cells, these fields:
- Always render with `overflow: hidden` (content clipped)
- Never show the `+N` overflow chip count

Additionally, `EmailsFieldDisplay` never passed `isFocused` at all, so the bug existed even outside the table performance optimization context.

## Fix

Remove the `isFocused` conditional and always render `<ExpandableList>` without the `isChipCountDisplayed` prop. This lets `ExpandableList`'s internal mouse-hover state control the overflow chip count display — consistent with how `RelationFromManyFieldDisplay` and `MultiSelectFieldDisplay` work.

**Changes:**
- `EmailsDisplay`: remove `isFocused` prop and `StyledContainer`, always render `<ExpandableList>`
- `PhonesDisplay`: same as above, also remove unused `styled` import
- `PhonesFieldDisplay`: remove `useFieldFocus` import and `isFocused` prop pass-through

## Root cause

Same root cause as #18778 (relation chip count) and #18781 (multi-select chip count) — the `FieldFocusStaticUnfocusedProvider` performance optimization broke components that conditionally rendered based on `isFocused`.

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/display/components/MultiSelectFieldDisplay.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/display/components/MultiSelectFieldDisplay.tsx
@@ -1,14 +1,10 @@
-import { useFieldFocus } from '@/object-record/record-field/ui/hooks/useFieldFocus';
 import { useMultiSelectFieldDisplay } from '@/object-record/record-field/ui/meta-types/hooks/useMultiSelectFieldDisplay';
-import { MultiSelectDisplay } from '@/ui/field/display/components/MultiSelectDisplay';
 import { ExpandableList } from '@/ui/layout/expandable-list/components/ExpandableList';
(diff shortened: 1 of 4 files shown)


charlesBochet:
@victorjzq Please stop, or I'll be force to ban you, you are not helping

charlesBochet:
Closing this, we are going to revisit focus and active states in upcoming weeks

PR #18783, opened 2026-03-19
fix: skip orphaned roleTargets during cache recomputation instead of throwing

victorjzq:
## Problem

Closes #18694

On instances that upgraded from pre-v1.13 versions (especially those that skipped intermediate versions), the `makeRoleUniversalIdentifierAndApplicationIdNotNullable` migration silently swallowed failures, leaving `role` and `roleTarget` rows with `null` `universalIdentifier`.

When any view mutation is performed, it triggers cache invalidation including `flatRoleTargetMaps` and `flatRoleTargetByAgentIdMaps`. These call `fromRoleTargetEntityToFlatRoleTarget`, which throws `FlatEntityMapsException` when a roleTarget's role or application has a missing `universalIdentifier`. This causes:
- Error log: "Role not found for flat role target" on every view mutation
- Brief UI hang waiting for cache invalidation

## Fix

Pre-filter roleTargets in both `computeForCache` implementations before passing them to `fromRoleTargetEntityToFlatRoleTarget`. RoleTargets whose role or application `universalIdentifier` is missing are skipped with a `WARN` log instead of throwing.

**Why this approach:**
- The strict validation in `fromRoleTargetEntityToFlatRoleTarget` is preserved as a programming-error safeguard (wrong map passed, etc.)
- Pre-filtering at the `computeForCache` layer is the correct place to handle data integrity issues from incomplete upgrades
- The `FixInvalidStandardUniversalIdentifiersCommand` in v1.19 upgrade path is the authoritative fix for the data; this is a resilience improvement for instances that haven't run it yet

**Files changed:**
- `workspace-flat-role-target-map-cache.service.ts` — skip orphaned roleTargets with warn log
- `workspace-flat-role-target-by-agent-id.service.ts` — same fix for agent-specific cache

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/display/components/MultiSelectFieldDisplay.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/display/components/MultiSelectFieldDisplay.tsx
@@ -1,14 +1,10 @@
-import { useFieldFocus } from '@/object-record/record-field/ui/hooks/useFieldFocus';
 import { useMultiSelectFieldDisplay } from '@/object-record/record-field/ui/meta-types/hooks/useMultiSelectFieldDisplay';
-import { MultiSelectDisplay } from '@/ui/field/display/components/MultiSelectDisplay';
 import { ExpandableList } from '@/ui/layout/expandable-list/components/ExpandableList';
 import { Tag } from 'twenty-ui/components';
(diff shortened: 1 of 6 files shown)


charlesBochet:
@victorjzq Last warning

charlesBochet:
@victorjzq No, we should make another command in 1-20 to backfill all those and make a constraint to make sure it's universalIdentifiers are non null

PR #18789, opened 2026-03-20
fix: insert-before folder for workspace DnD and add-to-nav drags

### LINKED ISSUE #2294: Can’t drop nav item from below folder to “just above” the folder
When reordering items in the workspace nav sidebar, you can't reliably drop an item that is **below** a folder so it ends up **just above** that folder. The "above folder" drop target either doesn't appear or the item is placed in the wrong position (e.g. below the folder or in the wrong slot). This only affects moving items **up** in the list to sit directly above a folder; moving items down or dropping onto/into the folder may still work.

## Steps to reproduce
1. Turn on **edit mode** for the nav sidebar (e.g. via the wrench/edit control or the relevant action).
2. In the Workspace section, ensure there is at least one **folder** with at least one **nav item below it** (e.g. order: … folder, then item A, then item B).
3. Start dragging the item that is **below** the folder (e.g. item A or B).
4. Drag it **up** toward the folder and try to drop it **just above** the folder (so it sits between the item above the folder and the folder).
5. Observe: the blue "drop above" line either doesn't appear above the folder, or it appears but the item is dropped in the wrong place (e.g. below the folder or not in the intended slot).
6. Optionally, compare with dragging an item that is **above** the folder **down** to just above the folder, or dropping onto/into the folder, to confirm those cases still behave as expected.

abdulrahmancodes:
Closes [#2294](https://github.com/twentyhq/core-team-issues/issues/2294)

https://github.com/user-attachments/assets/f7e0da72-dd08-46c4-92d7-be0453aaadd9


etiennejouan:
## Work in progress 🚧 

## Original PR Approach (replaced)

The PR solved this by intercepting `onDragMove` (fires on every pixel of mouse movement), manually reading DOM bounding rects, and remapping destinations with ~500 lines of pointer-math logic across 6 new utility files. This worked but was:

- **Expensive at runtime** — `getBoundingClientRect()` + `findIndex` scans on every mouse move
- **Complex** — two near-identical remap functions for hover vs drop, `preClearActiveDropTargetId` state hack
- **Leaking concerns** — DnD hook imported object permissions, views, and metadata

## New Approach: Let dnd-kit do the work

Instead of fighting dnd-kit's collision system, we give it the right droppable targets and let it handle everything natively.

### 1. Thin invisible droppable overlay on folder headers

A new `NavigationMenuItemInsertBeforeDroppableZone` component — a 10px absolutely-positioned overlay at the top of each folder header. It only renders when `isDragging` is true. Its droppable data points to the orphan list at the folder's index, with `insertBeforeItemId` set to the folder's ID.

### 2. `pointerIntersection` collision detector everywhere

The default dnd-kit collision detector uses the **dragged element's full bounding rect**, which meant the tall dragged item always overlapped the thin 10px zone. We switched all droppable slots to use `pointerIntersection` from `@dnd-kit/collision`, which only checks the **actual pointer position**. This gives precise control — the pointer must physically enter a zone to activate it.

### 3. Priority-based zone selection

With both the folder header (priority 3) and the insert-before zone (priority 5) using `pointerIntersection`:

- **Pointer in top 10px** → both detect a collision → insert-before wins (5 > 3)
- **Pointer below 10px** → only folder header detects → drop into folder

No manual pointer-Y math needed — dnd-kit resolves this from the droppable geometry and priorities.

etiennejouan (approved):

PR #18791, opened 2026-03-20
Fix orphan navigation menu items for deleted views

channi23:
## Summary

Fixes #18757

This fixes a set of Favorites / navigation-menu-item integrity problems related to deleted views, stale hidden items, and upgraded workspaces with orphaned navigation items.

## What changed

- delete `navigationMenuItem` entries when their favorited view is deleted
- keep the client metadata store in sync immediately when a view is deleted
- determine whether a view is already favorited from visible valid navigation items instead of raw stale items
- add a `1.20.0` upgrade repair command that deletes orphan navigation menu items and normalizes positions
- add regression coverage for deletion of both record-based and view-based navigation menu items

## Details

Server:
- extend `NavigationMenuItemDeletionService` so cleanup applies to deleted views as well as deleted records
- add regression tests covering record-based deletion, view-based deletion, and no-op behavior
- add `DeleteOrphanNavigationMenuItemsCommand` to remove orphaned items pointing to:
  - deleted views
  - deleted records
  - missing folders
- normalize positions per scope (`userWorkspaceId + folderId`) after repair
- wire the new repair command into the `1.20.0` upgrade flow

Frontend:
- add `useRemoveNavigationMenuItemByViewId`
- remove the related navigation item from client metadata immediately when deleting a view
- use sorted / visible navigation items for favorite detection so stale hidden rows do not block re-adding a favorite

## Why

Issue `#18757` reports mismatches between Favorites shown in the UI and rows users can still find in the database. We found that current Favorites behavior is driven by `navigationMenuItem`, not the legacy `favorite` table, and that stale / orphaned `navigationMenuItem` rows could:
- remain after deleting a favorited view
- stay hidden from the UI if they point to invalid targets
- still cause the UI to think a view was already favorited
- persist in workspaces with migration damage from skipped sequential upgrades

This patch addresses those cases directly and adds an upgrade-time repair path for older corrupted workspaces.

## Validation

Passed:
- `./node_modules/.bin/jest --config packages/twenty-server/jest.config.mjs --runInBand packages/twenty-server/src/engine/metadata-modules/navigation-menu-item/services/__tests__/navigation-menu-item-deletion.service.spec.ts`
- `./node_modules/.bin/tsc -p packages/twenty-front/tsconfig.json --noEmit --pretty false`

Known unrelated existing failure:
- `./node_modules/.bin/tsc -p packages/twenty-server/tsconfig.json --noEmit --pretty false`

The server typecheck failure is pre-existing and unrelated to this branch. Current errors are around `@file-type/pdf` module resolution and `is-psl-parsed-domain.type.ts`.


packages/twenty-docs/scripts/generate-documentation-paths.ts
```diff
@@ -97,5 +97,5 @@ export type DocumentationPath =
 
 fs.writeFileSync(path.join(outputDir, 'DocumentationPaths.ts'), pathsContent);
 
-console.log(`Generated documentation constants to ${outputDir}`);
-console.log(`Total paths: ${sortedPaths.length}`);
+process.stdout.write(`Generated documentation constants to ${outputDir}\n`);
```
charlesBochet: remove

packages/twenty-front/src/modules/command-menu/components/__stories__/CommandMenu.stories.tsx
```diff
@@ -294,3 +296,35 @@ export const SubPageNavigation: Story = {
     },
   ],
 };
+
+export const PinnedActionTooltip: Story = {
```
charlesBochet: What is this test and how is it related to the issue?

packages/twenty-front/src/modules/command-menu/components/CommandMenuButton.tsx
```diff
@@ -59,7 +59,11 @@ export const CommandMenuButton = ({
           ariaLabel={resolvedLabel}
         />
       ) : (
-        <div id={`command-menu-item-entry-${command.key}`} key={command.key}>
+        <div
+          id={`command-menu-item-entry-${command.key}`}
+          key={command.key}
+          data-testid="tooltip"
```
charlesBochet: same

packages/twenty-front/src/modules/navigation-menu-item/edit/hooks/useRemoveNavigationMenuItemByViewId.ts
```diff
@@ -0,0 +1,55 @@
...
+import { metadataStoreState } from '@/metadata-store/states/metadataStoreState';
+import { useApolloCoreClient } from '@/object-metadata/hooks/useApolloCoreClient';
+import {
+  FindManyNavigationMenuItemsDocument,
+  type NavigationMenuItem,
+} from '~/generated-metadata/graphql';
+import { isDefined } from 'twenty-shared/utils';
+import { useStore } from 'jotai';
+
+export const useRemoveNavigationMenuItemByViewId = () => {
```
charlesBochet: this is not a bad idea, let's not update apollo cache though. metadata queries should be network only (let's not do this change now and see if just remove the cache logic is enough)

let's call the hook: useOptimisticRemoveNavigationMenuItemsByViewId

packages/twenty-front/src/modules/views/view-picker/hooks/useDestroyViewFromCurrentState.ts
```diff
@@ -67,16 +70,22 @@ export const useDestroyViewFromCurrentState = (viewBarInstanceId?: string) => {
       return;
     }
 
+    if (!viewPickerReferenceViewId) {
```
charlesBochet: why this change now? I'm surprised that we need it / it was not broken before (as I see a performViewAPIDestroy below)

packages/twenty-server/src/database/commands/upgrade-version-command/1-20/1-20-delete-orphan-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,253 @@
...
+    workspaceId,
+    options,
+  }: RunOnWorkspaceArgs): Promise<void> {
+    const { flatViewMaps, flatObjectMetadataMaps } =
+      await this.workspaceCacheService.getOrRecompute(workspaceId, [
+        'flatViewMaps',
+        'flatObjectMetadataMaps',
+      ]);
+
+    const navigationMenuItems = await this.navigationMenuItemRepository.find({
```
charlesBochet: use flatNavigationMenuItemMaps from the workspaceCache too

packages/twenty-server/src/database/commands/upgrade-version-command/1-20/1-20-delete-orphan-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,253 @@
...
+    const activeViewIds = new Set(
+      Object.values(flatViewMaps.byUniversalIdentifier)
+        .filter((view): view is NonNullable<typeof view> => isDefined(view))
+        .filter((view) => view.deletedAt === null)
+        .map((view) => view.id),
+    );
+
+    const folderIds = new Set(
+      navigationMenuItems
+        .filter(
```
charlesBochet: no you should be using navigationMenuItem type here

packages/twenty-server/src/database/commands/upgrade-version-command/1-20/1-20-delete-orphan-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,253 @@
...
+      .map((item) => item.id);
+
+    const orphanRecordItemIds =
+      await this.findOrphanRecordNavigationMenuItemIds(
+        workspaceId,
+        navigationMenuItems,
+        flatObjectMetadataMaps,
+      );
+
+    const orphanNavigationMenuItemIds = Array.from(
```
charlesBochet: I'm not sure we should delete orphan menu items, that's fine to have empty folders I believe

packages/twenty-server/src/database/commands/upgrade-version-command/1-20/1-20-delete-orphan-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,253 @@
...
+      const existingRecordIds =
+        recordIdsByObjectMetadataId.get(item.targetObjectMetadataId) ?? [];
+      existingRecordIds.push(item.targetRecordId);
+      recordIdsByObjectMetadataId.set(
+        item.targetObjectMetadataId,
+        existingRecordIds,
+      );
+    }
+
+    for (const [
```
charlesBochet: this code is really not great peformance wise (for loop that is getting repository)

That's being said I don't think we have to worry about orphans (see comment above)

packages/twenty-server/src/database/commands/upgrade-version-command/1-20/1-20-delete-orphan-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,253 @@
...
+        ) {
+          navigationMenuItemIdsToDelete.push(item.id);
+        }
+      }
+    }
+
+    return navigationMenuItemIdsToDelete;
+  }
+
+  private async normalizeNavigationMenuItemPositions(
```
charlesBochet: no need to normalize

packages/twenty-server/src/engine/metadata-modules/navigation-menu-item/services/__tests__/navigation-menu-item-deletion.service.spec.ts
```diff
@@ -0,0 +1,129 @@
+import { NavigationMenuItemDeletionService } from 'src/engine/metadata-modules/navigation-menu-item/services/navigation-menu-item-deletion.service';
+
+describe('NavigationMenuItemDeletionService', () => {
```
charlesBochet: we don't write spec service anymore, they are not useful

packages/twenty-server/src/engine/metadata-modules/navigation-menu-item/services/navigation-menu-item-deletion.service.ts
```diff
@@ -40,9 +40,10 @@ export class NavigationMenuItemDeletionService {
     ).filter(
       (item): item is NonNullable<typeof item> =>
         isDefined(item) &&
-        isDefined(item.targetRecordId) &&
-        !isDefined(item.viewId) &&
-        deletedRecordIdsSet.has(item.targetRecordId),
+        ((isDefined(item.targetRecordId) &&
```
charlesBochet: this logic is too complex for a human to read, make an util or a private function that write clear logic

charlesBochet (changes requested):
Hi @channi23, thanks for helping, the overall approach is good, I've left comments

Can you take them, and QA intensively to make sure your fixes are working?

packages/twenty-front/src/modules/command-menu/components/__stories__/CommandMenu.stories.tsx
channi23: you're right to question this, this test and the tooltip-related changes are not directly related to the original issue.

i added them while trying to address a ci failure coming from a storybook interaction test (`expect(element).toBeVisible()` failing on a tooltip). my initial assumption was that the failure was due to the tooltip not being properly rendered or detectable in the test environment, so i introduced the `data-testid` and an additional test to stabilize it.

looking back, this was not the correct approach since it mixes unrelated concerns into this pr. i will remove these changes and instead address the ci issue separately in a more targeted way.

anyway i will remove these changes and handle the ci issue separately in a focused way

channi23:
> Hi @channi23, thanks for helping, the overall approach is good, I've left comments
> 
> Can you take them, and QA intensively to make sure your fixes are working?

hey @charlesBochet , i actually spent some time on this earlier but i didn’t properly check the review comments at that point. looks like i missed the timing on that opportunity.

looks like you’ve already started addressing these changes.

i was planning to go through the comments and fix them myself to better understand the context and the product, felt like a good chance to level up.

anyway, really appreciate the feedback and insights here, super helpful.

let me know if there’s anything i can pick up or help with, i’m around and ready to jump in.

charlesBochet:
@channi23 I'm doing it actually, you can learn from the changes :) Overall PR is sound, some details need to be a bit sharper

charlesBochet (approved):
PR ready to merge, thanks a lot @channi23

packages/twenty-server/src/database/commands/upgrade-version-command/1-20/1-20-delete-orphan-navigation-menu-items.command.ts
```diff
@@ -0,0 +1,91 @@
...
+    protected readonly twentyORMGlobalManager: GlobalWorkspaceOrmManager,
+    protected readonly dataSourceService: DataSourceService,
+    @InjectRepository(NavigationMenuItemEntity)
+    private readonly navigationMenuItemRepository: Repository<NavigationMenuItemEntity>,
+    private readonly workspaceCacheService: WorkspaceCacheService,
+  ) {
+    super(workspaceRepository, twentyORMGlobalManager, dataSourceService);
+  }
+
+  override async runOnWorkspace({
```
prastoin, line 38: We should add a FK to the navigation menu item entity directly to avoid this to re-occurs

prastoin: Also might directly rely on the FK on delete cascade instead of programmatically define this business logic

channi23: There is already a FK with ON DELETE CASCADE on navigationMenuItem.viewId, but views are soft-deleted (deletedAt), so the cascade does not fire on that path.

prastoin: Indeed my bad, might still wanna add a custom constraint to avoid such data setup to re-occur in the future
Or we might wanna directly deprecate the view soft deletion as discussed in the past

prastoin: The side effect implementation introduced in this PR is idiomatic in the end 👍

PR #18798, opened 2026-03-20
fix: add missing LEFT JOIN for relation fields in groupBy orderByForRecords

charlesBochet:
## Summary

Fixes "missing FROM-clause entry for table" SQL error when using `orderByForRecords` with a relation field (e.g. `{ company: { name: "AscNullsFirst" } }`) in a `groupBy` query.

### Root cause

This is a regression from #18005 (Feb 17). That PR correctly removed a duplicate `applyOrderToBuilder()` call on the groupBy subquery (which was conflicting with the `ROW_NUMBER() OVER (... ORDER BY ...)` window function), but in doing so it also removed the LEFT JOINs that `applyOrderToBuilder` was adding for relation fields.

After that change, ordering relied solely on `getOrderByRawSQL()` inside `applyPartitionByToBuilder()`, which builds raw SQL for the window function but never added the required JOINs. Scalar field ordering (e.g. `name`, `position`) kept working since those don't need JOINs, but relation field ordering (e.g. `company.name`) broke.

### Fix

- `getOrderByRawSQL` now returns `relationJoins` alongside the SQL string so callers get the join info they need
- `applyPartitionByToBuilder` in `GroupByWithRecordsService` adds the required LEFT JOINs before building the `ROW_NUMBER()` window function, mirroring the pattern used by `applyOrderToBuilder` in the `findMany` path

## Test plan

- [x] Manually tested locally with a GraphQL query matching the customer's failing pattern (`orderByForRecords: [{ company: { name: "AscNullsFirst" } }]`)
- [x] Added integration tests for ascending and descending relation field ordering in `group-by-with-records-resolver.integration-spec.ts`
- [x] Typecheck passes
- [x] Lint passes
- [x] CI green

### DIFF
--- a/packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query.parser.ts
+++ b/packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query.parser.ts
@@ -13,6 +13,7 @@ import { GraphqlQueryOrderGroupByParser } from 'src/engine/api/graphql/graphql-q
 import {
   GraphqlQueryOrderFieldParser,
   type OrderByClause,
+  type RelationJoinInfo,
 } from 'src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-order/graphql-query-order.parser';
 import {
   GraphqlQuerySelectedFieldsParser,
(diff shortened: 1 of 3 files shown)


FelixMalfait:

Your preview environment is available at: http://bore.pub:54026

thomtrp (approved):

PR #18799, opened 2026-03-20
fix: update workspace members state on profile onboarding completion

charlesBochet:
## Summary

Fixes #18610

After completing the CreateProfile onboarding step, `currentWorkspaceMembersState` (plural, used by `useActorFieldDisplay` to resolve "Created by" names) was never updated with the new name — only `currentWorkspaceMemberState` (singular) was. This caused "Created by" fields to display "Untitled" for the remainder of the session.

**Root cause analysis:**

The issue reporter attributed the bug to empty `core.user.firstName/lastName`, but `createdBy` display doesn't use `core.user` at all. The actual flow:

1. Sign-up creates workspace member with empty names (copied from empty `core.user`)
2. `GetCurrentUserDocument` loads `currentWorkspaceMembersState` with empty names
3. CreateProfile step updates workspace member in DB + `currentWorkspaceMemberState` (singular) ✓
4. `currentWorkspaceMembersState` (plural) is **never refreshed** — the query is skipped once `currentUser` exists
5. `useActorFieldDisplay` looks up from the stale plural state → empty name → "Untitled"

**Fix:** Update `currentWorkspaceMembersState` alongside `currentWorkspaceMemberState` in the CreateProfile submit handler.

### DIFF
--- a/packages/twenty-front/src/pages/onboarding/CreateProfile.tsx
+++ b/packages/twenty-front/src/pages/onboarding/CreateProfile.tsx
@@ -11,6 +11,7 @@ import { SubTitle } from '@/auth/components/SubTitle';
 import { Title } from '@/auth/components/Title';
 import { currentUserState } from '@/auth/states/currentUserState';
 import { currentWorkspaceMemberState } from '@/auth/states/currentWorkspaceMemberState';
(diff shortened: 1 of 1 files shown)


FelixMalfait:

Your preview environment is available at: http://bore.pub:2549

thomtrp (approved):

PR #18802, opened 2026-03-20
Fix page scrolling when layout customization bar is visible

abdulrahmancodes:


etiennejouan (approved):
The drawer inherits its height from the flex layout in DefaultLayout, which already correctly distributes the remaining space after the customization bar

PR #18809, opened 2026-03-20
fix: prevent localStorage bloat from derived fields on mock metadata

FelixMalfait:
## Summary

- On unauthenticated pages (login), mocked object metadata was being hydrated into the store with `readableFields` and `updatableFields` attached. These derived arrays duplicate every field per object, inflating `objectMetadataItems` in localStorage from ~70 KB to ~2 MB. Combined with other metadata keys, this exceeded Safari's 5 MB quota and caused `QuotaExceededError`, preventing real data from replacing mock data on login.
- Extracted flat mock data into a dedicated file (`generatedMockObjectMetadataItemsWithRelated.ts`) and use it for store hydration, keeping the enriched version (`generatedMockObjectMetadataItems`) only for tests.
- Completed `splitObjectMetadataItemWithRelated`'s contract by stripping `readableFields` and `updatableFields` at runtime (not just via TypeScript's `Omit`), matching the `FlatObjectMetadataItem` return type that omits all four relational properties.

## Root cause

1. `MinimalMetadataLoadEffect` loads mocked metadata on unauthenticated pages.
2. `generatedMockObjectMetadataItems` was produced by `enrichObjectMetadataItemsWithPermissions`, which attaches `readableFields` and `updatableFields` (full copies of the `fields` array).
3. `splitObjectMetadataItemWithRelated` only destructured `fields` and `indexMetadatas` — `readableFields`/`updatableFields` leaked into `...objectProperties` at runtime because `Omit` is a compile-time-only guard.
4. These bloated objects were written to localStorage, consuming ~2 MB instead of ~70 KB.
5. On login, the intermediate state (old composite mock `current` + new flat real `draft`) exceeded the 5 MB quota, causing a `QuotaExceededError` deadlock.

## Test plan

- [ ] Open the app in a fresh Safari private window (empty localStorage)
- [ ] Verify the login page loads without errors
- [ ] Log in and verify metadata loads correctly
- [ ] Check `localStorage` size — `metadataStoreState__objectMetadataItems` should be ~70 KB, not ~2 MB
- [ ] Run existing tests: `npx nx test twenty-front` — no regressions from the mock data split


FelixMalfait:

Your preview environment is available at: http://bore.pub:60699

charlesBochet (approved):

PR #18813, opened 2026-03-20
fix: resolve "Created By" showing "Untitled" for workspace members

oniani1:
## Summary

Fixes #18194 — The "Created By" column shows "Untitled" instead of the user's name when creating records manually.

**Root causes found:**

1. **`useLoadCurrentUser.ts`** — The `deletedWorkspaceMembers` data is fetched via GraphQL (the query fragment includes it) but never stored in `currentWorkspaceDeletedMembersState`. While `UserMetadataProviderInitialEffect` correctly stores it on initial load, `useLoadCurrentUser` (called from 8 places: sign-in, settings changes, etc.) doesn't. This means records created by deleted members can't resolve their name and fall back to the stored `fieldValue.name`, which may be empty → "Untitled".

2. **`buildCreatedByFromFullNameMetadata`** — Produces `" "` (single space) when both `firstName` and `lastName` are empty strings (the default for users who sign up with email only, per `sign-in-up.service.ts`). This stores a whitespace-only name in the database.

3. **`useActorFieldDisplay`** — When constructing the display name from a found workspace member, `` `${name.firstName} ${name.lastName}` `` also produces `" "` for empty-name users, rendering as visual blank space.

## Fix

- **`useLoadCurrentUser.ts`:** Store `deletedWorkspaceMembers` in state, matching the pattern already used in `UserMetadataProviderInitialEffect.tsx` (lines 156-158).
- **`buildCreatedByFromFullNameMetadata`:** Trim the constructed name and fall back to `"Anonymous"` when empty — prevents whitespace-only names from being stored in the database.
- **`useActorFieldDisplay`:** Trim the constructed name and fall back to `fieldValue.name` when the workspace member has an empty name — prevents blank display.

## Test plan

- [x] Backend typecheck passes
- [x] oxlint passes on all changed files (0 warnings, 0 errors)
- [x] Prettier passes on all changed files
- [ ] Manual: Create a record with a user that has no first/last name set → "Created By" should show "Anonymous" instead of "Untitled"
- [ ] Manual: Delete a workspace member, then view records they created → "Created By" should still show their name

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/hooks/useActorFieldDisplay.ts
+++ b/packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/hooks/useActorFieldDisplay.ts
@@ -45,9 +45,10 @@ export const useActorFieldDisplay = (): ActorFieldDisplayValue | undefined => {
   }
 
   const { name, avatarUrl } = relatedWorkspaceMember;
+  const fullName = `${name.firstName} ${name.lastName}`.trim();
   return {
     fieldValue,
-    name: `${name.firstName} ${name.lastName}`,
+    name: fullName || fieldValue.name,
     avatarUrl: avatarUrl,
   };
 };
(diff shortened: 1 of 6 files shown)


charlesBochet:
This is already fixed on main, likely AI generated, closing :)

PR #18821, opened 2026-03-21
fix: trigger data fetch at all scroll speeds in virtualized table

### LINKED ISSUE #18819: [Bug] Trigger to load new entries to late
The trigger to load new entries when scrolling down is to late.

We have the case, that we scroll down slowly and see already the placeholder but they are not replaced by real data.
Only when we scroll down further, you see in the network trace the new call for new data. 
My expectation would be, that new data is already loaded when I start scrolling down to see in the best place no placeholder and have a smoother experience

Image

oniani1:
Fixes #18819

When scrolling the record table at moderate speed (roughly 50-200 rows/s), data fetching never triggers. The scroll speed thresholds that control low-details rendering mode were also gating `triggerFetchPages()`, creating a dead zone where neither the fast-scroll nor slow-scroll branch would call it.

Moved `triggerFetchPages()` outside the speed threshold branches so it always fires on scroll. The speed thresholds now only control the visual rendering mode (skeletons vs full cells), which was the original intent.

This is safe because `triggerFetchPages` is already debounced (25ms, maxWait 2s) and `triggerFetchPagesWithoutDebounce` early-returns when low-details mode is active during fast scrolling.

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-table/virtualization/components/RecordTableVirtualizedRowTreadmillEffect.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-table/virtualization/components/RecordTableVirtualizedRowTreadmillEffect.tsx
@@ -195,12 +195,10 @@ export const RecordTableVirtualizedRowTreadmillEffect = () => {
           SCROLL_SPEED_THRESHOLD_IN_ROWS_PER_SECOND_TO_DEACTIVATE_LOW_DETAILS
         ) {
           deactivateLowDetailsDebounced();
-
-          triggerFetchPages();
         }
-      } else {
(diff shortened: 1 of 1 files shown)


charlesBochet:
@oniani1 The issue is discussing Kanban board, please stop opening PRs without testing them

charlesBochet:
Can't reproduce the issue on main, I'm closing this PR
@oniani1 please stop opening PR without taking time to understand product / issues, this is actually quite disrespectful. You are not helping at all

oniani1:
You're right, I apologize. I should have reproduced the issue first instead of just reading the code. Won't happen again.

PR #18824, opened 2026-03-21
fix: board view loads all records instead of showing skeleton placeholders

charlesBochet:
## Summary

- The record board (Kanban view) only loaded the first 10 records per column, then showed skeleton placeholder cards for the rest without ever fetching the remaining data
- **Root cause**: The `RecordBoardFetchMoreInViewTriggerComponent` (IntersectionObserver) is positioned at the bottom of the entire board. With 10 real cards + 10 skeleton cards per column (~3500px of content), the trigger div was pushed far beyond the 1600px `rootMargin` detection zone, so `recordBoardShouldFetchMore` never became `true` and fetch-more was never triggered
- **Fix**: The initial query now sets `recordBoardShouldFetchMore = true` when columns need more data, and `triggerRecordBoardFetchMore` uses an internal `while` loop to load all remaining pages in a single invocation — making it immune to the InView component racing to reset the flag between React render cycles

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-board/components/RecordBoardFetchMoreInViewTriggerComponent.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-board/components/RecordBoardFetchMoreInViewTriggerComponent.tsx
@@ -1,16 +1,19 @@
 import { styled } from '@linaria/react';
-import { useEffect } from 'react';
+import { useEffect, useMemo } from 'react';
 import { useInView } from 'react-intersection-observer';
 
 import { RECORD_BOARD_COLUMN_PADDING_AND_BORDER_WIDTH } from '@/object-record/record-board/constants/RecordBoardColumnPaddingAndBorderWidth';
 
 import { RECORD_BOARD_COLUMN_WIDTH } from '@/object-record/record-board/constants/RecordBoardColumnWidth';
+import { RECORD_BOARD_QUERY_PAGE_SIZE } from '@/object-record/record-board/constants/RecordBoardQueryPageSize';
 import { recordBoardIsFetchingMoreComponentState } from '@/object-record/record-board/states/recordBoardIsFetchingMoreComponentState';
 import { recordBoardShouldFetchMoreComponentState } from '@/object-record/record-board/states/recordBoardShouldFetchMoreComponentState';
+import { visibleRecordFieldsComponentSelector } from '@/object-record/record-field/states/visibleRecordFieldsComponentSelector';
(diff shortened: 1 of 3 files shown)


FelixMalfait:

Your preview environment is available at: http://bore.pub:30694

packages/twenty-front/src/modules/object-record/record-board/components/RecordBoardFetchMoreInViewTriggerComponent.tsx
```diff
@@ -32,10 +38,24 @@ export const RecordBoardFetchMoreInViewTriggerComponent = () => {
     recordBoardIsFetchingMoreComponentState,
   );
 
+  const visibleRecordFields = useAtomComponentSelectorValue(
+    visibleRecordFieldsComponentSelector,
+  );
+
+  // Scale detection zone with card height (driven by visible field count)
+  // so the trigger fires before skeleton cards become visible to the user
+  const rootMargin = useMemo(() => {
```
charlesBochet: no need for useMemo

packages/twenty-front/src/modules/object-record/record-board/components/RecordBoardFetchMoreInViewTriggerComponent.tsx
```diff
@@ -20,6 +23,9 @@ const StyledFetchMoreTriggerDiv = styled.div<{ width: number }>`
   min-width: ${({ width }) => width}px;
 `;
 
+const ESTIMATED_BOARD_CARD_HEIGHT_BASE = 60;
```
charlesBochet: let's avoid guessing but using constants / computation that should be consistent between css and this

packages/twenty-front/src/modules/object-record/record-board/hooks/useTriggerRecordBoardFetchMore.ts
```diff
@@ -97,6 +103,7 @@ export const useTriggerRecordBoardFetchMore = () => {
       .filter(isDefined);
 
     if (!isNonEmptyArray(recordGroupValuesThatShouldBeFetched)) {
+      store.set(recordBoardShouldFetchMoreCallbackState, false);
```
charlesBochet, line 106: is this still useful?

charlesBochet: Yes — it prevents a busy loop. When all columns are fully loaded, skeleton cards disappear so the trigger div moves closer to the viewport. `inView` stays `true`, keeping `shouldFetchMore = true`. Without clearing it here, the QueryEffect keeps calling `triggerRecordBoardFetchMore`, which early-exits (no columns need data) and sets `isFetchingMore = false`, which re-triggers the effect — infinite cycle.

Issue #6641, opened 2024-08-16
🧪 Introduction to E2E Playwright testing


BOHEUS:
This issue is a parent for all future PR and issues

# Description
Since more and more issues/requests appear for E2E tests, here are some things which need to be addressed before progressing:
- tests must consist of modules (since most of them will have overlaying path, e.g. each test will log in before accessing any data), this way it'll be easier to fix one thing in one file instead of multiple files
- tests must write logs in such form which is easy to process, understand and clear to find where the possible bug is, on top of that, making screenshots so finding process will be much easier
- tests must replicate as close as possible human behaviour, e.g. clicking on input before typing a word (so input will be focused)
- any variables such as URLs, fixed login data (like when hosting locally or in demo) must be imported from separate file like .env
- tests must be able to run in CI in order to ensure the quality of new features and the product (in CI tests must have the URL of existing instance like demo, not the localhost)
- tests must be versatile so anyone can basically run them after providing proper variables, be used in different environments (locally and in production/demo) and both in normal as well as regression testing
- tests must clear after finishing the test (basic assumption is every single test start from clean environment i.e. right after creating a new workspace) as some leftovers from previous tests may result in failed tests later

**While this shouldn't prevent from creating issues/requests for tests, this must be addressed ASAP so technical debt will be minimised**

Based on all of the things described above (some of them are taken from [here](https://github.com/twentyhq/twenty/discussions/6467) where I described them earlier), here's checklist
```[tasklist]
### TO DO
- [ ] Create structure and modules for each main view of the app as well as API (POM)
- [ ] Create components with locators without CSS or text references
- [ ] Create .env with all needed variables
- [ ] Create playwright.config.ts with all required options
- [ ] Analyse and create a cleaner which can run after each test and before/after all tests
- [ ] ~~Add Github workflow for running tests in CI~~ On hold
- [ ] Add proper documentation so there will be continuity in tests in case someone won't have time to maintain
- [x] Add proper drivers for OS shell in case test will require running shell command for e.g. restarting database
- [ ] Check the scrolling behaviour during tests
- [ ] Add storage states for saving auth data after login
- [ ] Add commands for creating containers for smtp4dev and webhook.site
```

*This checklist ^ may be updated in the future*

Materials:
- documentation for Playwright: https://playwright.dev/docs/intro
- XPath syntax https://www.w3schools.com/xml/xpath_syntax.asp
- blog with Playwright explanations https://playwrightsolutions.com/

[...]

PR #18827, opened 2026-03-22
fix(e2e): initialise missing locators in RecordDetails POM

### LINKED ISSUE #18826: fix(e2e): complete missing locator initialisations in RecordDetails POM
## Context

While exploring the `twenty-e2e-testing` package (related to #6641 and #8109), I noticed that `recordDetails.ts` has private locators declared but none initialised in the constructor, making all its methods non-functional.

## Proposed fix

Initialise all locators in the constructor using stable selectors (`aria-label`, `data-testid`, `role`) found by inspecting the record detail view.

Happy to submit a PR for this if it's still needed.

sharathsarangmath:
## What

Initialises the missing locator assignments in the `RecordDetails` POM 
class. All 13 private locators were declared but never assigned in the 
constructor, making every method in the class non-functional.

## How

- Added `objectType` parameter to constructor for dynamic tab locators
- Initialised stable locators using `aria-label`, `data-testid`, and 
  `getByRole` selectors found by inspecting the live app
- Tab locators use the pattern `tab-${objectType}-tab-${tabName}` 
  (e.g. `tab-person-tab-timeline`, `tab-company-tab-timeline`)
- Added TODO comment for `uploadProfileImageButton` and `addShowPageButton` 
  which have no stable selectors in the current UI

## Testing

- Ran existing test suite locally, all previously passing tests still pass
- No breaking changes

## Related

Closes #18826
Related to #6641

### DIFF
--- a/packages/twenty-e2e-testing/lib/pom/recordDetails.ts
+++ b/packages/twenty-e2e-testing/lib/pom/recordDetails.ts
@@ -2,24 +2,55 @@ import { Locator, Page } from '@playwright/test';
 
 export class RecordDetails {
   // TODO: add missing components in tasks, notes, files, emails, calendar tabs
+  //Initialised in constructor - stable data-testid locators
   private readonly closeRecordButton: Locator;
-  private readonly previousRecordButton: Locator;
-  private readonly nextRecordButton: Locator;
-  private readonly favoriteRecordButton: Locator;
-  private readonly addShowPageButton: Locator;
-  private readonly moreOptionsButton: Locator;
-  private readonly deleteButton: Locator;
-  private readonly uploadProfileImageButton: Locator;
   private readonly timelineTab: Locator;
   private readonly tasksTab: Locator;
   private readonly notesTab: Locator;
   private readonly filesTab: Locator;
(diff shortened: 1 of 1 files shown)


FelixMalfait:
Closing as I believe this isn't used. This was started but @BOHEUS but never properly implemented I think, we should probably clean it up

BOHEUS:
@FelixMalfait front end changed so much that locators are probably very outdated, yes, they need to be updated but first we need to decided what should be done exactly within e2e testing and update accordingly

FelixMalfait:
@BOHEUS we do have e2e tests, but the problem was the way this was introduced which was very horizontal (trying to cover everything at once but not even implementing it), while we try to have a more iterative approach (add working well-designs test one-by-one). So I think we should just clean this up / we should not keep dead code

sharathsarangmath:
Thanks for the clarification. I’ll look at the existing test coverage and propose a focused, end-to-end test for a specific flow.

FelixMalfait:
@sharathsarangmath thanks a lot but we're trying to focus on other priorities right now, we're temporarily holding on on outside contributions as they tend to slow us down on the main goals

PR #18828, opened 2026-03-22
fix: skip chat messages query when thread ID is the unknown-thread default

neo773:
Prevents invalid UUID error on initial mount before a real thread is selected


FelixMalfait:

Your preview environment is available at: http://bore.pub:24631

charlesBochet (approved):

PR #18840, opened 2026-03-23
Fix deploy: make agent modelId nullable, remove stale lockfile entry

charlesBochet:
## Summary

- **Fix migration failure**: The `MigrateModelIdsToCompositeFormat` migration was setting `agent.modelId = NULL` but the column had a `NOT NULL` constraint. The migration now first alters the column to be nullable, drops the default, then sets existing values to NULL so agents fall back to workspace defaults.
- **Make `modelId` nullable across the stack**: Updated the entity, DTOs (create input + output), the model registry service, agent executor, and workflow step cloning to handle `modelId: null` (meaning "use workspace default model").
- **Fix CI lockfile mismatch**: Regenerated `yarn.lock` to remove stale `@ai-sdk/groq` entries that were causing `yarn install --immutable` to fail in CI.

Fixes https://github.com/twentyhq/twenty-infra/actions/runs/23416208969/job/68112080874

PR #18847, opened 2026-03-23
Fix deploy: make agent modelId nullable, remove stale lockfile entry

charlesBochet:
## Summary

- **Fix migration failure**: The `MigrateModelIdsToCompositeFormat` migration was setting `agent.modelId = NULL` but the column had a `NOT NULL` constraint. The migration now first alters the column to be nullable, drops the default, then sets existing values to NULL so agents fall back to workspace defaults.
- **Make `modelId` nullable across the stack**: Updated the entity, DTOs (create input + output), the model registry service, agent executor, and workflow step cloning to handle `modelId: null` (meaning "use workspace default model").
- **Fix CI lockfile mismatch**: Regenerated `yarn.lock` to remove stale `@ai-sdk/groq` entries that caused `yarn install --immutable` to fail in CI.
- **Regenerate GraphQL types and SDK client** to reflect nullable `modelId`.
- **Remove obsolete test**: The "when modelId is missing" failing test is no longer valid since `modelId` is now optional.

Fixes https://github.com/twentyhq/twenty-infra/actions/runs/23416208969/job/68112080874

Supersedes #18840

## Test plan

- [ ] Verify migration runs successfully on a fresh database
- [ ] Verify agents with `modelId = null` correctly fall back to workspace smart model default
- [ ] Verify creating an agent without specifying `modelId` works
- [ ] Verify CI passes (lockfile is in sync)


FelixMalfait:

Your preview environment is available at: http://bore.pub:45121

PR #18852, opened 2026-03-23
Fix twenty app dev image

martmull:
as title


packages/create-twenty-app/src/utils/setup-local-instance.ts
```diff
@@ -70,32 +64,24 @@ export const setupLocalInstance = async (
...
+      console.log(chalk.green(`Server running on '${serverUrl}'`));
+      console.log('');
 
       return { running: true, serverUrl };
     }
 
     await new Promise((resolve) => setTimeout(resolve, 2000));
   }
 
+  console.log('');
```
charlesBochet: not pretty, don't we support line returns?

charlesBochet: not pretty, don't we support line returns?

packages/create-twenty-app/src/utils/cancel-install.ts
```diff
@@ -0,0 +1,5 @@
+import * as fs from 'fs-extra';
+
+export const cancelInstall = async (root: string) => {
+  await fs.remove(root);
```
prastoin: ***Question:*** wondering if this is really worth it, could leave the user handle the relicas himself
Could we either check before what would be failing before generating files instead ?

packages/create-twenty-app/src/create-app.command.ts
```diff
@@ -64,11 +65,13 @@ export class CreateAppCommand {
...
 
       if (!options.skipLocalInstance) {
-        // Auto-detect a running server first
         localResult = await setupLocalInstance(appDirectory);
 
         if (localResult.running && localResult.serverUrl) {
           await this.connectToLocal(appDirectory, localResult.serverUrl);
+        } else {
+          await cancelInstall(appDirectory);
+          process.exit(1);
```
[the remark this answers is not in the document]
prastoin: agreed

packages/twenty-sdk/src/cli/commands/server.ts
```diff
@@ -76,6 +87,8 @@ export const registerServerCommands = (program: Command): void => {
         return;
       }
 
+      checkDockerRunning();
```
prastoin: Shouldn't this be optional ? What if we're deploying on a prod env or and using the cli

packages/create-twenty-app/src/utils/cancel-install.ts
martmull: - we need files to check for server because we use twenty-sdk which is a yarn package
- i would prefer to keep the state clean instead of letting the user managing the cleanup

packages/twenty-sdk/src/cli/commands/server.ts
martmull: indeed, checking

packages/create-twenty-app/src/utils/setup-local-instance.ts
```diff
@@ -24,66 +23,73 @@ const isServerReady = async (port: number): Promise<boolean> => {
...
   serverUrl?: string;
 };
 
 export const setupLocalInstance = async (
   appDirectory: string,
+  preferredPort?: number,
 ): Promise<LocalInstanceResult> => {
-  console.log('');
-  console.log(chalk.blue('Setting up local Twenty instance...'));
+  const detectedPort = await detectRunningServer(preferredPort);
```
prastoin, line 49: Duplicated of `detectLocalServer` ?

packages/twenty-sdk/src/cli/commands/server.ts
```diff
@@ -134,19 +142,21 @@ export const registerServerCommands = (program: Command): void => {
           ],
           { stdio: 'inherit' },
         );
+        console.log('');
```
prastoin: Note: we could totally replace the .github/actions/spawn-twenty-docker-image/action.yaml by this comand usage

prastoin (approved):

PR #18853, opened 2026-03-23
fix: run AI catalog sync as standalone script to avoid DB dependency

FelixMalfait:
## Summary
- The `ci-ai-catalog-sync` cron workflow was failing because the `ai:sync-models-dev` NestJS command bootstraps the full app, which tries to connect to PostgreSQL — unavailable in CI.
- Converted the sync logic to a standalone `ts-node` script (`scripts/ai-sync-models-dev.ts`) that runs without NestJS, eliminating the database dependency.
- Removed the `Build twenty-server` step from the workflow since it's no longer needed, making the job faster.

## Test plan
- [x] Verified the standalone script runs successfully locally via `npx nx run twenty-server:ts-node-no-deps-transpile-only -- ./scripts/ai-sync-models-dev.ts`
- [x] Verified `--dry-run` flag works correctly
- [x] Verified the output `ai-providers.json` is correctly written with valid JSON (135 models across 5 providers)
- [x] Verified the script passes linting with zero errors
- [ ] CI should pass without requiring a database service

Fixes: https://github.com/twentyhq/twenty/actions/runs/23424202182/job/68135439740


FelixMalfait:

Your preview environment is available at: http://bore.pub:52646

PR #18857, opened 2026-03-23
fix: add security headers to file serving endpoints to prevent stored XSS

FelixMalfait:
## Summary

- File serving endpoints (`GET file/:fileFolder/:id` and `GET public-assets/...`) were piping S3/local file streams directly to the response without any HTTP headers, allowing a stored XSS attack via uploaded HTML files rendered inline on the CRM origin.
- Adds `Content-Type`, `Content-Disposition`, and `X-Content-Type-Options: nosniff` headers to all file serving responses. Only known-safe MIME types (images, PDF, plain text, audio, video) are served inline; everything else (HTML, SVG, XML, etc.) forces `Content-Disposition: attachment` to trigger download instead of rendering.
- New `setFileResponseHeaders` utility with an explicit allowlist of inline-safe MIME types.

## Test plan

- [x] Unit tests pass (9 tests including 2 new ones: header assertions and attachment-disposition for HTML)
- [x] Lint clean (`lint:diff-with-main`)
- [x] Typecheck clean (`nx typecheck twenty-server`)
- [ ] Manual: upload an HTML file via `uploadWorkflowFile`, access the returned URL — should download instead of rendering
- [ ] Manual: upload a PNG image, access the URL — should render inline with correct `Content-Type: image/png`
- [ ] Manual: verify `X-Content-Type-Options: nosniff` header is present on all file responses

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/file/services/file.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/file/services/file.service.ts
@@ -1,11 +1,9 @@
 import { Injectable } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
 
-import { extname } from 'path';
 import { type Readable } from 'stream';
 
 import { isNonEmptyString } from '@sniptt/guards';
-import { lookup } from 'mrmime';
 import { FileFolder } from 'twenty-shared/types';
 import {
   buildSignedPath,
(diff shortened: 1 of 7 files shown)


FelixMalfait:

Your preview environment is available at: http://bore.pub:34841

packages/twenty-server/src/engine/core-modules/file/services/file.service.ts
```diff
@@ -45,20 +47,26 @@ export class FileService {
...
+    const stream = await this.fileStorageService.readFile({
       resourcePath: filepath,
       fileFolder,
       applicationUniversalIdentifier: application.universalIdentifier,
       workspaceId,
     });
+
+    const ext = extname(filepath).slice(1).toLowerCase();
+    const mimeType =
+      (ext ? lookup(ext) : undefined) ?? 'application/octet-stream';
```
etiennejouan: mimeType already available in file entity. it should be fetch from there (file repository) instead of re-computed. @claude can you handle this ? thanks

etiennejouan (approved):

PR #18864, opened 2026-03-23
feat: add S3 presigned URL redirect for file downloads

FelixMalfait:
## Summary

- When `STORAGE_S3_PRESIGNED_URL_BASE` is configured, the file controller returns a **302 redirect** to a presigned S3 URL instead of proxying every byte through the server. This eliminates server bandwidth and CPU overhead for S3-backed deployments.
- For local storage or S3 without a public endpoint, behavior is unchanged (stream + pipe with security headers).
- Added `getPresignedUrl` to the `StorageDriver` interface (required method returning `string | null`), with implementations in S3Driver (uses a separate presign client with the public endpoint), LocalDriver (returns `null`), and ValidatedStorageDriver (path traversal protection + delegation).
- Added a unified `getFileResponseById` method in `FileService` that performs a single DB lookup and returns either a redirect URL or a stream, avoiding double lookups.
- Extracted `getContentDisposition` from the header util so both the proxy path and presigned URL path share the same inline/attachment allowlist.
- Added MinIO service to `docker-compose.dev.yml` (optional `s3` profile) for local S3 testing.
- Documented S3 presigned URL setup, CORS, and `nosniff` requirements in the self-hosting docs.

## Test plan

- [x] All 63 unit tests pass across 5 test suites (util, S3 driver, validated driver, file storage service, controller)
- [x] `npx nx typecheck twenty-server` passes
- [ ] Manual E2E test with MinIO: `docker compose --profile s3 up -d`, configure S3 env vars, verify `curl -I` returns 302 with `Location` header pointing to MinIO
- [ ] Verify local storage (no `STORAGE_S3_PRESIGNED_URL_BASE`) still streams files with 200 + security headers
- [ ] Verify public assets endpoint still proxies (no redirect)


FelixMalfait:

Your preview environment is available at: http://bore.pub:46352

packages/twenty-docker/docker-compose.dev.yml
```diff
@@ -38,5 +41,38 @@ services:
       retries: 10
     restart: unless-stopped
 
+  minio:
```
FelixMalfait: Let's not add this, it will confuse users ; our local storage driver is enough

packages/twenty-docs/developers/self-host/capabilities/setup.mdx
```diff
@@ -289,6 +289,33 @@ yarn command:prod cron:workflow:automated-cron-trigger
 **Environment-only mode:** If you set `IS_CONFIG_VARIABLES_IN_DB_ENABLED=false`, add these variables to your `.env` file instead.
 </Warning>
 
+## S3 Storage
+
+By default, Twenty stores files locally. To use S3 (or S3-compatible storage like MinIO), configure the `STORAGE_TYPE` and `STORAGE_S3_*` variables through the admin panel or `.env`.
```
FelixMalfait: I thought we had a warning somewhere about ephemeral storage already. Maybe worth insisting that for real production use-cases, setting a S3-compatible or persistent volume is crucial?

packages/twenty-docs/developers/self-host/capabilities/setup.mdx
```diff
@@ -289,6 +289,33 @@ yarn command:prod cron:workflow:automated-cron-trigger
...
+
+### Direct File Downloads (Presigned URLs)
```
FelixMalfait: I don't think we need to go as deep into the detail related to this specific PR. If there's a section on S3 storage it should be more generic on S3 storage config (tbh I thought we had such section already so check), not overly focused on the changes we're doing in this PR

packages/twenty-server/scripts/test-s3-presigned-url.sh
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env bash
```
FelixMalfait: This should not be commited! Remove it

packages/twenty-server/src/engine/core-modules/file-storage/drivers/__tests__/s3.driver.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  it('should create a separate S3 client for presigning with the public endpoint', () => {
+    new S3Driver({
+      bucketName: 'test-bucket',
+      region: 'us-east-1',
+      endpoint: 'http://internal:9000',
+      presignEndpoint: 'https://public.s3.com',
+    });
+
+    expect(S3).toHaveBeenCalledTimes(2);
```
FelixMalfait: I don't really get that test, can you explain why it has to be a separate S3? Do we really need such design / is that the standard way of doing things?

packages/twenty-server/src/engine/core-modules/file-storage/file-storage-driver.factory.ts
```diff
@@ -63,10 +63,15 @@ export class FileStorageDriverFactory extends DriverFactoryBase<StorageDriver> {
...
         );
+        const presignEndpointOverride = this.twentyConfigService.get(
+          'STORAGE_S3_PRESIGNED_URL_BASE',
+        );
+        const presignEndpoint = presignEndpointOverride ?? endpoint;
 
         rawDriver = new S3Driver({
           bucketName: bucketName ?? '',
           endpoint: endpoint,
+          presignEndpoint: presignEndpoint || undefined,
```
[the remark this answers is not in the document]
FelixMalfait: Reverted

etiennejouan (approved):
Tested with S3 (dev bucket), it works fine.
We could have an issue for our cloud set up,  there's a forcePathStyle:true that will break custom domain setups (CloudFront, CDN, reverse proxy).

The current implementation creates a second S3 client (presignClient) with STORAGE_S3_PRESIGNED_URL_BASE as its endpoint. This client inherits `forcePathStyle: true` from the main client, which means the bucket name is always prepended to the URL path: `<endpoint>/<bucket>/<key>`.
If `STORAGE_S3_PRESIGNED_URL_BASE` is a virtual-hosted URL (https://bucket.s3.region.amazonaws.com) or a custom domain mapped to a single bucket (https://cdn.example.com), the bucket ends up duplicated or added where it shouldn't be, causing a 404.

PR #18867, opened 2026-03-23
Animate the sidebar when opening/closing an element in the "opened" section

abdulrahmancodes:
https://github.com/user-attachments/assets/38f64750-7bca-4b10-8489-11d386a10298


etiennejouan (commented):
Image

@abdulrahmancodes it looks good and works well but the spacing seems unregular between "Opened" and following section. Can you set the same as between other sections ? (not seen anything in figma. If it doesn't seem logical, feel free to ask bonapara)

charlesBochet (approved):

PR #18871, opened 2026-03-23
Fix multi-select option removal crashing when records contain removed values

charlesBochet:
## Summary

- Fixes a bug where removing an option from a MULTI_SELECT field fails with `invalid input value for enum` when existing records contain the removed value alongside surviving values.
- The root cause was the `ELSE` branch in `updateArrayEnum` which tried to cast removed enum values (e.g. `DISTRIBUTOR`) to the new enum type that no longer includes them.
- The fix replaces the `ELSE` cast with a NULL-producing implicit CASE default and uses `array_agg(...) FILTER (WHERE mapped_value IS NOT NULL)` to silently strip removed values from existing arrays.

### Before (bug)
```sql
-- ELSE branch tries to cast removed value to new enum → crash
CASE unnest_value::text
  WHEN 'IMPL' THEN 'IMPL'::new_enum
  WHEN 'APP' THEN 'APP'::new_enum
  ELSE unnest_value::text::new_enum  -- 'DISTRIBUTOR' fails here
END
```

### After (fix)
```sql
-- No ELSE: removed values produce NULL, filtered out by array_agg
SELECT array_agg(mapped_value) FILTER (WHERE mapped_value IS NOT NULL)
FROM (
  SELECT CASE unnest_value::text
    WHEN 'IMPL' THEN 'IMPL'::new_enum
    WHEN 'APP' THEN 'APP'::new_enum
  END AS mapped_value
  FROM unnest(old_column) AS unnest_value
) enum_mapping
```

### DIFF
--- a/packages/twenty-server/src/engine/twenty-orm/workspace-schema-manager/services/workspace-schema-enum-manager.service.ts
+++ b/packages/twenty-server/src/engine/twenty-orm/workspace-schema-manager/services/workspace-schema-enum-manager.service.ts
@@ -327,7 +327,6 @@ export class WorkspaceSchemaEnumManagerService {
           escapedTable,
           escapedOldColumn,
           escapedNewColumn,
-          escapedNewEnumType,
           escapedOldEnumType: `${escapedSchema}.${escapeIdentifier(oldEnumTypeName)}`,
           caseStatements,
           mappedValuesCondition,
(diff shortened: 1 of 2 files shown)


FelixMalfait:

Your preview environment is available at: http://bore.pub:49926

martmull (approved):
lgtm

PR #18897, opened 2026-03-24
fix: expand kanban column drop zone to full height

JASSBR:
## Summary
- Added `flex: 1` to `StyledColumnContainer` in `RecordBoardColumns.tsx`
- The column wrapper wasn't stretching vertically, so the droppable area only covered the top of each column where cards existed
- Now the entire column height is a valid drop target

## Test plan
- [ ] Open a board/kanban view
- [ ] Drag a card from one column
- [ ] Drop it in the lower/empty area of another column
- [ ] Verify the drop registers correctly

Fixes #18842


thomtrp:
Hi @JASSBR, thank you for your contribution!

JASSBR:
> Hi @JASSBR, thank you for your contribution! Could you please provide before/after screenshots/recordings of the components? Sorry for asking but we are getting a lot of AI PRs that have not been properly tested.

Hi @thomtrp, thanks for the review!

  I've updated the PR with a more thorough fix. The initial approach (adding `flex: 1` to the column container) wasn't addressing the root cause.

  ### Root Cause
  The `@hello-pangea/dnd` `placeholder` was rendered outside the `innerRef` container (in `RecordBoardColumn`), while `innerRef` was on `StyledColumnCardsContainer`. This violated
   the library's API contract — when a card is dragged out, the `innerRef` container shrinks and the drop zone becomes smaller than the visible column.

  ### What changed
  - **Moved `innerRef` + `droppableProps` to `StyledColumn`** — this is the full-height column element, so the droppable area now covers the entire column
  - **Added `ignoreContainerClipping`** on `Droppable` — prevents the scroll wrapper from clipping the drop zone
  - **Removed explicit `height: 100%` / `min-height: 100%`** from `StyledColumn` — these were preventing `align-items: stretch` (flex default) from working, which is why shorter
  columns didn't match the height of taller ones
  - **Cleaned up `RecordBoardColumnCardsContainer`** — no longer needs `droppableProvided` as a prop

  ### Before
image
  ### After
image

  Tested locally — drag & drop now works across the full height of every column.

thomtrp (approved):
Thanks @JASSBR, LGTM ✅

PR #18902, opened 2026-03-24
fix: validate blocknote JSON in rich text fields

charlesBochet:
## Summary
- **Backend**: Add JSON validation for the `blocknote` subfield in rich text API inputs — rejects values that aren't valid JSON or aren't arrays (BlockNote content is always `PartialBlock[]`). This prevents corrupted data from being persisted to the database.
- **Frontend**: Replace all 5 unprotected `JSON.parse` calls on blocknote content with the safe `parseJson` utility from `twenty-shared`. Invalid content now degrades gracefully (empty block / empty string / unchanged passthrough) instead of crashing the app.
- **Tests**: Added integration tests for invalid blocknote JSON (both GraphQL and REST), unit tests for the new validation, and updated existing test constants to use valid BlockNote JSON.

## Context
A user reported a `SyntaxError: Expected ',' or ']' after array element` crash caused by malformed blocknote JSON stored in the database. The data had `"children":[]` nested inside the `content` array instead of as a sibling property. The API accepted this invalid JSON because it only validated that `blocknote` was a string, not that it contained valid JSON. On the frontend, 5 call sites used bare `JSON.parse` with no error handling, causing a white-screen crash.

## Test plan
- [x] Unit tests pass: `validate-rich-text-field-or-throw.util.spec.ts` (10/10)
- [x] Integration tests pass: `rich-text-field-create-input-validation` (8/8)
- [ ] Verify creating a note with valid rich text still works end-to-end
- [ ] Verify API returns clear error when blocknote contains invalid JSON
- [ ] Verify frontend renders empty block instead of crashing when encountering corrupted data

### DIFF
--- a/packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/validator-utils/validate-rich-text-field-or-throw.util.ts
+++ b/packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/validator-utils/validate-rich-text-field-or-throw.util.ts
@@ -1,34 +1,14 @@
 import { inspect } from 'util';
 
 import { msg } from '@lingui/core/macro';
-import { isNonEmptyString, isNull, isObject } from '@sniptt/guards';
-import {
-  compositeTypeDefinitions,
-  FieldMetadataType,
-} from 'twenty-shared/types';
+import { isNonEmptyString, isNull, isObject, isString } from '@sniptt/guards';
 
 import {
   CommonQueryRunnerException,
   CommonQueryRunnerExceptionCode,
 } from 'src/engine/api/common/common-query-runners/errors/common-query-runner.exception';
 
-const validateBlocknoteJsonOrThrow = (blocknote: unknown): void => {
-  if (!isNonEmptyString(blocknote)) {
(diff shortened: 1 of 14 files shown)


FelixMalfait:

Your preview environment is available at: http://bore.pub:29280

packages/twenty-server/src/engine/api/common/common-args-processors/data-arg-processor/validator-utils/validate-rich-text-field-or-throw.util.ts
```diff
@@ -1,55 +1,79 @@
...
+          `Invalid subfield ${inspect(subField)} for rich text field "${fieldName}"`,
+          CommonQueryRunnerExceptionCode.INVALID_ARGS_DATA,
+          { userFriendlyMessage: msg`Invalid value for rich text.` },
+        );
+    }
+  }
+
+  return value as {
+    blocknote?: string | null;
+    markdown?: string | null;
```
thomtrp, line 77: do we need that casting?

thomtrp (approved):

PR #18904, opened 2026-03-24
feat: make workflow objects searchable

FelixMalfait:
## Summary
- Adds `isSearchable: true` to the workflow standard object definition so new workspaces get searchable workflows automatically
- Adds a `upgrade:1-20:make-workflow-searchable` migration command that flips the `isSearchable` flag on the `objectMetadata` row for existing workspaces, with proper cache invalidation and metadata version increment
- Registers the command in the 1-20 upgrade module and the `upgrade.command.ts` orchestrator

The `searchVector` stored generated column already exists on the workflow table, so no data backfill is needed — this is purely a metadata flag change that makes the search service include workflows in results.

## Test plan
- [x] `--dry-run` logs what it would do without making changes
- [x] Actual run updates both workspaces and invalidates caches
- [x] Idempotent: re-running skips already-searchable workspaces
- [x] Typecheck passes
- [x] Lint passes on changed files
- [x] Prettier passes on changed files


FelixMalfait:

Your preview environment is available at: http://bore.pub:36191

PR #18907, opened 2026-03-24
Fix lambda driver

### LINKED ISSUE #2357: Fix subhosting lambda permission issues
as title

martmull:
as title

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/lambda.driver.ts
+++ b/packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/lambda.driver.ts
@@ -114,6 +114,9 @@ export interface LambdaDriverOptions extends LambdaClientConfig {
 
 export class LambdaDriver implements LogicFunctionDriver {
   private lambdaClient: Lambda | undefined;
+  private assumeRoleCredentials:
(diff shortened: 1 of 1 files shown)


packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/lambda.driver.ts
```diff
@@ -166,11 +155,42 @@ export class LambdaDriver implements LogicFunctionDriver {
...
 
-    return {
+    this.assumeRoleCredentials = {
       accessKeyId: Credentials.AccessKeyId,
       secretAccessKey: Credentials.SecretAccessKey,
       sessionToken: Credentials.SessionToken,
     };
+
+    this.credentialsExpiry = new Date(
+      Date.now() + (CREDENTIALS_DURATION_IN_SECONDS - 60 * 5) * 1000,
```
thomtrp, line 165: minus 60 ? why?

martmull: yes, it is ugly (it was there before)
This is a margin to refresh the credentials before hard expiry, if it expires in less than 5min, we refresh. credentials duration is 1h today

thomtrp (approved):

PR #18909, opened 2026-03-24
Seed company workflow for email upserts

Bonapara:
@thomtrp


Bonapara:
working locally after database reset

FelixMalfait:

Your preview environment is available at: http://bore.pub:36173

packages/twenty-server/src/engine/core-modules/workspace/services/workspace.service.ts
```diff
@@ -681,6 +684,12 @@ export class WorkspaceService extends TypeOrmQueryService<WorkspaceEntity> {
         },
       );
 
+    await ensureCreateCompanyWhenAddingNewPersonCodeStepLogicFunctions({
```
charlesBochet: we should use nest service injection ensureCreateCompanyWhenAddingNewPersonCodeStepLogicFunctions should likely be a service

Also: workspace.service should not take dependency on metadata-modules. If we want to create a logicFunction, we should use lower level services of logic functions from core-modules. Otherwise we will end up with circular dependencies

charlesBochet (changes requested):
Thank you @Bonapara, nice try :p

Plus fix the tests

charlesBochet (approved):

PR #18940, opened 2026-03-25
Clear navbar edit selection when closing the side panel

abdulrahmancodes:
In navbar edit mode, selecting an item for edit stored selectedNavigationMenuItemIdInEditModeState (and related pending-insertion state). Closing the side panel did not reset those atoms, so the nav item stayed visually selected. Reset both when the panel closes so the highlight matches the closed panel; reopening in edit mode then starts from the generic “new item” entry unless the user picks an item again.


packages/twenty-front/src/modules/side-panel/hooks/useSidePanelMenu.ts
```diff
@@ -29,6 +30,8 @@ export const useSidePanelMenu = () => {
 
     if (isSidePanelOpened) {
       store.set(addToNavPayloadRegistryState.atom, new Map());
+      store.set(selectedNavigationMenuItemIdInEditModeState.atom, null);
```
charlesBochet: sidePanel is a lower level module that should not take dependency on navItems which is a consumer of side panel.

We should have a higher level hook useNavigationMenuItemSidePanel that should update the stores (addToNavPayloadRegistryState actually) and call useSidePanelMenu

charlesBochet (changes requested):

packages/twenty-front/src/modules/side-panel/hooks/useSidePanelMenu.ts
abdulrahmancodes: Agreed, I removed nav state updates from `useSidePanelMenu`. Cleanup now lives in `navigation-menu-item` via `useResetNavigationMenuItemStateOnSidePanelClose`, wired to the existing side-panel close event (same idea as the workflow hook)

etiennejouan (commented):
I'm not sure to clearly understand how this hook is built and when/where to call it. The best would be to trigger when side panel is closed in context of nav edition. I'll continue searching 🚧 

Edit - older message 🔽 : 
> Do you think cleaning states in existing `useSidePanelCloseAnimationCompleteCleanup` could be an alternative ?

Not read Charles' comment, mine went against

charlesBochet (approved):

charlesBochet:
Better! Thank you @abdulrahmancodes

PR #18951, opened 2026-03-25
fix: wrap standard object/field metadata labels in msg for i18n extraction

FelixMalfait:
## Summary

- Standard object and field metadata labels were plain strings instead of being wrapped in Lingui `msg` template literals, which prevented extraction into translation catalogs. This caused "Uncompiled message detected!" warnings at runtime.
- The bug was introduced when the decorator-based approach (`@WorkspaceEntity({ labelSingular: msg`...` })`) was replaced by flat metadata builder utils with plain strings.
- Adds an `i18nLabel` helper to safely extract the message string from `MessageDescriptor` objects.
- Re-runs `lingui extract` and `lingui compile` to update all locale PO files and compiled catalogs.
- Adds an integration test that queries the metadata API with `x-locale` headers to verify locale-aware label resolution.
- Includes a ClickHouse usage event writer integration test (small unrelated fix noticed along the way).

## Test plan

- [ ] CI lint passes (prettier + oxlint)
- [ ] CI typecheck passes
- [ ] Server unit tests pass
- [ ] Integration tests pass (including new `object-metadata-i18n` test)
- [ ] No "Uncompiled message detected!" warnings for standard object/field labels at runtime


FelixMalfait:
Merging as I need to iterate more and run the cron to re-import and finish but would appreciate if you can have a look @prastoin thanks ; it's not perfect but I think it's a good enough hack for now

PR #18954, opened 2026-03-25
Fix: use user role for OAuth tokens bearing user context

ijreilly:
see [discord discussion](https://discord.com/channels/1130383047699738754/1486299347091198054/1486299351520383116)

## Summary

When an OAuth application token carries both `applicationId` and `userId`/`userWorkspaceId`, the auth context now uses the **user's role** for permissions instead of the application's `defaultRoleId`.

This fixes the case where external clients authenticating via OAuth (e.g. a client's external AI chat) were getting the app's permissions instead of the authenticated user's.

### What changed

- **`jwt.auth.strategy.ts`** (`validateApplicationToken`): when an application token includes user info, also resolve `workspaceMemberId` and `workspaceMember` from the workspace cache (same pattern as `validateAccessToken`)
- **`workspace-auth-context.middleware.ts`** (`buildAuthContext`): when both `application` and `user` (with `workspaceMemberId`/`workspaceMember`) are present on the request, build a `UserWorkspaceAuthContext` instead of `ApplicationWorkspaceAuthContext`. Falls back to application context if workspace member cannot be resolved.

### Places impacted by this change (no code changes, behavior changes)

These places check `isApplicationAuthContext` or resolve roles from auth context. Since hybrid tokens (OAuth with user) now produce a `UserWorkspaceAuthContext`, they naturally flow into the `isUserAuthContext` branches:

| File | Impact |
|------|--------|
| `permissions.service.ts` — `resolveRolePermissionConfigFromAuthContext` | OAuth+user now uses user's role via `isUserAuthContext` branch instead of `application.defaultRoleId` |
| `common-api-context-builder.service.ts` — `getObjectsPermissions` | Same — OAuth+user falls into `isUserAuthContext` branch |
| `common-base-query-runner.service.ts` — `getRoleIdOrThrow` | Same — OAuth+user falls into `isUserAuthContext` branch |
| `actor-from-auth-context.service.ts` — `buildActorMetadata` | Records created via OAuth+user will show the **user's name** as actor instead of the application's name |
| `message-find-one.post-query.hook.ts` | OAuth+user now passes the `isUserAuthContext` check (previously would fail unless it was the Twenty standard application) |
| `front-component.resolver.ts` | Front component tokens include `userId` — they will now correctly use the user's role, fixing a pre-existing permission escalation where a user could access data through a front component's app role that exceeded their own |
| `logic-function-executor.service.ts` | **Not impacted** — only generates tokens with `applicationId` (no `userId`) |

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/auth/middlewares/__tests__/workspace-auth-context.middleware.spec.ts
+++ b/packages/twenty-server/src/engine/core-modules/auth/middlewares/__tests__/workspace-auth-context.middleware.spec.ts
@@ -6,7 +6,7 @@ import {
 } from 'src/engine/core-modules/auth/auth.exception';
 import { workspaceAuthContextStorage } from 'src/engine/core-modules/auth/storage/workspace-auth-context.storage';
 
-import { WorkspaceAuthContextMiddleware } from './workspace-auth-context.middleware';
+import { WorkspaceAuthContextMiddleware } from '../workspace-auth-context.middleware';
 
 const mockWorkspace = {
   id: 'workspace-id',
--- a/packages/twenty-server/src/engine/core-modules/auth/middlewares/workspace-auth-context.middleware.ts
+++ b/packages/twenty-server/src/engine/core-modules/auth/middlewares/workspace-auth-context.middleware.ts
@@ -39,6 +39,21 @@ export class WorkspaceAuthContextMiddleware implements NestMiddleware {
     }
 
     if (isDefined(req.application)) {
+      if (
+        isDefined(req.user) &&
+        isDefined(req.userWorkspaceId) &&
+        isDefined(req.workspaceMemberId) &&
+        isDefined(req.workspaceMember)
+      ) {
+        return buildUserAuthContext({
+          workspace: req.workspace!,
+          userWorkspaceId: req.userWorkspaceId,
+          user: req.user,
+          workspaceMemberId: req.workspaceMemberId,
+          workspaceMember: req.workspaceMember,
+        });
+      }
+
       return buildApplicationAuthContext({
         workspace: req.workspace!,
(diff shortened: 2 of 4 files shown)


packages/twenty-server/src/engine/core-modules/auth/middlewares/workspace-auth-context.middleware.ts
```diff
@@ -39,6 +39,21 @@ export class WorkspaceAuthContextMiddleware implements NestMiddleware {
     }
 
     if (isDefined(req.application)) {
+      if (
+        isDefined(req.user) &&
+        isDefined(req.userWorkspaceId) &&
+        isDefined(req.workspaceMemberId) &&
+        isDefined(req.workspaceMember)
+      ) {
+        return buildUserAuthContext({
```
ijreilly: or we can just move up #L63 check before this one but this adds clarity

packages/twenty-server/src/engine/core-modules/auth/middlewares/workspace-auth-context.middleware.ts
```diff
@@ -60,6 +53,13 @@ export class WorkspaceAuthContextMiddleware implements NestMiddleware {
       });
     }
 
+    if (isDefined(req.application)) {
```
ijreilly, line 56: a test protects us against potential undesired consequences from this being moved around again

packages/twenty-server/src/engine/core-modules/auth/middlewares/__tests__/workspace-auth-context.middleware.spec.ts
martmull, line 1: should be added in a __tests__ folder

martmull (approved):

PR #18961, opened 2026-03-25
Use record page layouts in the merge records feature

Devessier:
There was a design issue. It revealed that we weren't using record page layouts in the merge records feature.

## Before

image

## After

https://github.com/user-attachments/assets/c0fa9cf5-2b28-4696-bf20-8271dce9e62c


ehconitin (approved):
lgtm!

PR #18962, opened 2026-03-25
Fix record page layout upgrade commands

Devessier:
- Fix issue with name that must be first in the view field list
- Improve dry run and logs of backfill page layouts command


packages/twenty-server/src/database/commands/upgrade-version-command/1-20/1-20-backfill-page-layouts.command.ts
```diff
@@ -124,27 +177,35 @@ export class BackfillPageLayoutsCommand extends ActiveOrSuspendedWorkspacesMigra
...
         tabUniversalIdentifiers.add(tab.universalIdentifier);
 
+        if (
+          isDefined(
+            existingFlatPageLayoutTabMaps.byUniversalIdentifier[
+              tab.universalIdentifier
+            ],
+          )
+        ) {
+          return false;
```
prastoin, line 187: ***Remark:*** All these check should be redundant, as the builder will detect nothing has to be created if the universal identifier is already present in the fromMaps

packages/twenty-server/src/engine/metadata-modules/object-metadata/utils/compute-flat-view-fields-to-create.util.ts
```diff
@@ -29,6 +29,19 @@ export const computeFlatViewFieldsToCreate = ({
           field.universalIdentifier ===
             labelIdentifierFieldMetadataUniversalIdentifier),
     )
+    .sort((a, b) => {
+      const aIsLabel =
+        a.universalIdentifier ===
+        labelIdentifierFieldMetadataUniversalIdentifier;
+      const bIsLabel =
```
prastoin: ***Subjectice nitpick:*** naming, suggestion something like `aIsLabelIdentifierFieldMetadata`

packages/twenty-server/src/database/commands/upgrade-version-command/1-20/1-20-backfill-page-layouts.command.ts
Devessier: If some widgets have already been created, I still want to create the other that don't exist yet. As per my understanding, the builder would reject the whole transaction if one widget already exists. Am I wrong?

prastoin (commented):
Hey there, I'm not sure to understrand what we're doing with all the custom objects
Are duplicating all their view fields ? but we assign them not to view but record page fields view ?
```ts
        viewUniversalIdentifier: flatRecordPageFieldsView.universalIdentifier,
```

This command isn't idempotent is it ?

packages/twenty-server/src/database/commands/upgrade-version-command/1-20/1-20-backfill-page-layouts.command.ts
prastoin: Yep you're right `computeUniversalFlatEntityMapsFromToThroughMutation` would fail

Devessier:
> This command isn't idempotent is it ?

It is because if the feature flag is already defined, we exit early

prastoin:
> > This command isn't idempotent is it ?
> 
> It is because if the feature flag is already defined, we exit early

If it partially fails the universal identifier not being deterministic we would be duplicating a lot of data
I understand why you would need a single query runner here

As a quick fix I would recommend creating v5 uuid that could be deterministically generated as derivated on the parent universal identifier

prastoin (approved):
Unless I'm mistaken I was a fool regarding the partial upgrade
Both standard and custom are performed within a single transaction
Which mean that if the custom fails they will all be rollbacked
We don't care that they're not derivated from their parents

PR #18966, opened 2026-03-25
Fix SSE event stream errors when impersonating users with limited permissions

thomtrp:
Clear the SSE client before swapping auth tokens during impersonation, preventing a userWorkspaceId mismatch between the existing event stream (created under the admin's identity) and the new impersonation token.
Treat NOT_AUTHORIZED event stream errors as recoverable (destroy + recreate), matching the existing behavior for EVENT_STREAM_DOES_NOT_EXIST and EVENT_STREAM_ALREADY_EXISTS.


lucasbordeau (approved):
LGTM

PR #18969, opened 2026-03-25
Fix create new with filters

lucasbordeau:
Fixes https://github.com/twentyhq/twenty/issues/18949

## Problem
- Creating a record from filters with DATE_TIME fields produced empty objects instead of dates — `isPlainObject` from `twenty-shared` treats `Date` instances as plain objects, so `mergeCompositeValues` spread them into `{}`
- `buildRecordInputFromFilter` applied composite merge logic to all field types indiscriminately, including primitives, dates, and strings
- DATE_TIME "is before" filters produced exact boundary values instead of subtracting a minute

## Fix
- Composite and non-composite fields are now handled in separate branches — `buildRecordInputFromFilter` uses `isCompositeFieldType` to decide whether to merge or assign directly
- `mergeCompositeValues` extracted to its own file with a properly typed signature (`Record<string, unknown>`) — no more runtime type guessing
- DATE_TIME fields with `IS_BEFORE` operand subtract one minute using `subMinutes` from `date-fns`
- 13 unit tests for `mergeCompositeValues` covering all composite field types (currency, address, full name, links, emails, phones) and successive sub-field accumulation


packages/twenty-front/src/modules/object-record/record-table/utils/buildRecordInputFromFilter.ts
```diff
@@ -40,22 +42,35 @@ export const buildRecordInputFromFilter = ({
...
         return;
       }
-      recordInput[fieldMetadataItem.name] = mergeCompositeValues(
-        recordInput[fieldMetadataItem.name],
-        value,
-      );
+
+      if (isCompositeFieldType(fieldMetadataItem.type as FieldType)) {
+        recordInput[fieldMetadataItem.name] = mergeCompositeValues(
+          recordInput[fieldMetadataItem.name] as ObjectRecord | undefined,
```
charlesBochet: is the value really ObjectRecord?

packages/twenty-front/src/modules/object-record/record-table/utils/buildRecordInputFromFilter.ts
```diff
@@ -40,22 +42,35 @@ export const buildRecordInputFromFilter = ({
...
+
       if (!isDefined(value)) {
         return;
       }
-      recordInput[fieldMetadataItem.name] = mergeCompositeValues(
-        recordInput[fieldMetadataItem.name],
-        value,
-      );
+
+      if (isCompositeFieldType(fieldMetadataItem.type as FieldType)) {
```
charlesBochet: why do we need as here? isn't fieldMetadataItem.type of type FieldType?

packages/twenty-front/src/modules/object-record/record-table/utils/mergeCompositeValues.ts
```diff
@@ -0,0 +1,7 @@
+export const mergeCompositeValues = (
```
charlesBochet: I don't see anything tied to composite here, not sure this should be an util with dedicated tests. I feel testing properly packages/twenty-front/src/modules/object-record/record-table/utils/buildRecordInputFromFilter.ts is enough (with composite cases)

charlesBochet (changes requested):

charlesBochet (approved):

PR #18989, opened 2026-03-26
Fix dropping workspace nav items at the bottom of the list

abdulrahmancodes:
https://github.com/user-attachments/assets/64abd955-892e-48c8-bf7b-d4d8645cf33e


packages/twenty-front/src/modules/navigation-menu-item/display/sections/workspace/components/WorkspaceSectionListDndKit.tsx
```diff
@@ -50,6 +60,13 @@ export const WorkspaceSectionListDndKit = ({
     (item) => item.type === NavigationMenuItemType.FOLDER,
   ).length;
   const isAddMenuItemButtonVisible = isLayoutCustomizationModeEnabled;
+  const reorderOrphanCount = workspaceNavigationMenuItems.filter(
+    (item) => !isDefined(item.folderId),
+  ).length;
+  const orphanAppendDndIndex = Math.max(
```
etiennejouan: It becomes confusing to handle filteredItems and workspaceNavigationMenuItems, two different sources of nav menu items. If I understand well, it's because of permission ?
I discuss it with bonapara and he suggests to display unaccessible nav menu item (in edit mode only), editor could re-order or remove it, it should simplify logic

packages/twenty-front/src/modules/navigation-menu-item/display/sections/workspace/components/WorkspaceSectionListDndKit.tsx
```diff
@@ -78,35 +95,44 @@ export const WorkspaceSectionListDndKit = ({
...
-          index={filteredItems.length}
-          compact={!isAddMenuItemButtonVisible}
+      <StyledOrphanAppendSlotOverlap>
+        <NavigationMenuItemDroppableSlot
+          droppableId={
+            NavigationMenuItemDroppableIds.WORKSPACE_ORPHAN_NAVIGATION_MENU_ITEMS
+          }
+          index={orphanAppendDndIndex}
+          disabled={workspaceDropDisabled}
+          collisionPriority={FOLDER_HEADER_SLOT_COLLISION_PRIORITY}
```
etiennejouan, line 96: nit : should not be re-used for a non-folder context. All collision priority should also be centralized

etiennejouan (commented):
It works well ! 👍 

But nav menu items position computing seems very confusing. Sorry we should have brainstormed this earlier. It should simplify a lot to display not permitted nav menu item in edition mode, one single source of truth for nav menu item. Waiting for bonapara design.

When dragging on last position, "Add menu item" should not be highlighted (with blue background)

etiennejouan:
rger
CleanShot_2026-03-26_at_16 59 18


locker icon on navigation menu items in edition mode, when user does not have permission. Note that as record name can't be display then "Record" is set to replace (concerning object name it's metadata, it's ok).
Concerning the Object not shared view, thomast is working on it

etiennejouan (approved):

PR #18993, opened 2026-03-26
fix: hide objects with no addable views in sidebar view picker

abdulrahmancodes:


etiennejouan (commented):
I've asked for Bonapara product validation for this improvement (I'm not sure we want or don't want duplicate, even if it seems cleaner). Maybe he requests this changes ?

We should add this same view filtering logic (in a dedicated util ?) on ... ViewSystemPickerSubPage and ... ViewPickerSubView (less useful as it's not shown if already filter in ObjectPicker, but clearer ?). wdyt ?

We should also hide the view command menu item if no view available. Globally, hide all command menu item, if there is no available subpages/options

charlesBochet (approved):

charlesBochet:
Made some changes and slight refactor to your PR
Discussed with @Bonapara, we will also let the index views be selected for now.

@abdulrahmancodes we should base the whole code on navigationItem.type and not on the existence of .viewId to now if that's a navItem for views or not (same for all other categories), this usually simplifies the code

PR #19001, opened 2026-03-26
fix: sign file URLs in timeline and file loss on click outside

### LINKED ISSUE #18943: [Bug] Files field attachments disappear after re-editing / deleting, and timeline records cannot be downloaded afterwards
## Bug Description

Hi team, thank you for the great project.

I found an issue with the **Files** field on **v1.19.0**. It seems that attachments may disappear unexpectedly during editing, and afterwards they also cannot be recovered from the timeline.

### Issue 1

When uploading an attachment to a **Files** field, if I click the **pen/edit icon** on the right side again, make no changes, and then click anywhere else on the page, the uploaded attachment disappears.

There is also a related case:

If I upload **two attachments**, clicking the pen/edit icon alone does not immediately make them disappear, but if I then delete one of the attachments, the other one also disappears unexpectedly.

### Issue 2

After the attachment disappears, I tried to recover it from the **timeline/activity history**. I can still see the record showing that the attachment was uploaded before, but I cannot click or download the file from there.

## Expected behavior

- Clicking the **pen/edit icon** and then clicking elsewhere without making changes should **not remove existing attachments**.
- If multiple attachments are uploaded, deleting one attachment should **only remove that one**, and the remaining attachments should stay intact.
- If an upload record is shown in the **timeline/activity history**, the corresponding file should still be accessible or downloadable, or at least the UI should clearly indicate that the file is no longer available.

## Technical inputs

- Version: **v1.19.0**
- The issue happens on a **Files** field.
- The problem seems to occur during the **edit / re-open interaction** of the Files field.
- It may be related to frontend state handling when reopening the field editor, because attachments disappear even when no new action is confirmed.
- There also seems to be an inconsistency between the **timeline record** and the actual file availability, since the upload event remains visible but the file cannot be opened or downloaded afterwards.

bugisthegod:
Fixes: #18943
## Problem

  Two bugs related to the Files field:

  1. **File loss on click outside**: When `MultiItemFieldInput` was not in edit mode (input hidden), clicking outside would still call `validateInputAndComputeUpdatedItems()` with an empty `inputValue` and `itemToEditIndex = 0`, causing the first file to be silently deleted. 
image

  2. **Unsigned file URLs in timeline activity**: FileId, FileName stored in `timelineActivity.properties.diff` (before/after values) were not being signed (timeActivity.properties is stored as json in database), making them inaccessible from the 
frontend.

image

 ## Reproduction

https://github.com/user-attachments/assets/e75b842b-5cbb-46e8-a923-ac9df62deb98

## Changes
- `MultiItemFieldInput.tsx`: Wrap the validate + onChange logic in `if (isInputDisplayed)` so it only runs when the input is actually open.
- `timeline-activity-query-result-getter.handler.ts`: New handler that iterates over `properties.diff` fields and signs any file arrays found in `before`/`after` values (call same method: `fileUrlService.signFileByIdUrl` as table field view  `FilesFieldQueryResultGetterHandler`.
- `common-result-getters.service.ts`: Register the new handler for `timelineActivity`.

## After 

https://github.com/user-attachments/assets/368c7be1-3101-43a2-ac93-fe1b0d8a7a37


packages/twenty-server/src/engine/api/graphql/workspace-query-runner/factories/query-result-getters/handlers/timeline-activity-query-result-getter.handler.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    for (const [fieldName, fieldDiff] of Object.entries(diff)) {
+      if (!isObject(fieldDiff)) {
+        updatedDiff[fieldName] = fieldDiff;
+        continue;
+      }
+
+      const updatedFieldDiff: Record<string, unknown> = { ...fieldDiff };
+
+      for (const diffKey of ['before', 'after'] as const) {
```
etiennejouan: `before` should not be signed as file is deleted, then not accessible anymore.

etiennejouan (commented):
Hi @bugisthegod 👋  Thanks a lot for this new good contribution !

Your approach on timeline activities is technically very clean but on a product side it raises issues (not on you!!). How we will handle deleted (not accessible anymore) files ? file url is signed but points to nothing.

We should not sign any timeline activities file url as we don't know if file still exists (and asking db for each fileId existence would be a dramatic perf issue).

Then, we should have a specific logic on UX side. The same with have with other events enabling more informations display (with chevron down). In that case, for files field update event (in timeline activities), we would findOne the given record with given file record and check if it returns fileId (with url). If it has, we display fileChip clickable (to file preview) and if not fileChip in disable state (and cherry on the cake, with a "File not existing anymore" tooltip)

So, what I would do now is : removing timeline getter handler logic. And merging this PR with first fix. If you want a work on the second issue on an other PR, you're very welcome ! And if you want to do something else, I'll do it.

bugisthegod:
> Hi @bugisthegod 👋 Thanks a lot for this new good contribution !
> 
> Your approach on timeline activities is technically very clean but on a product side it raises issues (not on you!!). How we will handle deleted (not accessible anymore) files ? file url is signed but points to nothing.
> 
> We should not sign any timeline activities file url as we don't know if file still exists (and asking db for each fileId existence would be a dramatic perf issue).
> 
> Then, we should have a specific logic on UX side. The same with have with other events enabling more informations display (with chevron down). In that case, for files field update event (in timeline activities), we would findOne the given record with given file record and check if it returns fileId (with url). If it has, we display fileChip clickable (to file preview) and if not fileChip in disable state (and cherry on the cake, with a "File not existing anymore" tooltip)
> 
> So, what I would do now is : removing timeline getter handler logic. And merging this PR with first fix. If you want a work on the second issue on an other PR, you're very welcome ! And if you want to do something else, I'll do it.

Thanks a lot for your detailed review! This kind of product-level feedback is exactly what I need — it really helps me think beyond the technical fix and understand the product more deeply!!!

I have already removed the handler logic. And as what you said `The same with have with other events enabling more informations display (with chevron down)`. Is it more likely this: when the card is expanded, query each file by `fileId` to check if it still exists
image

etiennejouan (approved):
Thanks for your contribution !! 👏 

I'll continue the discussion on the issue page.

PR #19015, opened 2026-03-26
Fix: Cannot create two workflow agent nodes because these have the same name

thomtrp:
Currently, when trying to create a second step agent, we get the error agent already exists because the name is using workflow id. Using step id instead.

### DIFF
--- a/packages/twenty-server/src/modules/workflow/workflow-builder/workflow-version-step/workflow-version-step-operations.workspace-service.ts
+++ b/packages/twenty-server/src/modules/workflow/workflow-builder/workflow-version-step/workflow-version-step-operations.workspace-service.ts
@@ -704,8 +704,7 @@ export class WorkflowVersionStepOperationsWorkspaceService {
 
         const clonedAgent = await this.agentService.createOneAgent(
           {
-            label:
-              existingAgent.label + ' ' + clonedStepId.substring(0, 4),
+            label: existingAgent.label + ' ' + clonedStepId.substring(0, 4),
             icon: existingAgent.icon ?? undefined,
             description: existingAgent.description ?? undefined,
             prompt: existingAgent.prompt,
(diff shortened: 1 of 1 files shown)


packages/twenty-server/src/modules/workflow/workflow-builder/workflow-version-step/workflow-version-step-operations.workspace-service.ts
```diff
@@ -432,16 +432,9 @@ export class WorkflowVersionStepOperationsWorkspaceService {
...
-          await this.workflowCommonWorkspaceService.getWorkflowVersionOrFail({
-            workflowVersionId,
-            workspaceId,
-          });
-
         const newAgent = await this.agentService.createOneAgent(
           {
-            label:
-              'Workflow Agent' + workflowVersion.workflowId.substring(0, 4),
+            label: 'Workflow Agent ' + baseStep.id.substring(0, 4),
```
martmull, line 437: Don't really like this solution
remove

@Index('IDX_AGENT_NAME_WORKSPACE_ID_UNIQUE', ['name', 'workspaceId'], {
  unique: true,
  where: '"deletedAt" IS NULL',
})

in packages/twenty-server/src/engine/metadata-modules/ai/ai-agent/entities/agent.entity.ts

and apply a weak unicity check where we think it is useful

martmull (commented):
need change i guess

packages/twenty-server/src/modules/workflow/workflow-builder/workflow-version-step/workflow-version-step-operations.workspace-service.ts
FelixMalfait: Agree it could be nice but the product vision for what Agents really are is not 100% clear outside of workflows (settings etc) so I wouldn't over invest in that for now. Maybe @thomtrp's solution can be a good patch for now

FelixMalfait (approved):

packages/twenty-server/src/modules/workflow/workflow-builder/workflow-version-step/workflow-version-step-operations.workspace-service.ts
FelixMalfait: But probably you're right long-term @martmull but it will involve more changes when we truly expose agents in settings

PR #19020, opened 2026-03-26
Fix not shared chip height + hard coded border radius

Bonapara:
## Summary
- align the forbidden field "Not shared" chip with small chip dimensions in the object table
- use the shared small border radius token instead of a hardcoded value
- add `overflow: hidden` and `user-select: none` to match chip behavior more closely

## Testing
- Not run (not requested)


FelixMalfait:

Your preview environment is available at: http://bore.pub:25943

charlesBochet (approved):

PR #19027, opened 2026-03-27
fix: batch viewGroup mutations sequentially to prevent race conditions

cyphercodes:
## Bug Description

When reordering stages in the Kanban board, the frontend fires all viewGroup update mutations concurrently via Promise.all, causing race conditions in the workspace migration runner's cache invalidation, database contention, and a thundering herd effect that stalls the server.

## Changes

Changed `usePerformViewGroupAPIPersist` to execute viewGroup update mutations sequentially instead of concurrently. The `Promise.all` pattern fired all N mutations simultaneously, each triggering a full workspace migration runner pipeline (transaction + cache invalidation). The sequential `for...of` loop ensures each mutation completes (including its cache invalidation) before the next begins, eliminating the race condition.

## Related Issue

Fixes #18865

## Testing

This fix addresses the root cause identified in the Sonarly analysis on the issue. The concurrent mutation pattern was causing:
- PostgreSQL row-level lock contention on viewGroup rows
- Cache thundering herd from repeated invalidation/recomputation cycles
- Server stalls requiring container restarts

The sequential approach ensures proper ordering and prevents these race conditions.


charlesBochet:
@cyphercodes Thanks for giving a look, while reviewing your PR and the original issues, I think we should actually batch the call, which is way more efficient

charlesBochet (approved):

ch9hn:
Thanks for checking it an referring to my created issue [#18865](https://github.com/twentyhq/twenty/issues/18865).

PR #19039, opened 2026-03-27
fix: resolve settings/usage chart crash and add ClickHouse usage event seeds

FelixMalfait:
## Summary

- **Fix settings/usage page crash**: The `GraphWidgetLineChart` component used on `settings/usage` was crashing with "Instance id is not provided and cannot be found in context" because it requires `WidgetComponentInstanceContext` (for tooltip/crosshair component states) which is only provided inside the widget system. Wraps the standalone chart usages with the required context provider.
- **Avoid mounting `GraphWidgetLegend` when hidden**: The legend component calls `useIsPageLayoutInEditMode()` which requires `PageLayoutEditModeProviderContext` — another context only available inside the widget system. Since the settings page passes `showLegend={false}`, the fix conditionally unmounts the legend instead of always mounting it with a `show` prop. Applied consistently across all four chart types (line, bar, pie, gauge).
- **Add ClickHouse usage event seeds**: Generates ~400 realistic `usageEvent` rows spanning the past 35 days with weighted user activity, weekday/weekend patterns, and gradual ramp-up. Enables developers to see the usage analytics page with data locally.

## Test plan

- [ ] Navigate to `settings/usage` — page should render without errors
- [ ] Verify the daily usage line chart displays correctly
- [ ] Navigate to a user detail page from the usage list
- [ ] Verify the user detail chart renders without errors
- [ ] Run `npx nx clickhouse:seed twenty-server` and confirm usage events are seeded
- [ ] Verify chart legend still works correctly on dashboard widgets (no regression)


FelixMalfait:

Your preview environment is available at: http://bore.pub:12256

packages/twenty-front/src/modules/ai/components/internal/AIChatContextUsageButton.tsx
```diff
@@ -105,6 +95,19 @@ export const AIChatContextUsageButton = () => {
   const { t } = useLingui();
   const [isHovered, setIsHovered] = useState(false);
   const agentChatUsage = useAtomStateValue(agentChatUsageState);
+  const billing = useAtomStateValue(billingState);
+  const isBillingEnabled = billing?.isBillingEnabled ?? false;
+
+  // Values from the streaming API arrive as display credits (micro-credits / 1000).
```
FelixMalfait, line 101: Comment probably not needed

PR #19049, opened 2026-03-27
Fix INDEX view showing labelPlural instead of resolved view name in nav

charlesBochet:
## Summary
- Navigation sidebar was displaying "Notes" instead of "All Notes" for INDEX views
- `getNavigationMenuItemLabel` had a special case for INDEX views that returned `objectMetadataItem.labelPlural` instead of the already-resolved `view.name`
- Since `viewsSelector` already resolves `{objectLabelPlural}` templates via `resolveViewNamePlaceholders`, the INDEX special case was redundant and incorrect — removed it from both `getNavigationMenuItemLabel` and `getViewNavigationMenuItemLabel`
- Added unit tests covering all navigation menu item type branches

image


FelixMalfait:

Your preview environment is available at: http://bore.pub:23124

Bonapara (approved):

PR #19051, opened 2026-03-27
Refined demo workspace creation skill

lucasbordeau:
- AI can now create, update, and delete view filters with full field-type awareness — `ViewFilterToolsFactory` + `ViewFilterToolProvider` (new tool category `VIEW_FILTER`, permission-gated)
- AI can now create, update, and delete view sorts — `ViewSortToolsFactory` + `ViewSortToolProvider` (new tool category `VIEW_SORT`, permission-gated)
- `create_view` now accepts `calendarFieldName` + `fieldNames` to set view columns at creation time — `ViewToolsFactory` (injects `ViewFieldService`, creates view fields inline)
- Three new standard skills registered: `view-building`, `view-filters-and-sorts`, `custom-objects-cleanup` — `standard-skill.constant.ts` + `create-standard-flat-skill-metadata.util.ts`
- `workspace-demo-seeding` skill reworked: keeps standard objects (People, Companies, Opportunities), adds custom fields and renames the first N records by position instead of creating fully custom objects — preserves existing emails and calendar events on standard objects
- Archived objects no longer appear in the nav menu — `NavigationMenuItemObjectDeactivationListener` listens to `metadata.objectMetadata.updated` and deletes nav items when `isActive` flips to false
- AI chat token retry now tries the current in-memory token before triggering a full renewal, and no longer nulls out the token pair on renewal failure — `useAgentChat` **but this is still not stable**

TODO :

Still a bug on record table widget creation from the skill, seems like we should enforce the label identifier : 

image


packages/twenty-server/src/engine/metadata-modules/navigation-menu-item/listeners/navigation-menu-item-object-deactivation.listener.ts
```diff
@@ -0,0 +1,57 @@
...
+
+type ObjectMetadataUpdatedEventProperties = {
+  updatedFields: string[];
+  diff: Record<string, { before: unknown; after: unknown }>;
+  before: Record<string, unknown>;
+  after: Record<string, unknown>;
+};
+
+@Injectable()
+export class NavigationMenuItemObjectDeactivationListener {
```
FelixMalfait: Not sure if it's the kind of thing we want to do (with apps it will become even more complex!), it feels like cascading this kind of logic makes it hard to restore. Have we considered just hiding these navigation items when displaying them instead of doing a listener?

FelixMalfait: Overall, the less event-based logic the better, it's cause for bugs

packages/twenty-server/src/engine/core-modules/tool-provider/providers/view-sort-tool.provider.ts
```diff
@@ -0,0 +1,98 @@
...
+import {
+  type ToolDescriptor,
+  type ToolIndexEntry,
+} from 'src/engine/core-modules/tool-provider/types/tool-descriptor.type';
+import { toolSetToDescriptors } from 'src/engine/core-modules/tool-provider/utils/tool-set-to-descriptors.util';
+import { PermissionsService } from 'src/engine/metadata-modules/permissions/permissions.service';
+import { ViewSortToolsFactory } from 'src/engine/metadata-modules/view-sort/tools/view-sort-tools.factory';
+
+@Injectable()
+export class ViewSortToolProvider implements ToolProvider, OnModuleInit {
```
FelixMalfait: I think it was a mistake to do a a ToolProvider for View + ViewField, and now one for ViewSort. For me it should be a single View ToolProvider for all View-related tools

packages/twenty-server/src/engine/core-modules/tool-provider/providers/view-filter-tool.provider.ts
```diff
@@ -0,0 +1,98 @@
...
+import {
+  type ToolDescriptor,
+  type ToolIndexEntry,
+} from 'src/engine/core-modules/tool-provider/types/tool-descriptor.type';
+import { toolSetToDescriptors } from 'src/engine/core-modules/tool-provider/utils/tool-set-to-descriptors.util';
+import { PermissionsService } from 'src/engine/metadata-modules/permissions/permissions.service';
+import { ViewFilterToolsFactory } from 'src/engine/metadata-modules/view-filter/tools/view-filter-tools.factory';
+
+@Injectable()
+export class ViewFilterToolProvider implements ToolProvider, OnModuleInit {
```
FelixMalfait: Same comment, I think it doesn't deserve a dedicated provider, we should put everything into one View ToolProvider imo

packages/twenty-server/src/engine/metadata-modules/navigation-menu-item/listeners/navigation-menu-item-object-deactivation.listener.ts
FelixMalfait: Remove all this part

PR #19060, opened 2026-03-28
Handle SSE token expiration silently to avoid Sentry noise

jnMetaCode:
## Summary
- Fixes the SSE client error handler in `useTriggerEventStreamCreation` to detect token expiration errors (401/UNAUTHENTICATED) and handle them silently by triggering event stream reconnection
- Non-token-expiration errors continue to be reported to Sentry as before
- This prevents excessive Sentry error logs caused by normal idle session token expiry, which is a standard reconnection scenario

Fixes #18077

## Test plan
- [ ] Verify that when a session goes idle and the token expires, no error is sent to Sentry
- [ ] Verify that the SSE client reconnects normally after token expiration
- [ ] Verify that genuine SSE errors (non-auth) are still captured in Sentry


charlesBochet:

PR #19062, opened 2026-03-28
Fix address autocomplete duplicating city/state/country in Address 1

jnMetaCode:
## Summary
- Fixes the address autocomplete behavior where selecting a suggestion would store the full address string (e.g., "123 Main St, New York, NY 10001, USA") in the Address 1 field, even though city, state, postal code, and country were already extracted into their dedicated fields
- After extracting address components from place details, the fix strips those components from the Address 1 text and cleans up leftover comma separators, so only the street address remains

Fixes #18860

## Test plan
- [ ] Open an address field and type a partial address in Address 1
- [ ] Select an autocomplete suggestion from the dropdown
- [ ] Verify that Address 1 contains only the street address (e.g., "123 Main St") and not the full string with city/state/country/zip
- [ ] Verify that City, State, Post Code, and Country fields are still correctly populated
- [ ] Test with addresses from different countries to confirm the cleanup handles various formats


charlesBochet:

PR #19064, opened 2026-03-28
fix: resolve CODE step result variables in workflow templates

jnMetaCode:
## Summary
- Fixes workflow template variable substitution for CODE step outputs referenced via `{{step.result.field}}` path
- The `getWorkflowRunContext` function now exposes step results both at the top level (`{{step.field}}`) and under a `result` key (`{{step.result.field}}`), so both access patterns resolve correctly
- Adds tests for the new behavior in both `getWorkflowRunContext` and `variable-resolver`

## Problem
When a CODE step in a workflow produces output (e.g., `{ memberName: "Mike Chen", eventName: "Microsoft Gala" }`), subsequent steps like SEND_EMAIL that reference these values using `{{transform-step-001.result.memberName}}` get `undefined` instead of the actual value. This is because `getWorkflowRunContext` strips the `result` wrapper from step outputs, so only `{{step.field}}` works but `{{step.result.field}}` does not.

## Fix
Modified `getWorkflowRunContext` in `packages/twenty-shared/src/workflow/utils/getWorkflowRunContext.ts` to spread the result object at the top level AND include it under a `result` key. This preserves backward compatibility with existing UI-generated variables (`{{step.field}}`) while also supporting the `{{step.result.field}}` path that users naturally write based on the workflow run state structure.

## Test plan
- [ ] Verify `{{step.field}}` still resolves correctly (backward compatibility)
- [ ] Verify `{{step.result.field}}` now resolves correctly
- [ ] Verify multi-variable strings like `{{step.result.a}} and {{step.result.b}}` work
- [ ] Run existing unit tests for `getWorkflowRunContext` and `variable-resolver`

Fixes #17109

### DIFF
--- a/packages/twenty-shared/src/utils/__tests__/variable-resolver.test.ts
+++ b/packages/twenty-shared/src/utils/__tests__/variable-resolver.test.ts
@@ -130,6 +130,46 @@ describe('resolveInput', () => {
     ).toBe('{ "a": "str" }');
   });
 
+  describe('CODE step output variable resolution', () => {
+    const codeStepContext = {
+      'transform-step-001': {
+        memberName: 'Mike Chen',
+        eventName: 'Microsoft Gala',
+        result: {
+          memberName: 'Mike Chen',
+          eventName: 'Microsoft Gala',
+        },
+      },
+    };
+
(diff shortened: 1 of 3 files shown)


charlesBochet:

PR #19069, opened 2026-03-28
Fix record index sync when view fields arrive via SSE

channi23:
Fixes #19023
## What changed

This updates the record index/view field state flow so the current view can react to late-arriving `viewFields` coming from SSE without requiring a page refresh.

Changes:
- extracted a narrower `syncRecordIndexViewFields` path in `useLoadRecordIndexStates`
- kept the initial full record-index load for first entry into a view
- added a follow-up sync in `RecordIndexLoadBaseOnContextStoreEffect` when the same view receives updated `viewFields`
- updated `ViewBarRecordFieldEffect` so it re-syncs current record fields when `currentView.viewFields` changes instead of only initializing once

## Why

There is a race when a user navigates to a custom object while AI is still creating metadata. In that case, the record index can initialize from a partial view, and later SSE `viewFields` updates were not being applied to the active view state. That could leave the table visually empty or incomplete until a refresh.

## Impact

This should allow:
- record index columns to update live when view fields arrive via SSE
- view bar field state to update live as well
- the current view to stay usable without a refresh while AI-created metadata is still streaming in

## Validation

Validated locally with:
- `npx prettier --check` on modified files
- `npx oxlint --type-aware` on modified files

Manual verification:
- confirmed live record creation appeared without refresh
- manual AI/SSE testing was partially limited by Groq TPM/token caps on the selected model, but the state-sync path was verified in code and local behavior checks

image


charlesBochet (approved):
Changed a bit the strategy! Thank you @channi23

PR #19071, opened 2026-03-28
fix: use AND instead of OR in neq filter for null-equivalent values

oniani1:
Fixes #19070

The `neq` operator in `compute-where-condition-parts.ts` uses `OR` where it should use `AND` when handling null-equivalent values.

Currently generates:
```sql
field != '' OR field IS NOT NULL
```

For a row where `field = ''`:
- `'' != ''` = false
- `'' IS NOT NULL` = true
- `false OR true` = true -- row incorrectly passes the filter

The `eq` operator correctly uses `OR field IS NULL` because it's additive (match value or its null equivalent). By De Morgan's law, the negation `neq` needs `AND field IS NOT NULL` -- exclude if the value doesn't match AND is not a null equivalent.

With the fix:
```sql
field != '' AND field IS NOT NULL
```
- `'' != ''` = false, `'' IS NOT NULL` = true, `false AND true` = false -- correctly excluded
- `NULL != ''` = NULL, `NULL IS NOT NULL` = false, `NULL AND false` = false -- correctly excluded
- `'Alice' != ''` = true, `'Alice' IS NOT NULL` = true, `true AND true` = true -- correctly included

Affects `neq` filters on TEXT fields and all composite sub-fields (firstName, lastName, primaryEmail, primaryPhoneNumber, address sub-fields, etc.) when filtering against null-equivalent values.


etiennejouan (approved):
Hi @oniani1, LGTM Thank you for catching this bug and fixing it !

oniani1:
Hey, just checking in -- this one's been approved for about a week now. Is there anything else needed to get it merged, or does it need someone else to sign off?

etiennejouan:
> Hey, just checking in -- this one's been approved for about a week now. Is there anything else needed to get it merged, or does it need someone else to sign off?

It should have been merged. Maybe a CI issue, i'll handle it

oniani1:
Thanks for handling it!

PR #19088, opened 2026-03-29
feat: implement field calculations for numeric fields

mebishnusahu0595:
## Description
This PR implements **Field Calculations** for numeric fields, allowing users to define Excel-style formulas (e.g., `Price * (1 - Discount / 100)`) in the field settings. Calculated fields are automatically updated on the server-side whenever their dependencies change and are treated as read-only in the UI.

### Key Changes:
- **Metadata**: Added `calculationFormula` property to `FieldMetadataNumberSettings`.
- **Backend**: 
  - Introduced `FieldCalculationService` using `expr-eval` for formula parsing and evaluation.
  - Implemented topological sorting for multi-step calculations and circular dependency detection.
  - Updated `DataArgProcessorService` to trigger calculations during record creation and updates.
- **Frontend**:
  - Added a "Calculation formula" input in the number field settings form.
  - Implemented automatic read-only protection in record forms for calculated fields.

### Testing:
- Added comprehensive unit tests for `FieldCalculationService` covering basic arithmetic, complex formulas, and dependency sorting.
- Verified both backend and frontend builds successfully.

Closes #19087

### DIFF
--- a/packages/twenty-front/src/modules/object-record/read-only/utils/isRecordFieldReadOnly.ts
+++ b/packages/twenty-front/src/modules/object-record/read-only/utils/isRecordFieldReadOnly.ts
@@ -7,7 +7,7 @@ type IsRecordFieldReadOnlyParams = {
   isSystemObject?: boolean;
   fieldMetadataItem: Pick<
     FieldMetadataItem,
-    'id' | 'isUIReadOnly' | 'isCustom'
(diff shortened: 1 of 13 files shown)


FelixMalfait:
This needs deeper product input, if we introduce computed fields I think we would do it for more than just for the numeric fields, it needs to be built in a more abstract way. Thanks for trying this but it's unlikely we accept a feature without prior validation/product work

Issue #19090, opened 2026-03-29
IMAP sync fails with "Mailbox doesn't exist" for \Noselect container folders


John-H-Smith:
## Bug Description
IMAP folder sync calls `STATUS` on folders that have the `\Noselect` flag (pure container/hierarchy folders without actual mailbox storage), causing a warning and failed `uidValidity` lookup.
This happens when a mail client (e.g. Thunderbird) creates a folder that only serves as a parent for subfolders but doesn't hold messages itself. The IMAP server (Dovecot) correctly reports it with `\Noselect` in the `LIST` response, but Twenty still attempts a `STATUS` command on it.

## Steps to Reproduce
1. Connect an IMAP account (Dovecot) to Twenty
2. Have a folder hierarchy where a parent folder is a pure container (no `cur/new/tmp` in Maildir), e.g. `Others` with subfolders `Others.Subfolder1`, `Others.Subfolder2`
3. Trigger email sync
4. Observe worker logs

## Actual Behavior
```
worker-1  | [Nest] 35  - 03/29/2026, 7:57:00 PM    WARN [ImapGetAllFoldersService] Failed to get uidValidity for folder INBOX.Others:
worker-1  | [Nest] 35  - 03/29/2026, 7:57:00 PM    WARN [ImapGetAllFoldersService] Error: Mailbox doesn't exist: INBOX.Others
worker-1  |     at module.exports (/app/node_modules/imapflow/lib/commands/status.js:133:29)
worker-1  |     at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
worker-1  |     at async ImapFlow.run (/app/node_modules/imapflow/lib/imap-flow.js:3263:22)
worker-1  |     at async ImapFlow.status (/app/node_modules/imapflow/lib/imap-flow.js:2085:16)
worker-1  |     at async ImapGetAllFoldersService.getUidValidity (/app/packages/twenty-server/dist/modules/messaging/message-folder-manager/drivers/imap/services/imap-get-all-folders.service.js:102:28)
worker-1  |     at async ImapGetAllFoldersService.filterAndMapFolders (/app/packages/twenty-server/dist/modules/messaging/message-folder-manager/drivers/imap/services/imap-get-all-folders.service.js:58:33)
worker-1  |     at async ImapGetAllFoldersService.getAllMessageFolders (/app/packages/twenty-server/dist/modules/messaging/message-folder-manager/drivers/imap/services/imap-get-all-folders.service.js:32:29)
[...]

PR #19093, opened 2026-03-30
fix(imap): skip STATUS for Noselect folders to avoid mailbox errors

ekkoitac:
Fix Issue #19090: IMAP sync fails with Mailbox does not exist for Noselect container folders.

When syncing IMAP folders, skip calling STATUS on folders with the Noselect flag (pure container/hierarchy folders with no message storage). STATUS fails on these with Mailbox does not exist because they have no cur/new/tmp Maildir directories.

Fix: check for Noselect before calling getUidValidity; if set, skip the call and use null uidValidity (still works for parentFolderId lookups).

Stack trace reference: ImapGetAllFoldersService.filterAndMapFolders calls getUidValidity on every mailbox including Noselect ones.

### DIFF
--- a/packages/twenty-server/src/modules/messaging/message-folder-manager/drivers/imap/services/imap-get-all-folders.service.ts
+++ b/packages/twenty-server/src/modules/messaging/message-folder-manager/drivers/imap/services/imap-get-all-folders.service.ts
@@ -94,7 +94,12 @@ export class ImapGetAllFoldersService implements MessageFolderDriver {
     }
 
     for (const mailbox of mailboxList) {
-      const uidValidity = await this.getUidValidity(client, mailbox);
+      // Skip STATUS for \Noselect folders (container-only folders with no messages).
+      // getUidValidity would fail with "Mailbox doesn't exist" on these.
+      const isNoselect = mailbox.flags?.has('\\Noselect');
+      const uidValidity = isNoselect
+        ? null
+        : await this.getUidValidity(client, mailbox);
       const externalId = uidValidity
         ? `${mailbox.path}:${uidValidity}`
         : mailbox.path;


packages/twenty-server/src/modules/messaging/message-folder-manager/drivers/imap/services/imap-get-all-folders.service.ts
```diff
@@ -94,7 +94,12 @@ export class ImapGetAllFoldersService implements MessageFolderDriver {
     }
 
     for (const mailbox of mailboxList) {
-      const uidValidity = await this.getUidValidity(client, mailbox);
+      // Skip STATUS for \Noselect folders (container-only folders with no messages).
+      // getUidValidity would fail with "Mailbox doesn't exist" on these.
+      const isNoselect = mailbox.flags?.has('\\Noselect');
```
charlesBochet, line 99: I'm a bit lost in the naming: No select is a flag at mailbox level but the comment above is discussing No select folders.

Also it's quite hard to understand the link between isNoSelect and Uid validity concept. Could you explain?

charlesBochet (changes requested):

packages/twenty-server/src/modules/messaging/message-folder-manager/drivers/imap/services/imap-get-all-folders.service.ts
ekkoitac: Hi Charles, great questions — let me clarify:

On naming: You are right that in IMAP, the \Noselect flag is attached to a mailbox (folder) object. The variable name isNoselect is derived from the flag name, not the object type — essentially "does this mailbox have the \Noselect flag set". I can rename it to hasNoselectFlag or isNoselectMailbox if that reads more naturally.

On the link to uidValidity: The root cause is that STATUS (used internally by getUidValidity) calls the filesystem to check cur/new/tmp Maildir subdirectories. A \Noselect folder is a pure hierarchy/container folder — it has no message storage of its own, so those subdirectories do not exist. STATUS then fails with "Mailbox does not exist".

uidValidity exists to detect when a folder contents have been reset/replaced. For a container folder (\Noselect), this concept is moot — it has no messages and no UID state to preserve. So skipping the lookup is both safe and correct.

Does that help? Happy to improve the comment to make this reasoning more explicit in the code as well.

ekkoitac:
Hi Charles, great questions — let me clarify:

**On naming:** You are right that in IMAP, the \\Noselect flag is attached to a mailbox (folder) object. The variable name `isNoselect` is derived from the flag name, not the object type — essentially "does this mailbox have the \\Noselect flag set". I can rename it to `hasNoselectFlag` or `isNoselectMailbox` if that reads more naturally.

**On the link to uidValidity:** The root cause is that STATUS (used internally by getUidValidity) calls the filesystem to check cur/new/tmp Maildir subdirectories. A \\Noselect folder is a pure hierarchy/container folder — it has no message storage of its own, so those subdirectories do not exist. STATUS then fails with "Mailbox does not exist".

uidValidity exists to detect when a folder contents have been reset/replaced. For a container folder (\\Noselect), this concept is moot — it has no messages and no UID state to preserve. So skipping the lookup is both safe and correct.

Does that help? Happy to improve the comment to make this reasoning more explicit in the code as well.

neo773:
> Hi Charles, great questions — let me clarify:
> 
> **On naming:** You are right that in IMAP, the \Noselect flag is attached to a mailbox (folder) object. The variable name `isNoselect` is derived from the flag name, not the object type — essentially "does this mailbox have the \Noselect flag set". I can rename it to `hasNoselectFlag` or `isNoselectMailbox` if that reads more naturally.
> 
> **On the link to uidValidity:** The root cause is that STATUS (used internally by getUidValidity) calls the filesystem to check cur/new/tmp Maildir subdirectories. A \Noselect folder is a pure hierarchy/container folder — it has no message storage of its own, so those subdirectories do not exist. STATUS then fails with "Mailbox does not exist".
> 
> uidValidity exists to detect when a folder contents have been reset/replaced. For a container folder (\Noselect), this concept is moot — it has no messages and no UID state to preserve. So skipping the lookup is both safe and correct.
> 
> Does that help? Happy to improve the comment to make this reasoning more explicit in the code as well.

Closing for AI reply, no human writes 4 em dashes in a normal conversation and please refrain from opening other AI generated PRs

PR #19103, opened 2026-03-30
Fix - Not shared message on record index

etiennejouan:
quality-feedbacks : https://discord.com/channels/1130383047699738754/1486711185436053514/1486711185436053514
Screenshot 2026-03-30 at 09 42 48
Screenshot 2026-03-30 at 09 42 34


packages/twenty-front/src/modules/object-record/record-index/hooks/useHasCurrentViewNonReadableFields.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    const params = {
+      objectMetadataItem,
+      objectMetadataItems,
+      readableFieldIds,
+      objectPermissionsByObjectMetadataId,
+    };
+
+    for (const sort of view.viewSorts) {
+      const reason = getNonReadableFieldInfoForViewField({
```
Weiko: why is it named reason here? 
Actually it seems you renamed getNonReadableFieldReasonForViewField into getNonReadableFieldInfoForViewField but the file is still called getNonReadableFieldReasonForViewField and you use both reason/info
Not really important but probably easy to fix on your end!

packages/twenty-front/src/modules/object-record/record-index/utils/getNonReadableReasonForViewField.ts
```diff
@@ -0,0 +1,87 @@
...
+  objectMetadataItem: EnrichedObjectMetadataItem;
+  objectMetadataItems: EnrichedObjectMetadataItem[];
+  readableFieldIds: Set<string>;
+  objectPermissionsByObjectMetadataId: Record<
+    string,
+    ObjectPermissions & { objectMetadataId: string }
+  >;
+}): { fieldLabel?: string; objectLabel: string } | undefined => {
+  if (!readableFieldIds.has(fieldMetadataId)) {
+    const { fieldMetadataItem: blockedField } = getFieldMetadataItemById({
```
Weiko: nit: Could use a single call and reuse in both branches? Up to you if it seems clearer or not

packages/twenty-front/src/modules/object-record/record-index/utils/getNonReadableReasonForViewField.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  const targetObjectMetadataItem = objectMetadataItems.find(
+    (item) =>
+      item.nameSingular ===
+      fieldMetadataItem.relation?.targetObjectMetadata.nameSingular,
+  );
+
+  if (!isDefined(targetObjectMetadataItem)) {
+    return undefined;
```
Weiko: Not sure if we have a rule for that already. We probably prefer `return;` and it seems you are using both in this file

Weiko (approved):
LGTM!

PR #19105, opened 2026-03-30
Fix warning about ../../tsconfig.base.json not found

martmull:
as title


prastoin (approved):
Indeed that's a way to do :,) the trade off is very minimal
Thanks !!

PR #19109, opened 2026-03-30
Fix record name hover

bosiraphael:
## Before

https://github.com/user-attachments/assets/1972f3ac-17b2-4531-a64e-22f2672de71c

## After

https://github.com/user-attachments/assets/0fffdac6-0fcb-4446-9b7f-1791ec08e2df


Weiko (approved):
LGTM

PR #19114, opened 2026-03-30
[APP] Stricter API assertions

prastoin:
# Introduction

Improved the ci assertions to be sure the default logic function worked fully

## what's next
About to introduce a s3 + lambda e2e test to cover the lambda driver too in addition of the local driver


martmull (approved):

Weiko (approved):
LGTM

PR #19115, opened 2026-03-30
Fix navigation menu edition

bosiraphael:
# Description

Fix "Already in navbar" false positive in the sidebar object picker: `getObjectMetadataIdsInDraft` was collecting `objectMetadataId` from all navigation menu item types (OBJECT, VIEW, RECORD), so having a pinned view or record for an object type would incorrectly block re-adding that object. Now only OBJECT-type items contribute to the duplicate check.

# Video QA

## Before

https://github.com/user-attachments/assets/7b853a55-6d7a-444c-92ee-bf6a75d9927c

## After

https://github.com/user-attachments/assets/65e651e2-9df8-45fd-872d-00782d6b6490


Weiko (approved):
LGTM

PR #19117, opened 2026-03-30
Fix export view

bosiraphael:
# PR Description

**Export view** command sent { id: { in: [] } } to the server, which rejects empty arrays. It now falls back to the view's filters instead when there is no selection, which is the intended behavior.

# Video QA

## Before

https://github.com/user-attachments/assets/3e6263f9-9c66-4f67-a81e-e0571652147d

## After

https://github.com/user-attachments/assets/977a366d-1936-457f-96a1-a5d4fb8ac98e

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-index/export/hooks/useRecordIndexLazyFetchRecords.ts
+++ b/packages/twenty-front/src/modules/object-record/record-index/export/hooks/useRecordIndexLazyFetchRecords.ts
@@ -92,7 +92,11 @@ export const useRecordIndexLazyFetchRecords = ({
     recordIndexId,
   );
 
-  const queryFilter = computeContextStoreFilters({
+  const isEmptySelection =
+    contextStoreTargetedRecordsRule.mode === 'selection' &&
+    contextStoreTargetedRecordsRule.selectedRecordIds.length === 0;
+
+  const contextStoreFilter = computeContextStoreFilters({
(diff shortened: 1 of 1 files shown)


packages/twenty-front/src/modules/object-record/record-index/export/hooks/useRecordIndexLazyFetchRecords.ts
```diff
@@ -92,7 +92,11 @@ export const useRecordIndexLazyFetchRecords = ({
     recordIndexId,
   );
 
-  const queryFilter = computeContextStoreFilters({
+  const isEmptySelection =
+    contextStoreTargetedRecordsRule.mode === 'selection' &&
```
Weiko, line 96: is it still the case if nothing is selected? I would have assumed the logic would have been
```typescript
contextStoreTargetedRecordsRule.mode === 'selection' ? contextStoreFilter : findManyRecordsParams.filter
```

Weiko (approved):
Left a question but otherwise LGTM

packages/twenty-front/src/modules/object-record/record-index/export/hooks/useRecordIndexLazyFetchRecords.ts
Weiko: Addressed

PR #19126, opened 2026-03-30
Fix cropped calendar event name

bosiraphael:
## Before
CleanShot 2026-03-30 at 16 17 13 2@2x

## After
CleanShot 2026-03-30 at 16 16 56@2x


Weiko (approved):
LGTM

PR #19127, opened 2026-03-30
fixed#12847:arrow keys in select field search input on record table

rudra2609-06:
##Issue #12847 fixed

## Fix record table keyboard navigation

* Improve arrow key movement between cells
* Fix incorrect focus behavior
* Handle edge cases properly

This improves overall keyboard usability in the table.


rudra2609-06:
I do have tried local testing but due to consistent error i was unable to open it locally because of .env changes and some configs but issue is clearly fixed accordingly.

PR #19133, opened 2026-03-30
Fix widgets drag handles

bosiraphael:
- Remove drag handles and cursor resize for the placeholder
- Fix the drag handles no longer appearing on hover and always present drag handle for north and south

This was due to the linaria migration: adding a minus sign or px after a css variable isn't working, we have to use calc.

## Before

https://github.com/user-attachments/assets/cba398ab-92c8-4924-ba3c-1a28fb4fd163

## After

https://github.com/user-attachments/assets/aec675cd-b809-434d-a8a7-edb12ce37364


Devessier (approved):
:+1:

Weiko (approved):
LGTM

PR #19136, opened 2026-03-30
Fix docker installer startup ordering and tag alignment

PhilosophiMoonbeam:
## Summary
- keep the installer's Docker image tag and GitHub tag aligned by default
- start `db`, `redis`, and `server` first, then wait for server health before starting `worker`

## Why
On this host, the first boot spends several minutes in the server image entrypoint running database initialization, cache flushes, upgrades, and cron registration before `exec node dist/main`.

With current Docker Compose behavior, `docker compose up -d` for the whole stack fails early because `worker` depends on `server` being healthy. The installer exits before the server finishes first-boot initialization even though the server eventually becomes healthy if left running.

The installer also currently resolves the Docker image tag and GitHub branch independently, which can produce mismatched releases (for example `v1.19.1` image with `v1.19.13` compose/env files).

## Validation
- reproduced the install failure on a fresh self-hosted setup
- verified the server eventually reaches healthy state when started without the worker race
- syntax checked the updated installer with `bash -n packages/twenty-docker/scripts/install.sh`

### DIFF
--- a/packages/twenty-docker/scripts/install.sh
+++ b/packages/twenty-docker/scripts/install.sh
@@ -43,9 +43,11 @@ function on_exit {
 }
 trap on_exit EXIT
 
-# Use environment variables VERSION and BRANCH, with defaults if not set
+# Use environment variables VERSION and BRANCH, with defaults if not set.
+# By default, keep the Docker image tag and GitHub tag aligned to avoid
+# fetching compose/env files from a different release than the image itself.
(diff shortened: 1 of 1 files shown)


BOHEUS:
It'll be handled by core team later in the future as there are few problems with Github and Docker tags, also, Docker tags take priority over Github tags (there were situations in the past where Docker tag was released and Github tag wasn't which created discrepancy and confusion in community)

PR #19140, opened 2026-03-30
feat: add regex pattern validation to custom text fields

intivalderas:
Closes #18655

## Summary

Implements regex pattern validation for custom `TEXT` fields, as requested in #18655. Admins can configure a pattern and optional error message per field; values are validated on both the client and server.

## Changes

**`twenty-shared`**
- `FieldMetadataSettings.ts` — added `validationPattern` and `validationErrorMessage` to `FieldMetadataTextSettings`

**`twenty-server`**
- `validate-text-flat-field-metadata.util.ts` *(new)* — validates the regex is syntactically valid when a field is created/updated
- `flat-field-metadata-type-validator.service.ts` — wired `TEXT` to the new validator
- `validate-text-field-or-throw.util.ts` — enforces the pattern when a record value is saved
- `data-arg-processor.service.ts` — passes `settings` to the text validator for the `TEXT` case

**`twenty-front`**
- `FieldMetadata.ts` — extended `FieldTextMetadata` with `validationPattern` and `validationErrorMessage`
- `SettingsDataModelFieldTextForm.tsx` — toggle in the "Customize" card that reveals pattern + error message inputs; Zod validates regex syntax client-side before saving the field
- `TextFieldInput.tsx` — red border live feedback via `recordFieldInputIsFieldInErrorComponentState` + snackbar error that blocks save on invalid values

## How it works

**Configuring the pattern (admin)**

In Settings → Data model → [Object] → [Text field], a new "Validation pattern" toggle appears in the Customize card. Enabling it reveals:
1. A regex input (e.g. `^BE0[0-9]{9}$`) with inline syntax error if the expression is invalid
2. An optional custom error message shown to users (e.g. `Not a valid Belgian VAT number`)

**Runtime validation (end user)**

- Typing an invalid value shows a red border on the cell (same as URL/phone/email fields)
- Pressing Enter, Tab, or clicking outside blocks the save and shows a snackbar with the custom error message
- Server-side validation also rejects invalid values via the GraphQL API

## Test plan

- [ ] Create a custom text field, enable the validation toggle, set pattern `^BE0[0-9]{9}$` and error message `Not a valid Belgian VAT number`
- [ ] On a record, click the field — type an invalid value (`kkk`) — red border should appear
- [ ] Press Enter — save is blocked, snackbar shows the custom error message
- [ ] Type a valid value (`BE0123456789`) — border clears, Enter saves successfully
- [ ] Disable the toggle — pattern is cleared, field accepts any value again
- [ ] Via GraphQL mutation, attempt to set an invalid value — server should return the custom error message

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/input/components/TextFieldInput.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/input/components/TextFieldInput.tsx
@@ -10,6 +10,8 @@ import { FieldInputContainer } from '@/ui/field/input/components/FieldInputConta
 import { useSnackBar } from '@/ui/feedback/snack-bar-manager/hooks/useSnackBar';
 import { useAvailableComponentInstanceIdOrThrow } from '@/ui/utilities/state/component-state/hooks/useAvailableComponentInstanceIdOrThrow';
 import { useSetAtomComponentState } from '@/ui/utilities/state/jotai/hooks/useSetAtomComponentState';
+import { isNonEmptyString } from '@sniptt/guards';
+import { useLingui } from '@lingui/react/macro';
 import { useContext } from 'react';
 import { turnIntoUndefinedIfWhitespacesOnly } from '~/utils/string/turnIntoUndefinedIfWhitespacesOnly';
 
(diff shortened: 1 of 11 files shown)


intivalderas:
Fixed — the validation now trims the value before testing it against the regex, matching the trimmed value that gets persisted.

intivalderas:
Addressing the concern raised by @BOHEUS in #18655:

- **Empty fields on creation**: validation is skipped when the value is `null` or empty string, so newly created records with unfilled fields won't fail.
- **Partial updates**: the validator only runs on fields explicitly passed in the mutation — updating one field will never trigger validation on another untouched field.

charlesBochet:
Closing this one as this product change is not validated and I'm not sure it should be added to the engine

PR #19143, opened 2026-03-30
fix(server): normalize Asia/Calcutta timezone alias

Misrilal-Sah:
Likely maps to issue #19099 symptoms (timezone error causing empty calendar/group-by).


BOHEUS:
Hey, please don't use AI in pull requests as it doesn't solve a problem you mention, also it adds unnecessary noise. If you want to contribute, please make changes that truly resolve the problem, you tested them by yourself and makes all tests pass.

PR #19144, opened 2026-03-30
fix(server): support legacy RICH_TEXT_V2 field type

Misrilal-Sah:
Targets issue #19141 upgrade breakage.


BOHEUS:
Hey, please don't use AI in pull requests as it doesn't solve a problem you mention, also it adds unnecessary noise. If you want to contribute, please make changes that truly resolve the problem, you tested them by yourself and makes all tests pass.

PR #19146, opened 2026-03-31
Fix workspace invitation email matching

DJnxtgen:
## Summary
- normalize workspace invitation emails before storing them in invitation token context
- compare invitation emails case-insensitively during validation and lookup so mixed-case invites still resolve correctly
- add regression coverage for normalized storage and mixed-case invite acceptance

## Test plan
- [x] Reviewed diff against `origin/main`
- [ ] `npx nx test twenty-server --runInBand --testPathPattern=workspace-invitation.service.spec.ts` currently fails before this spec runs because `twenty-emails:build` cannot resolve `expr-eval-fork` from `twenty-shared/dist/utils.mjs`
- [ ] `npx jest packages/twenty-server/src/engine/core-modules/workspace-invitation/services/workspace-invitation.service.spec.ts --config packages/twenty-server/jest.config.mjs --runInBand` currently fails for the same unrelated `expr-eval-fork` module-resolution issue

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/workspace-invitation/services/workspace-invitation.service.spec.ts
+++ b/packages/twenty-server/src/engine/core-modules/workspace-invitation/services/workspace-invitation.service.spec.ts
@@ -168,6 +168,30 @@ describe('WorkspaceInvitationService', () => {
       ).resolves.not.toThrow();
     });
 
+    it('should normalize the invitation email before saving it', async () => {
+      const email = 'Test@Example.com';
+      const workspace = { id: 'workspace-id' } as WorkspaceEntity;
+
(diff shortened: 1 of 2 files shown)


BOHEUS:
There's already email sanitization on front in WorkspaceInviteTeam.tsx

DJnxtgen:
I did see the frontend sanitation you mentioned. From what i saw though, the bug was that invite creation/acceptance normalized the email on the frontend as lowercase, but the backend still compared the stored invitation email case-sensitively (appToken.context.email !== email) and queried it case-sensitively as well.

BOHEUS:
@DJnxtgen could you please point to an issue you mention? Also, you can reopen this PR with `/twenty pr open` command if you think that this changes should be merged

PR #19156, opened 2026-03-31
fix: allow arrow key cursor movement in Select field search input

jaydeep-pipaliya:
## Summary

Fixes arrow keys (left/right) not moving the cursor in Select and MultiSelect field search inputs on the record table.

**Root cause:** `useSelectableListHotKeys` registers ArrowLeft/ArrowRight handlers with `preventDefault: true` (default), which intercepts the keys before the browser can move the cursor in the search input.

**Fix:** When the event target is an `INPUT` or `TEXTAREA`, skip the selectable list navigation and let the browser handle cursor movement. Also passes `preventDefault: false` for these keys so the default text input behavior is preserved.

Fix #12847

## Test plan

- [ ] Open a record table with a Select or MultiSelect field
- [ ] Click on a Select cell to open the dropdown with search input
- [ ] Type some text in the search input
- [ ] Press Left/Right arrow keys — cursor should now move through the text
- [ ] Press Up/Down arrow keys — should still navigate through the dropdown options
- [ ] Verify arrow keys still work for list navigation when focus is NOT on an input

### DIFF
--- a/packages/twenty-front/src/modules/ui/layout/selectable-list/hooks/internal/useSelectableListHotKeys.ts
+++ b/packages/twenty-front/src/modules/ui/layout/selectable-list/hooks/internal/useSelectableListHotKeys.ts
@@ -152,15 +152,29 @@ export const useSelectableListHotKeys = (
 
   useHotkeysOnFocusedElement({
     keys: Key.ArrowLeft,
-    callback: () => handleSelect('left'),
+    callback: (keyboardEvent) => {
+      const target = keyboardEvent.target as HTMLElement;
+      if (target.tagName === 'INPUT' || target.tagName === 'TEXTAREA') {
+        return;
+      }
+      handleSelect('left');
+    },
     focusId,
     dependencies: [handleSelect],
+    options: { preventDefault: false },
   });
 
   useHotkeysOnFocusedElement({
     keys: Key.ArrowRight,
-    callback: () => handleSelect('right'),
+    callback: (keyboardEvent) => {
+      const target = keyboardEvent.target as HTMLElement;
+      if (target.tagName === 'INPUT' || target.tagName === 'TEXTAREA') {
+        return;
+      }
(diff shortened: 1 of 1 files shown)


thomtrp:
Hi @jaydeep-pipaliya, thank you for your contribution!

jaydeep-pipaliya:
Hi @thomtrp, thanks for the review!

Just checking — do you need live UI screenshots, or is the code-level before/after enough?

Here's the code change:

**Before** (`useSelectableListHotKeys.ts`):
```ts
useHotkeysOnFocusedElement({
  keys: Key.ArrowLeft,
  callback: () => handleSelect('left'),  // always intercepts, even in inputs
  focusId,
  dependencies: [handleSelect],
});
```

**After:**
```ts
useHotkeysOnFocusedElement({
  keys: Key.ArrowLeft,
  callback: (keyboardEvent) => {
    const target = keyboardEvent.target as HTMLElement;
    if (target.tagName === 'INPUT' || target.tagName === 'TEXTAREA') {
      return;  // let browser handle cursor movement
    }
    handleSelect('left');
  },
  focusId,
  dependencies: [handleSelect],
  options: { preventDefault: false },
});
```

The `SelectableList` hotkey handler was intercepting ArrowLeft/ArrowRight with `preventDefault: true` even when focus was on the search input, preventing cursor movement. The fix skips list navigation when the event target is an INPUT or TEXTAREA and sets `preventDefault: false` so the browser's default cursor behavior is preserved.

thomtrp:
Hi @jaydeep-pipaliya, live UI screenshots 🙏

jaydeep-devx:
Hi @thomtrp, I don't currently have the app running locally to record a live demo. I'll set up the dev environment and add the recording. Thanks for your patience!

thomtrp:
@jaydeep-devx feel free to ping me again once you have the recording. Closing for now 👍

PR #19183, opened 2026-03-31
fix: prevent browser scrollbar layout shift on file preview

bugisthegod:
I found this 'bug' when I open the pdf preview. It triggers a transient native scrollbar on the right side, shifting the entire page left and return back quickly. I think hiding the scrollbar would improve the user experience a little. wdyt?
It only happens in pdf. CSV and xlsx is fine, but I haven't test ppt, doc, odt... (It's not allowed to open in local dev).

## Before 

https://github.com/user-attachments/assets/629c7874-b8f9-4a9b-97dc-023bcc8804f4

## After

https://github.com/user-attachments/assets/19ace675-c742-422d-9fe4-eeccde16bb2d

### DIFF
--- a/packages/twenty-front/src/index.css
+++ b/packages/twenty-front/src/index.css
@@ -8,7 +8,6 @@ body {
 
 html {
   font-size: 13px;
-  overflow: hidden;
 }
 
 button {
@@ -22,4 +21,8 @@ form {
 /* https://stackoverflow.com/questions/44543157/how-to-hide-the-google-invisible-recaptcha-badge */
 .grecaptcha-badge {
   visibility: hidden !important;
+}
+
+canvas.hiddenCanvasElement {
+  position: absolute;
 }
--- a/packages/twenty-front/src/index.css
+++ b/packages/twenty-front/src/index.css
@@ -8,6 +8,7 @@ body {
 
 html {
   font-size: 13px;
+  overflow: hidden;
 }
 
 button {


packages/twenty-front/src/index.css
```diff
@@ -8,6 +8,7 @@ body {
 
 html {
   font-size: 13px;
+  overflow: hidden;
```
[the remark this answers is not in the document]
bugisthegod: I don't think so. I stand by my idea.

charlesBochet: this looks to be a heavy change. Could we understand why we only have this issue with file preview? I suspect there is a more local fix

charlesBochet (changes requested):

packages/twenty-front/src/index.css
bugisthegod: @cyntler/react-doc-viewer bundles pdfjs-dist internally but only includes its JavaScript — the following CSS rule from pdf.js's official stylesheet is missing:                                 
                                                                                                   
  canvas.hiddenCanvasElement {
    position: absolute;
  }

image
image

  During PDF rendering, pdf.js calls document.body.append(canvas) to inject a hidden canvas directly into <body>. Without this CSS rule, the canvas remains in normal document flow (position: static). During initialization, it briefly has non-zero dimensions, causing a transient height overflow on <body> that triggers a vertical scrollbar — resulting in a around 15px layout shift across the entire page.

  So I implement this css rules from https://github.com/wojtekmaj/react-pdf/issues/1815                  

image
                             
                                                                                                   
Hope this change does make sense to you.

PR #19218, opened 2026-04-01
Add APP_SECRET_PREVIOUS_KEYS config for JWT key rotation

FelixMalfait:
## Summary
Added support for JWT key rotation by introducing a new configuration variable that allows storing previous APP_SECRET values. This enables a graceful transition period where tokens signed with old keys are still accepted while new tokens are signed with the current key.

## Key Changes
- Added `APP_SECRET_PREVIOUS_KEYS` configuration variable to store comma-separated list of previous secret keys
- Marked as sensitive to prevent accidental exposure in logs
- Set as environment-only configuration for security purposes
- Made optional with empty string as default to maintain backward compatibility
- Tokens signed with previous keys will continue to be accepted during the rotation period

## Implementation Details
- The new variable is part of the SERVER_CONFIG group alongside the main APP_SECRET
- Uses STRING type to support multiple comma-separated values
- Includes descriptive metadata explaining its purpose for key rotation scenarios
- Optional decorator ensures existing deployments without this variable continue to work

https://claude.ai/code/session_0188Xia7v1AtRPe7LULjVrYF


packages/twenty-server/src/engine/core-modules/auth/strategies/jwt.auth.strategy.ts
```diff
@@ -74,6 +92,7 @@ export class JwtAuthStrategy extends PassportStrategy(Strategy, 'jwt') {
       jwtFromRequest: jwtFromRequestFunction,
       ignoreExpiration: false,
       secretOrKeyProvider: secretOrKeyProviderFunction,
+      algorithms: ['HS256', 'ES256'],
```
FelixMalfait, line 95: @charlesBochet

PR #19246, opened 2026-04-02
fix: prepend UTF-8 BOM to fix non-Latin characters in csv

### LINKED ISSUE #19230: Export View doesn't support arabic characterset
1. Export the peoples as CSV with names which has arabic character set
2. Open CSV

Image

bugisthegod:
Fixes: #19230 

CSV exports containing Arabic, Chinese, Japanese, Korean characters displayed as garbled text in Excel because the file lacked a UTF-8 BOM. So I added `\uFEFF` prefix to the CSV Blob so Excel and other spreadsheet apps correctly interpret the file as UTF-8

## Table value
image

## Before
image

## After
image

And those are my test data

Script | Example | Language
-- | -- | --
Arabic | مرحبا | Arabic, Persian, Urdu
Chinese | 你好 | Chinese
Japanese | こんにちは | Japanese
Korean | 안녕하세요 | Korean


FelixMalfait:
Thanks! This is great but I think there's a built-in feature in json-2-csv so I'll push to leverage that

FelixMalfait (approved):

FelixMalfait (dismissed):
Thanks for the fix — the BOM approach is correct and this is the standard solution for Excel + UTF-8 CSVs.

However, the `json-2-csv` library (which this codebase uses) already has a built-in [`excelBOM` option](https://github.com/mrodrig/json-2-csv#json2csv-1) designed for exactly this purpose. Instead of manually prepending `\uFEFF` to the Blob, you can just add `excelBOM: true` to the `json2csv()` call:

```typescript
return json2csv(sanitizedRows, {
  keys,
  emptyFieldValue: '',
  excelBOM: true,
});
```

This is cleaner because:
1. **Uses the library's built-in feature** — no custom `UTF8_BOM` constant or Blob manipulation needed
2. **BOM is part of the CSV string itself**, so it works regardless of how the output is consumed downstream
3. **The generic `downloader` function stays generic** — the current PR adds BOM in `downloader()` which accepts any `mimeType`, meaning non-CSV formats would also get an unwanted BOM
4. **Smaller diff** — one line vs. a new constant + Blob change

I've pushed an alternative implementation to branch `claude/review-pr-19246-wmwU2` if you'd like to reference it. The full change is just adding `excelBOM: true` to the options object at line 114 in `useRecordIndexExportRecords.ts`.

---

bugisthegod:
> Thanks for the fix — the BOM approach is correct and this is the standard solution for Excel + UTF-8 CSVs.
> 
> However, the `json-2-csv` library (which this codebase uses) already has a built-in [`excelBOM` option](https://github.com/mrodrig/json-2-csv#json2csv-1) designed for exactly this purpose. Instead of manually prepending `\uFEFF` to the Blob, you can just add `excelBOM: true` to the `json2csv()` call:
> 
> ```ts
> return json2csv(sanitizedRows, {
>   keys,
>   emptyFieldValue: '',
>   excelBOM: true,
> });
> ```
> 
> This is cleaner because:
> 
> 1. **Uses the library's built-in feature** — no custom `UTF8_BOM` constant or Blob manipulation needed
> 2. **BOM is part of the CSV string itself**, so it works regardless of how the output is consumed downstream
> 3. **The generic `downloader` function stays generic** — the current PR adds BOM in `downloader()` which accepts any `mimeType`, meaning non-CSV formats would also get an unwanted BOM
> 4. **Smaller diff** — one line vs. a new constant + Blob change
> 
> I've pushed an alternative implementation to branch `claude/review-pr-19246-wmwU2` if you'd like to reference it. The full change is just adding `excelBOM: true` to the options object at line 114 in `useRecordIndexExportRecords.ts`.
> 
> _Generated by [Claude Code](https://claude.ai/code)_

Thanks for the review! Good to learn this - I wasn't aware it existed in this codebase. I've updated the implementation to use it instead.

bugisthegod:
Hi @FelixMalfait F, I noticed the tests you added in the last commit might overlap with the csvDownloader tests I already had. Should I remove my csvDownloader describe block to avoid duplication, or would you prefer to keep both of them?

FelixMalfait:
Ah yes sorry indeed! I automated the fix with Claude and didn't noticed we pushed at the same time

PR #19256, opened 2026-04-02
Improve apps

martmull:
- simplify the base application template
- remove --exhaustive option and replace by a --example option like in next.js https://nextjs.org/docs/app/api-reference/cli
- Fix some bugs and logs
- add a post-card app in twenty-apps/examples/


charlesBochet (approved):

.github/workflows/ci-create-app-e2e.yaml
```diff
@@ -107,7 +105,7 @@ jobs:
           create-twenty-app --version
           mkdir -p /tmp/e2e-test-workspace
           cd /tmp/e2e-test-workspace
-          create-twenty-app test-app --exhaustive --display-name "Test App" --description "E2E test app" --skip-local-instance
+          create-twenty-app test-app --example https://github.com/twentyhq/twenty/tree/main/packages/twenty-apps/hello-world --display-name "Test App" --description "E2E test app" --skip-local-instance
```
prastoin: This assumes the `twenty-apps/hello-world` is always up to date
The goal of this ci is to cover that the previously generated exhaustive app would 100% functional, typecheck build dev execute logic function and client. But also covering generated tests etc.

packages/create-twenty-app/src/constants/template/src/constants/universal-identifiers.ts
```diff
@@ -0,0 +1,4 @@
+export const APP_DISPLAY_NAME = 'DISPLAY-NAME-TO-BE-GENERATED';
+export const APP_DESCRIPTION = 'DESCRIPTION-TO-BE-GENERATED';
+export const APPLICATION_UNIVERSAL_IDENTIFIER = 'UUID-TO-BE-GENERATED';
+export const DEFAULT_ROLE_UNIVERSAL_IDENTIFIER = 'UUID-TO-BE-GENERATED';
```
prastoin, line 4: Shouldn't be constant file but a template as it needs to be dynamically generated
Sed/search and replace in this proves it

packages/create-twenty-app/src/constants/template/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "twenty": "twenty",
+    "lint": "oxlint -c .oxlintrc.json .",
+    "lint:fix": "oxlint --fix -c .oxlintrc.json .",
+    "test": "vitest run",
+    "test:watch": "vitest"
+  },
+  "dependencies": {
+    "twenty-client-sdk": "TO-BE-GENERATED",
+    "twenty-sdk": "TO-BE-GENERATED"
+  },
```
prastoin, line 24: Same should be a template not a constant IMO

packages/create-twenty-app/src/constants/template/tsconfig.json
```diff
@@ -33,5 +33,10 @@
     "**/*.test.ts",
     "**/*.spec.ts",
     "**/*.integration-test.ts"
+  ],
+  "references": [
+    {
+      "path": "./tsconfig.spec.json"
+    }
```
prastoin, line 40: That's not a conventional solution style setup
We should have a root tsconfig.json not including anything, that references two sub typescript configuration projects ( spec and build/lib )

packages/create-twenty-app/src/utils/__tests__/app-template.spec.ts
```diff
@@ -14,40 +12,29 @@ jest.mock('fs-extra', () => {
...
-  includeExampleObject: true,
-  includeExampleField: true,
-  includeExampleLogicFunction: true,
-  includeExampleFrontComponent: true,
-  includeExampleView: true,
-  includeExampleNavigationMenuItem: true,
-  includeExampleSkill: true,
-  includeExampleAgent: true,
-  includeExampleIntegrationTest: true,
-};
```
prastoin, line 30: This was overcomplexified

packages/create-twenty-app/src/utils/__tests__/download-example.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+  it('should throw on invalid URL', () => {
+    expect(() => parseGitHubUrl('https://gitlab.com/user/repo')).toThrow(
+      'Invalid GitHub URL',
+    );
+  });
+
+  it('should throw on malformed URL', () => {
+    expect(() => parseGitHubUrl('not-a-url')).toThrow('Invalid GitHub URL');
+  });
+});
```
prastoin: My hunch is that it's too early to start cloning app repo like that:
- Version missmatch, we should infer the app version in order to match a functional joint-dev image. As the app might not be compatible with twenty latest instance 
- We will be cloning an unknown repo, yarn its dependencies on the end user's PC without him approving us doing so. That's a security concern
- What's the plus value versus cloning the repo directly ?

To me that's a blocker

prastoin (changes requested):

packages/twenty-docs/developers/extend/apps/getting-started.mdx
```diff
@@ -148,66 +148,42 @@ my-twenty-app/
...
+```
+
+To start from a more complete example with custom objects, fields, logic functions, front components, and more, use the `--example` flag:
+
+```bash filename="Terminal"
+npx create-twenty-app@latest my-twenty-app --example <name-or-url>
 ```
 
-By default (`--minimal`), only the core files are created: `application-config.ts`, `roles/default-role.ts`, `logic-functions/pre-install.ts`, and `logic-functions/post-install.ts`. Use `--exhaustive` to include all the example files shown above.
+Examples are sourced from the [twenty-apps/examples](https://github.com/twentyhq/twenty/tree/main/packages/twenty-apps/examples) directory on GitHub. You can also pass any GitHub URL as the example value.
```
prastoin: ***Remark:*** I feel like this isn't right, if the end user wants to create an empty aka minimal app he should be providing --minimal
I feel like a lot of people will end up on this documentation as a first step into the twenty app ecosystem
This docs does not get any hint or direction of what all the sexy things can be done through this feature

prastoin (commented):
H

prastoin (approved):
As discussed final detail is the branch to retrieve examples app from, shouldn't be name as highly prone to update and will be desync from the create twenty app

It should be derivated from the create-twenty-app version instead ( assuming we're versioning both the same way )

PR #19258, opened 2026-04-02
App uninstall cleanup

### LINKED ISSUE #16205: In self‑hosted Twenty with local storage, deleted avatars and attachments are not removed from disk
## Bug Description

In a self‑hosted Twenty instance using local storage (filesystem) for files, deleting an avatar or an attachment from the UI removes it logically in the app, but the underlying file is not deleted from the storage folder.
This looks functionally similar to the S3 storage issue described in [#13191](https://github.com/twentyhq/twenty/issues/13191) but happens when using local storage instead of S3

## Observed behavior:

- Workspace is self‑hosted (Docker), storage set to local filesystem (not S3).  
- Create an object (for example: user avatar, note, or record with attachment).  
- Upload an image/file.  
- From the UI, delete the avatar or remove the attachment.  
- After refresh, the avatar/attachment no longer appears in the app, but the corresponding file is still present on disk in the local storage directory (and continues to occupy space).  

## Expected behavior

When an avatar or attachment is deleted from the UI:

- The application should also delete the corresponding file from the local storage directory (or at least mark and clean it up via a background job).  
- Disk usage for file storage should not keep growing due to orphaned files that are no longer referenced by any object.

Twenty version: v1.12.0

Image

Image

prastoin:
# Introduction

Indirectly fixes https://github.com/twentyhq/twenty/issues/16205
TODO: ( implement the local driver delete )


charlesBochet:
Closing, let's reopen once ready!

PR #19276, opened 2026-04-02
Sync record selection with the edit mode

bosiraphael:
## PR description

The previous approach used a detached flag to represent selection mode, which could was not synced sync with actual UI selection state. This change ties the edit mode's selection/fallback behavior directly to whether records are truly selected in the record index.

- Removes `commandMenuItemEditSelectionModeState` (a standalone 'none' | 'selection' atom) and replaces it with `mainContextStoreHasSelectedRecordsSelector`, which derives selection state from the real record selection in the context store.
- Adds `useSelectFirstRecordForEditMode` to programmatically select the first record (table row or kanban card) when toggling to selection mode, and useResetRecordIndexSelection to clear selection when toggling off or exiting edit mode.
- Ensures record selection is properly cleaned up when exiting layout customization mode or closing the side panel.

### Table

https://github.com/user-attachments/assets/1d845809-8369-4872-b3a9-a0281b8e464b

### Board

https://github.com/user-attachments/assets/325c2d58-4ca0-408a-ab4a-66b97d9d70de


packages/twenty-front/src/modules/command-menu-item/server-items/edit/hooks/useSelectFirstRecordForEditMode.ts
Devessier, line 1: Clean 👍

Devessier (approved):
LGTM!

PR #19289, opened 2026-04-02
fix: create company workflow (use psl for domain suffix)

neo773:
Fixes edge case where suffix like `.com.cy` which is TLD for Cyprus would attach person to wrong company

image

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/application/application-package/constants/seed-dependencies/package.json
+++ b/packages/twenty-server/src/engine/core-modules/application/application-package/constants/seed-dependencies/package.json
@@ -38,6 +38,7 @@
     "lodash.snakecase": "^4.1.1",
     "lodash.upperfirst": "^4.3.1",
     "nodemailer": "^8.0.4",
+    "psl": "^1.15.0",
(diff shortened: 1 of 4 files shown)


packages/twenty-server/src/engine/core-modules/application/application-package/constants/seed-dependencies/package.json
```diff
@@ -38,6 +38,7 @@
     "lodash.snakecase": "^4.1.1",
     "lodash.upperfirst": "^4.3.1",
     "nodemailer": "^8.0.4",
+    "psl": "^1.15.0",
```
FelixMalfait, line 41: Wrong package.json?

neo773: Uhm just did what Mart mentioned

image

FelixMalfait: Ah I thought this was for our company creation automation in messaging

FelixMalfait (approved):

PR #19304, opened 2026-04-03
Fix search fallback command menu item

bosiraphael:
The search fallback was treated as a standard action and not as a fallback by the frontend.

## Before

https://github.com/user-attachments/assets/4d2f1e09-b109-46eb-89ff-ca589eeb7239

## After

https://github.com/user-attachments/assets/e4523047-eb1e-4386-9c19-57797e1ef26c


Weiko (approved):
LGTM

PR #19306, opened 2026-04-03
Add redis cache for cron triggers

thomtrp:
- WorkflowCronTriggerCronJob was querying every active workspace (~700) sequentially every minute to find cron triggers, causing regular CPU spikes to 100% on worker pods

- Added a Redis hashset cache that stores cron triggers. On cache miss (TTL expired, cold start, or explicit invalidation), a full scan rebuilds the cache

- Creating/deleting a new cron trigger updates the cache, only if exists

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/cache-storage/services/cache-storage.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/cache-storage/services/cache-storage.service.ts
@@ -342,7 +342,7 @@ end`;
     }) as Promise<number>;
   }
 
-  async hashDeleteIfExists({
+  async hashDelete({
     key,
     field,
   }: {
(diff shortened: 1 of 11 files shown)


packages/twenty-server/src/engine/core-modules/cache-storage/services/cache-storage.service.ts
```diff
@@ -286,6 +286,88 @@ export class CacheStorageService {
...
+  return 0
+end`;
+
+    return redisClient.eval(script, {
+      keys: [this.getKey(key)],
+      arguments: [field, value],
+    }) as Promise<number>;
+  }
+
+  async hashDeleteIfExists({
```
charlesBochet: is that useful?

charlesBochet (approved):

packages/twenty-server/src/engine/core-modules/cache-storage/services/cache-storage.service.ts
thomtrp: no!

PR #19322, opened 2026-04-03
Remove IS_DRAFT_EMAIL_ENABLED feature flag, always request gmail.compose scope

FelixMalfait:
## Summary
- Removes the `IS_DRAFT_EMAIL_ENABLED` feature flag entirely
- Always requests the `gmail.compose` OAuth scope during Google account connection, alongside the already-unconditional `gmail.send` scope
- The draft email workflow action in the frontend is now always available (no longer gated by the flag)

The `gmail.compose` scope was previously only requested when the feature flag was enabled (off by default), which meant email draft creation silently failed with permission errors for most workspaces.

## Test plan
- [ ] Connect a Google account and verify the OAuth consent screen requests the compose scope
- [ ] Create an email draft via the AI chat to confirm it works without errors
- [ ] Verify the "Draft Email" action appears in the workflow action selector without needing the feature flag


FelixMalfait:
@neo773 can you please re-open this (includes the json/html fix) when it's ready / approved by Google? Thanks a lot!

neo773:
hmm too many conflicts better to make a new one

PR #19328, opened 2026-04-04
fix: skip IMAP STATUS on \Noselect folders

DragonBot00:
Fix for #19090.

`filterAndMapFolders()` calls `getUidValidity()` on every mailbox in the loop, but `\Noselect` folders don't support the IMAP `STATUS` command, so it throws `NotFound` on each sync cycle.

The fix checks for `\Noselect` before calling `getUidValidity()`. Container folders still get added to `pathToExternalIdMap` so child folders resolve their parent references correctly.

Added tests for the skip behavior and folder hierarchy preservation.


charlesBochet:
@DragonBot00 Thank you for giving a try, we will go with: 
https://github.com/twentyhq/twenty/pull/19402

Issue #19336, opened 2026-04-04
Ex: In DarkMode, a black text / css missing in the applications permissions


neilk17:
## Bug Description

A clear and concise description of what the current behavior is.
Please also add **screenshots** of the existing application.

**Example:** 
Ex: In DarkMode, a black text / css missing in the applications permissions

<img width="612" height="430" alt="Image" src="https://github.com/user-attachments/assets/4206de6c-1121-4ffe-ac9d-fdf4f4539226" />

## Expected behavior

Use different css for dark mode 

**Example:** 
```
The blank square should be transparent (invisible)
```

neilk17:
I can fix this

BOHEUS:
@neilk17 feel free to make a PR

neilk17:
@BOHEUS 
I created a PR https://github.com/twentyhq/twenty/pull/19340 but this is probably getting fixed when the permissions stuff gets added/

PR #19340, opened 2026-04-05
Fix dark mode text color on permissions tab empty stat (#19336)

neilk17:
Before: black text in dark mode
image

After updated styling: the correct grey text in dark mode
Screenshot 2026-04-05 at 2 47 24 PM


charlesBochet (approved):
Thank you, used the occasion to refactor and apply the fix to other areas

PR #19342, opened 2026-04-05
Fix zh-CN sidebar translations in minimalMetadata API

yukuotec:
The minimalMetadata query was returning untranslated English labels for standard objects (Companies, People, Opportunities, etc.) when users requested zh-CN locale. This fix adds translation logic to MinimalMetadataService using the existing I18nService.

Changes:
- Add translateLabel() method to translate standard object labels
- Pass locale from GraphQL context through resolver to service
- Only translate non-custom objects (custom objects use user-defined labels)

### DIFF
--- a/packages/twenty-server/src/engine/metadata-modules/minimal-metadata/minimal-metadata.resolver.ts
+++ b/packages/twenty-server/src/engine/metadata-modules/minimal-metadata/minimal-metadata.resolver.ts
@@ -1,7 +1,8 @@
 import { UseGuards } from '@nestjs/common';
-import { Query } from '@nestjs/graphql';
+import { Context, Query } from '@nestjs/graphql';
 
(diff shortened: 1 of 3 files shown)


charlesBochet:
Thanks for the PR! I've pushed some changes on top of yours:

- The `translateLabel()` method was reimplementing logic that already exists in `resolveObjectMetadataStandardOverride()` (used by `ObjectMetadataResolver`, `ViewResolver`, and `ViewController`). The duplicate missed `standardOverrides.translations` and per-key override fallbacks, so I refactored to reuse the shared util instead.
- Added an integration test (`minimal-metadata-i18n.integration-spec.ts`) that sets locale to `fr-FR` and asserts translated labels on the `minimalMetadata` query — we didn't have coverage for this path.
- Cleaned up import ordering in the resolver.

charlesBochet (approved):

PR #19343, opened 2026-04-05
fix: invalidate GraphQL schema cache after metadata changes

deliveredbyai:
- Add WorkspaceCacheStorageService to RestApiMetadataService
- Invalidate GraphQL schema cache after create/update/delete operations
- Add comprehensive unit tests for cache invalidation
- Fix issue where custom objects created via Metadata API don't appear in GraphQL schema

This ensures that when custom objects are created, updated, or deleted through the Metadata API, the GraphQL schema cache is properly invalidated, making the changes immediately available through GraphQL queries.


charlesBochet:
HI @deliveredbyai 

Thanks for opening a PR. I don't understand the issue though, here is my point of view on existing implementation:
- The REST metadata API already gets full cache invalidation for free. 

The call chain is:
- RestApiMetadataService.create() builds a GraphQL mutation and posts it to the app's own /metadata endpoint
- That hits ObjectMetadataResolver.createOneObject()
- Which calls ObjectMetadataService.createOneObject()
- Which calls WorkspaceMigrationValidateBuildAndRunService.validateBuildAndRunWorkspaceMigration()
- Which calls WorkspaceMigrationRunnerService.run()
- Which calls invalidateCache() — bumping metadata version, flushing flat entity maps, recomputing ORM schemas, resolver name maps, etc.

So the extra flushVersionedMetadata call in the PR is redundant. The runner already handles all of this. And since the GraphQL SDL cache is keyed by (workspaceId, metadataVersion), bumping the version (which the runner does) is sufficient — the next request will compute a fresh SDL against the new version.

On integration tests: there are extensive integration specs for object metadata CRUD through the GraphQL metadata API (test/integration/metadata/suites/object-metadata/), covering create, update, delete, rename, relations, side-effects, etc.

In short, this PR solves a problem that doesn't exist (the runner already invalidates)

I'm going to close it, feel free to comment on it if I'm missing anything

PR #19344, opened 2026-04-05
fix: replace npm pkg set with node script in set-local-version target

charlesBochet:
## Summary

- Replaces `npm pkg set version=X` with a direct `node -e` script in the `set-local-version` Nx target
- Fixes `CI Create App E2E minimal` workflow failures on fork PRs where `npm pkg set` fails with `EJSONPARSE: Unexpected end of JSON input while parsing empty string`

The `npm` command has unreliable `package.json` resolution in certain CI checkout contexts (shallow clones of fork PR merge refs). Using `node` directly to read/write the JSON file avoids this entirely.

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -141,7 +141,7 @@
       "cache": false,
       "options": {
         "cwd": "{projectRoot}",
-        "command": "npm pkg set version={args.releaseVersion}"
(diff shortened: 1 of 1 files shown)


FelixMalfait (approved):

prastoin (commented):
In the first place thought that it was coming from being run in // between interdependent repos thorugh nx run-many
Using jq might easier here

PR #19348, opened 2026-04-05
Migrate messageFolder parentFolderId from UUID to externalId

neo773:
This PR migrates `messageFolder`.`parentFolderId` from an internal `UUID` reference to external provider id.
Eliminates unnecessary lookup and complexity


charlesBochet (approved):
LGTM

PR #19356, opened 2026-04-06
Instance commands and `upgrade_migrations` table

prastoin:
# Introduction
Now only using typeorm to generate migrations up and down statement
We handle and maintain our own migration table history

## What's new
Now all the instance commands will live within the same module and folder than the upgrade commands
Sequentiality comes from the timestamp located in the filename
Same sequentiality also applies to the workspace commands in the future, for the moment still expected a as code explicit declaration

( below screen is an example see below section )
image

## Existing 1.21 migrations
We won't start following this pattern in 1.21 yet at least not with the migration that has already been released as typeorm migrations in cloud production as they would rerun

## Small duplication
Duplicating the legacy typeorm and instance commands run in the `run-instance-commands` to avoid any merge of interest for the moment 

## Concurrency
Not handling any run in parrallel of the upgrade for the moment


packages/twenty-server/src/engine/core-modules/upgrade/upgrade-migration.entity.ts
```diff
@@ -0,0 +1,31 @@
...
+  name: string;
+
+  @Column({ type: 'varchar', nullable: false })
+  status: UpgradeMigrationStatus;
+
+  @Column({ type: 'integer', nullable: false, default: 1 })
+  attempt: number;
+
+  @Column({ type: 'varchar', nullable: false })
+  runByVersion: string;
```
charlesBochet: naming

charlesBochet (approved):

PR #19358, opened 2026-04-06
feat: allow custom app tabs on system object record pages

8Maverik8:
Currently, the Page Layout SDK allows defining custom tabs with FRONT_COMPONENT or IFRAME widgets via `definePageLayout`, but two blockers prevent them from appearing on system objects (Company, Opportunity, etc.):

1. Frontend: `PageLayoutTabsRenderer` filters tabs on system objects to a hardcoded allowlist (Home, Timeline, Overview, Flow), hiding any custom app tabs. This patch extends the filter to also show tabs containing FRONT_COMPONENT or IFRAME widgets.

2. Backend: `validateWidgetConfigurationInput` throws "not supported yet" for FIELDS, TIMELINE, TASKS, NOTES, FILES, EMAILS, and CALENDAR widget types, blocking SDK apps from defining page layouts that include these standard record page widgets alongside custom components. This patch allows these types to pass validation.

These changes enable Twenty SDK apps to add custom tabs to system object record pages — a common use case for CRM extensions (e.g., deal item builders, document generators on Opportunity pages).

### DIFF
--- a/packages/twenty-front/src/modules/page-layout/components/PageLayoutTabsRenderer.tsx
+++ b/packages/twenty-front/src/modules/page-layout/components/PageLayoutTabsRenderer.tsx
@@ -133,9 +133,16 @@ export const PageLayoutTabsRenderer = () => {
 
   const SYSTEM_OBJECT_TABS = ['Home', 'Timeline', 'Overview', 'Flow'];
 
+  const hasCustomAppWidget = (tab: { widgets: Array<{ type: string }> }) =>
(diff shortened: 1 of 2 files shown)


charlesBochet:
@8Maverik8 Thanks for opening a PR, it is much appreciated. Regarding the specific issue you are trying to solve, it's a bit early. The pageLayout experience is not ready yet and this should be supported out of the box once ready

PR #19372, opened 2026-04-07
fix: merge preview blank page when dry-run relation loading fails

### LINKED ISSUE #19312: Merging records page is empty if there are 2 records with the same name
**Scenario:**
1. Log in
2. Go to Companies (or any other record)
3. Create 2 records with the same name
4. Select both records and one additional record with different name
5. Open command menu
6. Choose "Merge records" option

**Actual:** Preview of merged record is empty

Image

**Expected:** Preview of merged record must show details the result of merge

NathanDrake2406:
## Summary

Fixes #19312

- **Server**: Wrap dry-run `processNestedRelationsHelper.processNestedRelations` in try-catch so the preview returns merged field data even when nested relation queries fail
- **Frontend**: Remove `errorPolicy: 'ignore'` from dry-run mutation, add `previewError` state, render error placeholder instead of blank page

### DIFF
--- a/packages/twenty-front/src/modules/object-record/hooks/useMergeManyRecords.ts
+++ b/packages/twenty-front/src/modules/object-record/hooks/useMergeManyRecords.ts
@@ -85,10 +85,9 @@ export const useMergeManyRecords = <
             conflictPriorityIndex: mergeSettings.conflictPriorityIndex,
             dryRun: preview,
           },
-          // Prevent cache updates during dry run to avoid overwriting original record data
+          // no-cache prevents dry-run data from overwriting live records in Apollo cache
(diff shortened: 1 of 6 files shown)


charlesBochet:
@NathanDrake2406 Thanks for opening a PR, it's much appreciated but this is not fixing the root cause. We should try to understand what's wrong here, it looks to be a race condition

NathanDrake2406:
> @NathanDrake2406 Thanks for opening a PR, it's much appreciated but this is not fixing the root cause. We should try to understand what's wrong here, it looks to be a race condition

I'm curious as well. Couldn't reproduce it and just gave up tracing it lol.

PR #19386, opened 2026-04-07
App feedbacks fix option id required in apps

martmull:
fixes https://discord.com/channels/1130383047699738754/1488226371032453292


packages/twenty-sdk/src/cli/utilities/build/manifest/utils/add-missing-field-option-ids.ts
```diff
@@ -0,0 +1,31 @@
...
+  if (fieldManifest.options === undefined || fieldManifest.options === null) {
+    return fieldManifest;
+  }
+
+  return {
+    ...fieldManifest,
+    options: fieldManifest.options.map((option) => ({
+      ...option,
+      id: option.id ?? v4(),
+    })),
```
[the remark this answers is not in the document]
prastoin: true

packages/twenty-sdk/src/cli/utilities/build/manifest/manifest-build.ts
```diff
@@ -145,7 +146,9 @@ export const buildManifest = async (
 
         const objectManifest: ObjectManifest = {
           ...extract.config,
-          fields: objectFieldsWithDefaults,
+          fields: objectFieldsWithDefaults.map((field) =>
```
prastoin: ***Polish:*** could pass function directly

packages/twenty-sdk/src/cli/__tests__/integration/utils/normalize-manifest.util.ts
```diff
@@ -1,10 +1,42 @@
...
+
+  if (field.options === undefined || field.options === null) {
+    return field;
+  }
+
+  return {
+    ...field,
+    options: field.options.map((option) => ({
+      ...option,
+      id: '[option-id]',
```
prastoin: I don't understand this, why not derivating the uuid from the parent field universal and option name instead if not provided ?

charlesBochet (commented):
My point of view:
- we don't know how to update options (as we don't have universalId, appId, etc. we don't know who owns the option aka if a user has changed them)
- we should only support SELECT / MULTISELECT creation / deletion in our app sync (or in the buildAndRun)
- therefore we should not treat option ids as universalIds, they are more like the id of an object which is workspace specific and out of scope here, therefore I would completely drop option ids

Vision post v2 launch:
- make option synchronizable (with universalIds) in the JSON (like we do for universalConfigurations & co)

prastoin (commented):
Hey there,
Regarding the option id being optional agreed with that but I don't think we should send a place holder but directly compute then deterministically from the parent field universal id and option name

As said by Sentry right now any app sync will lead in any enum field type option update

As discussed with Charles we might wanna totally disable the enum field type update synchronization in the non custom app ecosystem until we've refactor the option to be sync metadata entities themselves

packages/twenty-sdk/src/cli/__tests__/integration/utils/normalize-manifest.util.ts
martmull: we generate uuid with v4, it avoids to have mock uuid generators during comparaison

prastoin (approved):

PR #19394, opened 2026-04-07
fix(messaging): create messageThread.subject field metadata + column in 1-21 backfill

FelixMalfait:
## Summary
- The 1-21 \`upgrade:1-21:backfill-message-thread-subject\` command assumed the legacy \`sync-metadata\` flow would create the new \`messageThread.subject\` field metadata and column on existing workspaces. That sync was removed, so the column was never added and the backfill silently skipped.
- The command now ensures the field exists by computing the standard \`messageThread.subject\` flat field from the twenty-standard application and running it through \`WorkspaceMigrationValidateBuildAndRunService\` (same pattern used by the page-layout / command-menu-item backfills). This creates both the field metadata row and the workspace schema column.
- After ensuring the field, the existing \`UPDATE messageThread SET subject = ...\` runs as before.

## Test plan
- [ ] On a workspace with no \`subject\` column on \`messageThread\`, run \`yarn command:prod upgrade:1-21:backfill-message-thread-subject\` and confirm:
  - the field metadata row is created in \`core.\"fieldMetadata\"\`
  - the \`subject\` column is created on \`workspace_<id>.messageThread\`
  - existing message threads are backfilled from the most recent message
- [ ] Re-run on the same workspace and confirm it is a no-op (field already exists, no rows to update)
- [ ] Run on a workspace that already has the column but \`NULL\` subjects and confirm only the backfill runs

### DIFF
--- a/packages/twenty-server/src/database/commands/upgrade-version-command/1-21/1-21-workspace-command-backfill-message-thread-subject.command.ts
+++ b/packages/twenty-server/src/database/commands/upgrade-version-command/1-21/1-21-workspace-command-backfill-message-thread-subject.command.ts
@@ -1,18 +1,31 @@
 import { Command } from 'nest-commander';
+import { isDefined } from 'twenty-shared/utils';
+import { v4 } from 'uuid';
 
 import { ActiveOrSuspendedWorkspaceCommandRunner } from 'src/database/commands/command-runners/active-or-suspended-workspace.command-runner';
 import { WorkspaceIteratorService } from 'src/database/commands/command-runners/workspace-iterator.service';
 import { type RunOnWorkspaceArgs } from 'src/database/commands/command-runners/workspace.command-runner';
+import { ApplicationService } from 'src/engine/core-modules/application/application.service';
+import { type FlatFieldMetadata } from 'src/engine/metadata-modules/flat-field-metadata/types/flat-field-metadata.type';
+import { WorkspaceCacheService } from 'src/engine/workspace-cache/services/workspace-cache.service';
 import { getWorkspaceSchemaName } from 'src/engine/workspace-datasource/utils/get-workspace-schema-name.util';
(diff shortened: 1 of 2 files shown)


packages/twenty-server/src/database/commands/upgrade-version-command/1-21/1-21-workspace-command-backfill-message-thread-subject.command.ts
```diff
@@ -75,4 +74,125 @@ export class BackfillMessageThreadSubjectCommand extends ActiveOrSuspendedWorksp
...
+        'flatFieldMetadataMaps',
+      ]);
+
+    const messageThreadObject = Object.values(
+      flatObjectMetadataMaps.byUniversalIdentifier,
+    )
+      .filter(isDefined)
+      .find(
+        (object) =>
+          object.nameSingular === MESSAGE_THREAD_OBJECT_NAME_SINGULAR,
```
charlesBochet: nitpick: usually easier to find by universalId (we might also have utils to ease that)

packages/twenty-server/src/database/commands/upgrade-version-command/1-21/1-21-workspace-command-backfill-message-thread-subject.command.ts
```diff
@@ -75,4 +69,77 @@ export class BackfillMessageThreadSubjectCommand extends ActiveOrSuspendedWorksp
...
+
+      return;
+    }
+
+    const { twentyStandardFlatApplication } =
+      await this.applicationService.findWorkspaceTwentyStandardAndCustomApplicationOrThrow(
+        { workspaceId },
+      );
+
+    await this.fieldMetadataService.createOneField({
```
charlesBochet: why not using the validateBuildAndRun?

neo773:
We need to remove all references to workspace:sync-metadata from internal documentation, LLMs keep hallucinating it from time to time, I'm making a PR for this

FelixMalfait:
@neo773 exactly! This + the messaging being in both schema led to a few hallucinations I didn't catch during review, I thought I had catched all but no 😮‍💨

PR #19397, opened 2026-04-07
Workspace command decorators

prastoin:
# Introduction
Migrating the workspace commands to the decorator version + timestamp listing as for the instance commands

We've now been able to remove the upgrade command abstraction where we needed to import all modules and order them
Now they're dynamically retrieved at upgrade runtime, sorted by timestamp

## Instance and workspace commands name
The name is computed from the command metadata `version` `className` and `timestamp` we have a duplicate validation at module init from the unified registry


packages/twenty-server/src/database/commands/upgrade-version-command/__tests__/upgrade.command.spec.ts
```diff
@@ -546,16 +540,6 @@ describe('UpgradeCommandRunner', () => {
             'APP_VERSION is not defined, please double check your env variables',
         },
       },
-      {
-        title: 'when current version commands are not found',
```
Weiko, line 550: This is not tested anymore?

Weiko (approved):
LGTM! Nice! 👏

PR #19400, opened 2026-04-07
docs: remove reference to sync metadata

neo773:


packages/twenty-docs/developers/self-host/capabilities/upgrade-guide.mdx
```diff
@@ -371,12 +371,10 @@ Run the following commands:
 
 ```
 yarn database:migrate:prod
-yarn command:prod workspace:sync-metadata -f
```
Weiko, line 374: technically the command still exists at this stage (if users upgrade incrementally via git versions) so I think we can keep it here. We might clean up older versions upgrade instructions at some point though

Weiko (approved):
Small comment otherwise LGTM!

PR #19413, opened 2026-04-07
Add twenty sync command

martmull:
Add one shot app synchronisation `twenty sync command` command

Complementary with `twenty app dev` command which is watch mode

Fixes https://discord.com/channels/1130383047699738754/1489644493106839663

### DIFF
--- a/packages/create-twenty-app/package.json
+++ b/packages/create-twenty-app/package.json
@@ -1,6 +1,6 @@
 {
   "name": "create-twenty-app",
-  "version": "0.8.0",
+  "version": "0.9.0-canary.0",
   "description": "Command-line interface to create Twenty application",
   "main": "dist/cli.cjs",
   "bin": "dist/cli.cjs",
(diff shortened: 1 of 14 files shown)


charlesBochet (commented):
I'm not sure I like the fact that we introduce yet another command. I understand that this is quite different from a code perspective to watch vs to build as a one shot but it still feels off from an external perspective

Proposal:
yarn twenty dev --watch (which is the default / optional)
yarn twenty dev --once

Regarding the operation, we can keep it but call it dev. It won't be used in the yarn twenty dev (as it does not makes sense, we need the orchestrator to watch) but will be used by dev --once so it can be leveraged by CLIs

martmull:
`dev` does not make a lot of sense 
what about doing `twenty sync` and `twenty sync --watch`? (we can still support `twenty dev` as a legacy command)

charlesBochet (approved):

PR #19417, opened 2026-04-07
App feedbacks front

martmull:
## Before
image

## After
image


packages/twenty-front/src/modules/settings/logic-functions/components/SettingsLogicFunctionsFieldItemTableRow.tsx
```diff
@@ -20,23 +27,54 @@ export const SettingsLogicFunctionsFieldItemTableRow = ({
   to: string;
 }) => {
   const { theme } = useContext(ThemeContext);
+
+  const computeTrigger = () => {
```
charlesBochet, line 31: nitpick: should likely be an util with clearer naming (computeLogicFunctionTriggerShortname)

charlesBochet (approved):

PR #19427, opened 2026-04-08
Fix AI model registry staleness on self-hosted instances

### LINKED ISSUE #19422: On self-hosted instances, AI features fail with "No AI models are available.
API keys set as environment variables work correctly because process.env is available immediately at startup.

Steps to reproduce

1. Self-host Twenty (e.g. on Railway)
2. Go to Settings → AI → Models
3. Enter a Mistral (or any other) API key and enable the models
4. Try using AI Chat → error: "No AI models are available"

Expected behavior
API keys saved via the admin panel should be picked up by the model registry on the next request or server restart.

Workaround 
Set API keys as environment variables directly on the server (e.g. MISTRAL_API_KEY=... in Railway variables) instead of using the admin panel.

FelixMalfait:
## Summary

Fixes #19422.

Self-hosted users hitting "No AI models are available" after configuring API keys via the admin panel were victims of a stale `AiModelRegistryService` cache. Only the four `addAiProvider` / `removeAiProvider` / `addModelToProvider` / `removeModelFromProvider` mutations called `refreshRegistry()` — setting an API key through `set/update/deleteDatabaseConfigVariable` left the registry pointing at the pre-mutation provider state.

Rather than patch each mutation site (and re-introduce the same class of bug on the next one), the registry now invalidates lazily based on the LLM config-group hash, mirroring the pattern `WebSearchDriverFactory` already uses via `DriverFactoryBase`. Any mutation to an LLM-tagged config variable is picked up automatically on the next read — callers never have to remember to refresh.

- Extracted `getConfigGroupHash` into a shared util reused by both `DriverFactoryBase` and `AiModelRegistryService` (and switched the hash to `JSON.stringify` so object-typed config vars like `AI_PROVIDERS` actually contribute meaningfully).
- `AiModelRegistryService` gates all internal `Map` access behind private getters that call `ensureFresh()`, so future read paths can't accidentally observe stale state. Build path uses underscored backing fields directly to avoid recursing.
- Dropped the now-redundant `refreshRegistry()` public method and its four call sites in `admin-panel.resolver.ts`.

## Test plan

- [x] Existing admin-panel + ai-models specs pass (79 tests)
- [ ] Manual: on a self-hosted instance, set `OPENAI_API_KEY` (or another provider key) via the admin panel `setDatabaseConfigVariable` mutation and confirm AI features work without a server restart
- [ ] Manual: existing flows (`addAiProvider`, `addModelToProvider`, etc.) still pick up new providers on the next read


packages/twenty-server/src/engine/metadata-modules/ai/ai-models/services/ai-model-registry.service.ts
```diff
@@ -44,26 +47,66 @@ export interface RegisteredAIModel {
 @Injectable()
 export class AiModelRegistryService {
   private readonly logger = new Logger(AiModelRegistryService.name);
-  private modelRegistry: Map<string, RegisteredAIModel> = new Map();
-  private modelConfigCache: Map<string, AIModelConfig> = new Map();
-  private providerModelDefCache: Map<
+  private _modelRegistry: Map<string, RegisteredAIModel> = new Map();
```
charlesBochet: we don't follow this convention in the project

packages/twenty-server/src/engine/core-modules/twenty-config/dynamic-factory.base.ts
```diff
@@ -44,28 +41,7 @@ export abstract class DriverFactoryBase<TDriver> {
...
-  private getConfigVariablesByGroup(
-    group: ConfigVariablesGroup,
-  ): Array<keyof ConfigVariables> {
-    const metadata =
-      TypedReflect.getMetadata('config-variables', ConfigVariables) ?? {};
-
-    return Object.keys(metadata)
-      .filter((key) => metadata[key]?.group === group)
-      .map((key) => key as keyof ConfigVariables);
+    return getConfigGroupHash(this.twentyConfigService, group);
```
charlesBochet: we should not pass services to utils, it's a service itself then and should use injection

charlesBochet (commented):
I'm not sure to understand what was really wrong with the previous method. If ai provider are updated with the admin panel, we invalidate (like it was done). If they update env variables, this supposed that the container needs to be restarted. I don't think it's a good idea to try to hot load env variables and won't scale to other variables

packages/twenty-server/src/engine/core-modules/twenty-config/dynamic-factory.base.ts
FelixMalfait: Ah yes my bad, good point

FelixMalfait:
@charlesBochet you can also edit via the traditional config-var edition path (not just the AI tab) ; the hash-based approach is standard and already used for other drivers (email etc). Will double check to be sure but the overall approach makes sense to me!

packages/twenty-server/src/engine/metadata-modules/ai/ai-models/services/ai-model-registry.service.ts
```diff
@@ -44,26 +47,66 @@ export interface RegisteredAIModel {
...
+    string,
+    { providerName: string; modelDef: AiProviderModelConfig }
+  > {
+    this.ensureFresh();
+
+    return this._providerModelDefCache;
+  }
+
+  private ensureFresh(): void {
+    const configKey = getConfigGroupHash(
```
charlesBochet: naming: configKey is actually a hash?

charlesBochet:
@FelixMalfait Ok, got it, to reformulate if I'm getting it right:
- write path, we need to call refreshRegistry() when we set it via the admin panel (meaning that we kind of need a side effect mecanism when updating env vars)
- read path, the service reading the config is a singleton and we store the hash locally in the service and compare everytime we fetch

=> the read path is already in place in simpler forms in other areas so you also go with read path here.

LGTM

charlesBochet:
Let's just fix the small comments (non idomatic + naming + utils getting a service as parameter)

charlesBochet (approved):

PR #19428, opened 2026-04-08
Add a lock on function creation

thomtrp:
Fixes https://twenty-v7.sentry.io/issues/7368127776/?environment=prod&project=4507072499810304&query=is%3Aunresolved&referrer=issue-stream

- Add a distributed lock (via CacheLockService) around the Lambda executor build in LambdaDriver to prevent concurrent workflow runs from racing on CreateFunctionCommand for the same logic function
- Use double-checked locking: isAlreadyBuilt is checked lock-free first (fast path), then re-checked inside the lock to avoid redundant rebuilds

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/lambda.driver.ts
+++ b/packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/lambda.driver.ts
@@ -905,9 +905,9 @@ export class LambdaDriver implements LogicFunctionDriver {
       return;
     }
 
-    const buildLockTtlMs = 30_000;
+    const buildLockTtlMs = 120_000;
     const buildLockRetryMs = 500;
-    const buildLockMaxRetries = 10;
+    const buildLockMaxRetries = 240;
 
     await this.cacheLockService.withLock(
       async () => {
--- a/packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/lambda.driver.ts
+++ b/packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/lambda.driver.ts
@@ -1,6 +1,6 @@
 import { createHash } from 'crypto';
 import * as fs from 'fs/promises';
-import { resolve, join } from 'path';
+import { join, resolve } from 'path';
 
 import {
   CreateFunctionCommand,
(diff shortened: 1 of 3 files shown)


charlesBochet (approved):

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/lambda.driver.ts
```diff
@@ -892,17 +895,66 @@ export class LambdaDriver implements LogicFunctionDriver {
...
+    const buildLockTtlMs = 30_000;
+    const buildLockRetryMs = 500;
+    const buildLockMaxRetries = 10;
+
+    await this.cacheLockService.withLock(
+      async () => {
+        // Need to check again inside the lock in case lock was not acquired immediately.
+        if (await this.canSkipBuild(buildArgs)) {
+          return;
+        }
```
martmull, line 917: looks weird

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/lambda.driver.ts
martmull, line 888: can't we use `@WithLock` decorator with `flatApplicationId` parameter?

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/lambda.driver.ts
```diff
@@ -892,17 +895,66 @@ export class LambdaDriver implements LogicFunctionDriver {
...
+      flatLogicFunction,
+      flatApplication,
+      applicationUniversalIdentifier,
+    };
+
+    if (await this.canSkipBuild(buildArgs)) {
+      return;
+    }
+
+    const buildLockTtlMs = 30_000;
```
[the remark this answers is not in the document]
Weiko: +1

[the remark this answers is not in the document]
Weiko: To address imho

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/lambda.driver.ts
thomtrp: why so? First we check if the function has already been built. If not, we try to acquire the lock. But we don't know yet if another worker is currently holding the lock and building the function. Maybe we will only acquire the lock after some time and the function would have been already built

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/lambda.driver.ts
thomtrp: decorator inject the cacheLockService. Lambda driver does not support dep injections. That's why I am passing the service through args

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/lambda.driver.ts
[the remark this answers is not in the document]
thomtrp: nice I see you both conspire

prastoin:
Very cool job

PR #19433, opened 2026-04-08
Fix overridable entities logic for SSE

Weiko:
## Context
SSE metadata events for overridable entities (viewField, viewFieldGroup, pageLayoutWidget, pageLayoutTab) were sending raw flat entity data without resolving overrides into base properties or filtering isActive: false entities. This caused the frontend to display stale/incorrect values (e.g., a hidden viewField still appearing visible).

## Implementation
Add a sanitization step in MetadataEventPublisher.enrichMetadataEventBatch that resolves overrides and converts isActive transitions into the appropriate event types (deactivated -> delete, reactivated -> create), matching what GraphQL resolvers already do

### DIFF
--- a/packages/twenty-front/src/modules/page-layout/widgets/record-table/hooks/useSaveRecordTableWidgetsViewDataOnDashboardSave.ts
+++ b/packages/twenty-front/src/modules/page-layout/widgets/record-table/hooks/useSaveRecordTableWidgetsViewDataOnDashboardSave.ts
@@ -186,10 +186,12 @@ export const useSaveRecordTableWidgetsViewDataOnDashboardSave = () => {
           mapRecordFieldToViewFieldWithCurrentAggregateOperation,
         );
 
+        const existingViewFields = currentView.viewFields ?? [];
+
         const { viewFieldsToCreate, viewFieldsToUpdate } =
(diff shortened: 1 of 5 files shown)


prastoin (commented):
Challenging the fact not to have override grain in the sse metadata events, not sure how we handle the actual field and object standardOverrides. Feels a bit like a presentation endpoint logic

Though no strong opinion on it !

Weiko:
> Challenging the fact not to have override grain in the sse metadata events, not sure how we handle the actual field and object standardOverrides. Feels a bit like a presentation endpoint logic
> 
> Though no strong opinion on it !

Yes, I totally agree and that was my concern too. 
However, I feel like until we have a strong reason to expose overrides vs original values to clients I prefer to keep it simple on their side and align with resolver logic.
This is rarely exposed with other products I've checked - this is almost always "hidden". In fact, without the "reset" feature we are about to implement consumers wouldn't be really aware of "overrides" layer (and again I don't see a case where you want the original value - We've had this same discussion with product about use cases for extensibility/apps as well and we didn't see any, yet)

prastoin (approved):
As discussed ! LGTM

ps: might wanna double check and uniformize the object and field metadata sse overrides behavior too

PR #19439, opened 2026-04-08
fix: return 404 for deleted workspace webhook race

### LINKED ISSUE #15544: Workspace Not Found: Workspace not found for workspace
Workspace no longer exists but webhook controller is still pinged. Need to update error handling to not be sent in Sentry ?

Sentry Issue: [TWENTY-SERVER-5SM](https://twenty-v7.sentry.io/issues/6606088674/?referrer=github_integration)

```
Workspace Not Found: Workspace not found for workspace c63a75bc-5c89-4bc8-9b44-7159598b7d01
  File "../../../../../src/engine/twenty-orm/factories/workspace-datasource.factory.ts", line 343, in {WorkspaceDatasourceFactory#2}.getWorkspaceMetadataVersionFromCacheOrFromDB
    throw new TwentyORMException(
  File "../../../../../src/engine/twenty-orm/factories/workspace-datasource.factory.ts", line 88, in {WorkspaceDatasourceFactory#2}.create
    await this.getWorkspaceMetadataVersionFromCacheOrFromDB(workspaceId);
  File "../../../../src/engine/twenty-orm/twenty-orm-global.manager.ts", line 44, in {TwentyORMGlobalManager#2}.getRepositoryForWorkspace
    await this.workspaceDataSourceFactory.create(workspaceId);
  File "../../../../../../src/engine/core-modules/workflow/controllers/workflow-trigger.controller.ts", line 75, in {WorkflowTriggerController#2}.runWorkflow
    await this.twentyORMGlobalManager.getRepositoryForWorkspace<WorkflowWorkspaceEntity>(
  File "../../../../../../src/engine/core-modules/workflow/controllers/workflow-trigger.controller.ts", line 49, in {WorkflowTriggerController#2}.runWorkflowByPostRequest
    return await this.runWorkflow({
...
(1 additional frame(s) were not displayed)
```

Step to reproduce

https://github.com/user-attachments/assets/c8ac2bf4-6ad6-4e49-85bf-2dcde80f41c1

If waiting deleteWorkspace ends

https://github.com/user-attachments/assets/f4271f20-55c3-4e50-83d9-41732ed4b5fa

hussainarslan:
Handle late TwentyORM workspace-not-found exceptions in the workflow webhook REST exception filter so deleted workspaces return a 404 instead of surfacing as internal errors.

Also add a focused regression spec covering the deleted-workspace ORM codes and the existing workflow-trigger status mappings.

Closes #15544


charlesBochet:
@hussainarslan thanks for opening a PR. To address the race condition, we should instead try catch the call to twentyORM in the function and throw a WorkflowTriggerExceptionCode.NOT_FOUND, then it will naturally flow to the exception filter.

Regarding the spec, we don't usually write those as they don't bring a lot of value (especially nowaday with AI, developers will just change the test and never actually run it to detect regression). Our policy is: unit test on complex business logic, integration tests on API.
Here it will be very tricky to reproduce the race condition and I don't think it's really worth it, so let's just drop the test entirely

hussainarslan:
> @hussainarslan thanks for opening a PR. To address the race condition, we should instead try catch the call to twentyORM in the function and throw a WorkflowTriggerExceptionCode.NOT_FOUND, then it will naturally flow to the exception filter.
> 
> Regarding the spec, we don't usually write those as they don't bring a lot of value (especially nowaday with AI, developers will just change the test and never actually run it to detect regression). Our policy is: unit test on complex business logic, integration tests on API. Here it will be very tricky to reproduce the race condition and I don't think it's really worth it, so let's just drop the test entirely

@charlesBochet Made some changes according to your recommendations. Let me know if anything further is required to close this.

charlesBochet (approved):

charlesBochet:
Thank you!

PR #19441, opened 2026-04-08
Introduce updateWorkspaceMemberSettings and clarify product

ijreilly:
## Summary

Introduces a dedicated **metadata** mutation to update **standard (non-custom)** workspace member settings, moves profile-related UI to use it, and aligns **workspace member** record permissions with the rest of the CRM so users cannot escalate visibility via RLS by editing their own member record.

### Profile and appearance (standard fields)

- Users can still update **their own** standard workspace member fields that the product exposes in **Settings / Profile** (e.g. name, locale, color scheme, avatar flow) via the new **`updateWorkspaceMemberSettings`** mutation.
- The mutation returns a **boolean**; the app **merges** the updated fields into local state so the UI stays in sync without refetching the full workspace member record.
- **Locale** changes also keep **`userWorkspace`** in sync when a locale is present in the payload (including from the workspace `updateOne` path when applicable).

### Custom fields on workspace members

- The dedicated metadata mutation **rejects** any **custom** workspace member field (and unknown keys). Those updates must go through the normal **object** `updateOne` pipeline, which is subject to **object- and field-level** permissions like other records. But since we don't have object- and field-level permission configuration for system objects yet, this permission is derived from Workspace member settings permission.
- **Workspace member** is no longer exempt from ORM permission validation for updates merely because it is a **system** object. Users who **do not** have workspace member access (e.g. no **Workspace members** settings permission and no equivalent broad settings access on the role) **cannot** use `updateOne` on `workspaceMember` to change **custom** (or other) fields on their own row—even though that row is used for RLS predicates.
- This closes a path where someone could widen what they can see by writing to fields that drive row-level rules.

### Who can change another member

- Updating **another** user’s workspace member still requires **Workspace members** (or equivalent) settings permission, consistent with admin tooling.


charlesBochet:
@ijreilly Another proposal:
- Settings should use dedicated endpoint on metadata talking about userWorkspace / user (it's a close API like signIn, signUp, etc.) and not update workspaceMember on /graphql. This closed endpoint can edit workspaceMember as a side effect by bypassing permissions
- Then on /graphql we don't have the settings case anymore, we only need pre-hook / post-hook (exisitng) to avoid mutating things that are tied to the userWorkspace (theme, emails, etc.) <=> maybe we only authorize updating customs here in the pre-hook
- in the ORM / permission layer, when we compute the objectPermissions, we fake the one for workspaceMember by checking the settingsPermission to updateWorkspaceMember

I think this is actually simpler and aligned with the vision
@Weiko @FelixMalfait FYI

packages/twenty-server/src/modules/workspace-member/query-hooks/workspace-member-update-one.pre-query.hook.ts
```diff
@@ -40,45 +28,14 @@ export class WorkspaceMemberUpdateOnePreQueryHook
...
-      await this.userWorkspaceRepository.save({
-        ...userWorkspace,
-        locale: payload.data.locale,
-      });
-
-      await this.coreEntityCacheService.invalidate(
-        'userWorkspaceEntity',
-        authContext.userWorkspaceId,
       );
     }
```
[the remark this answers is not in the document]
ijreilly: right, updated this !

packages/twenty-front/src/modules/settings/profile/utils/mergeWorkspaceMemberSettingsIntoCurrent.ts
```diff
@@ -0,0 +1,102 @@
...
+    'name' in payload &&
+    payload.name !== null &&
+    typeof payload.name === 'object'
+  ) {
+    const namePayload = payload.name as Record<string, unknown>;
+    next = {
+      ...next,
+      name: {
+        ...previous.name,
+        ...(typeof namePayload.firstName === 'string'
```
charlesBochet: use typeguards

packages/twenty-front/src/modules/settings/profile/utils/mergeWorkspaceMemberSettingsIntoCurrent.ts
```diff
@@ -0,0 +1,102 @@
...
+): CurrentWorkspaceMember => {
+  const payload = update as Record<string, unknown>;
+  let next: CurrentWorkspaceMember = { ...previous };
+
+  if (
+    'name' in payload &&
+    payload.name !== null &&
+    typeof payload.name === 'object'
+  ) {
+    const namePayload = payload.name as Record<string, unknown>;
```
charlesBochet: remove as

packages/twenty-front/src/modules/settings/profile/utils/mergeWorkspaceMemberSettingsIntoCurrent.ts
```diff
@@ -0,0 +1,102 @@
...
+      ...next,
+      avatarUrl: value === '' || value === null ? null : (value as string),
+    };
+  }
+
+  if ('timeZone' in payload && payload.timeZone !== undefined) {
+    next = { ...next, timeZone: payload.timeZone as string | null };
+  }
+
+  if ('dateFormat' in payload && payload.dateFormat !== undefined) {
```
charlesBochet: use typeguard everywhere isDefined

packages/twenty-server/src/engine/core-modules/user-workspace/user-workspace.service.ts
```diff
@@ -62,10 +63,81 @@ export class UserWorkspaceService extends TypeOrmQueryService<UserWorkspaceEntit
...
+      locale,
+    });
+
+    await this.coreEntityCacheService.invalidate(
+      'userWorkspaceEntity',
+      userWorkspaceId,
+    );
+  }
+
+  async syncUserWorkspaceLocaleForWorkspaceMember({
```
charlesBochet: updateUserWorkspaceLocaleForWorkspaceMember

packages/twenty-server/src/engine/core-modules/user-workspace/user-workspace.service.ts
```diff
@@ -62,10 +63,81 @@ export class UserWorkspaceService extends TypeOrmQueryService<UserWorkspaceEntit
...
   ) {
     super(userWorkspaceRepository);
   }
 
+  async syncUserWorkspaceLocale({
+    locale,
+    userWorkspaceId,
+    throwIfUserWorkspaceMissing = false,
+  }: {
+    locale: UserWorkspaceEntity['locale'] | undefined;
```
charlesBochet: don't allow undefined, this simplifies. the caller should not call if undefined

packages/twenty-server/src/engine/core-modules/user/utils/assert-workspace-member-update-non-custom-fields.util.ts
```diff
@@ -0,0 +1,119 @@
+import { msg } from '@lingui/core/macro';
+
+import { STANDARD_OBJECTS } from 'twenty-shared/metadata';
+import { isDefined } from 'twenty-shared/utils';
+
+import { UserInputError } from 'src/engine/core-modules/graphql/utils/graphql-errors.util';
+import { WorkspaceManyOrAllFlatEntityMapsCacheService } from 'src/engine/metadata-modules/flat-entity/services/workspace-many-or-all-flat-entity-maps-cache.service';
+import { type FlatFieldMetadata } from 'src/engine/metadata-modules/flat-field-metadata/types/flat-field-metadata.type';
+
+const getRelationJoinColumnName = (
```
charlesBochet: this code should already exist somewhere else

charlesBochet: I would do an allow list for this whole file

packages/twenty-server/src/engine/metadata-modules/role/services/workspace-roles-permissions-cache.service.ts
```diff
@@ -137,47 +139,66 @@ export class WorkspaceRolesPermissionsCacheService extends WorkspaceCacheProvide
...
+                role,
+                rolePermissionFlags,
+                PermissionFlagType.WORKSPACE_MEMBERS,
+              );
 
-          const getPermissionValue = (
-            overrideValue: boolean | undefined,
-            defaultValue: boolean,
-          ) => (isSystem ? true : (overrideValue ?? defaultValue));
+            canRead = hasWorkspaceMembersPermissions;
```
charlesBochet: canRead = true (always)

packages/twenty-server/src/engine/core-modules/user/user.resolver.ts
```diff
@@ -437,6 +442,91 @@ export class UserResolver {
...
+        id: workspaceMember.id,
+        ...(input.update as Partial<WorkspaceMemberWorkspaceEntity>),
+      };
+
+    await this.globalWorkspaceOrmManager.executeInWorkspaceContext(() =>
+      workspaceMemberRepository.save(workspaceMemberUpdatePayload),
+    );
+
+    if (isDefined(input.update.locale)) {
+      await this.userWorkspaceService.syncUserWorkspaceLocaleForWorkspaceMember(
```
charlesBochet: as we have the workspaceMember, why not passing the userWorkspaceId directly
I'm not sure about the value of  this.userWorkspaceService.syncUserWorkspaceLocaleForWorkspaceMember

charlesBochet: at least it's a bit weird to do that much in the resolver and then redo that much in the service. Either we have the resolver super light and pass its input to the service, either we do a lot in the resolver and keep service light, here I feel we are over complexifiying

packages/twenty-server/src/modules/workspace-member/query-hooks/workspace-member-update-one.pre-query.hook.ts
```diff
@@ -45,40 +34,25 @@ export class WorkspaceMemberUpdateOnePreQueryHook
...
-          : undefined,
       },
     );
 
     // TODO: remove this code once we have migrated locale update to userWorkspace update
-    if (payload.data.locale && isUserAuthContext(authContext)) {
-      const userWorkspace = await this.userWorkspaceRepository.findOne({
-        where: {
-          id: authContext.userWorkspaceId,
+    if (isUserAuthContext(authContext) && isDefined(payload.data.locale)) {
```
charlesBochet: remove

packages/twenty-front/src/modules/settings/profile/hooks/useUpdateWorkspaceMemberSettings.ts
```diff
@@ -0,0 +1,69 @@
...
+  type CurrentWorkspaceMember,
+  currentWorkspaceMemberState,
+} from '@/auth/states/currentWorkspaceMemberState';
+import {
+  mergeWorkspaceMemberSettingsIntoCurrent,
+  type WorkspaceMemberSettingsUpdateInput,
+} from '@/settings/profile/utils/mergeWorkspaceMemberSettingsIntoCurrent';
+import { useSetAtomState } from '@/ui/utilities/state/jotai/hooks/useSetAtomState';
+
+const UPDATE_WORKSPACE_MEMBER_SETTINGS_MUTATION = gql`
```
charlesBochet: should go in its own file (graphql/queries)

charlesBochet (approved):

PR #19453, opened 2026-04-08
Reset Tab Page Layout

Weiko:
## Context
- Add "Reset to default" for page layout tabs and widgets — backend mutation resets overrides, reactivates deactivated children, and deletes custom children
- Fix override write bug where mutating an overridable property (e.g. widget title) on a standard-app entity incorrectly overwrote the base column instead of writing to the overrides JSONB — PageLayoutUpdateService now uses resolveFlatEntityOverridableProperties for accurate diffing and routes properties through sanitizeOverridableEntityInput
- Deprecate isOverridden - We need more time to think about this feature. Currently this adds too much complexity for a very small benefit

https://github.com/user-attachments/assets/a84546c8-1e15-4d9e-a489-0825cf8b8ed2


packages/twenty-server/src/engine/metadata-modules/page-layout/services/page-layout-reset.service.ts
```diff
@@ -187,6 +198,303 @@ export class PageLayoutResetService {
     return fromFlatPageLayoutWidgetToPageLayoutWidgetDto(updatedWidget);
   }
 
+  async resetPageLayoutTabToDefault({
```
Weiko, line 201: Next PR will introduce resetPageLayout (including tab and widgets) and I'll take this chance to refactor this whole service

packages/twenty-server/src/engine/metadata-modules/page-layout/services/page-layout-update.service.ts
```diff
@@ -346,11 +375,33 @@ export class PageLayoutUpdateService {
...
+        });
+
+        const editableProperties = {
           title: tabInput.title,
           position: tabInput.position,
+          ...(tabInput.icon !== undefined && { icon: tabInput.icon }),
           layoutMode: tabInput.layoutMode ?? existingTab.layoutMode,
+        };
+
+        const { overrides, updatedEditableProperties } =
```
charlesBochet, line 393: is that aligned with how it's done for objects?

packages/twenty-server/src/engine/metadata-modules/page-layout/services/page-layout-update.service.ts
```diff
@@ -346,11 +375,33 @@ export class PageLayoutUpdateService {
...
+            workspaceCustomApplicationUniversalIdentifier,
+          entityApplicationUniversalIdentifier:
+            existingTab.applicationUniversalIdentifier,
+          workspaceCustomApplicationUniversalIdentifier,
+        });
+
+        const editableProperties = {
           title: tabInput.title,
           position: tabInput.position,
+          ...(tabInput.icon !== undefined && { icon: tabInput.icon }),
```
charlesBochet, line 389: use isDefined

packages/twenty-server/src/engine/metadata-modules/view-field-group/dtos/view-field-group.dto.ts
```diff
@@ -38,7 +38,10 @@ export class ViewFieldGroupDTO {
   @Field(() => [ViewFieldDTO])
   viewFields?: ViewFieldDTO[];
 
-  @Field(() => Boolean, { nullable: false })
+  @Field(() => Boolean, {
+    nullable: false,
+    deprecationReason: 'isOverridden is deprecated',
```
charlesBochet, line 43: are we sure we want to remove this everywhere?

packages/twenty-server/src/engine/metadata-modules/utils/resolve-flat-entity-overridable-properties.util.ts
```diff
@@ -0,0 +1,21 @@
+import { isDefined } from 'twenty-shared/utils';
+
+type FlatEntityWithOverrides = {
+  [key: string]: unknown;
+  overrides: Record<string, unknown> | null;
+};
+
+export const resolveFlatEntityOverridableProperties = <
```
charlesBochet, line 8: naming

charlesBochet, line 8: applyFlatEntityOverrides

charlesBochet (approved):

packages/twenty-server/src/engine/metadata-modules/page-layout/services/page-layout-update.service.ts
Weiko: Not sure! I'll have to double check

packages/twenty-server/src/engine/metadata-modules/view-field-group/dtos/view-field-group.dto.ts
Weiko: Yes, this is too complex for a small benefit, we can go back to it after v2 for sure!

PR #19458, opened 2026-04-08
Fix workspace command name inserted in db

prastoin:
# Introduction
Persist in the db the registry computed name that contains version_classname_timestamp to be consistent with the instance command pattern too


Weiko (approved):
LGTM

PR #19459, opened 2026-04-08
Fix:19331 :  It's not possible to change the stage of an opportunity by moving it in the Kanban view

hemant838:
The change fixes a mismatch between two update paths.

When you edit an opportunity directly, the app sends only stage, so the board knows “this record may belong in a different column” and reloads. But when you drag a card in kanban, the app sends both stage and position. The old logic treated any update containing position as a pure reorder and skipped the board refresh entirely. That meant a cross-column move could be optimistically shown first, then the board state stayed stale and the card appeared to jump back.

What I changed is:

I moved the refresh decision into a pure helper at shouldInitializeRecordBoardFromUpdateInputs.ts.
I updated useGetShouldInitializeRecordBoardForUpdateInputs.ts to use that helper instead of the old if (updatedFieldNames.has('position')) return false rule.
The new behavior is:

position only: no board reinit, because that’s a same-column reorder.
stage only: reinit.
stage + position: reinit.
filtered field + position: reinit.
sorted field + position: reinit.
unrelated field + position: no reinit.
I also added tests:

shouldInitializeRecordBoardFromUpdateInputs.test.ts covers the decision matrix directly.
useGetShouldInitializeRecordBoardForUpdateInputs.test.tsx adds a regression test for the kanban-style stage + position update.
So the intent was to keep fast optimistic drag behavior for real reorders, while making cross-stage kanban moves refresh correctly and stop snapping back.

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-board/hooks/__tests__/useGetShouldInitializeRecordBoardForUpdateInputs.test.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-board/hooks/__tests__/useGetShouldInitializeRecordBoardForUpdateInputs.test.tsx
@@ -1,0 +1,80 @@
+import { renderHook } from '@testing-library/react';
+
+import { useActiveFieldMetadataItems } from '@/object-metadata/hooks/useActiveFieldMetadataItems';
+import { type FieldMetadataItem } from '@/object-metadata/types/FieldMetadataItem';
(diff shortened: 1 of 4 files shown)


thomtrp:
Hi @hemant838, thanks for contributing! Could you please fix your CI + provide before/after videos? 
When moving an opportunity between columns I see the stage getting updated so I wonder what's requiring a fix

thomtrp:
Cannot reproduce. Issue may have been solved already

PR #19460, opened 2026-04-08
Fix workspace export for indirect core entity relations

neo773:
Include indirectly linked core records in workspace exports such as users and index field metadata


charlesBochet (commented):
@neo773 This makes the code too complex IMO.

Is this issue for tables that don't have workspaceIds? which ones?
I would advocate to write an explicit logic for these tables if needed rather than complexiify all the other ones

PR #19462, opened 2026-04-08
Fix pre post logic function not executed

martmull:
- removes pre-install function 
- execute **asyncrhonously** post-install function at application installation
- add optional `shouldRunOnVersionUpgrade` boolean value on post-install function definition default false
- update PostInstallPayload to 
```
export type PostInstallPayload = {
  previousVersion?: string;
  newVersion: string;
};
```


packages/twenty-apps/examples/postcard/package.json
```diff
@@ -19,8 +19,8 @@
     "test:watch": "vitest"
   },
   "dependencies": {
-    "twenty-client-sdk": "0.8.0-canary.8",
-    "twenty-sdk": "0.8.0-canary.8"
+    "twenty-client-sdk": "latest",
+    "twenty-sdk": "latest"
```
prastoin: should we ?

packages/twenty-apps/examples/postcard/src/logic-functions/post-install.ts
```diff
@@ -0,0 +1,35 @@
...
+  return {};
+};
+
+export default definePostInstallLogicFunction({
+  universalIdentifier: '852c6321-1563-4396-b7c5-9d370f3d30a9',
+  name: 'post-install',
+  description: 'Runs after installation to set up the application.',
+  timeoutSeconds: 300,
+  handler,
+});
```
prastoin, line 35: lets add a specific step in the ci of this app that checks that these two has been created, or lets just create a new test in its integration tests

packages/twenty-sdk/src/cli/commands/app-command.ts
```diff
@@ -183,22 +182,20 @@ export const registerCommands = (program: Command): void => {
       async (
         appPath?: string,
         options?: {
-          preInstall?: boolean;
```
prastoin, line 186: ***Question:*** Why's that ?

packages/twenty-sdk/src/cli/utilities/build/manifest/manifest-build.ts
```diff
@@ -227,20 +230,17 @@ export const buildManifest = async (
         logicFunctions.push(config);
         logicFunctionsFilePaths.push(relativePath);
 
-        if (
-          targetFunctionName === TargetFunction.DefinePreInstallLogicFunction
-        ) {
-          preInstallLogicFunctionUniversalIdentifiers.push(
-            extract.config.universalIdentifier,
-          );
-        }
```
prastoin, line 239: Have we completely remove the pre install hook ?

packages/twenty-server/src/engine/core-modules/application/application-install/application-install.service.ts
```diff
@@ -109,7 +121,24 @@ export class ApplicationInstallService {
...
+          workspaceId: params.workspaceId,
+        });
+
+      const previousVersion = existingApplication?.version ?? undefined;
+
+      const newVersion = resolvedPackage.packageJson.version;
+
+      const isVersionUpgrade = isDefined(existingApplication);
+
       const { application, wasCreated } = await this.ensureApplicationExists({
```
prastoin, line 146: I feel like we should guard that you can't downgrade an app version once a higher has been installed

prastoin (commented):
Hey there nice job ! Left few comments and would like to discuss the motivation of the pre-install removal

packages/twenty-sdk/src/cli/commands/app-command.ts
martmull: we d'ont need the preInstall option for exec command because pre-isntall logic function is removed

packages/twenty-server/src/engine/core-modules/application/application-install/application-install.service.ts
martmull: Indeed, see this -> https://github.com/twentyhq/twenty/pull/19481

packages/twenty-apps/examples/postcard/package.json
martmull: don't really know how to handle this tbh

prastoin: me too lol, in the ci process it will always be an equivalent of the portal using the verdaccio locally published iamges
As it's getting overrided it does not really matter

packages/twenty-server/src/engine/core-modules/application/application-install/application-install.service.ts
prastoin: Currently being discussed

.github/workflows/ci-create-app-e2e-postcard.yaml
```diff
@@ -161,6 +161,24 @@ jobs:
...
+          echo "$LOGIC_FUNCTIONS_RESPONSE" | jq -e '.data.findManyLogicFunctions | map(select(.name == "post-install")) | length == 1' > /dev/null
+
+          # Verify the post-install function seeded 2 post card records during install
+          POST_CARDS_RESPONSE=$(curl -s -X POST "${{ env.TWENTY_API_URL }}/graphql" \
+            -H "Authorization: Bearer ${{ env.TWENTY_API_KEY }}" \
+            -H "Content-Type: application/json" \
+            -d '{"query":"query { postCards { edges { node { id name content } } } }"}')
+          echo "Post cards response: $POST_CARDS_RESPONSE"
+          echo "$POST_CARDS_RESPONSE" | jq -e '.data.postCards.edges | length == 2' > /dev/null
+
```
prastoin: Sorry again I feel like we should rather create an integration tests rather than making a call from the ci in the end
We could use the client to do so

prastoin (commented):
As discussed with @charlesBochet we prefer keeping the pre-install in the end

packages/twenty-docs/developers/extend/apps/building.mdx
```diff
@@ -672,45 +716,127 @@ yarn twenty exec --preInstall
...
   handler,
 });
 ```
 
-You can also manually execute the post-install function at any time using the CLI:
+**Use `pre-install` when a migration would otherwise destroy or corrupt existing data.** Because pre-install runs against the *previous* schema and its failure rolls back the upgrade, it is the right place for anything risky:
 
-```bash filename="Terminal"
-yarn twenty exec --postInstall
+- **Backing up data that is about to be dropped or restructured** — e.g. you are removing a field in v2 and need to copy its values into another field or export them to storage before the migration runs.
```
charlesBochet, line 778: I'm not sure we need to get into such details but LGTM at this point

charlesBochet (approved):

PR #19473, opened 2026-04-08
Avoid reporting recoverable SSE token expiry reconnects

corazzione:
## Summary
- treat unauthenticated and authorization-related SSE stream creation errors as recoverable reconnect cases
- avoid reporting those expected reconnect paths to Sentry during normal token expiry handling
- add focused coverage for the UNAUTHENTICATED SSE message path

Fixes #18077

## Validation
- corepack yarn exec jest --config packages/twenty-front/jest.config.mjs packages/twenty-front/src/modules/sse-db-event/hooks/__tests__/useTriggerEventStreamCreation.test.tsx --runInBand
- corepack yarn prettier --check packages/twenty-front/src/modules/sse-db-event/hooks/useTriggerEventStreamCreation.ts packages/twenty-front/src/modules/sse-db-event/hooks/__tests__/useTriggerEventStreamCreation.test.tsx

## Notes
- The repo currently enforces Node ^24.5.0; this environment is on 24.1.0, so I ran the commands with YARN_IGNORE_NODE=1 to get targeted validation running.

### DIFF
--- a/packages/twenty-front/src/modules/sse-db-event/hooks/useTriggerEventStreamCreation.ts
+++ b/packages/twenty-front/src/modules/sse-db-event/hooks/useTriggerEventStreamCreation.ts
@@ -20,6 +20,19 @@ import { isDefined } from 'twenty-shared/utils';
 import { v4 } from 'uuid';
 import { type EventSubscription } from '~/generated-metadata/graphql';
 
+const isRecoverableEventStreamError = ({
+  subCode,
+  code,
+}: {
+  subCode?: string;
+  code?: string;
+}) =>
+  subCode === 'EVENT_STREAM_DOES_NOT_EXIST' ||
+  subCode === 'EVENT_STREAM_ALREADY_EXISTS' ||
+  subCode === 'NOT_AUTHORIZED' ||
+  code === 'UNAUTHENTICATED' ||
+  code === 'FORBIDDEN';
+
 export const useTriggerEventStreamCreation = () => {
   const store = useStore();
   const setIsCreatingSseEventStream = useSetAtomState(
@@ -80,9 +93,17 @@ export const useTriggerEventStreamCreation = () => {
           }>,
         ) => {
           if (isDefined(value?.errors) && Array.isArray(value.errors)) {
-            captureException(
-              new Error(`SSE subscription error: ${value.errors[0]?.message}`),
-            );
+            const subCode = value.errors[0]?.extensions?.subCode;
+            const code = value.errors[0]?.extensions?.code;
+
+            if (!isRecoverableEventStreamError({ subCode, code })) {
+              captureException(
+                new Error(
+                  `SSE subscription error: ${value.errors[0]?.message}`,
+                ),
+              );
+            }
+
             store.set(shouldDestroyEventStreamState.atom, true);
 
             return;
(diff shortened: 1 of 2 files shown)


packages/twenty-front/src/modules/sse-db-event/hooks/useTriggerEventStreamCreation.ts
```diff
@@ -129,16 +150,11 @@ export const useTriggerEventStreamCreation = () => {
             if (event === 'next') {
               if (isDefined(result?.errors)) {
                 const subCode = result.errors[0]?.extensions?.subCode;
+                const code = result.errors[0]?.extensions?.code;
 
-                switch (subCode) {
-                  case 'EVENT_STREAM_ALREADY_EXISTS': {
-                    store.set(shouldDestroyEventStreamState.atom, true);
```
thomtrp, line 135: I see that we lose that set here. Would you know why?

packages/twenty-front/src/modules/sse-db-event/hooks/useTriggerEventStreamCreation.ts
```diff
@@ -20,6 +20,19 @@ import { isDefined } from 'twenty-shared/utils';
...
+  code,
+}: {
+  subCode?: string;
+  code?: string;
+}) =>
+  subCode === 'EVENT_STREAM_DOES_NOT_EXIST' ||
+  subCode === 'EVENT_STREAM_ALREADY_EXISTS' ||
+  subCode === 'NOT_AUTHORIZED' ||
+  code === 'UNAUTHENTICATED' ||
+  code === 'FORBIDDEN';
```
thomtrp, line 34: don't think UNAUTHENTICATED and FORBIDDEN alone are actually recoverable. I think subcode is the right thing to check

thomtrp (commented):
I am not convinced by the strategy:
- UNAUTHENTICATED and FORBIDDEN should not be considered as recoverable
- recoverable should decide if state should be reset or not. Not only if we should send the error to sentry

Thank you for contributing @corazzione but I think someone from the core team should take over this work. SSE is a tricky part. Sorry about that!

corazzione:
Thanks for the thorough review @thomtrp, really appreciate the feedback.

You're right on both points — I can see now that collapsing `UNAUTHENTICATED`/`FORBIDDEN` at the top-level `code` into `isRecoverable` is too broad. The intent was to catch the token expiry path (which does surface as `UNAUTHENTICATED` in the `next` message), but I agree that `subCode` is the more precise and safer signal to check, since it reflects the server's explicit intent rather than a generic auth failure.

Regarding the lost `store.set(shouldDestroyEventStreamState.atom, true)` on the `EVENT_STREAM_ALREADY_EXISTS` branch — you're absolutely right, that was an unintentional regression from the refactor. That call needs to stay.

And the broader point about `recoverable` controlling state reset (not just Sentry suppression) makes sense architecturally. I was too narrowly focused on the Sentry noise from the original issue.

I understand if the core team wants to take this over given the SSE complexity. If it's helpful, I'm happy to keep iterating based on your direction — just let me know.

packages/twenty-front/src/modules/sse-db-event/hooks/useTriggerEventStreamCreation.ts
thomtrp: actually it was already present line 161

PR #19481, opened 2026-04-09
Fix install and deploy commands

martmull:
- disable publish with existing version
- disable installation of app version already installed

### DIFF
--- a/packages/twenty-docs/developers/extend/apps/publishing.mdx
+++ b/packages/twenty-docs/developers/extend/apps/publishing.mdx
@@ -197,5 +197,10 @@ yarn twenty install
 ```
 
 <Note>
-Install is idempotent: if the version of the app currently registered on the server is already installed in your workspace, the command exits successfully as a no-op — file writes, manifest sync, and SDK regeneration are all skipped. To install a newer version, deploy or publish it first, then re-run `yarn twenty install`.
+The server enforces semver versioning on install, mirroring the rules on deploy:
+
+- Installing the same version that is already installed in your workspace is rejected with an `APP_ALREADY_INSTALLED` error.
+- Installing a lower version than the one currently installed is rejected with a `CANNOT_DOWNGRADE_APPLICATION` error.
+
+To install a newer version, deploy or publish it first, then re-run `yarn twenty install`.
 </Note>
--- a/packages/twenty-docs/developers/extend/apps/publishing.mdx
+++ b/packages/twenty-docs/developers/extend/apps/publishing.mdx
@@ -66,11 +66,17 @@ The share link uses the server's base URL (without any workspace subdomain) so i
 
 ### Version management
 
+The server enforces semver versioning on every deploy. Each new deploy must use a `version` in `package.json` that is **strictly higher** (per [semver](https://semver.org) ordering) than the version currently deployed for this app. Re-deploying the same version, or pushing a lower one, is rejected before the tarball is stored — you'll see a `VERSION_NOT_BUMPED` error from the CLI.
+
 To release an update:
 
-1. Bump the `version` field in your `package.json`
+1. Bump the `version` field in your `package.json` (e.g. `1.2.3` → `1.2.4`, `1.3.0`, or `2.0.0`)
 2. Run `yarn twenty deploy` (or `yarn twenty deploy --remote production`)
 3. Workspaces that have the app installed will see the upgrade available in their settings
+
+<Note>
+Pre-release tags work as expected: bumping `1.0.0-rc.1` → `1.0.0-rc.2` is allowed, and a final release like `1.0.0` is correctly recognized as higher than `1.0.0-rc.5`. The version in `package.json` must itself be a valid semver string — values like `v1.0.0` (with a leading `v`) are rejected as `INVALID_INPUT`.
+</Note>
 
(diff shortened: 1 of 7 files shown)


packages/twenty-docs/developers/extend/apps/publishing.mdx
```diff
@@ -189,3 +195,7 @@ You can also install apps from the command line:
 ```bash filename="Terminal"
 yarn twenty install
 ```
+
+<Note>
+Install is idempotent: if the version of the app currently registered on the server is already installed in your workspace, the command exits successfully as a no-op — file writes, manifest sync, and SDK regeneration are all skipped. To install a newer version, deploy or publish it first, then re-run `yarn twenty install`.
```
charlesBochet: note: I would prefer an error message: App is already installed (not mandatory)

packages/twenty-server/src/engine/core-modules/application/application-install/application-install.service.ts
```diff
@@ -117,6 +117,21 @@ export class ApplicationInstallService {
...
 
+      const incomingVersion = resolvedPackage.packageJson.version;
+
+      if (
+        !wasCreated &&
+        isDefined(application.version) &&
+        isDefined(incomingVersion) &&
+        application.version === incomingVersion
+      ) {
+        this.logger.log(
```
charlesBochet: I would prefer a an error here

Also I think we should block lower versions too

martmull: indeed, forgot install

packages/twenty-server/src/engine/core-modules/application/application-install/application-install.service.ts
```diff
@@ -117,6 +118,37 @@ export class ApplicationInstallService {
...
+      ) {
+        if (!isDefined(semver.valid(incomingVersion))) {
+          throw new ApplicationException(
+            `Invalid version "${incomingVersion}" in package.json. Must be a valid semver version.`,
+            ApplicationExceptionCode.INVALID_INPUT,
+          );
+        }
+
+        if (isDefined(semver.valid(application.version))) {
+          if (semver.eq(incomingVersion, application.version)) {
```
prastoin, line 136: ***Polish:*** That's nitpicking but that might cool to introduce a method equivalent to `packages/twenty-server/src/engine/core-modules/application/application-install/application-install.service.ts` but that compare all 3 semver

packages/twenty-server/src/engine/core-modules/application/application-registration/application-registration.exception.ts
```diff
@@ -32,6 +33,8 @@ const getExceptionUserFriendlyMessage = (
       return msg`The app source channel does not match the expected type.`;
     case ApplicationRegistrationExceptionCode.VARIABLE_NOT_FOUND:
       return msg`Application registration variable not found.`;
+    case ApplicationRegistrationExceptionCode.VERSION_NOT_BUMPED:
```
prastoin: ***Nitpick:*** `VERSION_ALREADY_EXISTS` seems more standard

packages/twenty-server/src/engine/core-modules/application/application-registration/application-tarball.service.ts
```diff
@@ -103,6 +104,33 @@ export class ApplicationTarballService {
...
+
+          if (!isDefined(semver.valid(incomingVersion))) {
+            throw new ApplicationRegistrationException(
+              `Invalid version "${incomingVersion}" in package.json. Must be a valid semver version.`,
+              ApplicationRegistrationExceptionCode.INVALID_INPUT,
+            );
+          }
+
+          if (
+            isDefined(semver.valid(currentVersion)) &&
```
prastoin, line 125: ***Remark:*** Seems redundant ?

martmull: no, check for semver version valid, then check for new version is higher. Not the same error messages

packages/twenty-server/src/engine/core-modules/application/application-install/application-install.service.ts
martmull: like a refacto?

prastoin: no a brand new one but that wrapps the semver quite verbose API such as the one I've shared

charlesBochet (approved):

packages/twenty-server/src/engine/core-modules/application/application-registration/application-tarball.service.ts
```diff
@@ -103,6 +104,33 @@ export class ApplicationTarballService {
...
+            semver.lte(incomingVersion, currentVersion)
+          ) {
+            throw new ApplicationRegistrationException(
+              `Cannot deploy ${universalIdentifier}@${incomingVersion}: version must be higher than the currently deployed version ${currentVersion}. Please bump the version in package.json.`,
+              ApplicationRegistrationExceptionCode.VERSION_ALREADY_EXISTS,
+            );
```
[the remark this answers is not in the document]
prastoin: cc @martmull

PR #19502, opened 2026-04-09
Fix system objects not appearing in sidebar View picker due to filtering mismatch

abdulrahmancodes:
## Summary
- System objects with valid views were incorrectly hidden in the View > System objects picker
- The system picker page used a different filter (`view.key !== ViewKey.INDEX`) than the parent page (`isViewDisplayableInNavigationMenu`), causing a mismatch where the "System objects" link was visible but clicking it showed an empty list
- Aligned filtering in `SidePanelNewSidebarItemViewSystemPickerSubPage` to use `isViewDisplayableInNavigationMenu` and exclude views already in the workspace, consistent with the non-system picker page


etiennejouan (approved):
LGTM !

PR #19505, opened 2026-04-09
Fix: Select next sidebar menu item after removing current item

abdulrahmancodes:


packages/twenty-front/src/modules/navigation-menu-item/edit/side-panel/hooks/useNavigationMenuItemEditOrganizeActions.ts
```diff
@@ -98,8 +110,33 @@ export const useNavigationMenuItemEditOrganizeActions =
...
+      if (isDefined(nextItem)) {
+        setSelectedNavigationMenuItemIdInEditMode(nextItem.id);
+        const link = getNavigationMenuItemComputedLink(
+          nextItem,
+          objectMetadataItems,
+          views,
+        );
+        if (isNonEmptyString(link)) {
+          navigate(link);
+        }
```
etiennejouan, line 132: I'm not sure this logic is needed. wdyt ?

etiennejouan (commented):
Works great ! 🙌

etiennejouan (approved):

packages/twenty-front/src/modules/navigation-menu-item/edit/side-panel/hooks/useNavigationMenuItemEditOrganizeActions.ts
abdulrahmancodes: It’s needed because when we select a nav item for editing, it also navigates to that page. Similarly, when we autoselect the next item, it should navigate as well

PR #19539, opened 2026-04-10
halftone generator v1 + new 3d shapes effect start

Bonapara:
## Summary
- refresh the website halftone studio with new rendering controls, UI updates, and export/runtime plumbing
- improve halftone output quality by opening highlights, grouping shadows, and reducing visible row artifacts
- update home/problem illustration assets, 3D model references, and related illustration components to use the new visuals
- adjust footer and layout-related website wiring to support the updated illustration experience

## Testing
- Not run (not requested)


mabdullahabaid (approved):
Resolved conflicts and inlined utilities. Code requires a lot of cleanup, but that's the part for when we're done with styles imo.

Build size is 30Mb, package size is 22.6Mb, so we're good. Let's get closer to shipping V1, then I suggest we spend a sprint making things more maintainable honestly and clean dead-files/dead-code 😅

PR #19540, opened 2026-04-10
  [AI] xAI native search tools + agent capabilities refactor

ehconitin:
closes https://discord.com/channels/1130383047699738754/1491127326321152030

## Summary
- **xAI Responses API**: switch from `providerOptions.searchParameters` to explicit `webSearch()` / `xSearch()` tool instances, aligning xAI with how Anthropic / OpenAI / Bedrock expose native tools
- **Capabilities refactor**: rename `NativeModelCapabilities` → `AgentCapabilities`, add `codeInterpreter`, extract `isAgentCapabilityEnabled` + `AGENT_CAPABILITY_DEFAULTS` to `twenty-shared/ai` so frontend and backend share one source of truth for default-enabled state
- **UI**: replace custom checkbox menu item with standard `MenuItemToggle`
- **Billing**: count both `web_search` and `x_search` as native live-search calls in step-cost aggregation
- **Lazy tool runtime**: exclude `directTools` from `learn_tools` lookup so native / MCP tools don't return spurious `notFound` errors
- **Action tool provider**: web search uses the shared opt-in helper instead of inline `!== false`, fixing default-on for agents without explicit `modelConfiguration`

before -

CleanShot 2026-04-10 at 14 25 10

after -

CleanShot 2026-04-10 at 14 24 44

------------------------------

CleanShot 2026-04-16 at 19 03 27@2x

CleanShot 2026-04-16 at 19 04 33@2x

CleanShot 2026-04-16 at 19 04 15@2x


FelixMalfait (changes requested):
## Review Summary

This PR bundles **three distinct concerns** into one: (1) a UI refactor (checkbox → toggle), (2) a rename (`NativeModelCapabilities` → `AgentCapabilities`) + adding `codeInterpreter`, and (3) xAI Responses API support with schema workarounds. The UI change and the rename are clean. The xAI integration has real architectural issues.

---

### The Good

- **UI refactor** (checkbox → `MenuItemToggle`): Clean, removes custom styled components in favor of a standard component. Good change.
- **Rename** `NativeModelCapabilities` → `AgentCapabilities` and `nativeCapabilities` → `capabilities`: Makes sense — these aren't "native model" capabilities anymore since they now include external capabilities like code interpreter and EXA web search.
- **Adding `codeInterpreter` to the capabilities system**: Straightforward extension.
- **xAI native tools** (`webSearch()`, `xSearch()`): Moving from `providerOptions.searchParameters` to explicit tool instances is the right direction — it aligns xAI with how Anthropic/OpenAI/Bedrock handle native tools.
- **`countNativeWebSearchCallsFromSteps`** now tracking `x_search` too — correct.
- **Tests for `wrapJsonSchemaForExecution` and `sanitizeToolSetForXaiResponses`**: Good to have.

---

### The Problems

#### 1. `sanitizeToolSetForXaiResponses` — patching symptoms, not root cause

This is the core issue. The function strips `additionalProperties: false` from every tool's JSON schema because the xAI Responses API chokes on it. But this is a **lossy transformation** applied to the entire tool set at the call site level (`prepareToolSetForModel`), which means:

- You're silently degrading schema strictness for *all* tools when running on xAI.
- If xAI fixes this bug tomorrow, you have to remember to remove this workaround from a util that's disconnected from the actual API call.
- The approach of "mutate the schema to fit the provider" doesn't scale — what happens when the next provider has a different schema quirk?

**Better approach:** This should be handled at the **SDK provider layer** (`buildXaiProvider`), not at the tool-set level. The `@ai-sdk/xai` provider's `responses()` call likely accepts middleware or schema transformation options. If not, the sanitization should at minimum be co-located with the provider construction in `sdk-provider-factory.service.ts` so the workaround is visible where the provider is built, not scattered across the tool pipeline.

#### 2. `wrapJsonSchemaForExecution` — defensive guards hiding a type problem

The original code:
```ts
const properties = (schema.properties as Record<string, unknown>) ?? {};
const required = (schema.required as string[]) ?? [];
```

The new code:
```ts
const properties =
  isObject(schema.properties) && !isArray(schema.properties)
    ? (schema.properties as Record<string, unknown>)
    : {};
const required = isArray(schema.required)
  ? schema.required.filter((item): item is string => typeof item === 'string')
  : [];
```

The function signature says `schema: Record<string, unknown>`. If callers are passing schemas where `properties` might be an array or `required` contains non-strings, that's a **caller problem** — those aren't valid JSON Schemas. Adding runtime guards here means you're accepting malformed input and silently swallowing it, which makes debugging harder.

The `...schema` spread and `typeof schema.type === 'string' ? schema.type : 'object'` changes are the actually useful fix here — they preserve `$defs` and other schema metadata that the old code was dropping. But the `isObject`/`isArray`/`typeof item === 'string'` guards are unnecessary defensive programming. If a schema arrives with `properties` as an array, something is very wrong upstream and you *want* that to fail loudly, not silently produce an empty properties object.

**Better approach:** Keep the `...schema` spread and the `Set` deduplication (those fix real bugs). Drop the type guards. If there's a specific caller passing bad schemas, fix that caller.

#### 3. `isCapabilityEnabled` — inconsistent default behavior

```ts
const isCapabilityEnabled = (capability: AgentCapabilityKey) => {
  if (capability === 'codeInterpreter') {
    return modelConfiguration.codeInterpreter?.enabled !== false; // default ON
  }
  return modelConfiguration[capability]?.enabled || false; // default OFF
};
```

Code interpreter defaults to **enabled** (`!== false`), while web search and twitter search default to **disabled** (`|| false`). This same `!== false` pattern appears in `action-tool.provider.ts`:

```ts
const isCodeInterpreterEnabledForAgent =
  context.agent?.modelConfiguration?.codeInterpreter?.enabled !== false;
const isWebSearchEnabledForAgent =
  context.agent?.modelConfiguration?.webSearch?.enabled !== false;
```

But wait — in the backend, *both* code interpreter and web search use `!== false` (default on), while in the frontend `isCapabilityEnabled`, only code interpreter uses that pattern. This is a **frontend/backend inconsistency** that will cause the toggle UI to show "off" while the backend treats it as "on" for web search on newly created agents.

**Better approach:** Pick one convention and apply it uniformly. If new capabilities should default to enabled, use `!== false` everywhere. If they should default to disabled, use `=== true` everywhere. Document the convention.

#### 4. `sanitizeXaiJsonSchemaValue` has a dead code branch

```ts
const sanitizeXaiJsonSchemaValue = (value: unknown): unknown => {
  if (isArray(value)) {
    return value.map(sanitizeXaiJsonSchemaValue);
  }

  if (!isObject(value) || value === null || isArray(value)) {
    //                      ^^^^^^^^^^^^^^   ^^^^^^^^^^^^^
    //                      already covered   already returned above
    return value;
  }
  ...
```

`isObject` from `@sniptt/guards` already returns `false` for `null`. And the `isArray` check is redundant since arrays already returned on line 5. This is the kind of "just to be safe" code that signals the author wasn't confident in what the guards do.

#### 5. Minor: duplicated log lines

Both `NativeModelToolProvider` and `ChatExecutionService` now log `Web search strategy: native (provider SDK) | external (EXA)` with identical format strings. This should be a single log at the decision point, not duplicated across consumers.

---

### Suggested Improvements

1. **Move xAI schema sanitization into the provider layer** — either in `buildXaiProvider` or as a thin wrapper around the model returned by `provider.responses()`. Keep provider quirks co-located with provider construction.
2. **Drop the type guards in `wrapJsonSchemaForExecution`** — keep the `...schema` spread and `Set` dedup (real fixes), remove the `isObject`/`isArray`/`typeof` guards (defensive noise). If schemas are malformed, fix the source.
3. **Unify the default-enabled convention** — `isCapabilityEnabled` in the frontend should match the backend's `!== false` pattern for all capabilities, or vice versa. Currently web search has different defaults on frontend vs backend.
4. **Clean up dead branches** in `sanitizeXaiJsonSchemaValue`.
5. **Consider splitting this PR** — the UI refactor + rename is independent from the xAI Responses API work. Reviewing and reverting are easier with smaller, focused PRs.

---

packages/twenty-shared/src/ai/constants/agent-capability-defaults.const.ts
```diff
@@ -0,0 +1,11 @@
+import { type AgentCapability } from '../types/agent-capability.type';
+
+export const AGENT_CAPABILITY_DEFAULTS = {
+  webSearch: true,
+  twitterSearch: false,
+  // Default-off: unlike webSearch (free via model-native capability), codeInterpreter
```
FelixMalfait: Websearch isn't "free" on native model, it's charged to (like 0.1$) per request
Most models providers do offer a native model but the reason we don't enable it natively is that it's more inpredictible in how they work (how do you upload file to them would differ for each for example).

We do have the "local" driver though which is free. But it's not recommended for production because it's insecure and can lead to a data breach / server hack

packages/twenty-server/src/engine/metadata-modules/ai/ai-chat/services/chat-execution.service.ts
```diff
@@ -115,9 +114,12 @@ export class ChatExecutionService {
         workspace.id,
       );
 
-    const toolContext = {
+    // Regular AI chat is not executing a saved agent, so agent-specific
+    // capability toggles should only apply in AgentAsyncExecutorService.
```
FelixMalfait: It feels like this comment describes the PR changes and not really a good fit to stay here long-term

Also in the chat the capabilities should be inherited from the model so for example if you pick Grok then you should get tweet search to be available. Or console should be available too

packages/twenty-server/src/engine/metadata-modules/ai/ai-billing/utils/count-native-web-search-calls-from-steps.util.ts
```diff
@@ -1,15 +1,21 @@
 import { type StepResult, type ToolSet } from 'ai';
 
-const WEB_SEARCH_TOOL_NAME = 'web_search';
+// Shared by billing and workflow execution logging because both treat these
+// as provider-native search tool calls.
+export const NATIVE_SEARCH_TOOL_NAMES = new Set(['web_search', 'x_search']);
```
FelixMalfait: Two exports in one file

FelixMalfait: I'm not convinced this makes sense. For me native search = web search ; X search is a different kind of tool, completely separate, and only available on Grok

packages/twenty-server/src/engine/metadata-modules/ai/ai-agent-execution/services/agent-async-executor.service.ts
```diff
@@ -258,6 +339,32 @@ export class AgentAsyncExecutorService {
...
+        : '';
+      const errorMessage =
+        error instanceof Error ? error.message : String(error);
+
+      this.logger.error(
+        `Workflow agent execution failed [workspace=${agent?.workspaceId} agent=${agent?.id} model=${resolvedModelForLog?.modelId}${statusSuffix}]: ${errorMessage}`,
+        error instanceof Error ? error.stack : undefined,
+      );
+
+      this.exceptionHandlerService.captureExceptions(
```
FelixMalfait: I don't feel like it's a good idea to start capturing exceptions like that. If the error needs to be catch by Sentry usually it's just re-thrown and then there's an interceptor or something that captures the exception?

packages/twenty-server/src/engine/metadata-modules/ai/ai-agent-execution/services/agent-async-executor.service.ts
```diff
@@ -258,6 +339,32 @@ export class AgentAsyncExecutorService {
...
+        : APICallError.isInstance(cause)
+          ? cause
+          : undefined;
+      const statusSuffix = apiCallError?.statusCode
+        ? ` status=${apiCallError.statusCode}`
+        : '';
+      const errorMessage =
+        error instanceof Error ? error.message : String(error);
+
+      this.logger.error(
```
FelixMalfait: Do we need these logs?

packages/twenty-server/src/engine/metadata-modules/ai/ai-agent-execution/services/agent-async-executor.service.ts
```diff
@@ -258,6 +339,32 @@ export class AgentAsyncExecutorService {
...
+      const cause =
+        error instanceof Error
+          ? (error as Error & { cause?: unknown }).cause
+          : undefined;
+      const apiCallError = APICallError.isInstance(error)
+        ? error
+        : APICallError.isInstance(cause)
+          ? cause
+          : undefined;
+      const statusSuffix = apiCallError?.statusCode
```
FelixMalfait: Do we need this?

packages/twenty-server/src/engine/metadata-modules/ai/ai-agent-execution/services/agent-async-executor.service.ts
```diff
@@ -99,6 +123,49 @@ export class AgentAsyncExecutorService {
...
+
+    const directToolsSection =
+      directToolNames.length > 0
+        ? `Direct native model tools available now: ${directToolNames.map((toolName) => `\`${toolName}\``).join(', ')}.`
+        : 'No direct native model tools are available.';
+
+    const sections = [
+      `## Available Workflow Tools
+
+${directToolsSection}
```
FelixMalfait: I don't really understand what you're doing here. 
For agent chat, all tools are available so we have 100+ tools and can't pass them in the tool array so we resort to a hack where we pass the list of tools in the system prompt and then the chat has to call learn_tool/execute_tool. 
For agents in workflow it's very different because it's a targeted action and a very limited number of tools, that's why we didn't have this learn tool / execute tool system and don't need to pollute with a huge system prompt. We can be more surgical, it's very different from a chat context which has access to the whole universe

packages/twenty-server/src/engine/core-modules/tool-provider/services/lazy-tool-runtime.service.ts
```diff
@@ -0,0 +1,85 @@
...
+export type LazyToolRuntime = {
+  toolCatalog: ToolIndexEntry[];
+  lazyToolCatalog: ToolIndexEntry[];
+  directTools: ToolSet;
+  directToolNames: string[];
+  runtimeTools: ToolSet;
+};
+
+@Injectable()
+export class LazyToolRuntimeService {
```
FelixMalfait: Not sure I understand why we need to introduce one more service like that?

packages/twenty-server/src/engine/core-modules/tool-provider/providers/native-model-tool.provider.ts
```diff
@@ -32,16 +34,20 @@ export class NativeModelToolProvider implements NativeToolProvider {
...
     }
 
-    if (!this.webSearchService.shouldUseNativeSearch()) {
-      return {};
-    }
+    const useProviderNativeWebSearch =
+      this.webSearchService.shouldUseNativeSearch();
+
+    this.logger.log(
+      `Web search strategy: ${useProviderNativeWebSearch ? 'native (provider SDK)' : 'external (EXA)'}`,
```
FelixMalfait: Do we need to add those logs?

packages/twenty-server/src/engine/metadata-modules/ai/ai-chat/services/chat-execution.service.ts
```diff
@@ -170,10 +159,30 @@ export class ChatExecutionService {
...
-        : { tools: {}, callableToolNames: [] };
+    const nativeSearchTools =
+      this.aiModelConfigService.getChatNativeSearchTools(registeredModel, {
+        useProviderNativeWebSearch: useNativeSearch,
+      });
+    const hasNativeWebSearch = Object.prototype.hasOwnProperty.call(
+      nativeSearchTools,
+      SEARCH_TOOL_NAMES.webSearch,
+    );
+    const hasNativeXSearch = Object.prototype.hasOwnProperty.call(
```
FelixMalfait: feels like a code smell? (maybe not)

PR #19542, opened 2026-04-10
Upgrade cli tool version in technical apps

martmull:
as title


prastoin (approved):
Why not latest everywhere ?

PR #19544, opened 2026-04-10
Skip email/calendar tab creation for custom object record page layouts

Weiko:
## Context
Emails and calendars can only be associated with specific objects (companies, people...) and we were adding those tabs for all custom objects. I'm removing these from the default config


martmull (approved):

Devessier (approved):
Okay for me. Matches `packages/twenty-front/src/modules/page-layout/constants/DefaultRecordPageLayout.ts`

PR #19556, opened 2026-04-10
fix: side panel close animation cleanup not firing due to invalid CSS transition

abdulrahmancodes:
## Summary

Fixes the side panel close animation cleanup (`sidePanelCloseAnimationCompleteCleanup`) never running during the close transition.

This eventually fixes the issue where in navbar edit mode, clicking on a nav item sometimes doesn't get selected, the side panel opens but shows `Select a navigation item to edit` instead of the selected item's settings. The root cause was stale `isSidePanelClosing` state from a previous close that wasn't cleaned up, causing the next open to run cleanup synchronously (without emitting the close event), which interfered with the navigation item selection flow.

### Root cause

The CSS `transition` on `StyledSidePanelWrapper` used `${themeCssVariables.animation.duration.normal}s`, which Linaria converts to a CSS custom property value like `width var(--t-animation-duration-normal)s`. After CSS variable substitution, `0.3` and `s` become two separate tokens, not a valid `<time>` dimension (`0.3s`). The browser rejects the entire transition declaration, falls back to `all 0s`, and the width change happens instantly with no `transitionend` event.

### Fix

- Use `calc(var(--t-animation-duration-normal) * 1s)` to properly construct the `<time>` value (consistent with ~30 other usages in the codebase).
- Filter `handleTransitionEnd` to only process `width` transitions on the wrapper element itself, ignoring bubbled child `background-color` transitions.

## Before

https://github.com/user-attachments/assets/496ccbff-dc96-487d-a994-451dbf2a5165

## After

https://github.com/user-attachments/assets/78255240-1d7d-42cd-9b56-25627613883a

### DIFF
--- a/packages/twenty-front/src/modules/side-panel/components/SidePanelForDesktop.tsx
+++ b/packages/twenty-front/src/modules/side-panel/components/SidePanelForDesktop.tsx
@@ -1,19 +1,19 @@
+import { tableWidthResizeIsActiveState } from '@/object-record/record-table/states/tableWidthResizeIsActivedState';
 import { SidePanelRouter } from '@/side-panel/components/SidePanelRouter';
 import { SidePanelWidthEffect } from '@/side-panel/components/SidePanelWidthEffect';
 import { SIDE_PANEL_CLICK_OUTSIDE_ID } from '@/side-panel/constants/SidePanelClickOutsideId';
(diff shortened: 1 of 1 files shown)


etiennejouan (approved):
🚀 👏

PR #19560, opened 2026-04-10
[Billing for self host] End dummy enterprise key validity 

ijreilly:
Screenshot 2026-04-10 at 16 40 07

Re-using hasValidEnterpriseKey to avoid breaking changes. 
This will be entirely removed in the next versions.


FelixMalfait:
Great! Looks like it breaks tests 
      Received: [{"extensions": {"code": "INTERNAL_SERVER_ERROR", "exceptionEventId": "mocked-exception-id", "userFriendlyMessage": "Row level permission predicate feature is disabled."}, "message": "Row level permission predicate feature is disabled"}]

FelixMalfait (approved):

PR #19561, opened 2026-04-10
Clean event creation exception

thomtrp:
https://twenty-v7.sentry.io/issues/7351816489/?environment=prod&project=4507072499810304&query=is%3Aunresolved&referrer=issue-stream

Those are expected error that should not reach sentry. These happen when stream TTL expires or user session ends


packages/twenty-front/src/modules/sse-db-event/components/SSEQuerySubscribeEffect.tsx
```diff
@@ -87,17 +88,12 @@ export const SSEQuerySubscribeEffect = () => {
       }
     } catch (error) {
       if (CombinedGraphQLErrors.is(error)) {
-        const subCode = error.errors[0]?.extensions?.subCode;
-        const code = error.errors[0]?.extensions?.code;
+        const subCode = error.errors[0]?.extensions?.subCode as
```
charlesBochet: use typeguard

packages/twenty-front/src/modules/sse-db-event/hooks/useTriggerEventStreamCreation.ts
```diff
@@ -80,9 +81,21 @@ export const useTriggerEventStreamCreation = () => {
           }>,
         ) => {
           if (isDefined(value?.errors) && Array.isArray(value.errors)) {
-            captureException(
-              new Error(`SSE subscription error: ${value.errors[0]?.message}`),
-            );
+            const subCode = value.errors[0]?.extensions?.subCode as
```
charlesBochet: same

charlesBochet (approved):

PR #19562, opened 2026-04-10
Fields widget draft view

Weiko:
- Defer backend view creation for FIELDS widgets to prevent orphan views when users cancel or leave before saving. Views are now created optimistically in Jotai state (client-side UUID) and only persisted to the database when the page layout is saved.
- Align the frontend's default field groups fallback with the backend logic: split into General/Other groups, hide relation fields by default, and filter invisible fields consistently across all code paths.


packages/twenty-front/src/modules/page-layout/hooks/useCreatePendingFieldsWidgetViews.ts
```diff
@@ -0,0 +1,94 @@
...
+            input: {
+              id: viewId,
+              name: viewName,
+              icon: 'IconList',
+              objectMetadataId,
+              type: ViewType.FIELDS_WIDGET,
+            },
+          },
+          objectMetadataId,
+        );
```
Devessier, line 81: We do one call for each created Fields widget. As far as I remember, there is no mutation to create several views at once. Do we want to create one to improve the robustness of the system, as the goal of this PR is to prevent orphan views?

Devessier (approved):
Small bug reported in PM. Otherwise, looks good to me!

packages/twenty-front/src/modules/page-layout/hooks/useCreatePendingFieldsWidgetViews.ts
Weiko: The vision might be to merge PageLayoutUpsert with view creation so we can handle it at that time. I think it's fine to keep it that way I don't expect users to create many FIELDS widget in a single layout edition

PR #19564, opened 2026-04-10
Add gql operationName metadata in sentry

etiennejouan:


packages/twenty-server/src/engine/api/graphql/direct-execution/hooks/use-direct-execution.hook.ts
```diff
@@ -67,6 +73,17 @@ export function useDirectExecution(
         return;
       }
 
+      if (Sentry.isInitialized()) {
```
charlesBochet, line 76: @etiennejouan isn't senty an Exception driver? shouldn't this be abstracted? Not sure what's the state of exception driver abstraction in the code base

Weiko: useSentryTracing hook is not abstract for example and is directly imported into graphql config logic so we have precedents but I agree we could for example introduce a setTags methods in ExceptionHandlerService instead (see useGraphQLErrorHandlerHook for example how it's used @etiennejouan)

charlesBochet: yes, I think it's not really abstracted and kind of fake, this can be a re-work for later

charlesBochet (approved):

packages/twenty-server/src/engine/api/graphql/direct-execution/hooks/use-direct-execution.hook.ts
etiennejouan: Sentry is used for exception reporting but also for tracing. Here we're using the tracing. (same as here packages/twenty-server/src/engine/api/graphql/graphql-config/graphql-config.service.ts#L78-L79)

We could add a tracing driver abstraction, to make it cleaner.

PR #19567, opened 2026-04-10
Fix view filter/sort deletion

### LINKED ISSUE #19543: Saving view doesn't work as expected
**Scenario 1:**
1. Log in
2. Go to Companies
3. Create new view
4. Apply filter in new view and save

**Actual:** Button must be clicked twice in order to save the filter

**Expected:** Filter should be saved after clicking the button once

**Scenario 2:**
1. Log in
2. Go to Companies
3. Create new view
4. Apply filter in new view and save
5. Remove applied filter and save

**Actual:** "Update view" button stays even though request to backend is sent and view isn't updated  

**Expected:** View is updated and button disappears

https://github.com/twentyhq/private-issues/issues/461

etiennejouan:
fixes https://github.com/twentyhq/twenty/issues/19543

+ bonus bug : when deleting an advanced filter, it triggers a destroy which cascade-deletes associated view filters. Then, view filters deletion throws.


packages/twenty-front/src/modules/views/hooks/useSaveRecordFilterGroupsToViewFilterGroups.ts
```diff
@@ -69,15 +71,31 @@ export const useSaveRecordFilterGroupsToViewFilterGroups = () => {
...
+        ...prev,
+        current: (prev.current as FlatViewFilter[]).filter(
+          (viewFilter) =>
+            !isDefined(viewFilter.viewFilterGroupId) ||
+            !destroyedIdsSet.has(viewFilter.viewFilterGroupId),
+        ),
+      }));
+    }
+
+    return { viewFilterGroupIdsToDestroy };
```
charlesBochet: I'm not a fan of this? Is this for optimistic rendering? why do we need to return them?

packages/twenty-front/src/modules/views/hooks/useSaveCurrentViewFiltersAndSorts.ts
```diff
@@ -16,8 +16,14 @@ export const useSaveCurrentViewFiltersAndSorts = () => {
...
   const saveCurrentViewFilterAndSorts = async () => {
     await saveRecordSortsToViewSorts();
-    await saveRecordFilterGroupsToViewFilterGroups();
-    await saveRecordFiltersToViewFilters();
+
+    const { viewFilterGroupIdsToDestroy } =
+      await saveRecordFilterGroupsToViewFilterGroups();
+
+    await saveRecordFiltersToViewFilters({
+      destroyedViewFilterGroupIds: viewFilterGroupIdsToDestroy,
```
charlesBochet: I'm confused, this should be backend driven therefore no need to do it on frontend except for optimistic (but we already get SSE event, if we want a fast opitmisitc but let's make the naming super clear + I feel we should a global strategy, this looks like a patch to me

charlesBochet (approved):

PR #19568, opened 2026-04-10
Implement full tab widget frontend

Devessier:
https://github.com/user-attachments/assets/488e0145-ff49-4205-8fd3-0eb4f614c054


Weiko (approved):
Some duplicated code as mentioned already but this can be improved later.
I see that you are renaming canvas, what are the impacts on existing page layouts?
The rest LGTM @Devessier

PR #19576, opened 2026-04-11
fix: allow overriding REACT_APP_SERVER_BASE_URL independently of SERVER_URL

salmonumbrella:
## Summary

- Allow `REACT_APP_SERVER_BASE_URL` to be set as an independent env var, overriding what gets injected into the frontend config
- Falls back to `SERVER_URL` when unset (preserving current behavior)

## Problem

When self-hosting Twenty behind a reverse proxy, VPN, or on platforms where the access hostname differs from the configured `SERVER_URL`, the frontend cannot reach the backend; `generateFrontConfig()` always injects `SERVER_URL` as `REACT_APP_SERVER_BASE_URL` into the frontend HTML. The browser then makes API calls to that hardcoded URL. If the user accesses Twenty via a different hostname the API calls fail with "Unable to Reach Back-end."

The frontend already has auto-detection logic that derives the backend URL from `window.location.origin` when `REACT_APP_SERVER_BASE_URL` is empty but there's currently no way to trigger it, since `SERVER_URL` is always injected and cannot be set to empty (validation requires a valid URL).

## Use Cases

This benefits any self-hosted deployment where the access hostname isn't known at config time:

- **VPN access** (Tailscale, WireGuard) — hostname differs from LAN
- **Reverse proxy** (Nginx, Caddy, Traefik) — custom domain fronting the app
- **Tunnel services** (Cloudflare Tunnel, ngrok) — dynamic hostnames
- **Home server platforms** (Umbrel, CasaOS, Unraid) — accessed via mDNS, IP, or VPN interchangeably

## Test Plan

- [x] Verify existing behavior: without REACT_APP_SERVER_BASE_URL set, frontend still uses SERVER_URL as before
- [x] Set REACT_APP_SERVER_BASE_URL="" frontend auto-detects backend URL from browser location
- [x] Set REACT_APP_SERVER_BASE_URL=http://custom-domain:3000 frontend uses the custom URL

### DIFF
--- a/packages/twenty-server/src/utils/generate-front-config.ts
+++ b/packages/twenty-server/src/utils/generate-front-config.ts
@@ -11,7 +11,8 @@ export function generateFrontConfig(): void {
   const configObject = {
     window: {
       _env_: {
-        REACT_APP_SERVER_BASE_URL: process.env.SERVER_URL,
+        REACT_APP_SERVER_BASE_URL:
+          process.env.REACT_APP_SERVER_BASE_URL ?? process.env.SERVER_URL,
       },
     },
   };


charlesBochet:
@salmonumbrella I don't fully understand the issue. Could you give a precise example of your setup and how requests / DNS flow?

salmonumbrella:
Sure! Here's a concrete example:

**Setup:** Twenty self-hosted on a home server (e.g. Umbrel) at LAN IP `192.168.1.50`. The user also accesses it remotely via Tailscale at `100.64.0.5`, or via a Cloudflare Tunnel at `https://twenty.example.com`.

**The problem:** `SERVER_URL` must be set to one specific URL (say `http://192.168.1.50:3000`). `generateFrontConfig()` then hardcodes that value into the HTML as `REACT_APP_SERVER_BASE_URL`. When the user opens Twenty from Tailscale (`100.64.0.5:3000`), the browser tries to call `http://192.168.1.50:3000` for API requests — which is unreachable from the Tailscale network. Result: "Unable to Reach Back-end."

**The fix:** One extra line — `process.env.REACT_APP_SERVER_BASE_URL ?? process.env.SERVER_URL`. If the self-hoster sets `REACT_APP_SERVER_BASE_URL=""`, the frontend's existing auto-detection kicks in and derives the API URL from `window.location.origin`. The browser then calls whichever hostname it used to load the page — LAN IP, Tailscale IP, or tunnel domain — and it just works. If `REACT_APP_SERVER_BASE_URL` is unset, behavior is identical to today (falls back to `SERVER_URL`).

charlesBochet:
why don't you setup your SERVER_URL to be your public IP (100.64.0.5:3000) or your public DNS?

salmonumbrella:
You're right that for most setups, a stable `SERVER_URL` is the correct answer. My specific case was Umbrel + Tailscale, where users legitimately access the same deployment via multiple hostnames (LAN mDNS, Tailscale, LAN IP) that a single `SERVER_URL` can't satisfy.

Rather than push a change here, I've moved the fix downstream to Umbrel's app packaging — they can neutralize the frontend config injection at the container level and let your existing `window.location.origin` fallback in `packages/twenty-front/src/config/index.ts` take over: getumbrel/umbrel-apps#5308. That keeps Twenty's code untouched and fixes the specific deployment pattern that broke.

Closing.

PR #19586, opened 2026-04-11
Add ClickHouse-backed metered credit cap enforcement

FelixMalfait:
## Summary
Implements a ClickHouse-backed polling system to enforce metered-credit caps for workflow executions, replacing reliance on Stripe billing alerts. The system re-evaluates tier caps against live pricing on every poll cycle, allowing price/tier changes to propagate immediately without recreating Stripe alert objects.

## Key Changes

- **BillingUsageCapService**: New service that queries ClickHouse for current-period credit usage and evaluates whether a subscription has reached its metered-credit allowance (tier cap + credit balance)
  - `isClickHouseEnabled()`: Checks if ClickHouse is configured
  - `getCurrentPeriodCreditsUsed()`: Sums creditsUsedMicro from usageEvent table for a workspace within a billing period
  - `evaluateCap()`: Determines if usage has reached the allowance by reading live pricing from the subscription

- **EnforceUsageCapJob**: Cron job that polls all active subscriptions and updates `hasReachedCurrentPeriodCap` on metered items
  - Runs every 2 minutes to keep cap enforcement in sync with live usage
  - Supports shadow mode (log-only) via `BILLING_USAGE_CAP_CLICKHOUSE_ENABLED` flag for safe rollout
  - Continues processing after per-subscription errors with detailed logging

- **EnforceUsageCapCronCommand**: CLI command to register the enforcement cron job

- **MeteredCreditService**: Extracted `extractMeteredPricingInfoFromSubscription()` as a pure function for callers that already hold the subscription with pricing loaded, avoiding redundant DB queries

- **Configuration**: Added `BILLING_USAGE_CAP_CLICKHOUSE_ENABLED` flag to control enforcement mode (active vs. shadow)

- **Constants**: Added `METERED_OPERATION_TYPES` to define which operation types count toward the metered product's credit cap

## Implementation Details

- The service queries ClickHouse for the sum of `creditsUsedMicro` in the current billing period, matching Stripe meter semantics
- Pricing is re-read on every evaluation, so tier changes propagate within one poll cycle without Stripe alert recreation
- The cron job only updates the database when the cap state actually changes (no-op if already in the correct state)
- Shadow mode allows safe validation before enabling enforcement; transitions are logged but not persisted
- Comprehensive test coverage for both the service and cron job, including error handling and state transitions

https://claude.ai/code/session_01VksTSrYLXJVCPVBQhQdBTe


packages/twenty-server/src/engine/core-modules/billing/constants/metered-operation-types.constant.ts
```diff
@@ -0,0 +1,12 @@
+/* @license Enterprise */
+
+import { UsageOperationType } from 'src/engine/core-modules/usage/enums/usage-operation-type.enum';
+
+// Operation types that count toward the WORKFLOW_NODE_EXECUTION metered product's
+// credit cap. The underlying Stripe meter is historically named WORKFLOW_NODE_RUN
+// but its unit is credits (creditsUsedMicro) sent via
+// StripeBillingMeterEventService.sendBillingMeterEvent. Keep this list in sync
+// with whatever UsageOperationType values are emitted under the metered product.
+export const METERED_OPERATION_TYPES: readonly UsageOperationType[] = [
```
FelixMalfait: I'm not sure I understand why we need to introduce this. Everything that generates a billingEvent should be taken account into the period's cap. In the future we will introduce additional limits like per-user cap (when you reply to this comment analyze what it would take to introduce per-user cap just out of curiosity).

packages/twenty-server/src/engine/core-modules/billing/crons/enforce-usage-cap.job.ts
```diff
@@ -0,0 +1,139 @@
...
+    });
+
+    let evaluated = 0;
+    let transitioned = 0;
+    let errors = 0;
+
+    for (const subscription of subscriptions) {
+      try {
+        const evaluation =
+          await this.billingUsageCapService.evaluateCap(subscription);
```
FelixMalfait: We have thousands of subscriptions. Is there a more efficient way to do this rather than looping over all subscriptions? We could at least do batches of 100 for example? Also take into account that most won't have been active at all since the last period. Let's be smart

packages/twenty-server/src/engine/core-modules/billing/crons/enforce-usage-cap.job.ts
```diff
@@ -0,0 +1,139 @@
...
+              `for subscription=${subscription.id} workspace=${subscription.workspaceId} ` +
+              `usage=${evaluation.usage} allowance=${evaluation.allowance} ` +
+              `tierCap=${evaluation.tierCap} creditBalance=${evaluation.creditBalance}`,
+          );
+          continue;
+        }
+
+        await this.billingSubscriptionItemRepository.update(
+          { id: meteredItem.id },
+          { hasReachedCurrentPeriodCap: shouldBeCapped },
```
FelixMalfait: when is this being reset? Do we have a cron or Stripe webhook that resets this at the beginning of each period I guess?

packages/twenty-server/src/engine/core-modules/billing/services/billing-usage-cap.service.ts
```diff
@@ -0,0 +1,126 @@
...
+    private readonly twentyConfigService: TwentyConfigService,
+  ) {}
+
+  isClickHouseEnabled(): boolean {
+    return Boolean(this.twentyConfigService.get('CLICKHOUSE_URL'));
+  }
+
+  // Returns the sum of creditsUsedMicro for a workspace in [periodStart, periodEnd).
+  // Matches Stripe meter semantics: StripeBillingMeterEventService sends
+  // creditsUsedMicro as the meter value, so summing the same field in ClickHouse
```
FelixMalfait: Can we please check micro vs kilo ; I thought it was a x1000 factor between the number of displayed credits vs Stripe credits but it might not be the case, maybe there's a x1M factor here and it's something different? Let's bring clarity into who works with what unit (Stripe, Clickhouse, Code, Display, etc.)

packages/twenty-server/src/engine/core-modules/twenty-config/config-variables.ts
```diff
@@ -815,6 +815,20 @@ export class ConfigVariables {
   @ValidateIf((env) => env.IS_BILLING_ENABLED === true)
   BILLING_STRIPE_WEBHOOK_SECRET: string;
 
+  // When true, workflow-credit cap enforcement is driven by a ClickHouse-backed
```
FelixMalfait: No need for long commentaries like that, that's what variable descriptions are for.

packages/twenty-server/src/engine/core-modules/billing/constants/metered-operation-types.constant.ts
FelixMalfait: Good call — removed the `METERED_OPERATION_TYPES` constant entirely. The ClickHouse query now sums all `creditsUsedMicro` for the workspace in the period, with no `operationType` filter. Every billing event counts toward the cap.

Regarding **per-user caps**: the `usageEvent` table in ClickHouse already has a `userWorkspaceId` column, so the ClickHouse query side is straightforward (add a `WHERE userWorkspaceId = ...` clause). On the enforcement side, `hasReachedCurrentPeriodCap` currently lives on `BillingSubscriptionItemEntity` (subscription-level). For per-user caps we'd need either a new entity (e.g. `BillingUserUsageCapEntity` keyed by `workspaceId + userId`) or a JSONB map on the subscription item. The cron would need to iterate users within each subscription. The hot-path check in `canWorkspaceUse` would also need to accept a userId and check the per-user cap alongside the subscription-level one. The ClickHouse and Stripe plumbing is already there; the main work would be the new entity/storage and the per-user evaluation loop.

---

packages/twenty-server/src/engine/core-modules/billing/crons/enforce-usage-cap.job.ts
FelixMalfait: Fixed — two improvements:

1. **Batched processing**: subscriptions are now loaded in batches of 100 using `take`/`skip` pagination with `createQueryBuilder`.
2. **Skip inactive subscriptions**: the query now uses `innerJoinAndSelect` on `billingSubscriptionItems` + `billingProduct` with a `WHERE product.metadata->>'productKey' = 'WORKFLOW_NODE_EXECUTION'` filter. This means only subscriptions that actually have a metered item are loaded — the majority without one are never fetched.

---

packages/twenty-server/src/engine/core-modules/billing/crons/enforce-usage-cap.job.ts
FelixMalfait: Yes — it's already reset in three places:

1. **`BillingWebhookInvoiceService`** — on `subscription_cycle` invoice webhook (new billing period start), resets `hasReachedCurrentPeriodCap = false` for all items in the subscription
2. **`BillingSubscriptionService`** — on trial period end
3. **`BillingSubscriptionUpdateService`** — on plan/price/interval changes (but not seats-only changes)

This cron then flips it `true` or back to `false` based on live ClickHouse usage within the current period. Added a comment in the cron pattern file documenting these reset points.

---

packages/twenty-server/src/engine/core-modules/billing/services/billing-usage-cap.service.ts
FelixMalfait: Investigated and added a unit reference comment at the top of the service. Here's the breakdown:

| Layer | Field / Unit | Value |
|-------|-------------|-------|
| **ClickHouse** | `creditsUsedMicro` (Int64) | 1 credit = $0.000001. Conversion: `DOLLAR_TO_CREDIT_MULTIPLIER = 1_000_000` credits per dollar |
| **Stripe meter** | `value` in `sendBillingMeterEvent` | Same unit — `usageEvent.creditsUsedMicro.toString()` is sent directly |
| **Stripe tier** | `tiers[0].up_to` | Same unit (credits) — this is the cap compared against the sum |
| **Display** | `toDollars()` | Converts credits → dollars: `credits / 1_000_000` |

So it's a **×1,000,000 factor** between dollars and credits (not ×1,000). The name "micro" is a bit misleading — each unit is a micro-dollar ($0.000001), but within the system they're just called "credits."

---

packages/twenty-server/src/engine/core-modules/twenty-config/config-variables.ts
FelixMalfait: Removed the comment block. The `description` in `@ConfigVariablesMetadata` covers it.

---

packages/twenty-server/src/engine/core-modules/billing/crons/enforce-usage-cap.job.ts
```diff
@@ -0,0 +1,157 @@
...
+      'BILLING_USAGE_CAP_CLICKHOUSE_ENABLED',
+    );
+
+    let evaluated = 0;
+    let transitioned = 0;
+    let errors = 0;
+    let offset = 0;
+
+    // Process subscriptions in batches, only loading those that have a
+    // metered (WORKFLOW_NODE_EXECUTION) item to skip the majority that don't.
```
etiennejouan: They all have it. We can drop the 
```
.andWhere("product.metadata->>'productKey' = :productKey", {
          productKey: BillingProductKey.WORKFLOW_NODE_EXECUTION,
        })
```
and all the table joins.

If we want to focus lighten the cron, we should filter out SUSPENDED workspaces (A trial ended sub without payment method will be past due + suspended)

packages/twenty-server/src/engine/core-modules/billing/crons/enforce-usage-cap.job.ts
```diff
@@ -0,0 +1,157 @@
...
+            this.logger.log(
+              `[shadow] would set hasReachedCurrentPeriodCap=${shouldBeCapped} ` +
+                `for subscription=${subscription.id} workspace=${subscription.workspaceId} ` +
+                `usage=${evaluation.usage} allowance=${evaluation.allowance} ` +
+                `tierCap=${evaluation.tierCap} creditBalance=${evaluation.creditBalance}`,
+            );
+            continue;
+          }
+
+          await this.billingSubscriptionItemRepository.update(
```
etiennejouan: nit: we could batch update

packages/twenty-server/src/engine/core-modules/billing/services/billing-usage-cap.service.ts
```diff
@@ -0,0 +1,115 @@
...
+      this.meteredCreditService.extractMeteredPricingInfoFromSubscription(
+        subscription,
+      );
+
+    if (!meteredPricingInfo) {
+      return { skipped: true, reason: 'no-metered-item' };
+    }
+
+    const [creditBalance, usage] = await Promise.all([
+      this.meteredCreditService.getCreditBalance(
```
etiennejouan, line 128: thought : with the 200 r/s Stripe limit, we will have issues when dealing with 10ks workspaces

etiennejouan (commented):
Left comments. Looks good to me.

I wonder how we could set up integration testing to ensure everything is working. The billing logic is becoming more and more complex. I'll dig a bit.
We should also remove the old alerting logic.

packages/twenty-server/src/engine/core-modules/billing/crons/enforce-usage-cap.job.ts
```diff
@@ -0,0 +1,157 @@
...
+        })
+        .orderBy('subscription.id', 'ASC')
+        .take(BATCH_SIZE)
+        .skip(offset)
+        .getMany();
+
+      for (const subscription of batch) {
+        try {
+          const evaluation =
+            await this.billingUsageCapService.evaluateCap(subscription);
```
FelixMalfait: This is within the look and is still going to spam clickhouse no? We need to step back again and be smarter for batching - it's not just protecting Postgres, it's also protecting Clickhouse. Consider every request in this PR, let's do something efficient.

packages/twenty-server/src/engine/core-modules/billing/services/billing-usage-cap.service.ts
FelixMalfait: I think this is wrong. I'm on one of our workspaces and I see "credits for this Period: 5k", It should be just 5 credits then. We need to find the root cause and fix it. Let's just do 1$=1credit from a user perspective ; and store everything (clickhouse stripe etc) in micro so 1$ = 1M microcredits

packages/twenty-server/src/database/commands/upgrade-version-command/1-22/1-22-instance-command-fast-1776078919203-add-credit-balance-to-billing-customer.ts
```diff
@@ -3,19 +3,19 @@ import { QueryRunner } from 'typeorm';
...
-@RegisteredInstanceCommand('1.23.0', 1785000000000)
-export class DropWorkspaceVersionColumnFastInstanceCommand
+@RegisteredInstanceCommand('1.22.0', 1776078919203)
+export class AddCreditBalanceToBillingCustomerFastInstanceCommand
   implements FastInstanceCommand
 {
   public async up(queryRunner: QueryRunner): Promise<void> {
     await queryRunner.query(
-      `ALTER TABLE "core"."workspace" DROP COLUMN IF EXISTS "version"`,
+      `ALTER TABLE "core"."billingCustomer" ADD COLUMN IF NOT EXISTS "creditBalanceMicro" integer NOT NULL DEFAULT 0`,
```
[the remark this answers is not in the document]
FelixMalfait: Good point

packages/twenty-server/src/engine/core-modules/billing/services/billing-usage-cap.service.ts
```diff
@@ -0,0 +1,202 @@
...
+
+      result.set(row.workspaceId, Number.isFinite(total) ? total : 0);
+    }
+
+    return result;
+  }
+
+  // Evaluates cap for a single subscription. Makes async calls to ClickHouse + Stripe.
+  // Used by non-cron callers. The cron uses evaluateCapBatch instead.
+  async evaluateCap(
```
etiennejouan, line 111: not used anymore

packages/twenty-server/src/engine/core-modules/billing/crons/enforce-usage-cap.job.ts
```diff
@@ -0,0 +1,234 @@
...
+    do {
+      batch = await this.billingSubscriptionRepository
+        .createQueryBuilder('subscription')
+        .innerJoinAndSelect(
+          'subscription.billingSubscriptionItems',
+          'item',
+          'item.billingSubscriptionId = subscription.id',
+        )
+        .innerJoinAndSelect('item.billingProduct', 'product')
+        .leftJoinAndSelect('product.billingPrices', 'price')
```
etiennejouan: products and prices are the same for all subs. It should be less costly to fetch them once.

etiennejouan (approved):

PR #19590, opened 2026-04-11
fix(api-key): batch role resolution with DataLoader to fix N+1

oniani1:
## Summary

The `role` @ResolveField on `ApiKeyResolver` calls `getRolesByApiKeys` with a single-element array per API key. When a query returns N API keys, this produces N separate DB queries to resolve their roles.

This adds an `apiKeyRoleLoader` to the existing DataLoader infrastructure. All API key IDs in a single GraphQL request are collected and resolved in one batched query.

- Before: N queries (one per API key)
- After: 1 query (batched via DataLoader)

## Changes

- `dataloader.service.ts` - new `createApiKeyRoleLoader` method, delegates to `ApiKeyRoleService.getRolesByApiKeys`
- `dataloader.interface.ts` - `apiKeyRoleLoader` added to `IDataloaders`
- `dataloader.module.ts` - import `ApiKeyModule` so `ApiKeyRoleService` is available
- `api-key.resolver.ts` - `role()` now uses `context.loaders.apiKeyRoleLoader.load()` instead of calling the service directly

## Test plan

- [ ] Verify `apiKeys { id role { label } }` query returns the same results as before
- [ ] Confirm only 1 role_target query fires regardless of how many API keys are returned

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/api-key/api-key.resolver.ts
+++ b/packages/twenty-server/src/engine/core-modules/api-key/api-key.resolver.ts
@@ -1,12 +1,5 @@
 import { UseGuards } from '@nestjs/common';
-import {
-  Args,
-  Context,
-  Mutation,
-  Parent,
-  Query,
-  ResolveField,
-} from '@nestjs/graphql';
+import { Args, Mutation, Parent, Query, ResolveField } from '@nestjs/graphql';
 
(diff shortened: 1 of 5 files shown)


charlesBochet (approved):
@oniani1 We don't need data loader in this case as we have cache already in place, refactored your PR in this way

PR #19593, opened 2026-04-11
test: fix failing useColorScheme test and remove FIXME

srijita2506:
Summary
This PR fixes a bug in the `useColorScheme` test suite and removes a lingering `FIXME` comment where the color scheme was unexpectedly unsetting during state updates.

Root Cause
Previously, the Jotai state was being initialized *inside* the `renderHook` callback using `useSetAtomState`. When the `setColorScheme` function was called, it triggered a hook re-render, which caused the callback to execute again and overwrite the new state with the hardcoded `'System'` initial state.

The Fix
- Removed the state initialization from inside the render cycle.
- Bootstrapped the state on a fresh store using `resetJotaiStore()` and `store.set()` *before* rendering the hook.
- Updated the mock `workspaceMember` to correctly use the `CurrentWorkspaceMember` type.
- Removed the `FIXME` comment and successfully asserted that the color scheme updates to `'Dark'`.

Testing
Ran tests locally to confirm the fix works as expected:
`corepack yarn jest --config packages/twenty-front/jest.config.mjs --testPathPattern=useColorScheme.test.tsx`


charlesBochet (approved):
Reviewed both changes, thank you!

PR #19594, opened 2026-04-12
fix: avoid reporting recoverable SSE token expiry to Sentry

reverb256:
## Problem

When a user session stays idle long enough for the token to expire, the SSE event stream creation path reports the resulting `UNAUTHENTICATED` and `FORBIDDEN` errors to Sentry. These are expected, recoverable reconnect cases — not bugs. ([Issue #18077](https://github.com/twentyhq/twenty/issues/18077))

## Solution

In the `message` handler of `useTriggerEventStreamCreation`, the switch on `subCode` only handled `EVENT_STREAM_ALREADY_EXISTS`. Added `UNAUTHENTICATED` and `FORBIDDEN` as additional recoverable cases, so they destroy the stream for clean recreation instead of being reported to Sentry.

This follows the same pattern already used in `SSEQuerySubscribeEffect.tsx` (lines 82-93), where these same subCodes are treated as recoverable in the subscription update path.

## Changes

- 1 file changed: `packages/twenty-front/src/modules/sse-db-event/hooks/useTriggerEventStreamCreation.ts`
- Added `UNAUTHENTICATED` and `FORBIDDEN` to the existing `switch(subCode)` recoverable cases
- The `store.set(shouldDestroyEventStreamState.atom, true)` on line 148 still runs unconditionally after the switch, so stream cleanup is preserved for all error paths

Fixes #18077

### DIFF
--- a/packages/twenty-front/src/modules/sse-db-event/hooks/__tests__/useTriggerEventStreamCreation.test.tsx
+++ b/packages/twenty-front/src/modules/sse-db-event/hooks/__tests__/useTriggerEventStreamCreation.test.tsx
@@ -1,0 +1,153 @@
+import { act, renderHook } from '@testing-library/react';
+
+import { useTriggerEventStreamCreation } from '@/sse-db-event/hooks/useTriggerEventStreamCreation';
+import { shouldDestroyEventStreamState } from '@/sse-db-event/states/shouldDestroyEventStreamState';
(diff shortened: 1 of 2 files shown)


charlesBochet:
Thanks for the contribution @reverb256! Closing this as the fix is already included in #19561 which takes a slightly different approach. Appreciate you digging into this!

PR #19596, opened 2026-04-12
fix: edit button not coming up on avatar right after image upload

### LINKED ISSUE #19595: The edit icon does not appear immediately after uploading an avatar for a new person
## Bug Description

- In the People table, if we select a person without an avatar and then upload a avatar image for it, we don't see the option to edit it immediately, even after we get the notification saying that the image upload is complete.
- The edit option becomes visible only after we refresh the page or first edit another field in the drawer before hovering on the avatar field again. Then it works properly.

https://github.com/user-attachments/assets/ac9318e9-5471-434c-8af3-5c20d0112460

## Expected behavior

The hover interaction and edit option should be immediately available once the image upload is complete

amlannandy:
## Summary

After uploading an image/file to the empty avatar field in the person tab, the edit icon next to the field would not appear until the browser was refreshed or another field was clicked.

### Root cause

- When user clicks over the avatar field, `recordFieldListCellEditModePosition` is set to `globalIndex`
- That is fine when a avatar already exists. But when there is no avatar already set, the native file picker is opened with no `onClose` handler attached.
- So after the file upload is completed, `recordFieldListCellEditModePosition` is never reset to null.
- `FieldsWidgetCellEditModePortal` stays anchored to the avatar file element
- When the user hovers over the same field again, its hover portal tries to compete to anchor for the same element
- So, `RecordInlineCellDisplayMode ` (the edit button) doesn't render

### Fix

- Pass the `onClose` function through `openFieldInput` to `openFilesFieldInput`
- `onClose` resets `recordFieldListCellEditModePosition` back to null, when the upload completes.

## Before

https://github.com/user-attachments/assets/ac9318e9-5471-434c-8af3-5c20d0112460

## After

https://github.com/user-attachments/assets/0d064a7f-95ad-4b92-a9ee-d9570f360972

Fixes #19595

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-board/record-board-card/anchored-portal/components/RecordBoardCardCellHoveredPortalContent.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-board/record-board-card/anchored-portal/components/RecordBoardCardCellHoveredPortalContent.tsx
@@ -48,6 +48,7 @@ export const RecordBoardCardCellHoveredPortalContent = () => {
         fieldDefinition,
         recordId,
         prefix: RECORD_BOARD_CARD_INPUT_ID_PREFIX,
+        onFileUploadClose: () => setRecordBoardCardEditModePosition(null),
       });
     }
   };
--- a/packages/twenty-front/src/modules/object-record/record-calendar/record-calendar-card/anchored-portal/components/RecordCalendarCardCellHoveredPortalContent.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-calendar/record-calendar-card/anchored-portal/components/RecordCalendarCardCellHoveredPortalContent.tsx
@@ -48,6 +48,7 @@ export const RecordCalendarCardCellHoveredPortalContent = () => {
         fieldDefinition,
         recordId,
         prefix: RECORD_CALENDAR_CARD_INPUT_ID_PREFIX,
+        onFileUploadClose: () => setRecordCalendarCardEditModePosition(null),
       });
     }
   };
--- a/packages/twenty-front/src/modules/object-record/record-field-list/anchored-portal/components/RecordFieldListCellHoveredPortalContent.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-field-list/anchored-portal/components/RecordFieldListCellHoveredPortalContent.tsx
@@ -52,7 +52,7 @@ export const RecordFieldListCellHoveredPortalContent = () => {
         fieldDefinition,
         recordId,
         prefix: instanceId,
-        onClose: () => setRecordFieldListCellEditModePosition(null),
+        onFileUploadClose: () => setRecordFieldListCellEditModePosition(null),
       });
     }
   };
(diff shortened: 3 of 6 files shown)


charlesBochet (commented):
Thanks for the fix! The result is correct — the edit button now properly reappears after upload. However, I'd like to suggest a cleaner approach that also fixes the same bug in other contexts.

## Current issues

1. **`onClose` is a generic name for a Files-field-only concern.** It's drilled through `openFieldInput` (the generic API) but only ever consumed by `openFilesFieldInput`. No other field type uses it.

2. **`useOpenFilesFieldInput` hardcodes table-specific cleanup via `isTableContext`.** The hook imports `RECORD_TABLE_CELL_INPUT_ID_PREFIX`, `RecordTableComponentInstanceContext`, and `recordTableCellEditModePositionComponentState` — it shouldn't know about tables.

3. **Board and Calendar cards have the same bug.** They also set an edit mode position on click but don't pass `onClose`, so the position is never cleaned up after file upload:
   - `RecordBoardCardCellHoveredPortalContent.tsx` — sets `recordBoardCardEditModePosition`
   - `RecordCalendarCardCellHoveredPortalContent.tsx` — sets `recordCalendarCardEditModePosition`

## Suggested approach

1. **Rename `onClose` → `onFileUploadClose`** in both `openFieldInput` and `openFilesFieldInput` to make the intent explicit.

2. **Remove the `isTableContext` hardcoding** from `useOpenFilesFieldInput`. Instead of branching on `prefix === RECORD_TABLE_CELL_INPUT_ID_PREFIX`, just call `onFileUploadClose?.()` in all cases. This removes the record-table imports and makes the hook fully context-agnostic.

3. **Have each caller pass its own `onFileUploadClose` callback:**
   - **Table** (`useOpenRecordTableCell.ts`): reset `recordTableCellEditModePositionComponentState`, call `goBackToPreviousDropdownFocusId()`, call `removeLastFocusItemFromFocusStackByComponentType()`
   - **Field list** (`RecordFieldListCellHoveredPortalContent.tsx`): reset `recordFieldListCellEditModePosition` (what you have now)
   - **Board card** (`RecordBoardCardCellHoveredPortalContent.tsx`): reset `recordBoardCardEditModePosition`
   - **Calendar card** (`RecordCalendarCardCellHoveredPortalContent.tsx`): reset `recordCalendarCardEditModePosition`

This way `useOpenFilesFieldInput` becomes context-agnostic, the bug is fixed for all contexts (not just field list), and the generic `openFieldInput` API clearly communicates that this parameter is Files-specific.

charlesBochet:
@amlannandy Thanks for opening a PR, I've refactored a bit, could you make sure it's working as expected on
- Table
- Field list 
- Board card
- Calendar card

amlannandy:
Hi @charlesBochet 
Thanks for the review and refactor! I've verified it at these 4 places. Attaching the recordings for you.

https://github.com/user-attachments/assets/e069fd7b-c6a7-40bc-bc9e-f44d0d7555bf


https://github.com/user-attachments/assets/9b700a84-9eda-403b-adbc-c2d5cfcdf157


https://github.com/user-attachments/assets/42222815-7942-49db-8e6b-7fc5f9393c07


https://github.com/user-attachments/assets/4cd97725-de67-429b-b597-c90edf23daeb

charlesBochet:
@amlannandy Thank you for taking the time. This is not exactly what I meant, you can also upload a file by clicking:
- on a TableCell (Field type FILE)
- from a card of board
- from a card of calendar

charlesBochet:
Ok, I've tested LGTM

charlesBochet (approved):

PR #19597, opened 2026-04-12
Refactor impersonation: token restoration instead of session destruction

FelixMalfait:
## Summary

- Replace `useImpersonationAuth` with `useImpersonationSession` hook that stores the admin's original tokens in `sessionStorage` before swapping to the impersonated user's session
- "Stop Impersonating" now restores the original tokens, resets Apollo cache, reloads the user, and navigates back — no re-login needed
- Fix stale Apollo cache bug by calling `client.clearStore()` on both impersonation start and stop
- Cross-workspace impersonation (new tab via redirect) falls back to `window.close()` then `signOut()`

## Architecture

| Flow | Mechanism | Stop Impersonating |
|---|---|---|
| Same-workspace (Members page) | In-place token swap + sessionStorage backup | Restore tokens → navigate back |
| Same-workspace (Admin panel) | In-place token swap + sessionStorage backup | Restore tokens → navigate back |
| Cross-workspace (Admin panel) | New tab via redirect (different domain) | Close tab / sign out |

## Test plan

- [ ] Impersonate a user from the Members page → verify session switches, banner shows
- [ ] Click "Stop Impersonating" → verify original admin session restores without re-login
- [ ] Verify navigation returns to the page where impersonation was initiated
- [ ] Impersonate a same-workspace user from admin panel → same restore behavior
- [ ] Impersonate a cross-workspace user from admin panel → opens new tab
- [ ] In cross-workspace tab, click "Stop Impersonating" → tab closes or signs out
- [ ] Refresh page while impersonating → "Stop Impersonating" still works (sessionStorage persists)
- [ ] Regular sign-out while impersonating → cleans up stored session properly

https://claude.ai/code/session_01SdHcodsLnF7R6Zg8fUtKj8


packages/twenty-front/src/pages/settings/admin-panel/SettingsAdminWorkspaceChatThread.tsx
```diff
@@ -160,7 +159,8 @@ export const SettingsAdminWorkspaceChatThread = () => {
                     .join('\n');
 
                   const toolParts = message.parts.filter(
-                    (part) => part.type === 'tool-call' && part.toolName,
+                    (part) =>
```
FelixMalfait, line 162: (Probably shouldn't be in this PR)

FelixMalfait:
## Security Review — `useImpersonationSession` hook

### Overview

This PR replaces `useImpersonationAuth` with `useImpersonationSession`, storing the admin's original `AuthTokenPair` in `sessionStorage` so "Stop Impersonating" restores the admin session without requiring re-login. The approach is architecturally sound, but has a few areas worth hardening.

---

### Findings

#### 1. **[Medium] Admin token pair stored in `sessionStorage` without integrity protection**

The admin's full `AuthTokenPair` (access token + refresh token) is serialized as plaintext JSON into `sessionStorage` during impersonation. While the existing token storage via `js-cookie` is also JS-accessible (not HttpOnly — it's set from client-side code), storing a second copy of the **admin's higher-privilege tokens** in `sessionStorage` creates a specific target during the impersonation window.

In an XSS scenario during impersonation, an attacker could:
- Read the admin's stored tokens from `sessionStorage`, gaining admin-level access
- Tamper with the stored tokens so that when the admin clicks "Stop Impersonating", a different token pair is loaded (session fixation)

**Recommendation (longer-term):** Consider server-side impersonation session management — store a session reference server-side and return a session ID, rather than keeping admin tokens client-side. This eliminates the client-side token exposure window entirely.

#### 2. **[Low] No structural validation on deserialized session data** ✅ Fixed

`stopImpersonating` previously cast `JSON.parse(raw)` directly to `StoredImpersonationSession` without validating the structure. A malformed or tampered `sessionStorage` entry would pass the `try/catch` (valid JSON) but could have missing/wrong-typed fields, leading to undefined behavior when setting `tokenPairState`.

**Fix applied:** Added `isValidStoredSession()` type guard that validates the expected shape (nested `tokenPair` with `accessOrWorkspaceAgnosticToken.token` and `refreshToken.token` as strings, plus `returnPath` as string). Invalid structures now trigger cleanup + sign-out.

#### 3. **[Low] `returnPath` not validated with existing `isValidReturnToPath`** ✅ Fixed

The codebase has an `isValidReturnToPath()` utility (blocks `//`-prefixed paths, onboarding routes, sign-in routes, etc.), but the stored `returnPath` was navigated to directly without validation. While `window.location.pathname` is safe at write-time and `react-router`'s `navigate()` is same-origin only, a tampered `sessionStorage` value could navigate the admin to an excluded internal route.

**Fix applied:** `session.returnPath` is now passed through `isValidReturnToPath()` before navigation, falling back to `'/'` if invalid.

#### 4. **[Info] Positive security properties**

- `signOut` → `clearSession()` calls `sessionStorage.clear()`, which wipes stored admin tokens on any sign-out (including if the impersonated user's session is terminated). This is correct.
- `client.clearStore()` is called on both impersonation start and stop, preventing Apollo cache cross-contamination between admin and impersonated user contexts.
- The `isImpersonating` flag is derived from the JWT payload server-side, not from client state — so it can't be spoofed.

---

### Fixes

Pushed to branch [`claude/security-review-pr-19597-AcR1G`](https://github.com/twentyhq/twenty/tree/claude/security-review-pr-19597-AcR1G) — see commit `09ef8449`:
- `isValidStoredSession()` type guard for structural validation
- `isValidReturnToPath()` check before navigating to stored return path

---

PR #19602, opened 2026-04-12
refactor(app-token): remove redundant ID and timestamp generation from before-create hook

srijita2506:
Description
This PR resolves a `FIXME` in the `BeforeCreateOneAppToken` hook by removing the manual generation of `id` and `updatedAt` for App Tokens, delegating this responsibility to the database.

Changes Made
* Removed manual `uuidv4()` and `new Date()` assignments from `before-create-one-app-token.hook.ts`.
* Removed the unused `uuid` import.

 Why this is safe to merge
The `FIXME` comment noted that these fields should be autogenerated by the database. Upon reviewing the migration history, the `AppToken` table was already created with default values at the database level (`DEFAULT uuid_generate_v4()` and `DEFAULT now()`) in `1700140427984-setupMetadataTables.ts`.

Because these defaults are already enforced on table creation, application-level generation is redundant. Removing it correctly enforces Separation of Concerns without requiring any new database migrations or instance commands for legacy databases.


srijita2506:
I noticed the CI failure in server-integration-test (9). Looking at the logs, it seems to be a workspace/metadata initialization error (relation "core.keyValuePair" does not exist) rather than an issue with the AppToken refactor. My changes only affect the manual ID/timestamp assignment in the hook.

PR #19605, opened 2026-04-12
Inline twenty-shared types in SDK declarations

charlesBochet:
## Summary
- `twenty-shared` is private and never published to npm, so SDK consumers couldn't resolve type imports like `from 'twenty-shared/types'` in the generated `.d.ts` files
- Replace `vite.config.sdk.ts` with `rollup-plugin-dts` which compiles `src/sdk/index.ts` directly into a self-contained `dist/sdk/index.d.ts` with all `twenty-shared` types inlined
- The JS output from `vite.config.sdk.ts` (`dist/sdk/*.js`) was unused — the main export already maps to `dist/index.mjs` from the node Vite config

## Changes
- **Deleted** `vite.config.sdk.ts` — its preserved-module JS output wasn't referenced by any `package.json` export
- **Added** `rollup.config.dts.mjs` — uses `rollup-plugin-dts` to compile SDK types from source with `twenty-shared` inlined (~850ms)
- **Updated** `project.json` — build/dev/build:sdk targets now use rollup instead of the removed vite config
- **Updated** `tsconfig.json` — removed `vite.config.sdk.ts` from include

## Test plan
- [ ] Run `npx nx build twenty-sdk` and verify `dist/sdk/index.d.ts` contains no `twenty-shared` references
- [ ] Verify `dist/index.mjs` and `dist/index.cjs` are still produced correctly
- [ ] Verify CLI (`dist/cli.cjs`) still works
- [ ] Verify `npx nx build:sdk twenty-sdk` works standalone


mohammadmanzu1-creator (approved):

mohammadmanzu1-creator:
Ui

mohammadmanzu1-creator:
});
};

function saveUnsavedData() {
  // How you do this depends on your app.
}

function displayMessage() {
  // Show a non-modal message to the user.
}
Another way is to call the connection’s close() method. However, you need to make sure your app is aware of this, as subsequent attempts to access the database will fail.

db.onversionchange = function() {
  saveUnsavedData().then(function() {
    db.close();
    stopUsingTheDatabase();
  });
};

function stopUsingTheDatabase() {
  // Put the app into a state where it no longer uses the database.
}
The new client (the one attempting the upgrade) can use the blocked event to detect if other clients are preventing the upgrade from happening. The blocked event fires if other clients still hold a connection to the database after their versionchange events have fired.

const request = indexedDB.open("library", 4); // Request version 4.
let blockedTimeout;

request.onblocked = function() {
  // Give the other clients time to save data asynchronously.
  blockedTimeout = setTimeout(function() {
    displayMessage("Upgrade blocked - Please close other tabs displaying this site.");
  }, 1000);
};

request.onupgradeneeded = function(event) {
  clearTimeout(blockedTimeout);
  hideMessage();
  // ...
};

function hideMessage() {
  // Hide a previously displayed message.
}
The user will only see the above message if another client fails to disconnect from the database. Ideally the user will never see this.

2. Constructs
A name is a string equivalent to a DOMString; that is, an arbitrary sequence of 16-bit code units of any length, including the empty string. Names are always compared as opaque sequences of 16-bit code units.

NOTE: As a result, name comparison is sensitive to variations in case as well as other minor variations such as normalization form, the inclusion or omission of controls, and other variations in Unicode text. [Charmod-Norm]
If an implementation uses a storage mechanism which does not support arbitrary strings, the implementation can use an escaping mechanism or something similar to map the provided name to a string that it can store.

To create a sorted name list from a list names, run these steps:

Let sorted be names sorted in ascending order with the code unit less than algorithm.

Return a new DOMStringList associated with sorted.

PR #19613, opened 2026-04-12
fix: prevent image upload panel from being clipped in side panel

### LINKED ISSUE #19571: Minimized Notes Tab not resizing for components
## Bug Description

In my Macbook Pro 14", on Twenty v1.21.0, when I'm on the opportunities page, with an opportunity open, and when adding a minimized note, and trying to add an image, the upload and embed modal is "cutted off".

Image

## Expected behavior

To have the image modal, or embedded upload, resized to fit the minimized note tab.

## Technical inputs
N/A

sanskar0627:
fixes: #19571

## What's happening
BlockNote's `.bn-panel` (the Upload/Embed popup for images, videos, audio) has a hardcoded `width: 500px` in the upstream `blocknoteStyles.css`. When you open it inside the side panel (~350px wide), it overflows past the container's `overflow: hidden` and gets clipped.
## Why percentage-based fixes don't work here
I initially tried `max-width: 100%` on `.bn-panel` but it doesn't work  the panel sits inside a Floating UI popover that's absolutely positioned with no explicit width. Its width comes from its content (the 500px panel), so `max-width: 100%` just resolves to that same 500px. It's a circular reference.
## The fix
```css
& .bn-mantine {
  container-type: inline-size;
}
& .bn-mantine .bn-panel {
  width: min(500px, 100cqi);
  box-sizing: border-box;
}
```
container-type: inline-size on .bn-mantine lets us reference the editor's actual width using cqi units. min(500px, 100cqi) keeps the panel at 500px in wide containers (main note view) and shrinks it to fit in narrow ones (side panel).

I scoped container-type to .bn-mantine (BlockNote's own root wrapper) rather than the outer StyledEditor div, so the containment context stays within BlockNote's own tree.

#Before
image

#After
image
image

Tested
Side panel: panel fits correctly, matches "Add image" button width 
Full-width editor: panel stays at 500px 
Slash menu, formatting toolbar, drag handles: all unaffected


FelixMalfait (approved):

sanskar0627:
@FelixMalfait The merge queue failure is from a flaky E2E test (create-record.spec.ts "Previous Companies" field timeout) that's completely unrelated to my CSS only change. Could you please re-add this to the merge queue? Or let me know if there’s anything I need to do

Thank you

PR #19622, opened 2026-04-13
Fix command `FixMessageThreadViewAndLabelIdentifierCommand`

prastoin:
## Introduction
View standard backfill seed has been introduced during a partial patch, a window exists where some workspaces got created without the view and wasn't included in the upgrade process

The fix isn't covering all the view that has been missed during that time, only the one required for the command to pass


Weiko (approved):
LGTM

PR #19628, opened 2026-04-13
Fix design

martmull:
## Before
image

## After

image

### DIFF
--- a/packages/twenty-ui/src/input/button/components/Button/Button.tsx
+++ b/packages/twenty-ui/src/input/button/components/Button/Button.tsx
@@ -18,6 +18,7 @@ export type ButtonVariant = 'primary' | 'secondary' | 'tertiary';
 export type ButtonAccent = 'default' | 'blue' | 'danger';
 
 export type ButtonProps = {
+  id?: string;
   className?: string;
   Icon?: IconComponent;
   title?: string;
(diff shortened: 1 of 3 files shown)


packages/twenty-ui/src/input/button/components/Button/Button.tsx
```diff
@@ -441,7 +442,9 @@ const computeButtonWrapperColor = (
             : themeCssVariables.font.color.inverted;
         case 'danger':
           return !inverted
-            ? themeCssVariables.font.color.danger
+            ? !disabled
+              ? themeCssVariables.font.color.danger
+              : themeCssVariables.color.red5
```
prastoin, line 447: ***Remark:*** Global impact

packages/twenty-ui/src/input/button/components/Button/Button.tsx
```diff
@@ -18,6 +18,7 @@ export type ButtonVariant = 'primary' | 'secondary' | 'tertiary';
 export type ButtonAccent = 'default' | 'blue' | 'danger';
 
 export type ButtonProps = {
+  id?: string;
```
prastoin, line 21: ***Remark:*** Seems weird that's the first time we have to pass an id down there, is it the standard pattern to display tooltip ? or even displaying tooltip on a button ?

prastoin (approved):

packages/twenty-ui/src/input/button/components/Button/Button.tsx
martmull: yes, but should be global

packages/twenty-ui/src/input/button/components/Button/Button.tsx
martmull: yes first time

Weiko (approved):
LGTM

PR #19636, opened 2026-04-13
Fix permission flag deletion validator

martmull:
As title

### DIFF
--- a/packages/twenty-server/src/engine/workspace-manager/workspace-migration/workspace-migration-builder/validators/services/flat-permission-flag-validator.service.ts
+++ b/packages/twenty-server/src/engine/workspace-manager/workspace-migration/workspace-migration-builder/validators/services/flat-permission-flag-validator.service.ts
@@ -229,13 +229,7 @@ export class FlatPermissionFlagValidatorService {
         flatEntityMaps: flatRoleMaps,
       });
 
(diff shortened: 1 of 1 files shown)


prastoin (approved):

packages/twenty-server/src/engine/workspace-manager/workspace-migration/workspace-migration-builder/validators/services/flat-permission-flag-validator.service.ts
```diff
@@ -229,13 +229,7 @@ export class FlatPermissionFlagValidatorService {
         flatEntityMaps: flatRoleMaps,
       });
 
-      if (!isDefined(referencedRole)) {
```
Weiko, line 232: Why this change @martmull ?

martmull: i unify the check with other role children (objectPermission and fieldPermission)
this permissionFlag check was not working properly, and throw an unexpected error during app uninstall

prastoin (commented):
I don't have a strong opinion regarding enforcing the related entity existence before deleting one
Though indeed this would not occur if we were building/validating the permissions before the roles in the orchestrator

PR #19641, opened 2026-04-13
Document isAuthRequired: true instead of false

martmull:


Weiko (approved):

packages/twenty-docs/developers/extend/apps/building.mdx
```diff
@@ -555,7 +555,7 @@ export default defineLogicFunction({
   httpRouteTriggerSettings: {
     path: '/webhook',
     httpMethod: 'POST',
-    isAuthRequired: false,
```
charlesBochet, line 558: is it? In my understanding:
- isAuthRequired force you to provide a JWT, 
- here you want to have a custom authentication (based on wehbook secret) which is not part of the framework so you should have a "public" endpoint that implement its own check

martmull: indeed, went too fast

PR #19642, opened 2026-04-13
Move view field label identifier deletion validation into the cross entity validation

prastoin:
## Introduction
In the same validate build and run we should be able to delete a view field targetting a label identifier and at the same create one that repoints to it again without failing any validation

Leading for this valdiation rule to be moved in the cross entity validation steps


packages/twenty-server/src/engine/metadata-modules/flat-view-field/validators/utils/validate-view-field-label-identifier-cross-entity.util.ts
```diff
@@ -0,0 +1,130 @@
...
+  const alreadyCheckedViewUniversalIdentifiers = new Set<string>();
+
+  for (const deleteAction of deletedViewFieldActions) {
+    const deletedViewField = findFlatEntityByUniversalIdentifier({
+      universalIdentifier: deleteAction.universalIdentifier,
+      flatEntityMaps: preDeletionFlatViewFieldMaps,
+    });
+
+    if (!isDefined(deletedViewField)) {
+      continue;
```
charlesBochet: throw

charlesBochet (approved):

PR #19644, opened 2026-04-13
refactor(settings): centralize protected route redirection

srijita2506:
The Problem

SettingsProtectedRouteWrapper was using Navigate to trigger redirects during the render phase. This caused race conditions if multiple wrappers failed at the same time

The Solution

1)Added a new Jotai atom (settingsRedirectPathState) to track pending settings redirects safely.

image

2)Removed the Navigate component from SettingsProtectedRouteWrapper. It now uses a useEffect hook to write the intended destination to the new Jotai atom after rendering is complete.

image

3)Modified usePageChangeEffectNavigateLocation to prioritize the new settingsRedirectPath so the global effect handles the actual route change.

image

4)Updated the cleanup logic in PageChangeEffect to clear the settingsRedirectPath unconditionally after any navigation occurs. This prevents the bug where an interrupted redirect (like onboarding) would leave a stale value in the atom

image


thomtrp:
Hi @srijita2506, thank you for contributing.
We try to focus on existing issues rather than refactor. Feel free to select good first issues available 🙏

srijita2506:
@thomtrp Thankyou.

Will definitely work on good first issues.

PR #19649, opened 2026-04-13
Move is active to fe

Weiko:
## Context
Moving isActive filtering to the frontend for page layout tabs and widgets, hiding inactive entities from the UI while keeping them in state for future reactivation

Next we will implement deactivated standard tab re-activation during tab creation (cc @Devessier)
📋 Menu (Slots)


packages/twenty-front/src/modules/page-layout/widgets/fields/hooks/useDeleteViewForFieldsWidget.ts
```diff
@@ -1,14 +1,21 @@
...
 
 export const useDeleteViewForFieldsWidget = () => {
   const { performViewAPIDestroy } = usePerformViewAPIPersist();
+  const { getViewFromState } = useGetViewFromState();
 
   const deleteViewForFieldsWidget = useCallback(
     async (viewId: string) => {
+      if (!isDefined(getViewFromState(viewId))) {
+        return;
+      }
```
Devessier: Why do we need that? 😮

packages/twenty-front/src/modules/page-layout/widgets/record-table/hooks/useDeleteViewForRecordTableWidget.ts
```diff
@@ -1,14 +1,21 @@
...
 export const useDeleteViewForRecordTableWidget = () => {
   const { performViewAPIDestroy } = usePerformViewAPIPersist();
+  const { getViewFromState } = useGetViewFromState();
 
   const deleteViewForRecordTableWidget = useCallback(
     async (viewId: string) => {
+      if (!isDefined(getViewFromState(viewId))) {
+        return;
+      }
+
```
Devessier: Same here. I'm unsure why we need that now 😮

Devessier (approved):
Masterclass!

PR #19658, opened 2026-04-13
Fix 1.17 upgrade crash when built-function folder is missing

charlesBochet:
## Summary

- Fix the v1.17 upgrade command (`upgrade:1-17:migrate-workflow-code-steps`) crashing when the `built-function/` folder doesn't exist in storage
- When built files are missing, the migration now builds from source using esbuild instead of failing

## Context

Serverless functions created in v1.14 only write **source files** (`serverless-function/{id}/draft/src/index.ts`). The **built files** (`built-function/{id}/draft/index.mjs`) are only created when a function is actually **executed** at runtime.

The v1.17 migration assumed both always exist. Users who created serverless functions but never ran them hit one of two errors depending on their storage driver:
- **Local storage**: `ENOENT: no such file or directory, scandir '.local-storage/.../built-function/{id}/draft'`
- **S3**: `Missing index.mjs or index.ts for serverless function {id}/draft in workspace {workspace-id}`

## Fix

`readOldFunctionFiles()` now:
1. Wraps the `built-function/` read in a try/catch — missing directory is no longer fatal
2. If built files are missing, builds from source using esbuild (same config used by `LogicFunctionResourceService.buildInMemory`)
3. Only throws if the source (`index.ts`) is genuinely missing

## Test plan

- [ ] Create serverless functions in v1.14 without executing them
- [ ] Upgrade through v1.15 → v1.16 → v1.17
- [ ] Verify upgrade completes without error
- [ ] Verify migrated logic functions work correctly after upgrade

### DIFF
--- a/packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-workflow-code-steps.command.ts
+++ b/packages/twenty-server/src/database/commands/upgrade-version-command/1-17/1-17-migrate-workflow-code-steps.command.ts
@@ -2,8 +2,13 @@ import { Logger } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
 
 import crypto from 'crypto';
+import { mkdtemp, mkdir, writeFile, readFile, rm } from 'fs/promises';
+import { tmpdir } from 'os';
+import { join } from 'path';
 
+import { build } from 'esbuild';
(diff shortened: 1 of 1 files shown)


charlesBochet:
Closing, this is deployed on top of v1.17.3 as v1.17.4

PR #19659, opened 2026-04-13
Remove app record if first install fails

martmull:
If application install fails, and if app was created, uninstall app

fixes https://discord.com/channels/1130383047699738754/1491822937462804590

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/application/application-install/application-install.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/application/application-install/application-install.service.ts
@@ -18,10 +18,7 @@ import { ApplicationRegistrationEntity } from 'src/engine/core-modules/applicati
 import { ApplicationRegistrationSourceType } from 'src/engine/core-modules/application/application-registration/enums/application-registration-source-type.enum';
 import { ApplicationEntity } from 'src/engine/core-modules/application/application.entity';
 import { ApplicationService } from 'src/engine/core-modules/application/application.service';
-import {
-  ApplicationPackageFetcherService,
-  type ResolvedPackage,
-} from 'src/engine/core-modules/application/application-package/application-package-fetcher.service';
+import { ApplicationPackageFetcherService } from 'src/engine/core-modules/application/application-package/application-package-fetcher.service';
 import { ApplicationSyncService } from 'src/engine/core-modules/application/application-manifest/application-sync.service';
 import { CacheLockService } from 'src/engine/core-modules/cache-lock/cache-lock.service';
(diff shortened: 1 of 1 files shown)


packages/twenty-server/src/engine/core-modules/application/application-install/application-install.service.ts
```diff
@@ -234,6 +236,20 @@ export class ApplicationInstallService {
...
+        this.logger.log(
+          'Cleaning up application installation due to error during install',
+        );
+        try {
+          await this.applicationSyncService.uninstallApplication({
+            applicationUniversalIdentifier: universalIdentifier,
+            workspaceId: params.workspaceId,
+          });
+        } catch {
+          // ignore errors during uninstall
```
charlesBochet: we should throw here (and log to sentry), uninstall should always work

packages/twenty-server/src/engine/core-modules/application/application-install/application-install.service.ts
```diff
@@ -111,6 +111,10 @@ export class ApplicationInstallService {
   ): Promise<boolean> {
     let resolvedPackage: ResolvedPackage | null = null;
 
+    const universalIdentifier = appRegistration.universalIdentifier;
+
+    let isVersionUpgrade = true;
```
charlesBochet: avoid let, this is side-effect prone, we should only use const

packages/twenty-server/src/engine/core-modules/application/application-install/application-install.service.ts
```diff
@@ -111,6 +111,10 @@ export class ApplicationInstallService {
   ): Promise<boolean> {
     let resolvedPackage: ResolvedPackage | null = null;
 
+    const universalIdentifier = appRegistration.universalIdentifier;
```
charlesBochet: what's the benefit of this line?

packages/twenty-server/src/engine/core-modules/application/application-install/application-install.service.ts
```diff
@@ -234,6 +236,20 @@ export class ApplicationInstallService {
         `Failed to install app ${appRegistration.universalIdentifier}: ${error}`,
       );
 
+      if (!isVersionUpgrade) {
+        this.logger.log(
+          'Cleaning up application installation due to error during install',
```
charlesBochet: this log is not helpful unless we log workspaceId + appUniversalId

charlesBochet (changes requested):

charlesBochet (approved):

PR #19660, opened 2026-04-13
Remove orphaned workspaceId column from BillingSubscriptionItemEntity

charlesBochet:
## Summary
- Removes the `workspaceId` column, `@Index()`, and `@ManyToOne` workspace relation from `BillingSubscriptionItemEntity`
- The entity defined these fields but they don't exist in the actual database table, causing `column X.workspaceId does not exist` errors at runtime
- The workspace relationship is already accessible through the parent `BillingSubscriptionEntity`


Weiko (approved):
LGTM

PR #19662, opened 2026-04-13
Deprecate IS_RECORD_TABLE_WIDGET_ENABLED feature flag

FelixMalfait:
## Summary
- Removes the `IS_RECORD_TABLE_WIDGET_ENABLED` feature flag, making the record table widget unconditionally available in dashboard widget type selection
- The flag was already seeded as `true` for all new workspaces and only gated UI visibility in one component (`SidePanelPageLayoutDashboardWidgetTypeSelect`)
- Cleans up the flag from `FeatureFlagKey` enum, dev seeder, and test mocks

## Analysis
The flag only controlled whether the "View" (Record Table) widget option appeared in the dashboard widget type selector. The entire record table widget infrastructure (rendering, creation hooks, GraphQL types, `RECORD_TABLE` enum in `WidgetType`) is independent of the flag and fully implemented. No backend logic depends on this flag.


charlesBochet (approved):
I don't think it was activated on new workspace but let's get it out for the next release

PR #19297, opened 2026-04-03
Fix sidebar active state for object nav items on record show pages

abdulrahmancodes:
Screenshot 2026-04-03 at 11 31 36 AM

PR #19664, opened 2026-04-14
Fix active navigation item disambiguation

abdulrahmancodes:
## Summary

- Introduce a `activeNavigationMenuItemState` Jotai atom (persisted via localStorage) to disambiguate active navigation items when multiple items share the same URL
- Add active item evaluation for record show pages with three scenarios:
  1. Navigating from a nav item → parent stays active + dedicated RECORD item also active
  2. Clicking a dedicated RECORD nav item → only that item active
  3. Navigating via search/direct link → OBJECT nav item fallback, or Opened section if none exists
- Extract shared active logic into `isNavigationMenuItemActive` utility to eliminate duplication between orphan items and folder items
- Support multiple simultaneously active items within folders via `Set<number>` instead of a single index


packages/twenty-front/src/modules/navigation-menu-item/display/folder/hooks/useNavigationMenuItemFolderOpenState.ts
```diff
@@ -38,24 +41,50 @@ export const useNavigationMenuItemFolderOpenState = ({
...
       );
-      return isLocationMatchingNavigationMenuItem(
-        currentPath,
-        currentViewPath,
-        item.type,
-        computedLink,
-      );
-    },
-  );
+      if (!isDefined(objectMetadataItem)) {
```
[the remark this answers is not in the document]
abdulrahmancodes: link items point to external URLs

packages/twenty-front/src/modules/navigation-menu-item/common/states/activeNavigationMenuItemState.ts
```diff
@@ -0,0 +1,9 @@
+import { type ActiveNavigationMenuItem } from '@/navigation-menu-item/common/types/ActiveNavigationMenuItem';
+import { createAtomState } from '@/ui/utilities/state/jotai/utils/createAtomState';
+
+export const activeNavigationMenuItemState =
+  createAtomState<ActiveNavigationMenuItem | null>({
+    key: 'activeNavigationMenuItemState',
+    defaultValue: null,
+    useSessionStorage: true,
```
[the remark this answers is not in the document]
abdulrahmancodes: cross-session persistence was never intended

packages/twenty-front/src/modules/navigation-menu-item/common/types/ActiveNavigationMenuItem.ts
```diff
@@ -0,0 +1,5 @@
+export type ActiveNavigationMenuItem = {
+  id: string;
+  path: string;
+  objectMetadataId: string;
+};
```
abdulrahmancodes: all three fields are needed: `id` for disambiguation, `path` to scope the atom to the current page, and `objectMetadataId` to keep the parent active on show pages where the URL has changed from the index page

etiennejouan: Yes, but path and objectMetadata could be deducted from navigationMenuItem id

packages/twenty-front/src/modules/navigation-menu-item/common/utils/isNavigationMenuItemActive.ts
```diff
@@ -0,0 +1,79 @@
+import { AppPath, NavigationMenuItemType } from 'twenty-shared/types';
+import { getAppPath, isDefined } from 'twenty-shared/utils';
+import { type NavigationMenuItem } from '~/generated-metadata/graphql';
+
+import { type ActiveNavigationMenuItem } from '@/navigation-menu-item/common/types/ActiveNavigationMenuItem';
+import { isLocationMatchingNavigationMenuItem } from '@/navigation-menu-item/common/utils/isLocationMatchingNavigationMenuItem';
+import { type EnrichedObjectMetadataItem } from '@/object-metadata/types/EnrichedObjectMetadataItem';
+
+type IsNavigationMenuItemActiveParams = {
+  navigationMenuItem: NavigationMenuItem | null;
```
etiennejouan: Sorry but we should refactor this.

navigationMenuItem should not be nullable here. `isNavigationMenuItemActive` should not accept this.

Opened section should not be based on component with nav menu item functional logic.

Wdyt ? It could be an interesting next PR.

etiennejouan: We could in this PR, create a `IsNavigationMenuItemActiveParams` util with navigationMenuItem not null.
+ a is.....Active without navMenuItem, it should be explicitly named. After refactor we would move it in its own file.

packages/twenty-front/src/modules/navigation-menu-item/common/utils/isNavigationMenuItemActive.ts
```diff
@@ -0,0 +1,79 @@
...
+  currentPathWithSearch: string;
+  activeNavigationMenuItem: ActiveNavigationMenuItem | null;
+};
+
+export const isNavigationMenuItemActive = ({
+  navigationMenuItem,
+  computedLink,
+  objectMetadataItem,
+  currentPath,
+  currentPathWithSearch,
```
etiennejouan: why not passing currentLocation directly ? or creating a useIsNavigationMe nuItemActive hook ? calling useLocation directly ?

packages/twenty-front/src/modules/navigation-menu-item/common/utils/isNavigationMenuItemActive.ts
```diff
@@ -0,0 +1,79 @@
...
+  activeNavigationMenuItem,
```
etiennejouan: if hook, state can be called

etiennejouan (commented):
@abdulrahmancodes, sorry to question the whole implementation but I'm reviewing and still finding it too complicated. It's better than previous one, but still too complicated (it's not your logic fault, intended behaviour is complicated).

I've stopped review and wonder if the solution could not be : 
- `lastClickedNavigationMenuItemId` (same as your activeNavigationMenu) but naming less misleading ? I prefer If because path and objectMetadataId can be deducted from nav menu item id
- `useIdentifyActiveNavigationMenuItemIdsAndObjectMetadataIdForOpenedSection`, a hook taking all nav menu items (favorites and workspace) and finding the nav menu items (can be object + record) that should be active. First, checking if lastClicked is defined + path is relevant. If not, searching for object-typed nav menu item then record-typed.
If it has not found any nav menu item, objectMetadataIdForOpenedSection is defined with the right.
It would be computed once in nav menu item sidebar, stored in state and read in opened section and nav menu item component. It would centralized in one and only one palce the logic around which item is active, opened vs nav menu items.

wdyt ?

etiennejouan (commented):
LGTM ! Done some changes, prefer waiting for your review before approving.

We could improve with an activeNavMenuItemIds state at the whole nav menu scope. To do later maybe

etiennejouan:
Good catch from @abdulrahmancodes : 

> when navigating from an OBJECT/VIEW nav item, the dedicated RECORD nav item for that record doesn't get activated

The logic implemented in this PR (last commit) is to activate RECORD and OBJECT when landing directly but activating only lastClickNavMenuItem when coming from navigation menu.

It may be against bonapara's [comment](https://github.com/twentyhq/twenty/pull/19297#issuecomment-4199600545).


We decided to merge and will come back to this if considered as a bug (it would be my mistake)

etiennejouan (approved):

PR #19326, opened 2026-04-03
fix(address): populate street line from place details

jeevan6996:
## Summary
- extract and expose `street` from Google place details (`street_number` + `route`) on the server DTO
- request and type `street` in front-end geo-map place details query
- use `placeData.street` as the preferred value for `addressStreet1` in address autofill
- add regression coverage for query fields and street-line precedence behavior

## Why
Address autocomplete selection currently writes full place text (including city/state/postcode/country) into `addressStreet1`, duplicating values already mapped to dedicated fields.

Fixes #18860

PR #19666, opened 2026-04-14
fix: extract street from Place Details so Address 1 shows only street after auto-fill

HamzaSwitch:
Fixes #18860.

When a user selects an address from the autocomplete suggestions, city/state/zip/country get correctly populated in their subfields, but Address 1 keeps the full address string (including the parts that are now in subfields). This is because the Place Details API returns structured `address_components` including `street_number` and `route`, but `sanitizePlaceDetailsResults` was only extracting city/state/country/postcode and ignoring the street components.

4 files changed:

- `sanitize-place-details-results.util.ts` — extract `street_number` + `route` from address_components, combine into a new `street` field
- `place-details-result.dto.ts` — expose `street` in the GraphQL DTO
- `placeApi.ts` (frontend type) — add `street` to `PlaceDetailsResult`
- `useAddressAutocomplete.ts` — prefer `placeData.street` over the raw autocomplete text for `addressStreet1`

Now when a user selects "123 Main St, New York, NY 10001", Address 1 shows "123 Main St" and the subfields show the rest.

### DIFF
--- a/packages/twenty-front/src/modules/geo-map/types/placeApi.ts
+++ b/packages/twenty-front/src/modules/geo-map/types/placeApi.ts
@@ -13,6 +13,7 @@ export type PlaceAutocompleteResult = {
 };
 
 export type PlaceDetailsResult = {
+  street?: string;
   state?: string;
   postcode?: string;
   city?: string;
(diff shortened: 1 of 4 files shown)


charlesBochet:
Closing as https://github.com/twentyhq/twenty/pull/19326/changes is older, thanks for taking a look @HamzaSwitch, much appreciated

PR #19671, opened 2026-04-14
Fix upgrade commands discovery outside of cli

prastoin:
# Introduction
We were allowing the sequence to be empty in the worker context that was facing an edge case importing the UpgradeModule through the WorkspaceModule god module, no commands were discovered and it was throwing as the sequence must have at least one workspace commands to allow a workspace creation

Though the issue was also applicable to the twenty-server `AppModule` too that was not discovering any commands

## Integration tests were passing
The integration test were importing the `CommandModule` at the nest testing app creating leading to asymmetric testing context
It was a requirement for a legacy commands import and global assignation

## Fix
The `UpgradeModule` now import both `WorkspaceCommandsProviderModule` and `InstanceCommandProviderModule` which ships the commands directly in the module
We could consider moving the commands into the `engine/upgrade` folder

## Concern
Bootstrap could become more and more long to load at both server and worker start
When this becomes a problem we will have to only import the latest workspace command or whatever
For the moment this is not worth it the risk to import not the latest workspace command

### DIFF
--- a/packages/twenty-client-sdk/src/metadata/generated/schema.graphql
+++ b/packages/twenty-client-sdk/src/metadata/generated/schema.graphql
@@ -2193,30 +2193,18 @@ type DeletedWorkspaceMember {
   userWorkspaceId: UUID
 }
 
-type BillingEntitlement {
-  key: BillingEntitlementKey!
-  value: Boolean!
+type Relation {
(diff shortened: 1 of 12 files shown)


Weiko (approved):
LGTM

PR #19675, opened 2026-04-14
Introduce standalone page

bosiraphael:
Add support for standalone pages: a new `PageLayout` type (`STANDALONE_PAGE`) that can be rendered independently at `/page/:pageLayoutId`, not tied to any record or object context.

- New `STANDALONE_PAGE` page layout type
- New `PAGE_LAYOUT` navigation menu item type: adds a `pageLayoutId` foreign key to `NavigationMenuItemEntity`, allowing sidebar items to link directly to standalone pages
- New `GLOBAL_OBJECT_CONTEXT` command menu availability type: separates object-context-dependent commands (Create Record, Import, Export, See Deleted, Create View, Hide Deleted) from truly global ones, so standalone pages only show relevant commands
- Frontend routing & rendering: adds a `/page/:pageLayoutId` route with its own page component, header, and command menu
- Widget rendering refactor
- Instance commands: two fast 1.22 migrations: `pageLayoutId` column + `STANDALONE_PAGE` enum, and `GLOBAL_OBJECT_CONTEXT` availability type enum
- Workspace command: backfills existing command menu items from `GLOBAL` to `GLOBAL_OBJECT_CONTEXT` where appropriate
- Dev seeds: adds a sample "Star History" standalone page with an iframe widget for local development


packages/twenty-front/src/hooks/usePageChangeEffectNavigateLocation.ts
```diff
@@ -39,11 +44,27 @@ export const usePageChangeEffectNavigateLocation = () => {
...
     (objectMetadataItem) => objectMetadataItem.namePlural === objectNamePlural,
   );
+
+  const pageLayoutId = params.pageLayoutId;
+  const isOnPageLayoutPage = isMatchingLocation(
+    location,
+    AppPath.PageLayoutPage,
+  );
+
+  const { data: pageLayoutData, loading: isPageLayoutLoading } = useQuery(
```
Weiko, line 61: Shouldn't we try to pre-load all page layouts before? I'm wondering if we have a state for that already

packages/twenty-front/src/modules/command-menu-item/components/StandalonePageCommandMenu.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+    return {
+      pageType: ContextStorePageType.Standalone,
+      isInSidePanel: false,
+      isPageInEditMode: false,
+      favoriteRecordIds: [],
+      isSelectAll: false,
+      hasAnySoftDeleteFilterOnView: false,
+      numberOfSelectedRecords: 0,
+      objectPermissions: {
```
Weiko, line 57: what's the strategy for permissions? shouldn't we pass that info?

bosiraphael: Those are the default values for the context store, because no object is defined on a standalone page.
But we could think about declaring permissions for standalone pages too.

Weiko (approved):
LGTM

PR #19681, opened 2026-04-14
fix: AI chat not showing credits exhausted or API key errors

FelixMalfait:
## Summary

- **Root cause**: `AgentChatResolver` was missing `@UseInterceptors(AgentGraphqlApiExceptionInterceptor)`. Without it, `BillingException` and `AgentException` were never converted to typed GraphQL errors — they fell through to the generic handler which stripped `subCode` and `userFriendlyMessage`, turning them into featureless `INTERNAL_SERVER_ERROR` responses.
- **Also fixed**: `API_KEY_NOT_CONFIGURED` was re-thrown raw in the exception handler instead of being mapped to `ForbiddenError`, so it lost its code through the same path.
- **Frontend**: The `useAgentChat` catch block silently swallowed mutation errors without setting `agentChatErrorComponentFamilyState`, so `AIChatErrorRenderer` never fired for mutation-time errors (credits exhausted, API key not configured).

## Changes

**Backend (2 files):**
- `agent-chat.resolver.ts` — Added `@UseInterceptors(AgentGraphqlApiExceptionInterceptor)`, aligning with `AgentResolver`
- `agent-graphql-api-exception-handler.util.ts` — Map `BILLING_CREDITS_EXHAUSTED` and `API_KEY_NOT_CONFIGURED` to `ForbiddenError(error)`, which preserves `subCode` and `userFriendlyMessage` via the `BaseGraphQLError` constructor

**Frontend (1 file):**
- `useAgentChat.ts` — Catch block now normalizes the `ApolloError` into an `Error` with `.code` (same pattern as the stream error handler in `useAgentChatSubscription`) and sets it on the error atom. Also clears previous errors on new sends.

## Test plan

- [ ] With billing enabled and credits exhausted, send a chat message → should show "credits exhausted" message with upgrade prompt
- [ ] With no AI API keys configured, send a chat message → should show "API key not configured" message
- [ ] Normal chat flow still works (no regressions from interceptor addition)
- [ ] Stream errors (mid-stream failures) still display correctly

### DIFF
--- a/packages/twenty-front/src/modules/ai/hooks/useAgentChat.ts
+++ b/packages/twenty-front/src/modules/ai/hooks/useAgentChat.ts
@@ -1,3 +1,4 @@
+import { CombinedGraphQLErrors } from '@apollo/client/errors';
 import { useApolloClient } from '@apollo/client/react';
 import { useStore } from 'jotai';
 import { useCallback, useState } from 'react';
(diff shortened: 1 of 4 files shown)


FelixMalfait:
Replaced by https://github.com/twentyhq/twenty/pull/19690 ; codex was in better shape than Opus today

PR #19686, opened 2026-04-14
Fix navbar folder opening lag by deferring navigation until expand animation completes

abdulrahmancodes:
### Before

https://github.com/user-attachments/assets/7d57aa55-8d4c-43e1-8835-f40206e5e453

### After

https://github.com/user-attachments/assets/2457c8ee-fcb9-41ae-a8f7-08f8c7b4a233

### DIFF
--- a/packages/twenty-front/src/modules/navigation-menu-item/display/folder/components/NavigationMenuItemFolderLayout.tsx
+++ b/packages/twenty-front/src/modules/navigation-menu-item/display/folder/components/NavigationMenuItemFolderLayout.tsx
@@ -1,5 +1,5 @@
 import { styled } from '@linaria/react';
-import { useState, type ReactNode } from 'react';
+import { useDeferredValue, useState, type ReactNode } from 'react';
 import { AnimatedExpandableContainer } from 'twenty-ui/layout';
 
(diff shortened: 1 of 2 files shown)


bosiraphael:
Hello @abdulrahmancodes. I don't think we should do this, it adds latency before the navigation. Navigating quickly is more important than having a smooth animation. This animation problem seems to be quite hard to fix, I struggled also trying to fix the navigation bar animation when leaving out the settings page. I've already discussed this with @charlesBochet, he said that it wasn't worth a big refactoring for now. I'll let @charlesBochet confirm though.

etiennejouan:
After discussing with Charles, we’re fine delaying navigation relative to folder opening but the other way around from Abdul’s approach: Abdul had folder open first, then navigate; I give a try with navigate first, then folder open.

Claude, my friend, implemented this with `useDeferredValue`: the expand state can stay one step behind the urgent update so the route change can settle first, then a follow-up render brings the folder opening in sync. 

It’s not a pattern we're fans of, but here it’s scoped to the folder layout. @bosiraphael and @abdulrahmancodes, wdyt ?

charlesBochet (approved):

PR #19696, opened 2026-04-14
Fix duplicate Fields widget

Weiko:
## Context
Tab duplication was broken after the view creation logic was deferred and moved to the FE.

- Duplicating a tab or a FIELDS widget now produces a fully independent copy: new view, new view field groups, new view fields — all with fresh IDs — while preserving any unsaved edits
  from the source widget.
- Removed the backend auto-seed of default view fields / view field groups in ViewService.createOne for FIELDS_WIDGET views. The frontend always sends the complete layout via
  upsertFieldsWidget, so the auto-seed was both redundant and the source of potential bugs.
- Extracted a shared useDuplicateFieldsWidgetForPageLayout hook used by both tab and widget duplication paths, plus a small useCloneViewInMetadataStore helper that clones the FlatView
  in the metadata store and returns the copied flat view fields/groups for the caller.


Devessier (approved):
LGTM

PR #19706, opened 2026-04-14
Link command menu items to specific page layout

bosiraphael:
- Add a `pageLayoutId` foreign key to `CommandMenuItem`, allowing command menu items to be scoped to a specific page layout instead of being globally available
- Filter command menu items by the current page layout on the frontend. Items with a `pageLayoutId` only appear when viewing that layout, while items without one remain globally visible
- Create an effect to track the current page layout ID
- Include a seed example: a "Show Notification" command pinned to the Star history standalone page layout


Weiko (approved):
LGTM

Weiko (approved):

PR #19718, opened 2026-04-15
Hide workflow manual trigger from command menu on "Select All"

thomtrp:
https://github.com/user-attachments/assets/e19c6d23-03c1-49c8-8b83-5c8f5f0f1135

### DIFF
--- a/packages/twenty-server/src/modules/workflow/workflow-trigger/workspace-services/workflow-trigger.workspace-service.ts
+++ b/packages/twenty-server/src/modules/workflow/workflow-trigger/workspace-services/workflow-trigger.workspace-service.ts
@@ -1,5 +1,5 @@
-import { Injectable, Logger } from '@nestjs/common';
 import { msg } from '@lingui/core/macro';
+import { Injectable, Logger } from '@nestjs/common';
 import { isNonEmptyString } from '@sniptt/guards';
 import { type ActorMetadata } from 'twenty-shared/types';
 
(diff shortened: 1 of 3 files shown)


bosiraphael (commented):
To limit the changes until we implement this I would just add a filter on the frontend side to the displayed command menu items. We check if `workflowVersionId` is defined and if the command menu item is of type `RECORD_SELECTION` and we filter out. What do you think @thomtrp?

packages/twenty-server/src/modules/workflow/workflow-trigger/workspace-services/workflow-trigger.workspace-service.ts
```diff
@@ -361,6 +363,10 @@ export class WorkflowTriggerWorkspaceService {
         case 'BULK_RECORDS': {
           availabilityType = CommandMenuItemAvailabilityType.RECORD_SELECTION;
 
+          // Prevent the command menu item from being shown until this is implemented
```
Weiko: Redundant

Weiko (commented):
LGTM, leaving the approval to @bosiraphael though!

packages/twenty-server/src/modules/workflow/workflow-trigger/workspace-services/workflow-trigger.workspace-service.ts
thomtrp: I just wanted to highlight that is a tmp fix. We want to enable workflow runs with select all at some point

FelixMalfait:
Is it that costly to implement? I feel most real use-cases for bulk trigger = you make a list and click select all. It feels odd to just hide it in that case without a user feedback, feels as much as a bug that it disappears as before in my opinion

thomtrp (commented):
@bosiraphael I ended simply adding a warning message. Much simpler

@FelixMalfait this is not really costly but still a strategy to decide and this is not something I want to add now with the upcoming code freeze. I think a warning message is the right call. Wdyt?

bosiraphael (approved):
LGTM

PR #19721, opened 2026-04-15
Add event forwarding stories to the front component renderer

bosiraphael:
Add Storybook stories and example components to test event forwarding through the front component renderer: form events (text input, checkbox, focus/blur, submit), keyboard events (key/code/modifiers), and host API calls (navigate, snackbar, progress, close panel)


packages/twenty-front-component-renderer/src/__stories__/tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
```
Weiko: Why do we need a tsconfig here?

Weiko (approved):
LGTM

packages/twenty-front-component-renderer/src/__stories__/tsconfig.json
bosiraphael: @prastoin and @charlesBochet excluded it from the root `tsconfig` so I needed another one.

charlesBochet: and why was it excluded?

PR #19724, opened 2026-04-15
Fix standalone page migration failing on navigationMenuItem enum type change

charlesBochet:
## Summary
- The `AddStandalonePageFastInstanceCommand` migration was failing with: `default for column "type" cannot be cast automatically to type core."navigationMenuItem_type_enum"`
- The migration was missing `DROP DEFAULT` / `SET DEFAULT` around the `ALTER COLUMN TYPE` for `navigationMenuItem.type`. PostgreSQL cannot automatically cast the existing default (`'VIEW'::old_enum`) to the new enum type.
- The same 3-step pattern (DROP DEFAULT → ALTER TYPE → SET DEFAULT) was already correctly applied for `pageLayout.type` in the same migration — this fix brings `navigationMenuItem.type` in line.


bosiraphael (commented):
LGTM

bosiraphael (approved):

Weiko (approved):

PR #19734, opened 2026-04-15
Support Select All for workflow manual triggers

### LINKED ISSUE #14192: Workflow cannot be triggered for all records
**Scenario:**
1. Log in
2. Create new workflow where trigger is selected record (rest of workflow isn't important)
3. Open record table of object chosen in step 2
4. Click main checkbox located in the header of record table
5. Check command menu

**Actual:** Workflow trigger disappears from command menu

https://github.com/user-attachments/assets/8c9acba5-5958-44c0-b8de-88221390b698

**Expected:** Workflow trigger is visible in command menu

**Note:** If all records are selected one by one, then workflow trigger appears in command menu, only disappears when main checkbox is checked by user

FelixMalfait:
Fixes https://github.com/twentyhq/twenty/issues/14192#issuecomment-4259437397

## Summary
- Move record fetching and payload building from the enrichment hook into `TriggerWorkflowVersionEngineCommand`, following the same component-based pattern as `DeleteRecordsCommand` and `RestoreRecordsCommand`
- The enrichment hook now only stores workflow metadata (`trigger`, `availabilityType`, `availabilityObjectMetadataId`); the component uses `useLazyFetchAllRecords` for exclusion mode (Select All) with full pagination
- `buildTriggerWorkflowVersionPayloads` is now a pure function accepting `selectedRecords: ObjectRecord[]` instead of reading from the Jotai store

Fixes the issue introduced by #19718 which blocked Select All with a warning toast instead of implementing it.

## Test plan
- [ ] Select individual records → run workflow trigger from command menu → works as before
- [ ] Click Select All → run workflow trigger from command menu → fetches all matching records and runs the workflow
- [ ] Select All with some records deselected → correctly excludes those records
- [ ] Global workflows (no object context) → run without payload as before
- [ ] Bulk record triggers → payload wraps records in `{namePlural: [records]}`

### DIFF
--- a/packages/twenty-front/src/modules/command-menu-item/engine-command/hooks/__tests__/useEnrichHeadlessCommandContextApiWithWorkflowVersionTriggerInformation.test.tsx
+++ b/packages/twenty-front/src/modules/command-menu-item/engine-command/hooks/__tests__/useEnrichHeadlessCommandContextApiWithWorkflowVersionTriggerInformation.test.tsx
@@ -9,28 +9,12 @@ import {
 } from '~/generated-metadata/graphql';
 
 const mockFindOneWorkflowVersion = jest.fn();
-const mockEnqueueWarningSnackBar = jest.fn();
-const mockBuildTriggerWorkflowVersionPayloads = jest.fn();
 
 jest.mock('@/object-record/hooks/useLazyFindOneRecord', () => ({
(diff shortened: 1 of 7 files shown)


thomtrp (approved):
LGTM 👍

PR #19746, opened 2026-04-16
Add isUnique update in query + invalidate cache on rollback

thomtrp:
Fix isUnique not sent in field update mutation: Added isUnique and settings to the Pick type in useUpdateOneFieldMetadataItem, which previously silently dropped these properties from the GraphQL payload.

Invalidate cache on failed migration rollback: Added invalidateCache() call in the migration runner's catch block so that a failed migration (e.g., unique index creation failing due to duplicate data) regenerates the Redis hash, preventing stale metadata from persisting in frontend localStorage indefinitely. Was


Weiko (approved):
LGTM

PR #19760, opened 2026-04-16
Add test tab to tool step

thomtrp:
So we can use those as variables.
When the function input is updated, invalidate the input using an effect.

Capture d’écran 2026-04-16 à 14 45 41


Weiko (approved):
LGTM! I'll let @martmull have a 2nd 👀

martmull (approved):

PR #19763, opened 2026-04-16
Add search to add column dropdown

abdulrahmancodes:
https://github.com/user-attachments/assets/4a64fff0-6495-4651-934b-43f4ad0dc966


etiennejouan (commented):
LGTM ! 

Before approving, could you please check sentry comment and take it into account ? search should not be display if availableFieldMetadataItemsToShow is empty + No results should not be display if no search term

etiennejouan (approved):
Looks very good to me !!! Thanks for changes you've done

PR #19777, opened 2026-04-16
Integration tests for Record Page Layouts

Devessier:


Weiko:
Re-investigating after launch

PR #19796, opened 2026-04-17
Fix: aggregate Calculate not updating in dashboard Table widgets

Weiko:
## Context
Picking an aggregate option (Count, Sum, Percentage Not Empty, etc.) in a dashboard Table widget footer did nothing visually — the value never appeared or updated

## Fix
RecordTableWidget was missing RecordIndexTableContainerEffect, which reactively syncs currentView.viewFields[].aggregateOperation from the Apollo cache into the viewFieldAggregateOperationState jotai atom that the footer reads

Screenshot 2026-04-17 at 14 17 47


bosiraphael (approved):
LGTM

PR #19797, opened 2026-04-17
chore: move TABLE_WIDGET view type migration to 1.23 fast instance command

charlesBochet:
## Summary

- Relocates `AddTableWidgetViewTypeFastInstanceCommand` from `1-22/` to `1-23/` and bumps its `@RegisteredInstanceCommand` version from `1.22.0` to `1.23.0`. The original timestamp `1775752190522` is preserved so the command slots chronologically into the existing 1.23 sequence; auto-discovered via `@RegisteredInstanceCommand`, no module wiring change needed.
- Same pattern as #19792 (move `pageLayoutWidget.conditionalAvailabilityExpression` to 1.23).


Weiko (approved):
LGTM

PR #19800, opened 2026-04-17
Add reset page layout in record page layout edit mode tab

Weiko:
## Context
Adds a burger-menu dropdown on the layout customization bar exposing a "Reset record page layout" action, so users can reset a record page layout straight from the edit bar (previously only available in object settings).

Screenshot 2026-04-17 at 18 03 19


charlesBochet (commented):
Code looks complex for no reason, avoid umbrella hooks

charlesBochet (approved):

PR #19808, opened 2026-04-17
Fix cross-app role ownership validation on permission updates

### LINKED ISSUE #19807: App sync still allows cross-app role retargeting through permission update flows
## Bug Description

PR #19783 blocks creating permission entities on roles owned by another application, but the same ownership check is still missing on the update paths for existing permission entities.

As a result, an app can still attempt to retarget an existing:
- permission flag
- object permission
- field permission

to a role owned by another application.

## Current behavior

Create validators call `validateRoleBelongsToCallerApplication()`, but the corresponding update validators do not.

At the same time:
- `roleUniversalIdentifier` participates in diffing / update generation for `permissionFlag`, `objectPermission`, and `fieldPermission`
- universal update relation resolution rewrites `roleUniversalIdentifier` to `roleId`
- update handlers persist the resolved payload with repository `.update(...)`

So the create path enforces app-role ownership, but the update path does not.

## Expected behavior

If an app sync tries to move an existing permission entity from one role to a role owned by another application, sync should fail with `ROLE_BELONGS_TO_ANOTHER_APPLICATION`, just like the create path does.

## Technical inputs

Affected validators:
- `packages/twenty-server/src/engine/workspace-manager/workspace-migration/workspace-migration-builder/validators/services/flat-permission-flag-validator.service.ts`
- `packages/twenty-server/src/engine/workspace-manager/workspace-migration/workspace-migration-builder/validators/services/flat-object-permission-validator.service.ts`
- `packages/twenty-server/src/engine/workspace-manager/workspace-migration/workspace-migration-builder/validators/services/flat-field-permission-validator.service.ts`

A good regression test is:
1. sync App A with a permission entity attached to App A's role
2. sync App B so App B's role exists
3. sync App A again, keeping the same permission entity `universalIdentifier` but moving it under App B's role
4. expect `METADATA_VALIDATION_FAILED` and a nested `ROLE_BELONGS_TO_ANOTHER_APPLICATION` error for that permission metadata type

## Tracking PR
- #19808 fixes the three update validators and adds regression coverage for permission flag / object permission / field permission retarget attempts.

shaun0927:
Closes #19807

## Summary
- apply `validateRoleBelongsToCallerApplication()` to the update validators for permission flags, object permissions, and field permissions
- keep create and update ownership checks symmetric for permission entities introduced by app sync
- add integration coverage for retargeting an existing permission entity from one app role to another app's role

## Why this change
PR #19783 closed the create path, but the update path still accepted the same retargeting shape:
- `roleUniversalIdentifier` is part of the diff / update model for these metadata types
- update relation resolution rewrites `roleUniversalIdentifier` to `roleId`
- update handlers persist the resolved payload

That means cross-app role ownership was enforced on create, but not on update.

## Test plan
- [x] `git diff --check`
- [x] added integration coverage for permission flag / object permission / field permission retarget attempts
- [ ] full local `nx test:integration twenty-server` run
  - blocked in my local workspace snapshot because PostgreSQL and ClickHouse were not available on the expected local ports


prastoin (commented):
Hello there, thanks for your contribution but this is missing coverage

prastoin:
Will attest and tackle related issue in a dedicated PR

PR #19814, opened 2026-04-17
Disable reset to default when custom tab or widget

Weiko:
## Context
"Reset to default" action is rejected by the backend for custom entities because there is no "default" concept for them

## Implementation
Grey out the Reset to default action on record page-layout tabs and widgets when the entity either has no applicationId yet (unsaved draft — previously slipped through the existing check), or belongs to the workspace custom application.

Screenshot 2026-04-17 at 18 50 31


packages/twenty-front/src/modules/page-layout/utils/isCustomOrUnsavedPageLayoutEntity.ts
```diff
@@ -0,0 +1,15 @@
+import { isNonEmptyString } from '@sniptt/guards';
+
+export const isCustomOrUnsavedPageLayoutEntity = ({
```
charlesBochet: isCustomOrNewlyCreatedPageLayout

charlesBochet (approved):

PR #19815, opened 2026-04-17
Harden cross-app role ownership validation on role create and update

shaun0927:
Follow-up to #19808 / #19807.

## Summary
- add `validateRoleBelongsToCallerApplication` to `validateFlatRoleCreation` so an app sync that re-declares a role whose `universalIdentifier` already belongs to another application fails with `ROLE_BELONGS_TO_ANOTHER_APPLICATION` before the centralized `ENTITY_ALREADY_EXISTS` check and before the optimistic role map is mutated.
- add the same check to `validateFlatRoleUpdate` so any update targeting an existing role owned by another application is rejected before `validateRoleIsEditable` or other update-time rules run.
- update the affected snapshot fixtures: both `failing-sync-application-cross-app-permission-on-standard-role` and `failing-sync-application-universal-identifier-conflict` now surface the new nested error in addition to `ENTITY_ALREADY_EXISTS`.

## Why
PR #19783 closed the cross-app role-retarget path on permission creation. PR #19808 mirrored the same guard on the permission update validators for `permissionFlag`, `objectPermission`, and `fieldPermission`. That leaves one remaining shape: the role entity itself. When an app sync declares a role whose `universalIdentifier` is already owned by another application, the centralized `ENTITY_ALREADY_EXISTS` does fire for collisions that reach the role-type map, but the per-type validator lets the build optimistically continue — and for scenarios where the central check misses the shape (e.g. narrower dependency-loading windows) the hijack can poison the optimistic role map with the caller's `applicationUniversalIdentifier`, silently defeating the downstream permission-update guards.

Running the ownership check at the role-type layer closes the asymmetry: the role CREATE and UPDATE validators now fail fast with a dedicated `ROLE_BELONGS_TO_ANOTHER_APPLICATION` error, matching the behaviour already present on permission-level validators.

## Test plan
- [x] `npx nx lint:diff-with-main twenty-server`
- [ ] full `nx run twenty-server:test:integration:with-db-reset`
  - not executed locally (no Postgres/ClickHouse running in this environment); affected suites:
    - `failing-sync-application-cross-app-permission-on-standard-role`
    - `failing-sync-application-universal-identifier-conflict`
    - `failing-role-creation` / `failing-role-update` (unchanged snapshots — same-app scenarios should still pass cleanly because the check only fires on cross-app mismatches)

### DIFF
--- a/packages/twenty-server/src/engine/workspace-manager/workspace-migration/workspace-migration-builder/validators/services/flat-role-validator.service.ts
+++ b/packages/twenty-server/src/engine/workspace-manager/workspace-migration/workspace-migration-builder/validators/services/flat-role-validator.service.ts
@@ -10,6 +10,7 @@ import { type FailedFlatEntityValidation } from 'src/engine/workspace-manager/wo
 import { getEmptyFlatEntityValidationError } from 'src/engine/workspace-manager/workspace-migration/workspace-migration-builder/builders/utils/get-flat-entity-validation-error.util';
 import { type FlatEntityUpdateValidationArgs } from 'src/engine/workspace-manager/workspace-migration/workspace-migration-builder/types/universal-flat-entity-update-validation-args.type';
 import { type UniversalFlatEntityValidationArgs } from 'src/engine/workspace-manager/workspace-migration/workspace-migration-builder/types/universal-flat-entity-validation-args.type';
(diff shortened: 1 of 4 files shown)


prastoin (commented):
Hello there, same than for your other PR should be dedicated tests coverage
Here unrelated tests snapshots catches your exception indirectly

PR #19827, opened 2026-04-17
Fix app design 6

martmull:
Unify application display page and isntalled page


packages/twenty-front/src/pages/settings/applications/tabs/SettingsApplicationDetailAboutTab.tsx
```diff
@@ -1,106 +1,190 @@
...
 
-  const [isLoading, setIsLoading] = useState(false);
+const StyledMainContent = styled.div`
+  flex: 1;
+  min-width: 0;
+  overflow: hidden;
+`;
 
-  const { enqueueErrorSnackBar, enqueueSuccessSnackBar } = useSnackBar();
+const StyledSectionTitle = styled.h2`
```
charlesBochet: I'm surprise we are re-inventing the wheel in all these common components. Compare with what we have in other Settings page

charlesBochet: twenty-front/src/pages/ should not contain the big components. This shold go to twenty-front/src/modules

Looks at other Settings page
~pages is kind of the entrypoint only

packages/twenty-front/src/pages/settings/applications/tabs/SettingsApplicationDetailContentTab.tsx
```diff
@@ -1,147 +1,283 @@
...
+  frontComponents?: { name: string; description?: string | null }[];
+};
+
+type SettingsApplicationDetailContentTabProps = {
+  applicationId: string;
+  installedApplication?: InstalledApplicationForContentTab;
+  manifestContent?: Manifest;
+};
+
+const useObjectAndFieldRows = ({
```
charlesBochet: we don't do that usually (declaring hooks in a component file), not sure about eh value

PR #19829, opened 2026-04-18
fix oauth permission check

neo773:
was regressed due to https://github.com/twentyhq/twenty/pull/19441


packages/twenty-server/src/engine/core-modules/auth/services/create-connected-account.service.ts
```diff
@@ -51,6 +51,7 @@ export class CreateConnectedAccountService {
         await this.globalWorkspaceOrmManager.getRepository<WorkspaceMemberWorkspaceEntity>(
           workspaceId,
           'workspaceMember',
+          { shouldBypassPermissionChecks: true },
```
FelixMalfait: For the job/command it feels correct to bypass but for the services you probably have a user context available you can pass?

FelixMalfait (approved):

FelixMalfait (approved):

PR #19850, opened 2026-04-19
fix(front): apply object type translations to details panel widget titles

### LINKED ISSUE #19790: Object Type translations are not applied to details panel
## Bug Description

Translations of object types (or customizations of their singular/plural labels) are not applied to the related field labels in a details panel. See screenshots below where mismatches are highlighted.

**Example:** 

Image 

Image

Image

## Expected behavior
Expected that translated or name customizations would propagate throughout the app

## Technical inputs
- version 1.21.2

aaryyann:
Closes #19790

## What
Object type translations and custom labels (renames) were not applied to
widget titles in the record details side panel (e.g. "People", "Opportunities",
"Company"), causing a mismatch with the rest of the app.

## Why
Standard page-layout widgets store their title as a static string seeded from
`standard-*-page-layout.config.ts` into `core.pageLayoutWidget.title`. That
value is fixed at seed time, so it doesn't follow subsequent object renames
or locale changes.

## How
- Added `useWidgetDerivedTitle` hook at
  `packages/twenty-front/src/modules/page-layout/widgets/hooks/useWidgetDerivedTitle.ts`.
  For a `FIELD` widget backed by a relation field, it resolves the target
  object from `objectMetadataItems` and returns its current `labelPlural`
  (ONE_TO_MANY) or `labelSingular` (MANY_TO_ONE). Non-relation widgets fall
  back to the stored `widget.title`.
- Wired the hook into `RecordPageWidgetRenderer` and passed the derived
  title to `WidgetCardShell` as an optional `title` prop, which overrides
  `widget.title` when present.
- Non-record-page widgets (dashboards etc.) are unaffected.

Translations work for free: the server's `MinimalMetadataService` already
applies Lingui translations to `labelPlural` / `labelSingular` before
returning them, so locale switches propagate without additional code.

## Testing
- Renamed Company → "Org" in Settings → Data model → reopened a Person
  record; the side-panel widget formerly titled "Company" now reads "Org".
- Renamed Person plural → "Humans" → reopened a Company record; the
  "People" widget now reads "Humans".
- Switched app locale; relation widget titles switched with it.
- Verified non-relation widgets (timeline, tasks, notes, system fields)
  are unchanged.

## Screenshot
image

Testing : 
image

### DIFF
--- a/packages/twenty-front/src/modules/page-layout/widgets/components/RecordPageWidgetRenderer.tsx
+++ b/packages/twenty-front/src/modules/page-layout/widgets/components/RecordPageWidgetRenderer.tsx
@@ -1,6 +1,7 @@
 import { type PageLayoutWidget } from '@/page-layout/types/PageLayoutWidget';
 import { WidgetCardShell } from '@/page-layout/widgets/components/WidgetCardShell';
 import { useWidgetActions } from '@/page-layout/widgets/hooks/useWidgetActions';
+import { useWidgetDerivedTitle } from '@/page-layout/widgets/hooks/useWidgetDerivedTitle';
 import { useWidgetRendererState } from '@/page-layout/widgets/hooks/useWidgetRendererState';
 import { useIsFeatureEnabled } from '@/workspace/hooks/useIsFeatureEnabled';
 import { styled } from '@linaria/react';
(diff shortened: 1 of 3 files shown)


Weiko (commented):
Hi @aaryyann, closing in favor of #19890, which landed yesterday with a slightly different approach.
Your hook covered relation-field widgets on record pages but tabs and non-relation standard widgets would have stayed hardcoded English. The prefer fix is to have the backend as a single source of truth with 1 key per widget (no particular path for relation widgets even though in practice they use the object's name)
Thanks for your contribution!

aaryyann:
@Weiko Thanks for the explanation! Makes sense to have backend as single source of truth. I'll check out to learn the pattern for future contributions.

PR #19859, opened 2026-04-19
feat: clean OAuth2 support for custom IMAP/SMTP

ViVaLaDaniel:
## Summary
This is a clean version of the OAuth2 support for custom IMAP and SMTP connections, as requested by Felix.

## Changes
- Updated ConnectionParameters to include uthMethod, ccessToken, and 
efreshToken.
- Updated ImapSmtpCaldavService to use OAuth2 authentication for both IMAP and SMTP when specified.

This PR contains ONLY the relevant code changes (2 files) and passes basic validation.

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/imap-smtp-caldav-connection/dtos/imap-smtp-caldav-connection.dto.ts
+++ b/packages/twenty-server/src/engine/core-modules/imap-smtp-caldav-connection/dtos/imap-smtp-caldav-connection.dto.ts
@@ -22,8 +22,17 @@ export class ConnectionParameters {
    * While encrypting it could provide an extra layer of defense, we have decided not to,
    * as database access implies a broader compromise. For context, see discussion in PR #12576.
    */
-  @Field(() => String)
-  password: string;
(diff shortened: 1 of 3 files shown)


ViVaLaDaniel:
Hi Felix! I'm very sorry for the mess in the first PR. My local dev environment got completely out of sync with the upstream main branch, which led to a lot of unrelated file changes and deleted assets being included in the commit.

I've now completely reset my workspace and created this new, clean PR. It contains only the two necessary files for OAuth2 support in IMAP/SMTP and should pass all your automated checks. Thanks for your patience and for pointing it out!

ViVaLaDaniel:
Quick update: I just pushed a few more fixes to resolve some TypeScript type mismatches in the DTOs that were caught by your CI. Now it should be fully passing. Thanks again!

neo773:
Closing this PR as it's not the right approach, OAuth requires an authentication url and a callback url
What's your use case for this? I would recommend opening a GitHub discussion first
And we already support OAuth for Google and Microsoft which covers 95% use cases

PR #19868, opened 2026-04-20
fix 1.22 upgrade command add-workspace-id-to-indirect-entities

### LINKED ISSUE #19863: Migration failure when upgrading from v1.21 to v1.23 via latest tag: AddWorkspaceIdToIndirectEntities fails with 'column already exists'
## Bug Description

When running `twentycrm/twenty:latest` and the image jumps multiple versions at once (e.g. v1.21 → v1.23), the upgrade migration sequence fails in a cascade that leaves the server in a crash-restart loop.

**Root cause chain:**

1. `1.22.0_AddWorkspaceIdToIndirectEntitiesFastInstanceCommand` tries to `ALTER TABLE core."twoFactorAuthenticationMethod" ADD "workspaceId" uuid` — but `workspaceId` already exists on that table. The migration has **no `IF NOT EXISTS` guard**, so it throws `ERROR 42701: column "workspaceId" of relation "twoFactorAuthenticationMethod" already exists` and is marked `failed`.

2. This blocks all subsequent 1.22.0 workspace commands from running.

3. 1.22.0 workspace commands (e.g. `AddSendEmailRecordSelectionCommandMenuItems`) fail with `column CommandMenuItemEntity.pageLayoutId does not exist` because the TypeORM entities already reference 1.23.0 columns (`pageLayoutId`, `conditionalAvailabilityExpression`) that haven't been added yet.

4. This in turn blocks all 1.23.0 instance migrations from running → full deadlock.

**Server enters infinite restart loop.** The `upgradeMigration` table accumulates rows with `status = 'failed'`, `attempt` incrementing each cycle.

## Steps to Reproduce

1. Run `twentycrm/twenty:v1.21.x` (or pin to an older digest)
2. Pull `twentycrm/twenty:latest` which resolves to v1.23.x
3. Restart containers → migration loop begins

## Expected Behavior

- `AddWorkspaceIdToIndirectEntities` should use `ADD COLUMN IF NOT EXISTS` for each table, or check existence before altering.
- Alternatively, the upgrade runner should catch `42701` (duplicate column) and treat it as a no-op success rather than a hard failure.

## Affected migration

```
packages/twenty-server/src/database/commands/upgrade-version-command/1-22/
  1-22-instance-command-fast-1775758621017-add-workspace-id-to-indirect-entities.ts
```

The `up()` method runs:
```ts
for (const table of TABLES) {
  await queryRunner.query(`ALTER TABLE "core"."${table}" ADD "workspaceId" uuid`);
}
```

A safe fix:
```ts
await queryRunner.query(`ALTER TABLE "core"."${table}" ADD COLUMN IF NOT EXISTS "workspaceId" uuid`);
```

## Manual workaround applied

1. Add `workspaceId` with `IF NOT EXISTS` to the 7 affected tables manually via psql
2. Apply all 1.23.0 fast-instance schema changes manually (columns + enum renames)
3. Update `upgradeMigration` records to `status = 'completed'`
4. Restart → server recovers

## Environment

- Self-hosted via Docker Compose
- Image: `twentycrm/twenty:latest` (digest `ecf0b2051037`, built 2026-04-19)
- Previous version: `v1.21.0` (image `8b8d82ff467c`)
- PostgreSQL 16-alpine

neo773:
/closes #19863

### DIFF
--- a/packages/twenty-server/src/database/commands/upgrade-version-command/1-22/1-22-instance-command-fast-1775758621017-add-workspace-id-to-indirect-entities.ts
+++ b/packages/twenty-server/src/database/commands/upgrade-version-command/1-22/1-22-instance-command-fast-1775758621017-add-workspace-id-to-indirect-entities.ts
@@ -21,7 +21,7 @@ export class AddWorkspaceIdToIndirectEntitiesFastInstanceCommand
   public async up(queryRunner: QueryRunner): Promise<void> {
     for (const table of TABLES) {
       await queryRunner.query(
-        `ALTER TABLE "core"."${table}" ADD "workspaceId" uuid`,
+        `ALTER TABLE "core"."${table}" ADD COLUMN IF NOT EXISTS "workspaceId" uuid`,
       );
     }
   }
@@ -29,7 +29,7 @@ export class AddWorkspaceIdToIndirectEntitiesFastInstanceCommand
   public async down(queryRunner: QueryRunner): Promise<void> {
     for (const table of TABLES) {
       await queryRunner.query(
-        `ALTER TABLE "core"."${table}" DROP COLUMN "workspaceId"`,
+        `ALTER TABLE "core"."${table}" DROP COLUMN IF EXISTS "workspaceId"`,
       );
     }
   }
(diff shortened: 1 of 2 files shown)


Weiko (approved):

Simon2046:
I'm stay in version 1.21.2. And can't upgrade to 1.22.x nor 1.23.x
I tried to upgrade 1.22.6 with docker but failed. With wich version I can upgrade successfuly?

Simon2046:
> I'm stay in version 1.21.2. And can't upgrade to 1.22.x nor 1.23.x I tried to upgrade 1.22.6 with docker but failed. With wich version I can upgrade successfuly?

Successfuly upgraded from 1.21.2 to 1.23.9

PR #19871, opened 2026-04-20
Fix AggregateCompanies query firing on sign-in page

abdulrahmancodes:


etiennejouan (commented):
Closing the PR. It works but the real problem is this aggregateRecords call that should not happen if user is not authenticated.
Thomas will work on this issue after release week.

PR #19872, opened 2026-04-20
Fix Email composer rich text to HTML conversion 

neo773:


packages/twenty-front/src/modules/activities/emails/constants/AttachmentKeywordsByLocale.ts
```diff
@@ -0,0 +1,231 @@
+export const ATTACHMENT_KEYWORDS_BY_LOCALE: Record<string, string[]> = {
```
FelixMalfait: It's not a bad idea to do it like this but I don't feel like it's the future. This will trigger a lot of false positives and can also miss things. Imo the right long-term feature is more about running a model on every message to proofread, it's not just around attachment but also checking spelling or making other smart recommandations for improvements

neo773: I could remove it for now then and revisit later

FelixMalfait:
@neo773 yes let's revert the attachment warning and merge just the HTML fix

cc @Bonapara fyi my comment before
> It's not a bad idea to do it like this but I don't feel like it's the future. This will trigger a lot of false positives and can also miss things. Imo the right long-term feature is more about running a model on every message to proofread, it's not just around attachment but also checking spelling or making other smart recommandations for improvements

FelixMalfait (approved):

PR #19884, opened 2026-04-20
fix(ui): make CardPicker hover cover the whole card and align content left

charlesBochet:
## Summary

Two small visual issues with the shared `CardPicker` (used in the Enterprise plan modal and the onboarding plan picker):

- Labels like \`Monthly\` / \`Yearly\` were center-aligned inside their cards while the subtitle (\`\$25 / seat / month\`) stayed left-aligned, because the underlying \`<button>\` element's default \`text-align: center\` was leaking into the children.
- The hover background was painted on the same element that owned the inner padding, so the hover surface didn't visually feel like the whole card.

This PR:
- Moves the content padding into a new \`StyledCardInner\` so the outer \`<button>\` is just the card chrome (border + radius + background + hover).
- Adds \`text-align: left\` so titles align with their subtitles.
- Hoists \`cursor: pointer\` out of \`:hover\` (it should be on by default for the card).

Affects:
- \`EnterprisePlanModal\` (Settings → Enterprise)
- \`ChooseYourPlanContent\` (onboarding trial picker)

### DIFF
--- a/packages/twenty-front/src/modules/ai/components/AgentChatProvider.tsx
+++ b/packages/twenty-front/src/modules/ai/components/AgentChatProvider.tsx
@@ -9,9 +9,9 @@ export const AgentChatProvider = ({
 }) => {
   const isAiEnabled = useIsFeatureEnabled(FeatureFlagKey.IS_AI_ENABLED);
 
-  if (!isAiEnabled) {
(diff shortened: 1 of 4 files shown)


Weiko (approved):
LGTM

PR #19912, opened 2026-04-21
fix(settings): force display "Standard" and "Custom" for app chips

charlesBochet:
## Summary
- Override the `AppChip` label so the Twenty standard application always renders as `Standard` and the workspace custom application always renders as `Custom`, instead of leaking each app's underlying name (e.g. `Twenty Eng's custom application`).
- Detection mirrors the logic already used in `useApplicationAvatarColors`, relying on `TWENTY_STANDARD_APPLICATION_UNIVERSAL_IDENTIFIER` / `TWENTY_STANDARD_APPLICATION_NAME` and `currentWorkspace.workspaceCustomApplication.id`.
- The `This app` label for the current application context and the original `application.name` fallback for any other installed app are preserved.

## Affected UI
- Settings → Data model → Existing objects (App column).
- Anywhere else `AppChip` / `useApplicationChipData` is used.

### DIFF
--- a/packages/twenty-front/src/modules/applications/hooks/useApplicationChipData.ts
+++ b/packages/twenty-front/src/modules/applications/hooks/useApplicationChipData.ts
@@ -7,6 +7,10 @@ import { currentWorkspaceState } from '@/auth/states/currentWorkspaceState';
 import { useAtomStateValue } from '@/ui/utilities/state/jotai/hooks/useAtomStateValue';
 import { t } from '@lingui/core/macro';
 import { useContext } from 'react';
+import {
+  TWENTY_STANDARD_APPLICATION_NAME,
+  TWENTY_STANDARD_APPLICATION_UNIVERSAL_IDENTIFIER,
+} from 'twenty-shared/application';
 import { isDefined } from 'twenty-shared/utils';
 
 type UseApplicationChipDataArgs = {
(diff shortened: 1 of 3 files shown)


packages/twenty-front/src/modules/applications/hooks/useApplicationChipData.ts
```diff
@@ -47,9 +51,26 @@ export const useApplicationChipData = ({
   const isCurrent =
     isDefined(currentApplicationId) && currentApplicationId === applicationId;
 
+  const isStandard =
+    application.universalIdentifier ===
+      TWENTY_STANDARD_APPLICATION_UNIVERSAL_IDENTIFIER ||
+    application.name === TWENTY_STANDARD_APPLICATION_NAME;
```
Weiko: is application.name really necessary?

Weiko (approved):

Qodo-Free-For-OSS:
Hi, `AppChip` no longer accepts `to`, `variant`, or `size` and no longer renders a `Chip`/`LinkChip`, so any existing consumers relying on link behavior or chip styling will fail to compile or lose expected interactivity/styling.

**Severity:** remediation recommended | **Category:** maintainability

**How to fix:** Restore AppChip backward compatibility

> ### Issue description
> `AppChip` was changed from a `Chip`/`LinkChip` wrapper to a custom `div` and its props were narrowed to only `applicationId`/`className`. This is a breaking change for any call sites that previously passed `to`, `variant`, or `size`, and it removes the link/styling behavior previously provided by `LinkChip`/`Chip`.
> 
> ### Issue Context
> In-repo usage appears limited, so CI may still pass, but `AppChip` is an exported component and can be consumed elsewhere in the monorepo or by future code.
> 
> ### Fix Focus Areas
> - packages/twenty-front/src/modules/applications/components/AppChip.tsx[1-48]
> 
> ### What to change
> - Reintroduce optional props: `to?: string`, `variant?: ChipVariant`, `size?: ChipSize`.
> - Render `LinkChip` when `to` is provided, otherwise render `Chip` (as before), using the same `Avatar` as `leftComponent`.
> - Keep the new label logic in `useApplicationChipData` unchanged so the displayed label remains `Standard`/`Custom`/`This app`.
> - If the new non-chip visual style is required, consider creating a new component (e.g., `AppInlineLabel`) and keep `AppChip` as a backward-compatible wrapper to avoid breaking consumers.

---
*[Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review - free for open-source.*

PR #19920, opened 2026-04-21
[Website] Fix flickering of faq illustration.

mabdullahabaid:
Before:

https://github.com/user-attachments/assets/da189675-5de2-47be-b491-0d52eb11331e

After:

https://github.com/user-attachments/assets/7e382523-3539-4978-873c-e4ea79e264a7


FelixMalfait (approved):
Approving for launch day but this is the kind of code I think we'll probably have to improve/refactor :)

PR #19935, opened 2026-04-21
Optimize 3d models

neo773:
Remove unused textures and UV data from 3D models
Also simplified meshes where possible and re-compressed
Verified that no visual regression was seen but needs to be double checked in case I missed something

Total model size: 5.1MB -> 1.5MB


FelixMalfait (approved):

neo773:
Prompt for future optimizations
cc @mabdullahabaid 

````
Optimize static assets in $ARGUMENTS (default: current project's public directory). Think like a demoscene programmer who had to fit Mario Kart onto a cartridge — every byte must justify its existence.

## Mindset

Never blindly run compression tools. First understand what the asset IS, how it's CONSUMED, and what the renderer ACTUALLY NEEDS. The biggest wins come from removing data that's invisible in the final output, not from tweaking encoder settings.

## Process

### 1. Inventory

Find all static assets (models, images, fonts, animations, audio, video). List them with file sizes, sorted largest first. Identify duplicates by checksum.

### 2. Trace the consumer

For each asset type, read the code that loads and renders it. Answer:
- What data does the renderer actually read?
- What gets thrown away or overridden by shaders/post-processing?
- Are there attributes, textures, channels, or metadata that exist in the file but never reach the screen?

This is the most important step. A 1024x1024 texture fed into a shader that only reads brightness is 99% waste. UV coordinates on a model rendered through a post-process effect are dead weight. An alpha channel on an image composited onto a solid background does nothing.

### 3. Remove the invisible

Strip everything the renderer doesn't need. Examples:
- Textures/UVs on 3D models rendered through halftone/toon/wireframe shaders
- Alpha channels on opaque images
- Unused color profiles or EXIF metadata
- Audio channels nobody hears (stereo → mono for UI sounds)
- Font glyphs for character sets never used

### 4. Reduce precision to match output

Match data precision to what's perceptible at render size:
- 1024x1024 texture displayed at 200px → resize to 256x256
- float32 normals through a low-frequency shader → int8 normals
- 60fps animation displayed at 30fps → halve keyframes
- 44.1kHz audio for a click sound → 22kHz

### 5. Optimize encoding

Only now apply format-specific compression. The right encoder depends on what's left after steps 3-4:
- Images: pick format by content (WebP/AVIF for photos, SVG for vectors, PNG for sharp edges)
- 3D models: Draco or meshopt after geometry simplification
- Fonts: subset to used glyphs, woff2
- Video: appropriate codec/bitrate for actual playback size

### 6. Verify

- Compare every file against original. REVERT any that got bigger.
- Visual/audio verification — confirm output looks/sounds correct.
- Print before/after table with per-file and total savings.

## Example: GLB 3D models through halftone shader

Investigated renderer, found models carry metallicRoughness textures + UV coordinates but are rendered through a two-pass halftone shader that reduces everything to monochrome dashes. The shader only reads aggregate RGB brightness — all PBR texture detail is invisible.

Pipeline applied:
1. Decode Draco → simplify mesh (--error 0.001) → weld → prune → re-encode Draco
2. Strip all textures + TEXCOORD_0 attributes via gltf-transform SDK script
3. Revert any file that got larger than original

Result: 5.1MB → 1.5MB (70% reduction) with zero visual difference.

## Rules

- Always read the rendering code before optimizing. No blind compression.
- Biggest wins come from removing unused data, not better encoding.
- Never keep an optimized file that's larger than the original.
- Test visually after every batch. Shaders can amplify subtle geometry changes.
- Check for exact duplicates (by checksum) across directories.


```

mabdullahabaid:
@neo773 perfect, thank you for sharing!

Issue #19900, opened 2026-04-20
Inline creation and editing for relational objects within Record Page Layouts


rs0125:
### Current Behavior
Currently, when working inside a record (e.g., an Opportunity) and needing to associate it with a new or updated relational object (like a Company or a Person/POC), the UI forces a context switch. Clicking to add or edit these relational fields opens a completely new modal or page.

For teams doing high-velocity data entry, this breaks the workflow. A team member often has to create the Opportunity, navigate away to create the Company, navigate away again to create the POC, and then tie them all together.

### The Proposed Solution
Introduce inline creation and editing for relational fields directly within the Record Page Layout widgets.

Instead of routing to a new modal, the UI could utilize an expandable accordion interface or a nested sub-form right inside the side panel.

- Inline Create: Clicking "Add new" drops down a nested form with the required fields for that related object.

- Inline Edit: Clicking an existing relation expands an accordion card allowing quick edits to the related object's core fields (e.g., updating a POC's phone number without leaving the Opportunity view).

### Why it matters
With the massive improvements shipped in the v1.22/v1.23 layout engine, the infrastructure is there for highly customizable record pages. Allowing inline relational editing would dramatically reduce clicks, eliminate context-switching, and massively improve the UX for sales and ops teams who need to manage complex data relationships from a single pane of glass.

### Additional Context
This aligns well with modern CRM UX standards (e.g., Linear's issue linking or Notion's inline page creation) where the user's primary context is strictly preserved.

Learnobug:
hey, i would love work on it, can I get this issue?

thomtrp:
Hi @rs0125, for feature requests, please use Github Discussions. Thank you!

PR #19941, opened 2026-04-21
feat: add inline creation and editing for relational objects in Recor…

Learnobug:
…d Page Layouts

Implements #19900 - when clicking "Add New" in a relation section, instead of routing to the side panel, an accordion-style inline form expands directly within the Record Page Layout widget.

- Add RecordDetailInlineRelationCreateForm: inline form component that renders a RecordFieldList for the new related record and provides Done/Discard actions
- Add useCreateRelatedRecord hook: creates a related record and wires up the relation without opening the side panel
- Modify RecordDetailRelationSection: manages inlineCreateRecordIds state and passes onInlineCreate callback down to the dropdown
- Modify RecordDetailRelationRecordsList: renders inline create forms above the existing relation records list
- Modify RecordDetailRelationSectionDropdownToOne and ToMany: use onInlineCreate when provided, falling back to side panel for junction relations and non-inline contexts


charlesBochet:
Hi @Learnobug Thanks a lot for the PR and sorry for the slow review. Unfortunately this one is not meeting our quality guidelines and while the feature would be useful we are not sure how we would integrate it. Let's keep working from discussions.
Closing this one

PR #19947, opened 2026-04-21
feat(sdk): confirm authentication method on remote add

FelixMalfait:
## Summary

`yarn twenty remote add` only prints `✓ Default remote set to X.` after authenticating. When using the OAuth path, the browser flow happens silently — there's no line that says _"you authenticated"_ — so users (including me this morning while installing a Twenty app) are left wondering whether auth actually completed and which method was used.

This PR adds explicit confirmation of the auth step:

**New remote via OAuth**
```
✓ Remote "myremote" added (https://app.twenty.com) via OAuth.
✓ Default remote set to "myremote".
```

**New remote via API key**
```
✓ Remote "myremote" added (https://app.twenty.com) via API key.
✓ Default remote set to "myremote".
```

**Re-authenticating an existing remote**
```
✓ Re-authenticated "myremote" via OAuth.
✓ Default remote set to "myremote".
```

## Implementation

- `authenticate()` now returns the method actually used (`'OAuth' | 'API key'`) instead of `void`. This correctly surfaces OAuth → API-key fallback: if OAuth fails and the user drops into the API-key prompt, the success line reflects that.
- New-remote and re-auth paths print distinct messages so the user can tell which path they took.
- No new API calls — method name comes from which branch of `authenticate()` succeeded.

## Test plan

- [x] `nx typecheck twenty-sdk` — clean
- [x] `nx lint twenty-sdk` — clean
- [ ] Manual smoke test: `yarn twenty remote add --as test --api-url ...` via OAuth, API key, and re-auth

### DIFF
--- a/packages/twenty-sdk/src/cli/commands/remote.ts
+++ b/packages/twenty-sdk/src/cli/commands/remote.ts
@@ -18,20 +18,29 @@ const deriveRemoteName = (url: string): string => {
   }
 };
 
-const authenticate = async (apiUrl: string, apiKey?: string): Promise<void> => {
-  const result = apiKey
-    ? await authLogin({ apiKey, apiUrl })
-    : await runOAuthWithApiKeyFallback(apiUrl);
+type AuthMethod = 'OAuth' | 'API key';
 
-  if (!result.success) {
-    console.error(chalk.red('✗ Authentication failed.'));
-    process.exit(1);
+const authenticate = async (
+  apiUrl: string,
+  apiKey?: string,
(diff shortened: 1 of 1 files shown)


FelixMalfait:
@martmull does this seem relevant to you?

This was suggested by Claude when I tried to install an app and asked what went wrong / what kind of feedback was lacking and made the experience harder for him (he didn't understand I had successfully logged in to a browser tab)

charlesBochet (approved):

PR #19949, opened 2026-04-21
fix(gmail): fallback to text/html when text/plain is missing

Wong789:
## Summary

When importing emails from non-Gmail servers (Yandex SES Android email, Amazon SES, and other SMTP providers), the email body may only contain 	ext/html without a 	ext/plain alternative. The original getBodyData function only looked for 	ext/plain, causing it to return undefined for these emails, resulting in empty body text.

## Changes

- Modified getBodyData to recursively search all parts for 	ext/plain
- Added fallback to 	ext/html if no 	ext/plain is found
- Returns HTML data so the caller can handle it appropriately

## Test Case

Emails from providers like Yandex SES Android email that send HTML-only emails will now have their body content properly extracted instead of being empty.

---
This PR relates to the general IMAP/email import functionality improvements.

### DIFF
--- a/packages/twenty-server/src/modules/messaging/message-import-manager/drivers/gmail/utils/get-body-data.util.ts
+++ b/packages/twenty-server/src/modules/messaging/message-import-manager/drivers/gmail/utils/get-body-data.util.ts
@@ -1,12 +1,48 @@
 import { type gmail_v1 as gmailV1 } from 'googleapis';
 
-export const getBodyData = (message: gmailV1.Schema$Message) => {
-  const firstPart = message.payload?.parts?.[0];
+/**
+ * Recursively finds a part with the given mimeType in the parts array.
+ */
+const findPartByMimeType = (
+  parts: gmailV1.Schema[] | undefined,
+  mimeType: string,
+): gmailV1.Schema | undefined => {
+  if (!parts) return undefined;
 
-  if (firstPart?.mimeType === 'text/plain') {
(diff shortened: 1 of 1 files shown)


Wong789:
Hi, I noticed this PR shows as "out-of-date with the base branch". Would you like me to rebase with the latest main? Or can you merge it as-is? Happy to update if needed. Thanks!

Wong789:
@cubic-dev-ai Thanks for the review! I believe this case is already handled - in the loop I skip any part that has a filename property (attachments). So text attachments would be filtered out before checking mimeType. Could you provide a specific test case where this would still incorrectly match a text attachment? Happy to address if I am missing something.

Wong789:
@cubic-dev-ai Hi! I wanted to follow up on two things:

1. **sentry-bot HIGH bug**: The report says "the new fallback to text/html in getBodyData is not handled by the caller, which always processes the body as text/plain". Is this something I should fix in this PR, or is it a separate issue in the caller code?

2. **Rebase**: This PR is out-of-date with main. Would you like me to rebase before merge? Happy to do so.

3. **Approval**: If the code looks good to you now, could you approve this PR so it can be merged? Thanks!

Wong789:
Hi! I've addressed all the review comments. The CI checks are passing. Could a human maintainer please review and approve this PR? The PR adds a fallback to text/html when text/plain is not available in email messages from non-Gmail SMTP providers. Thanks!

PR #19973, opened 2026-04-22
feat(app): infrastructure for pre-installed apps

FelixMalfait:
**PR 1 of 2.** Follow-up PR ships the Exa app, sets it as a default pre-installed app, and removes the current `WebSearchTool` / `WebSearchService` / `ExaDriver`. This PR adds the plumbing; no user-visible change yet.

## Summary

- Server admins can declare a list of npm app packages to auto-install on every new workspace and backfill onto existing workspaces via CLI.
- Server-level secrets (like Exa's API key) live on the `ApplicationRegistration` (one row per server, encrypted) and are injected into logic function execution env at runtime. No more per-workspace storage of global secrets.
- A generic `POST /app/billing/charge` endpoint lets app logic functions emit workspace usage events for metered features. Exa uses it in PR 2; future apps (call recorder, etc.) reuse it.
- `LogicFunctionToolProvider` tool name prefix changes `logic_function_` → `app_`. Shorter, accurate (they come from installed apps).

## What's in this PR

**Logic function executor — server-level variables**
- `LogicFunctionExecutorService.getExecutionEnvVariables` now resolves env vars in the order: hardcoded defaults → `ApplicationRegistrationVariable[]` (server-level) → `ApplicationVariable[]` (workspace-level override). The manifest `serverVariables` schema has existed; this closes the loop.

**Config**
- `PRE_INSTALLED_APPS` — comma-separated list of npm packages. Default: empty.

**\`PreInstalledAppsService\`** (new module)
- \`onApplicationBootstrap()\` — fetches each package's manifest from the app registry CDN, upserts an \`ApplicationRegistration\`, and seeds declared \`serverVariables\` from matching env vars (e.g. \`EXA_API_KEY\` env → encrypted registration variable).
- \`installOnWorkspace(workspaceId)\` — installs all pre-installed apps on a single workspace. Tolerates per-app failures.

**Auto-install on new workspace activation**
- \`WorkspaceService.prefillCreatedWorkspaceRecords\` invokes \`installOnWorkspace\` after prefilling standard records. Non-blocking on failure.

**Backfill CLI command**
- \`install-pre-installed-apps\` — iterates active and suspended workspaces, installs pre-installed apps that aren't yet installed. Idempotent. Run after changing \`PRE_INSTALLED_APPS\`.

**App billing endpoint**
- \`POST /app/billing/charge\`. Authenticated via \`APPLICATION_ACCESS\` token (already injected into logic function execution env as \`DEFAULT_APP_ACCESS_TOKEN\`). Body: \`{ creditsUsedMicro, quantity, unit, operationType, resourceContext? }\`. Emits \`USAGE_RECORDED\` with \`applicationId\` as \`resourceId\`. Generic — reusable by any app.

**Tool name prefix**
- \`LogicFunctionToolProvider.buildLogicFunctionToolName\` now produces \`app_<name>\` instead of \`logic_function_<name>\`. Only affects tools sourced from logic functions; other tool providers unchanged.

## Stats

- 16 files, +501 / −2
- 7 new files (1 command, 1 service × 2, 1 controller, 1 DTO, 2 modules)
- Typecheck: 7 pre-existing errors, zero new
- Prettier clean

## Behavior deltas

- **\`PRE_INSTALLED_APPS\` default = empty**: existing servers see no change on merge.
- **\`ApplicationRegistrationVariable\` is now read by the executor**: apps that were using manifest \`serverVariables\` but expecting them to be ignored by the executor will now see them injected. No apps ship with \`isTool: true\` logic functions today, so this is latent — first consumer is Exa in PR 2.
- **Tool prefix**: currently no logic-function tools are named \`logic_function_*\` in any production flow. The prefix change affects only future tools emitted by \`LogicFunctionToolProvider\`.

## Risks

- **CDN unavailability at startup**: if the app registry CDN is down, \`ensureRegistrationsExist\` logs warnings but doesn't block server start. Installation on new workspaces during this window will find no registrations and log a non-blocking error. Backfill command can retry after CDN recovers.
- **Cold-start overhead**: \`ensureRegistrationsExist\` is called once per process on bootstrap. Current configurable default is empty, so zero overhead. When an admin sets \`PRE_INSTALLED_APPS\`, they accept one HTTP call per package at boot.
- **Server-level variables flow**: \`ApplicationRegistrationVariable.encryptedValue\` is shared by all workspaces of a server. Appropriate for a single-tenant Exa key. Not appropriate for per-tenant keys — those go in workspace-level \`ApplicationVariable\` and override.

## Test plan

- [ ] \`npx nx typecheck twenty-server\` passes (verified: 7 pre-existing unrelated errors, zero new)
- [ ] Set \`PRE_INSTALLED_APPS=@twenty-apps/hello-world\` (or any real npm-published app), \`HELLO_WORLD_API_KEY=xxx\`, restart server: \`ApplicationRegistration\` row is upserted, \`ApplicationRegistrationVariable\` for HELLO_WORLD_API_KEY is populated (encrypted).
- [ ] Create a new workspace: the app is auto-installed, \`ApplicationEntity\` row created, \`LogicFunctionEntity\` rows created.
- [ ] Existing workspace: run \`yarn nx run twenty-server:command install-pre-installed-apps\`: apps install across all workspaces, idempotent on re-run.
- [ ] Trigger a logic function that reads \`process.env.HELLO_WORLD_API_KEY\`: value resolves from the server-level \`ApplicationRegistrationVariable\`.
- [ ] Log a charge from the handler: \`POST /app/billing/charge\` with \`Authorization: Bearer \$DEFAULT_APP_ACCESS_TOKEN\` body \`{creditsUsedMicro: 1000, quantity: 1, unit: "INVOCATION", operationType: "WEB_SEARCH"}\` → returns \`{success: true}\`, \`USAGE_RECORDED\` event emitted with correct \`resourceId=applicationId\`.
- [ ] Tool name generated by \`LogicFunctionToolProvider\` starts with \`app_\`.

## What's NOT in this PR (PR 2 scope)

- The Exa app itself (\`packages/twenty-apps/...\` directory)
- Removing \`WebSearchTool\`, \`WebSearchService\`, \`ExaDriver\`, \`web-search\` module
- Removing \`WEB_SEARCH_DRIVER\` config var
- Removing the current \`exa_web_search\` entry in \`ActionToolProvider\`
- Chat preload list updated to \`app_exa_web_search\`
- Frontend \`getToolDisplayMessage\` branch for \`app_exa_web_search\`
- Setting \`PRE_INSTALLED_APPS\` default to include \`@twenty-apps/exa\`


packages/twenty-server/src/engine/core-modules/billing/app-billing/dtos/charge.dto.ts
```diff
@@ -0,0 +1,29 @@
...
+  @IsInt()
+  @Min(0)
+  creditsUsedMicro!: number;
+
+  @IsInt()
+  @Min(1)
+  quantity!: number;
+
+  @IsEnum(UsageUnit)
+  unit!: UsageUnit;
```
FelixMalfait: It feels like Unit and OperationType can be a bit redundant: let's analyze and make a reco / see if it's relevant or not to have both?

packages/twenty-server/src/engine/core-modules/application/app-billing/app-billing.controller.ts
```diff
@@ -0,0 +1,57 @@
...
+import { Request } from 'express';
+import { isDefined } from 'twenty-shared/utils';
+
+import { AppBillingService } from 'src/engine/core-modules/application/app-billing/app-billing.service';
+import { ChargeDto } from 'src/engine/core-modules/application/app-billing/dtos/charge.dto';
+import { JwtAuthGuard } from 'src/engine/guards/jwt-auth.guard';
+import { NoPermissionGuard } from 'src/engine/guards/no-permission.guard';
+import { WorkspaceAuthGuard } from 'src/engine/guards/workspace-auth.guard';
+
+// JwtAuthGuard validates the bearer token (including APPLICATION_ACCESS
```
FelixMalfait: Check all comments of this PR and remove ALL the comments that are useless like that. This is something pretty obvious and used all across the app!

packages/twenty-server/src/engine/core-modules/application/app-billing/app-billing.controller.ts
```diff
@@ -0,0 +1,57 @@
...
+export class AppBillingController {
+  constructor(private readonly appBillingService: AppBillingService) {}
+
+  @Post('charge')
+  @UsePipes(new ValidationPipe({ whitelist: true, forbidNonWhitelisted: true }))
+  async charge(
+    @Req() request: Request,
+    @Body() charge: ChargeDto,
+  ): Promise<{ success: true }> {
+    // JwtAuthGuard may accept a valid non-application token (user access
```
FelixMalfait: Comment not needed

packages/twenty-server/src/engine/core-modules/billing/app-billing/app-billing.service.ts
```diff
@@ -0,0 +1,50 @@
...
+import { UsageResourceType } from 'src/engine/core-modules/usage/enums/usage-resource-type.enum';
+import { type UsageEvent } from 'src/engine/core-modules/usage/types/usage-event.type';
+import { WorkspaceEventEmitter } from 'src/engine/workspace-event-emitter/workspace-event-emitter';
+
+// Receives billing events from installed apps (running as logic functions)
+// and emits USAGE_RECORDED workspace events. Apps call this through the
+// POST /app/billing/charge endpoint using their injected app access token;
+// workspaceId + applicationId come from the token, never from the body.
+@Injectable()
+export class AppBillingService {
```
FelixMalfait, line 27: I feel like this could have been a submodule into billing rather than into app? Wdyt?

packages/twenty-server/src/engine/core-modules/application/application-marketplace/marketplace.service.ts
```diff
@@ -106,6 +106,42 @@ export class MarketplaceService {
...
   }
 
+  // Resolves the latest published version of an exact npm package name
+  // directly via `GET {registryUrl}/{packageName}` (standard npm registry
+  // API: returns a document whose `dist-tags.latest` names the most recent
+  // stable release). Used by PreInstalledAppsService to resolve admin-
+  // declared packages, which must succeed regardless of where the package
+  // ranks in keyword-search results — fetchAppsFromRegistry caps at 250
+  // results and any long-tail package could otherwise be silently missed.
+  async fetchLatestVersionFromRegistry(
```
FelixMalfait: I'm surprised we have to do this and that there isn't a more straightforward path? Imo we should rather fix the root cause (improve fetchAppsFromRegistry?) rather than add more separate code paths

packages/twenty-server/src/engine/core-modules/workspace/services/workspace.service.ts
```diff
@@ -819,6 +821,20 @@ export class WorkspaceService extends TypeOrmQueryService<WorkspaceEntity> {
       );
       this.exceptionHandlerService.captureExceptions([error as Error]);
     }
+
+    // Install apps flagged `isPreInstalled` asynchronously so a slow
+    // install doesn't block workspace activation. Failures are non-critical
+    // — the admin can backfill later via the `install-pre-installed-apps`
+    // CLI command.
+    void this.preInstalledAppsService
```
FelixMalfait: Is it common to use void in the project? I don't usually see it // not convinced we should

packages/twenty-server/src/engine/core-modules/billing/app-billing/app-billing.controller.ts
```diff
@@ -0,0 +1,83 @@
...
+export class AppBillingController {
+  constructor(
+    private readonly appBillingService: AppBillingService,
+    private readonly throttlerService: ThrottlerService,
+    private readonly twentyConfigService: TwentyConfigService,
+  ) {}
+
+  @Post('charge')
+  @HttpCode(HttpStatus.NO_CONTENT)
+  @UsePipes(new ValidationPipe({ whitelist: true, forbidNonWhitelisted: true }))
```
FelixMalfait, line 44: Not familiar with this, is this common?

packages/twenty-server/src/engine/core-modules/billing/app-billing/app-billing.controller.ts
```diff
@@ -0,0 +1,83 @@
...
+  async charge(
+    @Req() request: Request,
+    @Body() charge: ChargeDto,
+  ): Promise<void> {
+    // When billing is disabled the USAGE_RECORDED listener is a no-op, so
```
FelixMalfait: You added way to many comments in this PR, keep them short

PR #19977, opened 2026-04-22
scaffold record page layout + fields view when adding an object

Weiko:
## Summary

Extends `yarn twenty add` → **Object** so it scaffolds a complete record page out
of the box:

- A **record-page-fields view** (`<name>-record-page-fields.ts`, FIELDS_WIDGET)
pre-populated with the `name` field plus the auto-generated default fields (`createdAt`,
`updatedAt`, `createdBy`, `updatedBy`) — the default-field entries are emitted as
`generateDefaultFieldUniversalIdentifier({ objectUniversalIdentifier, fieldName: '...' })`
 calls rather than pre-computed UUIDs, so the generated file double-serves as
documentation for the public util.
- A **record page layout** (`<name>-record-page-layout.ts`) with a Home tab whose Fields
widget points at the new view (via `viewUniversalIdentifier`), plus a Timeline tab.
- The companion prompt now covers all three artefacts (was view + nav menu item).

Fix: Server-side, renames `viewId` → `viewUniversalIdentifier` on the universal-flat FIELDS
widget configuration so it is consistent with other universal-flat references. The DB-side
 DTO keeps `viewId` (now typed as `SerializedRelation`), and the conversion utils map
between the two.

Screenshot 2026-04-22 at 15 40 22

### DIFF
--- a/packages/twenty-sdk/src/cli/commands/add.ts
+++ b/packages/twenty-sdk/src/cli/commands/add.ts
@@ -3,7 +3,7 @@ import inquirer from 'inquirer';
 import { writeFile } from 'node:fs/promises';
 import { join, relative } from 'path';
 import { SyncableEntity } from 'twenty-shared/application';
-import { FieldMetadataType } from 'twenty-shared/types';
+import { FieldMetadataType, ViewType } from 'twenty-shared/types';
 import { assertUnreachable } from 'twenty-shared/utils';
 import { v4 } from 'uuid';
 
@@ -15,6 +15,7 @@ import { getLogicFunctionBaseFile } from '@/cli/utilities/entity/entity-logic-fu
 import { getNavigationMenuItemBaseFile } from '@/cli/utilities/entity/entity-navigation-menu-item-template';
 import { getObjectBaseFile } from '@/cli/utilities/entity/entity-object-template';
 import { getPageLayoutBaseFile } from '@/cli/utilities/entity/entity-page-layout-template';
+import { getRecordPageLayoutBaseFile } from '@/cli/utilities/entity/entity-record-page-layout-template';
 import { getRoleBaseFile } from '@/cli/utilities/entity/entity-role-template';
 import { getAgentBaseFile } from '@/cli/utilities/entity/entity-agent-template';
 import { getSkillBaseFile } from '@/cli/utilities/entity/entity-skill-template';
@@ -27,6 +28,7 @@ const APP_FOLDER = 'src';
 export class EntityAddCommand {
   private lastObjectUniversalIdentifier: string | undefined;
   private lastNameFieldUniversalIdentifier: string | undefined;
+  private lastObjectLabelSingular: string | undefined;
 
   async execute(entityType?: SyncableEntity, path?: string): Promise<void> {
     try {
@@ -59,7 +61,7 @@ export class EntityAddCommand {
       );
 
       if (entity === SyncableEntity.Object) {
-        await this.promptAndCreateViewAndNavigationMenuItem(name, path);
+        await this.promptAndCreateObjectCompanions(name, path);
       }
     } catch (error) {
       console.error(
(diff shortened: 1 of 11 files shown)


packages/twenty-server/src/engine/metadata-modules/flat-page-layout-widget/utils/from-page-layout-widget-configuration-to-universal-configuration.util.ts
```diff
@@ -295,7 +295,7 @@ export const fromPageLayoutWidgetConfigurationToUniversalConfiguration = ({
       return {
         ...rest,
         newFieldDefaultVisibility,
-        viewId: viewUniversalIdentifier,
```
prastoin, line 298: That looks odd

ijreilly:
should we have a syntax shortcut to have them added when declaring the object directly ? personally i m using ai to declare objects so not going through yarn add. maybe i m overlooking something
@Weiko

packages/twenty-server/src/engine/metadata-modules/flat-page-layout-widget/validators/utils/validate-fields-flat-page-layout-widget-for-creation.util.ts
```diff
@@ -40,17 +40,23 @@ export const validateFieldsFlatPageLayoutWidgetForCreation = (
...
 
-  const viewId = (
-    universalConfiguration as { configurationType: string; viewId?: unknown }
-  ).viewId;
+  const viewUniversalIdentifier = (
+    universalConfiguration as {
+      configurationType: string;
+      viewUniversalIdentifier?: unknown;
+    }
+  ).viewUniversalIdentifier;
```
prastoin, line 48: Could use `PageLayoutWidgetConfiguration` here ? would be cool to have the same pattern than for fieldSettings

prastoin (approved):
I assume that until now the viewId and viewUniversalIdentifier were iso ? leading to the configuration mismatch to still work out ?
Approving though but this PR seems to also fix a bug IMO

I'm a bit afraid of deterministic universal generation derivated from the object name singular only, if two apps declare the same object they will face universal identifier conflicts

packages/twenty-sdk/src/cli/commands/add.ts
```diff
@@ -291,6 +330,75 @@ export class EntityAddCommand {
...
+      chalk.cyan(
+        relative(CURRENT_EXECUTION_DIRECTORY, recordPageLayoutFilePath),
+      ),
+    );
+  }
+
+  private buildRecordPageFieldsViewFields(
+    nameFieldUniversalIdentifier: string | undefined,
+  ) {
+    const autoGeneratedFieldNames = [
```
prastoin, line 377: Seems a bit duplicated of what we have at PARTIAL_SYSTEM_FLAT_FIELD_METADATAS in packages/twenty-server/src/engine/metadata-modules/object-metadata/constants/partial-system-flat-field-metadatas.constant.ts

Though refactoring this would be quite invasive

Weiko:
> should we have a syntax shortcut to have them added when declaring the object directly ? personally i m using ai to declare objects so not going through yarn add. maybe i m overlooking something @Weiko

We could introduce a boolean in the objectManifest such as "deferPageLayoutCreation/deferIndexViewCreation" indeed but I feel like we prefer to have manifests to be very declarative so app maintainers have full control over the entities they bring to the workspace (until we introduce 2-way sync?). 
I've asked claude to create a page layout on an object and it managed to pick up correctly based on the manifest structure though 🤔

packages/twenty-sdk/src/cli/commands/add.ts
Weiko: Interesting

Weiko:
TODO next: move the util to twenty-shared and add better typing to narrow down to the 4 default field names

PR #19980, opened 2026-04-22
feat(app): migrate Exa web search to a pre-installed app

FelixMalfait:
**PR 2 of 2.** Stacked on #19973 (the infrastructure PR).

Moves Exa from a hardcoded action tool into a standalone app at `packages/twenty-apps/community/exa`, installable via the `PRE_INSTALLED_APPS` infrastructure from PR #1. The tool surfaces to the model as `app_exa_web_search` (logic-function-sourced, hence the `app_` prefix).

## What ships

**New app package** (`packages/twenty-apps/community/exa/`):
- `application.config.ts` — declares the Exa app with one server variable (`EXA_API_KEY`, required, secret). Server admins set `EXA_API_KEY` as an env var and `PreInstalledAppsService` seeds it into the registration at bootstrap.
- `logic-functions/exa-web-search.ts` — the runtime handler. Reads `EXA_API_KEY` from its injected execution env, calls Exa via the official SDK, records usage by `POST`ing to the generic `/app/billing/charge` endpoint (using the injected `TWENTY_APP_ACCESS_TOKEN`), returns the structured results.
- `logic-functions/exa-web-search.schema.ts` — JSON-schema input contract (query, optional category, optional numResults 1–30) matching the previous tool.

## What's removed

- `packages/twenty-server/src/engine/core-modules/web-search/` — entire module: service, drivers, types, interface.
- `packages/twenty-server/src/engine/core-modules/tool/tools/web-search-tool/` — `WebSearchTool`, schema, input type.
- `WebSearchTool` / `WebSearchService` injections from `ActionToolProvider`, and the `exa_web_search` entry in its toolMap + descriptor.
- `WebSearchModule` from `CoreEngineModule` imports.
- `WEB_SEARCH_DRIVER` config variable (no longer needed — Exa is an app now).
- `ToolModule` no longer provides `WebSearchTool`.

`EXA_API_KEY` config variable stays; its description now says it seeds the Exa app's server variables. Chat preload and frontend display already use the new `app_exa_web_search` name.

## Behavior

- **Exa app installed + key set**: chat and workflow agents see `app_exa_web_search` in the catalog. Handler runs as a logic function (LOCAL in dev, LAMBDA in prod), charges credits through the generic billing endpoint.
- **Exa app not installed**: no `app_exa_web_search` tool. Chat still has native `web_search` if the model supports it.
- **Native `web_search` and `app_exa_web_search`** coexist when both are available — the model picks based on descriptions (same as current post-#19969 behavior, but under a new name).

## Deployment coordination

Server admins must:
1. Publish `@twenty-apps/exa` to the app registry (or use the LOCAL source-type flow for dev).
2. Set `PRE_INSTALLED_APPS=@twenty-apps/exa` and `EXA_API_KEY=<key>` as env vars.
3. On new workspaces: the app auto-installs during activation.
4. On existing workspaces: run the `install-pre-installed-apps` CLI command added in PR #1.

## Stats

- 28 files, +308 / −412 (net removal — the deleted server-side code is larger than the new app).
- Typecheck: 7 pre-existing errors, zero new.
- Prettier clean.

## Test plan

- [ ] \`npx nx typecheck twenty-server\` passes
- [ ] \`npx nx typecheck twenty-front\` passes
- [ ] Build the Exa app via the Twenty CLI: manifest emits with \`isTool: true\` and \`toolInputSchema\`, handler compiles.
- [ ] Dev workspace with Exa installed: chat sees \`app_exa_web_search\`, calling it returns structured Exa results, USAGE_RECORDED event emitted with \`resourceType: APP\`, \`operationType: WEB_SEARCH\`.
- [ ] Workspace without Exa: no \`app_exa_web_search\` in catalog; native \`web_search\` (Anthropic) still works.
- [ ] Frontend: tool call renders as "Searching the web for ..." (same display as native).


packages/twenty-apps/internal/exa/tsconfig.json
```diff
@@ -0,0 +1,30 @@
+{
```
FelixMalfait, line 1: This was put in community module but I think we have a directory for internal apps no? Move it there if we do

packages/twenty-apps/community/exa/src/logic-functions/exa-web-search.ts
```diff
@@ -0,0 +1,174 @@
...
+  const apiUrl = env[DEFAULT_API_URL_NAME];
+  const token = env[DEFAULT_APP_ACCESS_TOKEN_NAME];
+
+  if (!apiUrl || !token) {
+    return;
+  }
+
+  try {
+    const response = await fetch(
+      `${apiUrl.replace(/\/$/, '')}/app/billing/charge`,
```
FelixMalfait: Is this really the clean way to make API call? Can't we use the twenty-sdk / do something cleaner?!

packages/twenty-apps/community/exa/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@twenty-apps/exa",
```
FelixMalfait: is this the right naming structure for our npm packages? I'm not sure

packages/twenty-apps/internal/exa/src/application.config.ts
```diff
@@ -0,0 +1,20 @@
+import { defineApplication } from 'twenty-sdk/define';
+
+export const APPLICATION_UNIVERSAL_IDENTIFIER =
+  '2b7f4a2e-9c4b-4a11-b63c-2e5e7d3f5a9a';
+
+export default defineApplication({
```
FelixMalfait, line 8: Does every application need to provision a role? It's great if we don't have too but I wonder if logic function work in that case? If they do then great!

packages/twenty-apps/community/exa/src/logic-functions/exa-web-search.schema.ts
```diff
@@ -0,0 +1,49 @@
+// JSON Schema describing the input accepted by the exa_web_search tool.
+// Kept as a plain JSON Schema (not Zod) because this is shipped through the
+// application manifest and consumed by Twenty's tool provider at runtime.
+
+export const DEFAULT_NUM_RESULTS = 10;
```
FelixMalfait: Let's try to follow the project's convention which is one export = one file ;
Maybe not everything needs to be exported / put into a separate file

PR #19981, opened 2026-04-22
Cleanup application and app registration test util

prastoin:
## Introduction
Centralizing integ test app cleanup
Role will be deleted by the app uninstall if exists


Weiko (approved):
LGTM

PR #19993, opened 2026-04-22
feat(migration): seed Exa pre-install from EXA_API_KEY env var

FelixMalfait:
## Summary

One-shot bridge for instances that set `EXA_API_KEY` as an env var under the old native `web-search` driver (removed in #19980). Prevents a gap between deploy and admin setup where Exa search is broken.

**Stacked on top of #19980** — depends on the `isPreInstalled` column + the `twenty-exa` npm package.

## What it does

A **slow instance command** that runs at deploy time:

1. Skips if `EXA_API_KEY` is unset → no-op on fresh instances.
2. Fetches the `twenty-exa` manifest from the app registry CDN and upserts the `ApplicationRegistration` via the existing catalog-sync path.
3. Seeds `EXA_API_KEY` onto the registration's server variable, encrypted via `SecretEncryptionService`. Never overwrites a value already edited through the admin UI.
4. Flips `isPreInstalled=true`. New workspaces auto-install from that point; existing workspaces are backfilled with the `install-pre-installed-apps` CLI command.

Idempotent — each step no-ops when its target already exists. Safe to rerun.

## Prerequisites

Before merging / deploying:

1. Publish `twenty-sdk@2.1.0` (new `./billing` subpath, introduced in #19980).
2. Publish `twenty-exa@0.1.0` to npm.

Otherwise the migration's CDN fetch returns no manifest and the command logs a warning and exits clean (no crash).

## Post-deploy

- Drop `EXA_API_KEY` from infra env vars — it now lives encrypted on the registration.
- Run `npx nx run twenty-server:command install-pre-installed-apps` to backfill existing workspaces (new workspaces pick it up automatically).

## Why a separate PR

Keeps #19980 focused on the architectural migration. This one is a deployment-lifecycle helper — reviewable in isolation and droppable without losing the core work if we change our mind on the bridging strategy.

## Test plan

- [ ] Unit tests for the command
- [ ] Verify server boots with new command registered (✅ `Registered 1 fast instance, 1 slow instance ... for 2.0.0`)
- [ ] Verify idempotency: run migration twice against a seeded instance
- [ ] Verify no-op when env var absent
- [ ] Verify no-op when registration already exists with isPreInstalled=true


prastoin:
@claude could you please review this PR ?

charlesBochet:
@FelixMalfait Closing this one

PR #20000, opened 2026-04-23
optimize workspace export command

neo773:
- Use `COPY` instead of `INSERT` for workspace tables, 40x faster imports
- Multi value statements, 2x smaller file size
- Bump Batch size to 10K , remove COUNT query
- Add `formatPgCopyField` utility with unit tests


packages/twenty-server/src/database/commands/workspace-export/workspace-export.service.ts
```diff
@@ -219,47 +221,103 @@ export class WorkspaceExportService {
...
+    let totalRows = 0;
+
+    for (let offset = 0; ; offset += BATCH_SIZE) {
+      const rows: Record<string, unknown>[] = await queryRunner.query(
+        `SELECT * FROM "${schemaName}"."${tableName}" ORDER BY "id" LIMIT ${BATCH_SIZE} OFFSET ${offset}`,
+      );
+
+      if (rows.length === 0) break;
+
+      if (!columnNames) {
```
charlesBochet, line 287: use isEmptyArray

packages/twenty-server/src/database/commands/workspace-export/workspace-export.service.ts
```diff
@@ -219,47 +221,103 @@ export class WorkspaceExportService {
...
+
+        if (!stream.write(values.join('\t') + '\n')) {
+          await once(stream, 'drain');
+        }
+      }
+
+      if (rows.length < BATCH_SIZE) break;
+    }
+
+    if (columnNames) {
```
charlesBochet: same here

charlesBochet (approved):

PR #20022, opened 2026-04-24
Clarify registry tools vs native model tool binding

ehconitin:
## Intent

This is a small foundation cleanup for the tool architecture.

The main decision is: registry tools and native SDK/model tools are different things.

- Registry tools have descriptors, schemas, catalog entries, and execute through `ToolExecutorService`
- Native model tools are opaque AI SDK objects, bound directly into the model `ToolSet`
- Surfaces still own their policy: chat, MCP, and workflow agents decide what they expose

## What changed

- Removed `NATIVE_MODEL` from `ToolCategory`
- Kept `ToolRegistryService` focused on registry-backed tools only
- Moved native model tool binding through `NativeToolBinderService`
- Reused native binding from chat instead of duplicating provider-specific web-search logic
- Kept MCP local execution exclusions in a dedicated constant
- Moved surface-specific constants into dedicated constant files

## What comes next

- Move hardcoded chat app preloads, like Exa web search, into app/manifest metadata
- Decide a clearer policy for local runtime tools like code interpreter and HTTP request
- Gradually document the three tool shapes: registry tools, native model tools, and local runtime tools


packages/twenty-server/src/engine/metadata-modules/ai/ai-chat/constants/ai-chat-tool-names-to-preload.const.ts
```diff
@@ -0,0 +1,6 @@
+import { COMMON_PRELOAD_TOOLS } from 'src/engine/core-modules/tool-provider/constants/common-preload-tools.const';
+
+export const AI_CHAT_TOOL_NAMES_TO_PRELOAD: string[] = [
+  ...COMMON_PRELOAD_TOOLS,
+  'app_exa_web_search',
```
FelixMalfait, line 5: (nit): // TODO: remove EXA could be kept 

 shouldn't be hardcoded like that

packages/twenty-server/src/engine/metadata-modules/ai/ai-models/services/ai-model-config.service.ts
```diff
@@ -42,57 +43,63 @@ export class AiModelConfigService {
...
-            tools.web_search =
-              bedrockProvider.tools.webSearch_20250305() as ToolSet[string];
-          }
+        const bedrockProvider = model.providerName
+          ? this.sdkProviderFactory.getRawBedrockProvider(model.providerName)
+          : undefined;
+
+        if (bedrockProvider) {
+          tools.web_search =
+            bedrockProvider.tools.webSearch_20250305() as ToolSet[string];
```
FelixMalfait: I'm surprised we had websearch for bedrock, don't thinks it works! @claude remove it and push to this branch

packages/twenty-server/src/engine/metadata-modules/ai/ai-chat/constants/ai-chat-tool-names-to-preload.const.ts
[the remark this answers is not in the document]
FelixMalfait: @claude is app_exa_web_search automatically excluded if it's not available? Or is there a risk of this being advertised as available while the exa app isn't installed?

FelixMalfait (approved):

PR #20025, opened 2026-04-24
feat(twenty-sdk): stub twenty-sdk/define in app bundles to drop zod + locales

charlesBochet:
## Problem

The twenty CLI's esbuild only externalizes \`twenty-client-sdk/core\` and \`twenty-client-sdk/metadata\` when bundling user apps. Everything else is inlined — including \`twenty-sdk/define\`, which statically imports \`zod\` (entry point re-exports every locale — ~80 files) and \`@sniptt/guards\`.

Result on twenty-eng (a real twenty app): a 30-line \`defineLogicFunction\` handler bundles to **1.26 MB** of \`.mjs\` + **3.3 MB** of \`.map\`. Every logic function carries the full zod + @sniptt payload.

The twenty runtime only reads \`default.config.handler\` (see \`packages/twenty-server/src/engine/metadata-modules/logic-function/constants/handler.contant.ts\`). \`defineXxx\` helpers only run build-time validation, and build-time manifest extraction already imports the real module directly in \`manifest-extract-config-from-file.ts\` (unchanged here) — so at app runtime, all the validation machinery + zod + sniptt is dead weight.

## Fix

- New zero-dependency runtime stub at \`packages/twenty-sdk/src/sdk/define/runtime-stub.ts\`. It passes \`config\` through untouched from every \`defineXxx\`, inlines the enum constants (values only, no \`twenty-shared/types\` barrel — that barrel re-exports \`richTextValueSchema\` which is what drags zod in), and re-exports \`STANDARD_OBJECTS\` from the zod-free \`twenty-shared/metadata\` subpath. The real runtime helpers (\`getPublicAssetUrl\`, \`generateDefaultFieldUniversalIdentifier\`) are preserved.
- New vite entry \`vite.config.define-stub.ts\` → \`dist/define-runtime-stub/index.{mjs,cjs}\`; wired into \`project.json\`'s \`build\` and \`build:sdk\` targets.
- New \`./define-runtime-stub\` subpath in \`package.json\`; \`types\` point at the existing \`./define\` declarations so consumers that somehow import the stub directly still get the full typed surface.
- CLI esbuild in \`build-application.ts\` and \`esbuild-watcher.ts\` now passes \`alias: { 'twenty-sdk/define': 'twenty-sdk/define-runtime-stub' }\` for both logic-function and front-component builds. Manifest extraction (\`manifest-extract-config-from-file.ts\`) is untouched, so build-time validation still runs.

## Impact (twenty-eng, 31 logic functions)

| | before | after | |
|---|---|---|---|
| largest logic-function \`.mjs\` | 1.26 MB | **332 KB** | −75% |
| largest \`.mjs.map\` | 3.3 MB | **744 KB** | −77% |
| total logic-function \`.mjs\` | ~39 MB | **~14.7 MB** | −62% |

Front components (~2 MB) still carry the same shape of bloat via \`twenty-sdk/front-component\`. That surface is different — its helpers have real runtime behavior that the host injects — so a blanket stub doesn't apply. Follow-up.

## Test plan

- [x] \`nx run twenty-sdk:build\` — full sdk build succeeds
- [x] \`nx run twenty-sdk:test:unit\` — 262/262 pass
- [x] \`nx run twenty-sdk:test:integration\` — 27/27 pass (rich-app integration builds real logic-function + front-component outputs)
- [x] Rebuilt SDK linked into twenty-eng, \`twenty build\` succeeds, handler wiring preserved (\`default.config.handler\` resolves to user handler via the passthrough)
- [ ] CI

### DIFF
--- a/packages/twenty-sdk/src/sdk/define/runtime-stub.ts
+++ b/packages/twenty-sdk/src/sdk/define/runtime-stub.ts
@@ -1,0 +1,259 @@
+/**
+ * Runtime stub for `twenty-sdk/define`.
+ *
+ * `twenty-sdk/define` is an authoring API: its `defineXxx` helpers only run
+ * build-time validation, and the full module transitively pulls in `zod`
+ * (including every locale) plus `@sniptt/guards`. In a built logic function
+ * or front component that only needs `default.config.handler` /
+ * `default.config.component` at runtime, those extras inflate bundles by
+ * ~1MB per file.
+ *
+ * The CLI aliases `twenty-sdk/define` to this stub when bundling user code,
+ * so the published SDK continues to do build-time validation while user
+ * bundles ship only the identity passthroughs and the handful of enum
+ * constants they actually reference. Tree-shaking removes the rest.
+ *
+ * Keep this file free of imports that transitively reach `zod` — i.e. do not
+ * re-export from the `twenty-shared/types` barrel, which pulls
+ * `richTextValueSchema` (zod) in along for the ride.
+ */
+
+export { STANDARD_OBJECTS as STANDARD_OBJECT } from 'twenty-shared/metadata';
+export { STANDARD_OBJECTS as STANDARD_OBJECT_UNIVERSAL_IDENTIFIERS } from 'twenty-shared/metadata';
+
+export { generateDefaultFieldUniversalIdentifier } from '@/sdk/define/objects/generate-default-field-universal-identifier';
+export { getPublicAssetUrl } from '@/sdk/define/get-public-asset-url';
+
+type PassthroughResult<T> = {
+  success: true;
+  config: T;
+  errors: readonly never[];
+};
+
+const passthrough = <T>(config: T): PassthroughResult<T> => ({
+  success: true,
+  config,
+  errors: [],
+});
+
+export const defineAgent = passthrough;
+export const defineApplication = passthrough;
+export const defineField = passthrough;
+export const defineFrontComponent = passthrough;
+export const defineLogicFunction = passthrough;
+export const defineNavigationMenuItem = passthrough;
+export const defineObject = passthrough;
+export const definePageLayout = passthrough;
+export const definePageLayoutTab = passthrough;
+export const definePostInstallLogicFunction = passthrough;
+export const definePreInstallLogicFunction = passthrough;
+export const defineRole = passthrough;
+export const defineSkill = passthrough;
+export const defineView = passthrough;
+
+export const createValidationResult = <T>({
+  config,
+  errors = [],
+}: {
+  config: T;
+  errors?: string[];
+}) => ({
+  success: errors.length === 0,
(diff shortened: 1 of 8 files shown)


packages/twenty-sdk/src/sdk/define/runtime-stub.ts
```diff
@@ -0,0 +1,259 @@
...
+ * bundles ship only the identity passthroughs and the handful of enum
+ * constants they actually reference. Tree-shaking removes the rest.
+ *
+ * Keep this file free of imports that transitively reach `zod` — i.e. do not
+ * re-export from the `twenty-shared/types` barrel, which pulls
+ * `richTextValueSchema` (zod) in along for the ride.
+ */
+
+export { STANDARD_OBJECTS as STANDARD_OBJECT } from 'twenty-shared/metadata';
+export { STANDARD_OBJECTS as STANDARD_OBJECT_UNIVERSAL_IDENTIFIERS } from 'twenty-shared/metadata';
```
Weiko, line 22: Why do we re-export the same STANDARD_OBJECTS value under two different names, am I missing something?

packages/twenty-sdk/src/sdk/define/runtime-stub.ts
```diff
@@ -0,0 +1,259 @@
...
+  success: errors.length === 0,
+  config,
+  errors,
+});
+
+export const validateFields = (): string[] => [];
+
+// ─── Enum constants ─────────────────────────────────────────────────────────
+// Values inlined (rather than re-exported from `twenty-shared/types`) to keep
+// this stub free of the zod-using `richTextValueSchema` re-export that lives
```
Weiko, line 70: What do we do in case of out of sync enums? How do we prevent that? Should we add a test that imports both and asserts equality? Or a CI?

packages/twenty-sdk/src/sdk/define/runtime-stub.ts
```diff
@@ -0,0 +1,259 @@
...
+ *
+ * Keep this file free of imports that transitively reach `zod` — i.e. do not
+ * re-export from the `twenty-shared/types` barrel, which pulls
+ * `richTextValueSchema` (zod) in along for the ride.
+ */
+
+export { STANDARD_OBJECTS as STANDARD_OBJECT } from 'twenty-shared/metadata';
+export { STANDARD_OBJECTS as STANDARD_OBJECT_UNIVERSAL_IDENTIFIERS } from 'twenty-shared/metadata';
+
+export { generateDefaultFieldUniversalIdentifier } from '@/sdk/define/objects/generate-default-field-universal-identifier';
```
Weiko, line 24: I get the base idea but we don't prevent those utils to re-import zod themselves, right?

Weiko (approved):
LGTM, left some comments and wondering if we can find a more robust solution, happy to discuss 🤔

PR #20032, opened 2026-04-24
backfill widget position from gridPosition (phase 1 of gridPosition removal)

Weiko:
## Context
Phase 1 of removing the legacy `gridPosition` field from `PageLayoutWidget` in favor of the new `position` discriminated union (`grid` / `vertical-list` / `canvas`). This PR is purely additive — `gridPosition` is still required and read everywhere; we just guarantee that every widget now also has a non-null `position` so a follow-up PR can drop `gridPosition` cleanly.

## Changes
- **Slow instance command** `BackfillPageLayoutWidgetPositionSlowInstanceCommand` (2.1.0):
 for every `core.pageLayoutWidget` row where `position IS NULL`, copies `gridPosition`
into `position` with `layoutMode: 'GRID'`. Historically only grid widgets used
`gridPosition`, so a single SQL update covers every existing row.
- **`PageLayoutDuplicationService`**: when duplicating a widget, also forwards
`originalWidget.position` (was previously only forwarding `gridPosition`).
- **Frontend default layouts** (10 `Default*PageLayout.ts` files): added a `position`
sibling to every widget, matching the parent tab's `layoutMode` — `VERTICAL_LIST` widgets
get `{ layoutMode, index }`, `CANVAS` widgets get `{ layoutMode }`

Screenshot 2026-04-24 at 16 23 17

### DIFF
--- a/packages/twenty-server/src/database/commands/upgrade-version-command/2-1/2-1-instance-command-slow-1795000002000-backfill-page-layout-widget-position.ts
+++ b/packages/twenty-server/src/database/commands/upgrade-version-command/2-1/2-1-instance-command-slow-1795000002000-backfill-page-layout-widget-position.ts
@@ -1,0 +1,32 @@
+import { DataSource, QueryRunner } from 'typeorm';
+
+import { RegisteredInstanceCommand } from 'src/engine/core-modules/upgrade/decorators/registered-instance-command.decorator';
+import { SlowInstanceCommand } from 'src/engine/core-modules/upgrade/interfaces/slow-instance-command.interface';
+
+@RegisteredInstanceCommand('2.1.0', 1795000002000, { type: 'slow' })
+export class BackfillPageLayoutWidgetPositionSlowInstanceCommand
+  implements SlowInstanceCommand
+{
+  async runDataMigration(dataSource: DataSource): Promise<void> {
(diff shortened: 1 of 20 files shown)


charlesBochet (approved):

packages/twenty-server/src/database/commands/upgrade-version-command/2-1/2-1-instance-command-slow-1795000002000-backfill-page-layout-widget-position.ts
```diff
@@ -0,0 +1,32 @@
...
+          SET "position" = jsonb_build_object(
+            'layoutMode',  'GRID',
+            'row',         "gridPosition"->'row',
+            'column',      "gridPosition"->'column',
+            'rowSpan',     "gridPosition"->'rowSpan',
+            'columnSpan',  "gridPosition"->'columnSpan'
+          )
+        WHERE "position" IS NULL
+          AND "gridPosition" IS NOT NULL`,
+    );
```
prastoin, line 22: ***Remark:*** this has literally no impact here, we're all good. Just noting that when a instance slow is run if not workspace are found the `runDataMigration` is skipped

prastoin (approved):

PR #20048, opened 2026-04-25
[AI] ai usage line chart date gap filling

ehconitin:
https://github.com/user-attachments/assets/ecd37dfb-daae-41c3-8316-a9d923463eca


charlesBochet (approved):

FelixMalfait:
Great! Just checking: did this fix all affected surfaces(I think there was more than one chart)? I assume it did (in which case no need to reply to this comment), but just checking

PR #20049, opened 2026-04-25
Fix default workspace resolution in single-workspace mode

Adarshkumar167:
## Summary

In single-workspace mode, `getDefaultWorkspace()` sorted workspaces by `createdAt DESC` and picked the first result. After a full dev seed (which creates 4 workspaces), this resolved to Empty4 — the most recently created workspace. The default login user `tim@apple.dev` is only a member of Apple and YCombinator, so login failed with "User is not a member of the workspace."

## Fix

Changed the sort order from `createdAt DESC` to `createdAt ASC` so the oldest workspace (Apple, seeded first) is selected as the default. Updated the warning message to reflect the actual fallback behavior.

## Testing

- Updated existing test for multiple workspace fallback behavior
- All 11 tests pass

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/domain/workspace-domains/services/__test__/workspace-domains.service.spec.ts
+++ b/packages/twenty-server/src/engine/core-modules/domain/workspace-domains/services/__test__/workspace-domains.service.spec.ts
@@ -213,7 +213,7 @@ describe('WorkspaceDomainsService', () => {
       expect(result?.id).toEqual('workspace-id');
     });
 
-    it('should return 1st workspace if multiple workspaces when IS_MULTIWORKSPACE_ENABLED=false', async () => {
+    it('should return first workspace if multiple workspaces exist when IS_MULTIWORKSPACE_ENABLED=false', async () => {
       jest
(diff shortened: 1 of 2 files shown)


charlesBochet:
@Adarshkumar167 How did you face this issue? are you trying to self-host (in this case seeds should not be there) or to contribute? (in which case the seed command should be --light and you should not face this issue

It does not seem a real issue to me so I'll close it but interested in how you faced the issue that led you to raise this PR

Adarshkumar167:
@charlesBochet 
Thanks for the context! To answer your question, I was setting up a local dev environment to contribute.
I followed the documented setup which uses npx nx database:reset twenty-server. That command defaults to the full seed (defaultConfiguration: "seed" in project.json), which creates all 4 workspaces. Logging in with tim@apple.dev then failed because getDefaultWorkspace() picked the wrong one.
I wasn't aware of the --light flag, it's used internally in init-db.sh for the Docker dev container, but isn't mentioned in any contributor-facing docs.

Screenshot 2026-04-26 at 8 00 29 PM


Would you be open to a small docs PR that mentions --light for local dev setup? That way new contributors don't hit the same confusion.

PR #20053, opened 2026-04-26
add UpsertViewWidget resolver

abdulrahmancodes:


packages/twenty-server/src/engine/metadata-modules/view/resolvers/view.resolver.ts
```diff
@@ -197,6 +202,19 @@ export class ViewResolver {
     return isDefined(deletedView);
   }
 
+  @Mutation(() => ViewDTO)
+  @UseGuards(NoPermissionGuard)
+  @UsePipes(ResolverValidationPipe)
+  async upsertViewWidget(
```
etiennejouan, line 210: Shouldn't we guard this resolver behind   `@UseGuards(SettingsPermissionGuard(PermissionFlagType.LAYOUTS))` ?

etiennejouan (commented):
BE code looks very good. Would be perfect to add integration testing with create/update scenarii on all fields, groups.

Concerning FE, it doesn't work after testing. Try to add a widget, save & refresh table

etiennejouan (commented):
Integration tests look very good ! 👏 


Concerning feature testing, I've tried to update a view widget via UI without success.

packages/twenty-front/src/modules/page-layout/widgets/record-table/utils/getRecordTableWidgetDraftViewFieldClientId.ts
```diff
@@ -0,0 +1,21 @@
...
+): string => {
+  if (isDefined(field.id)) {
+    return field.id;
+  }
+
+  if (isDefined(field.clientRecordFieldId)) {
+    return field.clientRecordFieldId;
+  }
+
+  return '';
```
etiennejouan, line 20: nit : in which case it can happen ?

packages/twenty-front/src/modules/page-layout/widgets/record-table/hooks/useRecordTableWidgetFilterCallbacks.ts
```diff
@@ -0,0 +1,83 @@
...
+            subFieldName: recordFilter.subFieldName ?? null,
+          })),
+          viewFilterGroups: currentRecordFilterGroups.map(
+            (recordFilterGroup) => ({
+              id: recordFilterGroup.id,
+              viewId,
+              parentViewFilterGroupId:
+                recordFilterGroup.parentRecordFilterGroupId ?? null,
+              logicalOperator:
+                recordFilterGroup.logicalOperator as unknown as ViewFilterGroupLogicalOperator,
```
etiennejouan: nit : use a mapper util to pass from RecordFilterGroup to ViewFilterGroup

etiennejouan (approved):
Very nice work 👏

packages/twenty-front/src/modules/page-layout/widgets/record-table/utils/getRecordTableWidgetDraftViewFieldClientId.ts
abdulrahmancodes: it shouldn’t happen in normal use. Persisted fields have `id`, newly added ones get `clientRecordFieldId`. The type just allows both to be optional, so this is a defensive fallback instead of throwing

PR #20054, opened 2026-04-26
fix(logic-function): serialize LocalDriver layer builds with cache lock

charlesBochet:
## Summary

Fixes a race condition in `LocalDriver` where concurrent `execute()` calls for the same application can trash each other's layer build, causing logic function executions to fail with either:

- `Error: ENOENT: process.cwd failed with error no such file or directory, the current working directory was likely removed without changing the working directory, uv_cwd` (the yarn-install child's `cwd` got wiped mid-run), or
- `ENOENT: no such file or directory, open '…/<pkg>/<file>.js.map'` raised from yarn's berry link step (files under the deps layer vanish while yarn is extracting).

Both are thrown from `…/deps/<checksum>/.yarn/releases/yarn-4.9.2.cjs` — i.e. the yarn process `copyYarnEngineAndBuildDependencies` spawns with `cwd: buildDirectory`.

## Root cause

`LocalDriver.createLayerIfNotExist` (and `ensureSdkLayer`) both follow a check-then-act pattern with no mutual exclusion:

```ts
if (await pathExists(depsNodeModulesPath)) return;
await fs.rm(depsLayerPath, { recursive: true, force: true });
await copyDependenciesInMemory(...);
await copyYarnEngineAndBuildDependencies(depsLayerPath); // spawns yarn with cwd=depsLayerPath
```

The deps layer path is shared across all `execute()` calls that match a given `yarnLockChecksum`. Two concurrent callers (e.g. a webhook invocation + a cron-triggered logic function firing while the first run's layer is still being built) both see no `node_modules`, both `fs.rm` the directory, and one's yarn child ends up with its `cwd` or its extraction target gone.

## Fix

Wrap the critical sections with `CacheLockService.withLock` — the same cache-backed lock the Lambda driver already uses for its layer/executor builds:

- `createLayerIfNotExist` → lock key `local-driver-deps-layer:${yarnLockChecksum ?? 'default'}`
- `ensureSdkLayer` → lock key `local-driver-sdk-layer:${workspaceId}:${applicationUniversalIdentifier}`

A fast-path `pathExists` check is kept outside the lock (so warm executions still skip lock acquisition entirely), and the existence + staleness check is **repeated inside the lock** so followers no-op after the leader finishes.

Lock parameters mirror the Lambda driver's layer-build lock (`ttl=120s`, `retry=500ms`, `maxRetries=240`).

`cacheLockService` is now passed to `LocalDriver` from `LogicFunctionDriverFactory` (it was already injected there for the Lambda driver).

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/local.driver.ts
+++ b/packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/local.driver.ts
@@ -24,15 +24,10 @@ import { copyYarnEngineAndBuildDependencies } from 'src/engine/core-modules/appl
 import type { LogicFunctionResourceService } from 'src/engine/core-modules/logic-function/logic-function-resource/logic-function-resource.service';
 import type { SdkClientArchiveService } from 'src/engine/core-modules/sdk-client/sdk-client-archive.service';
 
-// Concurrent execute() calls for the same application would otherwise race
-// in `createLayerIfNotExist` / `ensureSdkLayer`: both would observe the layer
-// as missing, both would `fs.rm` and repopulate the same directory, and yarn
-// install running in one call would see its cwd wiped by the other. Use the
-// shared cache-backed lock (same mechanism Lambda driver uses) to serialize
-// layer builds across processes.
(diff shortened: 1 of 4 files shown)


Weiko (approved):
LGTM

PR #20056, opened 2026-04-26
Redesign application content tab + logic function settings; add Layout detail pages

FelixMalfait:
## Summary

Iterative redesign of two related areas in settings, plus a new `pages/settings/layout/` folder for read-only entity detail pages.

### Application content tab

- **Grouped into three sections** — Data / Layout / Logic — each with one H2 + multiple `TableSection`-wrapped sub-tables (mirrors the role-permissions pattern). Replaces six per-category table/row components with one uniform `<SettingsApplicationContentSubtable>` + `ApplicationContentRow` shape (net **−~700 lines** across the refactor).
- **All 10 row categories now clickable** for installed apps:
  - Objects / Fields / Logic functions / Front components → existing detail pages
  - Agents → existing `AiAgentDetail`
  - Skills → existing `AiSkillDetail` (looked up by `Skill.applicationId + name`)
  - Roles → existing `RoleDetail` (looked up by `Role.universalIdentifier`)
  - Views / Page layouts / Navigation menu items → **new** detail pages (see below)
- **Lifecycle hooks visible** — `pre-install` / `post-install` logic functions are surfaced in the Trigger column instead of appearing as empty/misconfigured.

### Logic function settings (Triggers + Test tabs)

- Triggers tab is now editable (HTTP / Cron / Database event / AI tool) with a `<SettingsLogicFunctionTriggerSection>` wrapper that owns the toggle, header, and read-only short-circuit.
- HTTP section gets a Live URL field with copy-to-clipboard.
- Each section shows a **Sample input** preview (the JSON the function will receive) using the same payload builders the Test tab uses.
- Test tab: **Simulate trigger** buttons that prefill the JSON input from the configured trigger's schema. Replaces an unclickable `<Select>` (which auto-disables when there's only one option — the typical case).
- Read-only behavior for installed-app functions: explicit `<Callout>` notice when there's no trigger; trigger sections render as disabled controls when there is one.
- Removed the empty Environment Variables section from the Settings tab (it just told the user to go elsewhere).

### New `pages/settings/layout/` folder

Three new app-scoped detail pages so users can drill into entities the GraphQL `Application` type doesn't expose by id (keyed by manifest `universalIdentifier`):

- `ApplicationViewDetail` — type, object, visibility + Fields / Filters / Sorts subsections (field UIDs resolved to readable labels via `useFieldLabelByUid`)
- `ApplicationPageLayoutDetail` — type, object + per-tab subsections listing widgets
- `ApplicationNavigationMenuItemDetail` — type, destination (resolved), icon, color, position

Each page reads from the marketplace manifest the parent app page already loads (no extra queries). Folder set up so a future "Layout" settings tab can grow here (analogous to the existing `data-model/` folder under the Data tab).

### Other consistency fixes

- Breadcrumbs on every app-scoped entity detail page now include a category crumb so users know what they're looking at: `Workspace / Applications / Timely / Navigation menu items / Time entry`.
- Title fallback for nav menu items uses the resolved destination (`"Time entry"`) instead of the raw enum (`"OBJECT"`).
- New shared utils: `getNavigationMenuItemDestination`, `resolveManifestObjectLabel`, `getLogicFunctionTriggerLabel`, `<MonoText>`.

## Backend changes

Only one minor schema-shape change (additive): added `applicationId` to the `SkillFields` GraphQL fragment and `universalIdentifier` to the `RoleFragment` so the new lookups have what they need. Generated metadata schema patched in-tree to match — regenerate with `nx run twenty-front:graphql:generate --configuration=metadata` if it drifts.

## Test plan

- [ ] Application content tab on an installed app shows the 3 grouped sections; rows in each section are clickable
- [ ] Click an Object → existing object detail page
- [ ] Click a Field → existing field-edit page
- [ ] Click an Agent / Skill / Role → existing detail page
- [ ] Click a View / Page layout / Navigation menu item → new read-only detail page; subsections (Fields/Filters/Sorts for views, per-tab widgets for page layouts) populate correctly
- [ ] Breadcrumbs on every entity detail page have 5 crumbs ending in `<Category> / <Entity name>`
- [ ] Logic function Triggers tab: toggle each trigger type on/off, see the Sample input preview update; for installed apps, sections render as read-only
- [ ] Test tab: each "Simulate trigger" button prefills the JSON editor with the matching payload shape
- [ ] Functions list: a function configured as `post-install` shows "Post-install" in the Trigger column


packages/twenty-front/src/modules/logic-functions/utils/getLogicFunctionTriggerLabel.ts
```diff
@@ -0,0 +1,42 @@
...
+
+type LogicFunctionLike = {
+  universalIdentifier?: string | null;
+  isTool?: boolean;
+  cronTriggerSettings?: unknown;
+  httpRouteTriggerSettings?: unknown;
+  databaseEventTriggerSettings?: { eventName?: string } | null;
+};
+
+// Lifecycle hook universalIdentifiers come from the application manifest
```
charlesBochet: I don'tg et this comment. how a frontend util would impact install/upgrade flow

packages/twenty-front/src/modules/settings/applications/hooks/useApplicationContentSections.ts
```diff
@@ -0,0 +1,239 @@
...
+export const useApplicationContentSections = ({
+  installedApplication,
+  manifestContent,
+}: {
+  installedApplication?: InstalledApplicationForContentSections;
+  manifestContent?: Manifest;
+}) => {
+  const objectMetadataItems = useAtomStateValue(objectMetadataItemsSelector);
+
+  // Skills + roles are loaded so manifest entries can be linked to their
```
charlesBochet: I don't get it

packages/twenty-front/src/modules/settings/applications/hooks/useApplicationContentSections.ts
```diff
@@ -0,0 +1,239 @@
...
+  getNavigationMenuItemDisplayLabel,
+} from '~/pages/settings/layout/utils/getNavigationMenuItemDestination';
+import { resolveManifestObjectLabel } from '~/pages/settings/layout/utils/resolveManifestObjectLabel';
+
+type InstalledApplicationForContentSections = Pick<
+  Application,
+  'agents' | 'id'
+>;
+
+export const useApplicationContentSections = ({
```
charlesBochet: I recommend to avoid this kind of hook which is fetching a lot. Use a ComponentEffect instead write to states, you'll get more control. When using a componentEffect, you can gate it with if (installedApplication.id) for instance and avoid skiping logic which will lead to race conditions

charlesBochet: Also, regarding devXP, people will start re-using this umbrealla to get one piece of information without understanding that this is making that many queries

packages/twenty-front/src/modules/settings/applications/hooks/useApplicationContentSections.ts
```diff
@@ -0,0 +1,239 @@
...
+                  },
+                )
+              : undefined,
+          };
+        })
+        .filter((row) => isDefined(row.name)),
+    [manifestContent, objectMetadataItems, installedAppId],
+  );
+
+  const agentRows = useMemo((): ApplicationContentRow[] => {
```
charlesBochet: then you don't need useMemo anymore if using ComponentEffect because you stop adding a lot of dependencies to your react tree

packages/twenty-front/src/modules/settings/applications/hooks/useObjectAndFieldRows.ts
```diff
@@ -16,153 +17,127 @@ type InstalledApplicationForObjectRows = Omit<
   objects: { id: string }[];
 };
 
+// Framework plumbing that always carries app-contributed fields but isn't
+// useful to show as something the app extends.
+const FIELD_GROUP_DENY_LIST = new Set(['timelineActivity', 'favorite']);
+
 export const useObjectAndFieldRows = ({
```
charlesBochet: hook naming is not great, useComputeObjectAndFieldsContentForApplication
A bit "spaghetti" code as this seems really tied to the presentation layer (we are mapping all output in a specific format for presentation) but this hooks also does filtering. It's fine if this hook is not re-used out of it's context that's why naming should be super clear

packages/twenty-front/src/modules/settings/logic-functions/components/tabs/SettingsLogicFunctionTestTab.tsx
```diff
@@ -35,6 +72,43 @@ export const SettingsLogicFunctionTestTab = ({
...
 
+  const {
+    httpRouteTriggerSettings,
+    cronTriggerSettings,
+    databaseEventTriggerSettings,
+    toolInputSchema,
+    isTool,
+  } = formValues;
+
+  const triggers: SimulatedTrigger[] = useMemo(() => {
```
charlesBochet: nit: useMemo probably not useful here, let react do it his way

charlesBochet: what does "Simulated" naming goe for?

packages/twenty-front/src/modules/settings/logic-functions/components/triggers/SettingsLogicFunctionDatabaseEventTriggerSection.tsx
```diff
@@ -0,0 +1,75 @@
...
+  readonly,
+}: SettingsLogicFunctionDatabaseEventTriggerSectionProps) => {
+  const { t } = useLingui();
+
+  const [object, action] = isDefined(value)
+    ? value.eventName.split('.')
+    : ['', 'created'];
+
+  const updateOperation = (
+    _index: number,
```
charlesBochet: use object as parameter when multiple object + unused _index

packages/twenty-front/src/modules/settings/logic-functions/components/triggers/SettingsLogicFunctionHttpTriggerSection.tsx
```diff
@@ -0,0 +1,153 @@
...
+              onChange={(checked) => updateField('isAuthRequired', checked)}
+              disabled={readonly}
+              toggleSize="small"
+              color={theme.color.blue}
+            />
+            <StyledAuthLabel>{t`Require authentication`}</StyledAuthLabel>
+          </StyledAuthRow>
+          <SettingsLogicFunctionTriggerPayloadFormat
+            payload={buildHttpPayload(value)}
+            hint={t`Your handler receives this object. The body is null for GET requests; for POST/PUT/PATCH/DELETE it contains the parsed JSON sent by the client.`}
```
charlesBochet: not sure how lingui will handle this translation

packages/twenty-front/src/modules/settings/logic-functions/components/triggers/SettingsLogicFunctionTriggerPayloadFormat.tsx
```diff
@@ -0,0 +1,48 @@
...
+  font-size: ${themeCssVariables.font.size.xs};
+  font-weight: ${themeCssVariables.font.weight.semiBold};
+`;
+
+const StyledHint = styled.span`
+  color: ${themeCssVariables.font.color.tertiary};
+  font-size: ${themeCssVariables.font.size.sm};
+`;
+
+// Read-only preview of the payload the function will receive when triggered.
```
charlesBochet: remove comment + I don't get it :p

packages/twenty-front/src/pages/settings/applications/tabs/SettingsApplicationDetailContentTab.tsx
```diff
@@ -72,48 +94,153 @@ export const SettingsApplicationDetailContentTab = ({
...
           <H2Title
             title={t`Logic`}
-            description={t`Logic functions powering this app`}
+            description={t`Automation, AI, and access this app provides`}
           />
-          <SettingsLogicFunctionsTable logicFunctions={logicFunctionRows} />
+          <Table>
+            <SettingsApplicationContentSubtable
+              title={t`Logic functions`}
+              rows={filtered.logicFunctions}
```
charlesBochet, line 215: tbh it's a bit "spaghetti"
- prepare the data => rows
- we introduce a whole new typing
- pass it to components (new typing + drill down)

Is that because of marketplace vs installedApp?

It's fine but usually I'm more in favor of deriving data if possible when we needed rather that making the whole code area support a new typing + this is adding a lot on the data preparation

packages/twenty-front/src/pages/settings/applications/tabs/SettingsApplicationFrontComponentSettingsTab.tsx
```diff
@@ -0,0 +1,139 @@
...
+  componentName: string;
+  universalIdentifier?: string | null;
+  builtComponentChecksum: string;
+  isHeadless: boolean;
+  usesSdkClient: boolean;
+  createdAt: string;
+  updatedAt: string;
+};
+
+const StyledDescription = styled.div`
```
charlesBochet, line 24: When I see those, I'm always surprised it's not unified accross Settings > usually leads to inconsistencies

packages/twenty-front/src/pages/settings/applications/tabs/SettingsApplicationFrontComponentSettingsTab.tsx
```diff
@@ -0,0 +1,139 @@
...
+const StyledMonoText = styled.span`
+  color: ${themeCssVariables.font.color.primary};
+  font-family: ${themeCssVariables.code.font.family}, monospace;
+  font-size: ${themeCssVariables.font.size.sm};
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+`;
+
+const formatDateTime = (isoString: string): string => {
```
charlesBochet, line 40: use standard tooling :)

packages/twenty-front/src/pages/settings/layout/hooks/useFieldLabelByUid.ts
```diff
@@ -0,0 +1,32 @@
+import { objectMetadataItemsSelector } from '@/object-metadata/states/objectMetadataItemsSelector';
+import { useAtomStateValue } from '@/ui/utilities/state/jotai/hooks/useAtomStateValue';
+import { useMemo } from 'react';
+import { type Manifest } from 'twenty-shared/application';
+import { isDefined } from 'twenty-shared/utils';
+
+// Map<universalIdentifier, fieldLabel> over manifest fields + workspace fields,
```
charlesBochet: remove all these comments

packages/twenty-front/src/pages/settings/layout/hooks/useFieldLabelByUid.ts
```diff
@@ -0,0 +1,32 @@
...
+// for resolving fieldMetadataUniversalIdentifier references in view filters,
+// sorts, fields, and widget configurations.
+export const useFieldLabelByUid = (manifest: Manifest | undefined) => {
```
charlesBochet: I'm a bit confused as these hooks (not the first one I see) are based on objectMetadataItems, does it work for marketplace? is this code only for installed-apps. Not super clear reading the code

packages/twenty-front/src/pages/settings/layout/utils/getNavigationMenuItemDestination.ts
```diff
@@ -0,0 +1,105 @@
...
+export type NavigationMenuItemDestination =
+  | { kind: 'FOLDER' }
+  | { kind: 'LINK'; link: string | undefined }
+  | { kind: 'OBJECT'; label: string | undefined }
+  | { kind: 'PAGE_LAYOUT'; label: string | undefined }
+  | { kind: 'VIEW'; label: string | undefined }
+  | { kind: 'RECORD' }
+  | { kind: 'UNKNOWN' };
+
+export const getNavigationMenuItemDestination = (
```
charlesBochet: I feel this not specific to Settings, might be wrong

charlesBochet: and therefore alrady exists somewhere

packages/twenty-front/src/pages/settings/layout/utils/resolveManifestObjectLabel.ts
```diff
@@ -0,0 +1,34 @@
+import { type EnrichedObjectMetadataItem } from '@/object-metadata/types/EnrichedObjectMetadataItem';
+import { type Manifest } from 'twenty-shared/application';
+import { isDefined } from 'twenty-shared/utils';
+import { findObjectNameByUniversalIdentifier } from '~/pages/settings/applications/utils/findObjectNameByUniversalIdentifier';
+
+// Resolve an object's display label by universalIdentifier. Tries this app's
+// manifest first, then standard objects (matched by name, then optionally
+// labeled via the workspace's objectMetadataItems).
+export const resolveManifestObjectLabel = (
```
charlesBochet: this looks very temporary as we don't have the standard app yet

packages/twenty-front/src/pages/settings/layout/SettingsLayoutNavigationMenuItemDetail.tsx
```diff
@@ -0,0 +1,96 @@
...
+import {
+  type DetailRow,
+  SettingsLayoutDetailScaffold,
+} from '~/pages/settings/layout/components/SettingsLayoutDetailScaffold';
+import {
+  getNavigationMenuItemDestination,
+  getNavigationMenuItemDisplayLabel,
+} from '~/pages/settings/layout/utils/getNavigationMenuItemDestination';
+
+const renderDestinationLabel = (
```
charlesBochet: I feel we are recoding NavMenu here

charlesBochet (commented):
Overall additional comments 

- we already have isWorkspaceCustomApplication(application, currentWorkspace) in @/applications/utils/isWorkspaceCustomApplication, worth reusing for consistency.
- Pre-existing gap, newly reachable: isObjectMetadataReadOnly ignores applicationId, so the new "Fields added to other objects" rows can link to a fully editable SettingsObjectFieldEdit when an installed app contributed a field to a standard object. Not this PR's to fix, but a good follow-up since this PR surfaces the path.
- (Minor: pre-install / post-install hooks are still Test-runnable — probably intended, just worth confirming.)
- also lot of code in src/pages, should go under src/modules. IA is putting it here because the code was pre-existing I think

packages/twenty-front/src/modules/logic-functions/hooks/useLogicFunctionUpdateFormState.ts
```diff
@@ -57,6 +68,15 @@ export const useLogicFunctionUpdateFormState = ({
         timeoutSeconds: logicFunction.timeoutSeconds ?? 300,
         toolInputSchema:
           logicFunction.toolInputSchema || DEFAULT_TOOL_INPUT_SCHEMA,
+        cronTriggerSettings:
+          (logicFunction.cronTriggerSettings as CronTriggerSettings | null) ??
```
FelixMalfait: Do we need those casting? probably but double check

FelixMalfait:
@claude fix tests and push to this branch directly ; then also merge recent main

PR #19961, opened 2026-04-22
feat(community): add github-connector example app

charlesBochet:
## Summary

Adds a new community app at `packages/twenty-apps/community/github-connector` that demonstrates a complete, production-style GitHub integration built on the Twenty SDK.

It is extracted (and decoupled) from the internal `twenty-eng` workspace so external developers can use it as a reference for their own connectors.

What it ships:

- **Six synced objects**: `pullRequest`, `pullRequestReview`, `pullRequestReviewEvent`, `issue`, `projectItem`, `engineer`
- **Logic functions** for periodic backfills (PRs, reviews, issues, project items, contributors) and a single signed-webhook route trigger (`POST /github/webhook`) that performs idempotent upserts for `pull_request`, `pull_request_review`, `issues`, and `projects_v2_item` events
- **Views, navigation menu items and a GitHub folder** so the data is discoverable in the UI out of the box
- **Configurable repos / project numbers** via `GITHUB_REPOS` and `GITHUB_PROJECT_NUMBERS` application variables — no hardcoded org

## Authentication

Two interchangeable modes (PAT preferred for quick setup, GitHub App recommended for production):

1. **Personal Access Token** — set `GITHUB_TOKEN`. Used as-is for both REST and GraphQL.
2. **GitHub App** — set `GITHUB_APP_ID`, `GITHUB_APP_PRIVATE_KEY`, `GITHUB_APP_INSTALLATION_ID`. Issues a signed JWT, exchanges it for a short-lived installation token, and caches the token until expiry.

Webhook signature verification (`X-Hub-Signature-256`) is enforced when `GITHUB_WEBHOOK_SECRET` is set.

## Notes

- Built on `twenty-sdk@2.0.0` / `twenty-client-sdk@2.0.0`
- Decoupled from internal modules (`quality/bug`, `discord`, `release`, `code-build`, `project-management`) — `mustBeQa` is inlined and a local `github` nav folder replaces shared ones
- `npx twenty typecheck`, `yarn lint`, and `npx twenty build` all run cleanly
- Includes a comprehensive README with setup, env vars, webhook configuration, and the auth resolution flow

PR #20061, opened 2026-04-26
fix(logic-function): forward raw request body for HMAC signature verification

charlesBochet:
## Summary

- Add optional `rawBody?: string` to `LogicFunctionEvent` and forward it from the route trigger so HMAC-based webhook signatures (GitHub's `X-Hub-Signature-256`, Stripe, …) can be verified by user logic functions.
- Update `github-connector`'s `getRawBodyForSignature` to prefer `event.rawBody` (with the existing string/base64/null fallbacks kept for older runtimes).

## Why

GitHub computes `X-Hub-Signature-256` over the **raw bytes** of the request body. The receiver must verify against those exact bytes — key order, whitespace and unicode escaping all matter, so the parsed JSON body cannot be re-serialized to them.

Today the route trigger calls `extractBody(request)` which returns the parsed object only. NestJS already preserves the raw body on `request.rawBody` (the app is bootstrapped with `rawBody: true` in `main.ts`), but it was never propagated into `LogicFunctionEvent`.

As a result the github-connector's webhook handler always took the "raw body unavailable" branch and rejected every delivery (after #19961 / 962c2b3c14). With this change, signature verification can succeed end-to-end.

### DIFF
--- a/packages/twenty-shared/src/types/LogicFunctionEvent.ts
+++ b/packages/twenty-shared/src/types/LogicFunctionEvent.ts
@@ -1,44 +1,13 @@
-/**
- * AWS HTTP API v2 compatible request format for logic functions
- * @see https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html
- *
- * @typeParam TBody - The type of the request body. Defaults to `object` for parsed JSON bodies.
- */
 export type LogicFunctionEvent<TBody = object> = {
-  /** HTTP headers (filtered by forwardedRequestHeaders in route trigger) */
   headers: Record<string, string | undefined>;
-
-  /** Query string parameters (multiple values are joined with commas, e.g., "1,2,3") */
   queryStringParameters: Record<string, string | undefined>;
-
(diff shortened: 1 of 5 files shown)


Weiko (approved):
LGTM

prastoin (approved):
Nice !

packages/twenty-shared/src/types/LogicFunctionEvent.ts
```diff
@@ -1,31 +1,13 @@
...
   queryStringParameters: Record<string, string | undefined>;
-
-  /** Path parameters extracted from the route pattern (e.g., /users/:id → { id: '123' }). Multiple values are joined with commas. */
   pathParameters: Record<string, string | undefined>;
-
-  /** Request body */
   body: TBody | null;
-
-  /** Whether the body is base64 encoded */
+  rawBody?: string;
```
martmull, line 6: documentation updates are missing @charlesBochet (look for `RoutePayload` in `packages/twenty-docs/developers/extend/apps/logic-functions.mdx`)

PR #20064, opened 2026-04-26
feat(sentry): propagate workspace context to all spans

FelixMalfait:
## Summary

After the 2026-04-26 token-usage incident, identifying the responsible workspace from a Sentry trace required a Postgres scavenger hunt — Vercel AI SDK auto-instrumentation captures token counts and model name but no twenty-specific identifiers, and that same gap exists for every other auto-instrumented span (HTTP outbound, Postgres queries, GraphQL resolvers, Redis, etc.).

This PR plugs that gap globally, not just for AI:

- A small utility (`packages/twenty-server/src/engine/core-modules/sentry/utils/sentry-workspace-context.util.ts`) that writes workspace identifiers onto Sentry's active isolation scope as a `twenty` context block plus filterable tags and a `Sentry.setUser` call.
- Two hook points covering all server traffic:
  - **`WorkspaceAuthContextMiddleware`** — already runs after token hydration on the GraphQL, metadata, admin-panel, and REST routes. It now calls the utility once per authenticated request, before delegating to `withWorkspaceAuthContext`.
  - **`BullMQDriver.work` and `SyncDriver.processJob`** — every queue job now runs inside `Sentry.withIsolationScope` and applies workspace context from `job.data.workspaceId` (skipping silently for system jobs that don't carry one).
- A `beforeSendSpan` hook in `instrument.ts` that reads the scope's `twenty` context block back and projects it onto every span as `twenty.workspace.id` and (when available) `twenty.user_workspace.id` — dotted-namespace naming consistent with OTel/Sentry conventions like `user.id` and `http.response.status_code`. Spans without a workspace context (unauthenticated traffic) pass through untouched.

## Why this shape

Sentry's docs position `beforeSendSpan` as a per-span hook. The previous iteration set context only at AI-specific call sites, which left non-AI spans (DB queries, outbound HTTP, regular GraphQL queries, workflow steps not touching AI) entirely unenriched. Hooking the two existing global boundaries — auth middleware for HTTP/GraphQL/REST, and the queue driver `work()` callback for background jobs — covers every authenticated span across the app with no per-handler instrumentation.

## What's not in this PR

AI-specific identifiers (`twenty.agent.id`, `twenty.thread.id`, `twenty.turn.id`, `twenty.workflow_run.id`) are out of scope here. They're useful additions but require either propagating the IDs through the call stack or a more fine-grained scope (per-step, per-turn) than the request/job boundary, which is best handled in follow-up PRs that target those specific call sites.

## Test plan

- [ ] Make any authenticated GraphQL request locally and confirm the resulting span(s) in Sentry carry `twenty.workspace.id` and (for user-authenticated routes) `twenty.user_workspace.id`.
- [ ] Make any authenticated REST request and confirm the same.
- [ ] Trigger a queue job (chat stream, agent turn evaluation, workflow run, etc.) and confirm spans produced inside the worker carry `twenty.workspace.id`.
- [ ] Confirm that DB and outbound HTTP spans produced under the request/job also carry the workspace tag — these previously had no twenty-specific identifiers.
- [ ] In the Sentry UI, filter events by the `twenty.workspace.id` tag and confirm matching events appear.

## Notes for review

- Sentry init lives in `instrument.ts`, loaded before Nest bootstraps, so `beforeSendSpan` runs outside Nest DI and reads context off the isolation scope rather than holding a service reference.
- The middleware change is three lines; the BullMQ wrap is a single `Sentry.withIsolationScope` around the existing job handler body; the SyncDriver wrap mirrors it for the dev/test path. No new modules or DI providers.
- The previous iteration's `AiCallContextService` and per-handler `setContext` / `withContext` calls have been removed in favor of these two hooks.


FelixMalfait:

- #20064 — Sentry workspace/agent context on `gen_ai.*` spans (this PR)
- #20065 — Bill `executeAgent` in a finally block (catches failed runs and the eval pipeline)
- #20066 — Bill `POST /rest/ai/generate-text`

**Recommended merge order:** this PR (#20064) before #20065 — they touch `run-evaluation-input.job.ts` in different chunks and the executeAgent-finally PR rebases trivially on top. #20066 and #20067 are independent.

FelixMalfait:
Pushed [c90584a](https://github.com/twentyhq/twenty/commit/c90584af84) addressing the review feedback on idiomatic Sentry usage and OTel-style attribute naming.

**What changed:**

1. **Dropped the custom `AsyncLocalStorage`.** The service now uses Sentry's own primitives — `Sentry.withIsolationScope`, `Sentry.setUser`, `Sentry.setTag`, `Sentry.setContext`. These all write to the per-request isolation scope automatically. The `storage/` folder is gone.

2. **Span attributes now use dotted namespaces** (Sentry/OTel convention):
   - `twenty.workspace_id` -> `twenty.workspace.id`
   - `twenty.user_workspace_id` -> `twenty.user_workspace.id`
   - `twenty.agent_id` -> `twenty.agent.id`
   - `twenty.workflow_run_id` -> `twenty.workflow_run.id`
   - `twenty.turn_id` -> `twenty.turn.id`
   - `twenty.thread_id` -> `gen_ai.conversation.id` (the standard OTel gen_ai key)

3. **Two clear boundary patterns**, replacing the old single `run()`:
   - `setContext(ctx)` — for HTTP/GraphQL handlers and the workflow executor (request-scoped). Sentry's `httpIntegration` creates the isolation scope automatically per request on Node 22.12+, so we just write to it. Drops a level of indentation in resolvers.
   - `withContext(ctx, fn)` — for queue jobs (`stream-agent-chat`, `run-evaluation-input`, `evaluate-agent-turn`), where Sentry does *not* auto-isolate. Internally calls `Sentry.withIsolationScope`.

4. **`beforeSendSpan` reads from the isolation scope** via `Sentry.getIsolationScope().getScopeData().contexts.twenty` instead of the deleted custom ALS store.

**One deviation from "use the first-class API":** `Sentry.setConversationId` isn't yet exported in `@sentry/node` 10.27 (added in a later 10.x). Until we bump the SDK, `gen_ai.conversation.id` is set directly in `beforeSendSpan` from the stored thread_id. There's a TODO in `instrument.ts` flagging the swap.

Verified clean: `oxlint --type-aware` 0/9 errors, `prettier --check` clean, `tsgo` produces the same 203 pre-existing errors as `main` (none in the files I touched).

FelixMalfait:
Triaged @cubic-dev-ai's three findings:

- ✅ **threadId not tagged** — fixed in e05a17e53a. Added `twenty.thread.id` as a tag alongside the others. `setConversationId` still applies `gen_ai.conversation.id` to AI spans; the tag covers event-level filtering for consistency.
- ✅ **executeStep should use withContext** — fixed in e05a17e53a. Workflow runs originate from queue jobs as well as HTTP, and queue jobs don't get Sentry's automatic per-request isolation, so `withContext` is correct unconditionally.
- ❌ **`twenty.*_id` → `twenty.*.id` rename breaks documented keys** — keeping the new naming. The rename is the deliberate fix to follow Sentry/OTel dotted-namespace conventions (`user.id`, `gen_ai.conversation.id`, `http.response.status_code`). The previous PR description predates the rewrite — updating it now.

FelixMalfait:
Final architectural rewrite — the previous iterations of this PR sprinkled an `AiCallContextService` across every AI handler (resolvers, queue jobs, the workflow executor's `executeStep`). That patched the AI path but left every other span (DB queries, HTTP outbound, regular GraphQL queries, non-AI workflow steps) without a workspace identifier — exactly the situation that motivated the PR in the first place.

This rewrite drops that scaffold and hooks the two existing global boundaries Twenty already has:

- **HTTP / GraphQL / REST / metadata / admin-panel** flow through `WorkspaceAuthContextMiddleware` — every authenticated request now gets workspace context applied to the active Sentry isolation scope.
- **Every queue job** flows through `BullMQDriver.work` (and `SyncDriver.processJob` for dev/test parity) — each job is wrapped in `Sentry.withIsolationScope` with workspace context applied from `job.data.workspaceId`.

`beforeSendSpan` reads that context back and projects `twenty.workspace.id` (and `twenty.user_workspace.id` when present) onto every span. Spans outside an authenticated context (unauthenticated traffic, system jobs) pass through untouched.

Net effect: the diff is now ~170 lines (1 new utility file, 4 small modifications) vs. ~380 lines of per-handler instrumentation in the previous version.

In response to cubic's earlier review:
- **Finding 3** (`withContext` for `executeStep`) is now obsolete — `executeStep` no longer touches Sentry directly. The middleware/driver hooks cover both the HTTP-triggered and queue-triggered paths since both hit the global boundaries.
- The `ai-agent-monitor/run-evaluation-input.job.ts` overlap with #20065 is also gone — that file is reverted to main, so there is nothing to conflict with.
- AI-specific fields (`twenty.agent.id`, `twenty.thread.id`, `twenty.turn.id`, `twenty.workflow_run.id`) are out of scope for this PR and will land as follow-ups that can target the right per-step/per-turn scopes.

packages/twenty-server/src/engine/core-modules/message-queue/drivers/sync.driver.ts
```diff
@@ -62,7 +65,12 @@ export class SyncDriver implements MessageQueueDriver {
     const worker = this.workersMap[queueName];
 
     if (worker) {
-      await worker(job);
+      // Mirror the BullMQ driver: isolate the Sentry scope per-job and apply
+      // workspace context so beforeSendSpan can project attributes onto spans.
+      await Sentry.withIsolationScope(async () => {
```
FelixMalfait: We don't need Sentry for the sync driver

packages/twenty-server/src/engine/core-modules/sentry/utils/sentry-workspace-context.util.ts
```diff
@@ -0,0 +1,107 @@
...
+type WorkspaceSentryFields = {
+  workspaceId: string;
+  userWorkspaceId?: string;
+};
+
+// Apply Twenty workspace identifiers to the active Sentry isolation scope.
+// Sets user, tags, and a `twenty` context block. The beforeSendSpan hook in
+// instrument.ts reads this back and projects it onto every span as dotted
+// attributes (twenty.workspace.id, twenty.user_workspace.id).
+export const applyWorkspaceSentryContext = (
```
FelixMalfait: 1 file = 1 export

packages/twenty-server/src/engine/core-modules/message-queue/drivers/bullmq.driver.ts
```diff
@@ -108,25 +110,32 @@ export class BullMQDriver
...
-        await handler({ data: job.data, id: job.id ?? '', name: job.name });
-        const timeEnd = performance.now();
-        const executionTime = timeEnd - timeStart;
-
-        this.logger.log(
-          `Job ${job.id} with name ${job.name} processed on queue ${queueName} in ${executionTime.toFixed(2)}ms${workspaceSuffix}`,
-        );
-      },
+      async (job) =>
+        // Each job runs in its own Sentry isolation scope so the workspace
```
FelixMalfait: Remove this comment

packages/twenty-server/src/instrument.ts
```diff
@@ -45,6 +45,33 @@ if (process.env.EXCEPTION_HANDLER_DRIVER === ExceptionHandlerDriver.SENTRY) {
     profilesSampleRate: 0.3,
     sendDefaultPii: true,
     debug: process.env.NODE_ENV === NodeEnvironment.DEVELOPMENT,
+    beforeSendSpan: (span) => {
+      // Project the workspace context set by WorkspaceAuthContextMiddleware
```
FelixMalfait: Remove comment

packages/twenty-server/src/engine/core-modules/message-queue/drivers/sync.driver.ts
FelixMalfait: Fixed in 79121d79bf.

packages/twenty-server/src/engine/core-modules/sentry/utils/sentry-workspace-context.util.ts
FelixMalfait: Fixed in 79121d79bf.

packages/twenty-server/src/engine/core-modules/message-queue/drivers/bullmq.driver.ts
FelixMalfait: Fixed in 79121d79bf.

packages/twenty-server/src/instrument.ts
FelixMalfait: Fixed in 79121d79bf.

PR #20065, opened 2026-04-26
fix(ai-billing): bill executeAgent in a finally block so failed runs don't leak

FelixMalfait:
## Summary

`AgentAsyncExecutorService.executeAgent` consumes Anthropic tokens at two points (the main `generateText` and the optional structured-output sub-call). Billing was previously the **caller's** responsibility, executed only after `executeAgent` returned. If `executeAgent` threw — e.g. when `structuredResult.output == null` for a schema-mismatched response, or anything caught by the catch-and-rethrow — we paid Anthropic but never recorded a `usageEvent`. Likely the dominant source of the 716M-token-vs-3.27-credits discrepancy seen on the affected workspace in the 2026-04-26 incident.

## What changed

- Inject `AiBillingService` into `AgentAsyncExecutorService`. Add `workspaceId` (required), `userWorkspaceId`, and `operationType` (default `AI_WORKFLOW_TOKEN`) to `executeAgent`'s args.
- Capture `accumulatedUsage`, `cacheCreationTokens`, and `nativeWebSearchCallCount` into mutable locals as each `generateText` resolves. A throw between the main and structured-output calls still bills the first call's tokens; the schema-validation throw still bills the merged usage.
- Wrap the body in `try { ... } finally { ... }`. The finally calls `calculateAndBillUsage` and `billNativeWebSearchUsage`, each guarded by its own `try/catch + logger.error` so a billing exception can't mask the original execution error or block the second emit.
- `ai-agent.workflow-action.ts`: pass the new args; drop the now-redundant billing calls and `AiBillingService` injection. `AiBillingModule` removed from this action's module imports.
- `run-evaluation-input.job.ts`: pass `workspaceId` (already in `data`) and `userWorkspaceId: null`. **As a side effect, the eval pipeline now bills correctly** — closing an additional billing leak from the audit (`RunEvaluationInputJob` previously called `executeAgent` and discarded `executionResult.usage`).

## Behavior change worth calling out

Previously, failed agent executions were silently free. They will now be billed for the tokens Anthropic charged us. This is intentional and correct.

## Test plan
- [ ] Trigger a workflow agent action that succeeds — `usageEvent` count should match what was previously emitted.
- [ ] Trigger a workflow agent with a JSON response schema and ambiguous input that produces a non-schema-conforming output (`structuredResult.output == null`) — verify a `usageEvent` row is now written for the consumed tokens (was 0 rows previously).
- [ ] Trigger a `runEvaluationInput` GraphQL mutation — verify a `usageEvent` row is written (was 0 rows previously).

## Notes for review
- Conflicts trivially with the Sentry-context PR on `run-evaluation-input.job.ts`. Recommend merging Sentry-context first; this PR's 2-line argument addition rebases inside that PR's `aiCallContextService.run(...)` callback wrapper.
- A small follow-up after this lands: thread `billingContext` through the `experimental_repairToolCall` callback in `agent-async-executor.service.ts:198` (currently marked with a TODO from the title-gen+repair-tool PR).

### DIFF
--- a/packages/twenty-server/src/engine/metadata-modules/ai/ai-agent-execution/services/agent-async-executor.service.ts
+++ b/packages/twenty-server/src/engine/metadata-modules/ai/ai-agent-execution/services/agent-async-executor.service.ts
@@ -16,6 +16,7 @@ import { type Repository } from 'typeorm';
 
 import { isUserAuthContext } from 'src/engine/core-modules/auth/guards/is-user-auth-context.guard';
 import { type WorkspaceAuthContext } from 'src/engine/core-modules/auth/types/workspace-auth-context.type';
+import { ExceptionHandlerService } from 'src/engine/core-modules/exception-handler/exception-handler.service';
 import { type ToolProviderContext } from 'src/engine/core-modules/tool-provider/interfaces/tool-provider-context.type';
 import { NativeToolBinderService } from 'src/engine/core-modules/tool-provider/native/native-tool-binder.service';
 import { ToolRegistryService } from 'src/engine/core-modules/tool-provider/services/tool-registry.service';
@@ -69,6 +70,7 @@ export class AgentAsyncExecutorService {
     private readonly toolRegistry: ToolRegistryService,
     private readonly nativeToolBinder: NativeToolBinderService,
     private readonly aiBillingService: AiBillingService,
+    private readonly exceptionHandlerService: ExceptionHandlerService,
     @InjectRepository(RoleTargetEntity)
     private readonly roleTargetRepository: Repository<RoleTargetEntity>,
     @InjectRepository(WorkspaceEntity)
@@ -312,10 +314,9 @@ export class AgentAsyncExecutorService {
           userWorkspaceId,
         );
       } catch (billingError) {
-        this.logger.error(
-          `Failed to bill AI token usage for workspace ${workspaceId}`,
-          billingError instanceof Error ? billingError.stack : billingError,
-        );
+        this.exceptionHandlerService.captureExceptions([billingError], {
+          workspace: { id: workspaceId },
+        });
       }
 
       try {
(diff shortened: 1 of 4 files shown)


FelixMalfait:

- #20064 — Sentry workspace/agent context on `gen_ai.*` spans
- #20065 — Bill `executeAgent` in a finally block (this PR)
- #20066 — Bill `POST /rest/ai/generate-text`

**Recommended merge order:** #20064 before this PR — they touch `run-evaluation-input.job.ts` in different chunks and this PR rebases trivially on top.

**Follow-up after this lands:** thread `billingContext` through the `experimental_repairToolCall` callback in `agent-async-executor.service.ts:198` (currently a `TODO` from #20067).

packages/twenty-server/src/engine/metadata-modules/ai/ai-agent-execution/services/agent-async-executor.service.ts
```diff
@@ -269,6 +301,35 @@ export class AgentAsyncExecutorService {
...
+      try {
+        this.aiBillingService.calculateAndBillUsage(
+          agent?.modelId ?? AUTO_SELECT_SMART_MODEL_ID,
+          { usage: accumulatedUsage, cacheCreationTokens },
+          workspaceId,
+          operationType,
+          agent?.id ?? null,
+          userWorkspaceId,
+        );
+      } catch (billingError) {
```
FelixMalfait: Are we sure we want to catch this and not let it go to Sentry? Same for the billNativeWebSearchUsage?

FelixMalfait: Good catch -- fixed in f89d810665. Routed through `ExceptionHandlerService` (the canonical Twenty wrapper that hits the Sentry driver in prod). Kept the swallow so a billing failure cannot mask the original execution error, but it is no longer invisible.

FelixMalfait: No why not rethrow the error instead? @claude 

(same for both catch blocks)

FelixMalfait: Fixed in e151b61f18. Removed both try/catch wrappers and dropped the now-unused ExceptionHandlerService injection.

FelixMalfait:
Not perfect but better than before

PR #20066, opened 2026-04-26
fix(ai-billing): bill POST /rest/ai/generate-text usage to ClickHouse

FelixMalfait:
## Summary

`POST /rest/ai/generate-text` calls `generateText` and returns `usage` to the client without emitting a `usageEvent`. Authenticated, gated only by `PermissionFlagType.AI` — any workspace user with that permission could call it in a loop without billing. Identified during the 2026-04-26 incident audit.

## What changed

- Inject `AiBillingService` into `AiGenerateTextController`.
- Add `@AuthUserWorkspaceId() userWorkspaceId: string` to source the user-workspace identifier.
- Wrap the `generateText` call in `try { ... return ... } finally { ... }` so billing fires even if the controller throws after Anthropic was paid.
- Bill with `UsageOperationType.AI_WORKFLOW_TOKEN` and `cacheCreationTokens: result.usage.inputTokenDetails?.cacheWriteTokens ?? 0`.
- Inner `try/catch` around the billing emit so a billing error can't break the response.
- One-line module change: `AiGenerateTextModule` imports `AiBillingModule` (NestJS DI requirement).

## Test plan
- [ ] Call `POST /rest/ai/generate-text` with a small prompt; verify a `usageEvent` row appears in ClickHouse for the workspace with the correct token count and `operationType = AI_WORKFLOW_TOKEN`.
- [ ] Call with a malformed model id that throws after the API key is validated — verify no spurious billing call occurs (no Anthropic call was made).

## Notes for review
- Response shape unchanged.

### DIFF
--- a/packages/twenty-server/src/engine/metadata-modules/ai/ai-generate-text/controllers/ai-generate-text.controller.ts
+++ b/packages/twenty-server/src/engine/metadata-modules/ai/ai-generate-text/controllers/ai-generate-text.controller.ts
@@ -1,10 +1,19 @@
-import { Body, Controller, Post, UseFilters, UseGuards } from '@nestjs/common';
+import {
+  Body,
+  Controller,
+  Logger,
+  Post,
+  UseFilters,
+  UseGuards,
+} from '@nestjs/common';
 
 import { generateText } from 'ai';
 import { PermissionFlagType } from 'twenty-shared/constants';
 
 import { RestApiExceptionFilter } from 'src/engine/api/rest/rest-api-exception.filter';
+import { UsageOperationType } from 'src/engine/core-modules/usage/enums/usage-operation-type.enum';
 import type { WorkspaceEntity } from 'src/engine/core-modules/workspace/workspace.entity';
+import { AuthUserWorkspaceId } from 'src/engine/decorators/auth/auth-user-workspace-id.decorator';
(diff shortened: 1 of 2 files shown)


FelixMalfait:

- #20064 — Sentry workspace/agent context on `gen_ai.*` spans
- #20065 — Bill `executeAgent` in a finally block (catches failed runs and the eval pipeline)
- #20066 — Bill `POST /rest/ai/generate-text` (this PR)

This PR is independent of the others — no conflicts, no merge-order constraint.

packages/twenty-server/src/engine/metadata-modules/ai/ai-generate-text/controllers/ai-generate-text.controller.ts
```diff
@@ -50,18 +64,44 @@ export class AiGenerateTextController {
...
+              cacheCreationTokens:
+                result.usage.inputTokenDetails?.cacheWriteTokens ?? 0,
+            },
+            workspace.id,
+            UsageOperationType.AI_WORKFLOW_TOKEN,
+            null,
+            userWorkspaceId,
+          );
+        } catch (billingError) {
+          this.logger.error(
```
FelixMalfait: @claude why would we want to catch this? I guess if billing fails we want to know about it in Sentry and not catch/log instead. You can push to the branch directly

PR #20068, opened 2026-04-26
[AI] Add thread actions, filters, and archive support

ehconitin:
### Summary
- Add AI chat thread actions: rename, archive (soft-delete via `deletedAt`), and hard-delete with confirmation.
- Add chat thread filtering by status (active/archived/all), group-by mode, and last activity.
- Rework drawer/side-panel thread lists to share thread sections, item menus, archive icons, and empty-state behavior.
- Extend server chat thread model/API with `deletedAt`, mutations, broadcasts, and archive-aware stream guards.

### Decisions
- Two-stage lifecycle: Archive sets `deletedAt` (soft); Delete is a separate action on archived threads that hard-deletes the row. Aligns with Twenty's soft-delete convention (Felix's suggestion).
- `lastMessageAt` is derived from `MAX(agentMessage.createdAt)` on read, not stored. List query does inline aggregation for sort; `@ResolveField` covers single-thread / mutation paths so the schema contract is honest everywhere. Matches `timeline-messaging.service.ts` precedent and the existing `totalInputCredits` / `totalOutputCredits` `@ResolveField` pattern in the same resolver.
- Replaced auto-CRUD `chatThreads` (cursor-paginated Connection) with a custom `[AgentChatThreadDTO!]` resolver. Frontend metadata-store treats threads as a flat collection and filters/sorts client-side, so cursor pagination was performative.
- Sending in an archived chat unarchives it optimistically on the client and authoritatively on the server.
- Grouping and last-activity filtering use `lastMessageAt ?? updatedAt` so archive/rename don't bump threads in the list.
- Kept metadata-store core API unchanged; AI chat uses the same local cast pattern already used by other metadata-store partial updates.

https://github.com/user-attachments/assets/1b179b7b-1a2a-4a7a-aa0a-c88f6f051a87


FelixMalfait:
Nice! Can we show the "..." only on hover? It's a bit too much visual repetition otherwise

packages/twenty-front/src/modules/ai/components/AiChatThreadDeleteConfirmationModal.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+    await deleteChatThread(aiChatThreadPendingDelete.threadId);
+    setAiChatThreadPendingDelete(null);
+  };
+
+  // TODO: server uses TypeORM softDelete (sets deletedAt) but there is no
+  // restore UI and no cron purges deleted agent chat threads. Once a recovery
+  // flow lands (filter for "Deleted" + restoreChatThread mutation + retention
+  // job), update the copy below to reflect the actual retention window.
+  return (
```
FelixMalfait, line 38: I feel like archivedAt is the softDelete (you can set custom deletedAt column name in typeorm) ; and then "delete" is actually the hard delete? This todo implies we still have a "deletedAt" which shouldn't happen for me, worth tackling it

packages/twenty-front/src/modules/ai/components/NavigationDrawerAiChatContent.tsx
```diff
@@ -62,33 +73,53 @@ export const NavigationDrawerAiChatContent = () => {
...
+                title={dateGroup.title}
+                threads={dateGroup.threads}
+                currentThreadId={currentAiChatThread}
+                onThreadClick={handleThreadClick}
+                rightIcon={index === 0 ? filterDropdown : undefined}
+                alwaysShowRightIcon={index === 0}
+              />
+            ))}
+          </StyledSectionsContainer>
+        ) : (
```
FelixMalfait, line 105: I'm not sure I understand this ternary? Could you please double check it makes sense? (it probably does but I don't see why)

packages/twenty-front/src/modules/ai/constants/AgentChatThreadFilterStatus.ts
```diff
@@ -0,0 +1,8 @@
+export const AGENT_CHAT_THREAD_FILTER_STATUS = {
+  ACTIVE: 'active',
+  ARCHIVED: 'archived',
+  ALL: 'all',
+} as const;
+
+export type AgentChatThreadFilterStatus =
```
FelixMalfait: One export per file

packages/twenty-front/src/modules/ai/constants/AgentChatThreadGroupBy.ts
```diff
@@ -0,0 +1,7 @@
+export const AGENT_CHAT_THREAD_GROUP_BY = {
+  DATE: 'date',
+  NONE: 'none',
+} as const;
+
+export type AgentChatThreadGroupBy =
```
FelixMalfait: One export per file

packages/twenty-front/src/modules/ai/constants/AgentChatThreadLastActivityFilter.ts
```diff
@@ -0,0 +1,10 @@
+export const AGENT_CHAT_THREAD_LAST_ACTIVITY_FILTER = {
+  ALL: 'all',
+  ONE_DAY: '1d',
+  THREE_DAYS: '3d',
+  SEVEN_DAYS: '7d',
+  THIRTY_DAYS: '30d',
+} as const;
+
+export type AgentChatThreadLastActivityFilter =
```
FelixMalfait: On export per file

packages/twenty-server/src/database/commands/upgrade-version-command/2-2/2-2-instance-command-fast-1777190602079-add-archived-deleted-to-agent-chat-thread.ts
```diff
@@ -0,0 +1,27 @@
...
+@RegisteredInstanceCommand('2.2.0', 1777190602079)
+export class AddArchivedDeletedToAgentChatThreadFastInstanceCommand
+  implements FastInstanceCommand
+{
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      'ALTER TABLE "core"."agentChatThread" ADD "archivedAt" TIMESTAMP WITH TIME ZONE',
+    );
+    await queryRunner.query(
+      'ALTER TABLE "core"."agentChatThread" ADD "deletedAt" TIMESTAMP WITH TIME ZONE',
```
FelixMalfait: So you did delete it? Did I misunderstand the TODO then?

packages/twenty-server/src/engine/metadata-modules/ai/ai-chat/services/agent-chat-streaming.service.ts
```diff
@@ -139,6 +139,18 @@ export class AgentChatStreamingService {
...
   ): Promise<void> {
+    // Async stream cancellation: a worker's `finally` may land here after
+    // the thread was archived or soft-deleted.
+    const threadStatus = await this.threadRepository.findOne({
+      where: { id: threadId },
+      withDeleted: true,
+      select: ['id', 'archivedAt', 'deletedAt'],
+    });
+
+    if (!threadStatus || threadStatus.archivedAt || threadStatus.deletedAt) {
```
FelixMalfait: Doesn't really help for hard delete ; I can see reading this why we wouldn't want hard delete. For eval, logs etc it's good to have soft delete and makes things easier. But I feel like for privacy reasons we need to give people a way to completely erase this so I'd still go in that direction

packages/twenty-server/src/engine/metadata-modules/ai/ai-chat/entities/agent-chat-thread.entity.ts
```diff
@@ -69,9 +70,15 @@ export class AgentChatThreadEntity {
   @OneToMany(() => AgentMessageEntity, (message) => message.thread)
   messages: EntityRelation<AgentMessageEntity[]>;
 
+  @Column({ type: 'timestamptz', nullable: true })
```
FelixMalfait, line 79: Should this be the DeleteDateColumn? Not sure if it's a good idea, maybe it isn't....

packages/twenty-front/src/modules/ai/utils/groupThreadsByDate.ts
```diff
@@ -1,29 +1,95 @@
...
-    },
-    { today: [], yesterday: [], older: [] },
-  );
+  today = new Date(),
+): AgentChatThreadDateGroup[] => {
+  const groupedThreadsByDate = new Map<string, AgentChatThreadDateGroup>();
+
+  for (const thread of threads) {
+    // TODO: use a dedicated conversation activity timestamp once available;
+    // updatedAt also changes for archive/unarchive metadata updates.
```
FelixMalfait: Is there a touch() or something when we create a new message that updates the thread.updatedAt?

Not sure if it's worth introducing a field or if should just be a join to fetch the last message along with each thread? I guess it would work ok and can be done already? Or is it more complexity that what I foresee?

packages/twenty-front/src/modules/ai/utils/groupThreadsByDate.ts
```diff
@@ -1,29 +1,95 @@
...
+
 import { type AgentChatThread } from '~/generated-metadata/graphql';
 
-import { type DateGroupKey } from '@/ai/utils/dateGroupKey';
+export type AgentChatThreadDateGroup = {
+  id: string;
+  title: string;
+  threads: AgentChatThread[];
+};
+
```
FelixMalfait, line 11: Was this well tested with timezone?

packages/twenty-front/src/modules/ai/states/selectors/agentChatVisibleThreadsSelector.ts
```diff
@@ -0,0 +1,50 @@
...
+          break;
+        case AGENT_CHAT_THREAD_FILTER_STATUS.ARCHIVED:
+          if (!thread.archivedAt) return false;
+          break;
+        case AGENT_CHAT_THREAD_FILTER_STATUS.ALL:
+          break;
+      }
+
+      if (cutoffMs !== null) {
+        // TODO: use a dedicated conversation activity timestamp once available;
```
FelixMalfait: thread.lastMesage.created?

packages/twenty-front/src/modules/ai/constants/AgentChatThreadLastActivityFilter.ts
FelixMalfait: (also true more more files but won't comment on all)

packages/twenty-front/src/modules/ai/utils/groupThreadsByDate.ts
```diff
@@ -1,29 +1,95 @@
...
+  }
+
+  if (localDayDifference === 1) {
+    return {
+      id: 'yesterday',
+      title: t`Yesterday`,
+    };
+  }
+
+  if (localDayDifference >= 2 && localDayDifference <= 7) {
```
FelixMalfait, line 44: Do I understand correctly that it's one section by day? Maybe I misunderstood? Feels like a lot?

FelixMalfait: I wanted to check but it looks like OpenAI recently remove the time-based section? I see "Pin Chat" (is pinned at) and just a section for pinned, then one for pinned. So no good inspo

FelixMalfait (commented):
Great work!

ehconitin:
@cubic-dev-ai heads up -- you're reviewing against the old shape:

archivedAt was renamed to deletedAt (soft delete convention). There's no archivedAt anymore.
Archive = soft delete, Delete = hard delete on an already-archived thread. Two-stage, intentional.
(id, deletedAt) index follows existing precedent (agent.entity.ts, logic-function.entity.ts).
Idempotency + draft-restore points were already fixed.

Mind taking another pass?

FelixMalfait (approved):

PR #20078, opened 2026-04-27
Filtered upgrade logs when stopping before starting next instance segment

prastoin:
# Introduction
In a nutshell, added a more readable logs that relates that when performing a workspace fitlered workspace you can only browser the workspace commands sequence

This PR is not a fix, but a log improvement
As before when cross-upgrading a single workspace you would be facing a `instance` sync barrier error as not all your workspaces would have been updated

Now logging and early returning instead of letting the guard hard throw

## Tests
Created a dedicated test for instance prevention on filtered upgrade
Standardized  `migrationRecordToKey` usage across all upgrade integration tests suites


Weiko (approved):
LGTM

PR #20088, opened 2026-04-27
Fix stale UI state after stop-impersonation

FelixMalfait:
## Summary

A customer reported that after **Stop Impersonating**, the sidebar still
showed the impersonated user's pinned favorites, the AI chat tab toggle,
and the AI chat history — even though the original admin's session was
correctly restored.

## Root cause

The refactor in #19597 replaced the previous `signOut()`-based stop flow
with an in-place token swap, but only cleared Apollo cache + reloaded
the user. Several user-scoped client stores were left untouched:

- **`metadataStoreState`** is localStorage-backed (`navigationMenuItems`,
  `agentChatThreads`, `views`, `pageLayouts`, etc.) and only refreshed
  by `MinimalMetadataLoadEffect`. That effect is gated by
  `metadataLoadedVersion` + `desiredLoadState`, neither of which flips
  on a same-workspace token swap, so the effect never re-runs.
- **In-memory AI atoms** (`currentAiChatThreadState`, `agentChatInputState`,
  `hasInitializedAgentChatThreadsState`) keep pointing at the
  impersonated user's selected thread / input.
- **Session localStorage keys** (`agentChatDraftsByThreadIdState`,
  `lastVisitedObjectMetadataItemIdState`, `lastVisitedViewPerObjectMetadataItemState`,
  `playgroundApiKeyState`) carry the impersonated user's drafts and
  navigation state.

`clearSession()` (used by logout) avoids this because it calls
`applyMockedMetadata()` and flips `desiredLoadState` mocked↔real, which
chain-triggers a full metadata reload on next sign-in.

## Fix

Extract a `resetUserScopedClientState` helper inside
`useImpersonationSession` that:

1. Calls `clearSessionLocalStorageKeys()` to drop user-scoped localStorage
   keys.
2. Resets the in-memory AI session atoms.
3. Marks `metadataStoreState['agentChatThreads']` as `'empty'`.
   `useLoadStaleMetadataEntities` does **not** handle this entity key, so
   without an explicit reset to `'empty'` the
   `AgentChatThreadInitializationEffect` (which only fires on `'empty'`)
   would never refetch.
4. Calls `invalidateMetadataStore()` to clear all `currentCollectionHash`
   values and bump `metadataLoadedVersion`, forcing
   `MinimalMetadataLoadEffect` to re-run and refetch
   `navigationMenuItems`, `views`, `pageLayouts`, etc. against the new
   token.

The helper is applied to both `startImpersonating` and `stopImpersonating`
— start had the same latent bug; the impersonated user could see the
admin's favorites until the cache happened to refresh.

## Test plan

- [ ] As an admin user, pin some favorites in the sidebar
- [ ] Impersonate a user with different favorites → favorites should
      switch to the impersonated user's
- [ ] Click "Stop Impersonating" → sidebar should immediately show the
      admin's favorites (not the impersonated user's)
- [ ] As an admin **without** AI permission, impersonate a user **with**
      AI permission, open AI chat, send a message, then stop impersonating
      → AI chat history should be empty / inaccessible (the AI tab
      visibility itself is fixed in a separate PR)
- [ ] Type a draft in AI chat as the impersonated user → after stop, the
      draft should be gone
- [ ] Verify regular sign-out still works while impersonating
- [ ] Verify the impersonation banner still shows / hides correctly


packages/twenty-front/src/modules/auth/hooks/useImpersonationSession.ts
```diff
@@ -17,53 +12,53 @@ type StoredImpersonationSession = {
...
 
     if (!raw) {
-      // No stored session — likely a cross-workspace tab opened via redirect.
-      // Try closing the tab (works when opened via window.open or target=_blank).
+      // Cross-workspace tab opened via redirect — no stored admin session
+      // to restore. Close the tab; fall back to sign out if the browser
+      // blocks window.close().
       window.close();
-      // If window.close() was blocked by the browser, fall back to sign out.
       await signOut();
```
Weiko, line 57: is it reachable?

FelixMalfait:
Expanded the scope to also rewrite `clearSession` in `useAuth.ts` with the same reload approach. Same brittleness lived there: enumerated atom resets + `applyMockedMetadata` + Apollo `clearStore`, all missing the user-scoped metadata-store localStorage entries and any future user state added without remembering to update the list.

The new `clearSession` is 5 lines: clear sessionStorage, clear the user-scoped localStorage keys, null the cookie-backed token + workspace-domain atoms (synchronous so the next page boot reads cleared cookies), reload to `/sign-in`. Drops ~70 lines of imports and explicit resets across `useAuth.ts` + its test.

packages/twenty-front/src/modules/auth/hooks/useImpersonationSession.ts
FelixMalfait: Not always, that's what the comment was describing (window.close api available if it was opened programmatically I think but not otherwise)

FelixMalfait:
@claude fix CI and push to this branch

Qodo-Free-For-OSS:
Hi, `useAuth.clearSession()` now hard-navigates with `window.location.assign()`, so any code after `signOut()` (even when awaited) will typically never run. This breaks flows that do post-signout actions (e.g. `DeleteWorkspace` calling `redirectToDefaultDomain()` after `signOut()`).

**How to fix:** Avoid hard nav in signOut

> ### Issue description
> `clearSession()` performs `window.location.assign()`, which aborts subsequent JS execution. Several callers expect to run logic after `await signOut()` (e.g., domain redirects).
> 
> ### Issue Context
> Hard reload is useful for clearing user-scoped state, but the auth API must not silently invalidate downstream control flow.
> 
> ### Fix Focus Areas
> - packages/twenty-front/src/modules/auth/hooks/useAuth.ts[106-112]
> - packages/twenty-front/src/modules/settings/profile/components/DeleteWorkspace.tsx[23-30]
> 
> ### Suggested direction
> Either:
> 1) revert to SPA navigation (`navigate`) and explicitly clear the remaining user-scoped stores; or
> 2) keep hard reload but remove/relocate any post-signOut logic at call sites (e.g., perform domain redirect before calling signOut, or incorporate domain logic into clearSession via a parameter).

---

Qodo-Free-For-OSS:
Hi, `useAuth.signOut()` now triggers `window.location.assign()`, which is not mocked in this test file and commonly throws in jsdom or causes unwanted navigation side effects. The updated unit test calls `signOut()` without stubbing `window.location.assign`, making the test suite brittle.

**Severity:** action required | **Category:** reliability

**How to fix:** Mock window.location.assign in tests

> ### Issue description
> `signOut()` calls `window.location.assign()`, which can throw or cause flaky behavior under jsdom.
> 
> ### Issue Context
> The test `useAuth.test.tsx` calls `signOut()` and should remain unit-level (no real navigation).
> 
> ### Fix Focus Areas
> - packages/twenty-front/src/modules/auth/hooks/__tests__/useAuth.test.tsx[141-151]
> - packages/twenty-front/setupTests.ts[34-40]
> 
> ### Suggested change
> Stub `window.location.assign` in either `setupTests.ts` (global) or in `useAuth.test.tsx` (local), and assert it was called with `AppPath.SignInUp` if needed.

---
*Found by [Qodo](https://qodo.ai) code review*

FelixMalfait:
@claude I told you to fix CI but the tests are still red. Please fix CI and push to this branch

PR #20090, opened 2026-04-27
chore: export the enum view calendar layout to avoid typing errors on apps

abusarah-tech:
This is a small PR to fix an issue that came up when I created an app with a custom object. I think the enum `ViewCalendarLayout` was not caught as part of the exports. 

image


prastoin (approved):
Thanks !

PR #20067, opened 2026-04-26
fix(ai-billing): bill thread title generation and tool-call repair

FelixMalfait:
## Summary

Two `generateText` call sites were unbilled — identified during the 2026-04-26 incident audit:

- **Thread title generation** (`AgentTitleGenerationService.generateThreadTitle`) — fires once per new chat thread; low-volume but completeness matters.
- **Tool-call repair** (`repairToolCall` util, used inside `experimental_repairToolCall` callbacks) — can fire `MAX_STEPS` times per agent turn if a model gets stuck producing malformed tool calls.

## What changed

- `AgentTitleGenerationService` — inject `AiBillingService`, expand `generateThreadTitle` signature to accept `workspaceId` and `userWorkspaceId`, bill in a `finally` block with `UsageOperationType.AI_CHAT_TOKEN`. Restructured so the no-default-model short-circuit happens before the `try`, avoiding a fake billing call.
- `repair-tool-call.util.ts` — added an optional `billingContext` arg containing `aiBillingService`, `modelId`, `workspaceId`, `userWorkspaceId`, `operationType`. Wraps the `generateText` call in `try/finally`; bills in finally with the operation type provided by the caller. Optional so existing callers keep compiling.
- `chat-execution.service.ts` — threads `billingContext` (`AI_CHAT_TOKEN`) into the repair callback.
- `agent-chat.service.ts` — passes `workspaceId` and `thread.userWorkspaceId` to `generateThreadTitle`.
- `agent-async-executor.service.ts` — TODO comment marking that the repair callback should thread billing once `executeAgent` accepts `workspaceId` (depends on the executeAgent-finally PR).

## Follow-up

After the executeAgent-finally PR lands, `workspaceId` is in scope at the `experimental_repairToolCall` callback in `agent-async-executor.service.ts:198`. Thread `billingContext` through to fire repair-call billing for workflow agents too, and remove the TODO. Tiny follow-up.

## Test plan
- [ ] Create a new chat thread; verify a `usageEvent` row is written for the title generation call.
- [ ] Send a chat message that triggers tool-call repair (e.g. force a malformed tool call); verify a `usageEvent` row for the repair sub-call.

## Notes for review
- `billingContext` arg made **optional** to allow incremental rollout — the executeAgent-finally PR plus a small follow-up will close the agent-async-executor side.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #20019, opened 2026-04-24
Billing - optimize usageEvent CH table

etiennejouan:
- Update usageEvent clickhouse table, partitioning, indexing and projection (auto materialized view) to optimize credit usage queries
- Add caching for available credits and billing subscription

To do in next PR: deprecate enforceCapUsage cron. Bonus : real-time on billingSubscription

PR #20096, opened 2026-04-27
fix(billing): gate AI credit-cap at entry points instead of workflow executor

FelixMalfait:
## Background

The 2026-04-26 incident saw 716M Sonnet 4.6 tokens consumed in a single trial workspace. Two causes: failed agent executions weren't billed (addressed by #20065) and the credit-cap gate had been removed from `WorkflowExecutorWorkspaceService.executeStep` in #19904, leaving no enforcement point at all.

## Why not just revert #19904

#19904 was right that gating at the workflow executor is too coarse. When one user exhausted a workspace's credits via chat, *all* workflows hard-failed mid-run — including cheap DB/CRUD/branch automations costing essentially nothing. Reverting would re-introduce that cliff.

## New design: gate at the AI entry points

The chat resolver already gates this way (`agent-chat.resolver.ts:137-148`). This PR replicates the same pattern at every other point where the workspace can incur real AI cost:

- `executeAgent` in `agent-async-executor.service.ts`
- the REST handler in `ai-generate-text.controller.ts`
- `generateThreadTitle` in `agent-title-generation.service.ts`

In each, after auth/validation: skip if `IS_BILLING_ENABLED` is false; otherwise call `BillingService.canBillMeteredProduct(workspaceId, BillingProductKey.WORKFLOW_NODE_EXECUTION)`; on `false`, throw `BillingException(BILLING_CREDITS_EXHAUSTED)`. No new method, no new exception code, no new product key.

This matches industry convention (Lovable/Replit also gate at the expensive-operation boundary, not at every cheap step).

## Deliberately not gated

- `WorkflowExecutorWorkspaceService.executeStep` — the design choice is now intentional, so the #19904 TODO is replaced by a one-line absolute-behavior comment explaining why the gate isn't here. Cheap workflow steps (DB CRUD, branching, action steps) are not gated, so a chat-driven cap exhaustion does not block non-AI automations.
- `repair-tool-call.util` — repair is a sub-call inside an already-gated AI flow. If the parent is gated, repair will naturally not run. Adding a gate here adds complexity without value.

## Net effect

A workspace that exhausts credits via chat or AI agent stops making AI calls. Its non-AI workflows continue running normally. A workflow with both AI and non-AI steps fails at the AI step with `BILLING_CREDITS_EXHAUSTED`, but downstream non-AI steps that don't depend on the AI output still run.

## Conflicts

This PR overlaps with three other in-flight PRs in the same files. None of them touch the gate logic; rebasing on top of any of them is trivial:

- #20065 (agent-async-executor): adds `workspaceId` to `executeAgent` args and bills in `finally`. The gate at the top of `executeAgent` from this PR sits naturally above that.
- #20066 (REST controller): adds usage billing to the controller.
- #20067 (title gen): adds usage billing to title generation and tool-call repair.

Recommend landing #20065/#20066/#20067 first; this PR rebases trivially on top.

## Tests

Out of scope per the PR series convention. The existing chat-resolver gate isn't unit-tested either; this PR follows the same precedent. Follow-up: add integration coverage that exercises a workspace at `hasReachedCurrentPeriodCap=true` against each of the three new gates plus the pre-existing chat-resolver gate.

## Future follow-ups

- Per-user soft cap inside a workspace (the Lovable Business-tier pattern), so one user can't exhaust the workspace's cap.
- Pre-flight cost estimate so the user sees an "approaching cap" warning before the hard stop.
- Rename `BillingProductKey.WORKFLOW_NODE_EXECUTION` — the name predates this design choice and is misleading now that it gates AI entry points rather than workflow nodes.

## Test plan

- [ ] Trigger a workspace into `hasReachedCurrentPeriodCap=true`.
- [ ] Send a chat message — expect failure with `BILLING_CREDITS_EXHAUSTED`.
- [ ] Run a workflow whose only AI step is an `ai-agent` action — expect that step to fail with `BILLING_CREDITS_EXHAUSTED`, downstream non-AI steps still run.
- [ ] POST to `/rest/ai/generate-text` — expect `BILLING_CREDITS_EXHAUSTED`.
- [ ] Create a new chat thread (which kicks off `generateThreadTitle`) — expect `BILLING_CREDITS_EXHAUSTED`.
- [ ] Run a workflow with no AI step (only DB CRUD/branching/actions) — expect it to run unaffected.


packages/twenty-server/src/engine/metadata-modules/ai/ai-agent-execution/services/agent-async-executor.service.ts
```diff
@@ -113,6 +122,20 @@ export class AgentAsyncExecutorService {
     rolePermissionConfig?: RolePermissionConfig;
     authContext?: WorkspaceAuthContext;
   }): Promise<AgentExecutionResult> {
+    if (agent && this.twentyConfigService.get('IS_BILLING_ENABLED')) {
```
Weiko: nit: isDefined?

packages/twenty-server/src/engine/metadata-modules/ai/ai-chat/services/agent-title-generation.service.ts
```diff
@@ -20,13 +27,29 @@ export class AgentTitleGenerationService {
...
+    private readonly twentyConfigService: TwentyConfigService,
   ) {}
 
   async generateThreadTitle(
     messageContent: string,
     workspaceId: string,
     userWorkspaceId: string | null,
   ): Promise<string> {
+    if (this.twentyConfigService.get('IS_BILLING_ENABLED')) {
+      const canBill = await this.billingService.canBillMeteredProduct(
```
Weiko: nit: should IS_BILLING_ENABLED be part of canBillMeteredProduct?
nit2: should we add a canBillMeteredProductOrThrow?

Weiko (approved):
LGTM

packages/twenty-server/src/engine/metadata-modules/ai/ai-generate-text/controllers/ai-generate-text.controller.ts
```diff
@@ -43,6 +52,20 @@ export class AiGenerateTextController {
       );
     }
 
+    if (this.twentyConfigService.get('IS_BILLING_ENABLED')) {
+      const canBill = await this.billingService.canBillMeteredProduct(
+        workspace.id,
+        BillingProductKey.WORKFLOW_NODE_EXECUTION,
```
Weiko: Don't we have a more "generic" product key for AI? Is it always tied to workflow node exection? @FelixMalfait

FelixMalfait: WORKFLOW_NODE_EXECUTION is misnamed but we can't really change it easily (used by Stripe). It sould be called "credit" or "resource usage"

packages/twenty-server/src/engine/metadata-modules/ai/ai-chat/services/agent-title-generation.service.ts
FelixMalfait: @claude please address Weiko's comment. Take a step back / make sure to consider all impacts and report with a comment + push to the branch directly

Qodo-Free-For-OSS:
Hi, The new entry-point gates call BillingService.canBillMeteredProduct(), but that method can throw BillingException(BILLING_SUBSCRIPTION_NOT_FOUND) when a workspace has billing enabled but no subscription record. This makes the gate fail with an unrelated 404-style billing error instead of cleanly evaluating to “cannot bill”.

**Severity:** action required | **Category:** reliability

**How to fix:** Make canBill return boolean

> ## Issue description
> AI entry-point gating calls `BillingService.canBillMeteredProduct()` expecting a boolean, but it can throw if a subscription record is missing (`getCurrentBillingSubscriptionOrThrow`). This makes gating fail with `BILLING_SUBSCRIPTION_NOT_FOUND` instead of returning `false` (or a consistent billing-cap error).
> 
> ## Issue Context
> `canBillMeteredProduct()` is used as a predicate at multiple entry points. Throwing inside it introduces new failure modes and inconsistent error mapping.
> 
> ## Fix Focus Areas
> - packages/twenty-server/src/engine/core-modules/billing/services/billing.service.ts[70-105]
>   - Replace `getCurrentBillingSubscriptionOrThrow` with the non-throwing `getCurrentBillingSubscription` and return `false` when subscription is missing.
>   - Keep existing status/plan/product/cap checks unchanged.
> 
> (Alternative if you want to preserve throwing semantics: catch `BillingExceptionCode.BILLING_SUBSCRIPTION_NOT_FOUND` at each entry point and treat it as `canBill=false` / map to a consistent billing error.)

---

FelixMalfait:
@FelixMalfait Addressed @Weiko's feedback in `b98d203b641`. Took a step back since `main` had moved meaningfully — full report below.

## What changed

Two commits pushed:

1. **`97df41abbee`** — merge `origin/main` into the branch. `main` had landed several billing-related changes since this PR was opened, most importantly [#20019](https://github.com/twentyhq/twenty/pull/20019) which **removed `canBillMeteredProduct` from `BillingService`** and replaced it with `BillingUsageService.hasAvailableCredits` (a simpler workspace-scoped credits check, with caching). The chat resolver was migrated to the new API in that PR; this PR's three new gate sites still referenced the deleted method, so I had to migrate them too.

2. **`b98d203b641`** — refactor per @Weiko's two nits:

   - **"should `IS_BILLING_ENABLED` be part of `canBillMeteredProduct`?"** — moved the check inside `BillingUsageService.hasAvailableCredits`. Callers no longer wrap each gate in `if (twentyConfigService.get('IS_BILLING_ENABLED')) {…}`.
   - **"should we add a `canBillMeteredProductOrThrow`?"** — added `BillingUsageService.hasAvailableCreditsOrThrow(workspaceId)`. It calls `hasAvailableCredits` and throws `BillingException(BILLING_CREDITS_EXHAUSTED)` when no credits remain. All four AI entry points now use it.
   - **"`isDefined`?"** in `agent-async-executor.service.ts:125` — replaced `if (agent && …)` with `if (isDefined(agent))`.

## Impact

- **Net -55 lines** (5 files changed, +26 / -81) — the gate at every entry point collapsed from a 12-line `if/canBill/throw` block to one line: `await this.billingUsageService.hasAvailableCreditsOrThrow(workspaceId);`.
- Imports cleaned up: dropped `BillingException`, `BillingExceptionCode`, `BillingService`, `BillingProductKey`, `TwentyConfigService` from the four caller files (no longer needed at the call site since the helper handles them).
- `agent-chat.resolver.ts` (the canonical pattern from before this PR) also migrates to the new helper for consistency.
- Module wiring unchanged — `BillingModule` (which exports `BillingUsageService`) was already imported by all four caller modules.

## Other reviewer comments — status

- **@cubic-dev-ai (line 62)** "BillingException returns HTTP 400 instead of 402" — **not addressed in this iteration.** The REST controller uses `RestApiExceptionFilter`/`AiRestApiExceptionFilter`, neither of which is the dedicated `BillingApiExceptionFilter` (which would map to the right status code via `getBillingExceptionStatusCode`). Fix is to add `BillingApiExceptionFilter` to the controller's `@UseFilters` — small follow-up worth doing but felt out of scope for the Weiko refactor.
- **@Weiko (line 58)** about `WORKFLOW_NODE_EXECUTION` being misnamed for AI — already discussed in your reply ("misnamed but tied to Stripe"). Resolved.
- **Qodo bot** about `canBillMeteredProduct` throwing `BILLING_SUBSCRIPTION_NOT_FOUND` — partially mitigated. `hasAvailableCredits` uses `workspaceCacheService.getOrRecompute` rather than the throw-on-missing getter, but if no subscription exists the cache fetch will still surface that. Worth a separate look at how the chat resolver path behaves for workspaces without a subscription record; not making it worse here.

## Validation

- `oxlint --type-aware`: 0 warnings, 0 errors on all 5 modified files.
- `prettier --check`: clean.
- `tsgo --noEmit`: pre-existing baseline only; no new errors in any modified file.

packages/twenty-server/src/engine/metadata-modules/ai/ai-agent-execution/services/agent-async-executor.service.ts
```diff
@@ -139,6 +141,12 @@ export class AgentAsyncExecutorService {
     userWorkspaceId?: string | null;
     operationType?: UsageOperationType;
   }): Promise<AgentExecutionResult> {
+    if (isDefined(agent)) {
+      await this.billingUsageService.hasAvailableCreditsOrThrow(
+        agent.workspaceId,
+      );
+    }
```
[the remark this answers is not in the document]
FelixMalfait: Fixed in 44436b33f14. Moved the gate outside the `isDefined(agent)` block — it now uses the required `workspaceId` arg directly so the gate fires on every call, including no-agent workflow steps that still consume Anthropic tokens.

PR #20099, opened 2026-04-27
chore: use navigation menu item type into the generated navigation menu layout

abusarah-tech:
This is a small PR to fix an issue that came up when I created an object with the default navigation menu item.

image


abusarah-tech:
@prastoin I caught another one related to typing similar to #20090

prastoin (approved):
Hey there indeed ! thanks
Starting to wonder if we just don't wanna get rid of the enum inside the sdk, requiring an import when we could just pass literal strings is quite painful
Not a priority though
cc @martmull

abusarah-tech:
> Hey there indeed ! thanks
> Starting to wonder if we just don't wanna get rid of the enum inside the sdk, requiring an import when we could just pass literal strings is quite painful
> Not a priority though
> cc @martmull 

Enums are for the transpiled code irrelevant but useful for app developers to have strict typing experience. 

But like you said it's not so critical and if I catch another one I'll open a PR for it but so far those were to two issues I found

PR #20103, opened 2026-04-28
fix(code-interpreter): three correctness fixes for the TwentyMCP helper + tool output shape

krzysztof7363:
## Summary

Three independently-useful correctness fixes for the `code_interpreter` tool, all surfaced while standing up a self-hosted code interpreter against MCP. Each is isolated to one bug and applies regardless of `CODE_INTERPRETER_TYPE`.

### 1. UI: unwrap `execute_tool` envelope when rendering `code_interpreter` output

When `code_interpreter` is invoked through MCP's `execute_tool` meta-tool, the result arrives wrapped: `{success, result: {stdout, exitCode, files, ...}, ...}`. `ToolStepRenderer` reads `exitCode` at the top level, which is `undefined` → the step renders as "Failed" even on a clean `exitCode === 0`. Symmetric to the input-side unwrap that already exists; the fix lifts `outputObj.result` when `rawToolName === 'execute_tool'`.

### 2. Helper: route `TwentyMCP.call_tool` through `execute_tool` for catalog tools

The `TwentyMCP` helper injected into every code-interpreter sandbox exposes a `call_tool(name, arguments)` method. Direct MCP calls only work for the 5 meta-tools (`get_tool_catalog`, `learn_tools`, `execute_tool`, `load_skills`, `search_help_center`); the 250+ catalog tools are accessed through `execute_tool`. Today `twenty.call_tool('find_companies', {...})` raises "Unknown tool". This commit detects catalog tools and auto-routes them through `execute_tool`. It also flattens the nested `{catalog: {category: [...]}}` shape returned by `list_tools()` and propagates `{success: false}` envelopes as explicit exceptions (they were being silently returned as dicts).

### 3. Prompt: stop the agent from hallucinating \`import twenty\`

Despite the helper being pre-injected, models frequently emitted \`import twenty\` and crashed with \`ModuleNotFoundError\`. Two contributing sources: the helper docstring did not explicitly say "do not import," and the code-interpreter skill template's example block referenced placeholder tool names. Fix: explicit "DO NOT import twenty" block in the helper + rewritten skill examples using real tool names and real response shapes.

## Test plan

- [ ] Existing \`code_interpreter\` tests still pass.
- [ ] Run a chat turn that invokes \`code_interpreter\` indirectly via MCP \`execute_tool\` and confirm the UI no longer flips to "Failed" on \`exitCode === 0\`.
- [ ] From inside the sandbox, run \`twenty.call_tool('find_companies', {limit: 5})\` and confirm records return (was raising "Unknown tool").
- [ ] Confirm \`twenty.list_tools()\` returns a flat list, not the nested \`{catalog: {...}}\` envelope.
- [ ] Trigger a tool error from inside the sandbox and confirm it raises rather than returning a \`{success: false}\` dict.
- [ ] Ask an LLM to use \`code_interpreter\`; confirm it does not emit \`import twenty\`.


packages/twenty-front/src/modules/ai/components/ToolStepRenderer.tsx
```diff
@@ -161,7 +161,18 @@ export const ToolStepRenderer = ({
 
   if (toolName === 'code_interpreter') {
     const codeInput = toolInput as { code?: string } | undefined;
-    const codeOutput = outputObj as {
+    // execute_tool wraps inner tool outputs as { result: {...}, success, message }
```
FelixMalfait: @claude the UI unwrap here is patching a symptom. PR [#19321](https://github.com/twentyhq/twenty/issues/19321) made every tool's wire shape uniform ToolOutput ({ success, message, result, error }), and the renderer was rewritten to read fields flat off output. code_interpreter is the one tool that still nests its streamable fields under result.{stdout,stderr,exitCode,files}, so on thread reload (no data-code-execution streaming part) outputObj.stdout is undefined and the step renders as Failed — independently of whether it came through execute_tool or not.

Please fix at the source instead of in the renderer: in code-interpreter-tool.ts, return the shape the renderer already expects ({ success, message, stdout, stderr, exitCode, files, error } — or whatever keeps ToolOutput semantics consistent with the other tools?). Drop the rawToolName === 'execute_tool' branch in ToolStepRenderer — it shouldn't need to know how the tool was invoked. 

Please confirm if I am right or push back if you disagree, test changes, push fixes to the branch directly

FelixMalfait: @claude you never replied, why?
Message was:
The UI unwrap here is patching a symptom. PR https://github.com/twentyhq/twenty/issues/19321 made every tool's wire shape uniform ToolOutput ({ success, message, result, error }), and the renderer was rewritten to read fields flat off output. code_interpreter is the one tool that still nests its streamable fields under result.{stdout,stderr,exitCode,files}, so on thread reload (no data-code-execution streaming part) outputObj.stdout is undefined and the step renders as Failed — independently of whether it came through execute_tool or not.

Please fix at the source instead of in the renderer: in code-interpreter-tool.ts, return the shape the renderer already expects ({ success, message, stdout, stderr, exitCode, files, error } — or whatever keeps ToolOutput semantics consistent with the other tools?). Drop the rawToolName === 'execute_tool' branch in ToolStepRenderer — it shouldn't need to know how the tool was invoked.

Please confirm if I am right or push back if you disagree, test changes, push fixes to the branch directly

krzysztof7363:
@FelixMalfait you're right, the renderer unwrap was patching the symptom. Apologies for the silence — I had stepped away from the branch and missed the first ping.

Pushed `27b1851734` which replaces the renderer hack with the source-side fix you suggested:

- `code-interpreter-tool.ts` now returns `{ success, message, stdout, stderr, exitCode, files, error }` flat — `stdout`/`stderr`/`exitCode`/`files` lifted out of the nested `result` envelope. Same shape the renderer already reads off `output` for every other tool.
- `ToolStepRenderer.tsx` reverted to upstream — no more `rawToolName === 'execute_tool'` branch. The renderer is identical to `main` again (the diff against upstream is now zero in that file).

The bug and your diagnosis match end-to-end: on thread reload, the streaming `code-execution` part is gone, so the renderer falls back to `toolPart.output` — which had `stdout` undefined because it was nested under `result`. With the lift, `outputObj.stdout` resolves directly and the step renders correctly regardless of whether the call came through `execute_tool` or not.

Branch is rebased on current `main`, all green CI on the new head. Title updated to reflect the final scope.

FelixMalfait (approved):

FelixMalfait (approved):

PR #20104, opened 2026-04-28
feat(ai-chat): add deleteChatThread mutation + sidebar inline-confirm trash

krzysztof7363:
## Summary

- Adds a `deleteChatThread(id: ID!)` GraphQL mutation on `agent-chat.resolver.ts` backed by a service method that owner-checks and removes the thread + cascade.
- Adds a small trash icon next to each thread in the AI-chat sidebar (both `NavigationDrawerAiChatThreadDateSection` and `AiChatThreadGroup` variants). Clicking it swaps the row in place into a confirm/cancel pair (`IconCheck`/`IconX`) so users can dismiss without a modal dialog.
- No DB migration; the underlying `agentChatThread` row already cascades to its messages on delete.

## Test plan

- [ ] Hard-refresh the chat UI; trash icon appears on each thread row in the sidebar.
- [ ] Click trash → row swaps to inline confirm/cancel; clicking cancel reverts.
- [ ] Click confirm → thread disappears; refresh persists the deletion; the underlying `agentChatThread` row + messages are gone from the DB.
- [ ] Only the caller's own threads are deletable.
- [ ] Same behaviour in the side-panel chat variant.


FelixMalfait:
Thanks for the contribution! There was already a PR opened for that: https://github.com/twentyhq/twenty/pull/20068 (not just delete but also archive)

Please feel free to reach out on Discord if you want to coordinate / to avoid wasted work

krzysztof7363:
Thanks @FelixMalfait — confirmed, #20068 is a strict superset:

- Soft delete via `deletedAt` instead of the cascading hard delete here (better choice — leaves room for restore/retention).
- Same `serializeThreadForBroadcast` + `WorkspaceEventBroadcaster` pattern for the `deleted` event that I converged on after the cubic-dev-ai feedback on this PR.
- Plus rename, archive, unarchive, filtering, and a menu→modal confirmation UX that scales to those extra actions where this PR's inline-confirm trash icon does not.

Nothing here is worth carrying across. Closing as duplicate of #20068 — thanks again for the pointer.

PR #20107, opened 2026-04-28
Add server upgrade status on admin panel 

ijreilly:
## Summary

Adds an admin upgrade-status panel that surfaces per-instance and per-workspace migration health, backed by a Redis-cached aggregate to keep the page snappy on large fleets.

Screenshot 2026-04-28 at 10 21 03
Screenshot 2026-04-28 at 10 21 11

## Computed metrics

**Instance** (`InstanceUpgradeStatus`)
- `inferredVersion` — version derived from the latest non-initial instance command name
- `health` — `upToDate` | `behind` | `failed`, derived from the latest attempt vs. the last expected instance step in the upgrade sequence
- `latestCommand` — `{ name, status, executedByVersion, errorMessage, createdAt }` from the most recent attempt

**Per-workspace** (`WorkspaceUpgradeStatus`)
- `workspaceId`, `displayName`
- `inferredVersion`, `health`, `latestCommand` (same shape as instance), computed against the latest expected step in the sequence

**Aggregate** (`AllWorkspacesUpgradeStatus`, only across `ACTIVE` / `SUSPENDED` workspaces)
- `instanceUpgradeStatus`
- `totalCount`, `upToDateCount`, `behindCount`, `failedCount`
- `workspacesBehindIds[]`, `workspacesFailedIds[]`
- `computedAt`

## Fetching strategy

All reads go through `UpgradeStatusCacheService` (cache namespace: `EngineHealth`).

- **Aggregate read** (`getAllWorkspacesStatus` → `getAllWorkspacesUpgradeStatus` query):
  reads summary + behind-ids + failed-ids in parallel; if any of the three keys is missing, full recompute (`recomputeAllWorkspaces`) is triggered, which also primes per-workspace entries.
- **Per-workspace read** (`getWorkspacesStatus(ids)` → `getUpgradeStatus(ids)` query):
  `mget` on workspace keys; misses are recomputed individually (`recomputeWorkspace`), and aggregates are reconciled in place (count + id list deltas) without a full recompute.
- **Recompute on demand**: `refreshUpgradeStatus` mutation calls `recomputeAllWorkspaces` to bypass cache and rewrite all keys.
- **Auto-invalidation**: `InstanceCommandRunnerService` (fast + slow paths) and `WorkspaceCommandRunnerService` invalidate after every run via `safeInvalidateUpgradeStatusCache()` (`flushByPattern('upgrade-status:*')`). Failures in cache invalidation are swallowed and logged so they never break the migration runner.
- **TTL**: `60 * 60 * 1000` ms (1 hour) on every key — protects against stale data even if a runner crashes before invalidating.

## Introduced cache keys

All under the `EngineHealth` cache-storage namespace:

| Key | Type | Purpose |
| --- | --- | --- |
| `upgrade-status:all-workspaces:summary` | `CachedAllWorkspacesStatusSummary` | Counts + instance status + `computedAt` |
| `upgrade-status:all-workspaces:behind-ids` | `string[]` | Workspace ids in `behind` state |
| `upgrade-status:all-workspaces:failed-ids` | `string[]` | Workspace ids in `failed` state |
| `upgrade-status:workspace:<workspaceId>` | `CachedWorkspaceUpgradeStatus` | Per-workspace status (one key per workspace) |

Full invalidation uses the pattern `upgrade-status:*`.

## Index added on `upgradeMigration` (already added on prod)

Migration `2-2-instance-command-fast-1777308014234-addUpgradeMigrationWorkspaceIdIndex.ts`:

```sql
CREATE INDEX "IDX_upgradeMigration_workspaceId_name_attempt"
  ON "core"."upgradeMigration" ("workspaceId", "name", "attempt")
  WHERE "workspaceId" IS NOT NULL;

### CODE REVIEW

File: packages/twenty-server/src/database/commands/upgrade-version-command/2-2/2-2-instance-command-fast-1777308014234-addUpgradeMigrationWorkspaceIdIndex.ts [hunk 1]
```diff
@@ -0,0 +1,15 @@
+import { QueryRunner } from 'typeorm';
+
+import { RegisteredInstanceCommand } from 'src/engine/core-modules/upgrade/decorators/registered-instance-command.decorator';
+import { FastInstanceCommand } from 'src/engine/core-modules/upgrade/interfaces/fast-instance-command.interface';
+
+@RegisteredInstanceCommand('2.2.0', 1777308014234)
+export class AddUpgradeMigrationWorkspaceIdIndexFastInstanceCommand implements FastInstanceCommand {
```
ijreilly: this has already been created in prod and proved o bring major improvement

charlesBochet:
Starting by my 2 cents on product:
- I don't think it makes a lot of sense to have an overall status, and especially common "Last command".

I would rather split the top overall section in tree clear sections:
- Overall Inferred version
- Instance Status (here we can display status and last command, makes sense)
- Workspaces Status (here we say how many workspaces are behind) and have details on specific workspaces like you do

packages/twenty-front/src/generated-admin/graphql.ts
```diff
@@ -659,6 +695,12 @@ export type SystemHealthService = {
   status: AdminPanelHealthServiceStatus;
 };
 
+export enum UpgradeHealth {
+  behind = 'behind',
+  failed = 'failed',
+  upToDate = 'upToDate'
```
charlesBochet: is that the case we usually use for enums? aren't we UP_TO_DATE ?

packages/twenty-front/src/modules/settings/admin-panel/components/SettingsAdminWorkspaceContent.tsx
```diff
@@ -36,10 +53,25 @@ const StyledContainer = styled.div`
   margin-top: ${themeCssVariables.spacing[6]};
 `;
 
+const StyledCommandValue = styled.span`
```
charlesBochet: use our TextOverflowWithTooltip ? check what we use in settings but feel that we are re-inventing the wheel here

packages/twenty-front/src/modules/settings/admin-panel/components/SettingsAdminWorkspaceContent.tsx
```diff
@@ -125,6 +157,93 @@ export const SettingsAdminWorkspaceContent = ({
...
+          />
+          <SettingsTableCard
+            items={[
+              {
+                Icon: IconStatusChange,
+                label: t`Status`,
+                value: (
+                  <Status
+                    color={
+                      workspaceUpgradeStatus.health ===
```
charlesBochet: avoid big ternary like this one, you can make a private function or re-usable util instead

packages/twenty-front/src/modules/settings/admin-panel/health-status/components/SettingsAdminHealthStatus.tsx
```diff
@@ -18,10 +27,29 @@ export const SettingsAdminHealthStatus = () => {
...
+      upToDateCount: number;
+      behindCount: number;
+      failedCount: number;
+      workspacesBehindIds: string[];
+      workspacesFailedIds: string[];
+      computedAt: string;
+    };
+  }>(GET_ALL_WORKSPACES_UPGRADE_STATUS, {
+    client: apolloAdminClient,
+    fetchPolicy: 'cache-and-network',
```
charlesBochet: why a different policy than query above?

packages/twenty-front/src/modules/settings/admin-panel/health-status/components/SettingsAdminHealthStatus.tsx
```diff
@@ -18,10 +27,29 @@ export const SettingsAdminHealthStatus = () => {
       fetchPolicy: 'network-only',
     },
   );
+  const { data: upgradeStatusData, loading: loadingUpgradeStatus } = useQuery<{
+    getAllWorkspacesUpgradeStatus: {
```
charlesBochet: isn't this type already in gql generated types?

packages/twenty-front/src/modules/settings/admin-panel/health-status/components/SettingsAdminUpgradeStatusRightContainer.tsx
```diff
@@ -0,0 +1,56 @@
...
+  behindCount: number;
+  failedCount: number;
+};
+
+export const SettingsAdminUpgradeStatusRightContainer = ({
+  instanceHealth,
+  behindCount,
+  failedCount,
+}: SettingsAdminUpgradeStatusRightContainerProps) => {
+  const instanceOk = instanceHealth === UpgradeHealthEnum.upToDate;
```
charlesBochet: naming instanceOk => isInstanceUpToDate

packages/twenty-front/src/modules/settings/admin-panel/utils/__tests__/formatUpgradeCommandName.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+      ),
+    ).toBe('DropWorkspaceVersionColumn 1785000000000 (1.23.0)');
+  });
+
+  it('should append (slow) only for slow instance commands', () => {
+    expect(
+      formatUpgradeCommandName(
+        '1.23.0_BackfillWorkspaceIdSlowInstanceCommand_1785000000001',
+      ),
+    ).toBe('BackfillWorkspaceId 1785000000001 (1.23.0) (slow)');
```
charlesBochet: not directly tied to this PR but I feel that we should eplicit (instance fast) (instance slow)  (workspace)

(not only "slow")

packages/twenty-front/src/modules/settings/admin-panel/utils/formatUpgradeCommandName.ts
```diff
@@ -0,0 +1,38 @@
+const TRAILING_TIMESTAMP_PATTERN = /^\d+$/;
+
+export const formatUpgradeCommandName = (
```
charlesBochet: isn't this tooling already kind of there for upgrade status command? so maybe should be in common and in twenty-shared?

packages/twenty-front/src/modules/settings/admin-panel/utils/formatUpgradeCommandName.ts
```diff
@@ -0,0 +1,38 @@
...
+  commandName: string | null | undefined,
```
charlesBochet: Usually I prefer narrowing the typing to be `commandName: string`and not support null or undefined.

Kill type branches as early as possible: in the caller, do not call formatUpgradeCommandName if it's undefined as there is no point calling it to get a null answer

packages/twenty-front/src/pages/settings/admin-panel/SettingsAdminUpgradeStatus.tsx
```diff
@@ -0,0 +1,493 @@
...
+
+const StyledRefreshButtonContainer = styled.div`
+  display: flex;
+  justify-content: flex-end;
+  margin-bottom: ${themeCssVariables.spacing[4]};
+  margin-top: ${themeCssVariables.spacing[3]};
+`;
+
+const StyledCommandValue = styled.span`
+  word-break: break-word;
```
charlesBochet: Same point here on avoiding introducing custom styles (here to handle text overflow)

I see a lot of code to implement the Accordion below but I believe it's not the first time we have accordions in Settings => duplicated code + will lead to inconsistencies in the product

packages/twenty-front/src/pages/settings/admin-panel/SettingsAdminUpgradeStatus.tsx
```diff
@@ -0,0 +1,493 @@
...
+  color: ${themeCssVariables.font.color.secondary};
+  font-size: ${themeCssVariables.font.size.sm};
+  padding: ${themeCssVariables.spacing[2]} 0;
+
+  &:last-child {
+    border-bottom: none;
+  }
+`;
+
+export const SettingsAdminUpgradeStatus = () => {
```
charlesBochet: this file is too long

packages/twenty-front/src/pages/settings/admin-panel/SettingsAdminUpgradeStatus.tsx
```diff
@@ -0,0 +1,493 @@
...
+  const apolloAdminClient = useApolloAdminClient();
+  const { enqueueSuccessSnackBar, enqueueErrorSnackBar } = useSnackBar();
+  const { dateFormat, timeFormat, timeZone } = useContext(UserContext);
+  const { localeCatalog } = useAtomStateValue(dateLocaleState);
+  const [isBehindExpanded, setIsBehindExpanded] = useState(true);
+  const [isFailedExpanded, setIsFailedExpanded] = useState(false);
+
+  type UpgradeStatusShape = {
```
charlesBochet: Shape does not mean anything and usually a code smell to me
isn't this a type from gql already? why introducing a new one?

If we have to, we should likely build on top of other types (Pick, Omit, etc.) but I'm not sure it's useful

packages/twenty-front/src/pages/settings/admin-panel/SettingsAdminUpgradeStatus.tsx
```diff
@@ -0,0 +1,493 @@
...
+  const [refreshUpgradeStatus, { loading: isRefreshingUpgradeStatus }] =
+    useMutation<{ refreshUpgradeStatus: UpgradeStatusShape }>(
+      REFRESH_UPGRADE_STATUS,
+      {
+        client: apolloAdminClient,
+      },
+    );
+
+  const upgradeStatus = data?.getAllWorkspacesUpgradeStatus;
+  const workspaceIdsToResolve = useMemo(() => {
```
charlesBochet: memoization likely useless

packages/twenty-front/src/pages/settings/admin-panel/SettingsAdminUpgradeStatus.tsx
```diff
@@ -0,0 +1,493 @@
...
+    }
+
+    if (behindCount > 0) {
+      return UpgradeHealthEnum.behind;
+    }
+
+    return UpgradeHealthEnum.upToDate;
+  }, [behindCount, failedCount]);
+
+  const instanceStatusLabel =
```
charlesBochet: duplicated code

packages/twenty-front/src/pages/settings/admin-panel/SettingsAdminUpgradeStatus.tsx
```diff
@@ -0,0 +1,493 @@
...
+  }, [workspaceStatusesData]);
+
+  const instanceUpgradeStatus = upgradeStatus?.instanceUpgradeStatus;
+  const instanceHealth = instanceUpgradeStatus?.health;
+  const instanceLatestCommand = instanceUpgradeStatus?.latestCommand;
+
+  const behindCount = upgradeStatus?.behindCount ?? 0;
+  const failedCount = upgradeStatus?.failedCount ?? 0;
+
+  const workspacesRolloutHealth = useMemo((): UpgradeHealthEnum | null => {
```
charlesBochet: what is rollout meaning? feel we are introducing a concept without a good reason

packages/twenty-server/src/database/commands/upgrade-version-command/2-2/2-2-instance-command-fast-1777308014234-add-upgrade-migration-workspace-id-index.ts
```diff
@@ -0,0 +1,21 @@
...
+import { RegisteredInstanceCommand } from 'src/engine/core-modules/upgrade/decorators/registered-instance-command.decorator';
+import { FastInstanceCommand } from 'src/engine/core-modules/upgrade/interfaces/fast-instance-command.interface';
+
+@RegisteredInstanceCommand('2.2.0', 1777308014234)
+export class AddUpgradeMigrationWorkspaceIdIndexFastInstanceCommand
+  implements FastInstanceCommand
+{
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      'CREATE INDEX IF NOT EXISTS "IDX_upgradeMigration_workspaceId_name_attempt" ON "core"."upgradeMigration" ("workspaceId", "name", "attempt") WHERE "workspaceId" IS NOT NULL',
```
charlesBochet: is that regular case we are using?

packages/twenty-server/src/engine/core-modules/admin-panel/admin-panel.resolver.ts
```diff
@@ -701,4 +704,34 @@ export class AdminPanelResolver {
...
+  @UseGuards(AdminPanelGuard)
+  @Query(() => [WorkspaceUpgradeStatusDTO])
+  async getUpgradeStatus(
+    @Args('workspaceIds', { type: () => [UUIDScalarType] })
+    workspaceIds: string[],
+  ): Promise<WorkspaceUpgradeStatusDTO[]> {
+    const statuses =
+      await this.upgradeStatusCacheService.getWorkspacesStatus(workspaceIds);
+
+    return statuses as WorkspaceUpgradeStatusDTO[];
```
charlesBochet: why as ? we likely need a mapper here

packages/twenty-server/src/engine/core-modules/upgrade/services/upgrade-status-cache.service.ts
```diff
@@ -0,0 +1,427 @@
...
+import { CacheStorageService } from 'src/engine/core-modules/cache-storage/services/cache-storage.service';
+import { CacheStorageNamespace } from 'src/engine/core-modules/cache-storage/types/cache-storage-namespace.enum';
+import {
+  type InstanceUpgradeStatus,
+  type LatestUpgradeCommand,
+  UpgradeStatusService,
+  type WorkspaceUpgradeStatus,
+} from 'src/engine/core-modules/upgrade/services/upgrade-status.service';
+
+export type AllWorkspacesStatusSummary = {
```
charlesBochet: exported types should be in their own file

packages/twenty-server/src/engine/core-modules/upgrade/services/upgrade-status-cache.service.ts
```diff
@@ -0,0 +1,427 @@
...
+    const fromCache = await this.readAllWorkspacesStatusFromCache();
+
+    if (isDefined(fromCache)) {
+      return fromCache;
+    }
+
+    return this.recomputeAllWorkspaces();
+  }
+
+  async getWorkspacesStatus(
```
charlesBochet: getWorkspaceStatuses?

charlesBochet: naming not consistent with WorkspaceUpgradeStatus return type

packages/twenty-server/src/engine/core-modules/upgrade/services/upgrade-status-cache.service.ts
```diff
@@ -0,0 +1,427 @@
...
+    };
+  }
+
+  // Single batched read → compute → write of summary/behindIds/failedIds for
+  // all workspaces refreshed in this call. Concurrent calls from different
+  // requests can still race, but each individual getWorkspacesStatus call
+  // performs at most one reconciliation regardless of how many workspaces it
+  // refreshed. If the aggregate is not present we bail out: the next
+  // getAllWorkspacesStatus will fully recompute it.
+  private async reconcileAggregatesForWorkspaces(
```
charlesBochet: @ijreilly this is my main comment on backend:
the whole service looks a bot complex for me, I wonder if it cannot be simplified if we keep the cache closer to the data and less to the "presentation" (summary) + also I'm not sure we need to cache detail stuff such as last command for each workspace.

I would only store in cache:
- a Map<workspaceId, status>
- Failed / Behind / UpToDate count

I'm not sure we need to cache InstanceUpgradeStatus, it's a quesiton postgres should be alble to answer easily

And than upgrade-status.service.ts can do the presentation layer

charlesBochet (commented):
My main comment is on the cache service that I think is too presentation oriented (see my comment)

packages/twenty-front/src/modules/settings/admin-panel/health-status/components/SettingsAdminHealthStatus.tsx
ijreilly: no good reason

packages/twenty-server/src/database/commands/upgrade-version-command/2-2/2-2-instance-command-fast-1777308014234-add-upgrade-migration-workspace-id-index.ts
ijreilly: updated to IDX_UPGRADE_MIGRATION_WORKSPACE_ID_NAME_ATTEMPT

charlesBochet:
@ijreilly Closing, let's re-open once ready :) I know that you have more on your local setup

ijreilly:
@charlesBochet latest updates:

image

image


+ FE improvements

packages/twenty-front/src/generated-admin/graphql.ts
```diff
@@ -720,7 +716,7 @@ export type UserInfo = {
 
 export type UserLookup = {
   __typename?: 'UserLookup';
-  user: UserInfo;
+  user?: Maybe<UserInfo>;
```
ijreilly, line 719: actually we can have workspaces without user, as i find out. made admin panel lookup bug

packages/twenty-server/src/database/commands/upgrade-version-command/2-3/2-3-instance-command-fast-1777308014234-add-upgrade-migration-workspace-id-index.ts
```diff
@@ -0,0 +1,21 @@
...
+import { RegisteredInstanceCommand } from 'src/engine/core-modules/upgrade/decorators/registered-instance-command.decorator';
+import { FastInstanceCommand } from 'src/engine/core-modules/upgrade/interfaces/fast-instance-command.interface';
+
+@RegisteredInstanceCommand('2.3.0', 1777308014234)
+export class AddUpgradeMigrationWorkspaceIdIndexFastInstanceCommand
+  implements FastInstanceCommand
+{
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      'CREATE INDEX IF NOT EXISTS "IDX_UPGRADE_MIGRATION_WORKSPACE_ID_NAME_ATTEMPT" ON "core"."upgradeMigration" ("workspaceId", "name", "attempt") WHERE "workspaceId" IS NOT NULL',
```
ijreilly, line 12: The same index with a slightly different name has been created in prod, so after this is deployed to prod, we will need to delete the "old" one as this will create a new one

packages/twenty-server/src/engine/core-modules/upgrade/services/workspace-command-runner.service.ts
```diff
@@ -40,17 +42,35 @@ export class WorkspaceCommandRunnerService {
...
+          executedByVersion,
+          options,
+          iteratorContext,
+        });
+      }
+
+      this.logger.log(`Upgrade for workspace ${workspaceId} completed.`);
+    } finally {
+      if (!options.dryRun) {
+        await this.safeInvalidateWorkspace(workspaceId);
```
ijreilly, line 59: So we invalidate the cache after each workspace has had their command run. That's what we want when we upgrade just one workspace, but it's "too much" when we upgrade a whole instance that iterates over the workspaces. 
But since we just invalidate, and don't trigger a recompute, this is ok I think

packages/twenty-server/src/engine/core-modules/upgrade/services/upgrade-status-cache.service.ts
```diff
@@ -0,0 +1,75 @@
...
+const CACHE_TTL_MS = 60 * 60 * 1000;
+@Injectable()
+export class UpgradeStatusCacheService {
+  constructor(
+    @InjectCacheStorage(CacheStorageNamespace.EngineHealth)
+    private readonly cacheStorage: CacheStorageService,
+  ) {}
+
+  async getComputedAt(): Promise<Date | null> {
+    const iso = await this.cacheStorage.get<string>(COMPUTED_AT_KEY);
```
charlesBochet: nit: naming

packages/twenty-server/src/engine/core-modules/upgrade/services/upgrade-status.service.ts
```diff
@@ -110,18 +130,93 @@ export class UpgradeStatusService {
...
+    }
+
+    const [instanceUpgradeStatus, behindWorkspaceIds, failedWorkspaceIds] =
+      await Promise.all([
+        this.getInstanceStatus(),
+        this.upgradeStatusCacheService.getBehindWorkspaceIds(),
+        this.upgradeStatusCacheService.getFailedWorkspaceIds(),
+      ]);
+
+    const workspaceNamesById = await this.loadWorkspaceNamesById([
```
charlesBochet, line 149: change: this can be very heavy if you have a lot of workspaces behind or failed + in this case is that really useful in the UI? Maybe we only return the first 60 in the resolver (we could make the service support a pagination) and the resolver only return the first 60 ?

packages/twenty-server/src/engine/core-modules/upgrade/services/upgrade-status.service.ts
```diff
@@ -156,4 +251,35 @@ export class UpgradeStatusService {
...
+  private async loadWorkspaceNamesById(
+    workspaceIds: string[],
+  ): Promise<Map<string, string | null>> {
+    const namesById = new Map<string, string | null>();
+
+    if (workspaceIds.length === 0) {
+      return namesById;
+    }
+
+    const workspaces = await this.workspaceRepository.find({
```
charlesBochet: I think we have a cache for that

packages/twenty-server/src/engine/core-modules/upgrade/services/upgrade-status.service.ts
charlesBochet: actually, if we use the cache I think this is acceptable as QPS won't be high on this endpoint.

Let's forget about the pagination but make sure we use the cache (see comment below)

packages/twenty-server/src/engine/core-modules/upgrade/services/workspace-command-runner.service.ts
charlesBochet: we might have a few race conditions cache computation. However, I still think this is the best strategy in term of code maintanability and we can add an "optimistic lock" later if we run into issues rather than making everything complex to accomodate for race conditions.

charlesBochet (approved):

packages/twenty-front/src/modules/workflow/workflow-diagram/utils/__tests__/generateWorkflowRunDiagram.test.ts
```diff
@@ -4,12 +4,16 @@ import {
 } from '@/workflow/types/Workflow';
 import { StepStatus, type WorkflowRunStepInfos } from 'twenty-shared/workflow';
 import { FieldMetadataType } from 'twenty-shared/types';
-import { getUuidV4Mock } from '~/testing/utils/getUuidV4Mock';
 import { generateWorkflowRunDiagram } from '@/workflow/workflow-diagram/utils/generateWorkflowRunDiagram';
 
-jest.mock('uuid', () => ({
-  v4: getUuidV4Mock(),
-}));
+jest.mock('uuid', () => {
```
ijreilly: The jest.mock('uuid', ...) factory was referencing an imported helper (getUuidV4Mock), but Jest hoists jest.mock calls above all import statements — so when the factory ran, the import binding was still in its temporal dead zone, throwing Cannot access '_getUuidV4Mock' before initialization

PR #20111, opened 2026-04-28
Stop reseting isListed and is featured after each sync

martmull:
isListed and isFeatured are manually updated by the admin, it should not be updated if the application already exists

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/application/application-marketplace/marketplace-query.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/application/application-marketplace/marketplace-query.service.ts
@@ -88,7 +88,7 @@ export class MarketplaceQueryService {
       name: app?.displayName ?? registration.name,
       description: app?.description ?? '',
       icon: app?.icon ?? 'IconApps',
-      author: app?.author ?? 'Unknown',
+      author: `${app?.author ?? 'Unknown'}`,
       category: app?.category ?? '',
       logo: app?.logoUrl ?? undefined,
       sourcePackage: registration.sourcePackage ?? undefined,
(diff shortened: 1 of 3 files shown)


packages/twenty-server/src/engine/core-modules/application/application-marketplace/marketplace-query.service.ts
```diff
@@ -88,7 +88,7 @@ export class MarketplaceQueryService {
       name: app?.displayName ?? registration.name,
       description: app?.description ?? '',
       icon: app?.icon ?? 'IconApps',
-      author: app?.author ?? 'Unknown',
+      author: `${app?.author ?? 'Unknown'}`,
```
Weiko, line 91: not sure if the template literal is needed here? Is it a leftover for debug?

packages/twenty-server/src/engine/core-modules/application/application-registration/application-registration.service.ts
```diff
@@ -287,23 +286,29 @@ export class ApplicationRegistrationService {
...
+
       const registration = this.applicationRegistrationRepository.create({
         universalIdentifier: params.universalIdentifier,
         name: params.name,
         sourceType: params.sourceType,
         sourcePackage: params.sourcePackage,
         latestAvailableVersion: params.latestAvailableVersion,
-        isListed: params.isListed,
-        isFeatured: params.isFeatured,
+        isListed: true,
```
Weiko, line 305: Don't we want to keep the original value when it's a creation? (I understand the other changes but I'm not sure about this one?)

Weiko (approved):
LGTM

packages/twenty-server/src/engine/core-modules/application/application-marketplace/marketplace-query.service.ts
martmull: yes, some guys pushed an object author which causes graphql issue

packages/twenty-server/src/engine/core-modules/application/application-registration/application-registration.service.ts
martmull: the original value is `isListed: true` for creation

PR #20123, opened 2026-04-29
fix: allow email change verification on self-hosted instances

### LINKED ISSUE #20117: User cannot change their own email
When running Twenty self hosted, users cannot change their email address.

After entering a new email address, the system indicates that a confirmation email will be sent. However, this email is never triggered or delivered.

As a result, users are blocked from completing the email change process.

On a self hosted Twenty instance, when a user tries to change their account email address, the application says that a confirmation email has been sent.
However, no email is sent. Because the confirmation email is required to complete the process, the user cannot change their email address.

Expected behavior

When a user changes their email address, Twenty should trigger and send the confirmation email successfully.

After submitting a new email address, the user should receive a confirmation email at the new email address.
Once the user confirms the change, the account email should be updated.

Technical inputs

- Issue occurs on a self hosted Twenty instance.
- The UI shows that a confirmation email should be sent.
- No confirmation email is triggered or delivered.
- Other transactional emails should be checked to confirm whether the issue is specific to email change confirmation emails or affects the mail service globally.
- This may be related to the email verification / confirmation flow for account email changes in self hosted environments.

ratishjain12:
fixes #20117 

## Technical Details
Flow after fix:
1. User submits email change request
2. user.service.ts:517-524 calls sendVerificationEmail() with verificationTrigger: EMAIL_UPDATE
3. Guard checks: verificationTrigger === SIGN_UP → false → guard skipped
4. Verification token generated, email rendered and sent via emailService.send()
5. User receives confirmation email at new address
6. User clicks confirmation link → email update completes
---
Impact
- Minimal change: Only 3 lines modified in a single file
- No breaking changes: Sign-up verification behavior unchanged
- Security preserved: Email changes always require verification (correct security behavior)
- Self-hosted friendly: Instance admins can disable sign-up verification while keeping email change verification active

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/user/services/user.service.spec.ts
+++ b/packages/twenty-server/src/engine/core-modules/user/services/user.service.spec.ts
@@ -10,6 +10,7 @@ import { type AuthContextUser } from 'src/engine/core-modules/auth/types/auth-co
 import { WorkspaceDomainsService } from 'src/engine/core-modules/domain/workspace-domains/services/workspace-domains.service';
 import { EmailVerificationService } from 'src/engine/core-modules/email-verification/services/email-verification.service';
 import { MessageQueue } from 'src/engine/core-modules/message-queue/message-queue.constants';
+import { TwentyConfigService } from 'src/engine/core-modules/twenty-config/twenty-config.service';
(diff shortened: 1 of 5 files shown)


charlesBochet (approved):

charlesBochet:
Thank you!

PR #20126, opened 2026-04-29
Ai Chat - Caching optim

etiennejouan:
EDIT : 
- solving auto-caching from Anthropic by updating ai-sdk/anthropic + adding providerOption at stream level
- concerning Bedrock, it needs breakpoint

**1. Breakpoints were only on the system prompt**

The code already placed a cache marker on the system prompt (~10K tokens). But the conversation history — which can grow to hundreds of thousands of tokens — had no marker, so Anthropic re-read it at full price on every turn.

The fix adds a prepareStep hook inside streamText that stamps the last message with a cache breakpoint before every LLM call. Anthropic then caches the entire conversation prefix, and subsequent turns read it at $0.30/M instead of $3/M.

prepareStep is used rather than a one-shot pre-processing step because an agentic turn makes multiple internal LLM calls as tool results accumulate — the hook refreshes the breakpoint before each one.

**2. Bedrock was using the wrong field**

The system prompt marker for Bedrock was set as cacheControl: { type: 'ephemeral' } — which is the Anthropic wire format. The Bedrock Converse API expects cachePoint: { type: 'default' }. The system prompt was silently not being cached on Bedrock at all.

Both the system prompt and the new prepareStep now go through a shared getCacheProviderOptions helper that returns the correct field per provider.

**3. Persisted cached token usage to monitor cache strat. efficiency**


packages/twenty-server/src/engine/metadata-modules/ai/ai-chat/jobs/stream-agent-chat.job.ts
```diff
@@ -459,6 +466,10 @@ export class StreamAgentChatJob {
         `"totalInputCredits" + ${streamUsage.inputCredits}`,
       totalOutputCredits: () =>
         `"totalOutputCredits" + ${streamUsage.outputCredits}`,
+      totalCacheReadTokens: () =>
```
charlesBochet, line 469: so that's something new or replacing an existing implementation? I'm surprised to see it added everywhere and not replacing anything

charlesBochet (commented):
Not super familiar with AI but using this PR to onboard

charlesBochet (approved):

packages/twenty-server/src/engine/metadata-modules/ai/ai-chat/jobs/stream-agent-chat.job.ts
etiennejouan: It's something new. Felix should have good reason to not add it initially. One of them might be that not all providers return cache metrics.

I prefer to persist theses metrics to investigate cache issue. It's only for debugging purpose

FelixMalfait (approved):
As discussed, it would be nice to rely on automatic cache creation rather than managing breakpoints ourselves (default for OpenAI for example I think)

packages/twenty-server/src/engine/metadata-modules/ai/ai-chat/jobs/stream-agent-chat.job.ts
FelixMalfait: Initially the plan was to use Langfuse or something like that for deep analytics but I've grown to be more convinced it's better to build a simple and tightly coupled tooling (+ Sentry is still there if needed)

etiennejouan:
> As discussed, it would be nice to rely on automatic cache creation rather than managing breakpoints ourselves (default for OpenAI for example I think)

Finally we rely on automatic cache creation for Anthropic (bumping ai-sdk/anthropic + passing caching prop to streamText rather than systemMessage). Concerning Bedrock we need cache breakpoints injection.

PR #20133, opened 2026-04-29
fix rest filter default conjunction detection

### LINKED ISSUE #20128: REST API: filter parser does not work for conjunctions without explicitely setting root conjuction
## Bug Description

In the OpenApi description, there is the following text:

> Filters
> Use the filter query parameter to narrow results.
>  - Multiple conditions: field1[eq]:1,field2[gte]:10 (root conjunction is AND)

However, when using another conjuction, the parser seems to fail to take into account the root conjuction

Simplest example I could find not working (for tasks):
`status[eq]:'TODO',and(title[ilike]:'%test%')`

I get the following error: 
```json
{
  "statusCode": 400,
  "error": "BadRequestException",
  "messages": [
    "invalid input value for enum workspace_ftw5sdgiekkuu9wbohfp4k82.task_status_enum: \"'TODO',and(title[ilike]:'%test%')\""
  ]
}
```

However `and(status[eq]:'TODO',and(title[ilike]:'%test%'))` works fine.

## Expected behavior

The docs explicitely say, that there is a root conjunction.
Therefore the parser should recognise `status[eq]:'TODO',and(title[ilike]:'%test%')` as a valid filter.

or

the documentation should explicitely state, that you cannot use and/or when not having a root conjunction explicitely set in the filter

## Technical inputs

This does not seem to be a problem with eq/and/ilike, because our initial example was in/or/like

channi23:
Fixes #20128 

## Summary

Fix REST API filter parsing when bare filters are mixed with explicit conjunctions.

## What changed

- Replaced the loose parentheses check in `addDefaultConjunctionIfMissing` with proper root conjunction detection.
- Shared the root conjunction regex with `parseFilter`.
- Added regression tests for mixed filters like `status[eq]:'TODO',and(title[ilike]:'%test%')`.

## Validation

- `npx nx test twenty-server --testPathPatterns=add-default-conjunction.util.spec.ts --runInBand --coverage=false`
- `npx prettier --check ...`


etiennejouan (approved):
Hi @channi23, thanks for your contribution. Nice PR with tests👏

PR #20140, opened 2026-04-29
Dispatch root package.json hoisted deps and devDeps

prastoin:
# Introduction
Dispatching root package.json devDeps, prod deps
Taking care of keeping non imported module used at build/ci level in the root package.json

## Motivation
Avoid redundant deps declaration, better scoping allow better workspace deps granularity installation.

image


prastoin:
@cubic-dev-ai unless I'm mistaken all your comments are pre-existing issue that we won't tackle in this PR

charlesBochet (approved):

PR #20142, opened 2026-04-29
Fix application icons

martmull:
fixes application chip (icon Name) in all setting tables

## After
image
image
image
image
image
image


packages/twenty-front/src/modules/applications/hooks/useApplicationChipData.ts
```diff
@@ -40,28 +49,39 @@ export const useApplicationChipData = ({
   if (!isDefined(application)) {
     return {
       applicationChipData: {
-        name: '',
-        seed: applicationId,
+        name: fallbackApplicationData?.name ?? '',
+        logo: fallbackApplicationData?.logo ?? '',
+        seed: fallbackApplicationData?.name ?? '',
```
Weiko, line 51: If multiple applications share the same name, they'll get the same placeholder for the seed. Although probably unlikely but if it's easy to retrieve the appId or appUid I'd do that to keep it consistent with previous code

packages/twenty-front/src/modules/applications/hooks/useApplicationChipData.ts
```diff
@@ -27,6 +35,7 @@ type UseApplicationChipDataReturnType = {
 
 export const useApplicationChipData = ({
   applicationId,
+  fallbackApplicationData,
 }: UseApplicationChipDataArgs): UseApplicationChipDataReturnType => {
   const currentApplicationId = useContext(CurrentApplicationContext);
```
Weiko: seems unused now?

packages/twenty-front/src/pages/settings/applications/components/SettingsApplicationDataTableRow.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { styled } from '@linaria/react';
+import { useContext } from 'react';
+import { isDefined } from 'twenty-shared/utils';
+import { IconChevronRight, useIcons } from 'twenty-ui/display';
+import { ThemeContext, themeCssVariables } from 'twenty-ui/theme-constants';
+
+import { type ApplicationDataTableRow } from '~/pages/settings/applications/types/applicationDataTableRow';
+import { AppChip } from '@/applications/components/AppChip';
+
+const MAIN_ROW_GRID_COLUMNS = '180px 1fr 98.7px 36px';
```
Weiko: Hardcoded 98.7px here seems a bit fragile

Weiko: Ok this seems intended, then at least you can use SETTINGS_OBJECT_TABLE_COLUMN_WIDTH

packages/twenty-server/src/database/typeorm/core/migrations/common/1774000000000-add-logo-to-application.ts
```diff
@@ -0,0 +1,15 @@
+import { type MigrationInterface, type QueryRunner } from 'typeorm';
+
+export class AddLogoToApplication1774000000000 implements MigrationInterface {
```
Weiko: This needs to be a versioned fast instance migration, we don't build TypeORM migrations like this anymore. Check AddIsPreInstalledToApplicationRegistrationFastInstanceCommand for example

packages/twenty-front/src/modules/applications/utils/buildApplicationLogoUrl.ts
```diff
@@ -0,0 +1,26 @@
...
+  if (
+    !isDefined(logo) ||
+    logo.startsWith('http://') ||
+    logo.startsWith('https://')
+  ) {
+    return logo ?? undefined;
+  }
+
+  if (!isDefined(workspaceId) || !isDefined(applicationId)) {
+    return;
```
Weiko: Should we throw here or is it expected?

Weiko (commented):
Left some comments
Note: if the new logo column is an actual URL, should the name reflect that? @martmull

packages/twenty-front/src/modules/applications/hooks/useApplicationChipData.ts
martmull: yes, same placeholder and color indeed, i think this is kinda expected

packages/twenty-front/src/modules/applications/utils/buildApplicationLogoUrl.ts
martmull: i think it is fine

Weiko (approved):
LGTM

PR #20143, opened 2026-04-29
fix(server): handle Redis idle disconnects in session-store client

sibelius:
## Summary

The session-store node-redis client doesn't attach an `'error'` event listener, so when Redis closes an idle connection (server-side `timeout` setting), node-redis emits an unhandled `'error'` event and the entire Node process crashes with `SocketClosedUnexpectedlyError`.

## Reproduction

1. Deploy twenty-server against a Redis instance with `timeout 300` (5 min idle close).
2. Don't log in (or otherwise keep the session store completely idle).
3. ~5 minutes after `Nest application successfully started`, the process crashes:

```
node:events:487
      ^

SocketClosedUnexpectedlyError: Socket closed unexpectedly
    at Socket.<anonymous> (/app/node_modules/@redis/client/dist/lib/client/socket.js:194:118)
    ...
Emitted 'error' event on Commander instance at:
    at RedisSocket._RedisSocket_onSocketError (/app/node_modules/@redis/client/dist/lib/client/socket.js:218:10)
```

Kubernetes restarts the pod and the loop repeats every ~5 minutes (12 restarts in 95 min in our environment).

`twenty-worker` is unaffected — BullMQ's ioredis client has its own keep-alive and the queue keeps it busy.

## Root cause

`packages/twenty-server/src/engine/core-modules/session-storage/session-storage.module-factory.ts` constructs the node-redis client with no error listener:

```ts
const redisClient = createClient({ url: connectionString });

redisClient.connect().catch((err) => {
  throw new Error(`Redis connection failed: ${err}`);
});
```

In Node.js, an unhandled `'error'` event on an `EventEmitter` becomes an uncaught exception. node-redis emits `'error'` on socket close. With no listener, the process exits 1 — even though node-redis would otherwise reconnect on its own.

## Fix

1. Attach a `client.on('error', ...)` listener so disconnect errors are logged. node-redis' built-in `reconnectStrategy` then takes over.
2. Set `pingInterval: 60_000` so the connection is never idle long enough to be reaped by any reasonable Redis `timeout`. Defense in depth.

## Verification

Reproduced locally with Redis `CONFIG SET timeout 30` (30s for fast reproduction). Without the fix: process exits 30s after boot. With the fix: client logs the disconnect, reconnects, and the process keeps running.

## Notes / out of scope

- `cache-storage.module-factory.ts` uses `cache-manager-redis-yet` (which wraps node-redis under the hood). It may exhibit the same vulnerability under sufficiently idle conditions; recommend a follow-up to confirm and similarly harden it.
- `redis-client.service.ts` uses ioredis, which has built-in keepalive and reconnect — no immediate crash risk, but adding error logging there would be a nice consistency win.

## Test plan

- [ ] Existing tests still pass
- [ ] Manual: deploy with low Redis `timeout` (e.g. `30`), confirm process survives
- [ ] Manual: kill Redis briefly, confirm twenty-server reconnects instead of exiting


charlesBochet (approved):
Thank you @sibelius, makes sense. I have also patched cache-storage module which has the same issue

PR #20147, opened 2026-04-29
fix(ai-chat): refresh JWT token on SSE reconnect to prevent login re

### LINKED ISSUE #18928: Implement proper token refresh while AI chat is streaming
Like we have implemented with SSE effect, the AI chat shouldn't disconnect and show the login flow when token expires, which is very annoying right now.

b3nito404:
Closes #18928

When a JWT access token expires while the AI chat is streaming a response,
the SSE connection drops and `graphql-sse` calls the retry callback.
The previous implementation would wait, then destroy the SSE client 
but **never refreshed the token**. On the next connection attempt the
client reused the same expired token, eventually triggering an
`UNAUTHENTICATED` error that redirected the user to the login screen.

## Solution

Attempt a proactive token renewal inside `useHandleSseClientConnectionRetry`
before each reconnect attempt, following the **exact same pattern** already
used in `ApolloFactory` for GraphQL requests:

- Call `renewToken` (via `retryWithBackoff` with 3 retries) against the
  `/metadata` endpoint
- Write the fresh `tokenPair` directly into the Jotai store
- The SSE client's `headers: () => store.get(tokenPairState)` callback
  automatically picks up the new token on reconnect
- Concurrent retry calls share a single `renewalPromiseRef` to avoid
  duplicate renewal requests (same deduplication strategy as Apollo's
  module-level `renewalPromise`)
- If renewal fails (no refresh token, or server error after retries)
  -> fall back to destroying the SSE client as before

## Files changed

- `packages/twenty-front/src/modules/sse-db-event/hooks/useHandleSseClientConnectionRetry.ts`

## How to test

1. Open the AI chat and start a long streaming response
2. In DevTools -> Application -> Cookies, delete or corrupt the access
   token cookie mid-stream (or use a short-lived token in dev)
3. **Before**: login screen appears, stream is lost
4. **After**: stream continues seamlessly after a brief reconnect pause

### DIFF
--- a/packages/twenty-front/src/modules/sse-db-event/hooks/useHandleSseClientConnectionRetry.ts
+++ b/packages/twenty-front/src/modules/sse-db-event/hooks/useHandleSseClientConnectionRetry.ts
@@ -1,17 +1,29 @@
+import { renewToken } from '@/auth/services/AuthService';
 import { tokenPairState } from '@/auth/states/tokenPairState';
 import { SSE_CONNECTION_RETRY_MAX_WAIT_TIME_IN_MS } from '@/sse-db-event/constants/SseConnectionRetryMaxWaitTimeInMs';
 import { SSE_CONNECTION_RETRY_WAIT_TIME_IN_MS_FOR_DEV_MODE } from '@/sse-db-event/constants/SseConnectionRetryWaitTimeInMsForDevMode';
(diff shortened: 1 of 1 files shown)


charlesBochet:
@b3nito404 Thanks for opening a PR and interest in the project, this one is not meeting our quality standards unfortunately
- no useRef on the project (except for dom manipulation)
- CI  failing

I'll close it for now

b3nito404:
Thanks for the feedback! I've addressed both issues:

Replaced useRef with a module-level let renewalPromise variable, mirroring the exact pattern used in ApolloFactory.ts
Removed the CombinedGraphQLErrors import that was causing the CI failure

The fix is now a single file change, 100% aligned with the codebase conventions.

PR #20153, opened 2026-04-30
add NO_STAGE stage option + hide column by default; block empty stage in workflow UI

jsdevninja:
## Summary
- Add a **`NO_STAGE`** option to standard **Opportunity › Stage**.
- Seed the **Kanban (“by Stage”) view** with a **`noStage` view group mapped to `NO_STAGE`** and set **`isVisible: false`** so the column stays hidden until users explicitly show it (matches maintainer guidance to use a sentinel stage instead of nullable).
- Migrate existing workspaces with a **workspace upgrade command** that appends **`NO_STAGE`** to **`opportunity.stage`** options when missing.

## Workflow UX
**Do not expose an empty / “No {field}” select option when the underlying field is not nullable** (`FormFieldInput` → `FormSelectFieldInput`). Clearing a workflow variable fallback now picks the **first available option** instead of trying to persist an empty/`null`-like state for required selects like **stage**.
## Not changing
Per review feedback we are **not** making **`stage`** nullable (`null`), and we dropped the nullable migration.
## Regression / housekeeping
Remove the **`stage: null`** GraphQL integration expectation.

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-field/ui/form-types/components/__stories__/FormMultiSelectFieldInput.stories.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-field/ui/form-types/components/__stories__/FormMultiSelectFieldInput.stories.tsx
@@ -19,6 +19,7 @@ type Story = StoryObj<typeof FormMultiSelectFieldInput>;
 export const Default: Story = {
   args: {
     label: 'Work Policy',
+    onChange: fn(),
     defaultValue: ['WORK_POLICY_1', 'WORK_POLICY_2'],
     options: [
       {
--- a/packages/twenty-front/src/modules/page-layout/components/PageLayoutTabsRenderer.tsx
+++ b/packages/twenty-front/src/modules/page-layout/components/PageLayoutTabsRenderer.tsx
@@ -102,11 +102,24 @@ export const PageLayoutTabsRenderer = () => {
     isEditMode: isPageLayoutInEditMode,
   });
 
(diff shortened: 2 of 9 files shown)


jsdevninja:
Hi @FelixMalfait

Could you please review? Thanks

charlesBochet:
@jsdevninja Thanks for opening a PR but let's not change that for now. What you can do: add another stage option (NO_STAGE) and hide the column on your opportunity kanban view.

Regarding the workflow, we should rather not allow to put a "No stage" if the field is not nullable. Could you open a PR in this direction?

jsdevninja:
@charlesBochet I updated PR as well as title & description, but not reflected here since it's closed.
Please re-open the PR and review.
Thank you.

jsdevninja:
@charlesBochet Sorry for pining, but would you please reopen this PR?

charlesBochet:
@jsdevninja please make a new one :)

jsdevninja:
@charlesBochet Thanks! I understand the preference for a new PR.

I’d strongly prefer to continue on this one since it already has the full context and discussion, and it will keep the history cleaner on my side. Would it be okay to reopen it just for that reason? I’ll make sure to incorporate all the requested changes.

jsdevninja:
@charlesBochet Would be great if you can reopen this PR, since as soon as this is reopened, the new updated code will be reflected here.

charlesBochet:
I can't
image

jsdevninja:
@charlesBochet Just created a new PR:
https://github.com/twentyhq/twenty/pull/20344

PR #20154, opened 2026-04-30
fix(security): reject ?token= URL query parameter for authentication

FelixMalfait:
## Summary

Removes the `?token=` URL query-parameter fallback from JWT authentication. Every authenticated route (`/graphql`, `/metadata`, REST, etc.) used to accept a full workspace JWT in the URL alongside the `Authorization` header. The fallback was intended for the REST API Playground only, but it was wired into the global Passport JWT extractor and applied to every route.

URL-borne tokens leak into:
- Server access logs (nginx / Apache / CDN / proxy / load balancer)
- Log aggregators (Datadog, CloudWatch, Loki, Sumo, …)
- Browser history (and synced across devices)
- `Referer` headers when navigating to external pages
- Browser extensions with `tabs`/`webNavigation` permissions

A leaked log line was equivalent to a leaked workspace credential for the lifetime of the token.

## What changed

- **`jwt-wrapper.service.ts`** — `extractJwtFromRequest()` is now header-only (`ExtractJwt.fromAuthHeaderAsBearerToken()`). No URL fallback anywhere in the system.
- **`open-api.service.ts` / `base-schema.utils.ts`** — Dropped the `token?: string` plumbing that propagated the URL token into the schema description. The "Authentication" section gains a "Never put your token in a URL" warning. The "Usage with LLMs" section is rewritten to point at the **Twenty MCP server** (header-authenticated, exposes typed tools — the right tool for AI agents) instead of telling users to paste tokenized OpenAPI URLs into Cursor/ChatGPT.
- **`RestPlayground.tsx`** — Now fetches the OpenAPI schema with `Authorization: Bearer ${playgroundApiKey}` and passes the JSON document to Scalar via `spec.content` instead of constructing a URL with `?token=`. Aborts in-flight fetches on unmount/key change.
- **New integration test** — Asserts `?token=` is rejected on `/rest/*`, `/graphql`, `/metadata`, and that `/rest/open-api/core?token=` returns the unauthenticated base schema (no workspace object paths).

## Why not keep `?token=` scoped to the OpenAPI endpoint only

The first instinct was to narrow the fallback to just `/rest/open-api/*`, since that endpoint is what the Scalar playground component fetches. But the same log-leakage attack still applies to that endpoint — the workspace JWT would still sit in access logs, just from one URL pattern instead of all of them. The cleaner long-term fix is to remove the URL pattern entirely and let the playground fetch with a header (Scalar supports `spec.content` natively). For LLM agent use, the MCP server is a strictly better path — typed tools, OAuth or header-based API key auth, no tokens in URLs anywhere.

## Not affected

File downloads at `file-url.service.ts` also use `?token=` URLs but with separate, short-lived `FILE`-typed tokens validated by `file-by-id.guard.ts` directly (not via `extractJwtFromRequest`). That mechanism is scoped per-file with limited TTL and is acceptable.

## Action required for users

Anyone who previously pasted `?token=` URLs into LLM tools, scripts, bookmarks, or shared configs should rotate their workspace API keys. Those tokens are likely captured in server logs / chat histories somewhere.

## Test plan

- [x] `npx nx typecheck twenty-front` — clean
- [x] `npx nx lint:diff-with-main twenty-server` — clean
- [x] `npx nx lint:diff-with-main twenty-front` — clean
- [x] OpenAPI utils unit tests + snapshots — 11/11 pass
- [ ] Run the new integration test against a live server: `nx run twenty-server:test:integration:with-db-reset` and verify `url-token-auth-rejection.integration-spec.ts` passes
- [ ] Manually open Settings → Playground → REST, confirm the schema loads (now via Bearer header instead of `?token=` URL)
- [ ] Manually verify `POST /metadata?token=<jwt>` (no Authorization header) returns Forbidden, and the same request with the token in the header returns the user


charlesBochet:
LGTM on the code, let's make sure Rest Playground still working!

charlesBochet (approved):

PR #20158, opened 2026-04-30
Add twenty sdk server upgrade command

martmull:
##
The command pulls the image, compares it against the one the container was created from, and only recreates the container if the image actually changed. Your data volumes are preserved — only the container is replaced.


packages/twenty-sdk/src/cli/utilities/server/docker-container.ts
```diff
@@ -49,6 +49,66 @@ export const containerExists = (containerName = CONTAINER_NAME): boolean => {
   }
 };
 
+export const getImageForVersion = (version = 'latest'): string =>
+  `twentycrm/twenty-app-dev:${version}`;
+
+export const getContainerImage = (
```
charlesBochet: clarify why we need both getContainerImage and getContainerImageId ? this should be simplified!

packages/twenty-sdk/src/cli/operations/server-upgrade.ts
```diff
@@ -0,0 +1,148 @@
...
+        code: SERVER_ERROR_CODES.DOCKER_NOT_RUNNING,
+        message: 'Docker is not running. Please start Docker and try again.',
+      },
+    };
+  }
+
+  const containerName = isTest ? TEST_CONTAINER_NAME : CONTAINER_NAME;
+  const image = getImageForVersion(version);
+  const hasContainer = containerExists(containerName);
+  const previousImageId = hasContainer
```
charlesBochet: let's clarify Image / ImageId / Digest, make it simple, it was clear when you explain but not when I read the code

charlesBochet (approved):

PR #20171, opened 2026-04-30
[Website] Implement translations.

### LINKED ISSUE #2406: [Website] Translations with Lingui and Crowdin.

mabdullahabaid:
As title.


mabdullahabaid:
Ran yarn install while resolving merge conflicts with main, so yarn.lock should be good too!

packages/twenty-website-new/src/design-system/components/LocalizedText.tsx
```diff
@@ -0,0 +1,35 @@
+import type { I18n, MessageDescriptor } from '@lingui/core';
+import { Trans } from '@lingui/react';
+
```
FelixMalfait: 1 export per file ; but I don't think we need all this to be honest, this should be handled by Lingui tooling directly (e.g. the fallback if the string isn't translated), it doesn't feel righ

packages/twenty-website-new/src/lib/i18n/LocalizedText.tsx
```diff
@@ -0,0 +1 @@
+export { LocalizedText } from '@/design-system/components/LocalizedText';
```
FelixMalfait: Re-export?

packages/twenty-website-new/src/design-system/components/LocalizedText.tsx
```diff
@@ -0,0 +1,35 @@
...
+export type LocalizableText = string | MessageDescriptor;
```
FelixMalfait: Why? Let's just use MessageDescriptor directly, no need for fallback wrapper that loses typing

packages/twenty-shared/package.json
```diff
@@ -53,6 +56,7 @@
     "react-router-dom": "^6.4.4",
     "temporal-polyfill": "^0.3.0",
     "transliteration": "^2.3.5",
+    "uuid": "^9.0.0",
```
FelixMalfait: Where in this PR do you use this in twenty-shared? Same for other dependencies added?

packages/twenty-website-new/src/app/[locale]/(home)/helped.data.ts
```diff
@@ -1,46 +1,50 @@
...
     heading: {
-      text: 'In production.',
+      text: msg`In production.`,
       fontFamily: 'sans',
     },
   },
   heading: [
     {
-      text: 'Dev teams power\ncompany-wide\n',
+      text: msg`Dev teams power\ncompany-wide\n`,
```
Comment (FelixMalfait) [inline packages/twenty-website-new/src/app/[locale]/(home)/helped.data.ts]: We can deal with later but these \n will probably be an issue

Comment (FelixMalfait) [reply packages/twenty-website-new/src/app/[locale]/(home)/helped.data.ts]: The other method that consists of splitting explicitely in multiple variables isn't good either (although it was quick to implement). The right way is to use CSS. But we can fix this later / in a subsequent PR

packages/twenty-shared/package.json
mabdullahabaid: I tried fixing this deployment bug.

https://github.com/twentyhq/twenty-infra/actions/runs/25275248087/job/74104013108

Not the best solution I believe, so I removed it to keep the PR focused to translations only.

Edit: Charles just fixed the same issue and merged to main.

FelixMalfait (approved):

PR #20172, opened 2026-04-30
[Dashboards] [Warning] Remove gauge chart support and delete existing widgets

ehconitin:
## Summary

Removes gauge chart from the chart-type picker and deletes existing gauge widgets via a workspace migration. The gauge was rendering a hardcoded `0.7 / "Progress"` stub regardless of configuration -- never wired to real data.

The contract stays in place. We keep `WidgetConfigurationType.GAUGE_CHART`, the DTO, the GraphQL union member, and the gauge folder -- so stored gauge JSON still resolves through the schema. The render path falls through to `default: return null`, so any un-migrated gauge widget renders as an empty cell, not a crash.

This PR just removes existing gauge widgets if there are any (via `upgrade:2-3:delete-gauge-widgets`). The deliberate cleanup -- deleting the type definitions, the gauge folder, the DTO -- comes in a follow-up PR after the migration has run.


Qodo-Free-For-OSS:
Hi, Existing GAUGE_CHART widgets still map to GraphType.GAUGE in the side-panel, so the UI can show Gauge settings while the widget renderer now falls through to default and returns null (empty cell). This creates an inconsistent and misleading editing experience for any workspace that still has gauge widgets before the migration runs.

**How to fix:** Hide GAUGE in settings

> ### Issue description
> After removing the gauge renderer and picker entry, existing GAUGE_CHART widget configurations can still reach the side-panel settings flow (derived as `GraphType.GAUGE`). This results in gauge settings being shown/usable while the actual widget render path returns `null`.
> 
> ### Issue Context
> This inconsistency occurs in the migration window (before `upgrade:2-3:delete-gauge-widgets` has been run on all workspaces).
> 
> ### Fix Focus Areas
> - packages/twenty-front/src/modules/side-panel/pages/page-layout/utils/getCurrentGraphTypeFromConfig.ts[7-35]
> - packages/twenty-front/src/modules/side-panel/pages/page-layout/constants/GraphTypeInformation.ts[21-59]
> - packages/twenty-front/src/modules/side-panel/pages/page-layout/components/ChartSettings.tsx[63-86]
> 
> ### What to change
> - Ensure gauge configurations do **not** resolve to `GraphType.GAUGE` in the side panel during this phase (e.g., map GAUGE config to a supported graph type, or return a safe fallback state).
> - Alternatively, guard `ChartSettings` to hide settings / show a deprecation message when `configuration.configurationType === WidgetConfigurationType.GAUGE_CHART`.
> - Keep behavior consistent with `GraphWidget` (which now renders empty for GAUGE_CHART).

---

packages/twenty-server/src/database/commands/upgrade-version-command/2-3/2-3-workspace-command-1798000000000-delete-gauge-widgets.command.ts
```diff
@@ -0,0 +1,98 @@
...
+        { workspaceId },
+      );
+
+    const result =
+      await this.workspaceMigrationValidateBuildAndRunService.validateBuildAndRunWorkspaceMigration(
+        {
+          allFlatEntityOperationByMetadataName: {
+            pageLayoutWidget: {
+              flatEntityToCreate: [],
+              flatEntityToDelete: gaugeWidgets,
```
charlesBochet, line 75: @ehconitin what do we have in the widget configuration? If it's tied to another metadata (like fieldMetadata or something else), how does the validateBuildAndRun behave on deletion?

charlesBochet (commented):
Reviewed @ehconitin, left a question.

packages/twenty-server/src/database/commands/upgrade-version-command/2-3/2-3-workspace-command-1798000000000-delete-gauge-widgets.command.ts
ehconitin: Deletion path only checks the widget exists (validateFlatPageLayoutWidgetDeletion at https://github.com/twentyhq/twenty/blob/main/packages/twenty-server/src/engine/workspace-manager/workspace-migration/workspace-migration-builder/validators/services/flat-page-layout-widget-validator.service.ts#L120) -- doesn't validate config refs, so leftover/orphan aggregateFieldMetadataId etc. don't block it. No cleanup concern either -- refs live inside the widget's JSONB and vanish with the soft-delete -- nothing FKs to pageLayoutWidget.id

Same precedent as the 1-23 backfill!

ehconitin:
@charlesBochet answered!

packages/twenty-server/src/database/commands/upgrade-version-command/2-3/2-3-workspace-command-1798000000000-delete-gauge-widgets.command.ts
```diff
@@ -0,0 +1,98 @@
...
+import { ActiveOrSuspendedWorkspaceCommandRunner } from 'src/database/commands/command-runners/active-or-suspended-workspace.command-runner';
+import { WorkspaceIteratorService } from 'src/database/commands/command-runners/workspace-iterator.service';
+import { type RunOnWorkspaceArgs } from 'src/database/commands/command-runners/workspace.command-runner';
+import { ApplicationService } from 'src/engine/core-modules/application/application.service';
+import { RegisteredWorkspaceCommand } from 'src/engine/core-modules/upgrade/decorators/registered-workspace-command.decorator';
+import { WidgetConfigurationType } from 'src/engine/metadata-modules/page-layout-widget/enums/widget-configuration-type.type';
+import { WorkspaceCacheService } from 'src/engine/workspace-cache/services/workspace-cache.service';
+import { WorkspaceMigrationValidateBuildAndRunService } from 'src/engine/workspace-manager/workspace-migration/services/workspace-migration-validate-build-and-run-service';
+
+@RegisteredWorkspaceCommand('2.3.0', 1798000000000)
```
Weiko, line 13: Should now be a 2.4 command (2.3 has been released already, a CI will soon check that for us)

Weiko (commented):
Left a comment about the command version to bump, otherwise LGTM

packages/twenty-server/src/database/commands/upgrade-version-command/2-3/2-3-workspace-command-1798000000000-delete-gauge-widgets.command.ts
ehconitin: @claude, update the folder to 2.4 and the timestamp while you are at it :) thanks

ehconitin: ah no nvm! we have conflicts I will handle it

charlesBochet (approved):

PR #20181, opened 2026-05-01
feat(apps): generic OAuth provider support for app SDK

FelixMalfait:
## Summary

App developers can now declare third-party OAuth integrations (GitHub, Linear, Slack, etc.) in their manifest and the platform handles the full authorize → callback → token-exchange → refresh → injection lifecycle. The dev writes ~10 lines of config and reads tokens via `useOAuth('linear')` inside any logic function.

```ts
// app/src/oauth-providers/linear.ts
export default defineOAuthProvider({
  universalIdentifier: '...',
  name: 'linear',
  displayName: 'Linear',
  authorizationEndpoint: 'https://linear.app/oauth/authorize',
  tokenEndpoint: 'https://api.linear.app/oauth/token',
  scopes: ['read', 'write'],
  connectionMode: 'per-user',
  clientIdVariable: 'LINEAR_CLIENT_ID',
  clientSecretVariable: 'LINEAR_CLIENT_SECRET',
  tokenRequestContentType: 'form-urlencoded',
});

// app/src/logic-functions/handlers/...
const { accessToken } = useOAuth('linear'); // throws OAuthNotConnectedError if missing
```

## Architecture

- **Storage**: extends the existing `connectedAccount` table — new nullable `applicationOAuthProviderId` FK + new `app` value on the `ConnectedAccountProvider` enum. Existing Google/Microsoft flows are untouched.
- **OAuth flow**: a single `/apps/oauth/authorize` + `/apps/oauth/callback` controller pair handles every app provider. State travels in a JWT signed via the existing `JwtWrapperService` (new `APP_OAUTH_STATE` token type).
- **Token exchange**: goes through `SecureHttpClientService.createSsrfSafeFetch()` (so an installed app can't point `tokenEndpoint` at internal hosts).
- **Refresh**: piggybacks on the existing `ConnectedAccountRefreshTokensService` dispatch — Google/Microsoft drivers untouched, new app driver lives engine-side under `application-oauth-provider/refresh/`.
- **Injection**: the executor injects refreshed tokens as env vars (`OAUTH_<NAME>_ACCESS_TOKEN`, `_HANDLE`, `_SCOPES`, `_CONNECTED`); the SDK helpers `useOAuth` / `useOptionalOAuth` read them.
- **Frontend**: auto-rendered "OAuth Connections" section under each app's settings tab (no custom front component needed). App-managed connections are filtered out of `/settings/accounts` so the email/calendar page stays focused.
- **Disconnect**: best-effort revoke against the manifest's `revokeEndpoint` before deleting the row.

## Reference app

`packages/twenty-apps/internal/twenty-linear/` exercises the full pipeline:

- `defineOAuthProvider` for Linear
- `POST /linear/create-issue` and `GET /linear/teams` HTTP-route logic functions
- Vitest tests for the handlers

## Tests

- 14 server-side Jest tests: token-exchange util (form-urlencoded vs JSON, PKCE, error paths), flow service (authorize URL shape, state binding, ConnectedAccount upsert on first/reconnect, per-workspace mode, invalid state)
- 8 app-level Vitest tests: handler error paths, GraphQL request shape, Linear error propagation
- All 4 packages clean: `npx nx lint:diff-with-main` and `npx tsc --noEmit`

## Test plan

- [ ] Apply migration on a dev DB: `npx nx run twenty-server:database:migrate:prod`
- [ ] Regenerate frontend types: `npx nx run twenty-front:graphql:generate --configuration=metadata`
- [ ] Create a Linear OAuth app at https://linear.app/settings/api/applications/new with redirect URI `<SERVER_URL>/apps/oauth/callback`
- [ ] Deploy + install `twenty-linear` on a workspace, paste the Linear client id/secret into the app's variables
- [ ] Click "Connect Linear" in the app's settings tab → complete OAuth → verify `connectedAccount` row created with `provider = 'app'`
- [ ] Trigger `POST /linear/create-issue` with a valid teamId → verify issue lands in Linear
- [ ] Disconnect → verify the row is deleted and (if Linear's revoke endpoint is configured in the manifest) the revoke call fires
- [ ] Verify `/settings/accounts` does NOT show the Linear connection — it appears only under the Linear app's settings tab

## Out of scope (deliberately)

- **Cron + per-user providers**: a cron-triggered function with a per-user OAuth provider currently returns `CONNECTED=false` (no user context). The follow-up design is `useOAuthForUser(name, userWorkspaceId)` paired with a `POST /apps/oauth/connection-token` endpoint, deferred to keep this PR focused.
- **Token encryption at rest**: tokens stored as plain `varchar` matching the existing Google/Microsoft pattern. Worth a separate cross-cutting PR.
- **Manifest endpoint pinning**: a malicious app upgrade could change `tokenEndpoint` silently. Same trust model as logic-function source code (which already runs arbitrary server-side); worth tightening across the whole upgrade pipeline rather than just OAuth.
- **CLI helpers** (`twenty oauth show-callback-url`, `twenty oauth connect`): manual setup for v1.


packages/twenty-front/src/modules/settings/accounts/hooks/useMyConnectedAccounts.ts
```diff
@@ -4,12 +4,17 @@ import { useMyCalendarChannels } from '@/settings/accounts/hooks/useMyCalendarCh
...
 
 type CoreConnectedAccount = Omit<
   ConnectedAccount,
   'messageChannels' | 'calendarChannels'
 >;
 
+// Returns the user's email/calendar connected accounts. App-managed OAuth
+// connections (`provider = 'app'`) are excluded — they're surfaced under
+// each app's settings tab via `useMyAppConnectedAccounts` instead, so they
+// don't pollute the personal accounts page.
```
FelixMalfait: I see OIDC and SAML were already types of connected accounts?! So why didn't we need to exclude them too before then?

packages/twenty-shared/src/application/connectionProviderManifestType.ts
```diff
@@ -0,0 +1,31 @@
...
+  tokenEndpoint: string;
+  revokeEndpoint?: string;
+  scopes: string[];
+  clientIdVariable: string;
+  clientSecretVariable: string;
+  authorizationParams?: Record<string, string>;
+  tokenRequestContentType?: OAuthProviderTokenRequestContentType;
+  usePkce?: boolean;
+};
+
```
FelixMalfait: We should have just 1 export per file. Follow other file naming conventions in this folder

packages/twenty-shared/src/application/enums/syncable-entities.enum.ts
```diff
@@ -6,6 +6,7 @@ export enum SyncableEntity {
   Role = 'role',
   Skill = 'skill',
   Agent = 'agent',
+  ConnectionProvider = 'connectionProvider',
```
FelixMalfait, line 9: Please check the /skills folder in .cursor ; it contains skills about syncable entities. Load each skills and check that this has been wired properly

packages/twenty-server/test/integration/upgrade/suites/sequence-runner/__snapshots__/failing-sequence-runner.integration-spec.ts.snap
```diff
@@ -1,3 +1,3 @@
 // Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
 
-exports[`UpgradeSequenceRunnerService — failing sequence (integration) should throw when cursor command is not found in the sequence 1`] = `"Step "RemovedCommand" not found in upgrade sequence. The sequence only covers versions [1.21.0, 1.22.0, 1.23.0, 2.0.0, 2.1.0, 2.2.0]. Please upgrade to 1.21.0 first."`;
+exports[`UpgradeSequenceRunnerService — failing sequence (integration) should throw when cursor command is not found in the sequence 1`] = `"Step "RemovedCommand" not found in upgrade sequence. The sequence only covers versions [1.21.0, 1.22.0, 1.23.0, 2.0.0, 2.1.0, 2.2.0, 2.3.0]. Please upgrade to 1.21.0 first."`;
```
FelixMalfait: Looks like a short-term way to solve this / that the problem will popup again in the future. Raise a separate small PR for that and we'll merge main into our branch

packages/twenty-server/src/modules/connected-account/refresh-tokens-manager/services/connected-account-refresh-tokens.service.ts
```diff
@@ -141,6 +145,24 @@ export class ConnectedAccountRefreshTokensService {
...
+          try {
+            return await this.appOAuthRefreshAccessTokenService.refreshTokens(
+              connectedAccount,
+              refreshToken,
+            );
+          } catch (error) {
+            // Translate the engine-side exception so callers see a single
+            // exception class regardless of provider, matching what
+            // Google/Microsoft drivers throw.
+            if (error instanceof ApplicationOAuthProviderException) {
```
FelixMalfait: I'm not sure why we need to do that downstream. Is there something cleaner that should be done at the root? What are our options here?

packages/twenty-server/src/modules/connected-account/email-alias-manager/services/email-alias-manager.service.ts
```diff
@@ -43,6 +43,9 @@ export class EmailAliasManagerService {
       case ConnectedAccountProvider.IMAP_SMTP_CALDAV:
       case ConnectedAccountProvider.OIDC:
       case ConnectedAccountProvider.SAML:
+      case ConnectedAccountProvider.APP:
+        // App-managed OAuth connections aren't email accounts; alias
```
FelixMalfait: Not sure these kind of comments are really necessary, seems quite straightforward. Check all comments in this PR and only keep the relevant ones

packages/twenty-server/src/engine/metadata-modules/connected-account/entities/connected-account.entity.ts
```diff
@@ -56,6 +67,40 @@ export class ConnectedAccountEntity extends WorkspaceRelatedEntity {
   @Column({ type: 'uuid', nullable: false })
   userWorkspaceId: string;
 
+  // DB column stays OAuth-named (storage detail); TS / GraphQL surface
```
FelixMalfait: Remove all the un-necessary comments in this file

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
```diff
@@ -1140,6 +1140,11 @@ export class AuthService {
         return [];
       case ConnectedAccountProvider.IMAP_SMTP_CALDAV:
         return [];
+      case ConnectedAccountProvider.APP:
+        // App-managed OAuth connections never participate in SSO sign-in;
```
FelixMalfait: Remove

FelixMalfait: Check security aspect. Make sure we can't use our app-declare connected accounts to circumvent login. This should not have any impact on login

packages/twenty-server/src/engine/core-modules/application/application-variable/application-variable.service.ts
```diff
@@ -42,6 +42,48 @@ export class ApplicationVariableEntityService {
     });
   }
 
+  // Decrypted plaintext value. Server-side only — never expose via GraphQL.
```
FelixMalfait: I don't understand why we have changes to packages/twenty-server/src/engine/core-modules/application/application-variable/application-variable.service.ts in this PR, is it still relevant? Or is it legacy from a previous iteration? Remove if not relevant

packages/twenty-server/src/engine/core-modules/application/application-oauth-provider/refresh/types/app-oauth-tokens.type.ts
```diff
@@ -0,0 +1,8 @@
+// Refresh result returned by the app OAuth driver. Mirrors
+// `ConnectedAccountTokens` from the central refresh manager but redeclared
+// here so this engine-side driver has zero dependency on `modules/`.
```
FelixMalfait: But I guess this module as a dependency on the core so why not import it that way? (maybe a dumb question?)

packages/twenty-server/src/engine/core-modules/application/application-oauth-provider/connections/application-connections.controller.ts
```diff
@@ -0,0 +1,54 @@
...
+// application access token (already injected into the function runtime as
+// TWENTY_APP_ACCESS_TOKEN). Apps can only list their own connections.
+@Controller('apps/connections')
+@UseGuards(JwtAuthGuard, WorkspaceAuthGuard, NoPermissionGuard)
+export class ApplicationConnectionsController {
+  constructor(
+    private readonly listService: ApplicationConnectionsListService,
+  ) {}
+
+  @Post('list')
```
FelixMalfait, line 36: I would have expect a Get rather than a Post? Also isn't the standard to use graphql in apps? I think I would have expected a graphql resolver not a controler? Wdyt?

packages/twenty-server/src/engine/core-modules/application/application-oauth-provider/connections/dtos/app-connection.dto.ts
```diff
@@ -0,0 +1,12 @@
+// Wire shape returned to apps from POST /apps/connections/list. Matches the
+// SDK's `AppConnection` type. Kept minimal — extend when a real consumer
+// needs more fields.
+export type AppConnectionDto = {
```
FelixMalfait: Check ALL the types declared in this PR and make sure you use Pick whenever possible rather than redeclaring overlapping types. Confirm you checked all type files.

packages/twenty-apps/internal/twenty-linear/README.md
```diff
@@ -0,0 +1,54 @@
+# twenty-linear
```
FelixMalfait: Let's focus on what this app does for users in a first section ; and then only have a section for self-hosters. This readme is both used in the marketplace in Twenty cloud (so mostly non-technical people looking to click "install" on an already approved appRegistration), and also for self-hosters (more technical! appRegistration has probably already been picked up from npm automatically but app still needs to be configured). We don't want to confuse the first user category with the second

packages/twenty-front/src/pages/settings/applications/components/SettingsApplicationConnectScopePickerModal.tsx
```diff
@@ -0,0 +1,81 @@
...
+export const SettingsApplicationConnectScopePickerModal = ({
+  modalInstanceId,
+  providerDisplayName,
+  onConfirm,
+}: SettingsApplicationConnectScopePickerModalProps) => {
+  const { t } = useLingui();
+  const { closeModal } = useModal();
+  const [scope, setScope] = useState<'user' | 'workspace'>('user');
+
+  const options = [
```
FelixMalfait: This model pattern is extremely ugly and not like what we do before. Let's find something else than the modal. Could be for example instead of having one CTA, it's a CTA with an arrow for a second option, or "..." and then the two options etc

packages/twenty-shared/src/application/connectionProviderManifestType.ts
FelixMalfait: We should have just 1 export per file.

packages/twenty-apps/internal/twenty-linear/src/connection-providers/linear-connection.ts
```diff
@@ -0,0 +1,23 @@
+import { defineConnectionProvider } from 'twenty-sdk/define';
+
+import { LINEAR_CONNECTION_PROVIDER_UNIVERSAL_IDENTIFIER } from 'src/constants/universal-identifiers';
+
+export default defineConnectionProvider({
+  universalIdentifier: LINEAR_CONNECTION_PROVIDER_UNIVERSAL_IDENTIFIER,
+  name: 'linear',
+  displayName: 'Linear',
+  icon: 'IconBrandLinear',
```
FelixMalfait: I guess in most cases we will use the app's logo ; and if not I think having a logo is eventually useful but not an Icon (there will rarely be a Tabler Icon for the given provider, unless it's a very popular service)

packages/twenty-server/src/engine/core-modules/application/application-oauth-provider/application-connection-provider.resolver.ts
```diff
@@ -0,0 +1,51 @@
...
+
+@UseGuards(WorkspaceAuthGuard)
+@MetadataResolver(() => ApplicationConnectionProviderDTO)
+export class ApplicationConnectionProviderResolver {
+  constructor(
+    private readonly oauthProviderService: ApplicationOAuthProviderService,
+  ) {}
+
+  @Query(() => [ApplicationConnectionProviderDTO])
+  @UseGuards(NoPermissionGuard)
```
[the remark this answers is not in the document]
FelixMalfait: I think this is a false positive — happy to be talked out of it though.

The query returns provider metadata (name / displayName / icon / scopes / `isClientCredentialsConfigured`) for one application in the request workspace. No secrets cross the wire (`clientSecret` lives on the registration row, never in this DTO). Workspace boundary is enforced by `@AuthWorkspace` + the explicit `workspaceId` filter in `oauthProviderService.findManyByApplication`.

This list is what powers the `SettingsApplicationConnectionsSection` UI for **end users** clicking "Add connection" — gating it behind `SETTINGS_APPLICATIONS` would mean only workspace admins can see the available providers, which would block non-admin users from connecting their own personal credentials (which is the main reason for the per-credential `scope` UX in this PR).

If we agree, marking resolved — happy to revisit if you'd rather have a tighter "this is a configuration list, but it's not sensitive" guard.

packages/twenty-shared/src/application/enums/syncable-entities.enum.ts
FelixMalfait: This was the SyncableEntity-skill conversation — we pivoted to keeping `applicationOAuthProvider` (now renamed `ConnectionProvider` in the public API) as a regular workspace-scoped entity rather than wiring it through the workspace-migration pipeline. Reasoning:

- The OAuth provider table doesn't need `universalIdentifier`-based action handlers / dry-run / undo / dependency-DAG — manifest sync calls `upsertManyFromManifest` directly and that's it.
- It's the smaller of the two remaining "bespoke manifest sync" holdouts; the other is `applicationVariable` (issue #20215).

I opened #20215 to track converting both as one cleanup pass. Marking this resolved — it's tracked there.

packages/twenty-server/src/engine/core-modules/application/application-oauth-provider/refresh/types/app-oauth-tokens.type.ts
FelixMalfait: Not a dumb question — you were right, this redeclaration wasn't pulling its weight. Engine modules already import types from `modules/` (203 occurrences in `packages/twenty-server/src/engine/`), so the "zero dependency on modules/" comment was aspirational rather than enforced.

Switched to importing `ConnectedAccountTokens` directly and deleted `app-oauth-tokens.type.ts` in [`ff22b012`](https://github.com/twentyhq/twenty/pull/20181/commits/ff22b0128b).

packages/twenty-server/src/engine/core-modules/application/application-oauth-provider/connections/application-connections.controller.ts
FelixMalfait: Two layered questions — let me take them in order:

**POST vs GET.** I'd usually agree with GET-by-default, but `/list` takes a structured filter (`{ providerName?, userWorkspaceId?, scope? }`). GET-with-body is awkward (most clients/proxies strip request bodies on GET) and encoding it as query string brings its own problems — comma-separated arrays, URL length limits if the filter ever grows, harder to validate with `class-validator`. Other Twenty endpoints that take a body (e.g. `POST /apps/oauth/callback`) use POST for the same reason.

**REST vs GraphQL.** This one's not arbitrary either:

- The runtime caller is a logic-function bundle running in the app's Lambda sandbox. The SDK there is intentionally **tiny** — no Apollo Client, no `graphql` package, no schema codegen — just `fetch`. Pulling GraphQL into the runtime would balloon the bundle and re-introduce the dep on twenty-shared schema artifacts that we deliberately stripped.
- The endpoint is auth'd via the `TWENTY_APP_ACCESS_TOKEN` (an APPLICATION_ACCESS JWT injected at execution time), not a user session — different guard chain than the GraphQL surface, which is user-session-oriented.
- Same precedent as `POST /apps/oauth/{authorize,callback}` and the `/apps/connections/list` is a sibling of those.

If the broader plan is for app runtimes to start using GraphQL (with a slimmer client?), happy to align then — but the runtime-side cost feels too high right now for too little gain. WDYT?

packages/twenty-server/src/engine/core-modules/application/application-oauth-provider/connections/dtos/app-connection.dto.ts
FelixMalfait: Did the audit. Findings:

The 13 type/dto files added/modified in this PR break down into:

| File | Pick-able? |
|---|---|
| `AppConnectionDto` (server) and `AppConnection` (sdk) | **Yes — and they were duplicates.** Hoisted the canonical shape to `twenty-shared/application/appConnectionType.ts` and both sides now re-export from it. Drift fails typecheck. |
| `ApplicationConnectionProviderDTO`, `ConnectedAccountDTO`, `GetAppConnectionDto`, `ListAppConnectionsDto` | No — class-based with `class-validator`/`@nestjs/graphql` decorators. Decorators don't survive `Pick`. |
| `AppOAuthStateJwtPayload` | Already extends `CommonPropertiesJwtPayload` (the right pattern). |
| `TokenExchangeResponse` (server) vs `ConnectedAccountTokens` (modules) | Look similar but `refreshToken` and `scopes` are nullable here vs non-null over there. Pick'ing would force exposing a non-null where the upstream is nullable — wrong direction. |
| `AppOAuthTokens` (deleted) | Was a redundant re-declaration of `ConnectedAccountTokens`. Removed in same commit per [your other thread](#discussion_r2511317553). |
| `OAuthProviderTokenRequestContentType` | Plain string union. |
| `CreateIssueInput`, `CreateIssueMutationResult`, `LinearGraphQLResult` (twenty-linear) | Domain shapes for the Linear API — no parent in this codebase. |
| `DefinableEntity`, `DefineEntity` (sdk) | Discriminated union of manifest configs — already a union, not Pick-able. |

Net change: extracted the SDK↔server duplication into twenty-shared in [`ff22b012`](https://github.com/twentyhq/twenty/pull/20181/commits/ff22b0128b).

packages/twenty-apps/internal/twenty-linear/README.md
FelixMalfait: Restructured in [`ff22b012`](https://github.com/twentyhq/twenty/pull/20181/commits/ff22b0128b) along the lines you suggested:

- Top section is **What you can do** + **Installing**, written for a non-technical workspace user clicking "Install" in the marketplace. Mentions both scope choices in plain language.
- A note in that section says "if you see a 'OAuth not yet set up' notice, ask your admin to follow the setup below" — keeps the two audiences from getting tangled.
- Below the `---`, a **Self-hosting setup** block: opens with "this section is for Twenty server admins. If you're on Twenty Cloud, skip this". Three subsections (Linear OAuth app → wire creds into Twenty → optional source-build for app developers).

Let me know if the framing still drifts — happy to iterate.

packages/twenty-server/src/modules/connected-account/refresh-tokens-manager/services/connected-account-refresh-tokens.service.ts
```diff
@@ -141,6 +144,11 @@ export class ConnectedAccountRefreshTokensService {
           return await this.microsoftAPIRefreshAccessTokenService.refreshTokens(
             refreshToken,
           );
+        case ConnectedAccountProvider.APP:
+          return await this.appOAuthRefreshAccessTokenService.refreshTokens(
```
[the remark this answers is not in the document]
FelixMalfait: Already addressed in [`40ed3b9d`](https://github.com/twentyhq/twenty/pull/20181/commits/40ed3b9dbd) (`fix(apps): classify app OAuth refresh failures correctly + translate provider lookup errors`). The `AppOAuthRefreshAccessTokenService` now catches `ApplicationOAuthProviderException` thrown during provider lookup / credential resolution and translates it into a `ConnectedAccountRefreshAccessTokenException` with the right code (`PROVIDER_NOT_SUPPORTED` for missing-provider, `TEMPORARY_NETWORK_ERROR` vs `INVALID_REFRESH_TOKEN` for the actual refresh).

Marking resolved.

packages/twenty-front/src/pages/settings/applications/components/SettingsApplicationConnectScopePickerModal.tsx
FelixMalfait: Replaced the modal with an inline split-button in [`ff22b012`](https://github.com/twentyhq/twenty/pull/20181/commits/ff22b0128b). The "Add connection" CTA now opens a small `Dropdown` anchored to itself with two `MenuItem`s — "Just for me" (IconUser) and "Workspace shared" (IconUsers). No modal layer at all, sits flush below the connections list. Same pattern as `SettingsAdminQueueJobRowDropdownMenu` so it matches existing settings UI.

packages/twenty-apps/internal/twenty-linear/src/connection-providers/linear-connection.ts
FelixMalfait: Agreed — Tabler doesn't ship brand icons for most providers, so the field is going to be empty in practice almost everywhere. Dropped `icon: 'IconBrandLinear'` from twenty-linear in [`ff22b012`](https://github.com/twentyhq/twenty/pull/20181/commits/ff22b0128b) and left a comment in the file pointing at the bigger fix.

The bigger fix — adding a `logoUrl` field to `ConnectionProviderManifest` (mirroring `defineApplication({ logoUrl })`), threading it through the SDK CLI's asset upload, the server entity, and the frontend rendering — felt out of scope for this PR. Tracked it as #20216.

packages/twenty-docs/developers/extend/apps/connections.mdx
```diff
@@ -0,0 +1,195 @@
...
+  type: 'oauth',
+  oauth: {
+    authorizationEndpoint: 'https://linear.app/oauth/authorize',
+    tokenEndpoint: 'https://api.linear.app/oauth/token',
+    scopes: ['read', 'write'],
+    // Names of serverVariables on this app's defineApplication. The server
+    // admin fills these in once on the app registration; all workspaces
+    // installing the app share the same OAuth client.
+    clientIdVariable: 'LINEAR_CLIENT_ID',
+    clientSecretVariable: 'LINEAR_CLIENT_SECRET',
```
martmull, line 34: those values should probably be defined as serverVariables

packages/twenty-docs/developers/extend/apps/connections.mdx
```diff
@@ -0,0 +1,195 @@
...
+    // revokeEndpoint: 'https://example.com/oauth/revoke',
+  },
+});
+```
+
+Key points:
+
+- `name` is the unique identifier string used in `listConnections({ providerName })` (kebab-case, must match `^[a-z][a-z0-9-]*$`).
+- `displayName` shows in the per-app settings tab and in the AI tool list.
+- `clientIdVariable` / `clientSecretVariable` reference **server variables** declared on `defineApplication.serverVariables` (not application variables). The server admin fills them on the app registration, and all installed workspaces share them.
```
martmull: ok, documented here

martmull: we may add a small example then

```suggestion
```ts src/application-config.ts
import { defineApplication } from 'twenty-sdk/define';

export default defineApplication({
   ...
  serverVariables: {
     ...,
     LINEAR_CLIENT_ID: {...},
     LINEAR_CLIENT_SECRET: {...},
  },
});
```
```

File: packages/twenty-docs/developers/extend/apps/connections.mdx [hunk 20]
```diff
@@ -0,0 +1,195 @@
...
+      isRequired: true,
+    },
+    LINEAR_CLIENT_SECRET: {
+      description: 'OAuth client secret from your Linear OAuth application.',
+      isSecret: true,
+      isRequired: true,
+    },
+  },
+});
+```
```
martmull, line 72: jeez ^^

packages/twenty-docs/developers/extend/apps/connections.mdx
FelixMalfait: Good catch — the matching `serverVariables` block was hidden behind a separate accordion, easy to miss. Pulled both file snippets into the same accordion so the reader sees the full setup at once, and tightened the inline note to call out that `clientIdVariable` / `clientSecretVariable` are *names* (not values), with the actual credentials entered by the server admin through the registration UI (never committed). [`d515db90`](https://github.com/twentyhq/twenty/pull/20181/commits/d515db900e)

packages/twenty-docs/developers/extend/apps/connections.mdx
FelixMalfait: Not sure what "jeez" is reacting to here — the line at the time was `displayName: 'Linear'` inside the `defineApplication` example, which does duplicate the `displayName: 'Linear'` already on the connection provider. If that's the gripe: the duplication is on purpose — apps with multiple connection providers (e.g. a "Communications" app exposing both Slack and Discord) need a displayName per provider distinct from the app's own. For Linear it just happens to collapse to the same string.

But happy to be wrong about what you meant — let me know what you'd like changed.

packages/twenty-docs/developers/extend/apps/connections.mdx
FelixMalfait: Pulled both file snippets into the same accordion in [`d515db90`](https://github.com/twentyhq/twenty/pull/20181/commits/d515db900e), so the cross-reference is no longer split across two collapsible sections — the reader sees the connection-provider and its matching `serverVariables` declaration in one place.

FelixMalfait: Done in [`d515db90`](https://github.com/twentyhq/twenty/pull/20181/commits/d515db900e) — added the `defineApplication.serverVariables` snippet directly under the `defineConnectionProvider` example in the same accordion, with both `LINEAR_CLIENT_ID` and `LINEAR_CLIENT_SECRET` declared with full `description` / `isSecret` / `isRequired` props (rather than the suggestion's `{...}` placeholder, since this is the only place readers see what a serverVariable declaration looks like in this guide).

packages/twenty-front/src/modules/accounts/types/ConnectedAccount.ts
```diff
@@ -12,6 +12,10 @@ export type ConnectedAccount = {
   handleAliases: string[] | null;
   lastSignedInAt: string | null;
   userWorkspaceId: string;
+  applicationConnectionProviderId: string | null;
+  name: string | null;
+  scope: 'user' | 'workspace';
```
martmull: might be another naming because it is conflicting with existing `scopes` attribute

FelixMalfait: Good call — the one-letter difference between `scope` and `scopes` is a bug magnet. Renamed to `visibility` across the stack in [`05d4b0a2`](https://github.com/twentyhq/twenty/pull/20181/commits/05d4b0a209): DB column, `ConnectedAccountVisibility` type, entity / DTO / JWT payload / controller / flow service / connections-list service + filter DTO, twenty-shared `AppConnection` shape, SDK helpers (`findConnectionForRequest`, `listConnections` filter), frontend type / hook / UI / GraphQL query, generated metadata, docs, and twenty-linear. Values stay `'user' | 'workspace'`. 28 files but mechanical.

packages/twenty-front/src/modules/settings/applications/graphql/queries/findApplicationConnectionProviders.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const FIND_APPLICATION_CONNECTION_PROVIDERS = gql`
+  query ApplicationConnectionProviders($applicationId: UUID!) {
+    applicationConnectionProviders(applicationId: $applicationId) {
+      id
+      applicationId
+      type
+      name
+      displayName
+      icon
```
martmull: do we really want an icon?

packages/twenty-front/src/pages/settings/applications/hooks/useMyAppConnectedAccounts.ts
```diff
@@ -0,0 +1,34 @@
...
+// these out; this hook is the inverse, used by the per-app settings tab.
+type AppConnectedAccount = {
+  id: string;
+  handle: string;
+  provider: ConnectedAccountProvider;
+  authFailedAt: string | null;
+  applicationConnectionProviderId: string | null;
+  name: string | null;
+  visibility: 'user' | 'workspace';
+};
```
martmull: We should use generated types

packages/twenty-front/src/pages/settings/applications/tabs/SettingsApplicationConnectionsSection.tsx
```diff
@@ -0,0 +1,206 @@
...
+    refetchQueries: [{ query: GET_MY_CONNECTED_ACCOUNTS }],
+  });
+
+  if (loading || connectionProviders.length === 0) {
+    return null;
+  }
+
+  return (
+    <>
+      {connectionProviders.map((provider) => {
```
martmull, line 114: i think we should comply with existing list design. Lets create a table where each rows are pointing to a dedicated connectedAccountDetail page (ok to do it in another PR)

packages/twenty-sdk/src/cli/commands/add.ts
```diff
@@ -159,6 +160,16 @@ export class EntityAddCommand {
...
       }
 
+      case SyncableEntity.ConnectionProvider: {
+        const name = await this.getEntityName(entity);
+
+        const file = getConnectionProviderBaseFile({
+          name,
+        });
+
+        return { name, file };
```
martmull, line 177: client id and clientSecret server variables should be added to existing defineApplication.serverVariables

packages/twenty-server/src/engine/core-modules/i18n/locales/de-DE.po
```diff
@@ -197,7 +197,7 @@ msgid "A temporary error occurred. Please try again."
 msgstr "Es ist ein temporärer Fehler aufgetreten. Bitte versuchen Sie es erneut."
 
 #. js-lingui-id: reBsjo
-#: src/modules/connected-account/refresh-tokens-manager/exceptions/connected-account-refresh-tokens.exception.ts
+#: src/engine/metadata-modules/connected-account/exceptions/connected-account-refresh-tokens.exception.ts
```
martmull: we should not update generated files

martmull (commented):
comments

packages/twenty-front/src/modules/settings/applications/graphql/queries/findApplicationConnectionProviders.ts
FelixMalfait: Good question — no, we don't. Tabler doesn't ship brand icons for most OAuth providers (Notion, Slack, HubSpot, Salesforce, Stripe, GitLab…), so the field has been empty in practice. Dropped it everywhere in [`b931474f`](https://github.com/twentyhq/twenty/pull/20181/commits/b931474fbf): GraphQL DTO + query + frontend types + manifest type + entity + DB column (in-place migration edit). Tracked the proper replacement (a `logoUrl` field mirroring `defineApplication({ logoUrl })`) in #20216.

packages/twenty-front/src/pages/settings/applications/hooks/useMyAppConnectedAccounts.ts
FelixMalfait: Done in [`b931474f`](https://github.com/twentyhq/twenty/pull/20181/commits/b931474fbf). The hand-defined `AppConnectedAccount` is now a derived type from the codegen-generated `MyConnectedAccountsQuery`:

```ts
export type AppConnectedAccount =
  MyConnectedAccountsQuery['myConnectedAccounts'][number];
```

New fields added to the GraphQL query selection now flow through automatically.

packages/twenty-front/src/pages/settings/applications/tabs/SettingsApplicationConnectionsSection.tsx
FelixMalfait: Tracked as #20220 per your "ok to do it in another PR". Captured both halves of what you described: replacing `SettingsListCard` with the canonical Settings table, and a per-connection detail page where Reconnect / Disconnect / Rename live (instead of inline buttons that are easy to misclick). Happy to take that on once this PR lands — let me know if you'd rather pick a different table component than what's at e.g. `/settings/accounts` for the canonical pattern.

packages/twenty-sdk/src/cli/commands/add.ts
FelixMalfait: Done in [`b931474f`](https://github.com/twentyhq/twenty/pull/20181/commits/b931474fbf). New `appendServerVariablesToAppConfig` util, called from the connection-provider branch of `twenty add`. Handles three cases:

- **existing `serverVariables` block** → injects new entries after the opening `{` (preserves existing entries)
- **no block yet** → adds a fresh `serverVariables: {…},` property right before the closing `})` of `defineApplication`
- **already declared** → no-op, prints `(already declared)` so re-running is idempotent

Falls back to a printed `yellow` snippet when the file shape doesn't match (no `src/application.config.ts`, no `defineApplication(` call, etc.) — best-effort, never destructive. 6 specs cover every branch.

packages/twenty-server/src/engine/core-modules/i18n/locales/de-DE.po
FelixMalfait: Reverted all 30 .po files in [`b931474f`](https://github.com/twentyhq/twenty/pull/20181/commits/b931474fbf). The path comments got rewritten because an earlier commit on this branch moved `connected-account-refresh-tokens.exception.ts` from `src/modules/...` to `src/engine/metadata-modules/...` and someone re-ran lingui extract. The lingui extract bot picks the new source path up automatically on its next scheduled run, so committing the rewrites in this PR just adds noise — better to let the bot do it post-merge.

PR #20192, opened 2026-05-02
fix: handle missing file entity in avatar deletion listener

### LINKED ISSUE #20191: WorkspaceMemberAvatarFileDeletionListener crashes NestJS server with unhandled EntityNotFoundError on any workspaceMember update (theme change, avatar upload/remove)
## Bug Description

Any update to a `workspaceMember` record (theme change, avatar upload/remove, locale, timezone, etc.) triggers `WorkspaceMemberAvatarFileDeletionListener` which calls `FileCorePictureService.deleteCorePicture`. If no matching `FileEntity` exists in `core.file`, an unhandled `EntityNotFoundError` propagates and kills the entire NestJS server process, causing a 502 for all users until Docker restarts the container. The server then encounters the same orphaned `avatarUrl` reference again on the next `workspaceMember` update, creating a crash loop.

## Exact Stack Trace

```
FileException [Error]: Error retrieving file: Could not find any entity of type "FileEntity" matching:
    "_value": "core-picture/%",
    at FileController.getFileById (file.controller.js:84:19)
    code: 'INTERNAL_SERVER_ERROR',
    userFriendlyMessage: { id: 'XyOToQ', message: 'An error occurred.' }

EntityNotFoundError: Could not find any entity of type "FileEntity" matching:
    "_value": "core-picture/%",
    at FileCorePictureService.deleteCorePicture (file-core-picture.service.js:134:22)
    at WorkspaceMemberAvatarFileDeletionListener.deleteCorePictures (workspace-member-avatar-file-deletion.listener.js:39:13)
```

## Steps to Reproduce

1. Self-hosted Docker Compose, local storage, `latest` tag
2. Upload a profile picture for a workspace member — `avatarUrl` is written to `workspaceMember` but `core.file` record is not always committed atomically due to a race condition
3. Change theme (Light/Dark/System) or perform any other profile setting update
4. `WorkspaceMemberAvatarFileDeletionListener` fires on the `workspaceMember` update event
5. `deleteCorePicture` calls `findOneOrFail` on `FileEntity` — throws `EntityNotFoundError`
6. Unhandled exception kills the NestJS process
7. Docker restarts the container — all users receive 502 during restart window
8. On next `workspaceMember` update, same crash occurs → crash loop

## Confirmed Triggers

- Changing theme (Light/Dark/System)
- Uploading a profile picture
- Removing a profile picture
- Any other settings update that mutates the `workspaceMember` record

## Root Cause

`workspace-member-avatar-file-deletion.listener.js` line 39 calls `FileCorePictureService.deleteCorePicture` which uses `findOneOrFail` on `FileEntity`. When the `core.file` record is missing due to a failed or non-atomic upload transaction, the hard throw propagates unhandled and kills the entire NestJS process.

Additionally, the profile picture upload flow has a race condition where `workspaceMember.avatarUrl` is written with the new file ID before the corresponding `core.file` record is fully committed, leaving a dangling reference that permanently triggers the crash on any subsequent `workspaceMember` update.

## Expected Behavior

Missing `FileEntity` should be handled gracefully. The fix should be one of:

- Wrap `deleteCorePicture` call at line 39 in try/catch
- Use `findOne` with null check instead of `findOneOrFail`
- Ensure `core.file` record is committed atomically before writing `avatarUrl`

The server should **never** crash due to a missing file reference.

## Database Evidence

The orphaned reference pattern — `workspaceMember.avatarUrl` containing a file ID with no matching `core.file` record:

```sql
SELECT "userEmail", "avatarUrl"
FROM workspace_<id>."workspaceMember" wm
WHERE wm."avatarUrl" IS NOT NULL
AND NOT EXISTS (
    SELECT 1 FROM core.file f
    WHERE f.id::text = substring(wm."avatarUrl" FROM 'core-picture/([^?]+)')
);
```

Workaround to break crash loop (null out orphaned references):

```sql
UPDATE workspace_<id>."workspaceMember" wm
SET "avatarUrl" = NULL
WHERE wm."avatarUrl" IS NOT NULL
AND NOT EXISTS (
    SELECT 1 FROM core.file f
    WHERE f.id::text = substring(wm."avatarUrl" FROM 'core-picture/([^?]+)')
);
```

## Environment

- Self-hosted Docker Compose
- Storage: local
- Tag: `latest` (v2.2.0)
- PostgreSQL 16
- Ubuntu 24.04
- Cloudflare tunnel (cloudflared) as reverse proxy

## Related

- #8601 (closed — similar "File not found" on avatar, may be resurfaced by the v1.18.0 `avatarUrl → avatarFile` migration)

aayushbaluni:
## Problem
When a `workspaceMember` is updated (e.g., theme/locale/avatar changes), the `WorkspaceMemberAvatarFileDeletionListener` triggers file deletion. If the referenced file entity doesn't exist in the database, an unhandled `EntityNotFoundError` crashes the NestJS server, causing a 502 loop.

## Change
Wrap the file deletion call in a try-catch that gracefully handles `EntityNotFoundError` as a no-op — if the file doesn't exist, there's nothing to delete.

Fixes #20191.


martmull (approved):
thanks

etiennejouan (approved):

PR #20200, opened 2026-05-02
Fix/workspace member avatars 20193

### LINKED ISSUE #20193: Workspace member avatars not showing in table views and relation chips
Bug Description:
In table views and relation chips (e.g. Account Owner field on Companies), workspace member avatars are not displayed. The avatar picker/dropdown shows avatars correctly, but the table/chip rendering does not.
Root Cause (investigated):

workspace_[id].workspaceMember.avatarUrl is populated correctly with valid file URLs
core.userWorkspace.defaultAvatarUrl is empty for all workspace members
Table views and chips appear to rely on defaultAvatarUrl which is never populated
The dropdown picker fetches fresh data via GraphQL on each open, which works correctly

Expected Behavior:
Workspace member avatars should be visible in all table views and relation chips, consistent with the dropdown picker.
Environment:

Self-hosted, Docker Compose
Version: v2.2.0
All browsers affected
Fresh install (not an upgrade issue)

Pantkartik:
Fixes #20193

**Bug Description:**
Previously, workspace member avatars failed to render correctly in table views and relation chips (such as the Account Owner field). While the avatar picker dropdown correctly fetched fresh GraphQL data, table views and chips relied on the cached defaultAvatarUrl or avatarUrl fields, which were frequently resolving to empty strings or failing to parse external OAuth URLs correctly.

**Root Cause:**

- Empty String Defaults: Deleting an avatar or failing to retrieve one defaulted the database state to an empty string ("") instead of null, which caused frontend image components to break rather than render their fallback states.

- Missing Permanent URLs: The WorkspaceMemberTranspiler was strictly expecting internal signed URLs. If an avatar was an external OAuth URL, it incorrectly returned an empty string, breaking SSO profile pictures.

- Missing Fallbacks: New users lacked a proper Gravatar fallback assignment upon workspace creation.

**Changes Made:**

- user-workspace.service.ts: Updated the avatar computation logic during user creation to implement a reliable Gravatar fallback and correctly set missing avatars to null instead of empty strings. Updated the storage to use permanent file URLs.
- file-url.service.ts: Implemented a getRawFileUrl method to support rendering permanent, non-expiring file URLs for avatars.
- workspace-member-transpiler.service.ts: Refactored the URL transpilation logic to gracefully pass through external OAuth URLs (e.g., Google/Microsoft profile pictures) instead of stripping them.
- WorkspaceMemberPictureUploader.tsx: Fixed the frontend removal logic so that deleting a profile picture sets the avatarUrl to null (consistent with the backend) rather than an empty string.

**Testing:**

- Verified that avatars correctly display in relation chips and table views.
- Verified that external OAuth avatars load properly.
- Verified that deleting an avatar correctly resets the UI to the fallback initials component.


Pantkartik:
Hi maintainers 
Just a gentle follow-up on this PR.  
Would love your feedback whenever you get time. Thanks!

charlesBochet:
@Pantkartik This PR seems to be more about dates than avatar. I'm investigating the avatar issue but I don't think we want to support external urls

Pantkartik:
Thanks for the feedback @charlesBochet — makes sense.
I agree the calendar/date changes got mixed in due to the branch merge. I’ll clean this PR up and split/remove the unrelated commits so it focuses only on the workspace member avatar fixes.
For external URLs, should I completely drop support for OAuth image URLs (Google/Microsoft), or just normalize them differently instead of blocking them?

charlesBochet:
@Pantkartik After investigation:
- the fact that we don't display profile picture of workspaceMember is due to the fact that we only load minimal information for relations (here company.accountOwner). Minimal meaning id, labelIdentifer (name), imageIdentifier (avatarUrl). For new workspace workspaceMember imageIdentifier is not properly set and that's why it's not queries by the frontend
- we should not support gravatar, not sure why it was added
- avatarUrl is not containing external url images, not sure why you said so
- (not related to your change) user.defaultAvatarUrl is not used so dropping it

I'll push directly to your branch with the right fix

Pantkartik:
Thanks @charlesBochet for the detailed investigation, that makes sense now 
I understand the root cause is in minimal relation loading and missing imageIdentifier population rather than URL handling.
Feel free to push the fix directly to this branch or modify it as needed. I’ll also review the final changes to understand the correct approach. Thanks for guiding me on this!

charlesBochet:
Done @Pantkartik, I'm also adding an integration test

charlesBochet (approved):

charlesBochet (approved):
Updated the PR with the right strategy:
What's happening: the fact that we don't display profile picture of workspaceMember is due to the fact that we only load minimal information for relations (here company.accountOwner). Minimal meaning id, labelIdentifer (name), imageIdentifier (avatarUrl). For new workspace workspaceMember imageIdentifier is not properly set and that's why it's not queries by the frontend

Fix:
- backfill with an upgrade command
- set imageIdentifier on workspaceMember for new members

PR #20208, opened 2026-05-03
feat(apps): split AI tool and workflow action triggers in LogicFunction manifest

FelixMalfait:
## Summary

Replaces the bolted-on `isTool` + `toolInputSchema` fields on `LogicFunctionManifest` with two distinct, opt-in triggers that align with the existing `cron` / `databaseEvent` / `httpRoute` trigger pattern:

- **`toolTriggerSettings`** — exposes the function as an AI tool (chat / MCP / function calling). Uses standard JSON Schema (the format LLMs natively understand).
- **`workflowActionTriggerSettings`** — exposes the function as a step in the visual workflow builder. Uses Twenty's rich `InputSchema` so the builder can render proper `FieldMetadataType`-aware editors, variable pickers, labels, and an optional `outputSchema`.

A function can opt into none, one, or both. Each surface gets the schema format appropriate for it.

### Why

`isTool: true` previously exposed the function as both an AI tool AND a workflow node, with the same JSON Schema feeding both — but the workflow builder really wants Twenty's `InputSchema` (with `CURRENCY`, `RELATION`, `EMAILS`, etc.) and the AI surface really wants standard JSON Schema. Today the workflow builder hacks around this by treating JSON Schema as `InputSchema`, which silently breaks for any non-primitive field type. Splitting the triggers fixes that and lets each surface evolve independently.

### Migration

- **Fast** instance command adds the two new nullable columns.
- **Slow** instance command backfills `toolTriggerSettings` + `workflowActionTriggerSettings` from `isTool=true` rows (preserving today's both-surfaces behaviour) then drops the legacy columns.

### Stacked

Stacked on top of #20181. Merge that first, then this.

## Test plan

- [ ] CI green (oxlint, typecheck, jest, vitest)
- [ ] Run `--include-slow` upgrade against a workspace with existing `isTool=true` logic functions; verify both new columns populated and old columns dropped
- [ ] Verify AI chat sees migrated tool functions (Linear create-issue, Exa search) and can call them with the JSON Schema
- [ ] Add an AI-tool function from the Settings UI (toggles `toolTriggerSettings`) and verify it shows up in chat
- [ ] Add a workflow-action function from the Settings UI (toggles `workflowActionTriggerSettings`) and verify it appears in the workflow node picker
- [ ] In the workflow builder, edit a `LOGIC_FUNCTION` step and verify input fields render (no more JSON-Schema-as-InputSchema hack)
- [ ] Try defining a function with no triggers in the SDK and verify `defineLogicFunction` rejects it


packages/twenty-docs/developers/extend/apps/logic-functions.mdx
```diff
@@ -13,14 +13,22 @@ Each function file uses `defineLogicFunction()` to export a configuration with a
...
   const client = new CoreApiClient();
-  const name = 'name' in params.queryStringParameters
-    ? params.queryStringParameters.name ?? process.env.DEFAULT_RECIPIENT_NAME ?? 'Hello world'
-    : 'Hello world';
+  const name =
+    'name' in params.queryStringParameters
+      ? (params.queryStringParameters.name ??
+        process.env.DEFAULT_RECIPIENT_NAME ??
+        'Hello world')
+      : 'Hello world';
```
martmull: remove this change, it is less readable that before

packages/twenty-sdk/src/cli/utilities/build/manifest/manifest-build.ts
```diff
@@ -234,13 +233,8 @@ export const buildManifest = async (
 
         const relativeFilePath = relative(appPath, filePath);
 
-        const toolInputSchema =
-          rest.toolInputSchema ??
-          (await getInputSchemaFromSourceCode(fileContent));
```
martmull, line 239: inputSchema should be computed if not provided by the user during manifest build

martmull, line 239: please use jsonSchemaToInputSchema here

martmull (changes requested):
we need to compute tool input schema if not provided in defineLogicFunction during manifest building

packages/twenty-sdk/src/cli/utilities/build/manifest/manifest-build.ts
FelixMalfait: I think Opus executed on your schema. I'm not sure I'm convinced that the dynamic inference magic is something we want for apps as code (it's nice for workflows but for code?). Let me know if you think we should keep it or revert ; not a strong opinion on my side, not sure

FelixMalfait: Or actually I'll let you lead/decide since you own this PR now

martmull (approved):

PR #20217, opened 2026-05-04
fix: add loading indicator in table views during data fetch

### LINKED ISSUE #20211: UI: Missing loading indicator in table views during data fetch
Scope & Context

We are improving the user experience of table views. This issue focuses on providing proper feedback during data fetching in table views.

Current behavior

When navigating to or refreshing a table view, there is no visible loading indicator while data is being fetched. The UI appears blank or static, which can make it seem unresponsive.

Expected behavior

A loading indicator such as a spinner or skeleton should be displayed while data is being fetched. This should clearly communicate to the user that content is loading.

Technical inputs
Add a loading state in table view components
Use a spinner or skeleton loader component during API calls
Trigger loading state before fetch and reset after response
Ensure consistency across all table-based views

ArchitAgarwal04:
Fixes #20211

When navigating to or refreshing a table view, there was no visible 
feedback while data was being fetched — the table appeared blank or 
static.

- **Re-fetch scenario** (records already cached): Added a thin animated 
  indigo progress bar at the top of the table that slides while new data 
  loads — users can still see existing data.
- **Initial load** (no cached records): The existing skeleton row loader 
  was already present, but rendered 80 rows unnecessarily. Reduced to 8 
  rows for better performance.

## Changes
- `RecordTable.tsx` — Added animated loading bar overlay for re-fetch state
- `RecordTableBodyLoading.tsx` — Reduced skeleton rows from 80 → 8

## Testing
- Navigate to any table view → skeleton rows shown on first load
- Navigate away and back → thin loading bar at top during re-fetch
- Empty table state unaffected
- Kanban and Calendar views unaffected

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-table/components/RecordTable.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-table/components/RecordTable.tsx
@@ -1,7 +1,6 @@
 import { isNonEmptyString } from '@sniptt/guards';
 import { useRef } from 'react';
 import { styled } from '@linaria/react';
-import { keyframes } from '@linaria/core';
 
 import { useObjectPermissionsForObject } from '@/object-record/hooks/useObjectPermissionsForObject';
 import { hasRecordGroupsComponentSelector } from '@/object-record/record-group/states/selectors/hasRecordGroupsComponentSelector';
@@ -21,18 +20,6 @@ import { useAtomComponentStateValue } from '@/ui/utilities/state/jotai/hooks/use
 import isEmpty from 'lodash.isempty';
 import { themeCssVariables } from 'twenty-ui/theme-constants';
 
-const loadingSlide = keyframes`
-  0% {
-    transform: translateX(-100%);
-  }
-  60% {
-    transform: translateX(0%);
-  }
-  100% {
-    transform: translateX(100%);
-  }
-`;
-
 const StyledLoadingBarWrapper = styled.div`
   height: 2px;
   left: 0;
@@ -44,7 +31,7 @@ const StyledLoadingBarWrapper = styled.div`
 `;
 
 const StyledLoadingBar = styled.div`
-  animation: ${loadingSlide} 1.4s ease-in-out infinite;
+  animation: loading-slide 1.4s ease-in-out infinite;
   background: linear-gradient(
     90deg,
     transparent 0%,
@@ -54,6 +41,18 @@ const StyledLoadingBar = styled.div`
   );
   height: 100%;
   width: 60%;
+
+  @keyframes loading-slide {
+    0% {
+      transform: translateX(-100%);
+    }
+    60% {
+      transform: translateX(0%);
+    }
+    100% {
+      transform: translateX(100%);
+    }
+  }
 `;
 
 const StyledTableWrapper = styled.div`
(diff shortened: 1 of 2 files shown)


martmull:
Please, add screenshots Before and After when opening front PR

packages/twenty-front/src/modules/object-record/record-table/record-table-body/components/RecordTableBodyLoading.tsx
```diff
@@ -25,7 +25,7 @@ export const RecordTableBodyLoading = () => {
 
   return (
     <RecordTableBody>
-      {Array.from({ length: 80 }).map((_, rowIndex) => (
+      {Array.from({ length: 8 }).map((_, rowIndex) => (
```
martmull, line 28: why do you reduce length to 8?

packages/twenty-front/src/modules/object-record/record-table/components/RecordTable.tsx
```diff
@@ -86,13 +131,20 @@ export const RecordTable = () => {
...
+            </StyledLoadingBarWrapper>
+          )}
+          <RecordTableContent
+            tableBodyRef={tableBodyRef}
+            handleDragSelectionStart={handleDragSelectionStart}
+            handleDragSelectionEnd={handleDragSelectionEnd}
+            hasRecordGroups={hasRecordGroups}
+            recordTableId={recordTableId}
+          />
+        </StyledTableWrapper>
```
martmull, line 147: we might flash the whole table during refetch, can you provide a screen recording to check the behavior? Cheers

packages/twenty-front/src/modules/object-record/record-table/record-table-body/components/RecordTableBodyLoading.tsx
ArchitAgarwal04: 80 rows was excessive,  it was rendering thousands of unnecessary DOM nodes that the user could never even see as only 8–12 fit in a typical viewport , so I thought 8-12 rows would be better.

packages/twenty-front/src/modules/object-record/record-table/components/RecordTable.tsx
ArchitAgarwal04: Hi! I understand the concern about flashing. Looking at the implementation, flashing shouldn't occur because of this guard:

const isRefetching = isRecordTableInitialLoading && recordTableHasRecords;

When recordTableHasRecords is true (cached data exists) → only the thin progress bar shows, existing rows stay visible
The skeleton rows only render when recordTableHasRecords is false (no data yet)

So the table content never gets cleared during a refetch. I'm currently unable to run the full stack locally to provide a recording, but I'm happy to adjust the logic if you see a specific scenario where flashing could still occur. Let me know!

Bonapara:
Please join a video recording too

FelixMalfait:
@martmull are you sure the orginal issue is relevant? Looks like AI slop spam

martmull:
> are you sure the orginal issue is relevant? Looks like AI slop spam

indeed, i doubt it

FelixMalfait:
Closing since it was not a Core Team request. Please feel free to add a before/after recording and @Bonapara or @martmull will re-open if it improves the app. Thanks for the contribution and sorry for the confusion

PR #20219, opened 2026-05-04
fix(shared): @types/lodash.camelcase for Docker CD

charlesBochet:
Adds `@types/lodash.camelcase` to `twenty-shared`.

**Why:** `lodash.camelcase` has no bundled types. Those types used to come from the root `devDependencies`; after scoped/hoisted deps (#20140), they are no longer guaranteed in the Docker `common-deps` graph, so `twenty-shared:build` (pulled in before server Lingui) fails with TS7016. Declaring the types on the package that imports `lodash.camelcase` fixes CD.

### DIFF
--- a/packages/twenty-shared/package.json
+++ b/packages/twenty-shared/package.json
@@ -24,6 +24,7 @@
     "@types/babel__preset-env": "^7",
     "@types/handlebars": "^4.1.0",
     "@types/jest": "^30.0.0",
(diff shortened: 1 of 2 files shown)


prastoin (approved):
~~Not sure about the fix here, does this export runtime code ?~~ Discussed it's about fixing dockerfile cd

PR #20227, opened 2026-05-04
Add twenty version validation

### LINKED ISSUE #20204: App ↔ Twenty server version compatibility: declare requirements via engines.twenty
## Problem

There is no link between an app's manifest features and the Twenty server version required to support them. Concretely:

- Server has a real version (`APP_VERSION` env, semver-validated, follows git tags `v2.2.0`)
- twenty-sdk has a real version (independent npm package, `2.2.0`)
- An app pins its SDK in `package.json` (`"twenty-sdk": "2.1.0"`)
- An app's `package.json.engines` exists today only for Node/Yarn

…but **no field anywhere says \"this app requires Twenty server >= X\"**, and the server doesn't refuse to install apps whose features it doesn't understand.

## Failure mode

Ship `twenty-linear@0.1.3` (built against `twenty-sdk@2.2.0`, contains `oauthProviders[]` in its manifest). Install it on a Twenty server that pre-dates the OAuth provider system (PR #20181):

1. `application-install.service.ts` (the older one) iterates `manifest.oauthProviders ?? []` — the field doesn't exist in its types, silently ignored.
2. App installs cleanly. No error.
3. At runtime, `listConnections({ providerName: 'linear' })` fails because the controller doesn't exist.
4. User sees a generic \"Linear is not connected.\" No hint that the platform is too old.

The forward direction (new server + old app) is fine — install service tolerates missing fields. **The painful direction is \"newer manifest features on an older server,\" and we have zero defense against it.** This will get worse with every new SDK feature.

## Proposal — `engines.twenty` in the app's package.json

Mirror the well-trodden `engines.node` idiom:

\`\`\`jsonc
{
  \"name\": \"twenty-linear\",
  \"engines\": {
    \"node\": \"^24.5.0\",
    \"twenty\": \">=2.3.0\"   // ← new
  }
}
\`\`\`

The app developer declares the minimum Twenty server version their app needs — explicit, granular, dev-controlled. No manifest schema version, no capabilities API, no SDK-implied magic. One semver string in a familiar place.

### Required components

1. **Public `GET /api/server/info` endpoint**
   Tiny, unauthenticated:
   \`\`\`jsonc
   { \"version\": \"2.2.0\" }
   \`\`\`
   Existing source of truth: `APP_VERSION` env var. New: `packages/twenty-server/src/engine/core-modules/server-info/server-info.controller.ts` (~15 lines).

2. **SDK CLI pre-deploy check**
   In `packages/twenty-sdk/src/cli/utilities/api/application-api.ts` (or wherever `deploy` runs): hit `/api/server/info`, parse the app's `engines.twenty`, refuse with a clear error if the server is too old. Dev gets the failure at deploy time, before the tarball is uploaded.

3. **Server-side install check (authoritative)**
   In `packages/twenty-server/src/engine/core-modules/application/application-tarball.service.ts` (deploy) and `application-install.service.ts` (install on a different workspace): re-check `manifest.engines?.twenty` against `APP_VERSION`. CLI is best-effort; server is authoritative — covers tarballs uploaded out-of-band, marketplace-installed apps, etc.

4. **Twenty-linear declares its baseline**
   As part of this PR, `packages/twenty-apps/internal/twenty-linear/package.json` gets `\"engines\": { \"twenty\": \">=2.3.0\" }` (or whatever version the OAuth feature ships in).

5. **Documentation**
   Note in app-developer docs: \"if you use a feature added in Twenty vX, set `engines.twenty: \">=X\"`. Otherwise your app may install on older servers and fail at runtime.\"

### Behavior when `engines.twenty` is missing

Treat as \"any version\" (current behavior — backward-compatible). Surface a soft warning at deploy time only: \"No `engines.twenty` declared — your app may install on servers that don't support its features.\"

This is the standard ecosystem behavior (npm, VS Code, WordPress all leave the responsibility on the package author). Could be made stricter later if needed.

## Why not a `manifestSchemaVersion`

Considered and rejected. Reasons:

- Adds yet another versioning concept (the codebase already has app version, SDK version, server version, root semver, package.json `engines`) — more cognitive overhead than upside.
- Same intent as `engines.twenty` but coarser: a schema bump tells you \"some new field exists,\" not \"which feature you actually depend on.\" An app that doesn't use OAuth still gets blocked from older servers if the schema bumped.
- Requires diligence on every SDK PR (\"did you add a top-level field? bump the schema\"). `engines.twenty` puts the version pin where the actual dependency lives — in the consuming app.
- Pure-additive manifest features (no semantic change) don't need a schema bump but probably should warn for older servers.

## What this catches / what it doesn't

**Catches:**
- App built with new feature X, deployed to server that lacks it (the actual current pain).
- Tarball uploaded out-of-band by a developer to an old server.
- Marketplace install attempted on an old server.

**Does not catch (not in scope):**
- App developer who forgets to update `engines.twenty` after using a new feature. Mitigation: docs, soft warning when missing.
- Subtle behavioral changes in features that exist on both old and new (e.g. \"OAuth scope semantics changed in vX.Y\"). Mitigation: this would require either a finer-grained capabilities API or a server-version range constraint per feature — defer until we hit it.

## Critical files

**New:**
- `packages/twenty-server/src/engine/core-modules/server-info/server-info.module.ts`
- `packages/twenty-server/src/engine/core-modules/server-info/server-info.controller.ts`

**Modified:**
- `packages/twenty-server/src/engine/core-modules/application/application-tarball.service.ts` — check at deploy
- `packages/twenty-server/src/engine/core-modules/application/application-install/application-install.service.ts` — check at install
- `packages/twenty-sdk/src/cli/utilities/api/application-api.ts` — pre-deploy check
- `packages/twenty-apps/internal/twenty-linear/package.json` — declare `engines.twenty`
- App-developer docs

## Where this came up

PR #20181 (twenty-linear app, generic OAuth provider). The `oauthProviders[]` manifest field is the first new top-level feature that would silently no-op on older servers. Catching this generically — not just for OAuth — is the right pattern going forward.

martmull:
as title, server version is checked before app deploy, and app install commands

### New section in publishing doc
image


packages/twenty-server/src/engine/core-modules/application/application-install/application-install.service.ts
```diff
@@ -118,6 +120,22 @@ export class ApplicationInstallService {
...
+    if (
+      isDefined(requiredServerVersion) &&
+      isDefined(serverVersion) &&
+      isDefined(semver.valid(serverVersion)) &&
+      !semver.satisfies(serverVersion, requiredServerVersion)
+    ) {
+      throw new ApplicationException(
+        `App requires Twenty server ${requiredServerVersion} but this server is ${serverVersion}.`,
+        ApplicationExceptionCode.SERVER_VERSION_INCOMPATIBLE,
+      );
```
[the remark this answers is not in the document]
prastoin: @martmull

packages/twenty-server/src/engine/core-modules/application/application-install/application-install.service.ts
```diff
@@ -118,6 +120,22 @@ export class ApplicationInstallService {
...
+      resolvedPackage.packageJson.engines?.['twenty'];
+    const serverVersion = this.twentyConfigService.get('APP_VERSION');
+
+    if (
+      isDefined(requiredServerVersion) &&
+      isDefined(serverVersion) &&
+      isDefined(semver.valid(serverVersion)) &&
+      !semver.satisfies(serverVersion, requiredServerVersion)
+    ) {
+      throw new ApplicationException(
```
prastoin, line 149: ***Remark:*** I feel like we're merging three different errors here
- Invalid manifest twenty engine semver
- Invalid twenty-instance APP_VERSION ( can occur for self host building their custom image ) ps: we should also run the semver.valid on the `APP_VERSION`
- Incompatible peer deps

packages/twenty-server/src/engine/core-modules/application/application-registration/application-tarball.service.ts
```diff
@@ -75,6 +78,21 @@ export class ApplicationTarballService {
...
+        isDefined(requiredServerVersion) &&
+        isDefined(serverVersion) &&
+        isDefined(semver.valid(serverVersion)) &&
+        !semver.satisfies(serverVersion, requiredServerVersion)
+      ) {
+        throw new ApplicationRegistrationException(
+          `App requires Twenty server ${requiredServerVersion} but this server is ${serverVersion}.`,
+          ApplicationRegistrationExceptionCode.SERVER_VERSION_INCOMPATIBLE,
+        );
+      }
```
prastoin, line 110: Lets create a specific service that centralize the logic in order to avoid any inconsistency in the checking + keep the same granularity over exceptions

prastoin (commented):
Nice ! Left two comments

prastoin (approved):

prastoin (commented):
~~Already fixed on main~~ nevermind was a merge commit

PR #20228, opened 2026-05-04
feat(server): OTLP metrics export logs for troubleshooting

charlesBochet:
## Summary

Adds **grep-friendly** `console` logging around the OpenTelemetry metrics OTLP exporter in [`packages/twenty-server/src/instrument.ts`](packages/twenty-server/src/instrument.ts) so production / staging can confirm whether the app is exporting metrics and why exports fail.

## Log format

- Prefix: **`[Twenty OTEL metrics]`** (easy to filter in Loki / `kubectl logs | grep`).
- **Startup:** whether the OTLP reader is enabled, `exportIntervalMs`, and endpoint as `protocol//host/path` only (no credentials).
- **First successful export:** one `console.log` per process (`first export ok`) with metric data point count — avoids spamming every 10s.
- **Each failed export:** `console.warn` with result code, point count, and serialized error (including nested `AggregateError` causes when present).

### DIFF
--- a/packages/twenty-server/src/instrument.ts
+++ b/packages/twenty-server/src/instrument.ts
@@ -1,5 +1,6 @@
 import process from 'process';
 
+import { ExportResultCode } from '@opentelemetry/core';
 import opentelemetry from '@opentelemetry/api';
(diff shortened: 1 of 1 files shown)


Weiko (approved):
LGTM

PR #20229, opened 2026-05-04
fix: show active advanced filter count badge in dropdown button

### LINKED ISSUE #20207: Advanced filter button doesn't show count of active rules
Scope & Context
The filter dropdown has an "Advanced filter" button at the bottom. When a user has active advanced filter rules, the button is supposed to show a pill badge with the count of those rules — similar to how AdvancedFilterChip shows "2 advanced rules" in the view bar. Right now the count is hardcoded to 0 so the pill never appears, even when there are active advanced filters.

Current behavior
When you open the filter dropdown on any record table (e.g. People, Companies), the "Advanced filter" menu item at the bottom never shows a count badge, regardless of how many advanced filter rules are active. The pill component is rendered in the code but the count driving it is hardcoded to 0.

// ViewBarFilterDropdownAdvancedFilterButton.tsx line 37
const advancedFilterQuerySubFilterCount = 0; // TODO

Expected behavior
The "Advanced filter" button should show a pill badge with the number of active top-level advanced filter rules when at least one rule exists — the same number shown on the AdvancedFilterChip in the view bar. So if there are 2 active advanced rules, the button reads "Advanced filter" with a blue pill showing "2".

Technical inputs
In packages/twenty-front/src/modules/views/components/ViewBarFilterDropdownAdvancedFilterButton.tsx, replace the hardcoded advancedFilterQuerySubFilterCount = 0 with a real count derived from state
Import rootLevelRecordFilterGroupComponentSelector from @/object-record/advanced-filter/states/rootLevelRecordFilterGroupComponentSelector and useAtomComponentSelectorValue from @/ui/utilities/state/jotai/hooks/useAtomComponentSelectorValue — these are already used in AdvancedFilterChip.tsx for the same purpose
Import useChildRecordFiltersAndRecordFilterGroups from @/object-record/advanced-filter/hooks/useChildRecordFiltersAndRecordFilterGroups
Get the root filter group via useAtomComponentSelectorValue(rootLevelRecordFilterGroupComponentSelector), pass its id into useChildRecordFiltersAndRecordFilterGroups, and use childRecordFiltersAndRecordFilterGroups.length as the count
AdvancedFilterChip.tsx (same directory level) already does this exact computation at lines 43–50 — the fix is just wiring the same hooks into this component

wadeKeith:
## Summary

Replaces the hardcoded `0` in `ViewBarFilterDropdownAdvancedFilterButton` with the actual count of active advanced filter rules, matching the behavior of `AdvancedFilterChip` in the view bar.

## What changed

In `packages/twenty-front/src/modules/views/components/ViewBarFilterDropdownAdvancedFilterButton.tsx`:

- Imported `useAtomComponentSelectorValue` and `rootLevelRecordFilterGroupComponentSelector`
- Imported `useChildRecordFiltersAndRecordFilterGroups`
- Replaced `const advancedFilterQuerySubFilterCount = 0; // TODO` with the real computed count via the same hook pattern used in `AdvancedFilterChip.tsx`

The pill badge will now appear on the "Advanced filter" dropdown menu item showing the number of active advanced filter rules (e.g. "2" when two rules are active).

## References

- Fixes #20207

### DIFF
--- a/packages/twenty-front/src/modules/views/components/ViewBarFilterDropdownAdvancedFilterButton.tsx
+++ b/packages/twenty-front/src/modules/views/components/ViewBarFilterDropdownAdvancedFilterButton.tsx
@@ -11,6 +11,7 @@ import { useAtomComponentSelectorValue } from '@/ui/utilities/state/jotai/hooks/
 import { VIEW_BAR_FILTER_BOTTOM_MENU_ITEM_IDS } from '@/views/constants/ViewBarFilterBottomMenuItemIds';
 
 import { useChildRecordFiltersAndRecordFilterGroups } from '@/object-record/advanced-filter/hooks/useChildRecordFiltersAndRecordFilterGroups';
+import { useSetRecordFilterUsedInAdvancedFilterDropdownRow } from '@/object-record/advanced-filter/hooks/useSetRecordFilterUsedInAdvancedFilterDropdownRow';
 import { rootLevelRecordFilterGroupComponentSelector } from '@/object-record/advanced-filter/states/rootLevelRecordFilterGroupComponentSelector';
 import { useCreateEmptyRecordFilterFromFieldMetadataItem } from '@/object-record/record-filter/hooks/useCreateEmptyRecordFilterFromFieldMetadataItem';
 import { useCloseDropdown } from '@/ui/layout/dropdown/hooks/useCloseDropdown';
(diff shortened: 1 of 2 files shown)


charlesBochet:
Hi @wadeKeith, thanks for opening a PR. Right directoin but I wonder if you tried the result:

image

We can obviously not let it like this in the product ; I'll take over. Appreciate the help and please check the result of your change next time

charlesBochet (approved):

PR #20234, opened 2026-05-04
Fix unclear metadata validation errors

abdulrahmancodes:
https://github.com/user-attachments/assets/8f8f1122-3de1-4a9b-8bb4-a3c8d31e47ae


packages/twenty-server/src/engine/core-modules/graphql/hooks/use-graphql-error-handler.hook.ts
```diff
@@ -225,10 +225,14 @@ export const useGraphQLErrorHandlerHook = <
                       ...error,
                       extensions: {
                         ...error.extensions,
-                        userFriendlyMessage: i18n._(
-                          error.extensions.userFriendlyMessage ??
-                            defaultErrorMessage,
-                        ),
+                        userFriendlyMessage:
+                          typeof error.extensions.userFriendlyMessage ===
```
charlesBochet: could you explain why we need this? It does not look correct to me

packages/twenty-server/src/engine/workspace-manager/workspace-migration/interceptors/utils/get-primary-metadata-validation-user-friendly-message.util.ts
```diff
@@ -0,0 +1,32 @@
...
+    const failedValidations = metadataValidation.errors[metadataName];
+
+    if (!isDefined(failedValidations) || failedValidations.length === 0) {
+      continue;
+    }
+
+    for (const failedValidation of failedValidations) {
+      for (const validationError of failedValidation.errors) {
+        if (isNonEmptyString(validationError.userFriendlyMessage)) {
+          return validationError.userFriendlyMessage;
```
charlesBochet: so it takes the last one? Why?
Also, let's cover this util with tests

packages/twenty-server/src/engine/workspace-manager/workspace-migration/interceptors/workspace-migration-builder-exception-formatter.ts
```diff
@@ -32,15 +35,22 @@ export const workspaceMigrationBuilderExceptionFormatter = (
...
 
+  const primaryUserFriendlyMessage =
+    getPrimaryMetadataValidationUserFriendlyMessage({
+      errors,
+      summary,
+    });
+
   throw new BaseGraphQLError(
-    error.message,
+    primaryUserFriendlyMessage ?? error.message,
```
charlesBochet: I don't think it's what we want here, message should not be the userFriendlyMessage

packages/twenty-server/src/engine/workspace-manager/workspace-migration/interceptors/workspace-migration-builder-exception-formatter.ts
```diff
@@ -32,15 +35,22 @@ export const workspaceMigrationBuilderExceptionFormatter = (
...
     ErrorCode.METADATA_VALIDATION_FAILED,
     {
       code: 'METADATA_VALIDATION_ERROR',
       errors,
       summary,
-      message,
-      userFriendlyMessage: msg`Metadata validation failed`,
+      message: validationSummaryMessage,
+      userFriendlyMessage:
+        primaryUserFriendlyMessage ?? msg`Metadata validation failed`,
```
charlesBochet: is primaryUserFriendlyMessage a "msg" itself?

charlesBochet (changes requested):
Left comments

packages/twenty-server/src/engine/core-modules/graphql/hooks/use-graphql-error-handler.hook.ts
abdulrahmancodes: `userFriendlyMessage` is sometimes a lingui descriptor, and sometimes the server already sends a plain string (final validation text with dynamic bits). `i18n._()` is for descriptors, not arbitrary runtime strings, so when it’s already a string we pass it through unchanged

FelixMalfait: > sometimes the server already sends a plain string 

Seems like this is the root cause you should try to fix?

packages/twenty-server/test/integration/graphql/suites/view/view-field/__snapshots__/object-identifier-update-side-effect-on-view-field.integration-spec.ts.snap
```diff
@@ -177,9 +177,9 @@ exports[`View Field Resolver - Successful object metadata identifier update side
       "totalErrors": 1,
       "viewField": 1,
     },
-    "userFriendlyMessage": "Metadata validation failed",
+    "userFriendlyMessage": "Label identifier view field has to be in the lowest position",
```
FelixMalfait, line 180: Thanks for these fixes! It'll be helpful

packages/twenty-server/src/engine/workspace-manager/workspace-migration/interceptors/utils/get-primary-metadata-validation-user-friendly-message.util.ts
abdulrahmancodes: It doesn’t take the last one.  We return on the first non-empty `userFriendlyMessage` while walking `ALL_METADATA_NAME` in order, then each failed validation and its errors array. So it’s first match, not last.
We only do this when `totalErrors === 1` so we can show a single clear message; otherwise we return `undefined` and the caller can fall back to a generic / multi-error handling

charlesBochet: yes my mistake, we return the first one.

I'm not a fan of this implementation:
1) getPrimaryMetadataValidationUserFriendlyMessage => getMetadataValidationUserFriendlyMessage and let's make it always return something. If totalError > 1 we should return msg'Many validation errors'. If it's === 1 we should return the first one
2) userFriendlyMesages should always be a msg (I'm not sure about how msg is typed, if it's considered as a string or not but I'm worried about this isNonEmptyString here). The emitter should always return msg and then getMetadataValidationUserFriendlyMessage should always return a msg

packages/twenty-server/src/engine/workspace-manager/workspace-migration/interceptors/workspace-migration-builder-exception-formatter.ts
charlesBochet: See my latest comment

packages/twenty-server/src/engine/core-modules/graphql/hooks/use-graphql-error-handler.hook.ts
charlesBochet: yes, agree: See my latest comment. The emitter of the CustomException (the one setting the userFriendlyMessage should emit a msg IMO)

charlesBochet (approved):

charlesBochet:
Nice work!

PR #19890, opened 2026-04-20
Translate standard page layouts

Weiko:
## Context
Standard page layout tabs, page layout widgets, and view field group titles were hardcoded English in the backend. This PR brings them under the same translation pipeline as views.

Notes: Once a standard widget/tab/section title is overriden, the backend returns its value without translation

PR #20235, opened 2026-05-04
Fix empty record page on system objects for non-English workspace members

### LINKED ISSUE #20102: System object record pages render empty for non-English locales (e.g. message thread email body not shown)
## Bug Description

For workspace members whose locale is not English (e.g. `ja-JP`, `fr-FR`), record pages of system objects (such as `messageThread`) render an empty content area. No tabs appear, the active widget never mounts, and underlying GraphQL queries (e.g. `FindManyMessages` for an email thread) are never sent.

The most user-visible symptom is "the email body does not show on the message thread detail page", but the same root cause affects every system object's record page in non-English locales.

### Steps to reproduce
1. Set the workspace member locale to `ja-JP` (Settings → Experience).
2. Open any `messageThread` detail page (`/object/messageThread/<id>`).
3. Observe that the page body is an empty `<div>` — no tabs, no widgets, no email body.

Switching the locale back to English makes the page render correctly.

## Expected behavior

System object record pages (message threads, etc.) should render their tabs and widgets regardless of the user's locale. The email body, timeline, and other widgets should be visible for `ja-JP`, `fr-FR`, and any other non-English locale, just as they are for English.

### Root cause

`packages/twenty-front/src/modules/page-layout/components/PageLayoutTabsRenderer.tsx:105-111` filters tabs for system objects against hardcoded **English** titles:

```ts
const SYSTEM_OBJECT_TABS = ['Home', 'Timeline', 'Overview', 'Flow'];

const tabsForCurrentObject = isSystemObject
  ? tabsWithVisibleWidgets.filter((tab) =>
      SYSTEM_OBJECT_TABS.includes(tab.title),
    )
  : tabsWithVisibleWidgets;
```

However, the server-side `title` field resolver in `packages/twenty-server/src/engine/metadata-modules/page-layout-tab/resolvers/page-layout-tab.resolver.ts:48-68` returns the **translated** title via Lingui:

```ts
return resolvePageLayoutTabTitle({
  title: tab.title,
  applicationId: tab.applicationId,
  twentyStandardApplicationId: twentyStandardFlatApplication.id,
  overrides: tab.overrides,
  i18nInstance: i18n,
});
```

For `ja-JP`, `'Home'` → `'ホーム'`, so `['Home', ...].includes('ホーム')` is `false`. All tabs are filtered out, `tabsForCurrentObject` becomes `[]`, no `activeTabId` is set, and `<PageLayoutMainContent />` is never rendered. As a downstream effect, `EmailThreadWidget` never mounts and `useEmailThread` never runs (the `FindManyMessages` query is not even sent).

The DB / message data itself is fine — the issue is purely on the frontend rendering path.

### Suggested fix

Filter on a locale-independent attribute. Each standard tab has a stable `icon` field defined in `packages/twenty-server/src/engine/workspace-manager/twenty-standard-application/constants/standard-page-layout-tabs.template.ts` (e.g. `IconHome`, `IconTimelineEvent`, `IconSettings`). Switching the filter from `tab.title` to `tab.icon` (or `tab.position`) resolves the issue without touching the server resolver.

### Environment
- Reproduced on Twenty `v2.0.3`.
- The offending filter is unchanged on `main` as of this report, so it should reproduce on `v2.1.0` as well.
- Affects any non-English locale.

### Workaround
Temporarily switching the workspace member's language back to English (Settings → Experience) makes the widget render and the email body display.

Weiko:
Fixes https://github.com/twentyhq/twenty/issues/20102

## Context
Since #19890 (Translate standard page layouts), the server's `PageLayoutTab.title` resolver translates standard tab titles at query time. A workspace member in French viewing a `messageChannel` (or any system object with a standard page layout) receives `tab.title = "Accueil"` / `"Chronologie"` instead of `"Home"` / `"Timeline"`.

The `SYSTEM_OBJECT_TABS` guard in `PageLayoutTabsRenderer` was comparing against an English-only literal allow-list, so every tab was dropped, `sortedTabs` became empty, and `<PageLayoutMainContent />` never mounted — the record page rendered blank (no fields, timeline, email thread, etc.).

## Fix
Only run the allow-list filter when the resolved layout is the synthetic `DEFAULT_RECORD_PAGE_LAYOUT` (the client-side fallback for the few system objects with no server-side standard page layout config, e.g. `workspaceMember`, `attachment`, `message`). That layout ships hardcoded English tabs, so the English allow-list still works in every locale.

System objects that do have a server-side standard page layout (`messageChannel`, `connectedAccount`, `workflowRun`, …) are no longer filtered at all, the server only ever persists Home/Timeline/Flow tabs for them, so no filter is needed.

## Before
Screenshot 2026-05-04 at 15 15 54

## After
Screenshot 2026-05-04 at 15 15 36

## Note
Next step should be to backfill those system objects with real record page layouts so we can remove this filter logic

### DIFF
--- a/packages/twenty-front/src/modules/page-layout/components/PageLayoutTabsRenderer.tsx
+++ b/packages/twenty-front/src/modules/page-layout/components/PageLayoutTabsRenderer.tsx
@@ -3,6 +3,7 @@ import { type FlatObjectMetadataItem } from '@/metadata-store/types/FlatObjectMe
 import { PageLayoutLeftPanel } from '@/page-layout/components/PageLayoutLeftPanel';
 import { PageLayoutTabList } from '@/page-layout/components/PageLayoutTabList';
 import { PageLayoutTabListEffect } from '@/page-layout/components/PageLayoutTabListEffect';
+import { DEFAULT_RECORD_PAGE_LAYOUT_ID } from '@/page-layout/constants/DefaultRecordPageLayoutId';
 import { PAGE_LAYOUT_LEFT_PANEL_CONTAINER_WIDTH } from '@/page-layout/constants/PageLayoutLeftPanelContainerWidth';
 import { useCurrentPageLayoutOrThrow } from '@/page-layout/hooks/useCurrentPageLayoutOrThrow';
 import { useIsPageLayoutInEditMode } from '@/page-layout/hooks/useIsPageLayoutInEditMode';
@@ -25,7 +26,6 @@ import { styled } from '@linaria/react';
 import { isDefined } from 'twenty-shared/utils';
 import { useIsMobile } from 'twenty-ui/utilities';
 import { FeatureFlagKey } from '~/generated-metadata/graphql';
-
 const StyledContainer = styled.div<{ hasPinnedTab: boolean }>`
   display: grid;
   grid-template-columns: ${({ hasPinnedTab }) =>
@@ -104,11 +104,15 @@ export const PageLayoutTabsRenderer = () => {
 
   const SYSTEM_OBJECT_TABS = ['Home', 'Timeline', 'Overview', 'Flow'];
 
-  const tabsForCurrentObject = isSystemObject
-    ? tabsWithVisibleWidgets.filter((tab) =>
-        SYSTEM_OBJECT_TABS.includes(tab.title),
-      )
-    : tabsWithVisibleWidgets;
+  const isUsingSyntheticDefault =
+    currentPageLayout.id === DEFAULT_RECORD_PAGE_LAYOUT_ID;
+
+  const tabsForCurrentObject =
+    isSystemObject && isUsingSyntheticDefault
+      ? tabsWithVisibleWidgets.filter((tab) =>
+          SYSTEM_OBJECT_TABS.includes(tab.title),
+        )
+      : tabsWithVisibleWidgets;
 
   const { tabsToRenderInTabList, pinnedLeftTab } = getTabsByDisplayMode({
     tabs: tabsForCurrentObject,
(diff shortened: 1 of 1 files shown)


packages/twenty-front/src/modules/page-layout/components/PageLayoutTabsRenderer.tsx
```diff
@@ -104,11 +104,15 @@ export const PageLayoutTabsRenderer = () => {
 
   const SYSTEM_OBJECT_TABS = ['Home', 'Timeline', 'Overview', 'Flow'];
```
martmull, line 105: how is those changes are fixing the `non-English workspace members` issue?

martmull (commented):
don't understand how the changes fixes the "non-english" issue

packages/twenty-front/src/modules/page-layout/components/PageLayoutTabsRenderer.tsx
Weiko: When a page layout is missing, the FE tries to fallback to a "mock" layout with a common config ("tasks", "home", etc...)
However a system object shouldn't have "Tasks", "Notes", etc... so we were filtering out those tabs simply based on their names which is fine when we are in the "mock" case but now the backend returns tab as translated (so "Timeline" would be "Chronologie" in FR for example) so the filter wouldn't work properly in this case. The PR fixes that by ignoring the filter when a page layout exists (since the source of truth becomes the BE) and only remain for the "mock" fallback.
As mentioned in the description, this logic is a bit legacy and only there for system objects that don't have a record page layout but it shouldn't be the case as we need to backfill those in the DB at some point.
I'll try to add commands for the next release to backfill but for the time being this fix is a simple 1 liner that should be enough @martmull

martmull (approved):

PR #20239, opened 2026-05-04
Add check for manifest uuid version

### LINKED ISSUE #20199: Apps: split manifest validation into static-vs-workspace tiers and share between SDK + server
## Problem

Manifest validation today is split inconsistently between two places:

- **twenty-sdk** \`defineX\` factories (\`packages/twenty-sdk/src/sdk/define/**/*.ts\`) — partial, mostly \"is this field present\" checks. Runs at \`twenty deploy\` build time.
- **twenty-server** \`application-sync.service.ts\` — does the heavy structural validation (UUID format, name regex, cross-references). Runs only at \`install\` time, against a workspace.

This has three concrete symptoms:

1. **Late surfacing**: an app can be successfully \`twenty deploy\`-ed (registration created on the server) with structurally invalid content. Every workspace install then fails identically. Hit while testing PR #20181: invalid-v4 UUIDs in \`twenty-linear\`'s \`universal-identifiers.ts\` slipped through registration creation and only blew up on \`twenty install\`.

   \`\`\`
   Sync failed with 3 errors
   logicFunction: 2 errors
     1. ENTITY_MALFORMED: Invalid universalIdentifier: \"be9f3a7d-8c2e-4f66-d03e-5b0d7f1c6a8b\" is not a valid UUID, uuid version should be greater than 4
     2. ENTITY_MALFORMED: Invalid universalIdentifier: \"cf0a4b8e-9d3f-4a77-e14f-6c1e8a2d7b9c\" is not a valid UUID, uuid version should be greater than 4
   role: 1 error
     1. ENTITY_MALFORMED: Invalid universalIdentifier: \"ad8e2f6c-7b1d-4e55-cf2d-4a9c6e0b5f7a\" is not a valid UUID, uuid version should be greater than 4
   \`\`\`

2. **No local feedback**: \`twenty deploy\` does \`appBuild\` (which runs typecheck and packs the tarball) but doesn't run the same structural validation the server eventually applies. The dev only finds out after a full upload + install round-trip.

3. **Drift risk**: two places that need to stay in lockstep but live in different packages, owned by different services.

### (a) Two clearly-separated validation tiers

| Tier | Examples | When it runs |
|---|---|---|
| **Manifest-static** (no workspace context) | UUID v4+ format, name regex (\`/^[a-z][a-z0-9-]*$/\`), scope literals (\`'user' \\| 'workspace'\`), cross-references (\`clientIdVariable\` matches a declared \`applicationVariable\` key, \`defaultRoleUniversalIdentifier\` matches one of the declared roles), required-field presence | \`deploy\` (registration creation) — AND inline in SDK \`appBuild\` so it fails before upload |
| **Workspace-context** | universal-identifier collisions with already-installed apps, env-var presence, role-name conflicts with workspace seed roles, permission-grant cross-checks | \`install\` |

The current \`application-sync.service.ts\` validation should be split along this line; everything in the first column moves earlier.

### (b) Shared validator implementation

Extract the structural validation into a pure function in \`twenty-shared/application\` (e.g. \`validateAppManifest(manifest): ValidationResult\`). Call it from:

- \`twenty-sdk/src/cli/operations/build.ts\` (so \`twenty deploy\` and \`twenty dev\` fail locally before the network call)
- \`twenty-server\` registration controller (so a server reject happens at upload, not at install)
- \`twenty-server\` install service (defense in depth — same call, no behavior change)

Single source of truth → no drift.

### (c) Auto-generate UUIDs in scaffolds

The \`twenty add <entity>\` commands already generate UUIDs for some entities. Audit and make sure every scaffold that emits a \`universalIdentifier\` uses \`uuidv4()\` (not a hand-written placeholder). Removes the temptation to hand-author and fail (b) at all.

## Out of scope

- Schema-level validation of object/field metadata against the workspace's existing schema (that's a different surface, with very different rules).
- Versioning / breaking-change checks across upgrades.

## Acceptance

- Running \`twenty deploy\` on an app with an invalid universal-identifier UUID or a name that doesn't match the regex fails at the SDK CLI step (before any network call), with the same error wording the server would have produced.
- Running \`twenty deploy\` against a server that has stricter checks than the local SDK validator (e.g. workspace-context checks) is the only path where errors appear post-upload.
- \`twenty add\` for any entity type produces a manifest stub that always passes \`validateAppManifest\`.

martmull:
As title

image


packages/twenty-sdk/src/cli/utilities/build/manifest/manifest-validate.ts
```diff
@@ -99,16 +103,50 @@ const validateRelationFields = (
...
-  const duplicates = extractDuplicates(findUniversalIdentifiers(manifest));
+  const universalIdentifiers = findUniversalIdentifiers(manifest);
+
+  const duplicates = extractDuplicates(universalIdentifiers);
 
   if (duplicates.length > 0) {
     errors.push(`Duplicate universal identifiers: ${duplicates.join(', ')}`);
   }
 
+  errors.push(...validateUniversalIdentifierVersions(universalIdentifiers));
```
charlesBochet: I would prefer something in the style of duplicates:
const invalidUniversalIdentifiers = getInvalidUniversal...

if (invalidUniversalIdentifiers > 0) { push }

charlesBochet: Also, I think the validation is not only about version so we should remove this from external naming => validateUniversalIdentifiers (or getInvalidUniversalIdentifiers if we return the list)

charlesBochet (approved):
Approved with comment

PR #20241, opened 2026-05-04
Bump twenty current version

prastoin:
# Introduction
This PR introduces a workflow and nx command that allow bumping to a given version or incrementing the current `TWENTY_CURRENT_VERSION`

Combined with accurate on point cd triggered and CI upgrade sequence guard mutation workflow the window where a PR can corrupt an already released twenty version is mitigated


packages/twenty-server/scripts/bump-version.ts
```diff
@@ -0,0 +1,251 @@
...
+    parseVersion(newVersionArg);
+
+    if (!semver.gt(newVersionArg, currentVersion)) {
+      throw new Error(
+        `New version '${newVersionArg}' must be greater than current version '${currentVersion}'`,
+      );
+    }
+
+    if (nextVersions.length > 0) {
+      const highestNextVersion = nextVersions[nextVersions.length - 1];
```
Weiko, line 126: Should we "sort" this array? I guess this should always written in the correct order though, up to you

packages/twenty-server/scripts/bump-version.ts
```diff
@@ -0,0 +1,251 @@
...
+    ...skippedNextVersions,
+  ];
+  const updatedNextVersions =
+    remainingNextVersions.length > 0
+      ? remainingNextVersions
+      : [incrementMinorVersion(newVersion)];
+
+  console.log('\nApplying changes:');
+
+  updatePreviousVersions(updatedPreviousVersions);
```
Weiko, line 226: Should we try/catch rollback if one of the 3 fail to avoid broken state? This seems overkill for this kind of command so up to you

Weiko (approved):
LGTM

packages/twenty-server/scripts/bump-version.ts
prastoin: I assume it is

packages/twenty-server/scripts/bump-version.ts
prastoin: overkill to me as it's still a PR that needs to be reviewed anw

PR #20242, opened 2026-05-04
Fix unreliable SSE event stream updates during workflow form transitions

thomtrp:
Before - workflow run not up to date, needs refresh to see created company in some cases

https://github.com/user-attachments/assets/28517e97-2404-4f75-8bce-cc33e3cbea20

After 

https://github.com/user-attachments/assets/60f930cb-1265-4c50-8ec5-aa4f978b1873

## Summary
- Split `SSEQuerySubscribeEffect`'s single debounced `updateQueryListeners` into separate `syncAdditions` (leading edge, 1s debounce) and `syncRemovals` (trailing edge, 200ms debounce) callbacks. This prevents query unregistrations during component mount/unmount transitions from creating gaps where events are missed, while keeping new registrations immediate.
- Each sync path now updates `activeQueryListenersState` granularly (append-only for additions, filter-only for removals) instead of overwriting the entire state, eliminating a race condition where removals could mark unregistered queries as active.
- Mount `WorkflowRunSSESubscribeEffect` inside `WorkflowEditActionFormFiller` so the workflow-run query subscription stays active during form steps.
- Extract `buildSortedConnectionEdges` util that builds the resulting edge list of a cached record connection after new records are created. Position placeholders (`'first'` / `'last'`) bypass orderBy and are pinned to the front/back; sortable positions (numeric or undefined) are merged into existing edges and sorted by the connection's actual `orderBy`. This replaces the broken `length * position` insertion logic in `triggerCreateRecordsOptimisticEffect` that treated the sortable `position` field as a 0-1 ratio, causing new records from SSE to land at invisible indices in the cached list. Also fixes `totalCount` increment for batched creates, derives `pageInfo` cursors from the final array, and gracefully skips records whose `toReference` returns null.

## Test plan
- [x] Run a workflow with a form step — verify the workflow status updates live after form submission (no stuck "running" state)
- [x] Run the same workflow multiple times — verify company creation events appear live on the record index page for every run, not just the first
- [x] Click the "+" button to create a record in first position — verify it appears immediately at the top
- [x] Verify other SSE-backed live updates (record creation, deletion, updates) still work correctly

### DIFF
--- a/packages/twenty-front/src/modules/apollo/optimistic-effect/utils/triggerCreateRecordsOptimisticEffect.ts
+++ b/packages/twenty-front/src/modules/apollo/optimistic-effect/utils/triggerCreateRecordsOptimisticEffect.ts
@@ -1,6 +1,7 @@
 import { type ApolloCache, type StoreObject } from '@apollo/client';
 import { isNonEmptyString } from '@sniptt/guards';
 
+import { sortCachedObjectEdges } from '@/apollo/optimistic-effect/utils/sortCachedObjectEdges';
 import { triggerUpdateRelationsOptimisticEffect } from '@/apollo/optimistic-effect/utils/triggerUpdateRelationsOptimisticEffect';
 import { type EnrichedObjectMetadataItem } from '@/object-metadata/types/EnrichedObjectMetadataItem';
 import { type RecordGqlRefEdge } from '@/object-record/cache/types/RecordGqlRefEdge';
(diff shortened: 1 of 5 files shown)


packages/twenty-front/src/modules/apollo/optimistic-effect/utils/triggerCreateRecordsOptimisticEffect.ts
```diff
@@ -226,9 +227,21 @@ export const triggerCreateRecordsOptimisticEffect = ({
           return rootQueryCachedObjectRecordConnection;
         }
 
+        const rootQueryOrderBy = rootQueryVariables?.orderBy;
+
+        let sortedEdges = nextRootQueryCachedRecordEdges;
+
+        if (Array.isArray(rootQueryOrderBy) && rootQueryOrderBy.length > 0) {
```
charlesBochet: let's introduce an util: getSortedEdges({ newRecord, currentEdges, orderBy }) that would handle both sort and position.

We should also make sure that isRecordMatchingFilter is taking cursor and offset into account to not add the same record to the same query UNLESS apollo stores it as one query

charlesBochet (approved):

charlesBochet (approved):

PR #20246, opened 2026-05-04
New name not appearing when renaming "Stages" in data-model settings

Bonapara:
## Summary
- Resolve standard field `label`, `description`, and `icon` overrides through dedicated GraphQL field resolvers.
- Fall back to the source locale safely when the request locale is missing, and allow direct overrides to apply for non-source locales when translations are absent.
- Enrich metadata subscription payloads for both `before` and `after`, reusing the same override application path for field and object metadata.
- Update and extend tests to cover the revised override behavior.

## Testing
- Updated unit coverage for standard override resolution, including the non-source-locale fallback path.
- Not run (not requested).

### DIFF
--- a/packages/twenty-server/src/engine/metadata-modules/field-metadata/field-metadata.resolver.ts
+++ b/packages/twenty-server/src/engine/metadata-modules/field-metadata/field-metadata.resolver.ts
@@ -22,8 +22,15 @@ import { RelationDTO } from 'src/engine/metadata-modules/field-metadata/dtos/rel
 import { UpdateOneFieldMetadataInput } from 'src/engine/metadata-modules/field-metadata/dtos/update-field.input';
 import { FieldMetadataService } from 'src/engine/metadata-modules/field-metadata/services/field-metadata.service';
 import { fieldMetadataGraphqlApiExceptionHandler } from 'src/engine/metadata-modules/field-metadata/utils/field-metadata-graphql-api-exception-handler.util';
+import { resolveFieldMetadataStandardOverride } from 'src/engine/metadata-modules/field-metadata/utils/resolve-field-metadata-standard-override.util';
 import { fromFlatFieldMetadataToFieldMetadataDto } from 'src/engine/metadata-modules/flat-field-metadata/utils/from-flat-field-metadata-to-field-metadata-dto.util';
 import { PermissionsGraphqlApiExceptionFilter } from 'src/engine/metadata-modules/permissions/utils/permissions-graphql-api-exception.filter';
+
+// Keep @Parent() structurally typed so ResolverValidationPipe does not validate
+// FieldMetadataDTO date decorators on already-loaded parent records.
+type FieldMetadataStandardOverrideParent = Parameters<
+  typeof resolveFieldMetadataStandardOverride
+>[0];
 
 @UseGuards(WorkspaceAuthGuard)
 @UsePipes(ResolverValidationPipe)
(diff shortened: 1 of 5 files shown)


packages/twenty-server/src/engine/metadata-modules/field-metadata/utils/resolve-field-metadata-standard-override.util.ts
```diff
@@ -26,21 +28,17 @@ export const resolveFieldMetadataStandardOverride = (
...
     const translationValue =
-      fieldMetadata.standardOverrides.translations[locale]?.[labelKey];
+      fieldMetadata.standardOverrides.translations[safeLocale]?.[labelKey];
 
     if (isDefined(translationValue)) {
       return translationValue;
     }
   }
 
-  if (
```
charlesBochet, line 40: are you sure about this one? this will load the standardOverride (non translated) no matter what the locale is and bypass the fallback translation mecanism we have in place. => let's say rename "Stage" in "Status". As "Status" is translated in our files, Status will be translated in all languages.

this might be the right  way but (I actually prefer it) as we don't have translation UI, this was the accepted trade-off at the time, I'm not sure to understand why it's reconsidered now. If reconsidered, I'm not against it but this should be carefully considered as this will change behavior for users. I wonder if we should not rather wait to have a full translation UI to avoid confusing the users

packages/twenty-server/src/engine/metadata-modules/field-metadata/field-metadata.resolver.ts
```diff
@@ -38,6 +45,51 @@ export class FieldMetadataResolver {
...
   ) {}
 
+  @ResolveField(() => String, { nullable: true })
+  async label(
+    @Parent() fieldMetadata: FieldMetadataStandardOverrideParent,
+    @Context() context: I18nContext,
+  ): Promise<string> {
+    const i18n = this.i18nService.getI18nInstance(context.req.locale);
+
+    return resolveFieldMetadataStandardOverride(
```
charlesBochet, line 55: I guess this is mimicking the behavior of the dataLoder of fieldList. I'm not against it but I'm surprised you need to fix it, the frontend should not use it

packages/twenty-server/src/engine/subscriptions/metadata-event/metadata-event-publisher.ts
```diff
@@ -80,25 +81,37 @@ export class MetadataEventPublisher {
...
+        'after' in enrichedProperties &&
+        isDefined(enrichedProperties.after)
+      ) {
+        const enrichedAfter = enrichFieldMetadataEventWithRelations({
+          record: enrichedProperties.after as Record<string, unknown>,
+          flatFieldMetadataMaps,
+          flatObjectMetadataMaps,
+        });
+
+        enrichedProperties.after = this.applyStandardOverridesToMetadataRecord(
```
charlesBochet, line 106: we usually avoid mutating variables and prefer deconstruction. I think it's not a big deal here as enrichedProperties is deconstructed as { ... } but we should double check

packages/twenty-server/src/engine/metadata-modules/field-metadata/field-metadata.resolver.ts
charlesBochet: LGTM though

packages/twenty-server/src/engine/metadata-modules/field-metadata/utils/resolve-field-metadata-standard-override.util.ts
charlesBochet: LGTM though, I think it's only a few percentage impacted + this looks the right way to me

packages/twenty-server/src/engine/subscriptions/metadata-event/metadata-event-publisher.ts
charlesBochet: this will introduce a precedent in the code base but does not have an impact here so I won't fight for it

charlesBochet (approved):

PR #20249, opened 2026-05-04
fix: allow multiple filter conditions on the same field

### LINKED ISSUE #20224: Cannot combine two filter conditions on the same field (e.g. contains + does not contain)
## Bug description

It is impossible to apply two different filter conditions on the **same field** — for example, filtering a **multi-select / multi-choice field** where it **contains** "X" **and does not contain** "Y".

When attempting to add a second filter condition on a field that already has a filter, the existing condition is **replaced** instead of a new one being added alongside it.

## Steps to reproduce

1. Open any object list view (e.g. Companies, People)
2. Add a filter on a **multi-choice field** — e.g. **Tags contains "agency"**
3. Try to add a second filter on the same field — e.g. **Tags does not contain "freelance"**
4. The first filter ("contains agency") is **overwritten** by the second one ("does not contain freelance")

## Expected behavior

Both filter conditions should coexist as an AND:
- Tags **contains** "agency"
- Tags **does not contain** "freelance"

This was reported and supposedly fixed in #5378, but the issue has regressed (at least for multi-choice fields).

## Current behavior

Adding a second filter on the same multi-choice field replaces the first filter instead of adding a new condition.

QuantumByteMaster:
## Summary
- Fixes inability to combine multiple filter conditions on the same field (e.g. Tags contains "X" AND does not contain "Y")
- Removes the duplicate-filter check that was loading an existing filter for editing instead of initializing a fresh one
- Deletes the now-unused `findDuplicateRecordFilterInNonAdvancedRecordFilters` utility

## Root cause
`useInitializeFilterOnFieldMetadataItemFromViewBarFilterDropdown` called `findDuplicateRecordFilterInNonAdvancedRecordFilters` when a field was selected in the filter picker. If a filter already existed on that field, it reopened the existing filter instead of creating a new one — making it impossible to add a second condition.

The fix removes this check. Since `useUpsertRecordFilter` matches by filter ID (not fieldMetadataId), each new filter naturally gets its own slot in `currentRecordFilters`.

Editing existing filters is unaffected — that path goes through `useSetEditableFilterChipDropdownStates`, which is unchanged.

Fixes #20224


charlesBochet:
Hi @QuantumByteMaster, answered on the issue, this is actually possible using Advanced Filter feature, closing this PR

PR #20250, opened 2026-05-04
fix: show empty cell instead of 'Not shared' for soft-deleted related records

LambertArm:
## Summary

Close#20076

### Problem
When a related record is soft deleted, cells were showing "Not shared" (with lock icon) which is misleading. "Not shared" implies a permission issue, not a deleted record.

### Solution
Changed `RelationToOneFieldDisplay` and `RelationFromManyFieldDisplay` to return `null` instead of `<ForbiddenFieldDisplay />` when:
1. We have a foreign key but no record data (`RelationToOneFieldDisplay`)
2. We have junction records but can't extract target records (`RelationFromManyFieldDisplay`)

This shows empty cells instead of "Not shared" for soft-deleted related records.

### Changes
- `packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/display/components/RelationToOneFieldDisplay.tsx`
- `packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/display/components/RelationFromManyFieldDisplay.tsx`

### Testing
- Manual testing: Create a relation, link records, soft delete related record, verify cell shows empty instead of "Not shared"
- No existing tests broken (no tests were testing this specific behavior)

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/display/components/RelationFromManyFieldDisplay.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/display/components/RelationFromManyFieldDisplay.tsx
@@ -4,16 +4,15 @@ import { useActivityTargetObjectRecords } from '@/activities/hooks/useActivityTa
 import { type NoteTarget } from '@/activities/types/NoteTarget';
 import { type TaskTarget } from '@/activities/types/TaskTarget';
 import { useObjectMetadataItems } from '@/object-metadata/hooks/useObjectMetadataItems';
-import { CoreObjectNameSingular } from 'twenty-shared/types';
 import { RecordChip } from '@/object-record/components/RecordChip';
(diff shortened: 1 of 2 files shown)


charlesBochet:
Thanks for the fix! I agree "Not shared" is misleading for soft-deleted records, but returning `null` (empty cell) loses the ability to distinguish soft-deleted from RBAC-filtered records.

**Analysis:** Both cases (soft-delete + no-access) produce the same frontend state: FK exists, record is null. The PR treats them identically, but they have different semantics.

**Industry precedent:** Salesforce clears the lookup field value when a record is moved to the recycle bin ("Clear the value of this field" default). HubSpot removes associations on delete. Both sever the FK rather than leaving it dangling.

**Alternative approach:** Null the FK on soft-delete at the application level, honoring the `onDelete: SET_NULL` setting already declared in relation metadata. The DB already has `ON DELETE SET NULL` constraints for hard-deletes — we just replicate this for soft-deletes. This requires no frontend changes and keeps "Not shared" meaningful (only triggers for true RBAC cases).

I'll open a separate PR with this approach.

PR #20251, opened 2026-05-04
fix: Invalid configuration instead of related notes 

### LINKED ISSUE #20118: Invalid configuration instead of related notes
https://github.com/twentyhq/private-issues/issues/471

Martin-Blau:
## Summary
The Notes widget (and other record-bound widgets like Tasks, Files, Calendar, Emails) incorrectly displayed "Invalid Configuration" when rendered on dashboard or standalone page contexts. The root cause was an overly broad `ErrorBoundary` that caught all runtime errors uniformly and displayed a misleading error message.
## Related issue
Fixes: #20118 
## Problem Analysis
**Proximate Cause:**
- `NotesCard` calls `useTargetRecord()` which throws a generic `Error` when `targetRecordIdentifier` is undefined
- `ErrorBoundary` in `WidgetCardShell.tsx` catches this error and renders `PageLayoutWidgetInvalidConfigDisplay`
- This displays "Invalid Configuration" which is factually misleading

**Triggering Cause:**
- Commit 5cd8b7899d removed the feature flag gate on page layouts, making them standard for all workspaces
- This exposed record-bound widgets to dashboard contexts where `targetRecordIdentifier` is intentionally undefined

**Error Propagation Chain:**
```
WidgetContentRenderer → NoteWidget → NotesCard → useTargetRecord()
useTargetRecord() throws Error('useTargetRecord must be used within a record page context')
ErrorBoundary catches error → PageLayoutWidgetInvalidConfigDisplay renders misleading UI
```
## Solution
Introduced a distinction between **configuration errors** and **record context requirement errors** by:
1. Creating a custom error class `RecordContextRequiredError`
2. Updating `useTargetRecord()` to throw this specific error type
3. Creating a dedicated display component for record context errors
4. Updating the `ErrorBoundary` fallback to handle error types appropriately
## User Impact
| Before | After |
|--------|-------|
| "Invalid Configuration" (red badge) | "Record Required" (gray badge) |
| Misleading error message | Accurate context-aware message |
| Users think widget is broken | Users understand widget needs record context |


charlesBochet (commented):
Thanks for looking into this issue. The distinction between "Invalid Configuration" and "Record Context Required" is a good idea in principle, but I believe this PR doesn't address the actual production crash from #20118.

## The actual crash

The production stack trace shows:
```
TypeError: Cannot read properties of null (reading 'id')
    at SidePanelToggleButton-IP25eh3-.js:252:2861
```

This maps to `useActivities.ts` — specifically the `updateActivitiesInStore` callback:

```ts
const updateActivitiesInStore = useCallback(
  (activityTargets: (TaskTarget | NoteTarget)[]) => {
    for (const activityTarget of activityTargets) {
      const activity = activityTarget[objectNameSingular];
      // 💥 activity is null here → activity.id throws TypeError
      store.set(recordStoreFamilyState.atomFamily(activity.id), activity);
    }
  },
  [store, objectNameSingular],
);
```

When an activity target has a null related note/task (e.g. soft-deleted or unresolved), `activityTarget[objectNameSingular]` returns `null`, and `.id` throws.

## Why the ErrorBoundary approach doesn't help here

React `ErrorBoundary` (including `react-error-boundary`) **only catches errors during the React render lifecycle**. The `onCompleted` callback from Apollo runs **after** the query resolves, outside the render phase. So the `ErrorBoundary` in `WidgetCardShell` will never intercept this crash — it propagates as an unhandled `TypeError`.

## Minimal fix

The correct fix is a null guard in `updateActivitiesInStore`:

```ts
const updateActivitiesInStore = useCallback(
  (activityTargets: (TaskTarget | NoteTarget)[]) => {
    for (const activityTarget of activityTargets) {
      const activity = activityTarget[objectNameSingular];
      if (!isDefined(activity)) {
        continue;
      }
      store.set(recordStoreFamilyState.atomFamily(activity.id), activity);
    }
  },
  [store, objectNameSingular],
);
```

(Note: the code at line 57-61 already does `.filter(isDefined)` for the render path, but the `onCompleted` callback was missing the same guard.)

charlesBochet (approved):
Thanks @John-pillo1124 for opening a PR, wrong fix though! I've fixed your PR, tested locally (was able to reproduce and make sure the fix is working)

PR #20258, opened 2026-05-04
fix: nullify foreign keys on soft-delete to prevent misleading 'Not shared' display

### LINKED ISSUE #20076: Cell wrongfully shows "Not shared" after related record was soft deleted
**How to reproduce**

Delete timCook 

--> 

charlesBochet:
## Summary

- When a record is soft-deleted, FK references from other records were left dangling, causing the frontend to show "Not shared" (lock icon) instead of an empty cell — misleading since the record is deleted, not RBAC-filtered.
- Adds a `RelationOnDeleteService` that replicates PostgreSQL's `ON DELETE SET NULL` constraint behavior at the application level for soft-deletes. After soft-deleting records, it scans relation metadata for incoming `MANY_TO_ONE` fields with `onDelete: SET_NULL` and nullifies the FK columns on referencing records.
- This matches industry standard (Salesforce's "Clear the value of this field" default behavior) and requires no frontend changes.

Closes #20076
Alternative to #20250

## Test plan

- [x] Unit tests for `RelationOnDeleteService` (4 tests: empty ids, SET_NULL nullification, CASCADE not affected, ONE_TO_MANY not affected)
- [ ] Integration test: soft-deleting a company nullifies `person.companyId` and `opportunity.companyId`
- [ ] Integration test: soft-deleting a task does NOT nullify `taskTarget.taskId` (CASCADE relation)
- [ ] Verify frontend no longer shows "Not shared" for soft-deleted relations
- [ ] Verify frontend still shows "Not shared" for actual RBAC-filtered relations

### DIFF
--- a/packages/twenty-server/src/engine/api/common/common-query-runners/common-delete-many-query-runner.service.ts
+++ b/packages/twenty-server/src/engine/api/common/common-query-runners/common-delete-many-query-runner.service.ts
@@ -1,4 +1,4 @@
-import { Injectable } from '@nestjs/common';
+import { Inject, Injectable } from '@nestjs/common';
 
 import { QUERY_MAX_RECORDS_FROM_RELATION } from 'twenty-shared/constants';
 import { ObjectRecord } from 'twenty-shared/types';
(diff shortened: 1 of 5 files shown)


charlesBochet:
wrong approach, closing

PR #20259, opened 2026-05-04
fix: distinguish soft-deleted relations from RLS-restricted ones

### LINKED ISSUE #20076: Cell wrongfully shows "Not shared" after related record was soft deleted

charlesBochet:
## Summary

Fixes #20076

When a relation foreign key exists but the target record can't be loaded, the UI previously always showed "Not shared" (`ForbiddenFieldDisplay`). This was misleading because it couldn't distinguish between:
- **Soft-deleted records** (the target was deleted, FK is stale) → should show empty
- **RLS-restricted records** (the target exists but the user lacks row-level access) → should show "Not shared"

**Approach:** The frontend already knows whether row-level permission predicates are configured on each object (via `ObjectPermissions.rowLevelPermissionPredicates`). When the relation display encounters a non-null FK with a missing record:
- If the target object has **no RLS predicates** → the record must be soft-deleted → render empty
- If the target object **has RLS predicates** → the record may be access-restricted → render `ForbiddenFieldDisplay`

This is a minimal, frontend-only fix (2 files changed) with no backend changes needed.

## Changes
- `RelationToOneFieldDisplay.tsx` — uses `useObjectPermissionsForObject` to check target object's RLS predicates
- `RelationFromManyFieldDisplay.tsx` — same logic for junction relations using `getObjectPermissionsForObject`

## Test plan
- [ ] Soft-delete a company, verify that relation fields pointing to it show empty (not "Not shared")
- [ ] Configure RLS predicates on an object, verify that restricted records still show "Not shared"
- [ ] Verify no regression for normal relation display (non-deleted, accessible records)
- [ ] Verify object-level permission denial still shows "Not shared" at the `FieldDisplay` level (unchanged behavior)

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/display/components/RelationFromManyFieldDisplay.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/display/components/RelationFromManyFieldDisplay.tsx
@@ -3,9 +3,11 @@ import { useContext } from 'react';
 import { useActivityTargetObjectRecords } from '@/activities/hooks/useActivityTargetObjectRecords';
 import { type NoteTarget } from '@/activities/types/NoteTarget';
 import { type TaskTarget } from '@/activities/types/TaskTarget';
+import { getObjectPermissionsForObject } from '@/object-metadata/utils/getObjectPermissionsForObject';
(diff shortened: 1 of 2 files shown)


charlesBochet:
not the best approach, closing

PR #20260, opened 2026-05-04
fix: show empty cell instead of 'Not shared' for soft-deleted related records

### LINKED ISSUE #20076: Cell wrongfully shows "Not shared" after related record was soft deleted

charlesBochet:
## Summary

Fixes #20076 (supersedes #20250)

When a related record is soft-deleted, the frontend displays "Not shared" (lock icon) because it sees a populated FK but a null relation object. This is misleading -- the record was deleted, not permission-restricted.

**Backend fix** (`process-nested-relations-v2.helper.ts`):
- For MANY_TO_ONE relations, widen the relation query with `.withDeleted()` and include `deletedAt` in the select
- In `assignRelationResults`, if the matched record has `deletedAt` set, nullify both the FK and the relation object in the API response
- Records filtered by RLS are still not returned (even with `withDeleted()`), so they correctly continue to show "Not shared"
- Strip `deletedAt` from relation results before returning to the client

**Frontend fix** (`RelationFromManyFieldDisplay.tsx`):
- For ONE_TO_MANY junction relations, return `null` instead of `<ForbiddenFieldDisplay />` when junction records exist but target records are unavailable

### Three cases now handled correctly:

| Scenario | FK in response | Relation object | Frontend display |
|---|---|---|---|
| **Live record** | `"abc"` | `{ id: "abc", ... }` | Record chip |
| **Soft-deleted record** | `null` | `null` | Empty cell |
| **RLS-hidden record** | `"abc"` | `null` | "Not shared" |

## Test plan

- [ ] Create a record with a MANY_TO_ONE relation (e.g., a person linked to a company)
- [ ] Soft-delete the related record (the company)
- [ ] Verify the relation field shows an empty cell, not "Not shared"
- [ ] Restore the related record and verify the relation reappears
- [ ] Verify that RLS-hidden relations still show "Not shared"

### DIFF
--- a/packages/twenty-server/src/engine/api/common/common-nested-relations-processor/process-nested-relations-v2.helper.ts
+++ b/packages/twenty-server/src/engine/api/common/common-nested-relations-processor/process-nested-relations-v2.helper.ts
@@ -1,6 +1,7 @@
 import { Injectable } from '@nestjs/common';
 
 import { FieldMetadataType, type ObjectRecord } from 'twenty-shared/types';
+import { isDefined } from 'twenty-shared/utils';
 import { type FindOptionsRelations, type ObjectLiteral } from 'typeorm';
 
 import { RelationType } from 'src/engine/metadata-modules/field-metadata/interfaces/relation-type.interface';
@@ -173,7 +174,7 @@ export class ProcessNestedRelationsV2Helper {
       targetObjectNameSingular,
     );
 
-    const columnsToSelect = buildColumnsToSelect({
+    const columnsToSelect: Record<string, boolean> = buildColumnsToSelect({
       select: selectedFields,
       relations: nestedRelations,
       flatObjectMetadata: targetObjectMetadata,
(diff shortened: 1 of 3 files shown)


packages/twenty-server/src/engine/api/common/common-nested-relations-processor/process-nested-relations-v2.helper.ts
```diff
@@ -173,25 +174,30 @@ export class ProcessNestedRelationsV2Helper {
...
+      columnsToSelect.deletedAt = true;
+      targetObjectQueryBuilder = targetObjectQueryBuilder.withDeleted();
+    }
+
     targetObjectQueryBuilder = targetObjectQueryBuilder.setFindOptions({
       select: columnsToSelect,
     });
 
+    const joinColumnName =
+      sourceFieldMetadata.settings.joinColumnName ?? `${sourceFieldName}Id`;
```
ijreilly, line 195: not tied to this PR but does this fallback make sense ? how do we explain it?

ijreilly (approved):

packages/twenty-server/src/engine/api/common/common-nested-relations-processor/process-nested-relations-v2.helper.ts
charlesBochet: I think we should actually get rid of joinColumnName and always compute based on sourceFieldName but that's not what is in the code right now. 

I will open a PR to clarify, should be a good usecase for LLM

PR #20262, opened 2026-05-04
feat: expose upgrade status as Prometheus gauge metrics

charlesBochet:
## Summary

- Adds `UpgradeGaugeService` that exposes three observable Prometheus gauges based on the recently merged upgrade status service:
  - `twenty_upgrade_instance_health` — 1 (up-to-date), 0 (behind), -1 (failed)
  - `twenty_upgrade_workspaces_behind_total` — count of workspaces with pending upgrade commands
  - `twenty_upgrade_workspaces_failed_total` — count of workspaces with a failed upgrade command
- Follows the existing gauge pattern (`WorkspaceGaugeService`, `BillingGaugeService`, `DatabaseGaugeService`)

### Caching & QPS design

Prometheus scrapes every **15s** via `ServiceMonitor`. Each gauge uses the `MetricsService.createObservableGauge({ cacheValue: true })` pattern which caches the value in Redis for **60 seconds**. Under that, `UpgradeStatusService.getInstanceAndAllWorkspacesStatus()` uses `UpgradeStatusCacheService` with a **1-hour TTL** in Redis.

Result: at most 1 DB query per hour regardless of scrape frequency.

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/upgrade/upgrade-gauge.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/upgrade/upgrade-gauge.service.ts
@@ -1,0 +1,98 @@
+import { Injectable, Logger, OnModuleInit } from '@nestjs/common';
+
+import { UpgradeHealthEnum } from 'twenty-shared/types';
+
+import { MetricsService } from 'src/engine/core-modules/metrics/metrics.service';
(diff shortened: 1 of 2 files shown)


etiennejouan (approved):
LGTM
Prom collects as many data points as there are pods. Wonder if that could have an imapct.

PR #20261, opened 2026-05-04
Fix phone unique constraints

MkDev11:
## Summary

Closes #20195

Fix phone field unique constraints so phone numbers are considered unique by both `primaryPhoneNumber` and `primaryPhoneCallingCode`.

- Include `primaryPhoneCallingCode` in the shared phone composite unique constraint metadata
- Align the frontend settings composite field config with the backend metadata
- Return all included unique composite subfields when building create-many conflict fields
- Match composite unique conflict fields as a group during create-many upserts

## Root Cause

Phone composite metadata only marked `primaryPhoneNumber` as part of the unique constraint. That made different international phone numbers with the same national number conflict, for example `+1 123456789` and `+32 123456789`.

## Test Plan

- `yarn workspace twenty-shared build`
- `jest --runTestsByPath <index action handler and create-many utility specs>`
- `prettier --check <touched files>`
- `oxlint --type-aware <touched files>`
- `nx run twenty-shared:typecheck`
- `nx run twenty-server:typecheck`
- `nx run twenty-front:typecheck`

PR #20263, opened 2026-05-04
Fix phone unique constraints

### LINKED ISSUE #20195: Unique constraint on phone type field falsely throws an error
**Scenario:**
1. Log in
2. Go to Settings > Data model > People > Phones field
3. Toggle Unique on and save
4. Go to People view
5. Insert the same number in 2 records with different landline code (e.g. +1 123456789 and +32 123456789)

**Actual:** Error is thrown as phone number is the same despite different landline code

**Expected:** No error is thrown as validation checks both landline code as well as phone number itself

hexven0m:
## Summary

Closes #20195

Fix phone field unique constraints so phone numbers are considered unique by both `primaryPhoneNumber` and `primaryPhoneCallingCode`. Previously, two records with the same number but different international codes (e.g. `+1 123456789` and `+32 123456789`) were incorrectly flagged as duplicates.

## Changes

- Mark `primaryPhoneCallingCode` as part of the phone composite unique constraint (`twenty-shared` metadata + frontend settings config).
- Return **all** included unique subfields from `getConflictingFields` (not just the first match).
- Match composite unique fields **as a group** in `getMatchingRecordId` so every subfield must match the same existing record.
- Treat `undefined === undefined` as a match for optional subfields, so two phone entries with the same number and no calling code still deduplicate (prevents NULL bypass).
- Skip composite dedup only when the request supplies **no values** for the base field.

## Root Cause

Phone composite metadata only marked `primaryPhoneNumber` as part of the unique constraint, so different international numbers sharing the same national digits collided. The dedup utility also only looked at a single composite property at a time.

## Test Plan

- [x] Added unit tests for `getConflictingFields` covering the multi-property phone case.
- [x] Added unit tests for `getMatchingRecordId` covering:
  - all composite subfields match the same record
  - partial subfield mismatch returns `undefined`
  - both-undefined subfields match (NULL-bypass fix)
  - one-side-undefined does not match
  - empty record skips dedup
- [x] Added unit test for `computeFlatIndexFieldColumnNames` returning every unique subfield column for phone composite fields.
- [ ] Manual verification: create two People with same number and different calling codes — both should save without a unique-constraint error.
- [ ] Manual verification: create two People with same number and **no** calling code — second should be flagged as duplicate.

### DIFF
--- a/packages/twenty-front/src/modules/settings/data-model/constants/SettingsCompositeFieldTypeConfigs.ts
+++ b/packages/twenty-front/src/modules/settings/data-model/constants/SettingsCompositeFieldTypeConfigs.ts
@@ -209,7 +209,7 @@ export const SETTINGS_COMPOSITE_FIELD_TYPE_CONFIGS = {
             .primaryPhoneCallingCode,
         isImportable: true,
         isFilterable: true,
-        isIncludedInUniqueConstraint: false,
+        isIncludedInUniqueConstraint: true,
       },
       {
         subFieldName:
--- a/packages/twenty-server/src/engine/api/common/common-query-runners/common-create-many-query-runner/utils/__tests__/get-conflicting-fields.util.spec.ts
+++ b/packages/twenty-server/src/engine/api/common/common-query-runners/common-create-many-query-runner/utils/__tests__/get-conflicting-fields.util.spec.ts
@@ -54,6 +54,13 @@ describe('getConflictingFields', () => {
     isUnique: true,
   });
 
+  const phonesUniqueField = createMockField({
+    id: 'phones-unique-id',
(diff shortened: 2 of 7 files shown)


etiennejouan (commented):
LGTM - Thanks and congrats for your first contribution ! 
Could you, please, add integration tests here : packages/twenty-server/test/integration/graphql/suites/unique-field/unique-phones-field-null-equivalence.integration-spec.ts : 
-> 'should allow creating when same phoneNumber but different phoneCallingCode'

etiennejouan:
Sorry, but closing your PR as [an other one](https://github.com/twentyhq/twenty/pull/20261) has been opened earlier

PR #20298, opened 2026-05-05
Billing - Migrate from Stripe metering

etiennejouan:
**Overall strategy**
**1. Introduce “Billing V2” behind a workspace flag**
Gate the new model with FeatureFlagKey.IS_BILLING_V2_ENABLED so existing workspaces stay on the old behavior until they’re migrated or explicitly on V2.

**2. Replace workflow metered SKUs with a resource-credit product**
Conceptually, billable “workflow execution” usage is not the primary subscription line item anymore. Add a RESOURCE_CREDIT product (and keep WORKFLOW_NODE_EXECUTION as deprecated for the transition). Usage and limits are expressed through credit buckets (e.g. price metadata like credit_amount), so one product can represent pooled credits instead of a narrow workflow-only meter.

**3. Migrate subscriptions in two layers**
Schema/catalog: persist extra price metadata (instance upgrade) so the server knows credit amounts and can match Stripe prices to the new model.
Per workspace: the registered workspace command upgrade:2-2:migrate-to-billing-v2 finds subscriptions that still have WORKFLOW_NODE_EXECUTION, swaps those items to the right RESOURCE_CREDIT prices (using existing Stripe schedule + BillingSubscriptionUpdateService stack), then treats the workspace as V2 (flag). Workspaces without that legacy item or without a subscription are skipped.

**4. Unify subscription lifecycle + usage on the server**

**5. Refresh the product surface in Settings**

Test : 

- [x] Subscribe v1 +  Update subscribe + Migrate
- [x]  Subscribe v2 + Update subscribe


packages/twenty-server/src/engine/core-modules/billing/stripe/services/stripe-invoice.service.ts
```diff
@@ -40,4 +40,37 @@ export class StripeInvoiceService {
...
+      subscription: stripeSubscriptionId,
+      amount: diffAmountInCents,
+      currency,
+      description,
+    });
+
+    const invoice = await this.stripe.invoices.create({
+      customer: stripeCustomerId,
+      subscription: stripeSubscriptionId,
+    });
```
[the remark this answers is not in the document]
etiennejouan: but it works. passing subscription may change the default logic

packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-invoice.service.ts
```diff
@@ -100,26 +103,68 @@ export class BillingWebhookInvoiceService {
       return;
     }
 
-    if (periodStart) {
+    const isFirstPeriodAfterTrial =
```
Weiko, line 112: Can't we get this info from the invoice returned by stripe? The fallback is fine to me though but might want to double check

Weiko (approved):
LGTM

packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-invoice.service.ts
etiennejouan: Best check would be to know if it's first invoice with non null amount. But seems a bit too much

PR #20304, opened 2026-05-05
refactor: stop reading joinColumnName from relation field settings

charlesBochet:
## Summary

`joinColumnName` on relation field settings is always derivable from the field name (and the target object name for morph relations). This PR stops reading it from settings anywhere in production code; the stored value is no longer used.

The settings field is **not** removed from data yet — a follow-up can drop it once we are confident nothing depends on the stored value.

## Helpers

The helpers are split by layer because frontend and backend hold morph relations differently: the frontend has a base name plus a `morphRelations[]` array, the backend has one row per target with the name already morph-resolved.

| Helper | Layer | When to use |
|---|---|---|
| `computeRelationGqlFieldJoinColumnName` | Shared / frontend (`gqlField`) | Non-morph relation on the frontend. |
| `computeMorphRelationGqlFieldName` | Shared / frontend (`gqlField`) | Need the per-target morph gqlField name (e.g. `targetCompany`). |
| `computeMorphRelationGqlFieldJoinColumnName` | Shared / frontend (`gqlField`) | Per-target morph join column on the frontend. Prefer over the non-morph helper for any morph field — it forces the per-target inputs. |
| `computeMorphOrRelationFieldJoinColumnName` | Backend (`FlatFieldMetadata.name`) | Any backend read or write — the flat name is already morph-resolved, so one helper covers both cases. |
| `computeMorphRelationFlatFieldName` | Backend (`FlatFieldMetadata.name`) | **Mutation paths only** (create / update / object rename). Reads consume the stored `field.name` and never call this. |

## Test plan

- [x] Typecheck and lint (front, server, shared)
- [x] Existing unit tests pass
- [ ] CI green


packages/twenty-server/src/engine/api/common/common-nested-relations-processor/process-nested-relations-v2.helper.ts
```diff
@@ -201,13 +205,16 @@ export class ProcessNestedRelationsV2Helper {
...
 
     const fieldMetadataTargetRelationColumnName =
-      targetRelation &&
-      isFieldMetadataEntityOfType(
-        targetRelation,
-        FieldMetadataType.MORPH_RELATION,
-      )
-        ? `${targetRelation.settings?.joinColumnName}`
-        : `${targetRelationName}Id`;
+      computeRelationFieldJoinColumnName({
```
charlesBochet: this is suspicious too, does it actually make sense to compute a RelationFieldJoinColumn with Relation util for Morph?

packages/twenty-server/src/engine/api/common/common-query-runners/common-merge-many-query-runner.service.ts
```diff
@@ -405,7 +405,9 @@ export class CommonMergeManyQueryRunnerService extends CommonBaseQueryRunnerServ
         objectMetadata: objMetadata,
         fieldName: field.name,
         fieldId: field.id,
-        joinColumnName: relationSettings.joinColumnName,
+        joinColumnName: computeRelationFieldJoinColumnName({
```
charlesBochet: is this never morph?

ijreilly (approved):
nice !

PR #20305, opened 2026-05-05
fix(front): resolve labelIdentifier per target for morph relation depth=1

charlesBochet:
## Summary

On the show page, morph relations were showing "Untitled" entries for targets whose `labelIdentifier` is not `name` (for example `Note.title`). The GraphQL response only contained `id` for those records.

`generateDepthRecordGqlFieldsFromFields` was hardcoding the morph depth=1 sub-selection to `{ id, name }` for every target instead of resolving each target's `labelIdentifier` (and `imageIdentifier`) from `objectMetadataItems`, the way the non-morph relation branch already does. The morph branch was also ignoring `shouldOnlyLoadRelationIdentifiers`.

image


Weiko (approved):
LGTM

PR #20319, opened 2026-05-06
fix(front): scroll long content in rich text editor

### LINKED ISSUE #20309: Expanded panel does not scroll long RICH_TEXT content
## Bug Description

When viewing a field with long RICH_TEXT content in the expanded panel, the content does not scroll. As a result, part of the text becomes inaccessible and cannot be viewed.

**Example:** 
```
Steps to reproduce:

Open a record with a long RICH_TEXT field
Expand the panel
Try to scroll through the content
```

## Expected behavior

The expanded panel should allow vertical scrolling so the entire content can be accessed.

Actual behavior:
Scrolling is not possible, and part of the content is cut off.

Environment:
(please fill in: browser, OS, version)

Additional notes:
This makes it difficult to work with long-form content inside the expanded panel.
```

### PR DESCRIPTION
**Pull request by Nik1125**
## Summary

Fixes scroll behavior in the rich text editor.

Long content was unbounded — the popup grew off-screen, the expand-to-side-panel button became unreachable, and the side panel itself didn't scroll either.

Closes #20309

## Changes

- `RichTextFieldInput`: cap the popup at `min(60vh, 500px)` and wrap the editor in a scrollable region. Collapse button stays at the top via `align-items: flex-start`.

- `RecordInlineCellEditMode`: add `shift()` middleware to keep the popup inside the viewport after `flip()` triggers (previously the popup could extend above the viewport top with the collapse button out of reach).

- `SidePanelContainer` / `SidePanelRouter`: add `min-height: 0` to the flex-column chain so the existing `overflow-y: auto` on the content area can actually clip and scroll long children.

  The previous attempt in #20310 added the same `min-height: 0` plus a nested overflow wrapper inside the rich-text page; the nested wrapper turned out to be the reason scrolling didn't work.

## Test plan

- [x] Open a record with a long `RICH_TEXT` field
- [x] Click the field — popup opens bounded; long content scrolls inside
- [x] Click the expand button (top-right) — side panel opens
- [x] Side panel: long content scrolls vertically
- [x] Popup near the bottom of the viewport flips upward and stays fully inside the viewport (collapse button remains visible)
- [x] `lint:diff-with-main`, `typecheck`, prettier — green

https://github.com/user-attachments/assets/827be881-aadd-49ed-9ddc-7566c00cf4be

### CODE REVIEW

File: packages/twenty-front/src/modules/object-record/record-inline-cell/components/RecordInlineCellEditMode.tsx [hunk 1]
```diff
@@ -71,6 +72,7 @@ export const RecordInlineCellEditMode = ({
     placement: isCentered ? 'bottom' : 'bottom-start',
     middleware: [
       flip(),
+      shift({ padding: 8 }),
```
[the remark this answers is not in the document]
Nik1125: Good catch, fixed in e84fae6 — moved shift() to the end so it's the final positional correction.

packages/twenty-front/src/modules/side-panel/components/SidePanelContainer.tsx
```diff
@@ -22,6 +22,7 @@ const StyledSidePanelContainer = styled.div<{ isMobile: boolean }>`
 
     return `calc(100% - ${mobileOffset})`;
   }};
+  min-height: 0;
```
martmull, line 25: why is it useful for?

packages/twenty-front/src/modules/side-panel/components/SidePanelRouter.tsx
```diff
@@ -15,6 +15,7 @@ import { ThemeContext } from 'twenty-ui/theme-constants';
 
 const StyledSidePanelContent = styled.div`
   flex: 1;
+  min-height: 0;
```
martmull, line 18: same

packages/twenty-front/src/modules/object-record/record-inline-cell/components/RecordInlineCellEditMode.tsx
```diff
@@ -82,6 +83,7 @@ export const RecordInlineCellEditMode = ({
               crossAxis: -5,
             },
       ),
+      shift({ padding: 8 }),
```
martmull, line 86: why?

packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/input/components/RichTextFieldInput.tsx
```diff
@@ -31,16 +31,26 @@ const RichTextFieldEditor = lazy(() =>
...
   padding: ${themeCssVariables.spacing[2]} ${themeCssVariables.spacing[2]}
     ${themeCssVariables.spacing[2]} ${themeCssVariables.spacing[12]};
   position: relative;
   width: 480px;
 `;
 
+const StyledEditorScroll = styled.div`
+  align-self: stretch;
+  flex: 1;
+  min-height: 0;
```
martmull, line 50: why `min-height:0`?

packages/twenty-front/src/modules/side-panel/components/SidePanelContainer.tsx
Nik1125: This container is a flex item (flex: 1) inside the side-panel layout.

Flex items have an implicit `min-height: auto`, which prevents them from
shrinking below their content size — so when the rich-text editor inside
grows past the viewport, the existing
`max-height: calc(100% - …)` is effectively ignored and the content
overflows the panel.

Setting `min-height: 0` lets the flex item actually respect its
`max-height`, which is what makes the inner scroll wrapper in
`SidePanelRouter` become a real scroll boundary instead of just expanding
the panel.

Standard fix for the flexbox `overflow: auto` no-op:
https://css-tricks.com/flexbox-truncated-text/

packages/twenty-front/src/modules/side-panel/components/SidePanelRouter.tsx
Nik1125: Same reason as `SidePanelContainer`.

`StyledSidePanelContent` has `flex: 1` and `overflow-y: auto`, but a
flex item's default `min-height: auto` makes it grow to fit its children
instead of being constrained by the parent — so `overflow-y: auto`
never triggers and long content (rich-text bodies, sub-pages) just
pushes the panel beyond its bounds.

`min-height: 0` lets the flex item shrink to the parent's bounded
height, at which point `overflow-y: auto` becomes the actual scroll
container.

This is what makes long notes/tasks scroll inside the side panel instead
of overflowing.

packages/twenty-front/src/modules/object-record/record-inline-cell/components/RecordInlineCellEditMode.tsx
Nik1125: The popup uses a negative `mainAxis` offset (`-29` / `-26`) so it
overlays the inline cell.

When the cell sits near the bottom of the viewport, `flip()` flips
placement to `top`, but with a tall rich-text editor the resulting
popup can still extend above the viewport top — and there's no further
flip target, so the user can't reach the top of the content (which is
also where the expand-to-side-panel button lives).

`shift({ padding: 8 })` is floating-ui's viewport-clamping middleware —
it slides the popup along the main axis to keep it inside the viewport
with an 8px gutter when neither placement fits cleanly.

It's placed after `offset()` (per cubic's earlier comment) so the
negative offset is applied first and `shift()` is the final positional
correction.

https://floating-ui.com/docs/shift

packages/twenty-front/src/modules/object-record/record-field/ui/meta-types/input/components/RichTextFieldInput.tsx
Nik1125: Same flex-overflow rule.

`StyledEditorScroll` is a flex item inside `StyledContainer`, which now
has `max-height: min(60vh, 500px); overflow: hidden`.

Without `min-height: 0`, the wrapper's implicit `min-height: auto`
makes it grow to its BlockNote content's height instead of the parent's
bounded `max-height`, so `overflow-y: auto` on this wrapper never
kicks in and no scrollbar appears.

With `min-height: 0` the wrapper shrinks to fill the bounded parent,
and `overflow-y: auto` makes long rich-text content scroll inside the
popup.

This is what makes the expand-to-side-panel button reachable on tall
notes — the original bug from #20309.

Nik1125:
Addressed all review comments, ready for another look 🙏

Nik1125:
Hi @martmull, gentle ping on this one 🙏
All review comments have been addressed, the automated pre-review is green ("No issues detected"), CI is passing and there are no conflicts with main. The PR has been waiting for a re-review for about a week now.
Could you take another look when you get a chance? If you're swamped, no worries - happy for this to be reassigned to anyone else from the team who has bandwidth (cc @twentyhq/maintainers or whoever is appropriate). Just want to make sure it doesn't get lost. Thanks!

martmull (approved):
Sorry @Nik1125 this went off my attention for a while. Approved. Thanks for your patience!

Nik1125:
@martmull thank you so much ;)

PR #20322, opened 2026-05-06
fix(server): bypass workspace cache in onboardingStatus resolver

charlesBochet:
## Summary

In multi-instance deployments, `coreEntityCacheService` memoizes the workspace entity per server for ~10s, bypassing Redis hash invalidation. After `activateWorkspace`, if the next `currentUser` query is routed to a stale replica, the server returns `onboardingStatus: WORKSPACE_ACTIVATION` and `workspaceMember: null`, the client redirects to `/create/profile`, and submitting the form throws "User is not logged in". Reproduces on prod/staging only (local dev = single instance).

Fix: in `OnboardingService.getOnboardingStatus({ user, workspaceId })`, read the workspace directly from `WorkspaceEntity` repository (bypassing the per-instance core entity cache) so `onboardingStatus` reflects the freshest `activationStatus` right after `activateWorkspace`, even when the request hits a replica with a stale cached workspace.

## Test plan

- Prod/staging: sign up + create workspace, verify `/create/profile` works and form submits.
- Local: regression on the full onboarding flow.

### DIFF
--- a/packages/twenty-front/src/pages/onboarding/CreateWorkspace.tsx
+++ b/packages/twenty-front/src/pages/onboarding/CreateWorkspace.tsx
@@ -28,6 +28,14 @@ import { themeCssVariables } from 'twenty-ui/theme-constants';
 import { useMutation } from '@apollo/client/react';
 import { ActivateWorkspaceDocument } from '~/generated-metadata/graphql';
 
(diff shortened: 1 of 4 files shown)


Weiko (approved):
LGTM

PR #20324, opened 2026-05-06
Remove broken total count from workflow version

thomtrp:
## Problem

Opening a workflow run with a form step in the side panel, closing the form and reopening it crashes the app: `<SidePanelWorkflowStepInfo>` blows up on `workflow.versions.find` because `versions` is `null` in the Apollo cache.

## Root cause

`useWorkflowVersion` was selecting:

```ts
workflow: { id, name, statuses, versions: { totalCount: true } }
```
Twenty's GraphQL field generator doesn't support connection-level scalars — { totalCount: true } is interpreted as fields on the inner WorkflowVersion node, gets filtered out, and the query collapses to:
versions { edges { node { __typename } } }
The server returns versions: null for that empty-node selection. 

Why now
The selection has always been wrong, but two recent changes made it consistently surface:

Apollo Client v4 upgrade (#18584): stricter normalized writes, null always wins.
#20242: WorkflowRunSSESubscribeEffect in the form filler keeps SSE flowing, which re-fires useWorkflowVersion more often, making the bad query consistently the last writer.


Weiko (approved):
If not used then LGTM 👍

charlesBochet (approved):

PR #20326, opened 2026-05-06
Add isConfigured to application registration in App admin panel

martmull:
## After
Added "Configured" column in admin panel apps tab
image

Add a banner to ask to configure the application 
image


Weiko (commented):
N+1 queries issue here (1 query to ApplicationRegistrationVariable for each ApplicationRegistration), you can use a dataloader instead @martmull

Weiko (commented):
```typescript
  async isConfigured(applicationRegistrationId: string): Promise<boolean> {
    const variables = await this.findVariablesByRegistrationId(
      applicationRegistrationId,
    );

    const requiredSecrets = variables.filter((v) => v.isSecret && v.isRequired);

    return requiredSecrets.every((v) => v.isFilled);
  }
```
Why only "secret" variables?

Weiko (approved):
Not sure if "true" should be the default value 🤔 @martmull 
LGTM

PR #20331, opened 2026-05-06
refactor(agents): split tool resolution into native and action rails

ehconitin:
## Summary

Splits AI agent tool resolution into two independent rails:

- **Native tools** — capabilities baked into the model SDK (Anthropic/OpenAI `web_search`, xAI `web`/`x` provider options). Bound by `NativeToolBinderService`, controlled by per-agent `modelConfiguration` toggles. Opaque to Twenty — executed on the model provider's servers.
- **Action tools** — registry-scoped tools from `ToolRegistryService` (code interpreter, send email, record CRUD, etc.). Permission-gated via the agent's role. Executed on Twenty's server.

Both rails merge into a single `ToolSet` at call time. When both surfaces expose a search tool the model picks at runtime — coexistence is intentional (relevant once Exa returns as an action, see below).

## Notable changes worth calling out

**Contract change: `AgentAsyncExecutorService.executeAgent` no longer accepts `rolePermissionConfig`.** Workflow agents now scope exclusively by the agent's own permission-tab role (`unionOf: [agentRoleId]`). The previous role-merging path (caller role intersected with agent role) is removed. No agent role → no registry tools (fail-closed by design).

**`NativeToolBinderService` relocated** from `core-modules/tool-provider/native/` → `metadata-modules/ai/ai-models/services/`. The binder needs SDK-package knowledge, which lives in `ai-models`. Old location created a backwards module dependency.

**`NATIVE_MODEL_TOOLS_BY_SDK_PACKAGE` is exhaustive over `AiSdkPackage`** (`Record<>`, not `Partial<Record<>>`). Adding a new SDK without thinking about native tools now fails the build. SDKs without native tools (Bedrock, Google, Mistral, Azure, OpenAI-compatible) get explicit `{}` entries.

**Discriminated union `kind: 'sdk-tool' | 'provider-option'`** lets one registry describe both function tools (Anthropic/OpenAI) and runtime sources (xAI). Follows the local `tool-provider` convention from #19321.

## Deferred to follow-ups

- **Exa web search is dropped from this PR** (along with its `WEB_SEARCH_TOOL` permission flag and the Exa-specific gating). Exa comes back as an **action/app tool** once apps can define permission flags through the SDK — ongoing work in #20481.
- **xAI native search currently errors.** xAI deprecated its Live Search API (the `web`/`x` provider-option sources this rail maps to), so xAI returns `410` when native search is actually exercised. The code path itself is clear — it's only hit if you test xAI native tools. Fixed separately alongside the broader xAI model fixes.

## Conscious non-decisions

- **No "twenty-native" category.** `native` is reserved for model/provider SDK features; everything Twenty-owned is just a tool/action.
- **Coexistence over precedence.** No rule forcing an action search tool to override native search (or vice-versa) — when both exist, it's the user's choice in workflow agents and the model's choice in chat.


packages/twenty-server/src/engine/metadata-modules/ai/ai-agent-execution/services/agent-async-executor.service.spec.ts
```diff
@@ -0,0 +1,201 @@
+import { generateText } from 'ai';
```
FelixMalfait: Let's remove the service.spec they are not really useful ; we can do much more useful test (integration tests for example are usually much more efficient to catch real bugs)

ehconitin:
reopen after https://github.com/twentyhq/twenty/pull/20481

FelixMalfait:
Uploading Screen Recording 2026-05-28 at 21.08.31.mov…


Websearch doesn't appear until I change the model?

PR #20343, opened 2026-05-07
fix(messaging): preserve same-domain imports for institutional accounts

neo773:
Use case by a cloud customer workspace with 9 seats

Currently we don't treat all institutional domains as same, for example lot of European academic institutions use generic TLDs like .de, .fr

Customer's primary connected account email is `user@tu-darmstadt.de` but they would like to keep their emails exchanged with colleagues which is legitimate use case, but we drop it in `packages/twenty-server/src/modules/messaging/message-import-manager/utils/filter-out-internals.util.ts`

This PR adds a list of institutional domains and a new utility `isInstitutionalEmail` which checks for both existibg `emailProvidersSet` and new `universityDomainsSet`


charlesBochet:
Discussed with @Bonapara and you @neo773, let's not do that but rather add an option under Settings security to bypass internal email filtering.

image
Let's add:
Sync Internal Emails

Include emails where all participants share the same domain.

(toggled by default)

Also, hide this in case the option is not checked
image

PR #20344, opened 2026-05-07
add NO_STAGE stage option + hide column by default; block empty stage in workflow UI

### LINKED ISSUE #20139: Bug: For Non nullable fields (SELECT, MULTISELECT), do not allow workflow update / create / upsert to set "no selected"
## Scope & Context

We have multiple opportunity views to handle different types of situations. 
Here is a simple example to understand: 

- Default views (organizing opportunities by `stage`) are used to handle all incoming requests.
- When someone puts an opportunity under the "WIN" `stage`, we add a custom field called `winstage` set to "New Client".
- There is a view that displays all opportunities with a non-null `winstage` field, so the marketing team can handle the onboarding in a dedicated view.

## Current behavior

Once we set the `winstage`, the opportunity stays in the WIN `stage` column in the default view, resulting in an infinite stack of opportunities. 

We can't manually remove that stage (set it to null).
When using a Workflow to do it and selecting the default "no stage selected," it still doesn't remove the "WIN" value.

The workaround is to create a "HIDDEN" `stage` value and hide its column in the view. 

## Expected behavior

It should be possible to have a null `stage` to hide opportunities in kanban views based on enum fields.

jsdevninja:
## Summary
- Add a **`NO_STAGE`** option to standard **Opportunity › Stage**.
- Seed the **Kanban (“by Stage”) view** with a **`noStage` view group mapped to `NO_STAGE`** and set **`isVisible: false`** so the column stays hidden until users explicitly show it (matches maintainer guidance to use a sentinel stage instead of nullable).
- Migrate existing workspaces with a **workspace upgrade command** that appends **`NO_STAGE`** to **`opportunity.stage`** options when missing.
Closes #20139

## Workflow UX
**Do not expose an empty / “No {field}” select option when the underlying field is not nullable** (`FormFieldInput` → `FormSelectFieldInput`). Clearing a workflow variable fallback now picks the **first available option** instead of trying to persist an empty/`null`-like state for required selects like **stage**.
## Not changing
Per review feedback we are **not** making **`stage`** nullable (`null`), and we dropped the nullable migration.
## Regression / housekeeping
Remove the **`stage: null`** GraphQL integration expectation.

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-field/ui/components/FormFieldInput.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-field/ui/components/FormFieldInput.tsx
@@ -116,6 +116,7 @@ export const FormFieldInput = ({
       VariablePicker={VariablePicker}
       options={field.metadata?.options}
       readonly={readonly}
+      allowsEmptySelection={field.metadata.isNullable}
     />
   ) : isFieldFullName(field) ? (
     <FormFullNameFieldInput
--- a/packages/twenty-front/src/modules/object-record/record-field/ui/form-types/components/FormSelectFieldInput.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-field/ui/form-types/components/FormSelectFieldInput.tsx
@@ -26,6 +26,7 @@ type FormSelectFieldInputProps = {
   VariablePicker?: VariablePickerComponent;
   options: SelectOption[];
   readonly?: boolean;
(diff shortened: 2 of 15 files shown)


etiennejouan:
Hi @jsdevninja, thank you for your contribution. will start review as soon as CI turns green.

etiennejouan:
Closing this PR, feel free to contact me on Discord to re-open it

PR #20350, opened 2026-05-07
Oxlint ignore twenty-version constant

prastoin:
# Introduction
We could handle that by installing prettier on the package.json root and running it over the twenty versions files inside the ci.
But to be honest it seems redundant, and would bloat the ci in the end. Lets just not care about lint in these codegen files

Related https://github.com/twentyhq/twenty/pull/20345


Weiko (approved):
LGTM

PR #20355, opened 2026-05-07
Fix issue 33

Careeza:
Following https://github.com/twentyhq/twenty/pull/20032

## Context
~Phase 2 of the `gridPosition` removal (Phase 1 was #20032). Now that Phase 1 has rolled out and every widget has a non-null `position`, this PR drops `gridPosition` from the DB, entity, GraphQL/REST API, manifest type, AI tools, all standard and default layouts, and every test fixture. `position` is the single source of truth.~

Postponing phase 2, some part of the code was still writing in the old column (no real impact since the code was reading `new ?? legacy` but in DB position could still be empty


prastoin:
This was a test

PR #20356, opened 2026-05-07
refactor(file v2) - deletion

etiennejouan:


packages/twenty-server/src/engine/twenty-orm/interfaces/workspace-internal-context.interface.ts
```diff
@@ -22,4 +23,5 @@ export interface WorkspaceInternalContext {
   userWorkspaceRoleMap: UserWorkspaceRoleMap;
   eventEmitterService: WorkspaceEventEmitter;
   coreDataSource: DataSource;
+  messageQueueService?: MessageQueueService;
```
Weiko: Might consider using the existing database events and dispatch the job in a listener instead of exposing queues in typeorm 🤔 Not a strong opinion though!

etiennejouan: You're right...and I've already done it 3 months ago 🙄

charlesBochet (approved):
LGTM

PR #20357, opened 2026-05-07
Fix auto draft workflow

prastoin:
# Introduction
Cannot use github graphql mutation with a `GITHUB_TOKEN` needs a authenticated one
Dispatching to ci-priv to do so


Weiko (approved):
LGTM

PR #20360, opened 2026-05-07
feat(public-domain): bind public domains to apps + reorganize settings

FelixMalfait:
## Summary

- **Public domains can now be bound to a specific app.** When a request hits an app-bound public domain, route resolution restricts logic-function matching to that app's HTTP-routed functions only — isolating each app's routes to its own domain instead of letting routes from other apps in the workspace match nondeterministically.
- **Settings sidebar reorganized.** Removed the standalone Domains page. Workspace Domain → General. Approved Domains + Invitations → Members "Access" tab. Emailing Domains + Public Domains → Apps "Developer" tab. Roles → Members "Roles" tab.

## Why

The use case: someone building a partner portal app or a lead-collection app declares private objects (leads, partners…) plus a few public HTTP routes. Each app needs its own domain (`partners.acme.com`, `leads.acme.com`) without those domains exposing every other app's routes in the same workspace. Today's PublicDomainEntity is workspace-scoped only, so all HTTP-routed logic functions in a workspace compete for any public domain — first match wins nondeterministically.

## Backend

- Added nullable `applicationId` FK to `PublicDomainEntity` (cascade-deleted with the app); indexed for the route-trigger lookup.
- New fast instance command `2-4-instance-command-fast-1798000003000-add-application-id-to-public-domain` adds the column, index, and FK constraint.
- `createPublicDomain(domain, applicationId)` accepts an optional app binding; new `updatePublicDomain(domain, applicationId)` mutation rebinds/unbinds an existing domain. Both validate the application belongs to the workspace.
- `WorkspaceDomainsService.resolveWorkspaceAndPublicDomain(origin)` returns both the workspace and the matched public domain in one query — replacing the old back-to-back lookups in the route-trigger hot path. `getWorkspaceByOriginOrDefaultWorkspace` is preserved as a thin wrapper.
- `RouteTriggerService` filters `logicFunction` by `applicationId` when the matched public domain is app-scoped; falls back to workspace-wide when unbound.
- Three sequential validation queries in `createPublicDomain` now run in parallel via `Promise.all`.

## Frontend

| Old location | New location |
|---|---|
| Settings sidebar → Domains (standalone page) | Removed |
| Domains page → Workspace Domain | General page |
| Domains page → Approved Domains | Members → Access tab |
| Domains page → Emailing Domains | Apps → Developer tab |
| Domains page → Public Domains | Apps → Developer tab |
| Settings sidebar → Roles (standalone) | Members → Roles tab |
| `pages/settings/roles/` | `pages/settings/members/roles/` |

- The Public Domain detail page has an Application picker that uses `Select`'s native `emptyOption` + `null` value pattern (matches `SettingsDataModelObjectIdentifiersForm`).
- Members page tabs use the existing `TabListFromUrlOptionalEffect` mechanism (rendered automatically by `TabList`) for hash-based tab activation.
- `/settings/members/roles` redirects to `/settings/members#roles` so role sub-pages' `navigate(SettingsPath.Roles)` lands on the Members page with the Roles tab pre-selected.
- All affected breadcrumbs updated to nest under their new parents.
- `SettingsPath.Roles` and friends now nest under `members/`; `Subdomain` and `CustomDomain` under `general/`; `PublicDomain` and `EmailingDomain` under `applications/`.

## Test plan

- [x] `nx typecheck twenty-front` passes
- [x] `oxlint --type-aware` clean on all touched files
- [x] `prettier --check` clean on all touched files
- [x] Migration applied locally; `publicDomain.applicationId` (uuid, nullable) confirmed in DB
- [x] GraphQL schema exposes `PublicDomain.applicationId`, `createPublicDomain.applicationId`, `updatePublicDomain` mutation
- [x] **End-to-end route resolution scenarios verified locally:**
  - Domain bound to App A, function in App A → route matches ✅
  - Domain bound to App B, function in App A → route does NOT match (HTTP 404 `TRIGGER_NOT_FOUND`) ✅
  - Domain unbound (`applicationId = NULL`) → route matches workspace-wide ✅
  - Unknown path on bound domain → returns 404 cleanly ✅
- [x] UI sanity (browser-tested at `apple.localhost:3001`):
  - General page shows Workspace Domain card
  - Members page shows Team / Access / Roles tabs
  - Access tab combines Invite by link + by email + Approved Domains
  - Roles tab embeds the role list
  - `/settings/members/roles` direct URL → redirects + Roles tab pre-selected
  - Apps Developer tab shows Emailing Domains + Public Domains sections
  - Public Domain detail page has Application picker dropdown listing workspace apps
  - Sidebar nav: "Domains" and "Roles" no longer present (now folded into General/Members)

## Notes for reviewers

- Creating a public domain via the UI still requires Cloudflare credentials in the dev `.env` (`CLOUDFLARE_API_KEY`, `CLOUDFLARE_PUBLIC_DOMAIN_ZONE_ID`, `PUBLIC_DOMAIN_URL`). The DNS step is unchanged from main.
- The `applicationId` column is nullable, so existing public-domain rows continue to work workspace-wide — no data backfill required.
- `SettingsRolesContainer` was deleted (no longer referenced after `SettingsRoles` index page was removed).


packages/twenty-front/src/modules/settings/domains/components/SettingPublicDomain.tsx
```diff
@@ -73,6 +88,22 @@ export const SettingPublicDomain = () => {
...
   );
 
+  const { data: applicationsData } = useQuery(FindManyApplicationsDocument);
+
+  const applicationPinnedOption: SelectOption<string | null> = useMemo(
+    () => ({ value: null, label: t`Workspace (all apps)` }),
+    [t],
+  );
+
+  const applicationOptions = useMemo<SelectOption<string | null>[]>(
```
FelixMalfait: is the useMemo really useful?

packages/twenty-front/src/modules/settings/domains/components/SettingPublicDomain.tsx
```diff
@@ -73,6 +88,22 @@ export const SettingPublicDomain = () => {
     FindManyPublicDomainsDocument,
   );
 
+  const { data: applicationsData } = useQuery(FindManyApplicationsDocument);
+
+  const applicationPinnedOption: SelectOption<string | null> = useMemo(
```
FelixMalfait: is the useMemo really useful?

packages/twenty-front/src/modules/settings/components/SettingsProtectedRouteWrapper.tsx
```diff
@@ -13,16 +14,19 @@ import {
...
 
 export const SettingsProtectedRouteWrapper = ({
   children,
   settingsPermission,
+  anySettingsPermission,
   requiredFeatureFlag,
 }: SettingsProtectedRouteWrapperProps) => {
   const hasAccessTokenPair = useHasAccessTokenPair();
   const hasPermission = useHasPermissionFlag(settingsPermission);
+  const hasAnyPermission = useHasAnyPermissionFlag(anySettingsPermission);
```
FelixMalfait: Why are we adding useHasAnyPermissionFlag / these change around "hasAnyPermission" in this PR? I don't see how it's related to the rest / which bug it's fixing?

packages/twenty-front/src/pages/settings/members/tabs/SettingsWorkspaceMembersAccessTab.tsx
```diff
@@ -0,0 +1,261 @@
...
+  const workspaceInvitations = useAtomStateValue(workspaceInvitationsState);
+  const setWorkspaceInvitations = useSetAtomState(workspaceInvitationsState);
+
+  const { data: invitationsData, error: invitationsError } = useQuery(
+    GetWorkspaceInvitationsDocument,
+  );
+
+  useSnackBarOnQueryError(invitationsError);
+
+  useEffect(() => {
```
FelixMalfait: Is this a pre-existing useEffect or one we introduce? useEffect probably isn't necessary here, this is not how we usually do, fix it

packages/twenty-front/src/pages/settings/members/tabs/SettingsWorkspaceMembersAccessTab.tsx
```diff
@@ -0,0 +1,261 @@
...
+      enqueueErrorSnackBar({
+        message: t`Error resending invitation`,
+        options: {
+          duration: 2000,
+        },
+      });
+    }
+  };
+
+  const getExpiresAtText = (expiresAt: string) => {
```
FelixMalfait, line 123: Can't we use an existing utils that display something nicer? I know this is a refactor and probably already was there but could be a good opportunity to refactor if we find a more powerful util. Maybe there isn't an existing util and they display too approximate values though (e.g. 1 month instead of 27 days for example wouldn't be satisfying)

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-trigger/triggers/route/route-trigger.service.ts
```diff
@@ -63,11 +61,15 @@ export class RouteTriggerService {
       ),
     );
 
+    // App-scoped public domain → restrict matches to that app's logic functions.
```
FelixMalfait, line 64: This makes me think we should probably scope route by apps in the future, otherwise today conflicting routes from two apps create a bug. Let's study options (from adding a uuid to something nicer...) and create an issue to followup on that

packages/twenty-shared/src/types/SettingsPath.ts
```diff
@@ -85,9 +84,9 @@ export enum SettingsPath {
   AdminPanelApplicationRegistrationDetail = 'admin-panel/applications/registrations/:applicationRegistrationId',
   AdminPanelWorkspaceChatThread = 'admin-panel/workspaces/:workspaceId/threads/:threadId',
 
-  Roles = 'roles',
-  RoleCreate = 'roles/create',
-  RoleDetail = 'roles/:roleId',
-  RoleObjectLevel = 'roles/:roleId/object/:objectMetadataId',
-  RoleAddObjectLevel = 'roles/:roleId/add-object-permission',
+  Roles = 'members/roles',
```
FelixMalfait, line 87: Check if they are twenty-docs pages that had references to the urls we changed that need to be update

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-trigger/triggers/route/route-trigger.service.ts
FelixMalfait: Followup tracked at #20422 — proposed options outlined there.

PR #20369, opened 2026-05-07
Protect sendEmail endpoint and thread user context through logic function executor

martmull:
- Thread userId and userWorkspaceId through LogicFunctionExecutorService.execute() so
  application access tokens carry user context when available. This allows logic functions
  triggered by authenticated HTTP routes to call sendEmail with proper user identity, making
  the existing verifyOwnership() check work naturally.
  - Gate the sendEmail resolver with
  SettingsPermissionGuard(PermissionFlagType.SEND_EMAIL_TOOL) instead of NoPermissionGuard,
  ensuring only callers with the SEND_EMAIL_TOOL permission can send emails.


prastoin (commented):
Hey there, nice
We should in the best of the world create an integration test that covers this

FelixMalfait (approved):

PR #20383, opened 2026-05-08
fix: sort SELECT/MULTI_SELECT fields by option position instead of alphabetically

### LINKED ISSUE #2415: [Feature] Sort select fields by position
## Bug Description

- When apply "Sort" for Priority, it works normally (A)

Image

- When apply "Filter" above (Status), it failed to sort in order (B)

Image

## Expected behavior
As defined in Object field (Priority), Sort order should follow this order as well (work for A)

Image

But as you can see in (B), applying "Filter" will broke the sort order

QuantumByteMaster:
Fixes twentyhq/core-team-issues#2415

## Problem

When sorting by a SELECT or MULTI_SELECT field with an active filter, results were ordered alphabetically instead of by the defined option order.

Root cause: the order parser was applying `LOWER(column::text)` to SELECT fields, which sorts alphabetically. Without a filter this was hidden by default DB row order, but became visible once a filter changed the result set.

## Solution

For SELECT/MULTI_SELECT fields, generate `ARRAY_POSITION(ARRAY[...], column)` using the option positions stored in field metadata, so records sort by the user-defined option order regardless of whether a filter is active.

Example generated SQL:
```sql
ORDER BY ARRAY_POSITION(ARRAY['URGENT','HIGH','MEDIUM','LOW'], "task"."priority") ASC NULLS LAST


prastoin (commented):
Hey there thanks for your interest and contributing
Please do not jump on unverified issues but wait for product acknowledgment

PR #20386, opened 2026-05-08
i18n-website

hexven0m:


charlesBochet:
What you are doing: opening a PR that is already opened by somebody help (here somebody is an AUTOMATION) is actually harming the community. Thanks for playing against open-source.

PR #20412, opened 2026-05-09
Prevent non-admin users from impersonating admin users

FelixMalfait:
## Summary
- Adds a privilege check to workspace-level impersonation: non-admin users can no longer impersonate users who have `canAccessFullAdminPanel` or `canImpersonate` flags
- Adds the same check in JWT token validation as defense-in-depth (invalidates existing impersonation sessions targeting admin users)
- Adds 3 unit tests covering: non-admin → admin blocked, non-admin → canImpersonate blocked, admin → admin allowed

## Test plan
- [x] Unit tests pass (14/14 in `impersonation.service.spec.ts`)
- [x] Typecheck passes
- [ ] Verify workspace-level impersonation of regular users still works normally
- [ ] Verify server-level impersonation by admins is unaffected

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/auth/strategies/jwt.auth.strategy.ts
+++ b/packages/twenty-server/src/engine/core-modules/auth/strategies/jwt.auth.strategy.ts
@@ -338,6 +338,21 @@ export class JwtAuthStrategy extends PassportStrategy(Strategy, 'jwt') {
       );
     }
 
+    const targetHasAdminPrivileges =
+      impersonatedUserWorkspace.user.canImpersonate === true ||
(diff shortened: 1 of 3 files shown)


packages/twenty-server/src/engine/core-modules/auth/strategies/jwt.auth.strategy.ts
```diff
@@ -338,6 +338,21 @@ export class JwtAuthStrategy extends PassportStrategy(Strategy, 'jwt') {
       );
     }
```
FelixMalfait, line 340: Not 100% necessary but better check twice at 2 levels for security stuff

PR #20416, opened 2026-05-09
fix(front-component): preserve caret position on controlled input/textarea updates

### LINKED ISSUE #20409: Cursor jumps to end of <input>/<textarea> in front-component sandbox on every keystroke
## Bug Description

In a Twenty App front-component (running in the remote-DOM sandbox), a controlled `<input>` (or `<textarea>`) pre-populated with text cannot be edited in the middle: clicking inside the existing text and typing inserts the character at the correct position, but the caret then immediately jumps to the END of the input on every keystroke. This makes mid-string editing impossible - users would have to do something weird like cut from the insertion point to the end, then paste back again.

Typing at the end of an input is unaffected, which is why the bug went unnoticed for a while in components where users only ever appended (e.g. search fields).

**Minimum reproduction** 
Shoudl be able to drop into any front-component:

```tsx
import { useState } from 'react';

const Repro = () => {
  const [text, setText] = useState('Hello world');
  return (
    <input
      type="text"
      value={text}
      onChange={(e) => {
        const v = (e as any).detail?.value;
        if (typeof v === 'string') setText(v);
      }}
    />
  );
};
```
Click between Hell and o, type any character. Expected: caret at position 5. Observed: caret at position 11 (end). The character appears in between Hell and o, but the caret moves to the end.

## Expected behavior
The caret should remain at its current insertion point after a keystroke, matching native browser <input> behaviour and React's standard controlled-input handling in the main app DOM. Mid-string editing should be possible!

## Technical inputs
Reproduced on twenty-sdk@2.1.0 against a Twenty server on v2.2.0 (self-hosted). Same behaviour in two completely different React patterns, confirming this is a host-side value-application issue, not a React-pattern problem:

Parent-state-controlled list of inputs (e.g. extract-tasks-panel.tsx editing AI-proposed task titles)
React.memo'd sub-components with local state per row (the LineItemRow pattern from upstream / from our quote-sections-panel.tsx)

Things ruled out as React-side workarounds:
 - `defaultValue=` (uncontrolled) - the sandbox <input> doesn't honour it; input renders blank
 - `onChange={() => {}}` (no-op, hoping the DOM owns the value natively): typing is silently rejected; value never changes
 - Local state + `React.memo` isolation: siblings don't re-render but the cursor still jumps on every same-input keystroke
 - `useRef` to `setSelectionRange` after render: refs attach but DOM methods (.setSelectionRange, .focus, .value) are inaccessible from the worker (per the SDK's documented sandbox limits)

The behaviour seems to suggest that the sandbox <input> host wrapper applies value-prop changes via plain inputElement.value = X, which natively resets the caret to the end. Standard React in normal DOM would avoid this, I think, via its input-value-setter that preserves selectionStart/selectionEnd. The remote-DOM bridge appears not to do this.

Also affects `<textarea>`  (Incidentally, `<textarea>` IS supported in the sandbox, even though it doesn't appear in any upstream front-component code we've seen.)

QuantumByteMaster:
## Problem

In the front-component sandbox, typing in the middle of a pre-filled
`<input>` or `<textarea>` caused the caret to jump to the end on every
keystroke. Characters appeared at the correct position, but editing
mid-string was effectively broken.

Root cause: the remote-DOM bridge round-trips every keystroke through the
worker. By the time the updated `value` prop arrives back at the host,
React applies it by setting `inputElement.value = X` directly, which
browsers always reset the caret to the end.

Typing at the end was unaffected, which is why this went unnoticed in
search fields and similar append-only inputs.

## Fix

For text-like `<input>` types and `<textarea>`, the `value` prop is now
applied imperatively through a ref callback instead of being passed as a
React controlled prop:

- If the DOM value already matches the incoming prop, the assignment is
  skipped entirely.
- If a write is needed and the element is focused, `selectionStart` and
  `selectionEnd` are captured before the assignment and restored
  afterwards with `setSelectionRange`.

Non-text input types (checkbox, radio, file, color, range) and all other
host elements are unaffected.

## Testing

Drop the repro from the issue into any front-component, click between two
characters in the pre-filled value, and type — the caret should now stay
at the insertion point.

Fixes #20409


charlesBochet:
Thanks @QuantumByteMaster I'm taking a look, this will take a while as I need to understand in details what's happening

charlesBochet (approved):

charlesBochet:
Thank you, I've made fixes

PR #20426, opened 2026-05-10
fix(marketplace): garbage collect orphaned application registration rows

Pantkartik:
### Context

> Currently, the MarketplaceCatalogSyncService only performs upsert operations for NPM packages. This causes orphaned applicationRegistration rows to persist in the database when:

- A package's universalIdentifier (UUID) is rotated/changed between versions.
- A package is unpublished or removed from the marketplace registry.
- A package is republished under a different name or author but serves the same purpose.
- This leads to "ghost" entries in the marketplace catalog and non-functional install buttons due to UUID mismatches.

### Changes
This PR implements a garbage collection mechanism to keep the marketplace catalog in sync with the registry.

**ApplicationRegistrationService:**
Added unlistOrphanedRegistrations(activeUids: string[]): A bulk update method that sets isListed: false for any NPM-source registrations whose universalIdentifier is not in the list of currently active IDs.
Modified upsertFromCatalog: Ensures that if an existing (potentially previously unlisted) registration is found in a sync, it is marked as isListed: true.
**MarketplaceCatalogSyncService:**
Modified syncRegistryApps to collect syncedUniversalIdentifiers during the synchronization loop.
Triggers unlistOrphanedRegistrations post-loop to clean up stale entries.
Added error handling to ensure successful syncs are tracked even if some individual package fetches fail.
### Tests:

- Added unit tests for MarketplaceCatalogSyncService to verify correct tracking of active UIDs and handling of partial failures.
- Added unit tests for ApplicationRegistrationService to verify the bulk unlisting logic and upsert behavior.

### Infrastructure:
Updated packages/twenty-server/jest.config.mjs with twenty-shared and @/ module mappings to fix test execution in the twenty-server environment.
### Verification Results
Unit Tests: All tests in packages/twenty-server/src/engine/core-modules/application passed (103 total).
Manual Verification: Verified that rotation of UUIDs correctly hides the old registration while keeping the latest one functional.
### How to verify

- Trigger a catalog sync (e.g., via the scheduled task or manual trigger).
- Check the applicationRegistration table for NPM packages.
- Observe that old/orphaned registrations have isListed set to false.
- Search for NPM packages in the Marketplace UI; orphaned entries should no longer appear.


Pantkartik:
@charlesBochet @FelixMalfait

I've updated the logic to address the P1 safety violation regarding the marketplace GC:

Safety Guard: The cleanup now only runs if the sync completes successfully without critical errors.

Data Integrity: If the registry API returns an empty list or fails, the service skips unlisting to prevent accidental wipeouts.

CI Check: All server integration tests are now passing.

Ready for review!

packages/twenty-server/src/engine/workspace-manager/twenty-standard-application/utils/__tests__/get-standard-object-metadata-related-entity-ids.util.spec.ts
```diff
@@ -1,6 +1,7 @@
 import { getStandardObjectMetadataRelatedEntityIds } from 'src/engine/workspace-manager/twenty-standard-application/utils/get-standard-object-metadata-related-entity-ids.util';
 
-let uuidCounter = 0;
+// eslint-disable-next-line no-var
+var uuidCounter = 0;
```
prastoin, line 4: Why ?

prastoin (commented):
Please associate your PR to an existing issue
Once done please feel free to reach out I'll reopen this PR

Pantkartik:
Hello @prastoin, thanks for the feedback!

Regarding the var change: it was a workaround for the jest.mock hoisting issue where the factory closure couldn't access uuidCounter when declared with let. However, I'll update this to use mockImplementation inside beforeEach to avoid using var and keep the code clean.



I've pushed the fix for the test file. Could you please reopen this PR for review? Thank you!

PR #20427, opened 2026-05-10
Slack workflow connector

abdulrahmancodes:
https://github.com/user-attachments/assets/5a746414-988b-473c-9401-b8863a3e1c15

https://github.com/user-attachments/assets/0cdebdb1-f7c8-43cb-beef-f279387b6ce9

https://github.com/user-attachments/assets/df31c631-0781-42d8-8e6e-e5a16573ee3b

https://github.com/user-attachments/assets/6adaeae4-f3c9-4a5f-b0df-50c1f9a78428


packages/twenty-server/src/modules/connected-account/refresh-tokens-manager/services/connected-account-refresh-tokens.service.ts
```diff
@@ -117,6 +118,29 @@ export class ConnectedAccountRefreshTokensService {
           new Date(Date.now() - tokenExpirationTime)
         );
       }
+      case ConnectedAccountProvider.APP: {
```
abdulrahmancodes: Slack only provides access token and no refresh token if token rotation is not enabled. The previous flow threw before checking whether the access token was still usable, and ConnectedAccountTokens did not allow `refreshToken: null`.

packages/twenty-apps/internal/twenty-slack/src/logic-functions/handlers/slack-add-reaction-handler.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  if (!connectionResult.success) {
+    return {
+      success: false,
+      message: 'Slack is not connected',
+      error: connectionResult.error,
+    };
+  }
+
+  const client = new WebClient(connectionResult.accessToken);
```
martmull: refactor this to the same utils `getSlackClient`

packages/twenty-apps/internal/twenty-slack/src/application.config.ts
```diff
@@ -0,0 +1,29 @@
...
+  displayName: 'Twenty Slack',
+  description:
+    'Connect Slack to Twenty. Each workspace member (or a shared workspace connection) can authenticate Slack; workflow steps then post messages, ephemerals, updates, deletes, and reactions on behalf of that connection.',
+  logoUrl: 'public/twenty-slack.svg',
+  defaultRoleUniversalIdentifier: DEFAULT_ROLE_UNIVERSAL_IDENTIFIER,
+  serverVariables: {
+    SLACK_CLIENT_ID: {
+      description:
+        'OAuth client ID from your Slack app (api.slack.com/apps).',
+      isSecret: true,
```
martmull: why?

packages/twenty-apps/internal/twenty-slack/src/connection-providers/slack-connection.ts
```diff
@@ -0,0 +1,20 @@
...
+  type: 'oauth',
+  oauth: {
+    authorizationEndpoint: 'https://slack.com/oauth/v2/authorize',
+    tokenEndpoint: 'https://slack.com/api/oauth.v2.access',
+    revokeEndpoint: 'https://slack.com/api/auth.revoke',
+    scopes: ['chat:write', 'reactions:write'],
+    clientIdVariable: 'SLACK_CLIENT_ID',
+    clientSecretVariable: 'SLACK_CLIENT_SECRET',
+    tokenRequestContentType: 'form-urlencoded',
+    usePkce: false,
```
martmull: `usePkce` should be true for best security, why set it to false?

packages/twenty-apps/internal/twenty-slack/src/logic-functions/slack-update-message.ts
```diff
@@ -0,0 +1,44 @@
...
+    inputSchema: [
+      {
+        type: 'object',
+        properties: {
+          slackChannelId: { type: 'string' },
+          messageTimestamp: { type: 'string' },
+          newMessageText: { type: 'string' },
+          messageFormat: { type: 'string', enum: ['plain', 'markdown'] },
+        },
+      },
```
martmull: can't you infer workflowSetting.inputSchema from `slackUpdateMessageInputSchema`? I think we already have transform utils in the codebase and it will avoid dupicating the input schemas

martmull (commented):
Very nice, I have some comments about code style, then i will need to test the app when passed.

packages/twenty-apps/internal/twenty-slack/src/application.config.ts
```diff
@@ -0,0 +1,29 @@
...
+  APPLICATION_UNIVERSAL_IDENTIFIER,
+  DEFAULT_ROLE_UNIVERSAL_IDENTIFIER,
+} from 'src/constants/universal-identifiers';
+
+export default defineApplication({
+  universalIdentifier: APPLICATION_UNIVERSAL_IDENTIFIER,
+  displayName: 'Twenty Slack',
+  description:
+    'Connect Slack to Twenty. Each workspace member (or a shared workspace connection) can authenticate Slack; workflow steps then post messages, ephemerals, updates, deletes, and reactions on behalf of that connection.',
+  logoUrl: 'public/twenty-slack.svg',
```
martmull, line 13: we will need some more information for the marketplace

```typescript
...
  author?: string;
  category?: string;
  logoUrl?: string;
  screenshots?: string[];
  aboutDescription?: string;
  websiteUrl?: string;
  termsUrl?: string;
  emailSupport?: string;
  issueReportUrl?: string;
```

packages/twenty-apps/internal/twenty-slack/src/application.config.ts
abdulrahmancodes: Updated

packages/twenty-apps/internal/twenty-slack/src/logic-functions/slack-update-message.ts
abdulrahmancodes: Thanks, didn't know

packages/twenty-server/src/engine/core-modules/application/application-oauth-provider/refresh/types/app-oauth-tokens.type.ts
```diff
@@ -4,5 +4,5 @@
 
 export type AppOAuthTokens = {
   accessToken: string;
-  refreshToken: string;
```
charlesBochet, line 7: why this change?

packages/twenty-server/src/modules/connected-account/refresh-tokens-manager/services/connected-account-refresh-tokens.service.ts
```diff
@@ -117,6 +118,29 @@ export class ConnectedAccountRefreshTokensService {
...
+        if (!isDefined(connectedAccount.accessToken)) {
+          return false;
+        }
+
+        if (!isNonEmptyString(connectedAccount.refreshToken)) {
+          return true;
```
charlesBochet: why is the accessTokenValid if we have noRefreshToken?

charlesBochet (changes requested):

packages/twenty-server/src/modules/connected-account/refresh-tokens-manager/services/connected-account-refresh-tokens.service.ts
martmull: for logic function, no refresh token is provided (no need to because we have one access token for each execution with an expiry that equals the function timeout)

prastoin:
Hey there, feel free to reach out if you any question regarding conflicts you're facing with https://github.com/twentyhq/twenty/pull/20441
Please be sure to still encrypt your new integration token

packages/twenty-server/src/engine/core-modules/application/application-oauth-provider/refresh/types/app-oauth-tokens.type.ts
abdulrahmancodes: Slack only provides access token and no refresh token if token rotation is not enabled. The previous flow threw before checking whether the access token was still usable

packages/twenty-server/src/modules/connected-account/refresh-tokens-manager/services/connected-account-refresh-tokens.service.ts
charlesBochet: but in this case, why do we consider the accessToken always valid?

packages/twenty-server/src/engine/core-modules/application/application-oauth-provider/refresh/types/app-oauth-tokens.type.ts
charlesBochet: Ok

packages/twenty-server/src/modules/connected-account/refresh-tokens-manager/services/connected-account-refresh-tokens.service.ts
martmull: i was speaking of role related application access token, not connected account my bad

charlesBochet: @abdulrahmancodes can we adress that?

packages/twenty-server/src/modules/workflow/workflow-builder/workflow-schema/workflow-schema.workspace-service.ts
```diff
@@ -449,7 +449,8 @@ export class WorkflowSchemaWorkspaceService {
         };
       }
       case WorkflowActionType.CODE:
-      case WorkflowActionType.HTTP_REQUEST: {
+      case WorkflowActionType.HTTP_REQUEST:
+      case WorkflowActionType.LOGIC_FUNCTION: {
```
charlesBochet, line 453: could you explain this change too @abdulrahmancodes?

abdulrahmancodes: Resolving `Current item` for the iterator walks the source step’s `outputSchema` and infers the array element shape. That only ran for CODE/HTTP, so `LOGIC_FUNCTION` fell through to the default → generic leaf → the picker couldn’t expand `Current item` into fields like id / name. I added `LOGIC_FUNCTION` to the same branch because it uses the same outputSchema shape

packages/twenty-server/src/modules/connected-account/refresh-tokens-manager/services/connected-account-refresh-tokens.service.ts
abdulrahmancodes: @charlesBochet with no refresh token there’s nothing to call on the refresh path, so we return `true` here to take the "decrypt and reuse access token" branch instead of failing. So with no refresh token we always reuse the stored access token. There’s nothing else we can do until the user reconnects if the provider revokes or expires it.

martmull (approved):

packages/twenty-server/src/modules/connected-account/refresh-tokens-manager/services/connected-account-refresh-tokens.service.ts
charlesBochet: IMO, we should not modify this service and lie on internal implementation

we should not try to refresh the tokens at all if doesn't make sense. We probably need an util shouldTryRefreshTokens and gate the call to refreshAndSaveTokens higher in the code stack

charlesBochet (approved):

PR #20430, opened 2026-05-10
Refactor SAML relayState structure

prastoin:
# Introduction
Restructure the RelayState and avoid asserting the idp identifier from this opaque blob
Inferring the id from the secured validated and signed request params


martmull (approved):

prastoin:
E2E failing, checking if flaky. If not will investigate

prastoin:
On it

PR #20440, opened 2026-05-11
Stop rejecting application token on calendar and message events requests

### LINKED ISSUE #20423: Backend logic functions cannot authenticate to /graphql — runtime-injected TWENTY_APP_ACCESS_TOKEN is rejected and SDK cannot refresh it without a browser
# Backend logic functions cannot authenticate to `/graphql` — runtime-injected `TWENTY_APP_ACCESS_TOKEN` is rejected and SDK cannot refresh it without a browser

## Summary

A custom Twenty app with a `cron` (or any non-frontend) logic function
cannot make authenticated `/graphql` calls. Twenty's twenty-client-sdk
reads `process.env.TWENTY_APP_ACCESS_TOKEN`, the request is rejected
with `Authentication is required`, and the SDK's token-refresh path
relies on `globalThis.frontComponentHostCommunicationApi` which only
exists inside the React host — it's `undefined` on the worker pod.

Net effect: every Twenty app whose backend handler needs to read
workspace data (`calendarEvents`, `people`, `companies`, etc.) has to
ask the operator to paste a long-lived workspace API key as an
`applicationVariable`. That's a friction point that defeats the
"install from npm and you're done" story for `apps`.

## Versions

- Twenty: `v2.2.0` (chart-deployed, in-cluster Postgres + Redis + Linode S3)
- `twenty-sdk` / `twenty-client-sdk`: `2.3.0`

## Repro

Minimal app — one cron logic-function that reads `calendarEvents`:

```ts
import { CoreApiClient } from 'twenty-client-sdk/core';
import { defineLogicFunction } from 'twenty-sdk/define';

const handler = async () => {
  const client = new CoreApiClient(); // picks up TWENTY_APP_ACCESS_TOKEN
  const r = await client.query({
    calendarEvents: {
      __args: { first: 5 } as any,
      edges: { node: { id: true, title: true } },
    },
  } as any);
  console.log(r);
};

export default defineLogicFunction({
  universalIdentifier: '<v4-uuid>',
  name: 'probe',
  timeoutSeconds: 30,
  handler,
  cronTriggerSettings: { pattern: '* * * * *' },
});
```

Role grants `canReadAllObjectRecords: true`.

Deploy + install via SDK CLI:

```
yarn twenty deploy -r my-server
yarn twenty install -r my-server
```

## Expected

Cron tick reads workspace data successfully — install + go.

## Actual

Cron handler logs:

```
cron-dispatch: calendarEvents query failed: GenqlError: Authentication is required
errors: [ { message: 'Authentication is required', extensions: [Object] } ]
```

This happens regardless of role permissions — `canRead*All*ObjectRecords`
+ explicit per-object `objectPermissions` both granted, no change.

## Root cause (from reading the SDK)

`twenty-client-sdk/dist/core.mjs`:

```js
constructor(...) {
  // tokenFromHeaders is null in cron context
  this.authorizationToken =
    tokenFromHeaders
    ?? processEnvironment[APP_ACCESS_TOKEN_ENV_KEY]   // TWENTY_APP_ACCESS_TOKEN
    ?? processEnvironment[API_KEY_ENV_KEY]            // TWENTY_API_KEY
    ?? null;
}

shouldRefreshToken(response) {
  if (response.status === 401) return true;
  return hasAuthenticationErrorInGraphqlPayload(response.payload); // ← matched
}

async requestRefreshedAccessToken() {
  const refreshAccessTokenFunction =
    globalThis.frontComponentHostCommunicationApi?.requestAccessTokenRefresh;
  if (typeof refreshAccessTokenFunction !== "function") {
    return null;   // ← worker pod has no host-comm API; refresh always null
  }
  ...
}
```

So:

1. The injected `TWENTY_APP_ACCESS_TOKEN` (type `APPLICATION_ACCESS`,
   ~30min TTL based on inspecting the JWT) is rejected by `/graphql`
   with the in-payload "Authentication is required" error — even
   immediately after install when the token is fresh and the role
   has full read permissions.
2. The SDK detects auth failure and tries to refresh.
3. The refresh hook only exists inside the browser host
   (`frontComponentHostCommunicationApi`). On the worker, it's
   `undefined`.
4. Refresh returns null → original error propagates → handler fails.

It's possible that (1) is itself a separate bug (apps' tokens *should*
be accepted on `/graphql`); but even if (1) is fixed, the missing
backend refresh path in (3) means tokens minted at install time will
eventually expire with no recovery.

## Workaround in use today

Add a `TWENTY_API_KEY` `applicationVariable`, instruct the operator
to paste a workspace admin API key from Settings → APIs & Webhooks,
override the SDK's chosen token at runtime:

```ts
const adminKey = process.env.TWENTY_API_KEY ?? '';
delete process.env.TWENTY_APP_ACCESS_TOKEN;
const client = new CoreApiClient();
if (adminKey) (client as any).setAuthorizationToken?.(adminKey);
```

This is what we ship today (`@vexaai/twenty-app`) and what we'd like
to delete the moment the platform supports a real backend auth path.

## Suggested fixes (in priority order)

1. **Accept `APPLICATION_ACCESS` tokens on `/graphql`** for app-bound
   workspace queries, gated by the `defaultRole`'s permissions. If
   the gate is intentional, document it.
2. **Provide a backend refresh path.** Either inject a long-lived
   service-account-style token at handler boot (mint a fresh one
   server-side per invocation), or expose a non-browser callable for
   `requestAccessTokenRefresh` (e.g. an HTTP endpoint the worker can
   hit using the application's signing context).
3. **Document the gap in the apps SDK docs** so app authors stop
   re-discovering this pattern (we've seen multiple apps in the wild
   with the same paste-an-API-key workaround).

Happy to PR if there's directional agreement.

## Repo / context

- Vexa for Twenty: https://github.com/vexa-ai/vexa-twenty-app
- Affected handler: `app/src/logic-functions/cron-dispatch.ts`

martmull:
fixes https://github.com/twentyhq/twenty/issues/20423 by authorizing application token to perform calendarEvents and message queries

### DIFF
--- a/packages/twenty-server/src/engine/twenty-orm/repository/permissions.utils.ts
+++ b/packages/twenty-server/src/engine/twenty-orm/repository/permissions.utils.ts
@@ -119,7 +119,6 @@ export const validateOperationIsPermittedOrThrow = ({
     WORKSPACE_MEMBER_OBJECT_UNIVERSAL_IDENTIFIER;
 
   // TODO: this should be improved, we may have more complex permission configuration for is system objects
-  // see https://discord.com/channels/1130383047699738754/1503320724704854036 for context
   if (objectMetadataIsSystem && !isWorkspaceMemberObject) {
     return;
   }
--- a/packages/twenty-server/src/engine/twenty-orm/repository/permissions.utils.ts
+++ b/packages/twenty-server/src/engine/twenty-orm/repository/permissions.utils.ts
@@ -118,6 +118,8 @@ export const validateOperationIsPermittedOrThrow = ({
     objectMetadata.universalIdentifier ===
     WORKSPACE_MEMBER_OBJECT_UNIVERSAL_IDENTIFIER;
 
+  // TODO: this should be improved, we may have more complex permission configuration for is system objects
+  // see https://discord.com/channels/1130383047699738754/1503320724704854036 for context
   if (objectMetadataIsSystem && !isWorkspaceMemberObject) {
     return;
   }
--- a/packages/twenty-server/src/modules/calendar/common/query-hooks/calendar-event/calendar-event-find-many.post-query.hook.ts
+++ b/packages/twenty-server/src/modules/calendar/common/query-hooks/calendar-event/calendar-event-find-many.post-query.hook.ts
@@ -7,6 +7,8 @@ import { type WorkspaceAuthContext } from 'src/engine/core-modules/auth/types/wo
 import { ForbiddenError } from 'src/engine/core-modules/graphql/utils/graphql-errors.util';
 import { ApplyCalendarEventsVisibilityRestrictionsService } from 'src/modules/calendar/common/query-hooks/calendar-event/services/apply-calendar-events-visibility-restrictions.service';
 import { type CalendarEventWorkspaceEntity } from 'src/modules/calendar/common/standard-objects/calendar-event.workspace-entity';
+import { isApiKeyAuthContext } from 'src/engine/core-modules/auth/guards/is-api-key-auth-context.guard';
+import { isApplicationAuthContext } from 'src/engine/core-modules/auth/guards/is-application-auth-context.guard';
 
 @WorkspaceQueryHook({
   key: `calendarEvent.findMany`,
@@ -29,11 +31,13 @@ export class CalendarEventFindManyPostQueryHook
 
     // TODO: this check should be removed, see https://discord.com/channels/1130383047699738754/1503320724704854036 for context
     if (
-      authContext.type !== 'user' &&
-      authContext.type !== 'apiKey' &&
-      authContext.type !== 'application'
+      !isUserAuthContext(authContext) &&
+      !isApiKeyAuthContext(authContext) &&
+      !isApplicationAuthContext(authContext)
     ) {
-      throw new ForbiddenError('Authentication should be user scoped');
+      throw new ForbiddenError(
+        'Authentication error, auth context should be user, apiKey or application',
+      );
     }
 
     const workspace = authContext.workspace;
--- a/packages/twenty-server/src/modules/calendar/common/query-hooks/calendar-event/calendar-event-find-many.post-query.hook.ts
+++ b/packages/twenty-server/src/modules/calendar/common/query-hooks/calendar-event/calendar-event-find-many.post-query.hook.ts
@@ -29,7 +29,7 @@ export class CalendarEventFindManyPostQueryHook
     const isUserContext = isUserAuthContext(authContext);
     const userId = isUserContext ? authContext.user.id : undefined;
 
-    // TODO: this check should be removed, see https://discord.com/channels/1130383047699738754/1503320724704854036 for context
+    // TODO: this check should be removed
     if (
       !isUserAuthContext(authContext) &&
       !isApiKeyAuthContext(authContext) &&
--- a/packages/twenty-server/src/modules/calendar/common/query-hooks/calendar-event/calendar-event-find-many.post-query.hook.ts
+++ b/packages/twenty-server/src/modules/calendar/common/query-hooks/calendar-event/calendar-event-find-many.post-query.hook.ts
@@ -5,7 +5,6 @@ import { WorkspaceQueryHookType } from 'src/engine/api/graphql/workspace-query-r
 import { isUserAuthContext } from 'src/engine/core-modules/auth/guards/is-user-auth-context.guard';
 import { type WorkspaceAuthContext } from 'src/engine/core-modules/auth/types/workspace-auth-context.type';
 import { ForbiddenError } from 'src/engine/core-modules/graphql/utils/graphql-errors.util';
-import { TWENTY_STANDARD_APPLICATION } from 'src/engine/workspace-manager/twenty-standard-application/constants/twenty-standard-applications';
 import { ApplyCalendarEventsVisibilityRestrictionsService } from 'src/modules/calendar/common/query-hooks/calendar-event/services/apply-calendar-events-visibility-restrictions.service';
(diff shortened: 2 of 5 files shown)


FelixMalfait (approved):

packages/twenty-server/src/engine/twenty-orm/repository/permissions.utils.ts
```diff
@@ -118,6 +118,8 @@ export const validateOperationIsPermittedOrThrow = ({
     objectMetadata.universalIdentifier ===
     WORKSPACE_MEMBER_OBJECT_UNIVERSAL_IDENTIFIER;
 
+  // TODO: this should be improved, we may have more complex permission configuration for is system objects
+  // see https://discord.com/channels/1130383047699738754/1503320724704854036 for context
```
Weiko: If you really want to comment I'd link to a GitHub issue instead

packages/twenty-server/src/modules/calendar/common/query-hooks/calendar-event/calendar-event-find-many.post-query.hook.ts
```diff
@@ -28,17 +27,13 @@ export class CalendarEventFindManyPostQueryHook
     const isUserContext = isUserAuthContext(authContext);
     const userId = isUserContext ? authContext.user.id : undefined;
 
-    const isTwentyStandardApplication =
-      authContext.type === 'application' &&
-      authContext.application.universalIdentifier ===
-        TWENTY_STANDARD_APPLICATION.universalIdentifier;
-
+    // TODO: this check should be removed, see https://discord.com/channels/1130383047699738754/1503320724704854036 for context
```
Weiko: Same, this is not public

packages/twenty-server/src/modules/calendar/common/query-hooks/calendar-event/calendar-event-find-one.post-query.hook.ts
```diff
@@ -28,17 +27,13 @@ export class CalendarEventFindOnePostQueryHook
...
 
-    const isTwentyStandardApplication =
-      authContext.type === 'application' &&
-      authContext.application.universalIdentifier ===
-        TWENTY_STANDARD_APPLICATION.universalIdentifier;
-
+    // TODO: this check should be removed, see https://discord.com/channels/1130383047699738754/1503320724704854036 for context
     if (
-      !isUserContext &&
+      authContext.type !== 'user' &&
```
Weiko: What could be a possible value then? Is this check still relevant?

packages/twenty-server/src/modules/calendar/common/query-hooks/calendar-event/calendar-event-find-one.post-query.hook.ts
```diff
@@ -28,17 +27,13 @@ export class CalendarEventFindOnePostQueryHook
...
       authContext.type !== 'apiKey' &&
-      !isTwentyStandardApplication
+      authContext.type !== 'application'
     ) {
```
Weiko, line 37: Nit: I believe we have guards for that type of check

packages/twenty-server/src/modules/calendar/common/query-hooks/calendar-event/calendar-event-find-one.post-query.hook.ts
```diff
@@ -28,17 +27,13 @@ export class CalendarEventFindOnePostQueryHook
...
-      throw new ForbiddenError('Authentication is required');
+      throw new ForbiddenError('Authentication should be user scoped');
```
Weiko: I don't believe the message helps, especially if we accept api keys and applications

packages/twenty-server/src/modules/messaging/common/query-hooks/message/message-find-one.post-query.hook.ts
```diff
@@ -27,17 +24,13 @@ export class MessageFindOnePostQueryHook
...
-      authContext.application.universalIdentifier ===
-        TWENTY_STANDARD_APPLICATION.universalIdentifier;
-
+    // TODO: this check should be removed, see https://discord.com/channels/1130383047699738754/1503320724704854036 for context
     if (
-      !isUserAuthContext(authContext) &&
-      !isApiKeyAuthContext(authContext) &&
-      !isTwentyStandardApplication
+      authContext.type !== 'user' &&
+      authContext.type !== 'apiKey' &&
```
Weiko: Why removing guards?

packages/twenty-server/src/modules/calendar/common/query-hooks/calendar-event/calendar-event-find-one.post-query.hook.ts
martmull: indeed, updated it too quickly

packages/twenty-server/src/modules/calendar/common/query-hooks/calendar-event/calendar-event-find-one.post-query.hook.ts
martmull: ok

packages/twenty-server/src/engine/twenty-orm/repository/permissions.utils.ts
martmull: ok it is for us tbh

packages/twenty-server/src/modules/messaging/common/query-hooks/message/message-find-one.post-query.hook.ts
martmull: to unify, will add guards instead

FelixMalfait:
I will fix the E2E tests

PR #20441, opened 2026-05-11
Encrypt connected account `accessToken` and `refreshToken`

prastoin:
# Introduction
Encrypt the `connectedAccount` `accessToken` and `refreshToken` using `APP_SECRET` in order to mitigate potential data leak or `core` table compromise

## Decrypt
Temporary allow already plain text stored token to be retrieve without decryption until the slow instance has been passed
Will uncomment the invariant check in a patch when the instance slow has fully be run

## Standards
- Token are encrypted as quickly as possible
- A token cannot be written in database non encrypted by mistake using a custom constraint ( `enc:` prefix )

## What's next
We should standardize not managing secret as is in the the services and layer, they should be encrypted on the flight the earliest and should never be logged
Will create a dedicated pattern afterwards for `applicationVariables` secrets too


packages/twenty-server/src/database/commands/upgrade-version-command/2-5/__tests__/2-5-instance-command-slow-1798000004000-encrypt-connected-account-tokens.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+    const connectedAccountTokenEncryptionService =
+      buildFakeTokenEncryptionService();
+    const command = new EncryptConnectedAccountTokensSlowInstanceCommand(
+      connectedAccountTokenEncryptionService,
+    );
+
+    return { command, connectedAccountTokenEncryptionService };
+  };
+
+  describe('runDataMigration', () => {
```
martmull, line 112: nice, tests on migrations, i have not seen that since a while <3

packages/twenty-server/src/database/commands/upgrade-version-command/2-5/2-5-instance-command-slow-1798000004000-encrypt-connected-account-tokens.ts
```diff
@@ -0,0 +1,131 @@
...
+          sets.push(`"refreshToken" = $${params.length}`);
+        }
+
+        if (sets.length === 0) {
+          continue;
+        }
+
+        await dataSource.query(
+          `UPDATE "core"."connectedAccount"
+              SET ${sets.join(', ')}
```
martmull, line 94: we use params.length as values there i don't understand

packages/twenty-server/src/database/commands/upgrade-version-command/2-5/2-5-instance-command-slow-1798000004000-encrypt-connected-account-tokens.ts
```diff
@@ -0,0 +1,131 @@
...
+        if (
+          row.accessToken !== null &&
+          !row.accessToken.startsWith(CONNECTED_ACCOUNT_TOKEN_ENCRYPTION_PREFIX)
+        ) {
+          params.push(
+            this.connectedAccountTokenEncryptionService.encrypt(
+              row.accessToken,
+            ),
+          );
+          sets.push(`"accessToken" = $${params.length}`);
```
martmull, line 71: `params.length`? I don't get it

martmull: ahh ok, we use param identifers $1, $2, ... ok that is tough to read but as it is backend by a test why not. If you feel it, just use values for something more readable

martmull (approved):
approved with comment

packages/twenty-server/src/database/commands/upgrade-version-command/2-5/2-5-instance-command-slow-1798000004000-encrypt-connected-account-tokens.ts
prastoin: I don't want to leak the plain text existing access token and refresh token in the upgrade logs that's why

prastoin: I don't want to leak the plain text existing access token and refresh token in the upgrade logs that's why

Also following parameterized queries pattern here too

PR #20443, opened 2026-05-11
Stop rejecting application install when APP_VERSION is wrong

martmull:
as title

allows to install https://github.com/JordanChoo/twenty-multi-pipeline locally

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/application/application-package/application-version-validation.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/application/application-package/application-version-validation.service.ts
@@ -2,8 +2,8 @@ import { Injectable } from '@nestjs/common';
 
 import semver from 'semver';
 import { isDefined } from 'twenty-shared/utils';
-
-import { TwentyConfigService } from 'src/engine/core-modules/twenty-config/twenty-config.service';
+import { UpgradeMigrationService } from 'src/engine/core-modules/upgrade/services/upgrade-migration.service';
+import { extractVersionFromCommandName } from 'src/engine/core-modules/upgrade/utils/extract-version-from-command-name.util';
 
 export type VersionValidationFailureReason =
   | 'INVALID_REQUIRED_VERSION'
@@ -20,11 +20,13 @@ export type VersionValidationResult =
 
 @Injectable()
 export class ApplicationVersionValidationService {
-  constructor(private readonly twentyConfigService: TwentyConfigService) {}
+  constructor(
+    private readonly upgradeMigrationService: UpgradeMigrationService,
+  ) {}
 
-  validateServerCompatibility(
+  async validateServerCompatibility(
     requiredServerVersion: string | undefined,
-  ): VersionValidationResult {
+  ): Promise<VersionValidationResult> {
     if (!isDefined(requiredServerVersion)) {
       return { compatible: true };
     }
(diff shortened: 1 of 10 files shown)


prastoin (commented):
Hey there,
I don't think that's the best fix
We would still want to throw if the semver is invalid
Unless I'm mistaken what we want to fix is the marketplace on the dev env which does not have an APP_VERSION by definition
We should rather have an extra layer, such as an env variable or dedicated bypass rather than this global one

Weiko:
> Hey there, I don't think that's the best fix We would still want to throw if the semver is invalid Unless I'm mistaken what we want to fix is the marketplace on the dev env which does not have an APP_VERSION by definition We should rather have an extra layer, such as an env variable or dedicated bypass rather than this global one

I agree 👍

martmull:
i think the APP_VERSION validation should not belong to the install app operation

prastoin:
We might wanna provide a static value that would pass the configService validation that would serve only for main
=> But I think in your use case, and for most of them, this is totally decorelated
What we should be using is the instance version, inferred from the played instance commands and not the `APP_VERSION` which now only be used for display purposes

Even main technically has a version it's just unstable by definition

packages/twenty-server/src/engine/core-modules/application/application-package/application-version-validation.service.ts
```diff
@@ -37,21 +39,29 @@ export class ApplicationVersionValidationService {
...
+      await this.upgradeMigrationService.getLastAttemptedInstanceCommand();
+
+    const inferredServerVersion = isDefined(migration)
+      ? extractVersionFromCommandName(migration.name)
+      : undefined;
 
-    if (!isDefined(serverVersion) || !isDefined(semver.valid(serverVersion))) {
+    if (
+      !isDefined(inferredServerVersion) ||
+      !isDefined(semver.valid(inferredServerVersion))
```
prastoin, line 46: Just so you know the version returned here will never contain a patch different than 0
This means that a twenty app will only be able to declare MAJOR.MINOR peer dependencies to twenty

prastoin (approved):
Looks good to me !

packages/twenty-server/src/engine/core-modules/application/application-package/application-version-validation.service.ts
```diff
@@ -37,21 +39,29 @@ export class ApplicationVersionValidationService {
       };
     }
 
-    const serverVersion = this.twentyConfigService.get('APP_VERSION');
+    const migration =
+      await this.upgradeMigrationService.getLastAttemptedInstanceCommand();
```
Weiko: Not for now but we might want an higher level service to serve that info (this seems a bit complex right now when you don't really know how migrations work) (FYI @prastoin)

Weiko (approved):
Much better! LGTM!

packages/twenty-server/src/engine/core-modules/application/application-package/application-version-validation.service.ts
```diff
@@ -37,21 +39,29 @@ export class ApplicationVersionValidationService {
       };
     }
 
-    const serverVersion = this.twentyConfigService.get('APP_VERSION');
+    const migration =
```
charlesBochet: we should have a service for that, we should not re-implement it here

packages/twenty-server/src/engine/core-modules/application/application-package/application-version-validation.service.ts
charlesBochet: I would do it now :)

charlesBochet:
LGTM, but we should standardize version check

packages/twenty-server/src/engine/core-modules/application/application-package/application-version-validation.service.ts
martmull: done

PR #20438, opened 2026-05-11
feat(field-metadata): configurable default sort sub-field for FullName and Address

FelixMalfait:
## Summary

Adds a **Default Sort Sub-Field** advanced-mode setting on **FullName** and **Address** composite field types so admins can pick which sub-field a column sort defaults to. Closes the long-standing complaint that sorting by a "Full Name" column sorts by `firstName` only — admins can now pick `lastName`.

- **FullName** now uses the configured sub-field as the primary sort key, with the other as a tie-breaker (preserves the existing dual-key stable ordering).
- **Address becomes sortable** for the first time. Default sort sub-field falls back to `addressCity` when unset, or the first enabled sub-field if city is disabled.
- The Address default-sort dropdown is dynamically constrained to the currently-enabled `subFields` — disabling a sub-field removes it from the sort options, and the runtime falls back gracefully if a previously-saved default is later disabled.

## Why

Customer asked: "I want to sort my People table by lastName, but the Full Name column sorts by firstName." The backend already required the client to pick a sub-field — the frontend was hardcoding both sub-fields with same direction (firstName as primary). This change makes the choice configurable per field, with sensible defaults.

## What changed

**Backend:** No backend changes. The GraphQL order parser already required a sub-field for composite sorts — this PR just makes the frontend configurable about which one it sends. No DB migration needed; `FieldMetadata.settings` is JSONB and the new key is optional.

**Types (twenty-shared):**
- New `FullNameFieldsType.ts` with `ALLOWED_FULL_NAME_SUBFIELDS` / `AllowedFullNameSubField` (mirrors `AddressFieldsType.ts`)
- New `DEFAULT_VISIBLE_ADDRESS_SUBFIELDS` constant — canonical list of the 6 user-visible address sub-fields
- Extended `FieldMetadataSettings.ts` with `FieldMetadataFullNameSettings` and `defaultSortSubField` on `FieldMetadataAddressSettings`

**Sort logic (twenty-front):**
- `Address` added to `SORTABLE_FIELD_METADATA_TYPES`
- New helpers: `getDefaultSortSubFieldForFullName`, `getDefaultSortSubFieldForAddress`, `getEnabledAddressSubFields`
- `getOrderByForFieldMetadataType` now reads `field.settings.defaultSortSubField`

**Settings forms (twenty-front):**
- Address form gets a new "Default Sort Sub-Field" dropdown, constrained to enabled sub-fields via live `watch('settings.subFields')`
[...]

PR #20445, opened 2026-05-11
[breaking: deploy server before front] feat(view-sort): pick sort sub-field inline on the chip

FelixMalfait:
## Summary

Lets users choose which sub-field of a composite column to sort by — directly from the sort chip — by clicking the sub-field label and picking from a dropdown. Persists per view via a new nullable \`subFieldName\` column on \`ViewSort\`.

Replaces twentyhq/twenty#20438, which proposed a field-settings (admin) configuration for the same problem. The chip-level approach is more discoverable (the option lives where the user is looking) and per-view, so different views on the same object can sort by different sub-fields.

### What changes for users

- **FullName columns**: previously sorted by \`firstName\` and \`lastName\` together as a stable dual-key sort. Now the user can pick which sub-field is primary (the other is the tie-breaker). Default remains \`firstName\` primary, \`lastName\` tie-breaker.
- **Address columns**: previously not sortable at all (not in \`SORTABLE_FIELD_METADATA_TYPES\`). Now sortable, with a chip dropdown listing each enabled sub-field. Default is \`addressCity\` if enabled, else the first enabled sub-field. Disabling a sub-field at the field-metadata level (existing setting) removes it from the dropdown.
- **Other composite types** (Currency, Phones, Emails, Links, Actor) and scalar fields keep their existing single-key sort behavior.

### UX

```
┌─────────────────────────┐    ┌─────────────────────────┐
│ ↑ Name · Last name  ✕ │    │ ↑ Address · City  ✕ │
└────────┬────────────────┘    └────────┬────────────────┘
         ▼ (click sub-field)            ▼
   ┌────────────┐                ┌────────────┐
   │ First name │                │ Address 1  │
   │ Last name ✓│                │ Address 2  │
   └────────────┘                │ City      ✓│
                                 │ State      │
                                 │ Postcode   │
                                 │ Country    │
                                 └────────────┘
```

The chip body still toggles direction on click — the \`Dropdown\`'s internal wrapper calls \`stopPropagation\` so the sub-field click doesn't bubble to the chip's onClick.

## What changed

**Backend:**
- \`ViewSortEntity\` — new nullable \`subFieldName: varchar\` column
- \`ViewSortDTO\`, \`CreateViewSortInput\`, \`UpdateViewSortInputUpdates\` — new \`@Field(() => String, { nullable: true })\`
- \`FLAT_VIEW_SORT_EDITABLE_PROPERTIES\` — \`'subFieldName'\` added so the property flows through the update merge path
- \`ALL_ENTITY_PROPERTIES_CONFIGURATION_BY_METADATA_NAME.viewSort\` — new \`subFieldName\` entry with \`toCompare: true\` so cache diffs notice it
- \`fromCreateViewSortInputToFlatViewSortToCreate\` — threads \`subFieldName\` through
- Instance command migration (\`add-sub-field-name-to-view-sort\`) — single \`ALTER TABLE core.viewSort ADD subFieldName varchar\` / \`DROP\`

**Frontend:**
- \`RecordSort\` and \`ViewSort\` types — \`subFieldName?: string | null\`
- \`VIEW_SORT_FRAGMENT\` — adds \`subFieldName\` so the field round-trips
- \`mapRecordSortToViewSort\` + \`areViewSortsEqual\` — carry the new field through, include it in the diff so the usual \`useSaveRecordSortsToViewSorts\` create/update flow fires when it changes
- \`useSaveRecordSortsToViewSorts\` — passes \`subFieldName\` in both \`CreateViewSortInput\` and \`UpdateViewSortInputUpdates\`
- \`getOrderByForFieldMetadataType(field, direction, subFieldName?)\` — new optional third arg. \`turnSortsIntoOrderBy\` threads \`sort.subFieldName\` into it.
- \`Address\` added to \`SORTABLE_FIELD_METADATA_TYPES\`
- New helpers: \`getEnabledAddressSubFields\` (filters by the field's \`subFields\` setting, falls back to the 6 default visible address sub-fields), \`getDefaultSortSubFieldForAddress\`, \`getDefaultSortSubFieldForFullName\`
- New shared types/constants: \`AllowedFullNameSubField\`, \`ALLOWED_FULL_NAME_SUBFIELDS\`, \`DEFAULT_VISIBLE_ADDRESS_SUBFIELDS\`
- \`SortOrFilterChip\` — new \`labelSubField?: ReactNode\` slot; renders as \` · {sub-field}\` with subdued weight after the main label
- \`EditableSortChip\` — builds options from field metadata (\`ALLOWED_FULL_NAME_SUBFIELDS\` for FullName, \`getEnabledAddressSubFields\` for Address), uses i18n-wrapped labels, persists picks via \`upsertRecordSort\`

## Test plan

- [x] \`npx nx typecheck\` passes for twenty-shared, twenty-front, twenty-server
- [x] \`oxlint --type-aware\` on all 19 frontend + 9 server changed files: 0 errors
- [x] \`prettier --check\`: clean
- [x] 16 unit tests pass — \`getOrderByForFieldMetadataType\` covers the new \`subFieldName\` override branch for FULL_NAME and ADDRESS; \`getDefaultSortSubFieldForAddress\` covers the city/first-enabled fallback path; \`getDefaultSortSubFieldForFullName\` exercises its constant
- [ ] Manual: sort a People view by Full Name → click the chip's sub-field label → switch between First name and Last name → reload page → choice is preserved
- [ ] Manual: sort a Company view by Address → confirm dropdown lists only enabled sub-fields → disable Address \`addressCity\` in field settings → confirm dropdown options update and runtime falls back to the first enabled sub-field


packages/twenty-front/src/modules/object-metadata/constants/AddressDefaultSortSubField.ts
```diff
@@ -0,0 +1,4 @@
+import { type AllowedAddressSubField } from 'twenty-shared/types';
```
charlesBochet: shared types naming seems off: does not includes "sort"

packages/twenty-front/src/modules/object-metadata/utils/__tests__/getOrderByForFieldMetadataType.test.ts
```diff
@@ -0,0 +1,89 @@
...
+    });
+
+    it('uses the per-sort subFieldName as the primary sort key', () => {
+      const field = buildField({
+        type: FieldMetadataType.FULL_NAME,
+        name: 'name',
+      });
+
+      const result = getOrderByForFieldMetadataType(
+        field,
```
charlesBochet: getOrderByForFieldMetadataType({ field, orderByDirection, compositeSubField? }) would be clearer than getOrderByForFieldMetadataType(a, b, c)

packages/twenty-front/src/modules/object-metadata/utils/__tests__/resolveAddressSortSubField.test.ts
```diff
@@ -0,0 +1,50 @@
+import { resolveAddressSortSubField } from '@/object-metadata/utils/resolveAddressSortSubField';
+
+describe('resolveAddressSortSubField', () => {
+  it('returns the requested sub-field when it is enabled', () => {
+    expect(
+      resolveAddressSortSubField(
```
charlesBochet: same

packages/twenty-front/src/modules/object-metadata/utils/__tests__/resolveAddressSortSubField.test.ts
```diff
@@ -0,0 +1,50 @@
+import { resolveAddressSortSubField } from '@/object-metadata/utils/resolveAddressSortSubField';
+
+describe('resolveAddressSortSubField', () => {
```
charlesBochet, line 3: why do we need an util for that, is that dynamic?

charlesBochet: understood, I didn't know it was possible in current product :p

packages/twenty-front/src/modules/object-metadata/utils/getEnabledAddressSubFields.ts
```diff
@@ -0,0 +1,19 @@
...
+  type FieldMetadataType,
+} from 'twenty-shared/types';
+
+type AddressSettings = FieldMetadataSettingsMapping[FieldMetadataType.ADDRESS];
+
+export const getEnabledAddressSubFields = (
+  settings: FieldMetadataItem['settings'] | null | undefined,
+): readonly AllowedAddressSubField[] => {
+  const addressSettings = settings as AddressSettings | null | undefined;
+  if (addressSettings?.subFields && addressSettings.subFields.length > 0) {
```
charlesBochet: isNonEmptyArray?

packages/twenty-front/src/modules/object-metadata/utils/getEnabledAddressSubFields.ts
```diff
@@ -0,0 +1,19 @@
...
+import {
+  type AllowedAddressSubField,
+  type FieldMetadataSettingsMapping,
+  type FieldMetadataType,
+} from 'twenty-shared/types';
+
+type AddressSettings = FieldMetadataSettingsMapping[FieldMetadataType.ADDRESS];
+
+export const getEnabledAddressSubFields = (
+  settings: FieldMetadataItem['settings'] | null | undefined,
```
charlesBochet: I thought it was possible to type FieldMetadataItem<Address>['settings] but I might be wrong here, just double check. This part of the typing has holes

packages/twenty-front/src/modules/object-metadata/utils/getOrderByForFieldMetadataType.ts
```diff
@@ -1,32 +1,51 @@
...
-  field: Pick<FieldMetadataItem, 'id' | 'name' | 'type'>,
+  field: Pick<FieldMetadataItem, 'id' | 'name' | 'type' | 'settings'>,
   direction: OrderBy | null | undefined,
+  subFieldName?: string | null,
 ): RecordGqlOperationOrderBy => {
   switch (field.type) {
-    case FieldMetadataType.FULL_NAME:
+    case FieldMetadataType.FULL_NAME: {
+      const primarySubField = resolveFullNameSortSubField(subFieldName);
+      const [firstSubField, lastSubField] = ALLOWED_FULL_NAME_SUBFIELDS;
```
charlesBochet: this part looks convoluted to me, not sure why we don't directly reference ALLOWED_FULL_NAME_SUBFIELDS

packages/twenty-front/src/modules/object-metadata/utils/resolveFullNameSortSubField.ts
```diff
@@ -0,0 +1,21 @@
...
+import { ALLOWED_FULL_NAME_SUBFIELDS } from 'twenty-shared/constants';
+import { type AllowedFullNameSubField } from 'twenty-shared/types';
+import { isDefined } from 'twenty-shared/utils';
+
+const isAllowedFullNameSubField = (
+  value: string | null | undefined,
+): value is AllowedFullNameSubField =>
+  ALLOWED_FULL_NAME_SUBFIELDS.includes(value as AllowedFullNameSubField);
+
+export const resolveFullNameSortSubField = (
```
charlesBochet: I'm not a fan of this API, maybe it's just naming but I have trouble understanding it.

Maybe: resolvePrimaryFullNameSortSubField({ requestedPrimarySubField? })

packages/twenty-shared/src/types/AllowedFullNameSortSubField.ts
```diff
@@ -0,0 +1,4 @@
+import { type ALLOWED_FULL_NAME_SORT_SUBFIELDS } from '@/constants/AllowedFullNameSortSubfields';
+
+export type AllowedFullNameSortSubField =
+  (typeof ALLOWED_FULL_NAME_SORT_SUBFIELDS)[number];
```
FelixMalfait, line 4: @claude there is inconsistent naming for types across this /types/ folder, please create an issue in twentyhq/twenty to fix it ; assign it to CoreTeamTwenty. Decide which way to go based on what is more common across the codebase

packages/twenty-front/src/modules/views/editable-chip/components/EditableSortChip.tsx
```diff
@@ -1,51 +1,178 @@
...
 
   const { fieldMetadataItem } = useFieldMetadataItemByIdOrThrow(
     recordSort.fieldMetadataId,
   );
 
-  const handleClick = () => {
-    const newSort: RecordSort = {
+  const subFieldDropdownId = `sort-subfield-${recordSort.fieldMetadataId}`;
+
+  const subFieldState = useMemo<SubFieldState | undefined>(() => {
```
FelixMalfait: useMemo necessary?

prastoin (commented):
Hey there, just so you know that now by default the CD always deploys the server, waiting for full sync, before releasing the front

packages/twenty-front/src/modules/views/editable-chip/components/EditableSortChip.tsx
FelixMalfait: Done — removed the `useMemo` and now just compute it as a plain function call per render. Work is cheap (string lookups + a small map), so memoization wasn't pulling its weight.

packages/twenty-front/src/modules/object-metadata/constants/AddressDefaultSortSubField.ts
FelixMalfait: Inlined the single-use default-sort-sub-field constants into their resolvers (`firstName` / `addressCity`) and deleted the standalone constant files in the latest commit. On the type-naming asymmetry: `AllowedFullNameSortSubField` was renamed to include `Sort` because it's sort-specific (`firstName`/`lastName` only), but `AllowedAddressSubField` is intentionally generic — it's used outside of sort too (`AddressInput`, `formatAddressDisplay`, `joinAddressFieldValues`, `FieldMetadataSettings.subFields`, etc.), so adding `Sort` to its name would be misleading.

packages/twenty-front/src/modules/object-metadata/utils/__tests__/getOrderByForFieldMetadataType.test.ts
FelixMalfait: Done in earlier commits — `getOrderByForFieldMetadataType` now takes `{ field, orderByDirection, compositeSubField }`, and `getOrderByForRelationField` got the same treatment.

packages/twenty-front/src/modules/object-metadata/utils/getOrderByForFieldMetadataType.ts
```diff
@@ -1,101 +1,140 @@
...
   switch (field.type) {
-    case FieldMetadataType.FULL_NAME:
+    case FieldMetadataType.FULL_NAME: {
+      const primarySubField = resolvePrimaryFullNameSortSubField({
+        requestedPrimarySubField: compositeSubField,
+      });
+      const secondarySubField =
+        primarySubField === 'firstName' ? 'lastName' : 'firstName';
+      const direction = orderByDirection ?? 'AscNullsLast';
+      // Each sub-field is its own array entry rather than two keys inside
```
FelixMalfait: @claude remove this comment and push to the branch directly

packages/twenty-front/src/modules/object-metadata/utils/__tests__/getOrderByForRelationField.test.ts
```diff
@@ -52,22 +52,18 @@ describe('getOrderByForRelationField', () => {
...
+    const result = getOrderByForRelationField({
       field,
       relatedObjectMetadataItem,
-      'DescNullsLast',
-    );
+      orderByDirection: 'DescNullsLast',
+    });
 
-    // Should produce nested structure with composite field
+    // Composite labels emit one array entry per sub-field — see the
```
FelixMalfait: Remove comment

packages/twenty-front/src/modules/object-metadata/utils/__tests__/getOrderByForFieldMetadataType.test.ts
```diff
@@ -0,0 +1,93 @@
...
+      ]);
+    });
+
+    it('uses the per-sort compositeSubField as the primary sort key', () => {
+      const field = buildField({
+        type: FieldMetadataType.FULL_NAME,
+        name: 'name',
+      });
+
+      expect(
```
charlesBochet, line 38: nit: still looks odd: compositeSubField => primaryCompositeSubField, as we return several orderBy including other subfields

packages/twenty-front/src/modules/views/editable-chip/components/EditableSortChip.tsx
```diff
@@ -1,51 +1,181 @@
...
 type EditableSortChipProps = {
   recordSort: RecordSort;
 };
 
+type SubFieldOption = {
+  value: AllowedFullNameSortSubField | AllowedAddressSubField;
+  label: string;
+};
+
+type SubFieldState = {
```
charlesBochet: state is a bad suffix for frontend variable as it has a strong meaning (jotai)

packages/twenty-front/src/modules/views/editable-chip/components/EditableSortChip.tsx
```diff
@@ -1,51 +1,181 @@
...
 
   const { fieldMetadataItem } = useFieldMetadataItemByIdOrThrow(
     recordSort.fieldMetadataId,
   );
 
-  const handleClick = () => {
-    const newSort: RecordSort = {
+  const subFieldDropdownId = `sort-subfield-${recordSort.fieldMetadataId}`;
+
+  const buildSubFieldState = (): SubFieldState | undefined => {
```
charlesBochet: code quality: I don't like that we add Full Name and Address edge case in EditableSortChip, not very scalable. I would have added simple utils / abstraction here

packages/twenty-front/src/modules/views/editable-chip/components/EditableSortChip.tsx
```diff
@@ -1,51 +1,181 @@
...
-    upsertRecordSort(newSort);
+    });
   };
 
+  const handleSubFieldSelect = (value: string) => {
+    upsertRecordSort({ ...recordSort, subFieldName: value });
+    closeDropdown(subFieldDropdownId);
+  };
+
+  // Fence the dropdown from the chip's outer onClick (which toggles
```
charlesBochet: codesmell, seems we are hacking here. We either want the dropdown behavior, either we want to toggle direction on click

charlesBochet (commented):
LGTM except EditableSortChip that had its codequality downgraded (see comment)!

Nice product addition! Can you also add screen recording of new behavior with or without composite case?

+ complete integration tests on viewSort

packages/twenty-front/src/modules/views/editable-chip/components/EditableSortChip.tsx
FelixMalfait: Agreed

packages/twenty-front/src/modules/views/editable-chip/components/EditableSortChip.tsx
FelixMalfait: I can check but it felt normal to me to have a hacky code because the product behavior by itself is hacky, I inserted a link that opens a dropdown within a clickable chip. Not sure if it's the right long term ux, I think the best would have been a chip divided into 3 clickable part (field | subfield | x) but we don't have that component and I don't think it's worth doing it now

FelixMalfait: But Claude: please explore and review all alternatives / tell me if there are other behaviors/paths that are less hacky I should consider

PR #20455, opened 2026-05-11
fix(kanban): give title full width when card is not hovered

FelixMalfait:
## Summary

On kanban cards, the title was being truncated even when the checkbox wasn't displayed. The checkbox is hidden via `opacity: 0` on the card's non-hovered state, which keeps it in flex flow and still reserves its ~24px of width — so the title's flex item was shrinking unnecessarily.

This change collapses the checkbox container's `max-width` to `0` (with `overflow: hidden`) while it's hidden, and expands it back to the checkbox's natural size (`spacing[6]` = 24px) on hover or when selected. The existing `transition: all ease-in-out 160ms` animates the title expanding into the freed space.

### Before
Title truncates with ellipsis even though the checkbox slot is empty:

before

### After
Title uses the full row width when not hovered; the checkbox slides in on hover (or when the card is selected) and the title reflows.

### Tooltip
The full title is already exposed on hover when truncated — `RecordChip` → `Chip` already wraps the label in `OverflowingTextWithTooltip`, which detects overflow (`scrollWidth > clientWidth`) and renders an `AppTooltip` with the full text. No additional wiring needed.

## Test plan

- [ ] On a kanban board, verify a long record title now uses the full card width when the card is not hovered (no ellipsis if the title fits).
- [ ] Hover the card: the checkbox slides in smoothly (animated), and the title reflows (may now truncate if it doesn't fit).
- [ ] Hover the (now-truncated) title: tooltip with the full title appears.
- [ ] Select the card via the checkbox: checkbox stays visible (and title stays in its hover-state width) without hovering.
- [ ] Compact view (eye icon) still renders correctly.


charlesBochet:
@FelixMalfait Can you add before and after screen recording? Fix looks legit but hard to review CSS without screen capture. Also make sure that all added CSS properties are actually useful (we are doing max-width: 0 + overflow: hidden here, unlikely that they are both useful)

FelixMalfait:
How it looks after:
Screenshot 2026-05-12 at 11 08 12
Screenshot 2026-05-12 at 11 08 27

FelixMalfait:
Pushed 80756b6 addressing your review points:

**Charles' redundancy concern** — you were right, dropped `overflow: hidden`. After tracing each property's role:
- `max-width: 0` is the load-bearing one (collapses the flex slot so the title gets its space back)
- `opacity: 0` handles the visual hide + fade transition
- `overflow: hidden` was just clipping the painted output, which `opacity: 0` already accomplishes for the hidden state

So `overflow: hidden` was actually only doing work in the *shown* state — and it was the cause of the cropped hover ring (next point).

**Cropped hover ring on the right (Felix)** — with `overflow: hidden` on a container whose `max-width` equalled the inner Checkbox's intrinsic 24px, the right edge of the hover background was getting clipped by ~1px. Removing `overflow: hidden` makes the boundary irrelevant since nothing exceeds the container in the shown state.

**Side-effect mitigation** — without `overflow: hidden`, the hidden-state checkbox (opacity 0) still has its 24px bounding box, which would extend past the card's right edge and could intercept clicks in the gutter. Gated it with `pointer-events: none/auto`.

**Tooltip (Felix)** — confirmed it's already wired through `RecordChip → Chip → OverflowingTextWithTooltip`. The component checks `scrollWidth > clientWidth` on mouseenter and renders an `AppTooltip` with the full title when it's truncated. No code change needed; it should already show on the kanban titles on the preview env.

FelixMalfait:
Pushed 28c5628 — you were right, the tooltip wasn't actually firing on the kanban. Root cause turned out to be in LinkChip, not the kanban code:

Walking the DOM tree from the title text element outward on a kanban card with an overflowing title, the widths were:

| element | scrollWidth | clientWidth |
|---|---|---|
| Text DIV (`StyledOverflowingText`) | 205 | 205 |
| Chip `StyledContainer` (max-width: 100%) | 227 | 227 |
| Link `<a>` (max-width: none) | 227 | 227 |
| StyledLinkContainer SPAN | 227 | 184 ← first real clip |
| ancestors with `overflow: hidden` | … visual clipping | |

`OverflowingTextWithTooltip` only checks `scrollWidth > clientWidth` on the text element itself — and the text element thinks it's at content width because the constraint chain breaks at the `<a>`. The `<a>` had `max-width: none` so it grew to its content (chip's content), and the chip's `max-width: 100%` of an unconstrained parent is content-width too.

Fix: propagate the parent's width by giving the `<a>` `max-width: 100%; min-width: 0` in LinkChip's `StyledLinkContainer`. Now `scrollW=205 / clientW=142 / overflowing=true`, tooltip renders correctly on hover, and table-view chips still detect overflow normally (verified `Enterprise iPad Deployment` scrollW=171, clientW=124).

Why table view always worked: the table cell width directly constrains the chip via its grid layout, so the `<a>`-grows-to-content issue never bit there. Kanban hits it because the chip is in a regular flex column without a fixed-width cell ancestor.

kanban-tooltip-fix

packages/twenty-front/src/modules/object-record/record-board/record-board-card/components/RecordBoardCardHeader.tsx
```diff
@@ -41,6 +41,11 @@ const StyledRecordChipContainer = styled.div`
   display: flex;
   flex: 1 1 auto;
   overflow: hidden;
+
+  & a {
```
charlesBochet: we usually avoid that, would be better to fix the behavior of the link itself. Open to exceptions though

FelixMalfait:
Moved back to LinkChip in 443dcea. The change is `max-width: 100%; min-width: 0` on the inner `<a>` (replacing default `max-width: none / min-width: auto`).

**Areas to eyeball — places where the chip would previously grow past a narrow container and now truncates with ellipsis instead:**

LinkChip's consumers are: `RecordChip` (used everywhere records appear inline), `MentionRecordChip` (rich-text @-mentions), `RecordLink` (AI/copilot record refs), and the workspace name in `SettingsAdminWorkspaceContent`.

The risky case in each is: parent narrower than the chip's natural width *and* no `overflow: hidden` ancestor (i.e. the chip was previously overflowing its container visibly). After this change those chips truncate cleanly at the container boundary.

Spots worth a quick visual check:
- **@-mentions inside a narrow rich-text block** — e.g. comment thread in a side-panel-width note, or a multi-line mention chain in a tasks side panel. Long display names that used to overflow the paragraph will now ellipsize.
- **AI copilot chat bubbles** with record links (`RecordLink`) — chat bubble is narrow, long record names that previously stretched the bubble will now ellipsize inside it.
- **Activity timeline entries** in the side panel — a long related-record name in a "linked X with Y" sentence used to push past the entry width; now it truncates. Tooltip fires on hover so the full name is still discoverable.
- **Workspace name in admin panel** — typically a wide key-value row, very low risk.

Spot-checked locally in:
- Kanban (`By Stage` opportunities) — 47 chips correctly overflow-detect; tooltip fires on hover ✓
- Table view — chips still overflow correctly (`scrollW > clientW` unchanged) ✓
- Detail view — relation chips and timeline references render normally ✓

Worth a careful eye on the preview env before merge — particularly mentions in notes/tasks and the AI chat panel if you use those.

PR #20456, opened 2026-05-11
Fix(UI): Side panel having two scrollbars

### LINKED ISSUE #20417: Side panel shows two scrollbars
## Bug Description
In the right-side panel, the list currently has two separate scrollbars. When scrolling to the bottom, the first scroll ends before the full list is reached, requiring users to scroll again inside another container to access the remaining items.
This creates a confusing user experience and makes navigation less smooth.

A screenshot demonstrating the issue:
Image

## Expected behavior

The side panel list should have only one scrollbar, allowing users to scroll through the entire list smoothly without needing to interact with multiple scroll areas.

## Technical inputs
It appears there may be an unnecessary wrapper element causing the issue.

**Example:** 
The `StyledSidePanelList` div may not be needed. Even with `overflow-y: hidden`( It just has this one css property ), its presence seems to create an additional scroll layer, resulting in the double scroll behavior. Removing this element may resolve the problem.

git-init-priyanshu:
Fixes: #20417

Screenshot:
image


charlesBochet (approved):
Makes sense, thank you!

PR #20458, opened 2026-05-11
feat(front-component-renderer): forward file input metadata

charlesBochet:
## Summary

`<input type=\"file\">` inside front-components was silently non-functional:
- The host-side `serializeEvent` did not read `target.files`, so the worker received an empty `onChange` detail.
- `SerializedEventData` had no `files` field.
- The `html-input` schema in `AllowedHtmlElements` exposed neither `accept`, `multiple`, nor `capture` — the worker could not even configure the picker.

This PR forwards file metadata (`name`, `size`, `type`, `lastModified`) through the existing serialized event detail and accepts the missing attributes on the `html-input` remote element. A new Storybook play test guards the regression by uploading single and multiple files via `userEvent.upload`.

Reading file contents inside the worker is intentionally out of scope here and will need a separate host API bridge (the host has the `File` objects on the real input element; passing bytes through `postMessage` is a bigger design call).

### DIFF
--- a/packages/twenty-front-component-renderer/src/host/utils/createHtmlHostWrapper.ts
+++ b/packages/twenty-front-component-renderer/src/host/utils/createHtmlHostWrapper.ts
@@ -72,7 +72,9 @@ const parseCssString = (
   return style;
 };
 
-const serializeFileList = (files: unknown): SerializedFileData[] | undefined => {
+const serializeFileList = (
+  files: unknown,
+): SerializedFileData[] | undefined => {
   if (!isObject(files)) return undefined;
   const fileListLike = files as { length?: unknown } & Record<number, unknown>;
   if (!isNumber(fileListLike.length)) return undefined;
--- a/packages/twenty-front-component-renderer/src/host/utils/createHtmlHostWrapper.ts
+++ b/packages/twenty-front-component-renderer/src/host/utils/createHtmlHostWrapper.ts
@@ -53,7 +53,9 @@ const parseCssString = (
 
   for (const declaration of declarations) {
     const colonIndex = declaration.indexOf(':');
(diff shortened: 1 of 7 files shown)


packages/twenty-front-component-renderer/src/host/utils/createHtmlHostWrapper.ts
```diff
@@ -69,6 +72,37 @@ const parseCssString = (
   return style;
 };
 
+const serializeFileList = (
+  files: unknown,
+): SerializedFileData[] | undefined => {
+  if (!isObject(files)) return undefined;
```
martmull: code style return in brackets

packages/twenty-front-component-renderer/src/host/utils/createHtmlHostWrapper.ts
```diff
@@ -126,6 +160,9 @@ const serializeEvent = (event: unknown): SerializedEventData => {
     if (isBoolean(targetRecord.muted)) serialized.muted = targetRecord.muted;
     if (isNumber(targetRecord.playbackRate))
       serialized.playbackRate = targetRecord.playbackRate;
+
+    const files = serializeFileList(targetRecord.files);
+    if (isDefined(files)) serialized.files = files;
```
martmull: brackets

martmull (approved):

packages/twenty-front-component-renderer/src/host/utils/createHtmlHostWrapper.ts
charlesBochet: yes, the whole file is like this. I'll make a commit to fix the whole file

PR #20465, opened 2026-05-11
fix(ci): probe real schema in breaking-changes server readiness check

FelixMalfait:
## Summary

The `GraphQL and OpenAPI Breaking Changes Detection` workflow has been posting graphql-inspector stack traces as PR comments — see [#20445 comment](https://github.com/twentyhq/twenty/pull/20445#issuecomment-4421142635) for an example.

### Root cause

- The wait step probed readiness with `curl -s URL > /dev/null 2>&1`, which exits 0 for **any** HTTP response — including 5xx and GraphQL error JSON. NestJS opens the HTTP listener before the workspace schema cache is fully populated, so the wait often completed while the server still served auth/metadata error JSON.
- The introspection download therefore wrote a small (~154-byte) error payload instead of the real schema. `jq empty` in the validation step only checks JSON *syntax*, so `{"errors":[...]}` passed validation.
- `graphql-inspector diff` then failed with `Unable to read JSON file: ... Not valid JSON content`, the workflow swallowed the error into the diff markdown, and the bot posted that stack trace verbatim on the PR.

In the failing run, the main-branch files were 154 B (GraphQL) and 112 B (REST 500); the current-branch files in the same run were 600 KB–2.8 MB.

### Fix

- Wait steps now POST an authenticated introspection (`{ __schema { queryType { name } } }`) and require `.data.__schema` plus a 2xx response from `/rest/open-api/core` (`curl -f`) before declaring the server ready.
- Validation step now checks for the expected shape (`.data.__schema` for GraphQL, `.openapi`/`.swagger` for OpenAPI) and includes the first 200 bytes of any bad payload in the warning, so when something genuinely goes wrong the next debugger has a real lead instead of a generic stack trace.

## Test plan

- [ ] CI runs against this branch — the workflow's own readiness probes are now exercised against the real server, so a green run validates the new check.
- [ ] If the readiness probe still passes but downloads regress, the strengthened validation step will surface the payload in the workflow logs instead of posting a graphql-inspector stack trace on the PR.

### DIFF
--- a/.github/workflows/ci-breaking-changes.yaml
+++ b/.github/workflows/ci-breaking-changes.yaml
@@ -166,14 +166,11 @@ jobs:
 
           ADMIN_TOKEN=$(jq -r '.APPLE_JANE_ADMIN_ACCESS_TOKEN' packages/twenty-server/test/integration/constants/test-tokens.json)
 
-          # The HTTP listener accepts connections before the workspace schema cache is
-          # populated, so a plain port-open check passes while introspection still errors.
-          # Probe with an authenticated introspection and require a 2xx OpenAPI response.
           while [ $elapsed -lt $timeout ]; do
             GRAPHQL_RESPONSE=$(curl -s -X POST "http://localhost:${{ env.CURRENT_SERVER_PORT }}/graphql" \
               -H "Content-Type: application/json" \
               -H "Authorization: Bearer ${ADMIN_TOKEN}" \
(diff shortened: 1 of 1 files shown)


charlesBochet:
Looks not ready for review

FelixMalfait:
It works: https://github.com/twentyhq/twenty/actions/runs/25712443942

charlesBochet (approved):

PR #20467, opened 2026-05-11
feat(server): asymmetric JWT signing with kid + key rotation table

charlesBochet:
## Context

Today every JWT issued by Twenty (access, refresh, login, file, etc.) is HMAC-signed with a per-token-type secret derived from the global `APP_SECRET`. Rotating that secret invalidates **every** active token at once and there is no way to scope a leak to a subset of tokens.

This PR is the first slice of a broader effort to **decouple stateful encryption (`APP_SECRET`-derived secrets) from stateless encryption (JWTs)**. It introduces an asymmetric (private/public key) signing path for `ACCESS` and `REFRESH` tokens and a signing-key registry to enable **safe rotation**: leaked keys can be revoked by flipping `revokedAt`/`isCurrent` on the matching row without invalidating tokens issued by other keys.

> Out of scope (intentionally): swapping stateful encryption for `APP_SECRET`, asymmetric signing for token types other than `ACCESS`/`REFRESH`, an admin-panel rotation UI, and an enterprise re-encryption command. Those will land in follow-up PRs.

## What changes

- **New `core.signingKey` table** (instance command `2.5.0` / `1778550000000`) storing both the public key (PEM, in clear) and the private key (PEM, encrypted with `APP_SECRET` via `SecretEncryptionService`). One row is marked `isCurrent = true` (enforced by a partial unique index). The row's UUID `id` is used directly as the JWT `kid`.
  - When a key is rotated out, its `privateKey` is nulled (we never keep historical private keys) but the `publicKey` row stays so previously issued tokens can still be verified.
- **`JwtKeyManagerService`** lazily loads-or-generates the current signing key on first use:
  - If a row with `isCurrent = true` exists → decrypts and uses it.
  - Otherwise → generates a fresh EC P-256 keypair, encrypts the private key, inserts the row (UUID id = kid). Handles concurrent insert races via the unique constraint.
- **`JwtWrapperService.signAsync()`** signs `ACCESS`/`REFRESH` payloads with `ES256` and a `kid` header. Falls back to `HS256` if no signing key is available (boot-time DB error, transient failure).
- **Dual-path verification** in both `JwtWrapperService.verifyJwtToken` and the Passport `JwtAuthStrategy.secretOrKeyProvider`:
  - JWT with a `kid` header → resolve the public key PEM by id and verify with `ES256`,
  - otherwise → fall back to the existing `APP_SECRET`-derived `HS256` path (unchanged).
- **`AccessTokenService` / `RefreshTokenService`** now sign through `signAsync` (single public surface; the routing detail stays internal to the wrapper).
- **Public key cache**: a new `SigningKeyEntityCacheProviderService` plugs into `CoreEntityCacheService` (`signingKeyPublicKey` namespace) and serves PEMs by id, with the standard local-memo + Redis layering.
- **PEM strings end-to-end**: `jsonwebtoken` accepts PEM strings directly for both sign and verify, so the manager never converts to a Node `KeyObject` and the cache hands the PEM straight to `jwt.verify`.

## Why ES256 (and not EdDSA / RS256)

- `@nestjs/jwt` is backed by `jsonwebtoken`, which does **not** support EdDSA today.
- ES256 keys are tiny (~120 bytes vs 1.6 kB for RS256), signatures are short (~64 bytes), and signing/verification is fast — important since JWT verification runs on every authenticated request.
- ES256 is widely supported and standardized (RFC 7518), with mature ecosystem support.

## Why store the private key in DB (not env)

- No new secret to provision: existing instances already have `APP_SECRET`, which we reuse to encrypt the private key at rest.
- Self-healing: a fresh instance auto-generates its first signing key on first boot. Nothing to copy/paste.
- Rotation is a SQL operation against `core.signingKey`, not a redeploy + env mutation.

## Backward compatibility

- All previously-issued tokens (no `kid`) keep verifying through the legacy HS256 path with their existing `APP_SECRET`-derived secret. No forced re-login.
- Token types not in scope (`WORKSPACE_AGNOSTIC`, `API_KEY`, `FILE`, `LOGIN`, `EMAIL_VERIFICATION`, etc.) keep their current HS256 behavior unchanged — they still go through the synchronous `JwtWrapperService.sign(payload, options)` with a caller-supplied secret.
- `signWithAppSecret` is kept intentionally as the HS256 fallback path; it will be deprecated in a follow-up PR.
- If the DB lookup/generation fails for any reason, the wrapper logs the error and falls back to HS256 — no startup crash, no silent regression.

## Rotation story

1. Bootstrap: first signing call lazily inserts a row in `core.signingKey` with `isCurrent = true`, `privateKey = encrypt(pem_A)`. New tokens carry `kid_A`.
2. Rotate: `UPDATE core."signingKey" SET "isCurrent" = false, "privateKey" = NULL WHERE id = '<kid_A>';` then insert a new row with `isCurrent = true`. New tokens carry `kid_B`. Tokens still in flight with `kid_A` keep verifying because the public-key row for `kid_A` is still there.
3. Revoke: `UPDATE core."signingKey" SET "revokedAt" = now() WHERE id = '<kid_A>';`. All tokens with `kid_A` now fail verification cleanly with `UNAUTHENTICATED` (no 500).
4. Tokens with no `kid` (legacy) are unaffected throughout.

## Test plan

- [x] Unit: `JwtWrapperService` dual-path verification (HS256 no-kid vs ES256 with-kid), unknown-kid → `UNAUTHENTICATED`, `signAsync` happy path + `null` when no key, `signAsync` rejection for non-rotatable types.
- [x] Unit: `JwtAuthStrategy` `secretOrKeyProvider` dual-path resolution and algorithm validation.
- [x] All existing JWT/auth/application unit tests adjusted to the renamed public method.
- [x] Integration (`jwt-key-rotation.integration-spec.ts`):
  - **Happy path**: signed-up user's `ACCESS` token has `alg=ES256` + correct UUID `kid`, the `isCurrent=true` row exists in `core.signingKey`, `getCurrentUser` resolves.
  - **Legacy fallback**: hand-crafted no-kid HS256 token verifies via the legacy `APP_SECRET`-derived path.
  - **Previous-key rotation**: token signed by a hardcoded *previous* key whose row is pre-inserted with `privateKey = NULL` (rotated-out) still verifies — proves the leaked-key revocation flow works in both directions.
  - **Unknown kid**: token signed with an orphan UUID `kid` is cleanly rejected (no 500).
- [x] `npx nx test twenty-server`
- [x] `npx nx run twenty-server:lint`


packages/twenty-server/src/database/commands/upgrade-version-command/2-5/2-5-instance-command-fast-1778550000000-create-jwt-public-key-table.ts
```diff
@@ -0,0 +1,34 @@
...
+@RegisteredInstanceCommand('2.5.0', 1778550000000)
+export class CreateJwtPublicKeyTableFastInstanceCommand
+  implements FastInstanceCommand
+{
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE TABLE IF NOT EXISTS "core"."jwtPublicKey" (
+        "id" uuid NOT NULL DEFAULT uuid_generate_v4(),
+        "kid" varchar(128) NOT NULL,
+        "publicKey" text NOT NULL,
```
charlesBochet: We usually don't use text in this project. Is there a strong reason why here?

packages/twenty-server/src/database/commands/upgrade-version-command/2-5/2-5-instance-command-fast-1778550000000-create-jwt-public-key-table.ts
```diff
@@ -0,0 +1,34 @@
...
+
+@RegisteredInstanceCommand('2.5.0', 1778550000000)
+export class CreateJwtPublicKeyTableFastInstanceCommand
+  implements FastInstanceCommand
+{
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE TABLE IF NOT EXISTS "core"."jwtPublicKey" (
+        "id" uuid NOT NULL DEFAULT uuid_generate_v4(),
+        "kid" varchar(128) NOT NULL,
```
charlesBochet: we usually use varchar varying in this project. Even if it makes sense here, is there any real advantage worse breaking the consensus

packages/twenty-server/src/database/commands/upgrade-version-command/2-5/2-5-instance-command-fast-1778550000000-create-jwt-public-key-table.ts
```diff
@@ -0,0 +1,34 @@
...
+        "publicKey" text NOT NULL,
+        "algorithm" varchar(32) NOT NULL DEFAULT 'ES256',
+        "status" varchar(16) NOT NULL DEFAULT 'active',
```
charlesBochet: remove status, introduce revokedAt

packages/twenty-server/src/engine/core-modules/auth/strategies/jwt.auth.strategy.ts
```diff
@@ -46,7 +49,26 @@ export class JwtAuthStrategy extends PassportStrategy(Strategy, 'jwt') {
   ) {
     const jwtFromRequestFunction = jwtWrapperService.extractJwtFromRequest();
     // @ts-expect-error legacy noImplicitAny
-    const secretOrKeyProviderFunction = async (_request, rawJwtToken, done) => {
+    const secretOrKeyProviderFunction = (_request, rawJwtToken, done) => {
+      const kid = extractKidFromTokenHeader(rawJwtToken);
+
+      if (typeof kid === 'string' && kid.length > 0) {
```
charlesBochet: use typeGuard: isNonEmptyString

packages/twenty-server/src/engine/core-modules/auth/strategies/jwt.auth.strategy.ts
```diff
@@ -46,7 +49,26 @@ export class JwtAuthStrategy extends PassportStrategy(Strategy, 'jwt') {
...
+        jwtKeyManagerService
+          .getPublicKeyByKid(kid)
+          .then((publicKey) => {
+            if (publicKey === null) {
+              done(new Error('Unknown JWT signing key'), null);
```
charlesBochet: shoudn't we throw AuthExction here?

getPublicKeyByKid => getActivePublicKeybyKid (we should exclude revoked ones)

packages/twenty-server/src/engine/core-modules/auth/strategies/jwt.auth.strategy.ts
```diff
@@ -466,3 +489,19 @@ export class JwtAuthStrategy extends PassportStrategy(Strategy, 'jwt') {
     );
   }
 }
+
+const extractKidFromTokenHeader = (rawJwtToken: string): string | undefined => {
+  try {
+    const decoded = jwt.decode(rawJwtToken, { complete: true });
+
+    if (decoded === null || typeof decoded === 'string') {
```
charlesBochet: same

packages/twenty-server/src/engine/core-modules/auth/token/services/access-token.service.ts
```diff
@@ -141,6 +141,15 @@ export class AccessTokenService {
       impersonatedUserWorkspaceId: payloadOriginalUserWorkspaceId,
     };
 
+    const asymmetricToken = await this.jwtWrapperService.signAsymmetric(
+      jwtPayload,
+      { expiresIn },
+    );
+
+    if (asymmetricToken !== null) {
```
charlesBochet: same

packages/twenty-server/src/engine/core-modules/auth/token/services/refresh-token.service.ts
```diff
@@ -137,19 +133,32 @@ export class RefreshTokenService {
...
+      sub: payload.userId,
+      type: JwtTokenTypeEnum.REFRESH,
+    };
+
+    const asymmetricToken = await this.jwtWrapperService.signAsymmetric(
+      jwtPayload,
+      { expiresIn, jwtid: refreshToken.id },
+    );
+
+    if (asymmetricToken !== null) {
```
charlesBochet: same

packages/twenty-server/src/engine/core-modules/auth/token/services/refresh-token.service.ts
```diff
@@ -137,19 +133,32 @@ export class RefreshTokenService {
 
     await this.appTokenRepository.save(refreshToken);
 
+    const jwtPayload: RefreshTokenJwtPayload = {
+      ...payload,
+      sub: payload.userId,
+      type: JwtTokenTypeEnum.REFRESH,
+    };
+
+    const asymmetricToken = await this.jwtWrapperService.signAsymmetric(
```
charlesBochet: I feel we should refactor to hide je whole signing logic within jwtWrapperService. We should have a .sign({ jwtPayload }, {expiresIn, jwtId} and internally we have signAsymetric and signWithAppSecret method

packages/twenty-server/src/engine/core-modules/jwt/services/__tests__/jwt-key-manager.service.spec.ts
```diff
@@ -0,0 +1,206 @@
...
+  });
+
+  return {
+    privatePem: privateKey.export({ format: 'pem', type: 'pkcs8' }).toString(),
+    publicPem: publicKey.export({ format: 'pem', type: 'spki' }).toString(),
+    publicJwk: publicKey.export({ format: 'jwk' }),
+  };
+};
+
+const buildRepository = (initialRows: Array<Record<string, unknown>> = []) => {
```
charlesBochet: remove this test

packages/twenty-server/src/engine/core-modules/jwt/services/__tests__/jwt-wrapper.service.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+  const publicJwk = publicKey.export({ format: 'jwk' });
+
+  return {
+    privateKey,
+    publicKey,
+    kid: computeJwkThumbprint(publicJwk),
+  };
+};
+
+describe('JwtWrapperService', () => {
```
charlesBochet: same, we don't write service spec anymore. If there is any complex business logic that we want to test it should be extracted to utils and test unitary

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-key-manager.service.ts
```diff
@@ -0,0 +1,209 @@
...
+  const canonical = canonicalizeJwk(jwk);
+  const digest = createHash('sha256').update(canonical).digest();
+
+  return base64UrlEncode(digest);
+};
+
+const canonicalizeJwk = (jwk: JsonWebKey): string => {
+  if (jwk.kty === 'EC') {
+    if (
+      typeof jwk.crv !== 'string' ||
```
charlesBochet: use typeguard everywhere (isDefined, isNonEmptySgtring)

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-key-manager.service.ts
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    try {
+      await this.jwtPublicKeyRepository.insert({
+        kid: args.kid,
+        publicKey: args.publicKeyPem,
+        algorithm: 'ES256',
+        status: 'active',
+      });
+    } catch (error) {
```
charlesBochet: this service should throw CustomException everywhere

charlesBochet: we should create our own class JWTKeyManagerException

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-key-manager.service.ts
```diff
@@ -0,0 +1,209 @@
...
+    return publicKey;
+  }
+
+  computeKidFromPublicKey(publicKey: KeyObject): string {
+    const jwk = publicKey.export({ format: 'jwk' });
+
+    return computeJwkThumbprint(jwk);
+  }
+
+  resetForTests(): void {
```
charlesBochet: remove, we usually don't have test code in our real services

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-key-manager.service.ts
```diff
@@ -0,0 +1,209 @@
...
+};
+
+@Injectable()
+export class JwtKeyManagerService {
+  private readonly logger = new Logger(JwtKeyManagerService.name);
+
+  private currentSigningKeyPromise: Promise<CurrentSigningKey | null> | null =
+    null;
+
+  private readonly publicKeyCache = new Map<string, KeyObject>();
```
charlesBochet: I wonder if we should rather use our cache system like we have done for users

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-key-manager.service.ts
```diff
@@ -0,0 +1,209 @@
...
+      typeof jwk.x !== 'string' ||
+      typeof jwk.y !== 'string'
+    ) {
+      throw new Error('Invalid EC JWK: missing crv, x or y');
+    }
+
+    return JSON.stringify({ crv: jwk.crv, kty: 'EC', x: jwk.x, y: jwk.y });
+  }
+
+  if (jwk.kty === 'RSA') {
```
charlesBochet: is this needed? I thought we were only supporting ES256 for now?

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
```diff
@@ -132,4 +153,64 @@ export class JwtWrapperService {
...
+    } catch (error) {
+      throw this.toAuthException(error);
+    }
+  }
+
+  private tryDecodeHeader(token: string): jwt.JwtHeader | undefined {
+    try {
+      const decoded = jwt.decode(token, { complete: true });
+
+      if (decoded === null || typeof decoded === 'string') {
```
charlesBochet: same everywhere

packages/twenty-server/test/integration/graphql/suites/auth/jwt-key-rotation/jwt-key-rotation.integration-spec.ts
```diff
@@ -0,0 +1,255 @@
...
+  appSecretBody: string,
+): string =>
+  createHash('sha256')
+    .update(`${HS256_APP_SECRET}${appSecretBody}${type}`)
+    .digest('hex');
+
+const decodeJwtCompleteOrThrow = (token: string) => {
+  const decoded = jwt.decode(token, { complete: true });
+
+  if (decoded === null || typeof decoded === 'string') {
```
charlesBochet: use typeguard evrerywhere

packages/twenty-server/test/integration/graphql/suites/auth/jwt-key-rotation/jwt-key-rotation.integration-spec.ts
```diff
@@ -0,0 +1,255 @@
...
+
+      expect(errors).toBeUndefined();
+      expect(data?.currentUser).toBeDefined();
+    });
+  });
+
+  describe('Legacy HS256 fallback (no kid)', () => {
+    it('verifies a hand-crafted no-kid HS256 ACCESS token via the legacy fallback', async () => {
+      const { payload } =
+        await signUpFreshUserAndCreateWorkspace('hs256-fallback');
```
charlesBochet: why do we need to create workspaces and users to tests keys? we should only test the happy path for token generation already done above but then to test thant previous kid still valid, that old jwt is still working we just have to generate the right jwt without needing to sign up or create workspace

packages/twenty-server/src/engine/core-entity-cache/types/core-entity-cache-key.type.ts
```diff
@@ -6,12 +9,22 @@ export type CoreEntityCacheDataMap = {
...
+type CoreEntityCacheKeyConfig = {
+  prefix: string;
+  isValidId: (entityId: string) => boolean;
+};
+
+export const CORE_ENTITY_CACHE_KEYS: Record<
+  CoreEntityCacheKeyName,
+  CoreEntityCacheKeyConfig
+> = {
+  workspaceEntity: { prefix: 'workspace', isValidId: isValidUuid },
```
charlesBochet: why this change? How validity is a concern of the cache key here?
Should be responsibility of higher level service

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-key-manager.service.ts
```diff
@@ -0,0 +1,172 @@
...
+  private readonly logger = new Logger(JwtKeyManagerService.name);
+
+  private currentSigningKeyPromise: Promise<CurrentSigningKey | null> | null =
+    null;
+
+  constructor(
+    private readonly twentyConfigService: TwentyConfigService,
+    @InjectRepository(JwtPublicKeyEntity)
+    private readonly jwtPublicKeyRepository: Repository<JwtPublicKeyEntity>,
+    private readonly coreEntityCacheService: CoreEntityCacheService,
```
charlesBochet, line 35: this cace should JWTPublicKeyEntityCacheProviderService, see how it's done for user

packages/twenty-server/src/engine/core-entity-cache/services/core-entity-cache.service.ts
```diff
@@ -84,7 +84,7 @@ export class CoreEntityCacheService implements OnModuleInit {
   ): Promise<CoreEntityCacheDataMap[K] | null> {
     this.evictExpiredLocalEntries();
 
-    if (!isDefined(entityId) || !isValidUuid(entityId)) {
+    if (!isDefined(entityId)) {
```
charlesBochet: let's revert this change now

charlesBochet:
image

image

packages/twenty-server/src/database/commands/upgrade-version-command/2-5/2-5-instance-command-fast-1778550000000-create-signing-key-table.ts
```diff
@@ -0,0 +1,34 @@
...
+@RegisteredInstanceCommand('2.5.0', 1778550000000)
+export class CreateSigningKeyTableFastInstanceCommand
+  implements FastInstanceCommand
+{
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE TABLE IF NOT EXISTS "core"."signingKey" (
+        "id" uuid NOT NULL DEFAULT uuid_generate_v4(),
+        "publicKey" character varying NOT NULL,
+        "privateKey" character varying,
```
prastoin, line 15: ***Question:*** ~~Shouldn't this be encrypted ?~~ => It is, should we introduce the constraint here as discussed here too ? or is it planned for later ?

packages/twenty-server/src/engine/core-modules/auth/strategies/jwt.auth.strategy.ts
```diff
@@ -46,21 +49,62 @@ export class JwtAuthStrategy extends PassportStrategy(Strategy, 'jwt') {
   ) {
     const jwtFromRequestFunction = jwtWrapperService.extractJwtFromRequest();
     // @ts-expect-error legacy noImplicitAny
```
prastoin: ***Polish:*** lets strictly type params here

packages/twenty-server/src/engine/core-modules/auth/strategies/jwt.auth.strategy.ts
```diff
@@ -466,3 +511,58 @@ export class JwtAuthStrategy extends PassportStrategy(Strategy, 'jwt') {
...
+
+const ASYMMETRIC_TOKEN_TYPES: ReadonlySet<JwtTokenTypeEnum> = new Set([
+  JwtTokenTypeEnum.ACCESS,
+  JwtTokenTypeEnum.REFRESH,
+]);
+
+const shouldUseAsymmetricKey = (args: {
+  header: jwt.JwtHeader | undefined;
+  payload: JwtPayload | undefined;
+}): boolean => {
```
prastoin: ***Suggestion:*** would be cool if this returned a predicate making the `kid` defined as string in the header

***Suggestion:*** make a dedicated util

packages/twenty-server/src/engine/core-modules/auth/strategies/jwt.auth.strategy.ts
```diff
@@ -46,21 +49,62 @@ export class JwtAuthStrategy extends PassportStrategy(Strategy, 'jwt') {
...
+                ),
+                null,
+              );
+
+              return;
+            }
+
+            done(null, publicKeyPem);
+          })
+          .catch((error) => done(error, null));
```
prastoin: ***Polish:*** Lets avoid combining chaining promises and fluent promises patterns

packages/twenty-server/src/engine/core-modules/auth/strategies/jwt.auth.strategy.ts
```diff
@@ -46,21 +49,62 @@ export class JwtAuthStrategy extends PassportStrategy(Strategy, 'jwt') {
...
+            : 'workspaceId' in decodedPayload
+              ? decodedPayload.workspaceId
+              : undefined;
+
+        if (!isNonEmptyString(appSecretBody)) {
+          throw new AuthException(
+            'Token invalid.',
+            AuthExceptionCode.UNAUTHENTICATED,
+          );
+        }
```
prastoin: ***Hunch:*** Would decompose in two explicit code flows:
- Legacy jwt decode process
- Target one with public key

packages/twenty-server/src/engine/core-modules/auth/token/services/access-token.service.ts
```diff
@@ -141,16 +141,11 @@ export class AccessTokenService {
...
-      expiresAt,
-    };
+    const token = await this.jwtWrapperService.signAsync(jwtPayload, {
+      expiresIn,
+    });
+
+    return { token, expiresAt };
   }
 
   async validateToken(token: string): Promise<AuthContext> {
```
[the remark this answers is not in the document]
prastoin: out of scope but true

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
```diff
@@ -132,4 +155,131 @@ export class JwtWrapperService {
...
+    }
+
+    if ('userId' in payload && isNonEmptyString(payload.userId)) {
+      return payload.userId;
+    }
+
+    return undefined;
+  }
+
+  private shouldVerifyAsymmetric(args: {
```
prastoin: ***Polish:*** duplicated

prastoin (commented):
first review

packages/twenty-server/src/engine/core-modules/auth/token/services/application-token.service.ts
```diff
@@ -148,11 +148,11 @@ export class ApplicationTokenService {
...
+  async validateApplicationAccessToken(
     token: string,
-  ): ApplicationAccessTokenJwtPayload {
+  ): Promise<ApplicationAccessTokenJwtPayload> {
     try {
-      this.jwtWrapperService.verifyJwtToken(token);
+      await this.jwtWrapperService.verifyJwtToken(token);
 
       const payload =
         this.jwtWrapperService.decode<ApplicationAccessTokenJwtPayload>(token, {
```
prastoin, line 158: ***Polish:*** out of scope but redundant decode here, could consume the decoded token from the verify method

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
```diff
@@ -132,4 +155,131 @@ export class JwtWrapperService {
...
+      return jwt.verify(args.token, publicKeyPem, {
+        ...args.options,
+        algorithms: ['ES256'],
+      });
+    } catch (error) {
+      throw this.toAuthException(error);
+    }
+  }
+
+  private tryDecodeHeader(token: string): jwt.JwtHeader | undefined {
```
prastoin: duplicated

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
```diff
@@ -132,4 +155,131 @@ export class JwtWrapperService {
...
+
+    if (!isDefined(signingKey)) {
+      return null;
+    }
+
+    return jwt.sign(payload as object, signingKey.privateKeyPem, {
+      ...options,
+      algorithm: 'ES256',
+      keyid: signingKey.id,
+    } as jwt.SignOptions);
```
prastoin: why force cast ?

packages/twenty-server/src/engine/core-modules/jwt/services/signing-key-entity-cache-provider.service.ts
```diff
@@ -0,0 +1,31 @@
...
+  constructor(
+    @InjectRepository(SigningKeyEntity)
+    private readonly signingKeyRepository: Repository<SigningKeyEntity>,
+  ) {
+    super();
+  }
+
+  async computeForCache(entityId: string): Promise<string | null> {
+    const row = await this.signingKeyRepository.findOne({
+      where: { id: entityId, revokedAt: IsNull() },
```
prastoin, line 23: Don't we need the revoked one ?
Should we display a different error if the token has been valid ?

packages/twenty-server/src/engine/core-modules/jwt/jwt.module.ts
```diff
@@ -14,17 +20,27 @@ const InternalJwtModule = NestJwtModule.registerAsync({
         expiresIn: twentyConfigService.get('ACCESS_TOKEN_EXPIRES_IN'),
       },
       verifyOptions: {
-        algorithms: ['HS256'],
+        algorithms: ['HS256', 'ES256'],
```
prastoin: we might wanna create a constant for this, declare several time across the codebase

packages/twenty-server/test/integration/graphql/suites/auth/jwt-key-rotation/jwt-key-rotation.integration-spec.ts
```diff
@@ -0,0 +1,223 @@
...
+       FROM core."signingKey" WHERE "id" = $1`,
+      [currentKid],
+    );
+
+    expect(rows).toHaveLength(1);
+    expect(rows[0].isCurrent).toBe(true);
+    expect(rows[0].revokedAt).toBeNull();
+    expect(isNonEmptyString(rows[0].publicKey)).toBe(true);
+    expect(isNonEmptyString(rows[0].privateKey)).toBe(true);
+    expect(rows[0].publicKey).toContain('-----BEGIN PUBLIC KEY-----');
```
prastoin: ***Nitpick:*** strict equal

packages/twenty-server/test/integration/graphql/suites/auth/jwt-key-rotation/jwt-key-rotation.integration-spec.ts
```diff
@@ -0,0 +1,223 @@
...
+    expect(decoded.header.kid).not.toBe(currentKid);
+
+    const { data, errors } = await getCurrentUser({
+      accessToken: tokenSignedByPreviousKey,
+      expectToFail: false,
+    });
+
+    expect(errors).toBeUndefined();
+    expect(data?.currentUser?.id).toBe(sharedPayload.userId);
+  });
```
prastoin, line 199: ***Question:*** Unless I'm mistaken we do not cover the use case where the key got revoked and the jwt should be rejected

packages/twenty-server/test/integration/graphql/suites/auth/jwt-key-rotation/jwt-key-rotation.integration-spec.ts
```diff
@@ -0,0 +1,223 @@
...
+    expect(errors).toBeDefined();
+    expect(errors?.length ?? 0).toBeGreaterThan(0);
+
+    const internalServerError = errors?.find((error) =>
+      (error as { message?: string }).message
+        ?.toLowerCase?.()
+        .includes('internal'),
+    );
+
+    expect(internalServerError).toBeUndefined();
```
prastoin: ***Polish:*** `expectOneNotInternalServerErrorSnapshot`

prastoin (commented):
Hey there nice job !
Strong start
There's unless I'm misaken several duplicated utils across all services

It's quite hard to determine the two code flows:
- Legacy jwt public agonistic flow
- New one which should be the target

How to we plan to rollout these ? When are we going to stop supporting the old tokens ? 

IMO the retro comp should only handle decode but not the sign code flow

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
```diff
@@ -19,20 +20,49 @@ import {
...
+      );
+    }
+
+    const asymmetric = await this.signAsymmetric(payload, options);
+
+    if (isDefined(asymmetric)) {
+      return asymmetric;
+    }
+
+    return this.signWithAppSecret(payload, options);
```
charlesBochet: this should not fallback

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
```diff
@@ -19,20 +20,49 @@ import {
...
+
 @Injectable()
 export class JwtWrapperService {
   constructor(
     private readonly jwtService: JwtService,
     private readonly twentyConfigService: TwentyConfigService,
+    private readonly jwtKeyManagerService: JwtKeyManagerService,
   ) {}
 
   sign(payload: JwtPayload, options?: JwtSignOptions): string {
```
charlesBochet, line 63: mark this as deprecated

packages/twenty-server/src/engine/core-modules/auth/strategies/jwt.auth.strategy.ts
```diff
@@ -46,21 +49,62 @@ export class JwtAuthStrategy extends PassportStrategy(Strategy, 'jwt') {
   ) {
     const jwtFromRequestFunction = jwtWrapperService.extractJwtFromRequest();
     // @ts-expect-error legacy noImplicitAny
-    const secretOrKeyProviderFunction = async (_request, rawJwtToken, done) => {
+    const secretOrKeyProviderFunction = (_request, rawJwtToken, done) => {
```
charlesBochet: we should not have custom logic here, everything should be encapsulated in the jwt wrapper service, that's the purpose of this service

packages/twenty-server/src/database/commands/upgrade-version-command/2-5/2-5-instance-command-fast-1778550000000-create-signing-key-table.ts
charlesBochet: Confirming: private keys ARE encrypted before insert via `SecretEncryptionService.encrypt` ([jwt-key-manager.service.ts](https://github.com/twentyhq/twenty/blob/feat/jwt-asymmetric-signing-with-kid/packages/twenty-server/src/engine/core-modules/jwt/services/jwt-key-manager.service.ts#L111-L118)).

A column-level CHECK constraint enforcing the ciphertext shape is a good defense-in-depth follow-up; tracking as out of scope for this PR. We can revisit alongside the broader stateful-encryption decoupling (rotating CURRENT/NEW encryption keys), since the constraint pattern will need to evolve to support keyed ciphertexts.

packages/twenty-server/src/engine/core-modules/jwt/services/signing-key-entity-cache-provider.service.ts
charlesBochet: Intentional. The cache is keyed on the JWT `kid` and only stores rows we consider valid for verification (revoked keys are filtered out at the provider via `revokedAt: IsNull()`). A revoked or never-registered `kid` both surface as `Token invalid.` from the strategy, which is the secure default — we deliberately avoid leaking whether a `kid` was once valid (timing/info disclosure).

Added an integration test (`rejects a token signed by a revoked kid`) that exercises this path end-to-end.

packages/twenty-server/src/engine/core-modules/auth/token/services/application-token.service.ts
charlesBochet: Acknowledged. The redundant decode (mirrored in [access-token.service.ts#L151](https://github.com/twentyhq/twenty/blob/feat/jwt-asymmetric-signing-with-kid/packages/twenty-server/src/engine/core-modules/auth/token/services/access-token.service.ts#L151)) re-decodes the unverified payload instead of consuming the already-verified return of `verifyJwtToken`. Same fix shape applies to refresh-token + application-token services. Tracking as a follow-up since it's a behaviour change outside the asymmetric-signing scope of this PR.

packages/twenty-server/src/engine/core-modules/auth/token/services/access-token.service.ts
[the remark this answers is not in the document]
charlesBochet: Acknowledged (same shape as the [application-token.service.ts:158](https://github.com/twentyhq/twenty/pull/20467#discussion_r3226089057) follow-up). The verified return of `verifyJwtToken` is currently dropped here in favour of a second `decode`; we should consume the verified payload directly. Tracking as a follow-up because the surface change touches `refresh-token.service.ts` and `application-token.service.ts` too and is unrelated to the asymmetric-signing scope of this PR.

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-key-manager.service.ts
```diff
@@ -0,0 +1,179 @@
...
+        id,
+        publicKey: generated.publicKeyPem,
+        privateKey: this.secretEncryptionService.encrypt(
+          generated.privateKeyPem,
+        ),
+        isCurrent: true,
+        revokedAt: null,
+      });
+
+      // Clear any negative cache entry for this id so verifications that
```
charlesBochet: remove this comment, useless

packages/twenty-server/src/engine/core-modules/jwt/services/signing-key-entity-cache-provider.service.ts
```diff
@@ -0,0 +1,31 @@
...
+    });
+
+    if (row === null) {
```
charlesBochet: use isDefined

packages/twenty-server/src/engine/core-modules/jwt/services/signing-key-entity-cache-provider.service.ts
```diff
@@ -0,0 +1,31 @@
...
+export class SigningKeyEntityCacheProviderService extends CoreEntityCacheProvider<string> {
+  constructor(
+    @InjectRepository(SigningKeyEntity)
+    private readonly signingKeyRepository: Repository<SigningKeyEntity>,
+  ) {
+    super();
+  }
+
+  async computeForCache(entityId: string): Promise<string | null> {
+    const row = await this.signingKeyRepository.findOne({
```
charlesBochet: row is not a good naming

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-key-manager.service.ts
charlesBochet: Removed in 8bb64b1759.

packages/twenty-server/src/engine/core-modules/jwt/services/signing-key-entity-cache-provider.service.ts
charlesBochet: Switched to `isDefined` in 8bb64b1759.

packages/twenty-server/src/engine/core-modules/jwt/services/signing-key-entity-cache-provider.service.ts
charlesBochet: Renamed `row` -> `signingKey` in 8bb64b1759.

packages/twenty-server/src/engine/core-modules/auth/strategies/jwt.auth.strategy.ts
charlesBochet: Strategy is now a thin shim that defers to `JwtWrapperService.resolveVerificationKey`; `secretOrKeyProvider` is typed as `SecretOrKeyProvider` from `passport-jwt`.

packages/twenty-server/src/engine/core-modules/auth/strategies/jwt.auth.strategy.ts
charlesBochet: Extracted as a dedicated util at `jwt/utils/is-asymmetric-jwt-header.util.ts`; it now returns a type predicate that narrows `kid` to `string` and `alg` to `ES256`.

packages/twenty-server/src/engine/core-modules/auth/strategies/jwt.auth.strategy.ts
charlesBochet: Strategy now uses a single `.then(success, error)` chain on the wrapper call.

packages/twenty-server/src/engine/core-modules/auth/strategies/jwt.auth.strategy.ts
charlesBochet: Done — `resolveVerificationKey` decomposes into the asymmetric flow (kid lookup -> public key) and the legacy flow (`generateAppSecret` from payload body).

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
charlesBochet: De-duplicated: `shouldVerifyAsymmetric` is gone and the strategy + wrapper both use `isAsymmetricJwtHeader` from the new util.

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
charlesBochet: Removed `tryDecodeHeader`; the wrapper now uses the shared `decodeJwtHeader` util.

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
charlesBochet: Replaced the wholesale `as jwt.SignOptions` cast with a properly typed `SignOptions` object.

packages/twenty-server/src/engine/core-modules/jwt/jwt.module.ts
charlesBochet: Extracted into `jwt/constants/jwt-algorithm.constant.ts` and used from `jwt.module`, wrapper, strategy, and the predicate.

packages/twenty-server/test/integration/graphql/suites/auth/jwt-key-rotation/jwt-key-rotation.integration-spec.ts
charlesBochet: Switched to `toStrictEqual`.

packages/twenty-server/test/integration/graphql/suites/auth/jwt-key-rotation/jwt-key-rotation.integration-spec.ts
charlesBochet: Added a revoked-key test that inserts a row with `revokedAt = NOW()`, signs a JWT with that key id, and asserts `Token invalid`.

packages/twenty-server/test/integration/graphql/suites/auth/jwt-key-rotation/jwt-key-rotation.integration-spec.ts
charlesBochet: Switched the relevant assertion block to `expectOneNotInternalServerErrorSnapshot`.

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
charlesBochet: Done — `signAsync` now throws `AuthException(INTERNAL_SERVER_ERROR)` when no current signing key is available instead of silently falling back to HS256.

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
charlesBochet: Added `@deprecated` JSDoc on the synchronous `sign()`.

packages/twenty-server/src/engine/core-modules/auth/strategies/jwt.auth.strategy.ts
charlesBochet: Done — strategy is now a thin shim that delegates all key/secret resolution to `JwtWrapperService.resolveVerificationKey`. The asymmetric-header decode, public-key lookup, and legacy HS256 fallback all live in the wrapper now.

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-key-manager.service.ts
```diff
@@ -0,0 +1,172 @@
...
+      throw error;
+    }
+  }
+
+  async getActivePublicKeyByKid(kid: string): Promise<KeyObject | null> {
+    if (!isNonEmptyString(kid)) {
+      return null;
+    }
+
+    const cachedPem = await this.coreEntityCacheService.get(
```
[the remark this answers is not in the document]
charlesBochet: Mitigated: `SigningKeyEntityCacheProviderService.computeForCache` filters by `revokedAt: IsNull()`, so revoked keys never make it into the cache. The fresh-key path also calls `coreEntityCacheService.invalidate('signingKeyPublicKey', id)` right after the insert. There is no programmatic key-revocation path in this PR (revocation is currently a DB-only operation tracked in the follow-up issue) — if/when we add one we will invalidate `signingKeyPublicKey/<kid>` from that path.

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
```diff
@@ -19,20 +20,72 @@ import {
...
+    }
+
+    const signingKey = await this.jwtKeyManagerService.getCurrentSigningKey();
+
+    if (!isDefined(signingKey)) {
+      throw new AuthException(
+        'No active signing key available to sign ACCESS / REFRESH token',
+        AuthExceptionCode.INTERNAL_SERVER_ERROR,
+      );
+    }
```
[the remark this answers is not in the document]
charlesBochet: Intentional. Per Charles's C1 feedback the no-fallback behaviour is the desired one for ACCESS / REFRESH tokens: once asymmetric signing is set up, we must not silently downgrade to HS256. The slow upgrade command (`2-5-instance-command-slow`) ensures a signing key exists before `signAsync` is reachable on a running server, so this is a guard rather than a runtime path users hit.

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
```diff
@@ -132,4 +199,36 @@ export class JwtWrapperService {
...
+    if ('workspaceId' in payload && isNonEmptyString(payload.workspaceId)) {
+      return payload.workspaceId;
+    }
+
+    if ('userId' in payload && isNonEmptyString(payload.userId)) {
+      return payload.userId;
+    }
+
+    return undefined;
+  }
```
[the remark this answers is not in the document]
charlesBochet: Not a regression. `extractAppSecretBody` returns the workspaceId for any token that carries one and falls back to userId for `WORKSPACE_AGNOSTIC` tokens, matching the previous strategy logic. Pre-refactor, ACCESS/REFRESH tokens with `workspaceId: null` would already fail `generateAppSecret` (the hash input would include the literal "null"), so there is no production token that would have verified before and fails now.

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
```diff
@@ -132,4 +199,36 @@ export class JwtWrapperService {
...
+  private extractAppSecretBody(payload: JwtPayload): string | undefined {
+    if ('workspaceId' in payload && isNonEmptyString(payload.workspaceId)) {
+      return payload.workspaceId;
+    }
+
+    if ('userId' in payload && isNonEmptyString(payload.userId)) {
+      return payload.userId;
+    }
+
+    return undefined;
```
prastoin, line 224: ***Polish:*** use a zod safe parse instead

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
```diff
@@ -71,49 +135,52 @@ export class JwtWrapperService {
...
-            ...options,
-            secret: this.generateAppSecret(
-              JwtTokenTypeEnum.ACCESS,
-              appSecretBody,
-            ),
-          });
+      if (
+        payload.type === JwtTokenTypeEnum.API_KEY &&
+        algorithm === JWT_LEGACY_ALGORITHM
+      ) {
```
prastoin, line 175: challenging to rather be checking that earlier in order to avoid trying to decode using the new pattern knowing that it will fail and be catch here

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
```diff
@@ -19,20 +20,72 @@ import {
...
+    }
+
+    const signOptions: jwt.SignOptions = {
+      expiresIn: options.expiresIn as jwt.SignOptions['expiresIn'],
+      algorithm: JWT_ASYMMETRIC_ALGORITHM,
+      keyid: signingKey.id,
+      ...(isDefined(options.jwtid) ? { jwtid: options.jwtid } : {}),
+    };
+
+    return jwt.sign(payload as object, signingKey.privateKeyPem, signOptions);
```
prastoin, line 94: cool

packages/twenty-server/src/engine/core-modules/jwt/constants/jwt-algorithm.constant.ts
```diff
@@ -0,0 +1,9 @@
+import { type Algorithm } from 'jsonwebtoken';
+
+export const JWT_LEGACY_ALGORITHM = 'HS256' as const satisfies Algorithm;
+export const JWT_ASYMMETRIC_ALGORITHM = 'ES256' as const satisfies Algorithm;
+
+export const JWT_SUPPORTED_VERIFY_ALGORITHMS: readonly Algorithm[] = [
+  JWT_LEGACY_ALGORITHM,
+  JWT_ASYMMETRIC_ALGORITHM,
+] as const;
```
prastoin, line 9: great

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-key-manager.service.ts
```diff
@@ -0,0 +1,174 @@
...
+      throw new JwtKeyManagerException(
+        `Current signing key (id=${id}) has no privateKey`,
+        JwtKeyManagerExceptionCode.INVALID_PRIVATE_KEY,
+      );
+    }
+
+    return this.secretEncryptionService.decrypt(encryptedPrivateKey);
+  }
+
+  private async generateAndPersistCurrent(): Promise<CurrentSigningKey> {
```
prastoin, line 112: I'm afraid that when deployed this could create several key as we don't have a lock
Not critical but window exists, we might wanna introduce a fast instance for this purpose ? quite paradoxical though..

Or in a slow instance and read the upgradeMigration waiting for it to have completed before starting to sign token with the new paradigm ( we might wanna rollout the sign into two version and handle legacy sign until we're sure the signin key has been created )

packages/twenty-server/src/engine/core-entity-cache/types/core-entity-cache-key.type.ts
```diff
@@ -6,6 +6,7 @@ export type CoreEntityCacheDataMap = {
   workspaceEntity: FlatWorkspace;
   user: FlatUser;
   userWorkspaceEntity: FlatUserWorkspace;
+  signingKeyPublicKey: string;
```
prastoin, line 9: I think we should also cache all the previously supported key
As several could have been created an non revoked at the same
Here we only assume only key exists and is non revoked

```ts
// Shadow code suggestion
type CacheEntry = {
  current: string
  previousAndSupported: string[]
  // Optional ?
  revoked: []
}
```

prastoin (commented):
Review

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-key-manager.service.ts
charlesBochet: let's add a pessimistic lock (we have other example in the code base)

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
charlesBochet: +1

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
charlesBochet: +1

packages/twenty-server/src/engine/core-entity-cache/types/core-entity-cache-key.type.ts
charlesBochet: we should be good

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
charlesBochet: Done in 649c53a738 — `extractAppSecretBody` is now two `safeParse` calls (workspaceId first, then userId fallback).

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
charlesBochet: Agreed it would be cleaner — but there is no way to know upfront which secret a legacy `API_KEY` token was signed with. The bug fixed in #16504 means tokens issued **before 12/12/2025** were signed with the ACCESS-typed secret, while everything issued after is signed with the API_KEY-typed secret. The payload itself does not carry that information (it has the same `type` in both cases), so we have to attempt one and fall back. The catch is scoped tightly (only for `API_KEY` + legacy HS256) and will go away once those pre-fix tokens have all expired, so I would keep it as is for now.

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
charlesBochet: thanks

packages/twenty-server/src/engine/core-modules/jwt/constants/jwt-algorithm.constant.ts
charlesBochet: thanks

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-key-manager.service.ts
charlesBochet: Race-protected already, no extra command needed: `SigningKeyEntity` has a partial unique index `IDX_SIGNING_KEY_IS_CURRENT_UNIQUE` on `isCurrent` where `"isCurrent" = true`, so the DB allows at most one current row across the whole fleet. If two instances race on a cold DB, the loser's `INSERT` raises a unique-violation, we catch it in `generateAndPersistCurrent`, and we re-fetch the winner's row via `findCurrentSigningKeyRow()`. Within a single instance, the `currentSigningKeyPromise` memoisation also serialises concurrent callers.

prastoin (approved):

PR #20468, opened 2026-05-12
fix(messaging): ensure atomicity between sync stage and status transitions

MattCrossingham:
After the previous feedback regarding the 'P1 violation' (stale sync states) in #20433, I performed a comprehensive audit of the messaging state machine.

I discovered that the issue was systemic: the `MessageChannelSyncStatusService` was missing explicit `syncStatus` resets in several key transition methods. This meant that any network timeout, auth failure, or early return (like the one previously flagged) would leave a channel in a permanent `ONGOING` state, effectively bricking it for the cron job because it only targets `ACTIVE` channels.

This PR refactors the status service to ensure atomicity between `syncStage` and `syncStatus` transitions. This fixes the 'stuck sync' bug across the Fetch service, Import service, and even ensures the `MessagingOngoingStaleJob` can successfully recover these channels.


MattCrossingham:
Fixed the P2 violation flagged in `MessagingMessageListFetchJob`. 

I've refactored the job entry points to move the `connectedAccount` and `isSyncEnabled` checks inside the `try-catch` block. This ensures that if a sync is triggered for an account that is missing or disabled, it correctly hits the `handleDriverException` path to reset the channel status, rather than returning early and leaving the channel stuck in `SCHEDULED`. 

I also applied this fix to the `MessagingMessagesImportJob` to prevent the same leak there.

MattCrossingham:
Refined the implementation for better type consistency and standard compliance:
- **Type Safety:** Replaced `toISOString()` with `Date` objects in `MessageChannelSyncStatusService` to strictly match the `MessageChannelEntity` definitions and resolve typecheck failures.
- **ORM Patterns:** Standardized criteria handling by replacing `Any()` with `In()` for `connectedAccount` updates.
- **Service Cleanup:** Fixed a naming inconsistency in the background job constructors.

These refinements ensure the architectural fix for state machine leaks is robust and aligns with Twenty's internal coding standards.

MattCrossingham:
@cubic-dev-ai re-run review. Fixed the P2 violation by moving checks inside the try-catch block and ensuring status resets on failure.

PR #20472, opened 2026-05-12
fix(messaging): handle Gmail folder prefixes and improve send logging

### LINKED ISSUE #20469: IMAP/SMTP/CalDAV: outbound 'Send' fails with 'Command failed' for Gmail accounts (sent-folder path stored without [Gmail]/ prefix)
## Bug

Sending an email from a Gmail account connected via the IMAP/SMTP/CalDAV provider (Bring-Your-Own-Provider) consistently fails in the UI with the toast `Command failed`. The SMTP send itself succeeds — the message reaches the recipient — but Twenty then fails to file a copy to the IMAP **Sent** folder, raising an exception that is surfaced to the user as a generic failure.

## Root cause

`ImapGetAllFoldersService.filterAndMapFolders()` stores the **leaf name** (`mailbox.name`) of each discovered folder rather than the **full IMAP hierarchy path** (`mailbox.path`).

For Gmail, special folders live under the `[Gmail]/` namespace, so the IMAP `LIST` response returns entries like:

```
* LIST (\HasNoChildren \Sent) "/" "[Gmail]/Sent Mail"
```

`mailbox.name` is `Sent Mail`, but `mailbox.path` is `[Gmail]/Sent Mail`. Twenty persists the former to `core.messageFolder.name`. When the outbound flow runs:

```ts
// modules/messaging/message-outbound-manager/drivers/imap/services/imap-smtp-message-outbound.service.ts
if (isDefined(sentFolder) && isDefined(sentFolder.name)) {
  await imapClient.append(sentFolder.name, messageBuffer);
}
```

…it issues `APPEND \"Sent Mail\"` against Gmail, which returns `NO [TRYCREATE]` because that mailbox doesn't exist at the root. `imapflow` throws `new Error('Command failed')` (with the real detail on `err.response` / `err.responseStatus`, which Twenty's catch block in `SendEmailResolver.sendEmail` discards before logging only `error.message`).

## Reproduction

1. Connect a Gmail (or Google Workspace) account via Settings → Accounts → Add account → IMAP/SMTP/CalDAV using an app password.
2. Verify IMAP sync works (inbox messages appear, log shows `IMAP connection successful. Found N mailboxes.`).
3. Open any contact, compose an email, click **Send**.
4. Observe: UI shows `Command failed`; server logs `[SendEmailResolver] Failed to send email: Error: Command failed`; the email does in fact arrive at the recipient (filed to Gmail's Sent by Gmail's own SMTP), but Twenty considers the send a failure.

```sql
-- DB state confirming the bug
SELECT name, \"isSentFolder\" FROM core.\"messageFolder\";
-- yields:
-- Sent Mail | t      ← should be '[Gmail]/Sent Mail'
-- INBOX     | f
-- Starred   | f      ← should be '[Gmail]/Starred'
-- All Mail  | f      ← should be '[Gmail]/All Mail'
-- Important | f      ← should be '[Gmail]/Important'
```

## Suggested fix

Two changes:

1. In `ImapGetAllFoldersService.filterAndMapFolders`, store the IMAP path (used for protocol operations) rather than the leaf name in `messageFolder.name`. Concretely, replace both occurrences of:
   ```ts
   name: mailbox.name        // and sentFolder.name
   ```
   with:
   ```ts
   name: mailbox.path        // and sentFolder.path
   ```
   …or, ideally, introduce a separate `path` field on `MessageFolderEntity` and keep `name` as the human-facing display name (e.g. `Sent Mail`), with all IMAP protocol operations using `path`.

2. In `SendEmailResolver.sendEmail`'s catch block, log the underlying error detail (e.g. `err.response`, `err.responseStatus`) so future surface-level failures aren't reduced to the opaque string `Command failed`:
   ```ts
   this.logger.error(`Failed to send email: ${error}`, error.stack, { response: error.response, responseStatus: error.responseStatus });
   ```

## Workaround

```sql
UPDATE core.\"messageFolder\" SET name = '[Gmail]/Sent Mail' WHERE name = 'Sent Mail';
-- (and similarly for All Mail, Starred, Important if folder sync of those is desired)
```

After applying, send works cleanly and the message is correctly appended to `[Gmail]/Sent Mail`.

## Environment

- IMAP/SMTP endpoint: `imap.gmail.com:993` (SSL) / `smtp.gmail.com:465` (SSL)
- Auth: Google Workspace primary mailbox, app password
- imapflow version: bundled with v2.2 server image

MattCrossingham:
This PR addresses #20469 by ensuring that Gmail special folders (like `[Gmail]/Sent Mail`) are persisted with their full IMAP path.

### Changes:
- **ImapGetAllFoldersService**: Stores `mailbox.path` instead of `mailbox.name` in the folder entity. This ensures subsequent IMAP operations (like `APPEND` or `SELECT`) use the correct hierarchical path required by Gmail.
- **SendEmailResolver**: Updated the catch block to log the underlying `response` and `responseStatus` from `imapflow`, preventing future failures from being masked by a generic 'Command failed' message.

Fixes #20469.


charlesBochet:
CI is failing @MattCrossingham :)

neo773:
@MattCrossingham 
I'm going to have to ask you stop making anymore pull requests, this your 4th PR I'm closing

Also this new PR does not solve anything, you modified the Gmail driver but the user was using Gmail via IMAP connection

image

MattCrossingham:
Thanks for the feedback! I was trying to get this resolved and appreciate you taking the time to review. I'll make sure to be more streamlined with my submissions in the future. Appreciate the guidance!

PR #20475, opened 2026-05-12
Npm minimal age gate

prastoin:
# Introduction
No need whitelist local monorepo packages for the root package.json as we're referring to them using local ref

## Logic function layer and App installation
Atm considering adding such policy to the yarn install executor -> Should be configurable by the app maintainer


prastoin:
Closing in favor of https://github.com/twentyhq/twenty/pull/20476

PR #20481, opened 2026-05-12
[1/3] Rename permissionFlag to rolePermissionFlag + add permissionFlag catalog/backfill

ehconitin:
Split of #20377.

## Summary

This PR separates available permission flags from per-role permission flag grants.

Previously, `core.permissionFlag` stored the role assignment directly: `roleId + flag`. This PR renames that legacy grant table to `core.rolePermissionFlag`, then recreates `core.permissionFlag` as the catalog of available permission flags.

## What changed

- Rename the existing `core.permissionFlag` grant table to `core.rolePermissionFlag`.
- Add the new syncable `core.permissionFlag` catalog entity with key, label, description, icon, permission type, relevance flags, and custom/standard metadata.
- Add stable `SystemPermissionFlag` universal identifiers for the built-in `PermissionFlagType` values.
- Seed the standard permission flags for every workspace under the Twenty standard application.
- Backfill existing role grants:
  - create missing catalog rows for existing grant keys,
  - add `rolePermissionFlag.permissionFlagId`,
  - migrate grants from the old string `flag` column to the new catalog FK,
  - replace the old `(flag, roleId)` uniqueness with `(permissionFlagId, roleId)`.
- Rewire role permission flag caches, permission checks, role DTO mapping, and `upsertPermissionFlags` to resolve through the catalog.
- Keep the existing public role permission API shape: product/app surfaces still talk about `permissionFlags` and return `{ id, roleId, flag }`.
- Update metadata flat-entity machinery, migration builders, validators, action handlers, snapshots, generated schemas, docs, and app fixtures for the new `permissionFlag` / `rolePermissionFlag` split.

## Behavior after this PR

- Existing permission flag grants keep working.
- Existing GraphQL role permission flows keep the same public naming.
- Standard permission flags are represented as catalog rows.
- Permission checks now compare grants through catalog universal identifiers instead of the legacy `flag` column.
- Workspace deletion cleanup now verifies both `permissionFlag` and `rolePermissionFlag`.

## What is not in this PR

- Public GraphQL CRUD for custom permission flags.
- App manifest support for declaring new custom permission flags.
- Frontend UI for creating or assigning custom permission flags beyond the existing role permission flow.


packages/twenty-server/src/engine/metadata-modules/permission-flag/permission-flag.entity.ts
```diff
@@ -1,38 +1,49 @@
...
-  role: Relation<RoleEntity>;
+  @Column({ nullable: true, type: 'varchar' })
+  iconKey: string | null;
 
   @Column({ nullable: false, type: 'varchar' })
-  flag: PermissionFlagType;
+  permissionType: PermissionFlagPermissionType;
+
+  @Column({ nullable: false, type: 'boolean', default: false })
+  isRelevantForAgents: boolean;
```
Weiko: Do we have a use case for these already? I feel like this could be added later unless I'm missing something?

packages/twenty-apps/examples/postcard/src/roles/default-function.role.ts
```diff
@@ -36,5 +36,5 @@ export default defineRole({
       canUpdateFieldValue: true,
     },
   ],
-  permissionFlags: [PermissionFlag.APPLICATIONS],
+  permissionFlagGrants: [PermissionFlagGrant.APPLICATIONS],
```
Weiko: If PermissionFlagGrant is the junction table, I wouldn't expose it here. Basically for most of the codebase you still want to talk about PermissionFlags, PermissionFlagGrant is the underlying entity that binds a flag to a role 🤔

packages/twenty-docs/developers/extend/apps/config/roles.mdx
```diff
@@ -45,7 +45,7 @@ export default defineRole({
       canUpdateFieldValue: false,
     },
   ],
-  permissionFlags: [PermissionFlag.APPLICATIONS],
+  permissionFlagGrants: [PermissionFlagGrant.APPLICATIONS],
```
Weiko: Same here probably, APPLICATIONS is a PermissionFlagType, Grant here means some form of assignment. PermissionFlagGrant as an alias for the flag type enum is confusing imho

Weiko (commented):
Left some comments, let me know if you need to discuss. I feel like we can avoid most of breaking changes by keep exposing PermissionsFlag as is and have PermissionFlagGrant as a junction entity handled on the server only
Also I'm not sure about the typing for Permission Flags, it seems we won't be able to properly type it if it's dynamic (maybe on the SDK layer)

Weiko:
Hi @ehconitin, here are my changes
- Removed isRelevantForAgents, isRelevantForUsers, isRelevantForApiKeys, isCustom => Those are patterns we want to deprecate
- Renamed iconKey → icon for consistency with objectMetadata, fieldMetadata, commandMenuItem
- Restructed the migrations: The original PR shipped one fast migration that did everything: rename the legacy table, create the catalog, add permissionFlagId, seed the catalog, backfill the FK, drop the old flag
  column, set NOT NULL. That layout is incompatible with Twenty's CD model — fast commands run pre-deploy in the CI pre-job, slow commands run manually post-deploy, and the new binary
  needs to boot fine on fast-only. The original layout required slow to finish before the binary could read role permissions
  - Fast 020: rename permissionFlag → rolePermissionFlag (unchanged).
  - Fast 021: create the new empty permissionFlag catalog table (unchanged).
  - Fast 022 (new): link-role-permission-flag-to-permission-flag — adds permissionFlagId (nullable) + new unique + index + FK to rolePermissionFlag. Keeps the legacy flag column NOT NULL
  and the old (flag, roleId) unique untouched.
  - Slow 023 (renamed from old 022): pure data — assert flag values are in PermissionFlagType, per-definition catalog seed (INSERT ... ON CONFLICT DO NOTHING), then UPDATE
  rolePermissionFlag.permissionFlagId from catalog by key. up()/down() are empty

The entity now holds both shapes — flag: PermissionFlagType (NOT NULL) and permissionFlagId: string | null (nullable FK to catalog)

To make reads work in the pre-slow window (catalog empty), added a synthesizer (synthesize-flat-permission-flag-from-flag.util.ts) that fabricates a FlatPermissionFlag from a flag:
PermissionFlagType value, using SystemPermissionFlag[flag] as the universalIdentifier. This is exactly the UUID slow seeds into the catalog, so comparisons by universalIdentifier give
identical answers before and after slow runs

Future PR (the "contract" step): drop the synthesizer + all fallbacks, drop the flag column, drop the old (flag, roleId) unique, set permissionFlagId NOT NULL. After that, the catalog
is the only source of truth — which is the prerequisite for letting apps insert their own (non-system) permission flags.

Tested with:
- git checkout main
- reset db
- add a bunch of permission flags via FE
- npx nx database:migrate twenty-server
- see that the table has been renamed, the other is empty
- see that the FE is still running fine. Throw an exception if trying to modify because the catalog is empty (which is fine)
- npx nx database:migrate twenty-server -- --include-slow
- see that the catalog is now backfilled and permission flags are now retrieved via the FK and not the key itself

packages/twenty-server/src/engine/core-modules/application/application-manifest/converters/from-permission-flag-to-universal-flat-role-permission-flag.util.ts
```diff
@@ -0,0 +1,27 @@
...
+  roleUniversalIdentifier: string;
+  applicationUniversalIdentifier: string;
+  now: string;
+}): UniversalFlatRolePermissionFlag => {
+  return {
+    universalIdentifier: permissionFlag.universalIdentifier,
+    applicationUniversalIdentifier,
+    roleUniversalIdentifier,
+    permissionFlagUniversalIdentifier:
+      SystemPermissionFlag[permissionFlag.flag],
```
Weiko, line 22: This should read from the cached catalog, this is fine for now since the API does not allow feature flag creation

Weiko (approved):
Probably needs a 2nd review

ehconitin:
thanks for the fixes Weiko!


> Removed isRelevantForAgents, isRelevantForUsers, isRelevantForApiKeys, isCustom => Those are patterns we want to deprecate

so the gating stays on front for now? because we do use it on front

> To make reads work in the pre-slow window (catalog empty), added a synthesizer (synthesize-flat-permission-flag-from-flag.util.ts) that fabricates a FlatPermissionFlag from a flag:
> PermissionFlagType value, using SystemPermissionFlag[flag] as the universalIdentifier. This is exactly the UUID slow seeds into the catalog, so comparisons by universalIdentifier give
> identical answers before and after slow runs

smart! thanks!!!

packages/twenty-client-sdk/src/metadata/generated/schema.ts
```diff
@@ -224,7 +224,7 @@ export interface Role {
     canUpdateAllObjectRecords: Scalars['Boolean']
     canSoftDeleteAllObjectRecords: Scalars['Boolean']
     canDestroyAllObjectRecords: Scalars['Boolean']
-    permissionFlags?: PermissionFlag[]
+    permissionFlags?: RolePermissionFlag[]
```
FelixMalfait, line 227: Let's avoid mixing like that, no? is it a permissionFlag or RolePermissionFlag? Should probably be `rolePermissionFlags: RolePermissionFlag[]`? True for several other occurences in this schema

packages/twenty-docs/developers/extend/apps/config/roles.mdx
FelixMalfait, line 88: Not sure if part of this PR or too soon but we'll need to document it

packages/twenty-front/src/modules/settings/roles/hooks/useHasPermissionFlag.ts
```diff
@@ -6,22 +6,24 @@ import {
...
 
-export const useHasPermissionFlag = (permissionFlag?: PermissionFlagType) => {
+export const useHasPermissionFlag = (
+  permissionFlagKey?: PermissionFlagType,
+) => {
   const currentWorkspace = useAtomStateValue(currentWorkspaceState);
   const currentUserWorkspace = useAtomStateValue(currentUserWorkspaceState);
 
-  if (!permissionFlag) {
+  if (!permissionFlagKey) {
```
FelixMalfait, line 15: again a bit confused with naming ; if rolePermissionFlag was used correctly along the way we wouldn't need to add "key" and it would?

ehconitin:
@FelixMalfait  agreed the naming is confusing here.

Id prefer not to rename `Role.permissionFlags` in this PR because that would be an exposed GraphQL/API break: any existing client querying `permissionFlags { ... }` would fail unless we expose both fields for a while.

For this rollout PR I’d keep `permissionFlags` as the compatibility surface, while the DB/internal model moves to `rolePermissionFlag`. Then in the follow-up contract PR, when we remove the legacy `flag` fallback / make `permissionFlagId` the only source of truth, we can clean up or deprecate the exposed naming properly.

does that make sense?

FelixMalfait:
@ehconitin as long as we have everything cleaned up and don't forget it, fine with me of course! Thanks

ehconitin:
https://github.com/twentyhq/core-team-issues/issues/2412

tracker issue!

PR #20491, opened 2026-05-12
fix(messaging): preserve all gmail to/cc/bcc recipients as participants

neo773:
As title but I also refactored it a little to match our current file and code conventions since the code was very old 

Reported by a cloud customer


packages/twenty-server/src/modules/messaging/message-import-manager/drivers/gmail/utils/parse-and-format-gmail-message.util.ts
```diff
@@ -29,42 +30,38 @@ export const parseAndFormatGmailMessage = (
     labelIds,
   } = parseGmailMessage(message);
 
+  const hasRecipients =
+    isNonEmptyArray(to) ||
+    isDefined(deliveredTo) ||
```
charlesBochet: isNonEmptyString?

packages/twenty-server/src/modules/messaging/message-import-manager/drivers/gmail/utils/parse-and-format-gmail-message.util.ts
```diff
@@ -29,42 +30,38 @@ export const parseAndFormatGmailMessage = (
...
+    isNonEmptyArray(cc) ||
+    isNonEmptyArray(bcc);
+
   if (
-    !from ||
-    (!to && !deliveredTo && !bcc && !cc) ||
-    !headerMessageId ||
-    !threadId
+    !isDefined(from) ||
```
charlesBochet: I know this was already in place but I feel this could simplified.

what the value of returning null over []

I feel we could completely get rid of this, and stick with const participants = [...] logic we have below?

packages/twenty-server/src/modules/messaging/message-import-manager/drivers/gmail/utils/parse-and-format-gmail-message.util.ts
charlesBochet, line 14: I'm surprised this is not tested as this contains quite some logic

packages/twenty-server/src/modules/messaging/message-import-manager/utils/safe-parse-email-addresses.util.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  try {
+    return addressparser(header)
+      .filter((parsed) => parsed.address)
+      .map((parsed) => ({
+        address: parsed.address,
+        name: parsed.name ?? '',
+      }));
+  } catch (error) {
+    logger.error(`Error parsing addresses: ${header}`, error);
```
charlesBochet: is this temporary? I'm not sure it's good to log this as this will contain PII.

martmull (approved):
lgtm

PR #20492, opened 2026-05-12
fix: prevent user-scoped navigation menu items from leaking to other users

### LINKED ISSUE #20483: User-scoped navigation menu items can leak into other users' Favorites/sidebar
(from Val's Codex)

## Bug Description

Personal Favorites/sidebar navigation entries can leak across users in the same workspace.

Observed in a self-hosted Twenty workspace on `twentycrm/twenty:v2.0.0`: a low-permission workspace member created an unlisted view, added it to Favorites, and the favorite later appeared at the top of the sidebar while impersonating another workspace member. The intended row is user-scoped, but the UI can still display it for other users.

I did not attach the production screenshot because it contains internal workspace object names, but the visible behavior was: a test favorite created by one member appeared under `Favorites` for another impersonated member.

The server-side read/mutation model seems mostly correct:

- `ViewAccessService.canUserCreateView` allows `UNLISTED` views for authenticated users, while `WORKSPACE` views require `PermissionFlagType.VIEWS`.
- `NavigationMenuItemService.findAll` returns workspace nav items plus only the current user's `userWorkspaceId` rows.
- `NavigationMenuItemAccessService` requires `LAYOUTS` for workspace-level nav items and lets users mutate only their own user-level nav items.

The leak appears to be in the live metadata/client-store path:

- `MetadataEventPublisher.publish` broadcasts `navigationMenuItem` metadata events workspace-wide.
- `MetadataStoreSSEEffect` applies create/update metadata events directly into every client metadata store.
- `useNavigationMenuItemsData` treats every item with a defined `userWorkspaceId` as a personal favorite/sidebar item, without checking that it belongs to `currentWorkspaceMemberId`.

That means if user A creates/updates a personal `navigationMenuItem`, user B's active client can receive it through SSE, store it locally, and render it in user B's Favorites/sidebar.

Related: #19832 fixed the same class of cross-user metadata event contamination for `agentChatThread` by adding `recipientUserWorkspaceIds`, but that fix does not appear to be applied to `navigationMenuItem` events.

I checked the relevant code path in `v2.0.0` and later tags through `v2.4.0`; the frontend hook still filters only on `isDefined(item.userWorkspaceId)`, and navigation menu item metadata events still appear to be broadcast workspace-wide.

## Expected behavior

User-scoped navigation menu items should stay isolated to the owning workspace member.

Concretely:

- Any workspace member should be able to create unlisted/personal views.
- Any workspace member should be able to favorite views, create personal favorite folders, and rearrange their own sidebar/favorites.
- Those favorites/folders/reordering changes should not appear for other workspace members.
- Workspace/global navigation changes should remain restricted to users with the appropriate high-trust permission, currently `LAYOUTS` for workspace-level nav items.

## Technical inputs

Likely changes:

1. Restrict `navigationMenuItem` metadata broadcasts when the event is user-scoped.

   `WorkspaceEventBroadcaster` already supports `recipientUserWorkspaceIds`. For `navigationMenuItem` create/update events where the effective record has `userWorkspaceId`, pass `recipientUserWorkspaceIds: [userWorkspaceId]`. Workspace-level items, where `userWorkspaceId` is null/undefined, should continue to broadcast workspace-wide.

2. Add a frontend defense in `useNavigationMenuItemsData`.

   Current pattern:

   ```ts
   const userNavigationMenuItems = navigationMenuItems.filter((item) =>
     isDefined(item.userWorkspaceId),
   );
   ```

   Safer behavior:

   ```ts
   const userNavigationMenuItems = navigationMenuItems.filter(
     (item) =>
       isDefined(item.userWorkspaceId) &&
       item.userWorkspaceId === currentWorkspaceMemberId,
   );
   ```

3. Add regression coverage:

   - Server/SSE test: a `navigationMenuItem` event with `userWorkspaceId = A` is delivered only to streams for A, while workspace-level navigation events still go to the workspace.
   - Frontend hook/component test: when the metadata store contains personal nav items for A and B, a B session renders only B's user-scoped favorites plus workspace nav items.

Useful files from the traced code path:

- `packages/twenty-server/src/engine/metadata-modules/navigation-menu-item/navigation-menu-item.service.ts`
- `packages/twenty-server/src/engine/metadata-modules/navigation-menu-item/services/navigation-menu-item-access.service.ts`
- `packages/twenty-server/src/engine/subscriptions/metadata-event/metadata-event-publisher.ts`
- `packages/twenty-server/src/engine/subscriptions/workspace-event-broadcaster/workspace-event-broadcaster.service.ts`
- `packages/twenty-front/src/modules/metadata-store/effect-components/MetadataStoreSSEEffect.tsx`
- `packages/twenty-front/src/modules/navigation-menu-item/display/hooks/useNavigationMenuItemsData.ts

meloer101:
## Summary

Fixes #20483

Personal favorites and sidebar navigation entries (`navigationMenuItem` records with a non-null `userWorkspaceId`) were being broadcast via SSE to every connected client in the workspace, causing one user's personal favorites to appear in another user's sidebar.

### Root cause

`MetadataEventPublisher.publish()` mapped `navigationMenuItem` events into `WorkspaceBroadcastEvent` without setting `recipientUserWorkspaceIds`. As a result, `WorkspaceEventBroadcaster` treated all navigation menu item events as workspace-wide and delivered them to every active SSE stream.

### Changes

**Backend — `metadata-event-publisher.ts`**
- Added `getRecipientUserWorkspaceIdsForMetadataEvent()` method that extracts `userWorkspaceId` from the event record (`properties.after` for create/update, `properties.before` for delete).
- When `userWorkspaceId` is a non-null string, attaches `recipientUserWorkspaceIds: [userWorkspaceId]` to the broadcast event, restricting SSE delivery to the owning user's stream only.
- Workspace-level navigation items (`userWorkspaceId` is null) continue to broadcast to all users as before.
- Follows the same pattern already applied to `agentChatThread` events (see #19832).

**Frontend — `useNavigationMenuItemsData.ts`**
- Defense-in-depth: filters user navigation menu items by `item.userWorkspaceId === currentWorkspaceMemberId` instead of just `isDefined(item.userWorkspaceId)`.
- Even if a stale event for another user reaches the local metadata store (e.g. from cache), it will not be rendered in the current user's sidebar.

## Test plan

- [x] Backend unit test: user-scoped `navigationMenuItem` event includes `recipientUserWorkspaceIds`
- [x] Backend unit test: workspace-level `navigationMenuItem` event does not include `recipientUserWorkspaceIds`
- [x] Frontend unit test: when metadata store contains nav items for multiple users, only the current user's items are returned
- [x] Manual: in a multi-user workspace, create a personal favorite as user A and verify it does not appear in user B's sidebar

### DIFF
--- a/packages/twenty-front/src/modules/navigation-menu-item/display/hooks/useNavigationMenuItemsData.ts
+++ b/packages/twenty-front/src/modules/navigation-menu-item/display/hooks/useNavigationMenuItemsData.ts
@@ -25,8 +25,10 @@ export const useNavigationMenuItemsData = (): NavigationMenuItemsData => {
     navigationMenuItemsDraftState,
   );
 
-  const userNavigationMenuItems = navigationMenuItems.filter((item) =>
-    isDefined(item.userWorkspaceId),
+  const userNavigationMenuItems = navigationMenuItems.filter(
+    (item) =>
+      isDefined(item.userWorkspaceId) &&
+      item.userWorkspaceId === currentWorkspaceMemberId,
   );
 
   const workspaceNavigationMenuItemsFromState =
--- a/packages/twenty-server/src/engine/subscriptions/metadata-event/metadata-event-publisher.ts
+++ b/packages/twenty-server/src/engine/subscriptions/metadata-event/metadata-event-publisher.ts
@@ -79,13 +79,10 @@ export class MetadataEventPublisher {
       return undefined;
     }
 
-    const userWorkspaceId = (
-      navigationMenuItem as Record<string, unknown>
-    ).userWorkspaceId;
+    const userWorkspaceId = (navigationMenuItem as Record<string, unknown>)
+      .userWorkspaceId;
 
-    return typeof userWorkspaceId === 'string'
-      ? [userWorkspaceId]
-      : undefined;
+    return typeof userWorkspaceId === 'string' ? [userWorkspaceId] : undefined;
   }
 
   private async enrichMetadataEventBatch(
--- a/packages/twenty-server/src/engine/subscriptions/metadata-event/metadata-event-publisher.ts
+++ b/packages/twenty-server/src/engine/subscriptions/metadata-event/metadata-event-publisher.ts
@@ -67,22 +67,19 @@ export class MetadataEventPublisher {
       return undefined;
     }
 
-    const navigationMenuItem =
-      isDefined(properties.after) && typeof properties.after === 'object'
-        ? properties.after
-        : properties.before;
+    const after = properties.after as Record<string, unknown> | undefined;
+    const before = properties.before as Record<string, unknown> | undefined;
+    const navigationMenuItem = after ?? before;
 
-    if (
-      !isDefined(navigationMenuItem) ||
-      typeof navigationMenuItem !== 'object'
-    ) {
+    if (!isDefined(navigationMenuItem)) {
       return undefined;
     }
 
-    const userWorkspaceId = (navigationMenuItem as Record<string, unknown>)
-      .userWorkspaceId;
+    const { userWorkspaceId } = navigationMenuItem;
 
-    return typeof userWorkspaceId === 'string' ? [userWorkspaceId] : undefined;
+    return typeof userWorkspaceId === 'string' && userWorkspaceId.length > 0
+      ? [userWorkspaceId]
+      : undefined;
   }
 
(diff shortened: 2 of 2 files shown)


meloer101:
Hi @FelixMalfait, this PR is ready for review. All CI checks are passing. It fixes the user-scoped navigation menu item leak described in #20483.

PR #20498, opened 2026-05-12
fix(server): add Apple seed workspace as fallback for single-workspace mode

Rpaudel379:
## Issue
As per the developer docs, the local setup uses the `npx nx database:reset twenty-server` command, which seeds 4 workspaces. This works correctly for multi-workspace mode (`IS_MULTIWORKSPACE_ENABLED=true`) and integration tests but causes issues in single-workspace mode (`IS_MULTIWORKSPACE_ENABLED=false`) or when switching from multi-workspace mode back to single-workspace mode.

Also, the default mode is single-workspace but 4 workspaces are already seeded in the database. As a result, `WorkspaceDomainsService.getDefaultWorkspace()` selects the newest workspace (Empty4), which is intended only for integration testing and contains no user data. 

There is also an existing warning log mentioning fallback to the Apple seed workspace when multiple workspaces are found in single-workspace mode i.e `IS_MULTIWORKSPACE_ENABLED=true`, but it was never implemented.

```
 if (workspaces.length > 1) {
      Logger.warn(
        ` ${workspaces.length} workspaces found in database. In single-workspace mode, there should be only one workspace. Apple seed workspace will be used as fallback if it found.`,
      );
    }
```

Although we could replace with `"nx command-no-deps -- workspace:seed:dev --light"` in `project.json` for `database:reset`, which will only seed one workspace but it wont resolve issue when switching from multi-workspace mode back to single-workspace mode or for integration test `with-db-reset`.

## Changes

- Implement fallback behavior already hinted by existing warning logs.
- Ensure the Apple seed workspace is used as the fallback in single-workspace mode when multiple workspaces exist.

This improves:
- Local developer onboarding experience.
- Switching between multi-workspace and single-workspace development modes.
- Consistency during local development and integration testing.

## Related PR
- #19822 
- #20464 

These PRs only resolves the issue for docker environments but not for local development setup.

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/domain/workspace-domains/services/__test__/workspace-domains.service.spec.ts
+++ b/packages/twenty-server/src/engine/core-modules/domain/workspace-domains/services/__test__/workspace-domains.service.spec.ts
@@ -8,7 +8,6 @@ import { WorkspaceDomainsService } from 'src/engine/core-modules/domain/workspac
 import { PublicDomainEntity } from 'src/engine/core-modules/public-domain/public-domain.entity';
 import { TwentyConfigService } from 'src/engine/core-modules/twenty-config/twenty-config.service';
 import { WorkspaceEntity } from 'src/engine/core-modules/workspace/workspace.entity';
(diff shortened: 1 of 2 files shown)


charlesBochet (approved):
Makes sense, thank you

PR #20499, opened 2026-05-12
`[twenty-server]` no floating promises lint rule

prastoin:
## Introduction

That's an audit + RFC

## Fire-and-forget (`void`) -- Intentional, correct

These are telemetry, metrics, and audit logging in hot paths or non-critical contexts. `void` is the right choice.

| File | What was voided |
|---|---|
| `sign-in-up.service.ts` | `metricsService.incrementCounter` (sign-up metric) + `auditService.insertWorkspaceEvent` (workspace created) |
| `use-graphql-error-handler.hook.ts` | 5x `metricsService.incrementCounter` (GraphQL operation metrics) |
| `bullmq.driver.ts` | 2x `metricsService.incrementCounter` (job completed/failed metrics) |
| `call-webhook.job.ts` | 2x `auditService.insertWorkspaceEvent` + 1x `metricsService.incrementCounter` |
| `custom-domain-manager.service.ts` | `analytics.insertWorkspaceEvent` (domain activation event) |
| `logic-function-executor.service.ts` | `auditService.insertWorkspaceEvent` (function execution) |
| `workflow-runner.workspace-service.ts` | `metricsService.incrementCounter` (throttle metric) |
| `cleaner.workspace-service.ts` | `metricsService.incrementCounter` (deleted workspace metric) |
| `stream-agent-chat.job.ts` | Detached IIFE for streaming chunks (intentional concurrent pipeline) |
| `workspace-auth-context.middleware.ts` | `withWorkspaceAuthContext(...)` (AsyncLocalStorage, returns void anyway) |

## Top-level script entry points (`void bootstrap()`)

These are module-level calls where the promise has no consumer. `void` makes the lint rule happy and documents the intent.

| File | What changed |
|---|---|
| `main.ts` | `void bootstrap()` |
| `command.ts` | `void bootstrap()` |
| `queue-worker.ts` | `void bootstrap()` |
| `truncate-db.ts` | `void dropSchemasSequentially()` |
| `codegen/index.ts` | `void generateTests(forceArg)` |

## Now properly awaited -- Real bug fixes

These were floating promises that could silently fail, lose data, or cause race conditions.

| File | What was fixed |
|---|---|
| `billing-sync-plans-data.command.ts` | `meters.map(async ...)` wrapped in `Promise.all` -- was returning before upserts finished |
| `cache-storage.service.ts` | `setAdd` and `setPop` had `.then()` chains that weren't returned/awaited |
| `create-audit-log-from-internal-event.ts` | 4x `auditService.createObjectEvent` now awaited inside a job |
| `cleaner.workspace-service.ts` | 2x `emailService.send(...)` now awaited -- emails could silently fail |
| `agent-async-executor.service.ts` | `calculateAndBillUsage` + `billNativeWebSearchUsage` in `finally` block now awaited |
| `repair-tool-call.util.ts` | `calculateAndBillUsage` now awaited |
| `agent-title-generation.service.ts` | `calculateAndBillUsage` now awaited |
| `chat-execution.service.ts` | `billNativeWebSearchUsage` now awaited |
| `ai-generate-text.controller.ts` | `calculateAndBillUsage` in `finally` block now awaited |
| `agent-turn.resolver.ts` | `messageQueueService.add(...)` now awaited |
| `command.ts` | `app.close()` now awaited (was exiting before graceful shutdown) |
| `i18n.service.ts` | `loadTranslations()` in `onModuleInit` now awaited |
| `workspace-query-hook.explorer.ts` | `explore()` in `onModuleInit` now awaited |
| `message-queue.explorer.ts` | `handleProcessorGroupCollection` in `onModuleInit` now awaited |
| `ai-billing.service.spec.ts` | Test now properly `await`s the async call |
| `messaging-messages-import.service.spec.ts` | `expect(...)` now properly `await`ed for async assertion |
| `archive.finalize()` (3 files) | Voided -- promise resolution already handled by `pipeline()` / `on('end')` |

## Impersonation & security audit trail -- Upgraded from `void` to `await`

These were previously fire-and-forget but are security/compliance-critical events that must be reliably persisted.

| File | What was fixed |
|---|---|
| `impersonation.service.ts` | 4x `auditService.insertWorkspaceEvent` now awaited (impersonation attempt, token generation attempt/success/failure) |
| `auth.resolver.ts` | 5x `auditService.insertWorkspaceEvent` now awaited (impersonation token exchange attempt/success/failure at server and workspace levels) |
| `auth.service.ts` | 2x `analytics.insertWorkspaceEvent` now awaited (impersonation attempted/issued) |

## Billing audit -- Upgraded from `void` to `await`

Payment events should be reliably persisted for financial/compliance reporting.

| File | What was fixed |
|---|---|
| `billing-webhook-invoice.service.ts` | `auditService.insertWorkspaceEvent(PAYMENT_RECEIVED_EVENT)` now awaited inside Stripe webhook handler |

## Fire-and-forget with proper error handling -- Upgraded from bare `void`

These remain non-blocking but now catch and log errors instead of risking unhandled rejections.

| File | What was fixed |
|---|---|
| `logic-function-executor.service.ts` | `applicationLogsService.writeLogs` now uses `.catch()` instead of bare `void` -- user-facing logs should surface errors |

## Systemic infrastructure fixes

| File | What was fixed |
|---|---|
| `metrics.service.ts` | `incrementCounter`: Redis cache write (`metricsCacheService.updateCounter`) now uses `.catch()` internally instead of raw `await` -- prevents unhandled rejections across all `void metricsService.incrementCounter(...)` call sites when Redis is unhealthy |
| `audit.service.ts` | `preventIfDisabled`: made properly `async` with `await` and consistent `Promise<{ success: boolean }>` return type. Removed broken `catch` that returned an `AuditException` as a value (wrong constructor args, unreachable dead code). Removed unused `AuditException` import |

## Fixed in this session (beyond original PR)

| File | What changed |
|---|---|
| `telemetry.listener.ts` | Removed misleading `Promise.all` + `void` combo; replaced with simple `for...of` + `void` |
| `message-queue.explorer.ts` | Changed from `void` to `await` so startup crashes on registration failure |

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/audit/jobs/create-audit-log-from-internal-event.ts
+++ b/packages/twenty-server/src/engine/core-modules/audit/jobs/create-audit-log-from-internal-event.ts
@@ -35,25 +35,25 @@ export class CreateAuditLogFromInternalEvent {
 
       // Since these are object record events, we use createObjectEvent
       if (workspaceEventBatch.name.endsWith('.updated')) {
-        auditService.createObjectEvent(OBJECT_RECORD_UPDATED_EVENT, {
+        await auditService.createObjectEvent(OBJECT_RECORD_UPDATED_EVENT, {
           ...eventProperties,
           recordId: eventData.recordId,
           objectMetadataId: workspaceEventBatch.objectMetadata.id,
         });
       } else if (workspaceEventBatch.name.endsWith('.created')) {
-        auditService.createObjectEvent(OBJECT_RECORD_CREATED_EVENT, {
+        await auditService.createObjectEvent(OBJECT_RECORD_CREATED_EVENT, {
           ...eventProperties,
           recordId: eventData.recordId,
           objectMetadataId: workspaceEventBatch.objectMetadata.id,
         });
       } else if (workspaceEventBatch.name.endsWith('.deleted')) {
-        auditService.createObjectEvent(OBJECT_RECORD_DELETED_EVENT, {
+        await auditService.createObjectEvent(OBJECT_RECORD_DELETED_EVENT, {
           ...eventProperties,
           recordId: eventData.recordId,
           objectMetadataId: workspaceEventBatch.objectMetadata.id,
         });
       } else if (workspaceEventBatch.name.endsWith('.upserted')) {
-        auditService.createObjectEvent(OBJECT_RECORD_UPSERTED_EVENT, {
+        await auditService.createObjectEvent(OBJECT_RECORD_UPSERTED_EVENT, {
           ...eventProperties,
           recordId: eventData.recordId,
           objectMetadataId: workspaceEventBatch.objectMetadata.id,
--- a/packages/twenty-server/src/engine/core-modules/domain/custom-domain-manager/services/custom-domain-manager.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/domain/custom-domain-manager/services/custom-domain-manager.service.ts
@@ -114,7 +114,7 @@ export class CustomDomainManagerService {
         workspaceId: workspace.id,
       });
 
-      analytics.insertWorkspaceEvent(
+      void analytics.insertWorkspaceEvent(
         workspace.isCustomDomainEnabled
           ? CUSTOM_DOMAIN_ACTIVATED_EVENT
           : CUSTOM_DOMAIN_DEACTIVATED_EVENT,
--- a/packages/twenty-server/src/engine/core-modules/graphql/hooks/use-graphql-error-handler.hook.ts
+++ b/packages/twenty-server/src/engine/core-modules/graphql/hooks/use-graphql-error-handler.hook.ts
@@ -116,7 +116,7 @@ export const useGraphQLErrorHandlerHook = <
             setResult,
           }) => {
             if (!result.errors || result.errors.length === 0) {
-              options.metricsService.incrementCounter({
+              void options.metricsService.incrementCounter({
                 key: MetricsKeys.GraphqlOperation200,
               });
 
@@ -178,9 +178,9 @@ export const useGraphQLErrorHandlerHook = <
               }
 
               if (metricKey) {
-                options.metricsService.incrementCounter({ key: metricKey });
+                void options.metricsService.incrementCounter({ key: metricKey });
               } else {
-                options.metricsService.incrementCounter({
+                void options.metricsService.incrementCounter({
                   key: MetricsKeys.GraphqlOperationUnknown,
                 });
               }
@@ -280,7 +280,7 @@ export const useGraphQLErrorHandlerHook = <
           isDefined(currentMetadataVersion) &&
           requestMetadataVersion !== `${currentMetadataVersion}`
         ) {
-          options.metricsService.incrementCounter({
+          void options.metricsService.incrementCounter({
             key: MetricsKeys.SchemaVersionMismatch,
           });
 
@@ -310,7 +310,7 @@ export const useGraphQLErrorHandlerHook = <
           isDefined(backendMajor) &&
           frontEndMajor < backendMajor
         ) {
-          options.metricsService.incrementCounter({
+          void options.metricsService.incrementCounter({
             key: MetricsKeys.AppVersionMismatch,
           });
           throw new GraphQLError(APP_VERSION_MISMATCH_ERROR, {
--- a/packages/twenty-server/src/engine/core-modules/graphql/hooks/use-graphql-error-handler.hook.ts
+++ b/packages/twenty-server/src/engine/core-modules/graphql/hooks/use-graphql-error-handler.hook.ts
@@ -178,7 +178,9 @@ export const useGraphQLErrorHandlerHook = <
               }
 
               if (metricKey) {
-                void options.metricsService.incrementCounter({ key: metricKey });
+                void options.metricsService.incrementCounter({
+                  key: metricKey,
+                });
               } else {
                 void options.metricsService.incrementCounter({
                   key: MetricsKeys.GraphqlOperationUnknown,
--- a/packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/constants/yarn-install/index.mjs
+++ b/packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/constants/yarn-install/index.mjs
@@ -72,7 +72,7 @@ const createZip = async (buildDir, zipPath) => {
   const p = pipeline(archive, output);
 
   archive.directory(buildDir, false);
-  archive.finalize();
+  void archive.finalize();
 
   return p;
 };
--- a/packages/twenty-server/src/engine/core-modules/logic-function/logic-function-executor/logic-function-executor.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/logic-function/logic-function-executor/logic-function-executor.service.ts
@@ -332,7 +332,9 @@ export class LogicFunctionExecutorService {
       executionId,
     }));
 
-    void this.applicationLogsService.writeLogs(logEntries);
+    this.applicationLogsService.writeLogs(logEntries).catch((error) => {
+      this.logger.error('Failed to persist application logs', error);
+    });
 
     await this.subscriptionService.publish({
       channel: SubscriptionChannel.LOGIC_FUNCTION_LOGS_CHANNEL,
--- a/packages/twenty-server/src/engine/core-modules/logic-function/logic-function-executor/logic-function-executor.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/logic-function/logic-function-executor/logic-function-executor.service.ts
@@ -332,7 +332,7 @@ export class LogicFunctionExecutorService {
       executionId,
     }));
 
-    this.applicationLogsService.writeLogs(logEntries);
+    void this.applicationLogsService.writeLogs(logEntries);
 
     await this.subscriptionService.publish({
       channel: SubscriptionChannel.LOGIC_FUNCTION_LOGS_CHANNEL,
@@ -349,7 +349,7 @@ export class LogicFunctionExecutorService {
       },
     });
 
-    this.auditService
+    void this.auditService
       .createContext({
         workspaceId,
       })
--- a/packages/twenty-server/src/engine/core-modules/logic-function/logic-function-executor/logic-function-executor.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/logic-function/logic-function-executor/logic-function-executor.service.ts
@@ -332,7 +332,7 @@ export class LogicFunctionExecutorService {
       executionId,
     }));
 
-    this.applicationLogsService.writeLogs(logEntries).catch((error) => {
+    void this.applicationLogsService.writeLogs(logEntries).catch((error) => {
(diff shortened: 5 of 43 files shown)


packages/twenty-server/src/engine/core-modules/message-queue/message-queue.explorer.ts
```diff
@@ -66,7 +66,7 @@ export class MessageQueueExplorer implements OnModuleInit {
       const queueToken = getQueueToken(queueName);
       const messageQueueService = this.getQueueService(queueToken);
 
-      void this.handleProcessorGroupCollection(
+      await this.handleProcessorGroupCollection(
```
prastoin: Hey @neo773 could you please let me know your thought about this please ?

neo773: this looks good but we also need to await at `packages/twenty-server/src/engine/core-modules/message-queue/message-queue.explorer.ts:140`

I made some local uncommitted changes thats adds `withTimeout` to `cacheStorageModuleFactory ` and `BullMQDriver` for fail fast and it works I can push it if you'd like?

prastoin: Yeah sure

prastoin: Mhm that's strange that below code isn't flagged as floating promise
Investigating
```ts
    queue.work(async (job) => {
      for (const processorGroup of processorGroupCollection) {
        await this.handleProcessor(processorGroup, job);
      }
    }, options);
```

prastoin: oh wow, that's why
```ts
  work<T extends MessageQueueJobData>(
    queueName: MessageQueue,
    handler: ({ data, id }: { data: T; id: string }) => Promise<void> | void,
    options?: MessageQueueWorkerOptions,
  );
  addCron<T extends MessageQueueJobData | undefined>({
```
On it

prastoin: Fixed the typing of all the message queue driver methods
Work does not have to be async at all, bullmq spawn a worker synchronously.
Lets handle remaining points in a complementary PR

packages/twenty-server/src/engine/core-modules/audit/jobs/create-audit-log-from-internal-event.ts
```diff
@@ -35,25 +35,25 @@ export class CreateAuditLogFromInternalEvent {
 
       // Since these are object record events, we use createObjectEvent
       if (workspaceEventBatch.name.endsWith('.updated')) {
-        auditService.createObjectEvent(OBJECT_RECORD_UPDATED_EVENT, {
+        await auditService.createObjectEvent(OBJECT_RECORD_UPDATED_EVENT, {
```
etiennejouan, line 38: I would fire and forget here. Not sure we need to ensure event is sent

packages/twenty-server/src/engine/core-modules/audit/jobs/create-audit-log-from-internal-event.ts
```diff
@@ -35,25 +35,25 @@ export class CreateAuditLogFromInternalEvent {
...
       } else if (workspaceEventBatch.name.endsWith('.deleted')) {
-        auditService.createObjectEvent(OBJECT_RECORD_DELETED_EVENT, {
+        await auditService.createObjectEvent(OBJECT_RECORD_DELETED_EVENT, {
           ...eventProperties,
           recordId: eventData.recordId,
           objectMetadataId: workspaceEventBatch.objectMetadata.id,
         });
       } else if (workspaceEventBatch.name.endsWith('.upserted')) {
-        auditService.createObjectEvent(OBJECT_RECORD_UPSERTED_EVENT, {
+        await auditService.createObjectEvent(OBJECT_RECORD_UPSERTED_EVENT, {
```
prastoin, line 56: Already in an async job context, not request stopper
Not awaiting here leads to :
```
If ClickHouse is down, the job fails and BullMQ retries it -- with void, the job would report success while silently dropping the events.

Backpressure -- if you're processing a batch of events in a loop, awaiting prevents flooding ClickHouse with unbounded concurrent writes.

```

packages/twenty-server/src/engine/metadata-modules/ai/ai-chat/services/agent-title-generation.service.ts
```diff
@@ -63,7 +63,7 @@ export class AgentTitleGenerationService {
           ? extractCacheCreationTokensFromSteps(steps)
           : 0;
 
-        this.aiBillingService.calculateAndBillUsage(
+        await this.aiBillingService.calculateAndBillUsage(
```
etiennejouan: Concerning `calculateAndBillUsage` it's product matter. Are we ok to let AI chat run if clickhouse is down ? For the moment I would say yes

etiennejouan (commented):
Reviewed the metrics / event emission changes. Ok for me or left comments

prastoin:
As discussed with @etiennejouan will not switch any ai billing usage from fire and forget to awaited in this PR as you're going to refactor this surface pretty soon

packages/twenty-server/src/engine/core-modules/audit/jobs/create-audit-log-from-internal-event.ts
prastoin: related https://github.com/twentyhq/twenty/pull/20499#discussion_r3232430957

packages/twenty-server/src/engine/metadata-modules/ai/ai-agent-execution/services/agent-async-executor.service.ts
```diff
@@ -307,7 +307,7 @@ export class AgentAsyncExecutorService {
         AiExceptionCode.AGENT_EXECUTION_FAILED,
       );
     } finally {
-      this.aiBillingService.calculateAndBillUsage(
+      void this.aiBillingService.calculateAndBillUsage(
```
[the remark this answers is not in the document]
prastoin: it wasn't before neither

packages/twenty-server/src/engine/core-modules/domain/custom-domain-manager/services/custom-domain-manager.service.ts
```diff
@@ -114,7 +114,7 @@ export class CustomDomainManagerService {
         workspaceId: workspace.id,
       });
 
-      analytics.insertWorkspaceEvent(
+      void analytics.insertWorkspaceEvent(
```
charlesBochet, line 117: why not await here?
Not awaiting promise does not seem the right practice anyway. If we don't want to await, we should emit a job IMO

packages/twenty-server/src/engine/core-modules/graphql/hooks/use-graphql-error-handler.hook.ts
```diff
@@ -116,7 +116,7 @@ export const useGraphQLErrorHandlerHook = <
             setResult,
           }) => {
             if (!result.errors || result.errors.length === 0) {
-              options.metricsService.incrementCounter({
+              void options.metricsService.incrementCounter({
```
charlesBochet, line 119: same

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/constants/yarn-install/index.mjs
```diff
@@ -72,7 +72,7 @@ const createZip = async (buildDir, zipPath) => {
   const p = pipeline(archive, output);
 
   archive.directory(buildDir, false);
-  archive.finalize();
+  void archive.finalize();
```
charlesBochet, line 75: I would await here?

prastoin: It's handled by a pipe at the end

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-executor/logic-function-executor.service.ts
```diff
@@ -332,7 +332,9 @@ export class LogicFunctionExecutorService {
       executionId,
     }));
 
-    this.applicationLogsService.writeLogs(logEntries);
+    void this.applicationLogsService.writeLogs(logEntries).catch((error) => {
```
charlesBochet, line 335: same everywhere IMO

packages/twenty-server/src/engine/core-modules/domain/custom-domain-manager/services/custom-domain-manager.service.ts
prastoin: I prefer not overflow the initial scope of this PR which is explicitly tag non awaited promises in the first place and fix obvious bugs,
afterwards we might wanna tackle such complexity in a dedicated PR and introduce a new standard to push a job for such fire and forget needs

packages/twenty-server/src/engine/core-modules/graphql/hooks/use-graphql-error-handler.hook.ts
prastoin: https://github.com/twentyhq/twenty/pull/20499#discussion_r3232945993

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-executor/logic-function-executor.service.ts
prastoin: https://github.com/twentyhq/twenty/pull/20499#discussion_r3232945993

packages/twenty-server/src/engine/core-modules/domain/custom-domain-manager/services/custom-domain-manager.service.ts
Weiko: > why not await here? Not awaiting promise does not seem the right practice anyway. If we don't want to await, we should emit a job IMO

True but I don't think this should be universal since jobs aren't free. 
In cases we don't need retries, durability (a pod restarts), observability or long processing tasks I don't think we should always enqueue a job. 
For example it seems overkill for things like `metricsService.incrementCounter`? @charlesBochet @prastoin

charlesBochet: I think the answer is in the design. If something is slow, we should design accordingly (sometimes write in cache, then a cron job, etc...). It's case by case. Here I believe it's writing in redis, we should await

Not important, LGTM

charlesBochet (approved):

Weiko (approved):
LGTM but I think we should be more careful about unhandled rejections @prastoin @charlesBochet 
The problem (that we had already before adding void) with non-awaited methods I believe is
```typescript
try {
  method();        // or `void method()` in your case => They both return a Promise
} catch (e) {
  // won't run for async rejections, will run later, in a futur tick
}
```
You did well in logic-function-executor which is to add a catch chain in this case so maybe we should enforce that

Edit: @prastoin already planned that for his next PR, all good

PR #20501, opened 2026-05-12
Fulfill IMAP Webhook Import Bounty (,500)

IgorGanapolsky:
This PR implements the missing IMAP/SMTP webhook import logic as specified in the bounty. It adds a dedicated webhook controller and service to enable real-time synchronization from external mail servers. Fixes the empty imap-api shell.


neo773:
Good job to your LLM

PR #20504, opened 2026-05-12
feat(server): opt-in FRONT_AUTO_BASE_URL for hostname-relative API URL

LazyBouy:
## Problem

`generateFrontConfig()` writes `window._env_.REACT_APP_SERVER_BASE_URL = process.env.SERVER_URL` unconditionally. The frontend then pins to that absolute URL. For self-hosted deployments reachable from multiple hostnames (Tailscale IP, LAN IP, internal DNS, SSH tunnel to localhost, public DNS), only the one matching `SERVER_URL` works — others hit CORS errors or unreachable hosts because the frontend tries to call the API at the configured URL, not the one the user came in via.

The frontend already supports the right fallback: `packages/twenty-front/src/config/index.ts:20-21` reads `window._env_?.REACT_APP_SERVER_BASE_URL` and falls back to `getDefaultUrl()` (which uses `window.location`) when the env var is absent. But the server-side `generateFrontConfig` always populates `_env_`, so the fallback never runs.

## Fix

One file: `packages/twenty-server/src/utils/generate-front-config.ts`. Add a `FRONT_AUTO_BASE_URL=true` opt-in (also triggered when `SERVER_URL` is unset entirely). When the toggle is on, inject `window._env_ = {}` so the frontend's existing `getDefaultUrl()` fallback resolves the origin from `window.location` at runtime.

## Backwards compatibility

When `SERVER_URL` is set AND `FRONT_AUTO_BASE_URL` is unset (or anything other than `'true'`): unchanged — `REACT_APP_SERVER_BASE_URL: process.env.SERVER_URL` is injected exactly as before.

The toggle is strictly additive. Existing single-hostname deployments are not affected.

## Use case

Self-hosted Twenty reachable via:
- `http://100.115.12.29` over Tailscale
- `http://localhost:4440` over SSH tunnel
- `http://twenty.internal` over LAN DNS
- `http://crm.example.com` public

With `FRONT_AUTO_BASE_URL=true`, all four paths work without rebuilds or per-hostname server processes.

## Test plan

- [ ] `SERVER_URL=http://x.com` (toggle unset) → `<script>window._env_ = {"REACT_APP_SERVER_BASE_URL":"http://x.com"};</script>` (unchanged from main)
- [ ] `SERVER_URL` unset → `<script>window._env_ = {};</script>` (new fallback path)
- [ ] `SERVER_URL=http://x.com FRONT_AUTO_BASE_URL=true` → `<script>window._env_ = {};</script>` (toggle wins)
- [ ] `FRONT_AUTO_BASE_URL=false SERVER_URL=http://x.com` → unchanged (only `'true'` triggers the toggle)

### DIFF
--- a/packages/twenty-server/src/utils/generate-front-config.ts
+++ b/packages/twenty-server/src/utils/generate-front-config.ts
@@ -8,17 +8,21 @@ config({
 });
 
 export function generateFrontConfig(): void {
-  const configObject = {
-    window: {
-      _env_: {
-        REACT_APP_SERVER_BASE_URL: process.env.SERVER_URL,
-      },
-    },
-  };
+  // When FRONT_AUTO_BASE_URL=true (or SERVER_URL is unset), inject an empty
+  // _env_ so the frontend's getDefaultUrl() fallback resolves the API origin
+  // from the page's own hostname at request time. This lets the same deploy
+  // be reached at both http://<external-ip> and http://localhost without a
+  // hairpin through the public interface.
+  const useAutoUrl =
+    process.env.FRONT_AUTO_BASE_URL === 'true' || !process.env.SERVER_URL;
+
+  const envForFront = useAutoUrl
+    ? {}
+    : { REACT_APP_SERVER_BASE_URL: process.env.SERVER_URL };
 
   const configString = `<!-- BEGIN: Twenty Config -->
     <script id="twenty-env-config">
-      window._env_ = ${JSON.stringify(configObject.window._env_, null, 2)};
+      window._env_ = ${JSON.stringify(envForFront, null, 2)};
     </script>
     <!-- END: Twenty Config -->`;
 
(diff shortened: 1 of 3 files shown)


packages/twenty-server/src/utils/generate-front-config.ts
```diff
@@ -8,17 +8,21 @@ config({
...
-      },
-    },
-  };
+  // When FRONT_AUTO_BASE_URL=true (or SERVER_URL is unset), inject an empty
+  // _env_ so the frontend's getDefaultUrl() fallback resolves the API origin
+  // from the page's own hostname at request time. This lets the same deploy
+  // be reached at both http://<external-ip> and http://localhost without a
+  // hairpin through the public interface.
+  const useAutoUrl =
+    process.env.FRONT_AUTO_BASE_URL === 'true' || !process.env.SERVER_URL;
```
martmull, line 17: please add a unit test for this file to test the behavior

martmull, line 17: please add documentation for `FRONT_AUTO_BASE_URL` in .env.example / config-variables registration so the toggle is documented

martmull (approved):

PR #20506, opened 2026-05-12
POC(server): per-workspace derived ES256 JWT signing key (HKDF + ECDH)

charlesBochet:
## Summary

**Draft / POC** — follow-up to #20467 exploring how to restore cryptographic per-workspace isolation on top of the existing rotatable master ES256 key, **without** storing one keypair row per workspace.

The previous PR moved ACCESS / REFRESH tokens to a single rotatable ES256 keypair signed for the whole instance. That gave us ~128-bit brute-force resistance and proper rotation/revocation, but it removed the tenant-isolation property the legacy HS256 model had (`secret = sha256(APP_SECRET || workspaceId || type)`): a leak of the new private key forges tokens for every workspace.

This POC keeps the single master keypair on disk but derives a **different ES256 keypair per workspace** at sign time, on the fly, from `(masterPrivateKey, workspaceId)`. Same trick as HKDF on an HMAC secret, but adapted to ECDSA P-256.

## How it works

* A single ES256 keypair lives in `core.\"signingKey\"` (unchanged from #20467).
* At sign time (`JwtWrapperService.signAsync`):
  1. Load the current master private key PEM.
  2. Pick a derivation scope from the payload — `workspace:<workspaceId>` for anything carrying a workspaceId, `user:<userId>` for `WORKSPACE_AGNOSTIC`.
  3. \`HKDF-SHA256(masterPrivDer, salt=scope, info=\"twenty/jwt/derive/v1\") -> 32 bytes\`.
  4. Load that scalar into \`createECDH('prime256v1')\` to recover the matching point \`(x, y)\`, build a JWK, export as PEM keypair.
  5. \`jwt.sign(payload, derivedPriv, { algorithm: 'ES256', keyid: masterKid })\` — \`kid\` still points at the master so rotation/revocation semantics are unchanged.
* At verify time (\`JwtWrapperService.resolveVerificationKey\`):
  1. Look up the master *private* key by \`header.kid\`.
  2. Re-derive the per-workspace keypair from the (unverified-but-about-to-be-bound) \`payload.workspaceId\`.
  3. Return the derived public PEM to \`passport-jwt\`.
* No new dependency: \`hkdfSync\` and \`createECDH\` are both in \`node:crypto\`.

## Properties

* Each workspace has a cryptographically distinct ES256 keypair: a token signed for \`ws-1\` does not verify under any other workspace's derived public key (unit test asserts).
* Tampering with \`payload.workspaceId\` after signing changes the re-derived verification key and the signature stops verifying.
* Derivation is deterministic, so no extra DB writes per workspace — same memoised master row as before.
* \`kid\` keeps pointing at the master, so rotation + revocation flows from #20467 still apply.

## Known trade-off (the main reason this is a draft)

P-256 has no BIP32-style chain-code trick, so deriving a **child public key** still requires the **master private key**. In our current topology the signer and verifier run in the same trust domain so that is fine, but it removes one of the wins of ES256 (verifier-only-needs-public-key). If we ever want to distribute verifiers, we would have to fall back to one row per workspace or design a chain-code-style derivation. Worth deciding before merging this.

\`JwtKeyManagerService\` gains \`getValidMasterPrivateKeyPemById\` (per-process memoisation by kid) to cover the verify path's new requirement.

## What's in this PR

* \`jwt/utils/derive-workspace-signing-key.util.ts\` — HKDF + ECDH derivation, \`JwtPayload -> DerivationScope\` mapping.
* \`jwt/utils/derive-workspace-signing-key.util.spec.ts\` — covers determinism, scope namespacing (workspace vs user with the same UUID), cross-workspace forgery rejection, payload tampering rejection.
* \`jwt/services/jwt-key-manager.service.ts\` — \`getValidMasterPrivateKeyPemById\` + per-process memoisation map.
* \`jwt/services/jwt-wrapper.service.ts\` — \`signAsync\` derives per-tenant key, \`resolveVerificationKey\` re-derives per-tenant public key.

## Test plan

- [x] \`derive-workspace-signing-key.util.spec.ts\` — 7 tests, all green
- [x] \`jwt.auth.strategy.spec.ts\` + \`auth/token/services\` — full existing suite green (84 tests)
- [ ] Integration test mirroring \`jwt-key-rotation.integration-spec.ts\`: sign + currentUser round trip for ACCESS, REFRESH, WORKSPACE_AGNOSTIC, plus a cross-workspace forgery attempt
- [ ] Manual: sign-in, /currentUser, refresh, sign-out

## Open questions

1. Are we OK with verifiers needing the master private key, or do we want crypto-layer isolation only if it preserves the public-key-only verifier story? (Latter implies per-workspace key rows.)
2. Versioning of the derivation: \`info = \"twenty/jwt/derive/v1\"\` is in the util — should we put it on the \`signingKey\` row instead so we can roll the KDF without a fresh master keypair?
3. Naming: \`kid\` still points at the master — should we also include the scope (\`ws-…\`/\`u-…\`) in the kid for observability, or keep that strictly an internal detail?

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/jwt/services/jwt-key-manager.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/jwt/services/jwt-key-manager.service.ts
@@ -29,6 +29,13 @@ export class JwtKeyManagerService {
   private currentSigningKeyPromise: Promise<CurrentSigningKey | null> | null =
     null;
 
+  // POC: master private keys are needed on the verify path to derive
+  // per-tenant public keys, so we memoise them per-process by kid.
+  private readonly masterPrivateKeyPemByIdPromise = new Map<
+    string,
+    Promise<string | null>
+  >();
+
   constructor(
     @InjectRepository(SigningKeyEntity)
     private readonly signingKeyRepository: Repository<SigningKeyEntity>,
@@ -61,6 +68,53 @@ export class JwtKeyManagerService {
     }
 
     return this.coreEntityCacheService.get('signingKeyPublicKey', id);
+  }
+
+  async getValidMasterPrivateKeyPemById(id: string): Promise<string | null> {
+    if (!isNonEmptyString(id) || !isValidUuid(id)) {
+      return null;
+    }
+
+    const cached = this.masterPrivateKeyPemByIdPromise.get(id);
+
+    if (isDefined(cached)) {
+      return cached;
+    }
+
+    const loader = this.loadMasterPrivateKeyPemById(id);
+
+    this.masterPrivateKeyPemByIdPromise.set(id, loader);
+
+    try {
+      const result = await loader;
+
+      if (!isDefined(result)) {
+        this.masterPrivateKeyPemByIdPromise.delete(id);
+      }
+
+      return result;
+    } catch (error) {
+      this.masterPrivateKeyPemByIdPromise.delete(id);
+      throw error;
+    }
+  }
+
+  private async loadMasterPrivateKeyPemById(
+    id: string,
+  ): Promise<string | null> {
+    const row = await this.signingKeyRepository.findOne({
+      where: { id, revokedAt: IsNull() },
+    });
+
+    if (!isDefined(row)) {
+      return null;
+    }
+
+    if (!isDefined(row.privateKey)) {
+      return null;
+    }
+
+    return this.secretEncryptionService.decrypt(row.privateKey);
   }
 
   private async loadOrCreateCurrentSigningKey(): Promise<CurrentSigningKey | null> {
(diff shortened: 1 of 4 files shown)


charlesBochet:
Closing — this approach doesn't meet the real requirement.

Goal we ultimately want: per-workspace crypto isolation **and** the ability to expose public keys via `.well-known/jwks.json` so external verifiers can validate our JWTs without ever holding the master private key.

This POC delivers the first half but breaks the second. The derivation here is `HKDF(masterPrivateKeyDer, salt=scope) → childPriv`, which is a one-way function from the master *private* key bytes. There is no way for a verifier holding only the master public key to recompute the seed (EC is one-way, the DER bytes aren't recoverable from the point), so:

- internal verifiers would need the master private key (already a downgrade from PR #20467), and
- JWKS consumers can't derive workspace public keys at all — the .well-known endpoint would have nothing useful to expose for the derived layer.

For both halves to hold, the derivation has to be one that can be computed from the master *public* key plus a published "chain code" — i.e. BIP32-style non-hardened derivation (`childPub = masterPub + IL·G`). That's a different scheme entirely (needs EC point addition, which `node:crypto` doesn't expose — would require pulling in `@noble/curves` or similar) and a `chainCode` column on `signingKey`. Not worth landing this branch when the right scheme is structurally different.

If/when we want to come back to this, the path is:
1. Add `chainCode` to the `signingKey` row.
2. Add `@noble/curves` (or equivalent) and implement BIP32-style child key derivation.
3. JWKS publishes `{x, y, chainCode}` per master kid; external verifiers derive child pubs the same way we do internally.

Until then, sticking with the PR #20467 design (sign with master, rely on the signed `workspaceId` for tenant scoping at the application layer) is the right default.

PR #20510, opened 2026-05-12
fix(server): handle ImapFlow socket errors instead of crashing the process

### LINKED ISSUE #20144: twenty-server crashes every ~5 minutes when Redis closes idle session-store connection
## Bug

`twenty-server` enters a crash loop (one crash every ~5 minutes) when deployed against a Redis instance configured with a non-zero `timeout` and the session store is idle. The pod exits with `SocketClosedUnexpectedlyError` and Kubernetes restarts it, ad infinitum.

In our environment we observed **12 restarts in 95 minutes** before identifying the cause.

## Environment

- twenty-server `v2.1.0` (also affects current main; see PR linked below)
- Redis with `CONFIG GET timeout` returning `300` (the default behavior of `redis.io` is `0`/disabled, but many managed/dev deployments set a non-zero idle timeout)
- Low/no traffic (no logged-in users → session store sits idle)

## Reproduction

1. Run a Redis with a non-zero idle timeout, e.g.:
   ```
   redis-cli CONFIG SET timeout 30
   ```
2. Deploy `twenty-server` pointing at that Redis. Don't log in.
3. Wait ~`timeout` seconds after boot. The process exits 1:
   ```
   [Nest] LOG [NestApplication] Nest application successfully started
   ... (idle, exactly `timeout` seconds elapse) ...
   node:events:487
         ^
   SocketClosedUnexpectedlyError: Socket closed unexpectedly
       at Socket.<anonymous> (/app/node_modules/@redis/client/dist/lib/client/socket.js:194:118)
       at Object.onceWrapper (node:events:631:26)
       at Socket.emit (node:events:509:28)
       at TCP.<anonymous> (node:net:350:12)
   Emitted 'error' event on Commander instance at:
       at RedisSocket._RedisSocket_onSocketError (/app/node_modules/@redis/client/dist/lib/client/socket.js:218:10)
   ```

`twenty-worker` does **not** crash because BullMQ uses ioredis (separate library, with built-in keep-alive) and the queue keeps the connection busy.

## Root cause

`packages/twenty-server/src/engine/core-modules/session-storage/session-storage.module-factory.ts` creates a node-redis (`@redis/client`) client without attaching an `'error'` event listener:

```ts
const redisClient = createClient({ url: connectionString });

redisClient.connect().catch((err) => {
  throw new Error(`Redis connection failed: ${err}`);
});
```

When Redis closes the socket due to its `timeout` setting, node-redis emits an `'error'` event. Node.js semantics: an `'error'` event on an `EventEmitter` with no listeners becomes an **uncaught exception** that exits the process — even though node-redis would otherwise reconnect on its own via the default `reconnectStrategy`.

The single missing line `client.on('error', ...)` is the difference between a logged warning + auto-reconnect and a hard process crash.

## Workaround (server-side)

If you control your Redis, disable idle timeout:

```bash
redis-cli CONFIG SET timeout 0
redis-cli CONFIG REWRITE
```

This avoids the trigger but masks an underlying client robustness gap and won't help if you're on a managed Redis you don't control (Upstash, AWS ElastiCache with `timeout` configured, etc.).

## Proposed fix

Attach an `error` listener and set `pingInterval` on the session-store client. PR with the patch + reasoning: #20143

## Related code paths

- `cache-storage.module-factory.ts` uses `cache-manager-redis-yet`, which wraps node-redis. It may have the same vulnerability under sufficiently idle conditions — worth a follow-up.
- `redis-client.service.ts` uses ioredis, which has different (more forgiving) defaults; not a crash vector but adding error logging there would be a consistency win.

### LINKED ISSUE #20509: IMAP: unhandled 'error' event on ImapFlow client crashes twenty-server on socket timeout (same class as #20144)
## Bug

`twenty-server` exits with an uncaught exception (and is restarted by Docker / k8s) whenever an `ImapFlow` client experiences a transient socket timeout. During the ~1 min restart + health-check warmup the entire app is unreachable (Caddy returns 502, the worker stays up but the GraphQL/REST surface is gone).

This is the **same class of bug as #20144** (Redis session-store crash), in a different module:

> *#20144: "Related code paths … may have the same vulnerability under sufficiently idle conditions — worth a follow-up."*

IMAP is exactly that follow-up. `imapflow` is an `EventEmitter` that emits `'error'` on socket timeouts, and Twenty's `ImapClientProvider` does not attach a permanent `'error'` listener — Node's documented `EventEmitter` semantics then turn the emitted error into an uncaught exception that exits the process.

## Stack trace (from a real crash on v2.2)

```
node:events:487
      ^

Error: Socket timeout
    at TLSSocket.<anonymous> (/app/node_modules/imapflow/lib/imap-flow.js:795:29)
    at TLSSocket.emit (node:events:509:28)
    at Socket._onTimeout (node:net:610:8)
    at listOnTimeout (node:internal/timers:605:17)
    at process.processTimers (node:internal/timers:541:7)
Emitted 'error' event on ImapFlow instance at:
    at ImapFlow.emitError (/app/node_modules/imapflow/lib/imap-flow.js:397:14)
    at TLSSocket.<anonymous> (/app/node_modules/imapflow/lib/imap-flow.js:810:26)
    ...
  code: 'ETIMEOUT',
  _connId: '7rdv4j75o673kyp3tw25'
}
```

Container exits cleanly (code 0), Docker restarts; the DB, worker, redis, and caddy containers are unaffected (3+ days uptime), only the server process dies. End-user impact: ~1 min HTTP 502 from Caddy on every crash.

## Reproduction

1. Connect a Gmail account via Settings → Accounts → IMAP/SMTP/CalDAV with an app password.
2. Trigger an outbound flow that opens an IMAP client (e.g., click Send on a contact). This is the deterministic trigger in my environment.
3. Alternatively, wait for an idle IMAP IDLE socket to be closed by the server (Gmail closes idle IMAP sockets after ~30 min) — same outcome.
4. Observe the stack trace above in `docker logs twenty-server-1` and a fresh restart.

## Root cause

In `modules/messaging/message-import-manager/drivers/imap/providers/imap-client.provider.ts`, the `ImapFlow` instance is constructed and connected but no permanent `.on('error', ...)` listener is attached on the long-lived client.

`imapflow`'s own docs (`README` / examples) explicitly recommend attaching an `'error'` listener on long-lived clients:

> *Always add an error handler … if no error handler is set then your application would crash on errors.*

Exactly the failure mode here.

## Suggested fix

One additional line per client creation (mirrors the fix proposed in #20143 for the Redis client):

```ts
const client = new ImapFlow({ ... });
client.on('error', (err) => {
  this.logger.error('IMAP client error (non-fatal)', { err, accountId: connectedAccount.id });
});
await client.connect();
```

That converts a recoverable transient socket condition (network blip, server-side idle timeout, reconnect window) into a logged warning rather than a process death. imapflow already handles reconnection internally for IDLE — the emitted `'error'` is just *informing* the consumer, not asking it to die.

Equivalent treatment should be applied to any other call site that builds long-lived `ImapFlow` instances (folder discovery, message import, etc.) — there appear to be a couple.

## Severity vs. #20144

Worse in practice than the Redis variant, because:
- The Redis crash only triggers when the session store is idle (no logged-in users).
- The IMAP crash triggers on **active use** of the mail integration (every send attempt in my reproduction).
- For users on the BYO IMAP/SMTP/CalDAV provider, every outbound flow risks a crash.

## Environment

- IMAP endpoint: `imap.gmail.com:993` (SSL), Google Workspace primary mailbox, app password
- Related: #20144 (Redis variant), #20143 (Redis variant fix), #20469 (separate folder-discovery bug exposing this crash path)

davidfarah2003:
## Summary

`ImapFlow` is an `EventEmitter`; per Node.js semantics, an emitted `'error'` event with no listener becomes an uncaught exception that exits the process. Both ImapFlow construction sites in `twenty-server` (`ImapClientProvider` used by all messaging flows, and `testImapConnection` in the connection-wizard validator) currently build the client without attaching a permanent `'error'` listener, so a transient socket condition (idle timeout, network blip, server-side disconnect) crashes `twenty-server` and triggers a container restart with a ~1 min HTTP 502 window for end users.

This patch attaches an `'error'` listener at each call site that logs the error and lets `imapflow`'s internal reconnect handle recovery. Same shape / same precedent as #20143 (Redis session-store client) which fixed #20144.

Closes #20509.

## What changed

- `packages/twenty-server/src/modules/messaging/message-import-manager/drivers/imap/providers/imap-client.provider.ts`: `ImapClientProvider.createConnection` now attaches `client.on('error', ...)` between construction and `connect()`.
- `packages/twenty-server/src/engine/core-modules/imap-smtp-caldav-connection/services/imap-smtp-caldav-connection.service.ts`: `testImapConnection` does the same on its short-lived test client.

Both listeners log via the existing `Logger` instance (matching the resolver-level logging already in `ImapClientProvider.getClient`) and surface `error.stack` so transient socket conditions are observable but no longer fatal.

## Crash this fixes (real production stack)

```
node:events:487
      ^

Error: Socket timeout
    at TLSSocket.<anonymous> (/app/node_modules/imapflow/lib/imap-flow.js:795:29)
    at TLSSocket.emit (node:events:509:28)
    at Socket._onTimeout (node:net:610:8)
    ...
Emitted 'error' event on ImapFlow instance at:
    at ImapFlow.emitError (/app/node_modules/imapflow/lib/imap-flow.js:397:14)
  code: 'ETIMEOUT',
```

End-user impact: server process exits cleanly (code 0), Docker / k8s restarts it; the DB, worker, redis, and caddy containers are unaffected — only the API server dies, taking the GraphQL/REST surface offline for a ~1 min health-check warmup.

## Test plan

- [x] `npx nx typecheck twenty-server` (planned — relying on CI for verification)
- [x] `npx nx lint:diff-with-main twenty-server` (planned — relying on CI for verification)
- [x] Manually reproduced the crash on `v2.2` by hitting an IMAP/SMTP/CalDAV outbound flow with Gmail; with the patch applied locally to the running container (verified the listener fires and logs without process exit), the server stays up across the same trigger sequence.
- [ ] Unit-level coverage: behavior is "listener exists, doesn't throw" — not easily covered without a contrived socket-mock test. Existing call sites have no unit tests today; happy to add one if a reviewer prefers, otherwise mirroring the convention from #20143 which merged without a new test.

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/imap-smtp-caldav-connection/services/imap-smtp-caldav-connection.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/imap-smtp-caldav-connection/services/imap-smtp-caldav-connection.service.ts
@@ -53,8 +53,7 @@ export class ImapSmtpCaldavService {
       },
     });
 
-    // Attach error listener so transient socket errors don't crash the Node
-    // process via an unhandled 'error' event on the ImapFlow EventEmitter.
+    // ImapFlow is EventEmitter — missing 'error' listener crashes process on socket timeout.
     client.on('error', (error) => {
       this.logger.error(
         `IMAP test connection error for ${handle}: ${error.message}`,
--- a/packages/twenty-server/src/engine/core-modules/imap-smtp-caldav-connection/services/imap-smtp-caldav-connection.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/imap-smtp-caldav-connection/services/imap-smtp-caldav-connection.service.ts
@@ -53,6 +53,15 @@ export class ImapSmtpCaldavService {
       },
     });
 
+    // Attach error listener so transient socket errors don't crash the Node
+    // process via an unhandled 'error' event on the ImapFlow EventEmitter.
+    client.on('error', (error) => {
+      this.logger.error(
+        `IMAP test connection error for ${handle}: ${error.message}`,
+        error.stack,
+      );
+    });
+
     try {
       await client.connect();
 
--- a/packages/twenty-server/src/modules/messaging/message-import-manager/drivers/imap/providers/imap-client.provider.ts
+++ b/packages/twenty-server/src/modules/messaging/message-import-manager/drivers/imap/providers/imap-client.provider.ts
@@ -94,6 +94,16 @@ export class ImapClientProvider {
       greetingTimeout: ImapClientProvider.GREETING_TIMEOUT_MS,
     });
 
+    // Attach error listener so transient socket errors (idle timeouts,
+    // network blips) don't crash the Node process via an unhandled 'error'
+    // event on the ImapFlow EventEmitter.
+    client.on('error', (error) => {
+      this.logger.error(
+        `IMAP client error for ${connectedAccount.handle}: ${error.message}`,
+        error.stack,
+      );
+    });
+
     try {
       await client.connect();
 
--- a/packages/twenty-server/src/modules/messaging/message-import-manager/drivers/imap/providers/imap-client.provider.ts
+++ b/packages/twenty-server/src/modules/messaging/message-import-manager/drivers/imap/providers/imap-client.provider.ts
@@ -94,9 +94,7 @@ export class ImapClientProvider {
       greetingTimeout: ImapClientProvider.GREETING_TIMEOUT_MS,
     });
 
-    // Attach error listener so transient socket errors (idle timeouts,
-    // network blips) don't crash the Node process via an unhandled 'error'
-    // event on the ImapFlow EventEmitter.
+    // ImapFlow is long-lived EventEmitter — missing 'error' listener crashes process on socket timeout.
     client.on('error', (error) => {
       this.logger.error(
         `IMAP client error for ${connectedAccount.handle}: ${error.message}`,


davidfarah2003:
Why did you close the PR? @neo773

neo773:
@davidfarah2003 
Sorry about this, we've been dealing mass influx of AI generated spam and have had a few malicious actors who created lot of spam IMAP PRs so I abruptly closed it

neo773 (approved):
LGTM

PR #20511, opened 2026-05-12
Inject none secret env variables into front components

martmull:
## Summary
  - Inject non-secret application variables (`isSecret: false`) into front component `process.env` via the existing Web Worker `setWorkerEnv` mechanism
  - Filter secret variables server-side in the resolver so they never reach the browser
  - Set application variables before system variables (`TWENTY_API_URL`, `TWENTY_APP_ACCESS_TOKEN`) to prevent override
  - Wire up environment variable keys in the logic function code editor for TypeScript autocomplete

  ## Test plan
  - [x] Unit tests for `buildNonSecretEnvVar` (6 passing)
  - [x] Typecheck passes for `twenty-front` and `twenty-server`
  - [x] Install an app with both `isSecret: false` and `isSecret: true` variables, open a front component, verify only non-secret vars appear in `process.env`
  - [x] Open a logic function editor, verify autocomplete suggests declared variable keys


packages/twenty-client-sdk/src/metadata/generated/schema.graphql
```diff
@@ -323,6 +323,7 @@ type FrontComponent {
   isHeadless: Boolean!
   usesSdkClient: Boolean!
   applicationTokenPair: ApplicationTokenPair
+  environmentVariables: JSON
```
charlesBochet: I'm not sure we should pass them as "environmentVariable", I would pass them as "applicationVariables"

packages/twenty-docs/developers/extend/apps/layout/front-components.mdx
```diff
@@ -239,6 +239,37 @@ Available hooks:
...
 
+## Environment variables
+
+Application variables defined in [`defineApplication()`](/developers/extend/apps/config/application) with `isSecret: false` are automatically available as `process.env.VARIABLE_NAME` inside front components:
+
+```tsx src/front-components/greeting.tsx
+import { defineFrontComponent } from 'twenty-sdk/define';
+
+const Greeting = () => {
+  const recipientName = process.env.DEFAULT_RECIPIENT_NAME ?? 'World';
```
charlesBochet: I would not push the user to read process.env directly, i would introduce a nice util to wrap it, this util will be part of the front component api (like others)

martmull: why?

packages/twenty-server/src/engine/metadata-modules/front-component/utils/build-non-secret-env-var.ts
```diff
@@ -0,0 +1,20 @@
+import { type FlatApplicationVariable } from 'src/engine/metadata-modules/flat-application-variable/types/flat-application-variable.type';
+
+export const buildNonSecretEnvVar = (
```
charlesBochet: naming
stripSecretFromApplicationVariables

packages/twenty-server/src/engine/metadata-modules/front-component/front-component.resolver.ts
```diff
@@ -67,9 +71,29 @@ export class FrontComponentResolver {
...
+      ] ?? [];
+
+    const flatApplicationVariables = variableUniversalIdentifiers
+      .map(
+        (universalIdentifier) =>
+          applicationVariableMaps.byUniversalIdentifier[universalIdentifier],
+      )
+      .filter(isDefined);
+
+    const environmentVariables = buildNonSecretEnvVar(flatApplicationVariables);
```
charlesBochet: applicationVariables

packages/twenty-docs/developers/extend/apps/layout/front-components.mdx
charlesBochet: this way you can store them in process.env.applicationVariables under the hood btw and you don't have conflicts.

We should guide the user and process.env is not great IMO as it can contain other things that will conflict

charlesBochet (approved):

PR #20512, opened 2026-05-12
Simplify create-twenty-app command

martmull:
## Simplify `create-twenty-app` for zero-interaction use

  Makes `npx create-twenty-app@latest my-app` a fully non-interactive, single-command experience suitable for automated environments (Codex, Claude plugins).

  ### Changes

  - **Remove all interactive prompts** — app name, display name, description, and scaffold confirmation are now derived from CLI args with sensible defaults. `inquirer` dependency removed
   entirely.
  - **Replace OAuth with API key auth** — use the seeded dev API key (`DEV_API_KEY`) to authenticate against the Docker instance as `tim@apple.dev`, eliminating the browser-based OAuth
  flow.
  - **Docker-first with early validation** — check Docker is installed before scaffolding; if missing, print the install URL and exit. Detect alternative runtimes (Podman, nerdctl).
  - **Parallel image pull** — `docker pull` runs in the background during scaffold + dependency install, saving 10-30s on typical runs.
  - **Always pull latest image** — ensures the dev server is up-to-date on every run.
  - **Stop detecting port 3000** — only check port 2020 (Docker instance).
  - **Update CLI flags** — remove `--skip-local-instance` and `--yes`; add `--skip-docker`.
  - **Update CI workflows and docs** — align e2e workflows, package README, and template README/cd.yml with the new flow.


FelixMalfait:
Great! 

> Replace OAuth with API key auth 

Makes sense! But only for the twenty-app-dev image / docker path ; I think OAuth is still a better default for other remotes (cloud for example, if cloud becomes the new default as per this morning's discussion)

martmull:
> Makes sense! But only for the twenty-app-dev image / docker path ; I think OAuth is still a better default for other remotes (cloud for example, if cloud becomes the new default as per this morning's discussion)

indeed i have updated the behavior

packages/create-twenty-app/src/create-app.command.ts
```diff
@@ -28,39 +31,41 @@ type CreateAppOptions = {
...
       await this.validateDirectory(appDirectory);
 
-      const confirmed = await this.promptScaffoldConfirmation({
-        appName,
-        appDisplayName,
-        appDescription,
-        appDirectory,
-        autoConfirm: options.yes,
-      });
+      const hasRemoteAuth = skipLocalInstance && isDefined(apiUrl);
```
charlesBochet: remove is not correct naming as remote is any environment (including local docker)
I'm also not sure about this logic skipLocalInstance + isDefined(apiUrl), it seems to be only used to compute number of totalSteps, it seems fragile
why not create a stepCounter that would actually compute the step that will be executed rather than hardcoded it to 5 or 4?

packages/create-twenty-app/src/create-app.command.ts
```diff
@@ -104,38 +109,29 @@ export class CreateAppCommand {
 
       console.log('');
 
-      let hasLocalServer = false;
+      let serverUrl: string | undefined;
```
charlesBochet: why having both apiUrl and serverUrl.
I would only keep apiUrl and default it to localhost:2020 ?

packages/create-twenty-app/src/create-app.command.ts
```diff
@@ -280,167 +221,139 @@ export class CreateAppCommand {
...
-    ]);
+  private deriveRemoteName(url: string): string {
+    try {
+      return new URL(url).hostname.replace(/\./g, '-');
+    } catch {
+      return 'remote';
+    }
+  }
 
+  private async authenticateWithOAuth(apiUrl: string): Promise<boolean> {
```
charlesBochet, line 371: ok that's nice to support both. I would also add this as a parameter on the command, authentication-method = 'oAuth | devApiKey'

charlesBochet: I think it's clearer to parametrize as much as possible the command (with default) and to throw if a specific combination is not suitable, rather than try to take decision for the user that are heavy to maintain (computing serverUrl based on if we need to skipLocalInstance or not...)

charlesBochet (commented):
Very nice overall!

packages/create-twenty-app/src/create-app.command.ts
martmull: yes, then i remove skip-local-instance paramter

.github/workflows/ci-create-app-e2e-hello-world.yaml
```diff
@@ -109,7 +109,7 @@ jobs:
           create-twenty-app --version
           mkdir -p /tmp/e2e-test-workspace
           cd /tmp/e2e-test-workspace
-          create-twenty-app test-app --example hello-world --display-name "Test hello-world app" --description "E2E test hello-world app" --skip-local-instance
+          create-twenty-app test-app --example hello-world --display-name "Test hello-world app" --description "E2E test hello-world app" --api-url http://localhost:3000
```
charlesBochet, line 112: 2020 ?

packages/create-twenty-app/src/utils/docker-install.ts
```diff
@@ -0,0 +1,33 @@
+import { execSync } from 'node:child_process';
+
+const DOCKER_DOWNLOAD_URLS: Record<string, string> = {
+  darwin: 'https://docs.docker.com/desktop/setup/install/mac-install/',
+  win32: 'https://docs.docker.com/desktop/setup/install/windows-install/',
+  linux: 'https://docs.docker.com/engine/install/',
+};
+
+export const isDockerInstalled = (): boolean => {
```
charlesBochet, line 9: is is on purpose that this is part of create-twenty-app and not of the sdk that is able to pilot docker?

charlesBochet (approved):

packages/create-twenty-app/src/utils/docker-install.ts
martmull: this was only used here so i moved it

.github/workflows/ci-create-app-e2e-hello-world.yaml
martmull: no, because we want to use the installed instance and not scaffold a new one

PR #20513, opened 2026-05-12
feat(server): migrate all remaining JWT token types to ES256

charlesBochet:
## Summary

Extends the asymmetric signing work from #20467 to cover **every remaining `JwtTokenTypeEnum` value**: `LOGIN`, `WORKSPACE_AGNOSTIC`, `FILE`, `API_KEY`, `APPLICATION_ACCESS`, `APPLICATION_REFRESH`, `APP_OAUTH_STATE`, plus the ACCESS-shaped session token issued by the code interpreter tool.

After this PR, every JWT the server signs is ES256 with a `kid` pointing at the current `core."signingKey"` row, while legacy HS256 tokens (no `kid` header) remain verifiable indefinitely through the existing fallback in `JwtWrapperService.resolveVerificationKey`. No new entity / migration / config: this is a pure routing change on top of the infrastructure that already shipped.

## Why

`#20467` only flipped `ACCESS` and `REFRESH` to ES256. Every other JWT type was still HS256-signed against the global `APP_SECRET`, which kept the original blast radius (a leaked `APP_SECRET` invalidates *every* JWT type forever). Migrating the rest unifies the sign path on rotatable per-server private keys without forcing any token reissue.

### Sign side (8 services)
- `LoginTokenService.generateLoginToken`
- `TransientTokenService.generateTransientToken`
- `WorkspaceAgnosticTokenService.generateWorkspaceAgnosticToken`
- `ApplicationTokenService.signApplicationToken` (`APPLICATION_ACCESS` + `APPLICATION_REFRESH`)
- `ApiKeyService.generateApiKeyToken`
- `FileUrlService.signFileByIdUrl` / `signWorkspaceLogoUrl`
- `ConnectionProviderOAuthFlowService.signState` (`APP_OAUTH_STATE`)
- `CodeInterpreterTool.generateSessionToken`

Each call site swaps `jwtWrapperService.sign(payload, { secret: generateAppSecret(...), ... })` for `await jwtWrapperService.signAsync(payload, { expiresIn, [jwtid] })`. The `generateAppSecret` calls on the sign side are dropped (verifier-side `generateAppSecret` stays in `resolveVerificationKey` for the HS256 fallback).

### Verifier side
- `WorkspaceAgnosticTokenService.validateToken` now goes through `verifyJwtToken` instead of the bespoke `verify({ secret })` path, so new ES256 tokens are accepted while the legacy HS256 fallback inside `resolveVerificationKey` still serves the old shape.
- `JwtWrapperService.sign()` is kept (legacy compat / tests) but is now strictly deprecated — there are no remaining production callers.

### Async ripple (`signFileByIdUrl` was synchronous)
- `FileUrlService.signFileByIdUrl` and `signWorkspaceLogoUrl` are now `async`; the `signUrl` callback used by `getRecordImageIdentifier` is widened to accept `Promise<string | null>`.
- Every direct/indirect caller is updated: admin panel (user lookup + statistics + top workspaces), search service (`computeSearchObjectResults`, `getImageIdentifierValue`), workspace resolver (`logo` resolver, public workspace by domain/id), `WorkspaceMemberTranspiler` (now `async toWorkspaceMemberDto[s]` / `toDeletedWorkspaceMemberDto[s]` / `generateSignedAvatarUrl`), `UserService.loadSignedAvatarUrlsByUserId`, `UserWorkspaceService.castWorkspaceToAvailableWorkspace`, workspace-invitation, approved-access-domain, agent-chat-streaming, agent-message-part resolver, navigation-menu-item record identifier, file-ai-chat / file-core-picture / file-email-attachment / file-workflow / files-field services, rich-text & files-field query result getters, and the code-interpreter tool.

## Backward compatibility

- **Legacy HS256 tokens (no `kid`)** keep verifying via `resolveVerificationKey` → `extractAppSecretBody` → `generateAppSecret` for both `workspaceId`-bearing and `userId`-bearing payloads.
- The `API_KEY` HS256-via-ACCESS-secret fallback (#16504) still kicks in inside `verifyJwtToken` for pre-2025-12-12 API keys.
- No payload shape changes, no DB writes, no env var changes — old tokens issued by `main` continue to authenticate.

### Unit (all green locally — 63/63)
Updated specs for every migrated service to mock `signAsync` instead of `sign` and assert the new option shape:
- `login-token.service.spec.ts`, `transient-token.service.spec.ts`, `workspace-agnostic-token.service.spec.ts`, `application-token.service.spec.ts`, `api-key.service.spec.ts`, `connection-provider-oauth-flow.service.spec.ts`.

### Integration (`jwt-key-rotation.integration-spec.ts`)
- Existing ACCESS coverage (current key, legacy HS256 fallback, rotated-out key, revoked key, unknown kid) is preserved.
- New `it.each` assertion: `REFRESH`, `WORKSPACE_AGNOSTIC`, and `LOGIN` tokens emitted by the real signUp → signUpInNewWorkspace → getAuthTokensFromLoginToken pipeline are ES256 with a `kid` matching the current signing key — proves end-to-end that the migration didn't regress those flows.

## Open question (separate decision)

This PR keeps the legacy HS256 verification fallback **forever**. We may eventually want to sunset it for `API_KEY` once telemetry shows pre-migration tokens are gone, but that's a separate product/security decision and not part of this change.

## Test plan

- [ ] CI green
- [ ] `npx nx lint:diff-with-main twenty-server` passes
- [ ] `npx nx typecheck twenty-server` passes
- [ ] `jwt-key-rotation` integration suite passes (new + existing assertions)
- [ ] Manually verify: signing in issues an ES256 ACCESS / REFRESH token, generating an API key issues an ES256 token with `kid`, signed file URL JWT is ES256 with `kid`
- [ ] Pre-existing HS256 tokens still authenticate (covered by integration test, but worth a manual check with a token from `main`)


packages/twenty-server/test/integration/graphql/suites/auth/jwt-key-rotation/jwt-key-rotation.integration-spec.ts
```diff
@@ -53,8 +72,37 @@ const buildAccessTokenPayload = (payload: AccessTokenJwtPayload) => ({
...
+        }
+      }
+    }
+  `;
+
+  return await makeMetadataAPIRequest(
+    { query: mutation, variables: { appToken } },
+    undefined,
+  );
+};
```
prastoin: please create testing util

packages/twenty-server/test/integration/graphql/suites/auth/jwt-key-rotation/jwt-key-rotation.integration-spec.ts
```diff
@@ -240,4 +301,383 @@ describe('JWT Asymmetric Signing & Key Rotation (integration)', () => {
...
   });
+
+  it.each([
+    ['REFRESH', () => sharedRefreshToken],
+    ['WORKSPACE_AGNOSTIC', () => sharedWorkspaceAgnosticToken],
+    ['LOGIN', () => sharedLoginToken],
+  ])(
+    'signs %s tokens with ES256 + kid pointing at the current signing key',
+    (_label, getToken) => {
+      const decoded = decodeJwtCompleteOrThrow(getToken());
```
prastoin: That's not an integration test assertion by itself
Seems to be testing the before all ?
No need for each testing pattern here

packages/twenty-server/test/integration/graphql/suites/auth/jwt-key-rotation/jwt-key-rotation.integration-spec.ts
```diff
@@ -240,4 +301,383 @@ describe('JWT Asymmetric Signing & Key Rotation (integration)', () => {
...
+
+    expect(decoded.header.alg).toBe('HS256');
+    expect(decoded.header.kid).toBeUndefined();
+
+    const response = await renewTokenMutation(forgedToken);
+
+    expect(response.body.errors).toBeUndefined();
+    expect(
+      isNonEmptyString(
+        response.body.data?.renewToken.tokens.accessOrWorkspaceAgnosticToken
```
prastoin: ***Suggestion:*** I feel like we should have two integration tests suite, one for the legacy coverage and one for the new implem

packages/twenty-server/src/engine/api/common/common-result-getters/handlers/field-handlers/rich-text-field-query-result-getter.handler.ts
```diff
@@ -76,37 +76,39 @@ export class RichTextFieldQueryResultGetterHandler
...
-      return {
-        ...block,
-        props: {
-          ...block.props,
-          url,
-        },
-      };
-    });
+  ): Promise<RichTextBlock[]> => {
+    return Promise.all(
```
prastoin, line 83: ***Remark:*** Previously this could never fail, as it started to become async now it can
We might wanna return a default placeholder for failed to be processed url signature or using `promiseSettled` or make the atomic method non throwable

packages/twenty-server/src/engine/core-modules/file/file-url/file-url.service.ts
```diff
@@ -52,13 +52,7 @@ export class FileUrlService {
...
-    const secret = this.jwtWrapperService.generateAppSecret(
-      payload.type,
-      workspaceId,
-    );
-
-    const token = this.jwtWrapperService.sign(payload, {
-      secret,
+    const token = await this.jwtWrapperService.signAsync(payload, {
       expiresIn: fileTokenExpiresIn,
     });
```
prastoin, line 58: ***Suggestion:*** Imo this method should never throw. We might wanna return either null ro discriminated union `success` | `failure` ?
What do you think ?

prastoin (commented):
partial first review

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-wrapper.service.ts
```diff
@@ -54,32 +47,15 @@ export class JwtWrapperService {
...
     options: { expiresIn: string | number; jwtid?: string },
   ): Promise<string> {
-    if (!isAsymmetricSigningEligible(payload.type)) {
-      throw new AuthException(
-        `signAsync called with non-rotatable token type "${payload.type}"`,
-        AuthExceptionCode.INVALID_JWT_TOKEN_TYPE,
-      );
-    }
-
     const signingKey = await this.jwtKeyManagerService.getCurrentSigningKey();
```
prastoin, line 54: nice

packages/twenty-server/src/engine/core-modules/file/file-url/file-url.service.ts
charlesBochet: I would let it throw but I can rename the method to be clearer: signAsyncOrThrow

packages/twenty-server/src/engine/api/common/common-result-getters/handlers/field-handlers/rich-text-field-query-result-getter.handler.ts
charlesBochet: I think it's fine that it throws, I'll check what the behavior when it throw, it should be a nice error

packages/twenty-server/src/engine/core-modules/user-workspace/user-workspace.service.ts
```diff
@@ -584,20 +584,21 @@ export class UserWorkspaceService extends TypeOrmQueryService<UserWorkspaceEntit
     authProvider: AuthProviderEnum,
   ) {
     return {
-      availableWorkspacesForSignUp:
+      availableWorkspacesForSignUp: await Promise.all(
```
prastoin: ***Nitpick:*** Could start both the `availableWorkspacesForSignUp` and `availableWorkspacesForSignIn` promises at the same time in the same Promise.all

packages/twenty-server/test/integration/graphql/suites/auth/jwt-key-rotation/jwt-key-rotation.integration-spec.ts
charlesBochet: ok !

prastoin (approved):
Strongs work, code way cleaner without the generateAppSecret pattern
Some additional complexity as it introduces a lot of Promise.all

Still approving ! Very nice job

packages/twenty-server/src/engine/core-modules/file/file-url/file-url.service.ts
prastoin: I'm afraid of the side effect it will have on all the caller iterating in a `Promise.all`

prastoin: For example when rendering a block note we should not fail to completely render the note if only one image fails to be signed for some reason ?
Maybe I'm overengineering

packages/twenty-server/test/integration/graphql/suites/auth/jwt-key-rotation/jwt-key-rotation.integration-spec.ts
charlesBochet: done, extracted into dedicated util files

packages/twenty-server/test/integration/graphql/suites/auth/jwt-key-rotation/jwt-key-rotation.integration-spec.ts
charlesBochet: good catch, dropped

packages/twenty-server/test/integration/graphql/suites/auth/jwt-key-rotation/jwt-key-rotation.integration-spec.ts
charlesBochet: kept it throwing for now (signing failure means we have bigger problems), happy to revisit if it bites us

packages/twenty-server/src/engine/core-modules/file/file-url/file-url.service.ts
charlesBochet: renamed to signAsyncOrThrow, on the side-effect side I prefer to keep it simple for now, can be revisited if it becomes a real issue in practice

packages/twenty-server/test/integration/graphql/suites/auth/jwt-key-rotation/jwt-key-rotation.integration-spec.ts
charlesBochet: also split into two suites: jwt-asymmetric-signing + jwt-legacy-hs256-fallback

packages/twenty-server/src/engine/api/common/common-result-getters/handlers/field-handlers/rich-text-field-query-result-getter.handler.ts
charlesBochet: kept throwing - AuthException is properly mapped by the GraphQL error filter so no 500, and if signing breaks we have bigger issues than blocknote rendering

packages/twenty-server/src/engine/core-modules/user-workspace/user-workspace.service.ts
charlesBochet: done

PR #20516, opened 2026-05-13
fix(messaging): register error event handler on ImapFlow client

### LINKED ISSUE #20509: IMAP: unhandled 'error' event on ImapFlow client crashes twenty-server on socket timeout (same class as #20144)

ErnestHysa:
## Summary

Fixes #20509.

**Problem**: The `ImapFlow` client is created in `ImapClientProvider.getClient()` without an `error` event handler. Since `ImapFlow` extends Node.js `EventEmitter`, any socket-level or IMAP-protocol errors that occur after the client is instantiated (but are not connection-level errors that reject from `connect()`) either silently propagate or crash the process unhandled.

**Root cause**: `ImapFlow` is an EventEmitter. Without an `error` listener, throwing an error event causes Node.js to emit an unhandled `error` event, which terminates the process by default.

**Fix**: Added an `error` event handler to the `ImapFlow` client immediately after instantiation, before calling `connect()`:

```typescript
client.on('error', (error: Error) => {
  this.logger.error(
    `IMAP client error for ${connectedAccount.handle}: ${error.message}`,
    error.stack,
  );
});
```

The handler is registered **before** `connect()` so that errors occurring during the connection phase are also captured. The handler logs the full error message and stack trace for operator diagnostics.

Note: `closeClient()` already has a try/catch around `client.logout()`, so cleanup errors are handled separately and won't be affected.

**Tests added** in `imap-client.provider.spec.ts`:
- Verifies the `error` event handler is registered (and registered before `connect()`)
- Regression tests for provider validation (IMAP provider check, missing connection parameters)
- Regression tests for `closeClient()` logout behavior

**Testing**:
1. Connect any IMAP account
2. Simulate a network failure during an IMAP operation (e.g., unplug network, kill IMAP server)
3. Server logs should show: `IMAP client error for user@email.com: <error message>` with stack trace
4. Process should NOT crash — the error is caught and logged, allowing graceful degradation

---
/claim #20509

### DIFF
--- a/packages/twenty-server/src/modules/messaging/message-folder-manager/drivers/imap/services/__tests__/imap-get-all-folders.service.spec.ts
+++ b/packages/twenty-server/src/modules/messaging/message-folder-manager/drivers/imap/services/__tests__/imap-get-all-folders.service.spec.ts
@@ -232,4 +232,56 @@ describe('ImapGetAllFoldersService', () => {
       },
     );
   });
+
+  describe('Gmail namespace handling', () => {
+    it('should store the full hierarchical path as name, not the leaf name', async () => {
+      const mailboxList = [
+        createMockMailbox({
+          path: 'INBOX',
+          name: 'INBOX',
+        }),
+        createMockMailbox({
+          path: '[Gmail]/Sent Mail',
+          name: 'Sent Mail',
+        }),
+        createMockMailbox({
+          path: '[Gmail]/All Mail',
+          name: 'All Mail',
+        }),
+      ];
+
+      mockImapClient.list.mockResolvedValue(mailboxList);
+      mockImapClient.status.mockImplementation(async (path: string) => {
+        const uidMap: Record<string, bigint> = {
+          INBOX: BigInt(1),
+          '[Gmail]/Sent Mail': BigInt(2),
+          '[Gmail]/All Mail': BigInt(3),
+        };
+
+        return { uidValidity: uidMap[path] } as any;
+      });
+
+      imapFindSentFolderService.findSentFolder.mockResolvedValue({
+        path: '[Gmail]/Sent Mail',
+        name: 'Sent Mail',
+      });
+
+      const result = await service.getAllMessageFolders(
+        CONNECTED_ACCOUNT,
+        MESSAGE_CHANNEL,
+      );
+
+      const sentFolder = result.find((f) => f.isSentFolder);
+
+      expect(sentFolder).toBeDefined();
+      expect(sentFolder?.name).toBe('[Gmail]/Sent Mail');
+      expect(sentFolder?.externalId).toBe('[Gmail]/Sent Mail:2');
+
+      const allMailFolder = result.find(
+        (f) => !f.isSentFolder && f.externalId?.includes('All Mail'),
+      );
+
+      expect(allMailFolder?.name).toBe('[Gmail]/All Mail');
+    });
+  });
 });
--- a/packages/twenty-server/src/modules/messaging/message-folder-manager/drivers/imap/services/imap-get-all-folders.service.ts
+++ b/packages/twenty-server/src/modules/messaging/message-folder-manager/drivers/imap/services/imap-get-all-folders.service.ts
@@ -86,7 +86,7 @@ export class ImapGetAllFoldersService implements MessageFolderDriver {
 
       folders.push({
         externalId,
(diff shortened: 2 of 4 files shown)


ErnestHysa:
Thanks @cubic-dev-ai — you're right,  is semantically incorrect here. It always passes because the promise already resolved, regardless of whether  swallows the logout error or not.

Fixed: replaced with , which explicitly asserts that  returns  when logout fails (i.e., it swallows the error and doesn't re-throw). Force-pushed to the branch.

PR #20525, opened 2026-05-13
Fix standard React form event targets in front components

### LINKED ISSUE #20354: React Front component text fields/areas, checkboxes, and dropdown selectors not working at all
## Bug Description

React Front Components rendered inside a page layout widget do not appear to support editable native form controls correctly.

In a custom FRONT_COMPONENT used on a record page layout, button click handlers work, but interactive form fields do not work:

- **Text inputs** render, but typing does not update the field.
- **Textareas** render, but the user cannot edit their content.
- **Checkboxes** render, but clicking does not toggle them.
- **Select/dropdown fields** render, but selecting an option does not update state.
- **Regular buttons** still work, including async actions such as save/render.

This happens even with plain native HTML controls inside the React front component, not only with third-party editors.

**Screenshot:**

Image

## Expected behavior

Native form controls inside React Front Components should behave like normal browser controls:

- `<input>` should accept typed text and dispatch usable input/change events.
- `<textarea>` should allow editing existing text and dispatch usable input/change events.
- `<input type="checkbox">` should toggle and dispatch a usable change event.
- `<select>` should allow option selection and dispatch a usable change event.
- Controlled React state should be able to update from these events.

If Front Components intentionally do not support editable form controls in page layout widgets, the documentation should clearly state that limitation and recommend an alternative pattern.

## Technical inputs

- The component is defined with `defineFrontComponent` from `twenty-sdk/define`.
- It is mounted as a `FRONT_COMPONENT` widget in a `RECORD_PAGE` page layout.
- The issue happens with native controls such as:
```typescript
<input value={draft.name} onChange={(event) => setName(event.target.value)} />

<textarea
  value={source}
  onChange={(event) => setSource(event.target.value)}
/>

<input
  type="checkbox"
  checked={isActive}
  onChange={(event) => setIsActive(event.target.checked)}
/>

<select value={kind} onChange={(event) => setKind(event.target.value)}>
  <option value="INVOICE">Invoice</option>
  <option value="CONTRACT">Contract</option>
</select>

### PR DESCRIPTION
**Pull request by Qin-long**
Fixes #20354

## Problem

Front component form events currently expose serialized form state through a sandbox-specific event shape, such as `event.detail.value` and `event.detail.checked`.

That works for examples that explicitly read `event.detail`, but it is surprising for app authors writing standard React form handlers:

```tsx
onChange={(event) => {
  setValue(event.target.value);
}}
Internal app code already has to defend against multiple possible shapes:

// Values may live on e.detail.value, e.value, or e.target.value.
This suggests the sandbox event shape is leaking into userland.

Solution
This change keeps the existing event.detail behavior, but also syncs serialized event target properties back onto the remote element before dispatching the event.

That means both styles work:

// Existing sandbox-specific style
event.detail.value;

// Standard React style
event.target.value;
The same applies to checked, files, scroll/media target properties, and similar serialized target state.

What Changed
Added a shared helper to apply serialized event target properties onto the remote element.
Updated generated remote element event configs to dispatch serialized events through a custom event config.
Updated the remote-dom element generator so regenerated files preserve this behavior.
Updated Storybook form-event examples to use standard React event target reads.
Added/updated Storybook coverage for input, checkbox, textarea, select, submit, and caret preservation flows.
Validation
Ran git diff --check
Ran a targeted TypeScript error scan for the changed front component renderer files
Manually verified the Storybook FrontComponent/EventForwarding form event story locally:
text input updates state
checkbox updates state
submit reflects the updated JSON
Note: local Storybook verification on Windows required temporary local build/cache fixes that are not included in this PR, to keep this change focused on front component event behavior.

### CODE REVIEW

**Comment by charlesBochet** [comment]
I think this actually makes sense, I'm reviewing it but it might take time because I need a full setup to reproduce

**Comment by Qin-long** [comment]
> 我觉得这确实有道理，我正在审查，但可能需要一些时间，因为我需要一套完整的系统来重现这个问题。

Thanks for taking a look!

No rush at all. For reproduction, the important bit is not the existing Storybook demo behavior, because the old demo already worked by reading from `event.detail`.

The issue shows up when a front component author uses the standard React form handler shape:

```tsx
<input
  value={value}
  onChange={(event) => setValue(event.target.value)}
/>

Before this change, serialized form values were available on event.detail.value / event.detail.checked, but not reliably through event.target.value / event.target.checked.

This PR keeps the existing event.detail behavior intact and also applies the serialized target fields back onto the remote element before dispatching the event, so both styles work.

I also updated the Storybook form event example/tests to cover the standard React-style access pattern:

text input: event.target.value
checkbox: event.target.checked
textarea/select value reads
submit payload
caret preservation while editing in the middle of the input
I was able to verify the Storybook scenario locally. On Windows I did hit unrelated local build/static-dir issues while setting up Storybook, but those setup fixes are intentionally not included in this PR to keep the change focused.

**Review by charlesBochet — APPROVED** [review]

**Comment by prastoin** [comment]
Reviewing

File: packages/twenty-front-component-renderer/src/constants/SerializedEventData.ts [hunk 1]
```diff
@@ -43,3 +43,56 @@ export type SerializedEventData = {
   playbackRate?: number;
   files?: SerializedFileData[];
 };
+
+export const applySerializedEventTargetProperties = (
+  element: Record<string, unknown>,
+  eventData: SerializedEventData,
+): void => {
+  if ('value' in eventData) {
```
prastoin, line 51: ***Suggestion:*** `in` fully browse all key and nested classes prototype methods etc
whereas hasOwnProperty will only search for key properties
```ts
Object.prototype.hasOwnProperty.call(eventData, 'value')
```

Not critical really applicable neither critical here though

prastoin: Tbh I would basically replace this with a zod schema
Zod will always ( unless configured differently ) create a subset, stripping unknown keys out of an unknown set

prastoin: not a fan of a mutation neither here as theorically this could contain more properties than expected and they would be persisted

packages/twenty-front-component-renderer/src/remote/generated/remote-elements.ts
```diff
@@ -62,6 +67,26 @@ const HTML_COMMON_EVENTS_ARRAY = [
...
+): RemoteElementEventListenerDefinition => ({
+  dispatchEvent(this: Element, eventData: SerializedEventData) {
+    applySerializedEventTargetProperties(
+      this as unknown as Record<string, unknown>,
+      eventData,
+    );
+
+    return new CustomEvent(eventType, {
+      detail: eventData,
+    }) as RemoteEvent<SerializedEventData>;
```
prastoin, line 81:
```ts
    const serializedEvent = applySerializedEventTargetProperties(
      this as unknown as Record<string, unknown>,
      eventData,
    );

    return new CustomEvent(eventType, {
      detail: serializedEvent,
    }) as RemoteEvent<SerializedEventData>;
```

prastoin:
Double checking my initial hunch, as I'm unsure of the expected behavior here

prastoin:
Yeah my concerns are irrelevant
The whole goal is to make a safe whitelisted mutation of the given element

prastoin (approved):
Haven't check the tests

PR #20526, opened 2026-05-13
feat(ai-chat) - Stop ai thinking if credits exhausted

etiennejouan:
Billing is now decremented per-step, not per-turn. The onStepFinish callback in chat-execution.service.ts calls a new decrementAndCheckAvailableCredits method on each model step, so Redis is debited incrementally as the agent runs rather than all at once at the end.

Credit exhaustion stops the stream mid-run. When a step depletes the remaining credits, a hasNoMoreAvailableCredits flag is set and passed into the stopWhen predicate of streamText, causing the agent to halt before starting the next step.

A new credits-exhausted event is introduced. After the stream drains and the response is persisted, if credits ran out the job publishes a dedicated credits-exhausted event to the frontend instead of the normal message-persisted event.

The frontend handles this new event. useAgentChatSubscription has a new credits-exhausted case that sets a BILLING_CREDITS_EXHAUSTED-coded error on the atom, closes the writer, and stops the streaming state — triggering the existing AiChatCreditsExhaustedMessage UI.


packages/twenty-front/src/modules/information-banner/components/billing/InformationBannerEndTrialPeriod.tsx
```diff
@@ -18,8 +18,8 @@ export const InformationBannerEndTrialPeriod = () => {
       variant="secondary"
       message={
         hasPermissionToEndTrialPeriod
-          ? t`End trial period to continue using Workflow or AI features.`
-          : t`Contact your admin to continue using Workflow or AI features.`
+          ? t`Free trial credits exhausted. End trial period to continue using Workflow or AI features.`
```
FelixMalfait: nit: I remember this wasn't displaying very nicely on small screen and that's why we had to crop in the past, mut might not be accurate anymore

packages/twenty-server/src/engine/core-modules/billing/services/billing-usage.service.ts
FelixMalfait: nit: I feel like DX/naming is a bit unclear. I understand there is a temporary storage but that is never automatically persisted to clickhouse, but it's not clear from the service/method name

packages/twenty-server/src/engine/metadata-modules/ai/ai-chat/services/chat-execution.service.ts
```diff
@@ -303,11 +310,26 @@ export class ChatExecutionService {
...
+      const costInDollars = this.aiBillingService.calculateCost(
         registeredModel.modelId,
         { usage, cacheCreationTokens },
+      );
+      const creditsUsedMicro = Math.round(
+        convertDollarsToBillingCredits(costInDollars),
+      );
+
+      // Redis was already decremented per-step via onStepFinish; only emit the
+      // aggregated ClickHouse event here (one row per turn).
```
FelixMalfait: Maybe this comment can be removed if naming of service/function is more clear?

packages/twenty-server/src/engine/metadata-modules/ai/ai-chat/jobs/stream-agent-chat.job.ts
```diff
@@ -237,6 +238,8 @@ export class StreamAgentChatJob {
               conversationSizeTokens: data.conversationSizeTokens,
             });
```
FelixMalfait, line 240: Do we have the same issue for workflow agents which also run in a loop?

FelixMalfait (approved):
Great!

packages/twenty-server/src/engine/metadata-modules/ai/ai-chat/services/chat-execution.service.ts
etiennejouan: 🤢 missed comment..

I've renamed all functions with "cache" explicit kind of suffix. 
The whole service need clarification and to be independent from billing. It will be done with the [multi cap feature](https://github.com/twentyhq/core-team-issues/issues/2392).

packages/twenty-server/src/engine/metadata-modules/ai/ai-chat/jobs/stream-agent-chat.job.ts
etiennejouan: Screenshot 2026-05-13 at 15 30 27
good catch ! fixed

PR #20527, opened 2026-05-13
feat(twenty-server): one-hop relation filters in GraphQL API

FelixMalfait:
## Summary

Adds support for filtering records by fields on a related MANY_TO_ONE object via the GraphQL API. Backend only — no frontend, no REST, no view-filter persistence yet.

```graphql
{
  people(filter: { company: { name: { like: "%Airbnb%" } } }) {
    edges { node { id } }
  }
}
```

### Where the work lands

- **Schema** — `relation-field-metadata-gql-type.generator.ts` now emits `{relationName}: TargetFilterInput` alongside the existing `{joinColumnName}: UUIDFilter` for MANY_TO_ONE relations. Mirrors the order-by generator that already does this for sort. Lazy thunks in `object-metadata-filter-gql-input-type.generator.ts` handle the cycle between filter inputs.
- **Arg processor** — `FilterArgProcessorService` no longer hard-rejects accessing a relation by its name. When the value is a nested object on a MANY_TO_ONE field, it recurses into the target object's metadata so each leaf still gets validated and coerced. Depth-capped at 1.
- **Query parser** — new `parseRelationSubFilter` branch in `graphql-query-filter-field.parser.ts`. When triggered: looks up the target object metadata, calls `ensureRelationJoin` against the outer query builder, and recurses via a child `GraphqlQueryFilterConditionParser` scoped to the target. `and`/`or`/`not` inside the relation filter keep working because the child dispatches through the same `parseKeyFilter`.
- **Shared join utility** — `ensureRelationJoin.util.ts` is a single function that inspects `queryBuilder.expressionMap.joinAttributes` for the alias before adding a `LEFT JOIN`. Rewired the existing inline `qb.leftJoin` calls in the order parser and group-by service to use it, so filter-driven joins no longer collide with sort-driven joins on the same relation.

### Out of scope (explicit)

- ONE_TO_MANY reverse traversal (needs EXISTS subqueries)
- Aggregates (`company.people.count > 5` — needs HAVING)
- View-filter storage (no `relationPath` column on `ViewFilterEntity`)
- REST DSL changes
- Frontend filter-picker UX
- Nesting deeper than one hop (parser and arg-processor both reject)

### Open question for review

Permissions. The order-by-on-relation code path already lets users sort People by Company.name without a Company read-permission check, and this PR matches that behavior for filters — felt wrong to add a stricter gate only on the filter side. If we want object-permission gating on the relation target, it should be a follow-up that covers both paths consistently. The only attack surface today is existence inference via timing, identical to what sort already exposes.

## Test plan

- [x] `tsc --noEmit` — clean for changed files (5 unrelated pre-existing errors on main untouched)
- [x] `oxlint --type-aware` + `prettier --check` — 0 errors on all 17 changed/new files
- [x] `jest filter-arg-processor.service.spec` — 229 tests pass (the new optional `flatObjectMetadataMaps` arg is backwards-compatible)
- [x] Integration test (`filter-by-relation-field.integration-spec.ts`, 6 cases) — needs to be verified against a seeded test DB. Could not exercise the happy path in my isolated worktree; depth-2 rejection passed there.
- [ ] EXPLAIN ANALYZE on the integration test query to confirm the FK on `person.companyId` is indexed for both standard and custom MANY_TO_ONE relations.

### Integration test cases

1. Filter People by `company.name = "Airbnb"` (exact match)
2. Filter People by `company.name like "%irbnb%"`
3. Non-matching filter returns empty
4. Combined with a scalar filter at root via `and`
5. **Combined with `orderBy` on the same relation** — proves the join-dedupe works (without `ensureRelationJoin`, TypeORM throws "duplicate alias")
6. Depth-2 nesting (`company.accountOwner.name`) returns `INVALID_ARGS_FILTER`


packages/twenty-server/src/engine/api/common/common-args-processors/filter-arg-processor/filter-arg-processor.service.ts
```diff
@@ -26,15 +26,23 @@ import { buildFieldMapsFromFlatObjectMetadata } from 'src/engine/metadata-module
 import { isFlatFieldMetadataOfType } from 'src/engine/metadata-modules/flat-field-metadata/utils/is-flat-field-metadata-of-type.util';
 import { FlatObjectMetadata } from 'src/engine/metadata-modules/flat-object-metadata/types/flat-object-metadata.type';
 
+// Cap relation-traversal nesting depth. The schema allows arbitrary depth
+// because filter inputs reference each other lazily, but the query builder
+// can only realistically handle a single hop today — joins are added against
+// the root alias only. Bump this when reverse joins / multi-hop are wired up.
```
FelixMalfait: Remove comment

packages/twenty-server/src/engine/api/common/common-args-processors/filter-arg-processor/filter-arg-processor.service.ts
```diff
@@ -152,6 +172,54 @@ export class FilterArgProcessorService {
...
+
+          return this.validateAndTransformFilter(
+            filterValue as unknown as ObjectRecordFilter,
+            targetObjectMetadata,
+            flatObjectMetadataMaps,
+            flatFieldMetadataMaps,
+            targetFieldIdByName,
+            targetFieldIdByJoinColumnName,
+            depth + 1,
+          ) as unknown as Record<string, unknown>;
```
FelixMalfait: is it possible to avoid as unkown as and do a better typing all the way?

packages/twenty-server/src/engine/api/common/common-args-processors/filter-arg-processor/filter-arg-processor.service.ts
```diff
@@ -152,6 +172,54 @@ export class FilterArgProcessorService {
...
+      // (e.g. `{ company: { name: { like: "%X%" } } }`). Recurse into the
+      // target object's metadata so each leaf still gets validated and
+      // value-coerced. Only MANY_TO_ONE is supported for now; ONE_TO_MANY
+      // would need EXISTS-subquery support downstream.
+      if (
+        fieldMetadata.settings?.relationType === RelationType.MANY_TO_ONE &&
+        isDefined(flatObjectMetadataMaps) &&
+        isDefined(fieldMetadata.relationTargetObjectMetadataId)
+      ) {
+        if (depth >= MAX_RELATION_FILTER_DEPTH) {
```
FelixMalfait: Can we add a test with composite fields? e.g. I think we have an ARR field on company. I want to make sure there's no confusion on depth when it's a composite

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-filter/graphql-query-filter-field.parser.ts
```diff
@@ -1,33 +1,54 @@
...
+
+import { GraphqlQueryFilterConditionParser } from './graphql-query-filter-condition.parser';
 
 const ARRAY_OPERATORS = ['in', 'contains', 'notContains'];
 
+// Cap relation-traversal nesting at the parser layer too. The arg processor
+// rejects this earlier with a user-facing message; this guard exists so the
+// query builder never gets invoked with a shape it can't translate (e.g. if
+// the parser is called from a non-validated code path).
+const MAX_RELATION_FILTER_DEPTH = 1;
```
FelixMalfait: Should this be a shared const (1 file = 1 const) instead of replicating

packages/twenty-server/src/engine/api/common/common-args-processors/filter-arg-processor/filter-arg-processor.service.ts
```diff
@@ -26,15 +27,64 @@ import { buildFieldMapsFromFlatObjectMetadata } from 'src/engine/metadata-module
...
+      userFriendlyMessage: msg`Invalid filter: use "${joinColumnName}" to filter by this relation field`,
+    },
+  );
+}
+
+// A relation filter is treated as a traversal (`{ name: { eq: "X" } }`) when
+// at least one top-level key matches a field on the target object or is a
+// logical operator (`and`/`or`/`not`). Otherwise the value looks like a leaf
+// operator shape (`{ eq: "uuid" }`), which is invalid for relation traversal.
+const isRelationTraversalShape = ({
```
FelixMalfait: This feels hacky, let's explore other options? is it a pre-existing bug - I'm thinking for example if someone names a field "and", "no" or "or"? (maybe the 3 are in the list of forbidden names already, check)

packages/twenty-server/src/engine/api/common/common-args-processors/filter-arg-processor/filter-arg-processor.service.ts
```diff
@@ -143,29 +353,18 @@ export class FilterArgProcessorService {
       );
     }
 
+    // ONE_TO_MANY (and other non-MANY_TO_ONE relation types) accessed by name
```
FelixMalfait: Check all comments and only keep the useful one. Comments that stay should refer to an absolute truth not one relative to this PR (e.g. here you say "the legacy message" which can only be understood in the PR's context)

packages/twenty-server/src/engine/api/common/common-args-processors/filter-arg-processor/filter-arg-processor.service.ts
FelixMalfait: @claude wdyt?

packages/twenty-server/src/engine/api/common/common-args-processors/filter-arg-processor/filter-arg-processor.service.ts
```diff
@@ -26,15 +27,64 @@ import { buildFieldMapsFromFlatObjectMetadata } from 'src/engine/metadata-module
...
+const isRelationTraversalShape = ({
+  filterValue,
+  targetFieldIdByName,
+  targetFieldIdByJoinColumnName,
+}: {
+  filterValue: ObjectRecordFilter;
+  targetFieldIdByName: Record<string, string>;
+  targetFieldIdByJoinColumnName: Record<string, string>;
+}): boolean => {
+  if (typeof filterValue !== 'object' || filterValue === null) {
```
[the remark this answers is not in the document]
FelixMalfait: Yes see comment I made, agree I will look if we can do better

packages/twenty-server/src/engine/api/common/common-args-processors/filter-arg-processor/constants/all-leaf-filter-operator-names.constant.ts
```diff
@@ -0,0 +1,23 @@
+import { type FilterOperator } from 'src/engine/api/common/common-args-processors/filter-arg-processor/types/filter-operator.type';
+
+// Every leaf (per-field) filter operator that a value object's keys can be.
+// Logical operators `and`/`or`/`not` are intentionally excluded — they
+// combine sub-filters, they don't terminate one.
+export const ALL_LEAF_FILTER_OPERATOR_NAMES: readonly FilterOperator[] = [
```
FelixMalfait: I'm surprised this has to be redeclared and a list of filter operators like that doesn't already exist (for typing?). Is there an existing one we can extend? Or should this one be reused to improve overall typing in the codebase? @claude

packages/twenty-server/src/engine/api/common/common-args-processors/filter-arg-processor/filter-arg-processor.service.ts
```diff
@@ -26,15 +28,52 @@ import { buildFieldMapsFromFlatObjectMetadata } from 'src/engine/metadata-module
...
+  }
+
+  const keys = Object.keys(filterValue);
+
+  if (keys.length === 0) {
+    return false;
+  }
+
+  return keys.some((k) =>
+    (ALL_LEAF_FILTER_OPERATOR_NAMES as readonly string[]).includes(k),
```
FelixMalfait: As I understand it we should forbid all leaf operator names from being field names? Is that the case already? @claude

packages/twenty-server/src/engine/api/common/common-args-processors/filter-arg-processor/filter-arg-processor.service.ts
```diff
@@ -104,6 +173,123 @@ export class FilterArgProcessorService {
...
+      throw new CommonQueryRunnerException(
+        `Cannot filter by relation field "${key}"`,
+        CommonQueryRunnerExceptionCode.INVALID_ARGS_FILTER,
+        {
+          userFriendlyMessage: msg`Invalid filter: filtering by relation field "${key}" is not supported`,
+        },
+      );
+    }
+
+    if (isLeafOperatorShape(filterValue)) {
```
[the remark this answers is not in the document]
FelixMalfait: Good catch — fixed in d4a63a5 but with a slightly tighter shape than the suggestion.

Switched the relation-traversal handler to:

```ts
if (typeof filterValue !== "object" || filterValue === null) {
  throwUseJoinColumnInstead(key);
}
```

The `Array.isArray` branch in the suggestion is unreachable here because the upstream GraphQL/REST validators reject arrays before this code runs (the only place arrays appear in filters is inside `and` / `or` / `not`, which are handled by the boolean operator branch above). And the `isLeafOperatorShape` check was dropped entirely — see the reply on the operator-names constant for the reasoning.

packages/twenty-server/src/engine/api/common/common-args-processors/filter-arg-processor/constants/all-leaf-filter-operator-names.constant.ts
FelixMalfait: Resolved by dropping the constant entirely in d4a63a5.

You were right that this list felt redundant — and the deeper question was whether the check that used it was needed at all. Walking through the operator-as-field-name scenario again:

- If someone names a relation field `eq`, the filter `{ eq: { eq: "value" } }` would be ambiguous, but in practice the existing scalar/composite path already gives a sensible error here too: composite types like `fullName` would hit "doesn't have any X field" without any operator-name guard. Relations should behave the same way.
- The recursion now lets `{ relationField: { eq: "x" } }` fall through to the target-object lookup, where it naturally fails with `Object <target> doesn't have any "eq" field` — same error shape as the composite case.

So instead of introducing a new constant and a new error class, the simpler fix was to delete the special check and let the existing field-resolution error surface. That's what d4a63a5 does. (Snapshot for the integration test was updated to the new natural message.)

packages/twenty-server/src/engine/api/common/common-args-processors/filter-arg-processor/filter-arg-processor.service.ts
FelixMalfait: Stepping back from this in d4a63a5 — the check is gone, and the question itself turns out to be moot. Two reasons:

**1. Composite fields already coexist with operator names without a guard.**

If a composite sub-field were named `eq`, the recursion into the composite path would just try to resolve `eq` against the composite's allowed sub-fields and surface `Object <composite> doesn't have any "eq" field`. The codebase has never reserved operator names against composite sub-fields, so reserving them only for relation targets would be an inconsistent rule.

**2. The forbidden-names list already protects field creation, not filter parsing.**

Workspace field names go through `validateFieldNameAvailabilityOrThrow` / `BANNED_FIELD_NAMES` at creation time. That's where naming collisions should be prevented (and we can revisit that list separately if needed). The filter-arg processor shouldn't have to know about it — its job is to surface a sensible error if the metadata it's given somehow has a collision, and `Object <target> doesn't have any "eq" field` is already that error.

So d4a63a5 just removes the special case and lets the normal field-resolution path run for relation targets, matching how composites already behave.

packages/twenty-server/test/integration/graphql/suites/filter-by-relation-field.integration-spec.ts
```diff
@@ -0,0 +1,363 @@
...
+      (edge: { node: { company: { name: string } | null } }) =>
+        edge.node.company?.name ?? null,
+    );
+
+    // Airbnb, Notion, Stripe all contain "i"; ascending order
+    expect(companyNames).toEqual(['Airbnb', 'Airbnb', 'Notion', 'Stripe']);
+  });
+
+  it('should filter on a composite sub-field of the related object without tripping the depth cap', async () => {
+    // Person -> Company is one relation hop. annualRecurringRevenue is a
```
charlesBochet: remove most of the comments that are not helping

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
```diff
@@ -280,19 +281,12 @@ export class GroupByWithRecordsService {
...
-        );
-
         for (const joinInfo of relationJoins) {
-          if (!existingJoinAliases.has(joinInfo.joinAlias)) {
-            queryBuilder.leftJoin(
-              `${flatObjectMetadata.nameSingular}.${joinInfo.joinAlias}`,
-              joinInfo.joinAlias,
-            );
-          }
+          ensureRelationJoin(
```
charlesBochet: pass object parameter to utils
naming: ensureRelationJoin => addRelationJoinAliasToQueryBuilder

packages/twenty-server/src/engine/api/graphql/workspace-schema-builder/graphql-type-generators/input-types/relation-field-metadata-gql-type.generator.ts
```diff
@@ -78,16 +78,19 @@ export class RelationFieldMetadataGqlInputTypeGenerator {
   public generateSimpleRelationFieldFilterInputType({
     fieldMetadata,
     typeOptions,
+    context,
   }: {
     fieldMetadata: FlatFieldMetadata<
       FieldMetadataType.RELATION | FieldMetadataType.MORPH_RELATION
     >;
     typeOptions: { settings?: FlatFieldMetadata['settings'] };
+    context?: SchemaGenerationContext;
```
charlesBochet: why is context optional?

packages/twenty-server/src/engine/api/graphql/workspace-schema-builder/graphql-type-generators/input-types/relation-field-metadata-gql-type.generator.ts
```diff
@@ -104,12 +107,49 @@ export class RelationFieldMetadataGqlInputTypeGenerator {
...
+        const targetFilterInputTypeKey = computeObjectMetadataInputTypeKey(
+          targetObjectMetadata.nameSingular,
+          GqlInputTypeDefinitionKind.Filter,
+        );
+
+        const targetFilterInputType = this.gqlTypesStorage.getGqlTypeByKey(
+          targetFilterInputTypeKey,
+        );
+
+        if (
```
charlesBochet: 3 nested if, prefer early return

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/utils/ensure-relation-join.util.ts
```diff
@@ -0,0 +1,25 @@
+import { type ObjectLiteral } from 'typeorm';
+
+import { type WorkspaceSelectQueryBuilder } from 'src/engine/twenty-orm/repository/workspace-select-query-builder';
+
+// Adds a LEFT JOIN for a relation, but only if the same alias isn't already
```
charlesBochet: remove comment and improve funciton naming instead (see my other comment) :)

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-filter/graphql-query-filter-field.parser.ts
```diff
@@ -97,6 +134,82 @@ export class GraphqlQueryFilterFieldParser {
     }
   }
 
+  private parseRelationSubFilter(
+    queryBuilder: WhereExpressionBuilder,
+    outerQueryBuilder: WorkspaceSelectQueryBuilder<ObjectLiteral>,
+    parentAlias: string,
+    fieldMetadata: FlatFieldMetadata,
+    // oxlint-disable-next-line @typescripttypescript/no-explicit-any
+    filterValue: any,
```
charlesBochet: unknown?

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-filter/graphql-query-filter-field.parser.ts
```diff
@@ -40,13 +56,19 @@ export class GraphqlQueryFilterFieldParser {
...
     // oxlint-disable-next-line @typescripttypescript/no-explicit-any
     filterValue: any,
     isFirst = false,
     useDirectTableReference = false,
   ): void {
+    // `fieldIdByName` resolves relation fields by their entity name
+    // (`company`); `fieldIdByJoinColumnName` resolves them by their FK column
+    // (`companyId`). The two have different semantics — only the former opts
+    // into relation traversal.
+    const isAccessedByRelationName = isDefined(this.fieldIdByName[key]);
```
charlesBochet: this naming is not clear to me. it might be if we were in a relation specific context but here it's not specific to relations.

isFilterKeyARelation
isFilterKeyARelationJoinColumn

would be clearer to me

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-filter/graphql-query-filter-condition.parser.ts
```diff
@@ -16,36 +21,59 @@ export class GraphqlQueryFilterConditionParser {
...
+        this.populateBrackets(qb, queryBuilder, objectNameSingular, filter);
       }),
     );
   }
 
+  // Walks the entries of a filter object into an already-opened Brackets.
+  // Public so the field parser can recurse for relation traversal — the
+  // outer query builder is threaded through so JOINs can still be added
+  // even though we're inside a Brackets closure.
+  public populateBrackets(
```
charlesBochet: naming if public, the naming should be more specific? I don't get what it does

packages/twenty-server/test/integration/graphql/suites/filter-by-relation-field.integration-spec.ts
```diff
@@ -0,0 +1,363 @@
...
+  AIRBNB_ENGINEER: '20202020-dddd-4000-8000-000000000001',
+  AIRBNB_DESIGNER: '20202020-dddd-4000-8000-000000000002',
+  STRIPE_ENGINEER: '20202020-dddd-4000-8000-000000000003',
+  NOTION_ENGINEER: '20202020-dddd-4000-8000-000000000004',
+  UNAFFILIATED: '20202020-dddd-4000-8000-000000000005',
+};
+
+const ALL_TEST_PERSON_IDS = Object.values(TEST_PERSON_IDS);
+
+describe('Filter by relation field (e2e)', () => {
```
charlesBochet, line 33: let's also add a test for a morph case

charlesBochet (changes requested):
Bravo!!

packages/twenty-server/test/integration/graphql/suites/filter-by-relation-field.integration-spec.ts
FelixMalfait: Trimmed the narrative comments in 21e91e66af — kept only the ones that justify a non-obvious expected value (the "Airbnb/Notion/Stripe all contain i" line and the ARR threshold one).

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/group-by/services/group-by-with-records.service.ts
FelixMalfait: Done in 21e91e66af — `ensureRelationJoin` → `addRelationJoinAliasToQueryBuilder({ queryBuilder, parentAlias, relationName })` with the file renamed to match. Updated all 3 callers (graphql-query.parser, graphql-query-filter-field.parser, group-by-with-records.service).

packages/twenty-server/src/engine/api/graphql/workspace-schema-builder/graphql-type-generators/input-types/relation-field-metadata-gql-type.generator.ts
FelixMalfait: Made it required in 21e91e66af. Every callsite already passed it — pushed the requirement through the schema-builder chain (filter, order-by base, group-by generators) so the optionality is gone end-to-end.

packages/twenty-server/src/engine/api/graphql/workspace-schema-builder/graphql-type-generators/input-types/relation-field-metadata-gql-type.generator.ts
FelixMalfait: Flattened in 21e91e66af — each of the three generate…InputType methods now uses early returns for the three guards (no relationTargetObjectMetadataId / no targetObjectMetadata / no targetInputType), so the "build the field" path stays at the top level.

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/utils/ensure-relation-join.util.ts
FelixMalfait: Comment removed in 21e91e66af. The function got renamed to `addRelationJoinAliasToQueryBuilder` per your other comment — the idempotent-add behavior is left to be read off the body.

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-filter/graphql-query-filter-field.parser.ts
FelixMalfait: Typed it in 21e91e66af as `Partial<ObjectRecordFilter>` rather than `unknown` — that is what `populateBrackets` already required downstream, so it removes a cast rather than adding narrowing in this function.

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-filter/graphql-query-filter-field.parser.ts
FelixMalfait: Renamed to `isFilterKeyARelation` in 21e91e66af and dropped the explainer comment. The companion variable `fieldIdByJoinColumnName` already conveys the "JOIN column" path at the resolver level, so we kept just the single boolean here.

packages/twenty-server/src/engine/api/graphql/graphql-query-runner/graphql-query-parsers/graphql-query-filter/graphql-query-filter-condition.parser.ts
FelixMalfait: Renamed to `applyFilterEntriesToWhereBrackets` in 21e91e66af — name now says the WHERE-brackets context (it is the recursive form of `parse`, walking filter entries into an already-opened `Brackets`). Updated the single caller in the field parser too.

packages/twenty-server/test/integration/graphql/suites/filter-by-relation-field.integration-spec.ts
FelixMalfait: Added in 21e91e66af — new test `should filter pets by a MORPH_RELATION target field (rocket name)` which sets up a Pet with `polymorphicOwnerRocketId` and filters by `polymorphicOwnerRocket.name`. Exercises the same code path as RELATION (the field parser branches on `isMorphOrRelationFlatFieldMetadata`).

packages/twenty-server/src/engine/api/common/common-query-runners/utils/build-mutation-query-builder.util.ts
```diff
@@ -0,0 +1,62 @@
...
+  const filteredQueryBuilder = repository.createQueryBuilder(alias);
+
+  commonQueryParser.applyFilterToBuilder(filteredQueryBuilder, alias, filter);
+
+  const hasRelationTraversal =
+    filteredQueryBuilder.expressionMap.joinAttributes.length > 0;
+
+  if (!hasRelationTraversal) {
+    return filteredQueryBuilder;
+  }
```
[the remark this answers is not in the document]
FelixMalfait: Checked this — the no-traversal path does not need `.withDeleted()`. TypeORM auto-injects `deletedAt IS NULL` only when `queryType === "select"` at SQL-render time (see `QueryBuilder.createWhereExpression`, the `queryType === "select" && !withDeleted && metadata.deleteDateColumn` branch). The caller does `.restore()` on the returned builder before `.execute()`, and `restore()` flips `queryType` to `"restore"` on the cloned expressionMap before rendering — so the auto-filter is skipped without us touching `withDeleted`.

The traversal path is the asymmetric case: its inner builder is materialized via `getQuery()` while it is still `"select"` (it never gets morphed — only the outer builder does), which is why that branch needs `.withDeleted()`.

The existing `restore-many-object-records-permissions.integration-spec.ts` already exercises the scalar `restoreMany` against previously soft-deleted records and passes — both pre-PR and post-PR with this helper. So no fix needed.

packages/twenty-server/test/integration/graphql/suites/mutate-by-relation-field.integration-spec.ts
```diff
@@ -0,0 +1,350 @@
...
+
+    const findResponse = await makeGraphqlAPIRequest(findOperation);
+
+    expect(findResponse.body.data.people.edges).toEqual([]);
+  });
+
+  it('should keep scalar-only mutations working unchanged', async () => {
+    // No relation traversal — the builder helper should keep emitting the
+    // existing direct-WHERE SQL (no IN-subquery wrap). Behavior must match
+    // what the pre-PR code path produced.
```
FelixMalfait: Remove all comments that reference previous state in this PR. Only keep comments that will be useful 1 year from now and can be interepreted not relatively to this PR

PR #20528, opened 2026-05-13
feat(twenty-server): introduce ENCRYPTION_KEY env var with versioned envelope

charlesBochet:
## Summary

- Adds `ENCRYPTION_KEY` (primary) and `FALLBACK_ENCRYPTION_KEY` (decrypt-only fallback for rotation) env vars to twenty-server, with backward-compatible fallback to `APP_SECRET` when `ENCRYPTION_KEY` is unset.
- Introduces a versioned ciphertext envelope `enc:v2:<keyId>:<base64>` using AES-256-GCM with HKDF-SHA256 derived per-context keys. The 8-hex `keyId` fingerprint lets every row identify which physical key encrypted it, so rotation routes directly to primary or fallback without trial decryption; GCM's auth tag gives true integrity (legacy CTR has none).
- Migrates `ConnectedAccountTokenEncryptionService` to the new envelope and plumbs `workspaceId` through every caller, so per-workspace HKDF context binds each row to its tenant.

The remaining encryption sites (`jwt-key-manager`, `config-storage`, `postgres-credentials`, `application-variable`, TOTP) stay on the legacy unprefixed CTR path and will be migrated in follow-up PRs. The operator-facing rotation runbook is out of scope here.

### Format details

`enc:v{N}:{keyId}:{base64}` — `N=2` is the only version produced by new writes (`v1` exists for backward-compatible decryption of existing connected-account rows). `keyId = sha256(rawKey).slice(0,4).toString('hex')`. The CHECK constraint on `core.connectedAccount.{accessToken,refreshToken}` is relaxed from `LIKE 'enc:v1:%'` to `LIKE 'enc:v_:%'` so both versions pass.

### Key resolution

| `ENCRYPTION_KEY` | `FALLBACK_ENCRYPTION_KEY` | `APP_SECRET` | Encrypt with | Decrypt try order |
|---|---|---|---|---|
| set | set | (any) | `ENCRYPTION_KEY` | match `keyId` → primary → fallback |
| set | unset | (any) | `ENCRYPTION_KEY` | match `keyId` → primary |
| unset | set | set | `APP_SECRET` | match `keyId` → `APP_SECRET` → fallback |
| unset | unset | set | `APP_SECRET` | match `keyId` → `APP_SECRET` |
| unset | unset | unset | startup error | n/a |

## Test plan

- [x] `npx jest 'secret-encryption|connected-account-token-encryption|connected-account-refresh-tokens|encrypt-connected-account-tokens|connection-provider-oauth-flow'` — 87 tests pass
- [x] New `secret-encryption.service.versioned.spec.ts` covers: key resolution table (no-key error, APP_SECRET fallback, ENCRYPTION_KEY precedence), v2 round-trip with/without workspaceId, GCM tamper rejection, workspaceId-mismatch rejection, keyId-based primary→fallback routing, missing-key error names the fingerprint, v1 legacy decryption, no-prefix legacy decryption, malformed envelope rejection.
- [x] Updated `connected-account-token-encryption.service.spec.ts` covers workspaceId binding and HKDF context isolation.
- [x] Updated slow instance command spec verifies workspaceId is threaded through encryption and the relaxed `enc:v_:%` LIKE pattern matches both v1 and v2.
- [ ] Manual E2E: connect a Gmail account on a freshly deployed instance with `APP_SECRET` only → confirm `core.connectedAccount.accessToken` is `enc:v2:<keyId>:<base64>`.
- [ ] Manual E2E: rotate — set `ENCRYPTION_KEY=<new>` and `FALLBACK_ENCRYPTION_KEY=<old APP_SECRET>`, restart, confirm pre-rotation rows still decrypt and new rows carry the new `keyId`.
- [ ] Manual E2E: missing key — set `ENCRYPTION_KEY=<new>` without the fallback, confirm decrypt error names the old `keyId` so the operator can identify the missing key.


packages/twenty-server/src/engine/core-modules/secret-encryption/utils/aes-gcm-v2.util.ts
```diff
@@ -0,0 +1,75 @@
...
+    Buffer.from(rawKey),
+    ZERO_SALT,
+    buildInfo(workspaceId),
+    DERIVED_KEY_LENGTH,
+  );
+
+  return Buffer.from(derived);
+};
+
+export const encryptAesGcmV2 = (
```
charlesBochet: one util per file, we should  add or Throw if it can throw. we should create CustomException and we should pass object as params to ease devXP encryptAesGcmV2({ plaintext, ... })

charlesBochet: same for the whole PR

packages/twenty-server/src/engine/core-modules/secret-encryption/utils/envelope.util.ts
```diff
@@ -0,0 +1,56 @@
...
+export const ENVELOPE_PREFIX = 'enc:';
+export const ENVELOPE_V1_PREFIX = 'enc:v1:';
+export const ENVELOPE_V2_PREFIX = 'enc:v2:';
+
+export type ParsedEnvelope =
+  | { version: 1; payload: string }
+  | { version: 2; keyId: string; payload: string }
+  | { version: null };
+
+export const parseEnvelope = (value: string): ParsedEnvelope => {
```
charlesBochet: also utils naming should be more specific: here parseSecretEncryptionEnvelop, to ease devXP

packages/twenty-server/src/engine/core-modules/secret-encryption/utils/key-resolution.util.ts
```diff
@@ -0,0 +1,36 @@
+import { isNonEmptyString } from '@sniptt/guards';
+
+import { type EnvironmentConfigDriver } from 'src/engine/core-modules/twenty-config/drivers/environment-config.driver';
+
+export type ResolvedEncryptionKeys = {
+  primary: string;
+  fallback: string | null;
+};
+
+// Resolves the three env vars into a primary + optional fallback used by
```
charlesBochet: remove all comments, unless counter intuitive

packages/twenty-server/src/database/commands/upgrade-version-command/2-5/__tests__/2-5-instance-command-slow-1798000004000-encrypt-connected-account-tokens.spec.ts
```diff
@@ -2,19 +2,37 @@ import { type DataSource } from 'typeorm';
...
+
+const wrapAsV2 = (plaintext: string, workspaceId: string): string =>
+  `${SECRET_ENCRYPTION_ENVELOPE_V2_PREFIX}${FAKE_V2_KEY_ID}:CIPHER(${plaintext}|${workspaceId})`;
+
+const unwrapV1 = (ciphertext: string): string => {
+  const match = ciphertext.match(
+    new RegExp(`^${SECRET_ENCRYPTION_ENVELOPE_V1_PREFIX}V1CIPHER\\((.*)\\)$`),
+  );
+
+  if (match === null) {
```
charlesBochet: use typeguard always (isDefined, isNonEmptyString, etc.)

packages/twenty-server/src/engine/core-modules/secret-encryption/constants/secret-encryption.constant.ts
```diff
@@ -0,0 +1,11 @@
+export const SECRET_ENCRYPTION_ENVELOPE_PREFIX = 'enc:';
```
charlesBochet, line 1: filename should be .constant

packages/twenty-server/src/engine/core-modules/secret-encryption/utils/decrypt-aes-ctr-v1-or-throw.util.ts
```diff
@@ -0,0 +1,23 @@
+import { createDecipheriv, createHash } from 'crypto';
+
+const deriveCtrKey = (rawKey: string): string =>
+  createHash('sha512').update(rawKey).digest('hex').substring(0, 32);
+
+// AES-CTR has no integrity tag, so a wrong key silently returns garbage
```
charlesBochet: garbage? too familiar

packages/twenty-server/src/engine/metadata-modules/connected-account/services/connected-account-token-encryption.service.ts
```diff
@@ -14,69 +19,85 @@ export class ConnectedAccountTokenEncryptionService {
     private readonly secretEncryptionService: SecretEncryptionService,
```
prastoin, line 19: ***Polish:*** record args

packages/twenty-server/src/database/commands/upgrade-version-command/2-5/2-5-instance-command-slow-1798000004000-encrypt-connected-account-tokens.ts
```diff
@@ -14,12 +14,36 @@ const ACCESS_TOKEN_CHECK_CONSTRAINT_NAME =
 const REFRESH_TOKEN_CHECK_CONSTRAINT_NAME =
   'CHK_connectedAccount_refreshToken_encrypted';
 
+// Matches the new envelope: rows already in v2 are skipped on re-runs.
+// v1 and plaintext rows are upgraded to v2 (v1 must be upgraded so a later
```
prastoin: ***Remark:*** We don't need to be retro comp over the `v1` stuff as it has not been released yet

***Question:*** From my understanding this should be a dynamic util, the constraint entityId ?

prastoin (commented):
review 1/?

packages/twenty-server/src/engine/core-modules/application/connection-provider/__tests__/connection-provider-oauth-flow.service.spec.ts
```diff
@@ -21,11 +22,11 @@ import { JwtTokenTypeEnum } from 'src/engine/core-modules/auth/types/auth-contex
...
 import { SecureHttpClientService } from 'src/engine/core-modules/secure-http-client/secure-http-client.service';
 import { TwentyConfigService } from 'src/engine/core-modules/twenty-config/twenty-config.service';
+import { SECRET_ENCRYPTION_ENVELOPE_V2_PREFIX } from 'src/engine/core-modules/secret-encryption/constants/secret-encryption.constant';
 import { ConnectedAccountEntity } from 'src/engine/metadata-modules/connected-account/entities/connected-account.entity';
-import {
-  CONNECTED_ACCOUNT_TOKEN_ENCRYPTION_PREFIX,
-  ConnectedAccountTokenEncryptionService,
-} from 'src/engine/metadata-modules/connected-account/services/connected-account-token-encryption.service';
+import { ConnectedAccountTokenEncryptionService } from 'src/engine/metadata-modules/connected-account/services/connected-account-token-encryption.service';
+
```
prastoin, line 28: ***Question:*** why is it different in tests here ?

packages/twenty-server/src/engine/core-modules/secret-encryption/exceptions/secret-encryption.exception.ts
```diff
@@ -0,0 +1,46 @@
+import { type MessageDescriptor } from '@lingui/core';
+import { msg } from '@lingui/core/macro';
+import { assertUnreachable } from 'twenty-shared/utils';
+
+import { CustomException } from 'src/utils/custom-exception';
+
+export enum SecretEncryptionExceptionCode {
```
prastoin, line 7: ***Question:*** So we will be calling all the stateful encrypted data `secret` across the codebase right ?
And signed jwt for stateless ?

packages/twenty-server/src/engine/core-modules/secret-encryption/utils/parse-secret-encryption-envelope-or-throw.util.ts
```diff
@@ -0,0 +1,57 @@
...
+export const parseSecretEncryptionEnvelopeOrThrow = ({
+  value,
+}: {
+  value: string;
+}): ParsedSecretEncryptionEnvelope => {
+  if (!value.startsWith(SECRET_ENCRYPTION_ENVELOPE_PREFIX)) {
+    return { version: null };
+  }
+
+  if (value.startsWith(SECRET_ENCRYPTION_ENVELOPE_V1_PREFIX)) {
```
prastoin: ***Remark:*** I'm not to understand we want to handle this legacy which does not exists yet

packages/twenty-server/src/engine/core-modules/secret-encryption/utils/parse-secret-encryption-envelope-or-throw.util.ts
```diff
@@ -0,0 +1,57 @@
...
+  SecretEncryptionExceptionCode,
+} from 'src/engine/core-modules/secret-encryption/exceptions/secret-encryption.exception';
+import { type ParsedSecretEncryptionEnvelope } from 'src/engine/core-modules/secret-encryption/types/secret-encryption-envelope.type';
+
+export const parseSecretEncryptionEnvelopeOrThrow = ({
+  value,
+}: {
+  value: string;
+}): ParsedSecretEncryptionEnvelope => {
+  if (!value.startsWith(SECRET_ENCRYPTION_ENVELOPE_PREFIX)) {
```
prastoin, line 17: Same this has never been configured before for stateful encrypted string no ?

packages/twenty-server/src/engine/core-modules/secret-encryption/types/secret-encryption-envelope.type.ts
```diff
@@ -0,0 +1,4 @@
+export type ParsedSecretEncryptionEnvelope =
+  | { version: 1; payload: string }
+  | { version: 2; keyId: string; payload: string }
```
prastoin, line 2: ***Remark:*** From my understanding we should only keep 
```ts
  | { version: 2; keyId: string; payload: string }
  | { version: null }
```

packages/twenty-server/src/engine/core-modules/twenty-config/config-variables.ts
```diff
@@ -1151,6 +1151,28 @@ export class ConfigVariables {
...
+  @ConfigVariablesMetadata({
+    group: ConfigVariablesGroup.SERVER_CONFIG,
+    isSensitive: true,
+    description:
+      'Decrypt-only fallback key. During rotation, set this to the previous ENCRYPTION_KEY so rows encrypted with the old key remain readable.',
+    isEnvOnly: true,
+    type: ConfigVariableType.STRING,
+  })
+  @IsOptional()
+  FALLBACK_ENCRYPTION_KEY: string;
```
prastoin, line 1174: ***Challenge:*** Naming would rather be PREVIOUS than FALLBACK no ?

packages/twenty-server/src/engine/core-modules/twenty-config/config-variables.ts
```diff
@@ -1151,6 +1151,28 @@ export class ConfigVariables {
...
+  @ConfigVariablesMetadata({
+    group: ConfigVariablesGroup.SERVER_CONFIG,
+    isSensitive: true,
+    description:
+      'Primary key for at-rest encryption of secrets. Falls back to APP_SECRET when unset.',
+    isEnvOnly: true,
+    type: ConfigVariableType.STRING,
+  })
+  @IsOptional()
+  ENCRYPTION_KEY: string;
```
prastoin, line 1163: ***Question:*** Shouldn't this be required ?

prastoin (commented):
I feel like we're trying to cover legacy we does not exists
Unless I'm mistaken we should only handle previous encryption pattern
Unless I'm mistaken`enc` is just a notion you've just introduced for the jwt signature and `enc:v1` is one I've introduced yesterday but has never been released yet
We could simplify the whole services and utils

packages/twenty-server/src/engine/metadata-modules/connected-account/services/connected-account-token-encryption.service.ts
```diff
@@ -14,69 +19,85 @@ export class ConnectedAccountTokenEncryptionService {
...
   } {
     return {
-      encryptedAccessToken: this.encrypt(accessToken),
-      encryptedRefreshToken: this.encryptNullable(refreshToken),
+      encryptedAccessToken: this.encrypt(accessToken, workspaceId),
+      encryptedRefreshToken: this.encryptNullable(refreshToken, workspaceId),
     };
   }
+
+  private looksLikeCiphertext(value: string): boolean {
```
prastoin, line 118: ***Polish:*** naming `looksLike` -> `is`

packages/twenty-server/src/engine/core-modules/twenty-config/config-variables.ts
prastoin: ok we fallback on the app_secret until it's defined

packages/twenty-server/src/engine/core-modules/secret-encryption/exceptions/secret-encryption.exception.ts
charlesBochet: yes that what I had in mind

packages/twenty-server/src/engine/core-modules/secret-encryption/utils/parse-secret-encryption-envelope-or-throw.util.ts
charlesBochet: that's a fair point

packages/twenty-server/src/engine/core-modules/twenty-config/config-variables.ts
charlesBochet: I think fallback is better as previous seems to be tight to rotation and I wanted to decouple both. could be primary / secondary

packages/twenty-server/src/engine/metadata-modules/connected-account/services/connected-account-token-encryption.service.ts
```diff
@@ -14,69 +19,109 @@ export class ConnectedAccountTokenEncryptionService {
...
-      // v2.4.0 deployment-window tolerance. Should be patch to throw after v2.4.1
-      // throw new Error(
-      //   'ConnectedAccountTokenEncryptionService.decrypt received a value without the ' +
-      //     `'${CONNECTED_ACCOUNT_TOKEN_ENCRYPTION_PREFIX}' prefix. ` +
-      //     'This indicates the column was written without going through encrypt(), ' +
-      //     'or the value was read from a source other than core.connectedAccount.',
-      // );
+  // v2.4.0 rollout-window tolerance: rows written before the encryption
+  // backfill ran may still be plaintext. Returning them as-is lets the slow
+  // command finish; once it has run everywhere this branch can throw.
```
prastoin, line 57: Reading this comment I would have expected an assertion on the upgradeMigration table that would serve as correlation key before starting throwing

prastoin (approved):
Approved but disabled auto-merge as I've left a comment

packages/twenty-server/src/engine/metadata-modules/connected-account/services/connected-account-token-encryption.service.ts
charlesBochet: will remove the comment in next PR, it's actually inaccurate as we have smooth fallback mecanism

PR #20533, opened 2026-05-13
feat(twenty-front): relation traversal in filter dropdown (stacked)

FelixMalfait:
**Stacked on #20527** 

https://github.com/user-attachments/assets/48995655-401a-4c35-8094-e88da8408bdd

## Summary

Surfaces the one-hop relation traversal added in #20527 through the existing **composite sub-field dropdown pattern**. Clicking a MANY_TO_ONE relation field in the "+ Filter" picker now opens the same second-level dropdown that composite fields (FULL_NAME, ADDRESS, CURRENCY, etc.) already use — populated with the target object's filterable fields. Picking one (e.g. `Company → Name`) builds a filter that serializes to the nested GraphQL filter the backend now accepts: `{ company: { name: { ilike: "%X%" } } }`.

No new components. The whole feature reuses `AdvancedFilterSubFieldSelectMenu` + the existing `subFieldNameUsedInDropdownComponentState` + the existing `MenuItem hasSubMenu` indicator. Only the conditions that gate the sub-menu (and the sub-menu's content for relations) were broadened.

## What landed

| File | Change |
|---|---|
| `ObjectFilterDropdownFilterSelectMenuItem` | Sub-menu chevron now shows on MANY_TO_ONE relations (`isManyToOneRelationField` util). |
| `AdvancedFilterFieldSelectMenu` | Relation clicks open the sub-menu alongside composite clicks. |
| `AdvancedFilterSubFieldSelectMenu` | New branch: when the sub-menu type is `'RELATION'`, render the target object's filterable fields via `useFilterableFieldMetadataItems(targetObjectMetadataId)`. Composite logic untouched. |
| `objectFilterDropdownSubMenuFieldType` state | Widened to accept a `'RELATION'` sentinel. Role-permissions sub-field menu narrows it back out (it doesn't traverse relations). |
| `useSelectFieldUsedInAdvancedFilterDropdown` | New optional `targetFieldMetadataItem` arg. When present, the stored RecordFilter's `type` is the target field's type so the operand picker and value input render the target's operands (`'TEXT'` operators when filtering `company.name`, etc.). |
| `turnRecordFilterIntoGqlOperationFilter` (shared) | When the filter targets a `RELATION` field with a `subFieldName`, synthesize a field-metadata for the target, recurse to build the inner filter, then wrap it under the relation field's name → `{ relationName: { targetFieldName: { ...operator } } }`. |

`RecordFilter.subFieldName` stays narrowly typed as `CompositeFieldSubFieldName` so the wide downstream consumers (`shouldShowFilterTextInput`, composite handlers in the serializer, etc.) don't change. The relation target field's name is stored through a narrowly-scoped cast at the dropdown's storage point — the serializer checks `filter.type === 'RELATION'` before interpreting it as a target field name, so the cast can't be mis-read by composite-only code paths.

## Test plan

- [ ] Open a table view on People, click "+ Filter", click "Company" → sub-menu opens with Company's filterable fields
- [ ] Pick "Name" → operand picker shows TEXT operators (Contains, Equals, …)
- [ ] Type "Airbnb" → filter applies, table shows people whose company name contains "Airbnb"
- [ ] Verify network tab: the GraphQL filter variable is `{ company: { name: { ilike: "%Airbnb%" } } }`
- [ ] Same flow with a composite target field (e.g. `Company → annualRecurringRevenue → amountMicros`) — should work end-to-end (backend supports composite-within-relation; #20527 has an integration test covering this)
- [ ] Composite fields (FULL_NAME, ADDRESS) still open their normal sub-menu and filter correctly — no regression
- [ ] Role-permissions field-select sub-field menu is unaffected (it bails out early on the RELATION sentinel)

## Out of scope

- ONE_TO_MANY traversal (no backend support yet)
- Aggregates (`people.count > 5`)
- Persisting relation-traversal filters into a saved view (ViewFilter has no `relationPath` column yet; that's a separate slice)
- REST API DSL changes
- AI Tools


packages/twenty-front/src/modules/object-record/object-filter-dropdown/utils/isManyToOneRelationField.ts
```diff
@@ -0,0 +1,9 @@
+import { type FieldMetadataItem } from '@/object-metadata/types/FieldMetadataItem';
+import { FieldMetadataType, RelationType } from '~/generated-metadata/graphql';
+
+// MANY_TO_ONE relations are the only relation type the GraphQL filter API
+// supports traversing today, so they're also the only ones we expose for
+// drill-down in the filter dropdown.
+export const isManyToOneRelationField = (field: FieldMetadataItem): boolean =>
```
FelixMalfait: Why not creating a helper but more global? Let's analyze all places in the frontend codebase where we do this kind of comparaison and see if a bigger refactor is relevant @claude report

packages/twenty-shared/src/utils/filter/turnRecordFilterGroupIntoGqlOperationFilter.ts
```diff
@@ -18,6 +18,10 @@ export type RecordFilter = {
   type: FilterableAndTSVectorFieldType;
   recordFilterGroupId?: string | null;
   operand: ViewFilterOperand;
+  // Composite sub-field name (e.g. 'firstName') OR, when the filter is a
+  // one-hop relation traversal, the target field's name on the related
+  // object cast to this type. The serializer checks `type === 'RELATION'`
+  // before interpreting it as a target field name.
   subFieldName?: CompositeFieldSubFieldName | null | undefined;
```
FelixMalfait, line 21: Why add this comment now? I don't understand

packages/twenty-front/src/modules/object-record/record-filter/types/RecordFilter.ts
```diff
@@ -23,6 +23,10 @@ export type RecordFilter = {
   operand: ViewFilterOperand;
   positionInRecordFilterGroup?: number | null;
   label: string;
+  // Composite sub-field name (e.g. 'firstName') OR, when the filter is a
+  // one-hop relation traversal, the target field's name on the related
+  // object cast to this type. The serializer checks `type === 'RELATION'`
+  // before interpreting it as a target field name.
```
FelixMalfait: Why add this comment now?

packages/twenty-front/src/modules/object-record/object-filter-dropdown/states/objectFilterDropdownSubMenuFieldTypeComponentState.ts
```diff
@@ -2,8 +2,15 @@ import { ObjectFilterDropdownComponentInstanceContext } from '@/object-record/ob
 import { type CompositeFilterableFieldType } from '@/object-record/record-filter/types/CompositeFilterableFieldType';
 import { createAtomComponentState } from '@/ui/utilities/state/jotai/utils/createAtomComponentState';
 
+// Composite filterable types open a sub-menu for sub-field selection. MANY_TO_ONE
+// relations reuse the same sub-menu pattern to let the user pick a field on
+// the related object (one-hop relation traversal).
+export type ObjectFilterDropdownSubMenuFieldType =
+  | CompositeFilterableFieldType
+  | 'RELATION';
+
```
FelixMalfait: Todo: check behavior in normal filter AND advanced filter ; because composite level filtering was only available in advance filter I think

packages/twenty-shared/src/utils/filter/turnRecordFilterIntoGqlOperationFilter.ts
```diff
@@ -106,6 +106,44 @@ export const turnRecordFilterIntoRecordGqlOperationFilter = ({
...
   const isSubFieldFilter = isNonEmptyString(subFieldName);
 
+  // Relation traversal: the record filter targets a field on the related
+  // object (e.g. `company.name`). The frontend already stored the target
+  // field's type on `recordFilter.type`, so we synthesize a field metadata
+  // for the target, recurse to build the inner filter, then wrap the result
+  // under the relation field's name to produce
+  // `{ relationName: { targetFieldName: { operator: value } } }`.
+  if (
+    correspondingFieldMetadataItem.type === FieldMetadataType.RELATION &&
```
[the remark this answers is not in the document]
FelixMalfait: Stale finding from the pre-refactor diff. In the current code (`turnRecordFilterIntoGqlOperationFilter.ts:90-122` after the merge with main), the relation-traversal branch runs **before** `checkIfShouldComputeEmptinessFilter` — the comment above the if block explicitly calls this out:

> Run this BEFORE the emptiness shortcut so an "is empty" filter on `company.name` is evaluated against `Company.name`, not the FK column.

So an `IS_EMPTY` filter on `{ company: { relationTargetFieldMetadataId: <name id> } }` first recurses with `fieldMetadataId: <name id>` and `relationTargetFieldMetadataId: null`. The recursive call then enters the emptiness branch with the **target** field as `correspondingFieldMetadataItem`, producing `{ company: { name: { is: NULL } } }` rather than `{ companyId: { is: NULL } }`. Tracker cubic was looking at the old `isSubFieldFilter` shape.

packages/twenty-front/src/modules/object-record/advanced-filter/components/AdvancedFilterSubFieldSelectMenu.tsx
```diff
@@ -87,30 +90,53 @@ export const AdvancedFilterSubFieldSelectMenu = ({
...
     !isCompositeTypeNonFilterableByAnySubField(
       fieldMetadataItemUsedInDropdown.type,
     );
 
-  const selectableItemIdArray = [
-    '-1',
-    ...subFieldNames.map((subFieldName) => subFieldName),
-  ];
+  const selectableItemIdArray = isRelationSubMenu
+    ? ['-1', ...relationTargetFields.map((field) => field.id)]
```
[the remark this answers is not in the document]
FelixMalfait: Stale finding. In the current code (`AdvancedFilterSubFieldSelectMenu.tsx:128`) the relation-submenu branch builds its `selectableItemIdArray` directly from the target fields, no `-1` prepended:

```ts
const selectableItemIdArray = relationTargetFields.map((field) => field.id);
```

The `-1` is still present in the **composite** submenu branch (line 205-208) but that is intentional: it pairs with the rendered `<SelectableListItem itemId="-1">` "Any ${fieldLabel} field" entry that composite filters expose (line 230-249). So the only place `-1` appears as an item id, it has a matching rendered row.

packages/twenty-front/src/modules/object-record/object-filter-dropdown/utils/isManyToOneRelationField.ts
FelixMalfait: Done — moved to `@/object-metadata/utils/isManyToOneRelationField.ts` with a structural `{ type, relation? }` input so it can be reused beyond the filter dropdown. Refactored `prefillRecord`, `sanitizeRecordInput`, `filterSortableFieldMetadataItems`, and `spreadsheetImportHasNestedFields` to use it.

packages/twenty-shared/src/utils/filter/turnRecordFilterGroupIntoGqlOperationFilter.ts
FelixMalfait: Removed — the field name is self-describing.

packages/twenty-front/src/modules/object-record/record-filter/types/RecordFilter.ts
FelixMalfait: Removed — the field name is self-describing.

packages/twenty-front/src/modules/object-record/object-filter-dropdown/states/objectFilterDropdownSubMenuFieldTypeComponentState.ts
FelixMalfait: Verified: the dropdown sub-menu state is only consumed by the advanced filter and role-permissions code — the regular `ViewBarFilterDropdownFieldSelectMenuItem` renders without a sub-menu. So relation traversal follows the existing composite filtering pattern: advanced-filter-only. Removed the sentinel comment.

packages/twenty-shared/src/utils/filter/turnRecordFilterIntoGqlOperationFilter.ts
```diff
@@ -87,6 +87,40 @@ export const turnRecordFilterIntoRecordGqlOperationFilter = ({
...
+        [correspondingFieldMetadataItem.name]: innerFilter,
+      } as RecordGqlOperationFilter;
+    }
+    // Target field not in the provided fieldMetadataItems — fall through to
+    // the legacy relation-by-record path so the filter still works.
+  }
+
   const shouldComputeEmptinessFilter = checkIfShouldComputeEmptinessFilter({
     recordFilterOperand: recordFilter.operand,
     correspondingFieldMetadataItem,
```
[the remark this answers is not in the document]
FelixMalfait: Real bug — fixed. `ViewQueryParamsService` now propagates `relationTargetFieldMetadataId` onto the produced `RecordFilter` AND includes the target field in the `fields` array passed to `computeRecordGqlOperationFilter`, so the relation-traversal branch runs server-side too. Added a spec asserting the nested filter shape.

packages/twenty-server/src/engine/metadata-modules/view/services/view-query-params.service.ts
```diff
@@ -105,10 +116,16 @@ export class ViewQueryParamsService {
...
+      ),
+    );
+
+    const fields = Array.from(new Set(fieldIdsReferencedByFilters))
+      .map((fieldId) => {
         const field = findFlatEntityByIdInFlatEntityMaps({
-          flatEntityId: filter.fieldMetadataId,
+          flatEntityId: fieldId,
           flatEntityMaps: flatFieldMetadataMaps,
         });
```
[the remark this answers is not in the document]
FelixMalfait: How does it work for fields today? Is there a SQL cascade? Or an async cascade via a job?

packages/twenty-shared/src/utils/filter/turnRecordFilterIntoGqlOperationFilter.ts
```diff
@@ -87,6 +87,39 @@ export const turnRecordFilterIntoRecordGqlOperationFilter = ({
...
+    const targetFieldMetadataItem = fieldMetadataItems.find(
+      (field) => field.id === recordFilter.relationTargetFieldMetadataId,
+    );
+
+    if (isDefined(targetFieldMetadataItem)) {
+      const innerFilter = turnRecordFilterIntoRecordGqlOperationFilter({
+        recordFilter: {
+          ...recordFilter,
+          fieldMetadataId: targetFieldMetadataItem.id,
+          relationTargetFieldMetadataId: null,
```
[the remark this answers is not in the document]
FelixMalfait: Real bug — fixed two ways: (1) defensive — `turnRecordFilterIntoRecordGqlOperationFilter` now returns `undefined` (drops the filter) when `relationTargetFieldMetadataId` is set but the target isn't in `fieldMetadataItems`, instead of falling through to the legacy UUID-list path that could broaden destructive operations. (2) caller updates — `useGetRecordIndexTotalCount`, `getQueryVariablesFromFiltersAndSorts`, `useUpdateMultipleRecordsActions`, `useRecordIndexLazyFetchRecords`, `useFindManyRecordsSelectedInContextStore`, `RecordIndexContainerContextStoreNumberOfSelectedRecordsEffect`, `buildHeadlessCommandContextApi`, and `useRecordCalendarQueryDateRangeFilter` now pass the flattened cross-object field list. `computeContextStoreFilters` accepts an optional `flattenedFieldMetadataItems` param.

packages/twenty-server/src/engine/metadata-modules/view/services/view-query-params.service.ts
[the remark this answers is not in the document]
FelixMalfait: Not changing — silent-drop of filters whose source field metadata is missing is the existing convention in this code path (codified by the `should skip filters for deleted fields` test in the same spec, and by the `return undefined` on missing `correspondingFieldMetadataItem` at the top of `turnRecordFilterIntoRecordGqlOperationFilter`). My defensive fix in 4a37f65 extends that same convention to relation targets so destructive operations don't broaden via the legacy UUID-list fallback. Moving to a throw-and-notify model is a broader UX change that should land outside this PR, applied uniformly to both kinds of field deletion.

packages/twenty-server/src/engine/metadata-modules/view-filter/entities/view-filter.entity.ts
```diff
@@ -59,6 +64,20 @@ export class ViewFilterEntity
   @Column({ nullable: true, type: 'text', default: null })
   subFieldName: string | null;
 
+  @Column({ nullable: true, type: 'uuid', default: null })
+  relationTargetFieldMetadataId: string | null;
+
+  // ON DELETE SET NULL keeps the row when the target field is deleted: the
+  // load path drops any filter where relationTargetFieldMetadataId was set
+  // but is now null, because operand/value are still shaped for the original
+  // target.
```
FelixMalfait: Interesting. But it's not what we do for fieldMetadata though? Should we align to fieldMetadata's for consistency? This would be a pre-existing bug? Wdyt?

packages/twenty-server/src/engine/metadata-modules/view/services/view-widget-upsert.service.ts
```diff
@@ -719,7 +722,9 @@ export class ViewWidgetUpsertService {
           existingFilter.viewFilterGroupId !== inputFilter.viewFilterGroupId ||
           existingFilter.positionInViewFilterGroup !==
             inputFilter.positionInViewFilterGroup ||
-          existingFilter.subFieldName !== inputFilter.subFieldName;
+          existingFilter.subFieldName !== inputFilter.subFieldName ||
+          existingFilter.relationTargetFieldMetadataId !==
+            (inputFilter.relationTargetFieldMetadataId ?? null);
 
         if (hasChanged) {
           const {
```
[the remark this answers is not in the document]
FelixMalfait: Real bug — fixed in d39a2d0. The hasChanged branch now mirrors the create branch: includes relationTargetFieldMetadataId in the update payload AND resolves relationTargetFieldMetadataUniversalIdentifier alongside the source field and view-filter-group identifiers.

packages/twenty-front/src/modules/object-metadata/utils/isManyToOneRelationField.ts
```diff
@@ -0,0 +1,10 @@
+import { FieldMetadataType, RelationType } from '~/generated-metadata/graphql';
+
+type FieldWithRelation = {
+  type: FieldMetadataType;
+  relation?: { type: RelationType } | null;
+};
+
+export const isManyToOneRelationField = (field: FieldWithRelation): boolean =>
```
FelixMalfait: This looks suspicious, we already have isFieldRelationManyToOne, can't we re-use this?

packages/twenty-server/src/engine/metadata-modules/view-filter/entities/view-filter.entity.ts
FelixMalfait: Good catch — you're right, the other foreign keys on viewFilter (fieldMetadataId, viewId, viewFilterGroupId) all use ON DELETE CASCADE. My SET NULL was an outlier that left half-broken rows in the table and forced a runtime drop-if-null path. Switched relationTargetFieldMetadataId to CASCADE in d947fed: deleting a target field now atomically removes the dependent filters, same as deleting the source field. Migration FK constraint updated to match.

packages/twenty-front/src/modules/object-metadata/utils/isManyToOneRelationField.ts
FelixMalfait: Different input shapes:
- `isFieldRelationManyToOne` (record-field/ui/types/guards) takes `FieldDefinition<FieldMetadata>` and reads `field.metadata.relationType` — record-table / UI layer.
- `isManyToOneRelationField` (object-metadata/utils) takes `FieldMetadataItem` and reads `field.relation.type` — metadata-store layer.

The two layers use different property paths (`metadata.relationType` vs `relation.type`) so they can't trivially be merged without changing the input schema at one of the call sites. The filter dropdown / `prefillRecord` / `sanitizeRecordInput` / etc. work with `FieldMetadataItem`, not `FieldDefinition`. Happy to converge them later under a shared structural helper if we want — but it'd touch every existing caller. WDYT?

packages/twenty-front/src/modules/object-record/advanced-filter/hooks/useSelectFieldUsedInAdvancedFilterDropdown.ts
```diff
@@ -130,6 +147,19 @@ export const useSelectFieldUsedInAdvancedFilterDropdown = () => {
     const subFieldNameToUse =
       subFieldName ?? subFieldNameForNonFilterableWithAny;
 
+    const label = isRelationTraversal
+      ? `${fieldMetadataItem.label} → ${relationTargetFieldMetadataItem.label}`
+      : fieldMetadataItem.label;
+
+    const relationTargetField = isRelationTraversal
+      ? {
+          id: relationTargetFieldMetadataItem.id,
```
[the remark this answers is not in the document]
FelixMalfait: Real bug — fixed in 90a1381. Added a `skipFocusPush` opt-out to the hook, set by the two advanced-filter callers when the next step is the relation sub-menu (case 1) or when committing with a resolved target (case 2 — the target's value picker sets up its own focus). Role-permission callers keep the original push since their RELATION fields go directly to the entity picker.

packages/twenty-shared/src/utils/filter/turnRecordFilterIntoGqlOperationFilter.ts
```diff
@@ -87,6 +87,39 @@ export const turnRecordFilterIntoRecordGqlOperationFilter = ({
     return;
   }
 
+  // Must run before the emptiness shortcut so an "is empty" filter on
+  // `company.name` is evaluated against `Company.name`, not the FK column.
+  // If `relationTargetField` is missing the filter is dropped, not fed
+  // through the legacy relation-by-record path — that path parses `value`
+  // as a UUID list and would silently mishandle a target-field value like
+  // "Acme", risking broadening of destructive operations.
```
FelixMalfait: Comment looks suspicious, will open a refactor in a stacked PR

packages/twenty-front/src/modules/object-metadata/utils/isManyToOneRelationField.ts
charlesBochet: Indeed, I had the same challenge, to me the vision is to get rid of field definition and your new util is actually correct. 
If you want to avoid introducing it, you would need to call formatFieldMetadataItemAsFieldDefinition but it doesn't make sense

Short term: I think we should rename exisitng isFieldRelationManyToOne to isFieldDefinitionRelationManyToOne (but we should rename the whole util family then which is a noisy refactoring, out of scope for this PR)

FelixMalfait: Thanks — keeping the new util as-is then. Happy to do the `isFieldDefinition*` family rename in a follow-up PR whenever the wider FieldDefinition deprecation lands.

packages/twenty-front/src/modules/object-record/advanced-filter/components/AdvancedFilterFieldSelectMenu.tsx
```diff
@@ -112,19 +117,29 @@ export const AdvancedFilterFieldSelectMenu = ({
...
     );
 
+    const subMenuType: ObjectFilterDropdownSubMenuFieldType | null =
+      isManyToOneRelationField(selectedFieldMetadataItem)
+        ? RELATION_SUB_MENU_FIELD_TYPE
+        : isCompositeFilterableFieldType(filterType)
+          ? filterType
+          : null;
+
+    // The relation sub-menu uses its own focusId for selectable navigation;
```
charlesBochet: I struggle to see how this is different from composite sub field case?

packages/twenty-front/src/modules/object-record/advanced-filter/components/AdvancedFilterFieldSelectMenu.tsx
```diff
@@ -16,8 +16,13 @@ import { AdvancedFilterContext } from '@/object-record/advanced-filter/states/co
 import { ObjectFilterDropdownFilterSelectMenuItem } from '@/object-record/object-filter-dropdown/components/ObjectFilterDropdownFilterSelectMenuItem';
 import { fieldMetadataItemIdUsedInDropdownComponentState } from '@/object-record/object-filter-dropdown/states/fieldMetadataItemIdUsedInDropdownComponentState';
 import { objectFilterDropdownIsSelectingCompositeFieldComponentState } from '@/object-record/object-filter-dropdown/states/objectFilterDropdownIsSelectingCompositeFieldComponentState';
-import { objectFilterDropdownSubMenuFieldTypeComponentState } from '@/object-record/object-filter-dropdown/states/objectFilterDropdownSubMenuFieldTypeComponentState';
+import {
+  type ObjectFilterDropdownSubMenuFieldType,
+  RELATION_SUB_MENU_FIELD_TYPE,
```
charlesBochet: nit: looks wrong that the type is part of the state

packages/twenty-front/src/modules/object-record/advanced-filter/components/AdvancedFilterFieldSelectMenu.tsx
FelixMalfait: Good point — fixed in 49633515. Generalized to: any sub-menu type → skip the source-field push (`subMenuType !== null`). The asymmetry was an artifact of the original code's conditional push only firing for RELATION/SELECT types; composite source fields aren't either of those so the smell was invisible there, but the semantic is identical. Sub-field-select-menu now passes `skipFocusPush: true` unconditionally too — same reasoning (next dropdown is sub-field or target, never the source).

packages/twenty-front/src/modules/object-record/advanced-filter/components/AdvancedFilterSubFieldSelectMenu.tsx
```diff
@@ -57,18 +62,35 @@ export const AdvancedFilterSubFieldSelectMenu = ({
...
-  ) => {
-    if (!isDefined(selectedFieldMetadataItem)) {
-      return;
-    }
+  const isRelationSubMenu =
+    objectFilterDropdownSubMenuFieldType === RELATION_SUB_MENU_FIELD_TYPE &&
+    isDefined(fieldMetadataItemUsedInDropdown) &&
+    isManyToOneRelationField(fieldMetadataItemUsedInDropdown);
+
+  const targetObjectMetadataId = isRelationSubMenu
```
charlesBochet: naming + defaulting to '' looks very fragile

packages/twenty-front/src/modules/object-record/advanced-filter/components/AdvancedFilterSubFieldSelectMenu.tsx
```diff
@@ -91,8 +113,71 @@ export const AdvancedFilterSubFieldSelectMenu = ({
     return null;
   }
 
+  if (isRelationSubMenu && isDefined(fieldMetadataItemUsedInDropdown)) {
```
charlesBochet: this works but my feeling on this overall approach is that we try to make the relation case look like the composite case and we introduce if / else branching (if it's a relation, do not do the focus the same way, if it's a relation then we have targetObjectMetadataId but it's '' in composite case...). What we have learned in FE to ease maintainability is to cut the tree higher. Here I would have introduced two different specialized component:
- AdvancedFilterSubFieldSelectMenu.tsx => AdvancedFilterCompositeSubFieldSelectMenu
- AdvancedFilterRelationFieldSelectMenu

this way the whole logic is more robust to changes

charlesBochet: btw, subField here but targetField in the state, that really show the confusion
so maybe even: AdvancedFilterRelationTargetFieldSelectMenu

packages/twenty-front/src/modules/object-record/advanced-filter/hooks/useSelectFieldUsedInAdvancedFilterDropdown.ts
```diff
@@ -23,6 +25,11 @@ type SelectFilterParams = {
   fieldMetadataItemId: string;
   recordFilterId: string;
   subFieldName?: CompositeFieldSubFieldName | null | undefined;
+  relationTargetFieldMetadataItem?: FieldMetadataItem | null | undefined;
+  // Set when the next step opens its own dropdown with a different focusId
```
charlesBochet: a bit code smell to me, complexifying edge case that I don't really understand why it's different

packages/twenty-front/src/modules/object-record/object-filter-dropdown/states/objectFilterDropdownSubMenuFieldTypeComponentState.ts
```diff
@@ -2,8 +2,14 @@ import { ObjectFilterDropdownComponentInstanceContext } from '@/object-record/ob
...
 
+export const RELATION_SUB_MENU_FIELD_TYPE = 'RELATION' as const;
+
+export type ObjectFilterDropdownSubMenuFieldType =
+  | CompositeFilterableFieldType
+  | typeof RELATION_SUB_MENU_FIELD_TYPE;
+
 export const objectFilterDropdownSubMenuFieldTypeComponentState =
-  createAtomComponentState<CompositeFilterableFieldType | null>({
+  createAtomComponentState<ObjectFilterDropdownSubMenuFieldType | null>({
```
charlesBochet, line 6: same thing here, I feel we should leave this state alone and take relationTargetFieldMetadataIdUsedInDropdownComponentState as only source of truth (if we introduce two "SubField" component, we don't have to do this hack anymore)

packages/twenty-front/src/modules/object-record/record-filter/types/RecordFilter.ts
```diff
@@ -24,7 +25,16 @@ export type RecordFilter = {
   positionInRecordFilterGroup?: number | null;
   label: string;
   subFieldName?: CompositeFieldSubFieldName | null | undefined;
-  // RLS-specific: when set, filter compares against current user's field value
+  // Resolved at filter construction time (frontend mapping or chip edit) so
+  // the GraphQL builder doesn't need a cross-object field lookup. Null for
+  // non-relation filters; also null when the target field can no longer be
+  // resolved (e.g. it was deleted).
+  relationTargetField?: {
```
charlesBochet: if possible, we avoid that now and we prefer to derive when needed unless there are big performance trade-off. You only need the id in the state and then you can derive the rest form other states (objectMetadataItem)

packages/twenty-front/src/modules/settings/roles/role-permissions/object-level-permissions/record-level-permissions/components/SettingsRolePermissionsObjectLevelRecordLevelPermissionFieldSelectSubFieldMenu.tsx
```diff
@@ -94,7 +97,10 @@ export const SettingsRolePermissionsObjectLevelRecordLevelPermissionFieldSelectS
       advancedFilterFieldSelectDropdownId,
     );
 
-    if (!isDefined(objectFilterDropdownSubMenuFieldType)) {
+    if (
+      !isDefined(objectFilterDropdownSubMenuFieldType) ||
+      objectFilterDropdownSubMenuFieldType === RELATION_SUB_MENU_FIELD_TYPE
```
charlesBochet: yep :p another code smell that we should not try to hack composite sub field feature

packages/twenty-front/src/modules/views/hooks/useMapViewFiltersToFilters.ts
```diff
@@ -7,12 +9,20 @@ import { mapViewFiltersToFilters } from '@/views/utils/mapViewFiltersToFilters';
...
+
   const mapViewFiltersToRecordFilters = (
     viewFilters: ViewFilter[] | GqlViewFilter[],
   ) => {
     const filterableFieldMetadataItems =
       getFilterableFields(objectMetadataItem);
-    return mapViewFiltersToFilters(viewFilters, filterableFieldMetadataItems);
+    return mapViewFiltersToFilters(
+      viewFilters,
+      filterableFieldMetadataItems,
```
charlesBochet: sounds wrong to pass fieldMetadataItems twice, maybe filterableFieldMetadataItems is not needed anymore?

packages/twenty-front/src/modules/object-record/advanced-filter/components/AdvancedFilterFieldSelectMenu.tsx
FelixMalfait: Fair — moved the constant to `object-filter-dropdown/constants/RelationSubMenuFieldType.ts` and the type to `record-filter/types/ObjectFilterDropdownSubMenuFieldType.ts`. State file now only exports the atom.

packages/twenty-front/src/modules/object-record/advanced-filter/components/AdvancedFilterSubFieldSelectMenu.tsx
FelixMalfait: Agreed — extracted the relation branch into its own `AdvancedFilterRelationSubMenu` component that only mounts when in relation mode. The hook now always receives a real `targetObjectMetadataId`, no `''` sentinel. Also turned `isManyToOneRelationField` into a type guard so the call site can read `relation.targetObjectMetadata.id` after the check without a non-null assertion.

packages/twenty-server/src/engine/api/common/common-query-runners/common-group-by-query-runner.service.ts
```diff
@@ -222,15 +222,34 @@ export class CommonGroupByQueryRunnerService extends CommonBaseQueryRunnerServic
...
         value: convertViewFilterValueToString(viewFilter.value),
-        type: getFilterTypeFromFieldType(fieldMetadata.type),
+        type: getFilterTypeFromFieldType(
+          relationTargetField?.type ?? fieldMetadata.type,
+        ),
         operand: viewFilter.operand,
         recordFilterGroupId: viewFilter.viewFilterGroupId,
         positionInRecordFilterGroup: viewFilter.positionInViewFilterGroup,
         subFieldName: viewFilter.subFieldName as CompositeFieldSubFieldName,
+        relationTargetField: isDefined(relationTargetField)
```
charlesBochet: I'm lost, why is this needed in backend?

packages/twenty-server/src/engine/metadata-modules/flat-view-filter/utils/from-update-view-filter-input-to-flat-view-filter-to-update-or-throw.util.ts
```diff
@@ -87,5 +87,22 @@ export const fromUpdateViewFilterInputToFlatViewFilterToUpdateOrThrow = ({
       viewFilterGroupUniversalIdentifier;
   }
 
+  if (
+    updatedEditableFieldProperties.relationTargetFieldMetadataId !== undefined
```
charlesBochet, line 91: use typeguard

packages/twenty-server/src/engine/metadata-modules/view/services/view-query-params.service.ts
```diff
@@ -82,14 +82,31 @@ export class ViewQueryParamsService {
...
         return {
           id: viewFilter.id,
           fieldMetadataId: viewFilter.fieldMetadataId,
           value: viewFilter.value ?? '',
-          type: field.type,
+          type: relationTargetField?.type ?? field.type,
           recordFilterGroupId: viewFilter.viewFilterGroupId,
           operand: viewFilter.operand,
           subFieldName: viewFilter.subFieldName,
+          relationTargetField: isDefined(relationTargetField)
```
charlesBochet: why is that needed here too?

packages/twenty-shared/src/utils/filter/turnRecordFilterGroupIntoGqlOperationFilter.ts
```diff
@@ -19,6 +20,12 @@ export type RecordFilter = {
   recordFilterGroupId?: string | null;
   operand: ViewFilterOperand;
   subFieldName?: CompositeFieldSubFieldName | null | undefined;
+  // Resolved at filter construction time so traversal doesn't need a
```
charlesBochet: I'm not convinced this is a good idea

packages/twenty-shared/src/utils/filter/turnRecordFilterIntoGqlOperationFilter.ts
charlesBochet: agree

charlesBochet (changes requested):
Left comments!

packages/twenty-front/src/modules/object-record/advanced-filter/components/AdvancedFilterFieldSelectMenu.tsx
FelixMalfait: Right, after this refactor they are no longer treated as the same case. Source-field selection now branches: relation field opens `AdvancedFilterRelationTargetFieldSelectMenu`, composite field opens `AdvancedFilterCompositeSubFieldSelectMenu`. Each path sets its own dedicated state and runs through its own component.

packages/twenty-front/src/modules/object-record/advanced-filter/components/AdvancedFilterFieldSelectMenu.tsx
FelixMalfait: Agreed. `objectFilterDropdownSubMenuFieldTypeComponentState` is now `CompositeFilterableFieldType | null` (no more RELATION sentinel). Relation traversal uses a separate `objectFilterDropdownIsSelectingRelationTargetFieldComponentState` boolean signal.

packages/twenty-front/src/modules/object-record/advanced-filter/components/AdvancedFilterSubFieldSelectMenu.tsx
FelixMalfait: Agreed. The dual-purpose component is gone. The relation flow now has a dedicated `AdvancedFilterRelationTargetFieldSelectMenu` that reads the source field from `fieldMetadataItemUsedInDropdownComponentSelector` and derives `targetObjectMetadataId` only when the source is a many-to-one relation — no `''` fallback.

packages/twenty-front/src/modules/object-record/advanced-filter/components/AdvancedFilterSubFieldSelectMenu.tsx
FelixMalfait: Done. Split into two specialized components: `AdvancedFilterCompositeSubFieldSelectMenu` and `AdvancedFilterRelationTargetFieldSelectMenu` (named so it's clear we pick a *field*, not a sub-field). `AdvancedFilterFieldSelectDropdownContent` now dispatches between source / composite / relation-target based on dedicated state signals.

packages/twenty-front/src/modules/object-record/advanced-filter/hooks/useSelectFieldUsedInAdvancedFilterDropdown.ts
FelixMalfait: Done. `skipFocusPush` is gone. The hook only does state updates now. Each call site that opens a value-picker dropdown (advanced-filter leaf select, settings field select, relation target select) does its own `pushFocusItemToFocusStack` for RELATION/SELECT fields, right where the decision lives.

packages/twenty-front/src/modules/object-record/object-filter-dropdown/states/objectFilterDropdownSubMenuFieldTypeComponentState.ts
FelixMalfait: Done. `ObjectFilterDropdownSubMenuFieldType` is back to `CompositeFilterableFieldType` only — the `RELATION_SUB_MENU_FIELD_TYPE` constant is deleted. Relation traversal is signalled by the new `objectFilterDropdownIsSelectingRelationTargetFieldComponentState` boolean, and the resolved target id flows through `relationTargetFieldMetadataIdUsedInDropdownComponentState` as before.

packages/twenty-front/src/modules/object-record/record-filter/types/RecordFilter.ts
FelixMalfait: Done. `RecordFilter` now only carries `relationTargetFieldMetadataId: string | null`. The resolved target field's label/type/etc. is derived at render time from `objectMetadataItem` / flat field metadata items in `mapViewFiltersToFilters`, `useGetRecordFilterDisplayValue`, the shared GraphQL dispatcher, etc.

packages/twenty-front/src/modules/settings/roles/role-permissions/object-level-permissions/record-level-permissions/components/SettingsRolePermissionsObjectLevelRecordLevelPermissionFieldSelectSubFieldMenu.tsx
FelixMalfait: Done — the `=== RELATION_SUB_MENU_FIELD_TYPE` early-return is removed (along with the constant itself). This file is now strictly the composite path.

packages/twenty-front/src/modules/views/hooks/useMapViewFiltersToFilters.ts
FelixMalfait: Done. `mapViewFiltersToFilters` takes a single `fieldMetadataItems` list now. `useMapViewFiltersToFilters` just passes `flattenedFieldMetadataItems` through. Callers were updated accordingly (`useLoadRecordIndexStates`, `useApplyCurrentViewFiltersToCurrentRecordFilters`, etc.).

packages/twenty-server/src/engine/api/common/common-query-runners/common-group-by-query-runner.service.ts
FelixMalfait: Right, the embedding was unnecessary. Reverted — the backend now just propagates `relationTargetFieldMetadataId` (the id) onto the `RecordFilter`, and `turnRecordFilterIntoRecordGqlOperationFilter` resolves the target field from the flat field maps it already receives.

packages/twenty-server/src/engine/metadata-modules/flat-view-filter/utils/from-update-view-filter-input-to-flat-view-filter-to-update-or-throw.util.ts
FelixMalfait: Done — switched to `isDefined(updatedEditableFieldProperties.relationTargetFieldMetadataId)`.

packages/twenty-server/src/engine/metadata-modules/view/services/view-query-params.service.ts
FelixMalfait: Same as common-group-by — embedding was unnecessary. Reverted: `view-query-params.service.ts` now just sets `relationTargetFieldMetadataId` from the view filter and lets the shared dispatcher resolve the target field from `flatFieldMetadataMaps` at filter-compute time.

packages/twenty-shared/src/utils/filter/turnRecordFilterGroupIntoGqlOperationFilter.ts
FelixMalfait: Reverted. The shape extension is gone — the shared util reads `RecordFilter` with `relationTargetFieldMetadataId` (id only) from the canonical type, and the resolution happens inside `turnRecordFilterIntoRecordGqlOperationFilter` against `fieldMetadataItems`.

packages/twenty-shared/src/utils/filter/turnRecordFilterIntoGqlOperationFilter.ts
FelixMalfait: Stacked refactor is PR #20600 (`refactor/turn-record-filter-into-gql-operation-split`): the dispatcher `turnRecordFilterIntoRecordGqlOperationFilter` is split from the inner per-type `computeRecordGqlOperationFilterFromRecordFilter`, and the misleading 'legacy' wording is dropped. The id-only `relationTargetFieldMetadataId` model means the comment is no longer needed here either.

packages/twenty-front/src/modules/object-record/object-filter-dropdown/components/ObjectFilterDropdownInnerSelectOperandDropdown.tsx
```diff
@@ -32,11 +34,25 @@ export const ObjectFilterDropdownInnerSelectOperandDropdown = () => {
...
+  const { getFieldMetadataItemByIdOrThrow } =
+    useGetFieldMetadataItemByIdOrThrow();
+
+  const effectiveFieldMetadataItem = isDefined(
+    relationTargetFieldMetadataIdUsedInDropdown,
+  )
+    ? getFieldMetadataItemByIdOrThrow(
+        relationTargetFieldMetadataIdUsedInDropdown,
+      ).fieldMetadataItem
+    : fieldMetadataItemUsedInDropdown;
```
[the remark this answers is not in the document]
FelixMalfait: Fixed in 7158fab9fa. `ObjectFilterDropdownInnerSelectOperandDropdown` now uses `getFieldMetadataItemById` (non-throwing) for the relation target lookup — if the target field has been deleted, the resolved item is `undefined`, the existing `!isDefined(effectiveFieldMetadataItem)` guard short-circuits, and the component renders `null` instead of crashing.

packages/twenty-front/src/modules/object-record/advanced-filter/components/AdvancedFilterFieldSelectMenu.tsx
```diff
@@ -112,19 +123,50 @@ export const AdvancedFilterFieldSelectMenu = ({
...
       fieldMetadataItemId: selectedFieldMetadataItem.id,
       recordFilterId,
     });
 
-    if (isCompositeFilterableFieldType(filterType)) {
-      setObjectFilterDropdownSubMenuFieldType(filterType);
+    if (isRelationTraversalField) {
+      setFieldMetadataItemIdUsedInDropdown(selectedFieldMetadataItem.id);
+      setObjectFilterDropdownIsSelectingRelationTargetField(true);
+      return;
```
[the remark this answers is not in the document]
FelixMalfait: Real bug despite sentry's auto-resolved tag. Fixed in 50bd2a78a7: the source-field handler now only stages the choice for sub-menu paths (composite + relation) and defers `upsertRecordFilter` to the sub-menu's hook. Backing out of the relation target picker no longer leaves an orphan partial filter in the chip list. Same fix benefits composite back-out.

packages/twenty-server/src/engine/metadata-modules/view/services/view-widget-upsert.service.ts
```diff
@@ -706,6 +706,9 @@ export class ViewWidgetUpsertService {
           positionInViewFilterGroup:
             inputFilter.positionInViewFilterGroup ?? null,
           subFieldName: inputFilter.subFieldName ?? null,
+          relationTargetFieldMetadataId:
+            inputFilter.relationTargetFieldMetadataId ?? null,
+          relationTargetFieldMetadataUniversalIdentifier: null,
```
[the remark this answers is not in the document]
FelixMalfait: Fixed in 2bd03cab91. CREATE path now mirrors UPDATE: `relationTargetFieldMetadataId` is included in `foreignKeyValues` passed to `resolveEntityRelationUniversalIdentifiers`, and the resolved `relationTargetFieldMetadataUniversalIdentifier` is read from its result instead of being hardcoded to `null`.

packages/twenty-front/src/modules/object-record/advanced-filter/hooks/useApplyAdvancedFilterRelationTargetField.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  const setRelationTargetFieldMetadataIdUsedInDropdown =
+    useSetAtomComponentState(
+      relationTargetFieldMetadataIdUsedInDropdownComponentState,
+    );
+
+  const setObjectFilterDropdownCurrentRecordFilter = useSetAtomComponentState(
+    objectFilterDropdownCurrentRecordFilterComponentState,
+  );
+
```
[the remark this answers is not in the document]
FelixMalfait: Fixed in 203f72116a. The target picker now filters out many-to-one relations from `useFilterableFieldMetadataItems` results — only primitive/composite/leaf-relation fields appear, enforcing the one-hop constraint at the UI layer.

packages/twenty-front/src/modules/object-record/advanced-filter/components/AdvancedFilterFieldSelectMenu.tsx
```diff
@@ -112,19 +123,55 @@ export const AdvancedFilterFieldSelectMenu = ({
...
+
+    // For sub-menu paths (composite or relation traversal) we only stage
+    // the source field in the dropdown state and open the sub-menu —
+    // upsertRecordFilter happens in the sub-menu's hook once the user
+    // makes the final choice. Otherwise navigating back from the sub-menu
+    // would leave an orphan partial filter in the chip list.
+    if (isRelationTraversalField) {
+      setFieldMetadataItemIdUsedInDropdown(selectedFieldMetadataItem.id);
+      setObjectFilterDropdownIsSelectingRelationTargetField(true);
+      return;
```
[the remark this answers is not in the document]
FelixMalfait: Fixed in 88d0476da0. `useResetFilterDropdown` now clears `objectFilterDropdownIsSelectingRelationTargetField` alongside the composite flag, so dismissing the dropdown mid-pick (Escape, click-away) returns the next open to the main field-select menu.

packages/twenty-front/src/modules/views/utils/mapViewFiltersToFilters.ts
```diff
@@ -14,27 +14,38 @@ import { type ViewFilter } from '@/views/types/ViewFilter';
...
-      const filterType = getFilterTypeFromFieldType(
-        availableFieldMetadataItem.type,
-      );
+      const relationTargetFieldMetadataItem = isDefined(
+        viewFilter.relationTargetFieldMetadataId,
+      )
+        ? fieldMetadataItems.find(
+            (fieldMetadataItem) =>
+              fieldMetadataItem.id === viewFilter.relationTargetFieldMetadataId,
+          )
```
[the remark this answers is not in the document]
FelixMalfait: Fixed in e3638925a2. `mapViewFiltersToFilters` now drops the filter when `relationTargetFieldMetadataId` is set but the target field can't be resolved — mirroring the existing handling for unresolvable source fields. This avoids the UI/dispatcher mismatch where the chip showed up with the source filterType/label while the dispatcher silently dropped the GraphQL filter.

Issue #20535, opened 2026-05-13
MCP OAuth consent modal does not reopen after sign-in


Bonapara:
## Summary

When an unauthenticated user starts an OAuth authorization flow for the MCP endpoint, Twenty redirects them to the sign-in modal, but after completing sign-in the OAuth consent modal does not reopen. This blocks connecting clients such as the ChatGPT connector to `https://api.twenty.com/mcp`.

## Steps to reproduce

1. Start from a signed-out browser session.
2. Open an OAuth authorize URL similar to:

```text
https://app.twenty.com/authorize?response_type=code&client_id=<client-id>&redirect_uri=<connector-callback>&scope=api+profile&code_challenge=<challenge>&code_challenge_method=S256&resource=https%3A%2F%2Fapi.twenty.com%2Fmcp&state=<state>
```

3. Complete sign-in from the modal, for example with Google or Microsoft.
4. Observe the destination after authentication/workspace selection.

## Actual behavior

The sign-in flow completes, but the original `/authorize` request is lost. The OAuth consent modal does not open, so the client never receives an authorization code.

## Expected behavior

The original `/authorize` request should survive sign-in and workspace-domain redirects. After authentication, the user should land back on `/authorize` with the original OAuth parameters (`redirect_uri`, `scope`, `resource`, PKCE challenge, and `state`) and see the consent modal. Authorizing should redirect back to the client with `code` and the original `state`.

## Notes

Local email sign-in can preserve the authorize request through `returnToPath`, but social SSO redirects need to carry the same value through the provider redirect and backend verification redirect.

thomtrp:
closing given you merged a related pr @FelixMalfait. If not fixed please re-open

Issue #20296, opened 2026-05-05
ChatGPT Apps OAuth connect flow does not complete for Twenty MCP


Bonapara:
## Summary

Creating a ChatGPT app/connector against the Twenty MCP endpoint with OAuth support does not complete the connection flow. ChatGPT shows the app as OAuth-supported, but after clicking `Connect`, the app remains disconnected and shows no actions.

## Reproduction

1. In ChatGPT, enable Developer Mode for Apps/Connectors.
2. Create a new app with:
   - Name: `Twenty`
   - MCP Server URL: `https://api.twenty.com/mcp`
   - Authentication: `OAuth`
3. Open the app details page.
4. Click `Connect`.

## Observed behavior

- ChatGPT shows the app as OAuth-supported.
- The app details page still shows the `Connect` button after the flow.
- `Actions` says `No app actions available yet`.
- The connection does not appear to reach a completed authenticated state.

## Expected behavior

- ChatGPT should redirect/open the Twenty OAuth authorization page.
- After the user authorizes, ChatGPT should complete the OAuth callback.
- The app should show as connected and expose Twenty MCP tools/actions.

## Endpoint checks

The public endpoints appear reachable:

- `GET https://api.twenty.com/.well-known/oauth-protected-resource/mcp` returns protected resource metadata with:
  - `resource: https://api.twenty.com/mcp`
  - `authorization_servers: ["https://api.twenty.com"]`
  - scopes `api profile`
- `GET https://api.twenty.com/.well-known/oauth-authorization-server` returns OAuth discovery metadata with:
  - `authorization_endpoint: https://app.twenty.com/authorize`
  - `token_endpoint: https://api.twenty.com/oauth/token`
  - `registration_endpoint: https://api.twenty.com/oauth/register`
  - `code_challenge_methods_supported: ["S256"]`
- Dynamic client registration succeeds with `201` when posting to `https://api.twenty.com/oauth/register`.
- The generated `https://app.twenty.com/authorize?...` URL serves the Twenty frontend.

## Suspected issue

This may be an MCP OAuth compatibility issue around the OAuth `resource` parameter.

In the current code, the authorize page parses `client_id`, `code_challenge`, `redirect_uri`, and `state`, but does not preserve `resource`:

- `packages/twenty-front/src/pages/auth/Authorize.tsx`

The authorization code context stores redirect URI, client id, scope, and PKCE challenge, but not the requested resource:

[...]

PR #20537, opened 2026-05-13
fix(auth): preserve returnToPath across Google/Microsoft SSO redirects

FelixMalfait:
## Summary

Fixes the consent-modal-not-reopening half of [#20535](https://github.com/twentyhq/twenty/issues/20535): when a signed-out user opens an OAuth `/authorize?...` URL (e.g. ChatGPT connecting to `api.twenty.com/mcp`) and signs in with **Google or Microsoft**, the original `/authorize` request was lost and the consent screen never reopened.

### Root cause

`PageChangeEffect` already saves the deep link as `returnToPath` (Jotai atom) before navigating to `/welcome`. That atom is in-memory: it survives SPA navigation, and the cross-subdomain workspace hop is handled by `useBuildSearchParamsFromUrlSyncedStates` round-tripping the value through the URL.

But the social-SSO path leaves `app.twenty.com` entirely — `app.twenty.com/welcome` → `api.twenty.com/auth/google` → Google → `api.twenty.com/auth/google/redirect` → frontend — so the atom is wiped. None of the existing code paths plumbed `returnToPath` through that hop:
- `useAuth.buildRedirectUrl` packed `workspaceInviteHash`/`action`/etc. but not `returnToPath`.
- `SocialSSOState` / the Google + Microsoft strategies didn't carry it through the OAuth `state` blob.
- `signInUpWithSocialSSO` + `computeRedirectURI` didn't re-emit it on the redirect back to the frontend.

The email path worked because all transitions stay on the default frontend domain, so the atom survives until `SignInUpGlobalScopeForm` bakes it into the workspace URL.

### What changed

Plumb `returnToPath` through the SSO state the same way `workspaceInviteHash` and `action` already flow:

- **Frontend** (`useAuth.buildRedirectUrl`): read `returnToPath` from the Jotai store and append it to `/auth/google` / `/auth/microsoft` when set and structurally valid.
- **Server types** (`SocialSSOState`, `GoogleRequest['user']`, `MicrosoftRequest['user']`): add optional `returnToPath`.
- **Strategies** (`google.auth.strategy.ts`, `microsoft.auth.strategy.ts`): include `returnToPath: req.query.returnToPath` in the JSON `state` and read it back in `validate`.
- **auth.service.ts** (`signInUpWithSocialSSO`, `computeRedirectURI`): forward `returnToPath` on both branches — the multi-workspace redirect to `AppPath.SignInUp?tokenPair=...` and the single-workspace redirect to `<workspace>/verify?loginToken=...`. Validated via a new `isValidReturnToPath` helper so a tampered query value can't become an open-redirect vector.

After the round-trip, `useInitializeQueryParamState` rehydrates the atom from the URL and `usePageChangeEffectNavigateLocation` resolves it as the post-auth destination — same mechanism the email path already relied on.

Out of scope: the OAuth `resource` parameter handling tracked in [#20296](https://github.com/twentyhq/twenty/issues/20296) is independent and not addressed here.

## Test plan

- [x] `npx jest src/engine/core-modules/auth` (twenty-server) — 27 suites / 183 tests pass, including new `is-valid-return-to-path.util.spec.ts`.
- [x] `npx jest src/modules/auth` (twenty-front) — 13 suites / 52 tests pass, including two new cases in `useAuth.test.tsx` covering the happy path and the protocol-relative open-redirect guard.
- [x] `npx nx typecheck twenty-server` / `twenty-front` — clean.
- [x] `npx oxlint` + `prettier --check` on touched files — clean.
- [ ] Manual: signed-out user opens `https://app.twenty.com/authorize?client_id=...` → Continue with Google → completes Google → selects workspace → consent screen renders.
- [ ] Manual: same flow, single workspace — lands on consent screen directly after Verify.
- [ ] Manual: email path still works (regression).
- [ ] Manual: tamper `returnToPath=//evil.com` on the `/auth/google` URL → server validation rejects, user lands at default home, not at `evil.com`.

E2E note: existing `return-to-path.spec.ts` already covers deep links with query params through the email path. A mock OAuth provider would be needed to cover the SSO path end-to-end; unit coverage stands in for now.


packages/twenty-server/src/engine/core-modules/auth/utils/is-valid-return-to-path.util.ts
```diff
@@ -0,0 +1,20 @@
+// Server-side return-to-path validator.
+// Used to reject open-redirect attempts when forwarding the original deep link
+// (typically /authorize?...) back to the frontend after a social SSO callback.
+//
+// Mirrors the client-side check in
+// packages/twenty-front/src/modules/auth/utils/isValidReturnToPath.ts but
+// intentionally only enforces the structural rules that the server can reason
+// about without depending on frontend route constants. The frontend applies a
+// stricter check (excludes onboarding/auth paths) before navigating.
+export const isValidReturnToPath = (path: unknown): path is string => {
```
FelixMalfait: It feels like something we might have done elsewhere already. Please check across all the codebase for similar utils that could be reused

packages/twenty-front/src/modules/auth/hooks/__tests__/useAuth.test.tsx
```diff
@@ -85,6 +87,9 @@ const renderHooks = () => {
 describe('useAuth', () => {
   beforeEach(() => {
     jest.clearAllMocks();
+    // Reset the in-memory returnToPath between tests since the Jotai default
```
FelixMalfait: Remove comment

packages/twenty-server/src/engine/core-modules/auth/utils/is-valid-return-to-path.util.ts
FelixMalfait: Searched the codebase — the closest existing util is [`isSafeUrl`](packages/twenty-shared/src/utils/url/isSafeUrl.ts) in `twenty-shared`, which already rejects `//evil.com` and accepts only `http/https/mailto/tel` for absolute URLs. It's broader than what I needed (it accepts absolute URLs too), but composing it with a `startsWith('/')` guard gives exactly the open-redirect protection `returnToPath` requires.

Dropped the new util and its spec; the two call sites in `auth.service.ts` now use `isSafeUrl` inline. The frontend's [`isValidReturnToPath`](packages/twenty-front/src/modules/auth/utils/isValidReturnToPath.ts) remains the authoritative check (it additionally excludes onboarding/auth paths, which is frontend-specific). Done in 05d12213b4.

packages/twenty-front/src/modules/auth/hooks/__tests__/useAuth.test.tsx
FelixMalfait: Done in 05d12213b4.

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
```diff
@@ -995,6 +1007,11 @@ export class AuthService {
               targetedTokenType: JwtTokenTypeEnum.WORKSPACE_AGNOSTIC,
             }),
           }),
+          ...(returnToPath !== undefined &&
+          returnToPath.startsWith('/') &&
```
FelixMalfait: From what I read, this is basically equivalent to url.startsWith('/') && !url.startsWith('//'))

packages/twenty-front/src/modules/auth/hooks/useAuth.ts
```diff
@@ -524,9 +527,19 @@ export const useAuth = () => {
         url.searchParams.set('workspaceId', workspacePublicData.id);
       }
 
+      // Forward the in-memory returnToPath through the SSO redirect chain so
```
FelixMalfait: Comment not needed

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
```diff
@@ -761,17 +765,24 @@ export class AuthService {
...
       workspace,
       pathname: AppPath.Verify,
       searchParams: {
         loginToken,
         ...(billingCheckoutSessionState ? { billingCheckoutSessionState } : {}),
+        ...(returnToPath !== undefined &&
+        returnToPath.startsWith('/') &&
+        isSafeUrl(returnToPath)
+          ? { returnToPath }
+          : {}),
```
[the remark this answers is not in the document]
FelixMalfait: Good catch — `returnToPath` comes from JSON-parsed OAuth state so the type is effectively `unknown` at runtime. Switched to `typeof returnToPath === 'string'` in both call sites. Done in 0c648d9732.

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
FelixMalfait: Agreed — verified by walking through `isSafeUrl`'s branches: when combined with `x.startsWith('/')`, only the first branch (`startsWith('/') && !startsWith('//')`) can ever be reached. Dropped `isSafeUrl` and inlined the equivalent guard. Done in 0c648d9732.

packages/twenty-front/src/modules/auth/hooks/useAuth.ts
FelixMalfait: Done in 0c648d9732.

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
```diff
@@ -761,17 +761,24 @@ export class AuthService {
...
     billingCheckoutSessionState?: string;
+    returnToPath?: string;
   }) {
     const url = this.workspaceDomainsService.buildWorkspaceURL({
       workspace,
       pathname: AppPath.Verify,
       searchParams: {
         loginToken,
         ...(billingCheckoutSessionState ? { billingCheckoutSessionState } : {}),
+        ...(typeof returnToPath === 'string' &&
```
charlesBochet: use type guard isNonEmptyString

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
```diff
@@ -761,17 +761,24 @@ export class AuthService {
...
+        returnToPath.startsWith('/') &&
+        !returnToPath.startsWith('//')
```
charlesBochet: this looks suspicious? why?

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
```diff
@@ -995,6 +1003,11 @@ export class AuthService {
               targetedTokenType: JwtTokenTypeEnum.WORKSPACE_AGNOSTIC,
             }),
           }),
+          ...(typeof returnToPath === 'string' &&
```
charlesBochet: duplicated

FelixMalfait: Not worth extracting imo

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
FelixMalfait: thanks!

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
FelixMalfait: yes I've been a bit lazy. It did a few iterations to protect from malicious path and we ended up concluding isSafeUrl should be used and it had that, but I didn't question it. Will check. Maybe will remove from isSafeUrl too then

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
FelixMalfait: Done in f06cf1cdf6 — switched both call sites to `isNonEmptyString(returnToPath) && returnToPath.startsWith('/')`.

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
FelixMalfait: Walked through the consumers and you're both right — `!startsWith('//')` is redundant on the server:

- The server only uses `returnToPath` as a **query-param value** on a redirect URL pointing back to our own domain (`buildWorkspaceURL` / `buildBaseUrl`), never as a redirect target. So a value like `//evil.com` flowing through the server cannot itself drive navigation.
- Every frontend code path that ingests `returnToPath` into the `returnToPathState` atom (`useInitializeQueryParamState`, `useApolloFactory`, `useReturnToPath.saveReturnToPath`) gates the write on the frontend `isValidReturnToPath`, which already enforces `!startsWith('//')`. Downstream consumers read from that already-validated atom.

So the trust boundary is the frontend's `isValidReturnToPath`; the server's `!startsWith('//')` was belt-and-suspenders that didn't add real coverage. Dropped it in f06cf1cdf6.

(I'd keep it in `isSafeUrl` itself — that one is also used for rich-text URL validation where a protocol-relative `href` would actually navigate cross-origin. Different threat model.)

packages/twenty-server/src/engine/core-modules/auth/services/auth.service.ts
FelixMalfait: Agreed with @FelixMalfait — leaving inline. The expression is now down to a single line per call site (`isNonEmptyString(returnToPath) && returnToPath.startsWith('/')`) so the duplication is light.

packages/twenty-server/src/engine/core-modules/auth/strategies/google.auth.strategy.ts
```diff
@@ -59,6 +60,7 @@ export class GoogleStrategy extends PassportStrategy(Strategy, 'google') {
         workspacePersonalInviteToken: req.query.workspacePersonalInviteToken,
```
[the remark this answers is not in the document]
FelixMalfait: Looks serious but unrelated to my changes, will open a separate PR if true

PR #20538, opened 2026-05-13
fix: address review feedback on relation traversal filter (stacked on #20533)

oscomarch:
Stacked on #20533 — three issues flagged in the review.

## Changes

**1. IS_EMPTY / IS_NOT_EMPTY on relation sub-fields returned wrong filter**

Picking `Company → Name → is empty` was hitting the emptiness short-circuit before the traversal branch ran, producing `{ companyId: { is: NULL } }` instead of `{ company: { name: { is: NULL } } }`. Fixed by resolving `subFieldName` before the emptiness guard and skipping the guard for relation traversals.

**2. Ghost `-1` entry broke keyboard navigation in the relation sub-menu**

`selectableItemIdArray` included `'-1'` (the "Any field" composite item) in the relation branch even though it's never rendered there. First arrow-down was wasted on the invisible item. Removed it.

**3. `isRelationTraversalShape` could misclassify filters on fields named after GraphQL operators**

If a workspace had a custom field named `eq`, `in`, `like`, etc., a plain scalar filter `{ eq: "value" }` would be treated as a relation traversal and fail with a confusing error. Added `GRAPHQL_LEAF_OPERATORS` set; those keys are now excluded from the field-name check.

## Test plan

- [ ] `Company → Name → is empty` filter returns people with no company name, not people with no company
- [ ] Arrow-down in the relation sub-menu lands on the first field immediately
- [ ] Creating a custom field named `eq` and filtering by it does not throw a traversal error

### DIFF
--- a/packages/twenty-front/src/modules/object-record/advanced-filter/components/AdvancedFilterSubFieldSelectMenu.tsx
+++ b/packages/twenty-front/src/modules/object-record/advanced-filter/components/AdvancedFilterSubFieldSelectMenu.tsx
@@ -135,7 +135,7 @@ export const AdvancedFilterSubFieldSelectMenu = ({
     );
 
   const selectableItemIdArray = isRelationSubMenu
-    ? ['-1', ...relationTargetFields.map((field) => field.id)]
+    ? relationTargetFields.map((field) => field.id)
     : ['-1', ...subFieldNames.map((subFieldName) => subFieldName)];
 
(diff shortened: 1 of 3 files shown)


FelixMalfait:
This is quick :), I'm actually changing the implementation, second PR (frontend) not ready for review yet :) - will be much bigger as it involve a change on how view filters are stored

PR #20539, opened 2026-05-13
feat(upgrade): emit structured logfmt logs for upgrade flow

charlesBochet:
## Summary

Adds a small helper that lets every log line in the upgrade flow stay human-readable while emitting a structured tail that Loki / the upgrade-status Grafana dashboard can filter on.

Output shape per `logger.log()` call:

```
<humanMessage as-is, may span multiple lines>
[upgrade] event=<event> key=value …    ← always single line
```

Same call produces **one** structured Loki event regardless of how chatty the human-readable part gets — the dashboard's `|= "[upgrade]"` filter only matches the trailing line.

## Helper API

```ts
formatUpgradeLog({
  humanMessage: string,                  // free-form, multi-line OK, for engineers scrolling raw pod logs
  event: string,                         // required anchor for Loki filtering / dashboards
  logFields?: Record<string,             // short structured key=value tail
    string | number | boolean | null | undefined
  >,
});
```

- `humanMessage` is preserved as-is. A thrown `new Error('line one\nline two')` surfacing through `humanMessage` stays human-readable across multiple lines.
- `logFields` values are logfmt-escaped: whitespace / `=` / `"` get quoted, embedded `\` / `"` / `\n` / `\r` / `\t` are escaped, `null` / `undefined` emit literally (`key=null`, `key=undefined`) instead of being silently dropped — caught via `isDefined` from `twenty-shared/utils`.
- `event` itself runs through the same escape so an event name with whitespace or `=` can't break parsing.

## Example output

```
Initialized upgrade sequence: 8 step(s)
[upgrade] event=sequence.initialized stepCount=8 dryRun=false

Upgrading workspace abc-123 1/10
[upgrade] event=workspace.start workspaceId=abc-123 index=1 total=10 dryRun=false

Upgrade for workspace abc-123 completed.
[upgrade] event=workspace.success workspaceId=abc-123 executedByVersion=1.4.0 dryRun=false

Upgrade summary: 42 workspace(s) succeeded, 1 workspace(s) failed
[upgrade] event=summary totalSuccesses=42 totalFailures=1 dryRun=false

Upgrade failed: Workspace migration runner failed:
  - Option id is required
  - Option id is invalid
[upgrade] event=aborted totalSuccesses=41 totalFailures=2 dryRun=false
```

Loki query for the dashboard: `{namespace="twenty"} |= "[upgrade]" | logfmt event, workspaceId, command, executedByVersion`

## Scope

Only the **upgrade-specific** call sites carry the tag:

- `upgrade.command.ts` — `sequence.initialized`, `sequence.step` (verbose), `summary`, `aborted`
- `upgrade-sequence-runner.service.ts` — `sequence.stopped`, `sequence.aborted`
- `workspace-command-runner.service.ts` — `workspace.start`, `workspace.success`, `cache.invalidate.failed`

`instance-command-runner.service.ts` is intentionally **not** tagged — `runFastInstanceCommand` / `runSlowInstanceCommand` are also invoked from `RunInstanceCommandsCommand` (DB init / `run-instance-commands`), so an `[upgrade]` tag there would mislead at init time. Those lines stay plain-text; stacks still flow on their own via NestJS `logger.error(message, error.stack)`.

`chalk` is dropped from `upgrade.command.ts` — ANSI escapes break log parsers and chalk is a no-op without a TTY anyway.

## Tests

9 inline-snapshot tests in `format-upgrade-log.util.spec.ts` surface the actual output of every interesting shape (summary call site, multi-line humanMessage, quoted / escaped / control-character logField values, null/undefined fields, event name escaping). Snapshots double as documentation of what a real upgrade log line looks like.

## Test plan

- [x] Unit tests green (`jest format-upgrade-log`)
- [x] oxlint + prettier clean
- [x] tsgo typecheck clean on the upgrade module
- [x] CI green
- [ ] Smoke test on staging: run `upgrade` command, confirm `[upgrade]` structured lines surface in Loki and `| logfmt` extracts fields

### DIFF
--- a/packages/twenty-server/src/database/commands/upgrade-version-command/upgrade.command.ts
+++ b/packages/twenty-server/src/database/commands/upgrade-version-command/upgrade.command.ts
@@ -122,6 +122,9 @@ export class UpgradeCommand extends CommandRunner {
       );
     }
 
+    let totalSuccesses = 0;
+    let totalFailures = 0;
+
     try {
       const sequence = this.upgradeSequenceReaderService.getUpgradeSequence();
 
@@ -151,16 +154,17 @@ export class UpgradeCommand extends CommandRunner {
         );
       }
 
-      const { totalSuccesses, totalFailures } =
-        await this.upgradeSequenceRunnerService.run({
-          sequence,
-          options: {
-            ...options,
-            workspaceIds: isDefined(options.workspaceId)
-              ? Array.from(options.workspaceId)
-              : undefined,
-          },
-        });
+      const report = await this.upgradeSequenceRunnerService.run({
+        sequence,
+        options: {
+          ...options,
+          workspaceIds: isDefined(options.workspaceId)
+            ? Array.from(options.workspaceId)
+            : undefined,
+        },
+      });
+      totalSuccesses = report.totalSuccesses;
+      totalFailures = report.totalFailures;
 
       this.logger.log(
         formatUpgradeLog({
@@ -173,12 +177,6 @@ export class UpgradeCommand extends CommandRunner {
           },
         }),
       );
-
-      if (totalFailures > 0) {
-        throw new Error(
-          `Upgrade completed with ${totalFailures} workspace failure(s)`,
-        );
-      }
     } catch (error) {
       const errorMessage =
         error instanceof Error ? error.message : String(error);
@@ -191,6 +189,12 @@ export class UpgradeCommand extends CommandRunner {
         }),
       );
       throw error;
+    }
+
+    if (totalFailures > 0) {
+      throw new Error(
+        `Upgrade completed with ${totalFailures} workspace failure(s)`,
+      );
     }
   }
 }
--- a/packages/twenty-server/src/database/commands/upgrade-version-command/upgrade.command.ts
+++ b/packages/twenty-server/src/database/commands/upgrade-version-command/upgrade.command.ts
@@ -126,19 +126,27 @@ export class UpgradeCommand extends CommandRunner {
       const sequence = this.upgradeSequenceReaderService.getUpgradeSequence();
 
       this.logger.log(
-        formatUpgradeLog('sequence.initialized', {
-          stepCount: sequence.length,
-          dryRun: options.dryRun ?? false,
+        formatUpgradeLog({
+          message: `Initialized upgrade sequence: ${sequence.length} step(s)`,
+          event: 'sequence.initialized',
+          fields: {
+            stepCount: sequence.length,
+            dryRun: options.dryRun ?? false,
+          },
         }),
       );
 
       for (const [index, step] of sequence.entries()) {
         this.logger.verbose(
-          formatUpgradeLog('sequence.step', {
-            index,
-            kind: step.kind,
-            name: step.name,
-            version: step.version,
+          formatUpgradeLog({
+            message: `  [${index}] ${step.kind} — ${step.name} (${step.version})`,
+            event: 'sequence.step',
+            fields: {
+              index,
+              kind: step.kind,
+              name: step.name,
+              version: step.version,
+            },
           }),
         );
       }
@@ -155,10 +163,14 @@ export class UpgradeCommand extends CommandRunner {
         });
 
       this.logger.log(
-        formatUpgradeLog('summary', {
-          totalSuccesses,
-          totalFailures,
-          dryRun: options.dryRun ?? false,
+        formatUpgradeLog({
+          message: `Upgrade summary: ${totalSuccesses} workspace(s) succeeded, ${totalFailures} workspace(s) failed`,
+          event: 'summary',
+          fields: {
+            totalSuccesses,
+            totalFailures,
+            dryRun: options.dryRun ?? false,
+          },
         }),
       );
 
@@ -168,9 +180,14 @@ export class UpgradeCommand extends CommandRunner {
         );
       }
     } catch (error) {
+      const errorMessage =
+        error instanceof Error ? error.message : String(error);
+
       this.logger.error(
-        formatUpgradeLog('aborted', {
-          error: error instanceof Error ? error.message : String(error),
+        formatUpgradeLog({
+          message: `Upgrade failed: ${errorMessage}`,
+          event: 'aborted',
+          fields: { error: errorMessage },
         }),
       );
       throw error;
--- a/packages/twenty-server/src/database/commands/upgrade-version-command/upgrade.command.ts
+++ b/packages/twenty-server/src/database/commands/upgrade-version-command/upgrade.command.ts
@@ -122,9 +122,6 @@ export class UpgradeCommand extends CommandRunner {
       );
     }
 
-    let totalSuccesses = 0;
-    let totalFailures = 0;
-
     try {
       const sequence = this.upgradeSequenceReaderService.getUpgradeSequence();
 
@@ -154,17 +151,16 @@ export class UpgradeCommand extends CommandRunner {
         );
       }
 
-      const report = await this.upgradeSequenceRunnerService.run({
-        sequence,
-        options: {
-          ...options,
-          workspaceIds: isDefined(options.workspaceId)
-            ? Array.from(options.workspaceId)
-            : undefined,
-        },
-      });
-      totalSuccesses = report.totalSuccesses;
-      totalFailures = report.totalFailures;
+      const { totalSuccesses, totalFailures } =
+        await this.upgradeSequenceRunnerService.run({
+          sequence,
+          options: {
+            ...options,
+            workspaceIds: isDefined(options.workspaceId)
+              ? Array.from(options.workspaceId)
+              : undefined,
+          },
+        });
 
       this.logger.log(
         formatUpgradeLog({
@@ -177,6 +173,12 @@ export class UpgradeCommand extends CommandRunner {
           },
         }),
       );
+
+      if (totalFailures > 0) {
+        throw new Error(
+          `Upgrade completed with ${totalFailures} workspace failure(s)`,
+        );
+      }
     } catch (error) {
       const errorMessage =
         error instanceof Error ? error.message : String(error);
@@ -189,12 +191,6 @@ export class UpgradeCommand extends CommandRunner {
         }),
       );
       throw error;
-    }
-
-    if (totalFailures > 0) {
-      throw new Error(
-        `Upgrade completed with ${totalFailures} workspace failure(s)`,
-      );
     }
   }
 }
--- a/packages/twenty-server/src/database/commands/upgrade-version-command/upgrade.command.ts
+++ b/packages/twenty-server/src/database/commands/upgrade-version-command/upgrade.command.ts
@@ -127,9 +127,9 @@ export class UpgradeCommand extends CommandRunner {
 
       this.logger.log(
         formatUpgradeLog({
-          message: `Initialized upgrade sequence: ${sequence.length} step(s)`,
+          humanMessage: `Initialized upgrade sequence: ${sequence.length} step(s)`,
           event: 'sequence.initialized',
-          fields: {
+          logFields: {
             stepCount: sequence.length,
             dryRun: options.dryRun ?? false,
           },
@@ -139,9 +139,9 @@ export class UpgradeCommand extends CommandRunner {
       for (const [index, step] of sequence.entries()) {
         this.logger.verbose(
           formatUpgradeLog({
-            message: `  [${index}] ${step.kind} — ${step.name} (${step.version})`,
+            humanMessage: `  [${index}] ${step.kind} — ${step.name} (${step.version})`,
             event: 'sequence.step',
-            fields: {
+            logFields: {
               index,
               kind: step.kind,
               name: step.name,
@@ -164,9 +164,9 @@ export class UpgradeCommand extends CommandRunner {
 
       this.logger.log(
         formatUpgradeLog({
-          message: `Upgrade summary: ${totalSuccesses} workspace(s) succeeded, ${totalFailures} workspace(s) failed`,
+          humanMessage: `Upgrade summary: ${totalSuccesses} workspace(s) succeeded, ${totalFailures} workspace(s) failed`,
           event: 'summary',
-          fields: {
+          logFields: {
             totalSuccesses,
             totalFailures,
             dryRun: options.dryRun ?? false,
@@ -185,9 +185,8 @@ export class UpgradeCommand extends CommandRunner {
 
       this.logger.error(
         formatUpgradeLog({
-          message: `Upgrade failed: ${errorMessage}`,
+          humanMessage: `Upgrade failed: ${errorMessage}`,
           event: 'aborted',
-          fields: { error: errorMessage },
         }),
       );
       throw error;
--- a/packages/twenty-server/src/database/commands/upgrade-version-command/upgrade.command.ts
+++ b/packages/twenty-server/src/database/commands/upgrade-version-command/upgrade.command.ts
@@ -1,10 +1,10 @@
-import chalk from 'chalk';
 import { Command, CommandRunner, Option } from 'nest-commander';
 import { isDefined } from 'twenty-shared/utils';
 
 import { CommandLogger } from 'src/database/commands/logger';
 import { UpgradeSequenceReaderService } from 'src/engine/core-modules/upgrade/services/upgrade-sequence-reader.service';
 import { UpgradeSequenceRunnerService } from 'src/engine/core-modules/upgrade/services/upgrade-sequence-runner.service';
+import { formatUpgradeLog } from 'src/engine/core-modules/upgrade/utils/format-upgrade-log.util';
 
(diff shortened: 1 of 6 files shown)


packages/twenty-server/src/engine/core-modules/upgrade/utils/format-upgrade-log.util.ts
```diff
@@ -0,0 +1,36 @@
+type UpgradeLogValue = string | number | boolean | null | undefined;
+
+type UpgradeLogFields = Record<string, UpgradeLogValue>;
+
+const UPGRADE_LOG_PREFIX = '[upgrade]';
+
+const formatValue = (value: UpgradeLogValue): string | undefined => {
+  if (value === undefined || value === null) {
```
charlesBochet: always use typeguards

packages/twenty-server/src/engine/core-modules/upgrade/utils/format-upgrade-log.util.ts
```diff
@@ -0,0 +1,36 @@
...
+  const stringified = String(value);
+
+  if (/[\s"=]/.test(stringified)) {
+    return `"${stringified.replace(/\\/g, '\\\\').replace(/"/g, '\\"')}"`;
+  }
+
+  return stringified;
+};
+
+export const formatUpgradeLog = (
```
charlesBochet: I'm feeling we are re-inventing the wheel here, is that the best approach?

packages/twenty-server/src/engine/core-modules/upgrade/utils/format-upgrade-log.util.ts
```diff
@@ -0,0 +1,48 @@
...
+
+type FormatUpgradeLogParams = {
+  message: string;
+  event: string;
+  fields?: UpgradeLogFields;
+};
+
+const UPGRADE_LOG_PREFIX = '[upgrade]';
+
+const escapeForLogfmt = (value: UpgradeLogScalar): string => {
```
charlesBochet: no abbreviation

packages/twenty-server/src/engine/core-modules/upgrade/utils/format-upgrade-log.util.ts
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Emits a log line that reads naturally for humans yet is parseable by Loki's
+ * `| logfmt` decoder. The shape is:
+ *
+ *   [upgrade] <message> | event=<event> key=value ...
+ *
+ * The free-text message keeps a `git log`-style narrative for engineers
+ * scrolling pod logs; the structured tail lets the dashboard panel pull out
+ * `event`, `workspaceId`, `command`, `error`, etc.
+ */
```
charlesBochet: remove comment

packages/twenty-server/src/engine/core-modules/upgrade/utils/format-upgrade-log.util.ts
charlesBochet: Addressed in 149e425801 — see commit message for the breakdown of fixes.

packages/twenty-server/src/engine/core-modules/upgrade/utils/format-upgrade-log.util.ts
charlesBochet: Addressed in 149e425801 — see commit message for the breakdown of fixes.

packages/twenty-server/src/engine/core-modules/upgrade/utils/format-upgrade-log.util.ts
```diff
@@ -0,0 +1,36 @@
...
+
+const formatValue = (value: UpgradeLogValue): string | undefined => {
+  if (value === undefined || value === null) {
+    return undefined;
+  }
+
+  const stringified = String(value);
+
+  if (/[\s"=]/.test(stringified)) {
+    return `"${stringified.replace(/\\/g, '\\\\').replace(/"/g, '\\"')}"`;
```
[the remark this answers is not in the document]
charlesBochet: Addressed in 149e425801 — see commit message for the breakdown of fixes.

packages/twenty-server/src/engine/core-modules/upgrade/utils/format-upgrade-log.util.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
```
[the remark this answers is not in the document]
charlesBochet: Addressed in 149e425801 — see commit message for the breakdown of fixes.

packages/twenty-server/src/engine/core-modules/upgrade/utils/format-upgrade-log.util.ts
```diff
@@ -0,0 +1,36 @@
...
+  event: string,
+  fields: UpgradeLogFields = {},
+): string => {
+  const parts = [`event=${event}`];
+
+  for (const [key, value] of Object.entries(fields)) {
+    const formatted = formatValue(value);
+
+    if (formatted !== undefined) {
+      parts.push(`${key}=${formatted}`);
```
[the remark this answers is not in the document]
charlesBochet: Addressed in 149e425801 — see commit message for the breakdown of fixes.

packages/twenty-server/src/engine/core-modules/upgrade/services/workspace-command-runner.service.ts
```diff
@@ -66,9 +78,10 @@ export class WorkspaceCommandRunnerService {
       await this.upgradeStatusService.invalidateInstanceAndAllWorkspacesStatus();
     } catch (error) {
       this.logger.warn(
-        `Failed to invalidate upgrade-status cache for workspace ${workspaceId}: ${
-          error instanceof Error ? error.message : String(error)
-        }`,
+        formatUpgradeLog('cache.invalidate.failed', {
+          workspaceId,
```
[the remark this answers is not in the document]
charlesBochet: Addressed in 149e425801 — see commit message for the breakdown of fixes.

packages/twenty-server/src/engine/core-modules/upgrade/utils/format-upgrade-log.util.ts
charlesBochet: Addressed in 149e425801 — see commit message for the breakdown of fixes.

packages/twenty-server/src/engine/core-modules/upgrade/utils/format-upgrade-log.util.ts
charlesBochet: Addressed in 149e425801 — see commit message for the breakdown of fixes.

packages/twenty-server/src/engine/core-modules/upgrade/services/workspace-command-runner.service.ts
```diff
@@ -118,6 +149,24 @@ export class WorkspaceCommandRunnerService {
         });
       }
 
+      const errorMessage =
```
charlesBochet: why are we adding a log?

charlesBochet: Right — `WorkspaceIteratorService.iterate` already logs per-workspace errors at the end of the loop (including unwrapped `WorkspaceMigrationRunnerException` causes). The only thing my extra log added on top was the `[upgrade]` tag + structured fields, which let the dashboard's 'failures only' filter match per-workspace rows.

Dropped in 363e90fd9c — the iterator's log is now the single source for per-workspace failures. The dashboard's 'failures only' filter still matches `instance.failed`, `instance.data_migration.failed`, `aborted` and the aggregate `sequence.aborted`; just no per-workspace structured rows. If we later want those in Grafana, I'll extend the dashboard query to also `|~ "Error in workspace "` rather than re-adding the log here.

packages/twenty-server/src/database/commands/upgrade-version-command/upgrade.command.ts
```diff
@@ -122,47 +122,79 @@ export class UpgradeCommand extends CommandRunner {
...
+      const report = await this.upgradeSequenceRunnerService.run({
+        sequence,
+        options: {
+          ...options,
+          workspaceIds: isDefined(options.workspaceId)
+            ? Array.from(options.workspaceId)
+            : undefined,
+        },
+      });
+      totalSuccesses = report.totalSuccesses;
```
charlesBochet: why not destructuring like it was before

packages/twenty-server/src/database/commands/upgrade-version-command/upgrade.command.ts
```diff
@@ -122,47 +122,79 @@ export class UpgradeCommand extends CommandRunner {
...
-        chalk.blue(
-          `Upgrade summary: ${totalSuccesses} workspace(s) succeeded, ${totalFailures} workspace(s) failed`,
-        ),
+        }),
       );
 
-      if (totalFailures > 0) {
-        throw new Error(
-          `Upgrade completed with ${totalFailures} workspace failure(s)`,
+      for (const [index, step] of sequence.entries()) {
```
charlesBochet, line 139: also why inverting?

packages/twenty-server/src/database/commands/upgrade-version-command/upgrade.command.ts
charlesBochet: You're right. Sentry-bot flagged that partial failures log `event=aborted` in addition to `event=summary` (because the `if (totalFailures > 0) throw` rethrows into our own catch), and I restructured to fix that. Out of scope — reverted in 5c9ab7a4cf, destructuring back. Partial-failure runs will still emit both events; the structured counts in `event=summary` keep dashboards correct.

packages/twenty-server/src/database/commands/upgrade-version-command/upgrade.command.ts
charlesBochet: Reverted in 5c9ab7a4cf — control flow back to the original (try/throw/catch in place).

packages/twenty-server/src/engine/core-modules/upgrade/services/instance-command-runner.service.ts
```diff
@@ -144,9 +182,21 @@ export class InstanceCommandRunnerService {
...
 
       try {
-        this.logger.log(`${name} starting data migration...`);
+        this.logger.log(
+          formatUpgradeLog({
+            message: `${name} starting data migration...`,
+            event: 'instance.data_migration.start',
+            fields: { command: name },
+          }),
+        );
```
prastoin: just so you know instance command are run at db init which is kinda upgrade agnostic
same for the run-instance-commands command

packages/twenty-server/src/engine/core-modules/upgrade/utils/__tests__/format-upgrade-log.util.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  it('should escape an event name containing whitespace or =', () => {
+    expect(
+      formatUpgradeLog({
+        message: 'Something happened',
+        event: 'weird event=with-equals',
+      }),
+    ).toBe('[upgrade] Something happened | event="weird event=with-equals"');
+  });
+});
```
prastoin, line 149: Would have used snapshots, and covered complexier escaped values

packages/twenty-server/src/engine/core-modules/upgrade/utils/format-upgrade-log.util.ts
```diff
@@ -0,0 +1,48 @@
...
+  if (!NEEDS_QUOTING.test(raw) && !CONTROL_CHARACTERS.test(raw)) {
+    return raw;
+  }
+
+  const escaped = raw
+    .replace(/\\/g, '\\\\')
+    .replace(/"/g, '\\"')
+    .replace(/\n/g, '\\n')
+    .replace(/\r/g, '\\r')
+    .replace(/\t/g, '\\t');
```
prastoin, line 30: Not sure how this would behave with workspace migration orchestrator report error ?

packages/twenty-server/src/engine/core-modules/upgrade/utils/format-upgrade-log.util.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export const formatUpgradeLog = ({
+  message,
+  event,
+  fields = {},
+}: FormatUpgradeLogParams): string => {
+  const tailParts: string[] = [`event=${escapeLogValue(event)}`];
+
+  for (const [key, value] of Object.entries(fields)) {
+    if (!isDefined(value)) continue;
```
prastoin: we might wanna log null or undefined no ?

packages/twenty-server/src/engine/core-modules/upgrade/utils/__tests__/format-upgrade-log.util.spec.ts
prastoin: for example
```json
{
      "code": "METADATA_VALIDATION_FAILED",
      "errors": {
        "fieldMetadata": [
          {
            "errors": [
              {
                "code": "INVALID_FIELD_INPUT",
                "message": "Option id is required",
                "userFriendlyMessage": "Option id is required",
              },
              {
                "code": "INVALID_FIELD_INPUT",
                "message": "Option id is invalid",
                "userFriendlyMessage": "Option id is invalid",
                "value": null,
              },
            ],
            "flatEntityMinimalInformation": {
              "name": "testField",
              "objectMetadataUniversalIdentifier": Any<String>,
              "universalIdentifier": Any<String>,
            },
            "metadataName": "fieldMetadata",
            "status": "fail",
            "type": "update",
          },
        ],
      },
      "message": "Validation failed for 1 fieldMetadata",
      "summary": {
        "fieldMetadata": 1,
        "totalErrors": 1,
      },
      "userFriendlyMessage": "Option id is required",
    },
    "message": "Multiple validation errors occurred while updating field",
    "name": "GraphQLError",
  }
```

packages/twenty-server/src/engine/core-modules/upgrade/services/instance-command-runner.service.ts
charlesBochet: Good catch. Reverted instance-command-runner.service.ts back to plain-text logs in e118c9cf82 — `runFastInstanceCommand` / `runSlowInstanceCommand` are also called from `RunInstanceCommandsCommand` (DB init / `run-instance-commands`), so tagging those with `[upgrade]` would mislead operators reading init logs. The `[upgrade]` structured events now only come from upgrade-specific call sites (`upgrade.command`, `upgrade-sequence-runner`, `workspace-command-runner`).

packages/twenty-server/src/engine/core-modules/upgrade/utils/__tests__/format-upgrade-log.util.spec.ts
charlesBochet: Done in e118c9cf82 — converted all 7 existing cases to `toMatchInlineSnapshot()` and added an 8th case using a realistic multi-line METADATA_VALIDATION_FAILED-shaped error payload as the value. The snapshot shows `\\n`-escaped JSON staying on a single Loki line.

packages/twenty-server/src/engine/core-modules/upgrade/utils/format-upgrade-log.util.ts
charlesBochet: Type `UpgradeLogScalar = string | number | boolean` keeps callers from passing an Error/object directly — call sites pass `error.message` (or `String(error)` for non-Errors), which may itself be a multi-line JSON payload like the one you shared below. The CR/LF/tab escaping introduced for the cubic/Copilot comment keeps that payload on a single Loki line. New snapshot test built from your METADATA_VALIDATION_FAILED example demonstrates the actual output in e118c9cf82.

packages/twenty-server/src/engine/core-modules/upgrade/utils/format-upgrade-log.util.ts
charlesBochet: Agreed. Changed the helper in e118c9cf82 to emit `key=null` / `key=undefined` literally instead of dropping the field. New 'emits null and undefined fields explicitly' snapshot covers it.

packages/twenty-server/src/engine/core-modules/upgrade/utils/__tests__/format-upgrade-log.util.spec.ts
charlesBochet: Used a close paraphrase of this exact payload as the 8th snapshot test in e118c9cf82 — see the `handles a realistic multi-line orchestrator validation error` case. The full payload survives as a single Loki line with `\\n`-escaped contents.

prastoin (commented):
As discussed feel like we need two log modes:
- human readable
- one liner

As you said especially for self hosts

FelixMalfait (approved):

PR #20544, opened 2026-05-13
Fix time consuming search ilike fallback

Weiko:
## Context
When the tsvector full-text search returns 0 hits on the first page, SearchService falls back to ILIKE '%word%' over searchVector::text. The leading wildcard makes the GIN index unusable, so it seq-scans the table. 
On large searchable custom objects (e.g. a workspace with ~500k rows in _logs) a single fallback can take 2–3s, multiplied across all searchable objects in one request.

## Implementation
Wrap the fallback query in a tiny TypeORM transaction and apply a Postgres per-statement timeout via set_config('statement_timeout', ms, true) (= SET LOCAL). On timeout, Postgres throws 57014 (QUERY_CANCELED); we catch it, warn-log with workspace/object context, and return [] for that object 

## Note
This PR bounds the slow fallback and doesn't make it fast. The right structural fix is to let the fallback use an index. Since tsvector does not work with certain language (which is the reason why the ILIKE fallback was implemented in the first place), we should probably use the pg_trgm extension instead (@FelixMalfait)

### DIFF
--- a/packages/twenty-server/src/engine/api/graphql/workspace-query-runner/utils/is-query-canceled-error.util.ts
+++ b/packages/twenty-server/src/engine/api/graphql/workspace-query-runner/utils/is-query-canceled-error.util.ts
@@ -1,0 +1,6 @@
+import { POSTGRESQL_ERROR_CODES } from 'src/engine/api/graphql/workspace-query-runner/constants/postgres-error-codes.constants';
+
+export const isQueryCanceledError = (error: unknown): boolean =>
+  typeof error === 'object' &&
+  error !== null &&
+  (error as { code?: string }).code === POSTGRESQL_ERROR_CODES.QUERY_CANCELED;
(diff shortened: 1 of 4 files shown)


packages/twenty-server/src/engine/api/graphql/workspace-query-runner/utils/is-query-canceled-error.util.ts
```diff
@@ -0,0 +1,6 @@
+import { POSTGRESQL_ERROR_CODES } from 'src/engine/api/graphql/workspace-query-runner/constants/postgres-error-codes.constants';
+
+export const isQueryCanceledError = (error: unknown): boolean =>
+  typeof error === 'object' &&
```
charlesBochet: use typeGuard isDefined, etc.

charlesBochet (approved):

PR #20476, opened 2026-05-12
security: harden CI against supply-chain attacks

neo773:
- Pin all third-party actions to SHA
- Gate claude.yml triggers to internal authors with Harden-Runner egress audit
- Ignore fork-PR lifecycle scripts
- Narrow cross-repo dispatch payloads
- Add 7d npm release-age gate
- Add CODEOWNERS on .github/** and .yarnrc.yml

PR #20564, opened 2026-05-14
Regrant id token write to claude for oidc swap

prastoin:


.github/workflows/claude.yml
```diff
@@ -51,6 +51,7 @@ jobs:
       contents: write
       pull-requests: write
       issues: write
+      id-token: write
```
charlesBochet, line 54: why does claude needs access to OIDC?

prastoin: to swap for an anthropic token

prastoin: https://github.com/twentyhq/twenty/actions/runs/25843368469/job/75933222570

prastoin: Introduced when I've minimized workflow permissions in https://github.com/twentyhq/twenty/pull/20476/changes#diff-53fec9c265fdba82268df5cd90315b8da57cce43d8e20efbf5c0bdcd1de1342eR15

charlesBochet (approved):

PR #20568, opened 2026-05-14
feat: categorize filterable fields

camirian:
Groups filterable fields into Standard and Custom categories for better UX in the filter dropdown. /claim


FelixMalfait:
This is not something we want to do. Thanks

FelixMalfait:
Is there a reason/context behind why you would want that? Maybe we can help solve it in an other way!

PR #20579, opened 2026-05-14
Fix Gmail folder backfill when enabling sync

### LINKED ISSUE #17095: Messaging: Toggling Gmail folder sync ON doesn't backfill historical messages
When a user toggles a Gmail folder/label to sync after initial account connection, historical messages in that folder are not imported.

**Root cause:**
Gmail's `syncCursor` (historyId) is stored at the `messageChannel` level. The History API only returns changes *after* that historyId - it doesn't return messages that already existed in the folder.

**Expected behavior:**
Toggling a folder ON should trigger a full fetch of that folder's existing messages via `messages.list`, then continue with incremental sync.

**Affected providers:**
Gmail only. Microsoft and IMAP store `syncCursor` per-folder so they handle this correctly.

kkudumu:
## What changed

Fixes #17095.

When a Gmail folder is toggled from unsynced to synced, the service now marks only that folder for a one-time backfill by clearing the folder-level `syncCursor` and moving its Google message channel back to `MESSAGE_LIST_FETCH_PENDING` in the same transaction.

On the next Gmail list fetch, folders with `syncCursor === ''` are queried with a targeted Gmail label search. The existing channel-level history cursor is preserved, so normal incremental history sync continues from the previous cursor instead of triggering a broad full-channel rescan.

## Notes

- Scoped to Google channels only. IMAP/Microsoft are left unchanged because their folder-level cursor behavior already differs.
- No backfill marker is written when folders are disabled or when the folder was already synced.
- The folder update, folder backfill marker, and channel scheduling update are persisted in one transaction.

## Verification

- `PATH=/tmp/node24/bin:$PATH ../../node_modules/.bin/jest --config ./jest.config.mjs --runInBand --runTestsByPath src/engine/metadata-modules/message-folder/message-folder-metadata.service.spec.ts src/modules/messaging/message-import-manager/drivers/gmail/services/gmail-get-message-list.service.spec.ts`
- `PATH=/tmp/node24/bin:$PATH npx oxlint --type-aware -c .oxlintrc.json src/engine/metadata-modules/message-folder/message-folder-metadata.service.ts src/engine/metadata-modules/message-folder/message-folder-metadata.service.spec.ts src/modules/messaging/message-import-manager/drivers/gmail/services/gmail-get-message-list.service.ts src/modules/messaging/message-import-manager/drivers/gmail/services/gmail-get-message-list.service.spec.ts`
- `PATH=/tmp/node24/bin:$PATH npx tsgo -p tsconfig.json --noEmit`
- `PATH=/tmp/node24/bin:$PATH ./node_modules/.bin/prettier --check packages/twenty-server/src/engine/metadata-modules/message-folder/message-folder-metadata.service.ts packages/twenty-server/src/engine/metadata-modules/message-folder/message-folder-metadata.service.spec.ts packages/twenty-server/src/modules/messaging/message-import-manager/drivers/gmail/services/gmail-get-message-list.service.ts packages/twenty-server/src/modules/messaging/message-import-manager/drivers/gmail/services/gmail-get-message-list.service.spec.ts`
- `git diff --check`

### DIFF
--- a/packages/twenty-server/src/engine/metadata-modules/message-folder/message-folder-metadata.service.spec.ts
+++ b/packages/twenty-server/src/engine/metadata-modules/message-folder/message-folder-metadata.service.spec.ts
@@ -1,0 +1,216 @@
+import { Test, type TestingModule } from '@nestjs/testing';
+import { getRepositoryToken } from '@nestjs/typeorm';
+
+import {
+  ConnectedAccountProvider,
+  MessageChannelSyncStage,
+  MessageFolderPendingSyncAction,
+} from 'twenty-shared/types';
+import { In } from 'typeorm';
+
+import { ConnectedAccountMetadataService } from 'src/engine/metadata-modules/connected-account/connected-account-metadata.service';
+import { MessageChannelMetadataService } from 'src/engine/metadata-modules/message-channel/message-channel-metadata.service';
+import { MessageFolderEntity } from 'src/engine/metadata-modules/message-folder/entities/message-folder.entity';
+import { MessageFolderMetadataService } from 'src/engine/metadata-modules/message-folder/message-folder-metadata.service';
+
+describe('MessageFolderMetadataService', () => {
+  let service: MessageFolderMetadataService;
+  let messageFolderRepository: {
+    find: jest.Mock;
+    findOne: jest.Mock;
+    findOneOrFail: jest.Mock;
+    update: jest.Mock;
+  };
+  let messageChannelMetadataService: {
+    findById: jest.Mock;
+    update: jest.Mock;
+  };
+  let connectedAccountMetadataService: {
+    findById: jest.Mock;
+    getUserConnectedAccountIds: jest.Mock;
+  };
+
+  beforeEach(async () => {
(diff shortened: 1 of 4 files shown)


charlesBochet:
I feel it might be very compute expansive but might work. What do you think @neo773 ?

kkudumu:
Thanks for the review. I tightened the implementation so enabling a folder now resets only that Gmail label's history cursor, rather than forcing a whole-mailbox backfill. The folder sync update and channel cursor reset are also wrapped in the same transaction now, which addresses the atomicity issue Cubic flagged on the earlier commit.

Current CI note: the remaining red server shard is `test/integration/metadata/suites/field-metadata/relation/failing-field-metadata-relation-creation.integration-spec.ts`, failing on a duplicate `IDX_FIELD_METADATA_NAME_OBJECT_METADATA_ID_WORKSPACE_ID_UNIQUE` metadata-field constraint. This PR only changes the message folder metadata service and Gmail message-list service/specs, so I do not see a causal path from the Gmail label-history change to that metadata relation snapshot failure.

PR #20580, opened 2026-05-14
feat(secret-encryption): drop APP_SECRET from approved-access-domain validation and session cookies

charlesBochet:
## Summary

Continues retiring `APP_SECRET` as a hot signing secret (after the TOTP migration in #20577). This PR moves the last two cryptographic uses of `APP_SECRET` off it:

1. **Approved-access-domain validation tokens** — was a one-shot `sha256(JSON.stringify({id, domain, key: APP_SECRET}))` HMAC with no built-in expiry. Now a JWT signed by the workspace `signingKey` with a 7-day expiry and claims bound to `approvedAccessDomainId`, `workspaceId`, and `domain`.
2. **Express-session cookie signing** — was `sha256(APP_SECRET || 'SESSION_STORE_SECRET')`. Now `HKDF(ENCRYPTION_KEY, info='twenty:hmac:v1:session-cookie')` with `FALLBACK_ENCRYPTION_KEY` supported for rotation.

### Approved-access-domain — strict cutover

- `ApprovedAccessDomainService.mintValidationToken` issues a JWT via `JwtWrapperService.signAsyncOrThrow` (workspace `signingKey`, asymmetric ES256 with kid-based rotation built in).
- `validateApprovedAccessDomain` verifies the JWT, asserts `type === APPROVED_ACCESS_DOMAIN`, cross-checks `claim.approvedAccessDomainId` against the URL's `approvedAccessDomainId`, then re-checks `domain` and `workspaceId` against the stored row. Any failure maps to `APPROVED_ACCESS_DOMAIN_VALIDATION_TOKEN_INVALID`.
- **No legacy fallback:** any pending invitation link minted with the old SHA hash will fail validation and must be re-sent. Volume is small and admins can re-issue from settings — this is the cleanest cutover.

### Session cookies — bridged cutover

- `resolveSessionCookieSecretsOrThrow` returns an array `[HKDF(ENCRYPTION_KEY), HKDF(FALLBACK_ENCRYPTION_KEY)?, sha256(APP_SECRET || 'SESSION_STORE_SECRET')?]`.
- `express-session` signs new cookies with the first secret and verifies against any entry, so in-flight cookies signed under the legacy SHA keep verifying until `maxAge` (30 min) expires.
- New `deriveInstanceHmacKey` HKDF utility uses a dedicated `twenty:hmac:v1:` info prefix — distinct from the AEAD subkey prefix `twenty:enc:v2:` — so HMAC and encryption subkeys can never collide for the same raw `ENCRYPTION_KEY`.
- TODO comment marks the legacy slot for removal post-2.5.

### Notes on rotation behaviour

- Rotating `ENCRYPTION_KEY` while keeping the old value in `FALLBACK_ENCRYPTION_KEY` keeps cookies signed under either key verifying. New cookies sign under the new key. After all in-flight cookies expire (≤30 min), the fallback slot can be dropped from env.
- Rotating the workspace `signingKey` (already supported by `JwtKeyManagerService`) keeps already-issued approved-access-domain JWTs verifying via `kid` until their 7-day expiry.

## Test plan

- [x] Unit tests for `ApprovedAccessDomainService` cover: happy path, JWT verify failure, wrong token type, JWT id ≠ input id, JWT-claimed domain ≠ row, missing row, already-validated row.
- [x] Unit tests for `resolveSessionCookieSecretsOrThrow` cover: throws without keys, primary order (`ENCRYPTION_KEY` → APP_SECRET fallback), `FALLBACK_ENCRYPTION_KEY` placement, empty-string vars treated as unset, legacy slot omitted when `APP_SECRET` missing, HKDF domain separation across purposes.
- [x] `nx lint:diff-with-main twenty-server` — clean.
- [x] Full test surface across approved-access-domain, secret-encryption, session-storage — 78/78 pass.
- [ ] CI green.
- [ ] Manual smoke: boot with a dummy `ENCRYPTION_KEY`, confirm sign-in succeeds (session cookie works), create + validate an approved-access-domain end-to-end through the UI.

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/approved-access-domain/services/approved-access-domain.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/approved-access-domain/services/approved-access-domain.service.ts
@@ -22,6 +22,8 @@ import { WorkspaceDomainsService } from 'src/engine/core-modules/domain/workspac
 import { EmailService } from 'src/engine/core-modules/email/email.service';
 import { FileUrlService } from 'src/engine/core-modules/file/file-url/file-url.service';
 import { JwtWrapperService } from 'src/engine/core-modules/jwt/services/jwt-wrapper.service';
+import { decodeJwtHeader } from 'src/engine/core-modules/jwt/utils/decode-jwt-header.util';
+import { isAsymmetricJwtHeader } from 'src/engine/core-modules/jwt/utils/is-asymmetric-jwt-header.util';
 import { TwentyConfigService } from 'src/engine/core-modules/twenty-config/twenty-config.service';
 import { WorkspaceEntity } from 'src/engine/core-modules/workspace/workspace.entity';
 import { WorkspaceMemberWorkspaceEntity } from 'src/modules/workspace-member/standard-objects/workspace-member.workspace-entity';
@@ -144,6 +146,13 @@ export class ApprovedAccessDomainService {
   private async verifyValidationTokenOrThrow(
     validationToken: string,
   ): Promise<ApprovedAccessDomainJwtPayload> {
+    if (!isAsymmetricJwtHeader(decodeJwtHeader(validationToken))) {
+      throw new ApprovedAccessDomainException(
+        'Invalid approved access domain validation token',
+        ApprovedAccessDomainExceptionCode.APPROVED_ACCESS_DOMAIN_VALIDATION_TOKEN_INVALID,
+      );
+    }
+
     let payload: ApprovedAccessDomainJwtPayload;
 
     try {
--- a/packages/twenty-server/src/engine/core-modules/approved-access-domain/services/approved-access-domain.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/approved-access-domain/services/approved-access-domain.service.ts
@@ -1,7 +1,5 @@
-import { Injectable } from '@nestjs/common';
+import { Injectable, Logger } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
-
-import crypto from 'crypto';
 
 import { msg } from '@lingui/core/macro';
 import { render } from '@react-email/render';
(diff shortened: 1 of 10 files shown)


packages/twenty-server/src/engine/core-modules/secret-encryption/utils/resolve-session-cookie-secrets.util.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  if (!isNonEmptyString(rawPrimary)) {
+    throw new Error(
+      'Cannot derive session cookie secret: set ENCRYPTION_KEY (or APP_SECRET for legacy deployments).',
+    );
+  }
+
+  const secrets: string[] = [
+    deriveInstanceHmacKey({
+      rawKey: rawPrimary,
```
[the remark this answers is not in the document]
charlesBochet: Intentional — when `ENCRYPTION_KEY` is unset, `resolveEncryptionKeysOrThrow` already falls back to `APP_SECRET` as the primary at-rest encryption key (otherwise pre-rotation deployments would fail to boot). This util mirrors that contract for consistency: same precedence rule, same env-var semantics, no surprise behavior between encryption and session signing.

The cutover this PR delivers is structural rather than secret-substitution:

1. New cookies are signed by an **HKDF-derived subkey** with `info='twenty:hmac:v1:session-cookie'`, not by `APP_SECRET` directly. That domain-separates the cookie-signing key from the at-rest AEAD subkey (different `info` prefix, `twenty:enc:v2:` vs `twenty:hmac:v1:`), so a leak on one side no longer leaks the other — which was the original concern with `sha256(APP_SECRET + 'SESSION_STORE_SECRET')`.
2. Once an operator sets `ENCRYPTION_KEY`, `APP_SECRET` immediately stops being the signing input and only remains the legacy verification slot until in-flight cookies expire (30 min).

Hard-failing without `ENCRYPTION_KEY` is the right end-state but belongs to a follow-up PR (along with the same hard-fail for at-rest encryption) so we can stage the env-var rollout for self-hosters.

packages/twenty-server/src/engine/core-modules/approved-access-domain/services/approved-access-domain.service.ts
```diff
@@ -111,19 +124,62 @@ export class ApprovedAccessDomainService {
...
+      throw new ApprovedAccessDomainException(
+        'Invalid approved access domain validation token',
+        ApprovedAccessDomainExceptionCode.APPROVED_ACCESS_DOMAIN_VALIDATION_TOKEN_INVALID,
+      );
+    }
+
+    return payload;
   }
 
   async validateApprovedAccessDomain({
```
FelixMalfait, line 185: Nit: currently you can be logged in to workspaceB and validate the link for workspace A ; not a security issue / something that can be exploited really, but still feels a bit off, maybe we should pass workspaceId along

packages/twenty-server/src/engine/core-modules/twenty-config/config-variables.ts
```diff
@@ -1158,7 +1158,7 @@ export class ConfigVariables {
     group: ConfigVariablesGroup.SERVER_CONFIG,
```
FelixMalfait, line 1158: Let's make sure to update all docker configs and docs soon! I guess you're waiting for the release

FelixMalfait (approved):
Great

packages/twenty-server/src/engine/core-modules/approved-access-domain/services/approved-access-domain.service.ts
charlesBochet: Ok, I will treat this in a follow up PR today!

PR #20586, opened 2026-05-14
feat(admin-panel): signing keys management tab with usage tracking

charlesBochet:
## Summary
- Adds a new admin-only **Security** tab to the Admin Panel (alongside General/Apps/AI/Config/Health) containing a **Signing Keys** section. The tab is intentionally introduced now so the upcoming **Encryption rotation** work can land as a sibling section.
- Lists every JWT signing key with key id, `createdAt`, `revokedAt`, current/active/revoked status, and a **7-day verification count** read from Redis. A trailing row aggregates **legacy HS256** verifications so it is clear when the deprecated path is still in use.
- Lets an admin **revoke** a public key. Revoking the current key drops `isCurrent`, sets `revokedAt`, nulls the encrypted `privateKey` and clears the in-process cached current key; the existing lazy path in `JwtKeyManagerService.getCurrentSigningKey()` then mints a fresh current key on the next sign.

## Backend
- `SigningKeyVerifyCounterService` — bucketed Redis counter under the existing `EngineMetrics` namespace. 1-day UTC-aligned buckets, 8-day TTL refreshed on every increment, batched read via `mget`. Failures are swallowed and logged at `warn` so a Redis hiccup cannot break auth.
- `JwtWrapperService.verifyJwtToken` records verifies **after success** for both ES256 (`kid` as identifier) and HS256 (the literal `legacy` identifier).
- `JwtKeyManagerService.listSigningKeys()` and `revokeSigningKey(id)`: list ordered by `isCurrent DESC, createdAt DESC`; revoke is idempotent, validates the UUID, invalidates the public-key cache, and resets the cached current-key promise.
- `AdminPanelResolver.getSigningKeys` (query) and `revokeSigningKey` (mutation) are both decorated with `@UseGuards(AdminPanelGuard)` so they are admin-only, like the 35 existing admin-only methods on this resolver. `privateKey` is never returned over GraphQL.

## Frontend
- New `SECURITY` tab id wired into `SettingsAdminContent` and `SettingsAdminTabContent` (gated by `canAccessFullAdminPanel`).
- `SettingsAdminSecurity` / `SettingsAdminSigningKeysTable` strictly reuse existing admin-panel components: `Section`, `H2Title`, `Table`/`TableRow`/`TableCell`/`TableHeader` from `@/ui/layout/table`, `Tag`/`Button` from `twenty-ui`, and `ConfirmationModal` mirroring the queue retry/delete modals. Only one minimal styled helper for the monospaced UUID rendering.
- `useRevokeSigningKey` uses `useApolloAdminClient`, refetches `GetSigningKeys`, shows success/error snackbars (same pattern as `useRetryJobs`/`useDeleteJobs`).

image


FelixMalfait:
Great!!! Let's add it in health tab for now? (maybe we can rename it later), not worth a tab imo

packages/twenty-server/src/engine/core-modules/jwt/services/signing-key-verify-counter.service.ts
```diff
@@ -0,0 +1,155 @@
...
+// Note: this counter intentionally uses CacheStorageService directly rather
+// than MetricsCacheService. MetricsCacheService is built around Redis sets
+// keyed by a fixed MetricsKeys enum, dedup-by-eventId, 15s buckets and a
+// minutes-scale window. Per-kid signing key usage needs the opposite:
+// raw integer counters (incrBy), 1-day buckets over a 7-day window, and a
+// dynamic key suffix per kid. We'll generalize MetricsCacheService into a
+// reusable "bucketed raw counter" primitive once a second caller needs the
+// same shape.
+@Injectable()
+export class SigningKeyVerifyCounterService {
```
FelixMalfait: Curious why we need this? Seems like it will be a lot of calls - is it worth it?

packages/twenty-front/src/modules/settings/admin-panel/security/components/SettingsAdminSecurity.tsx
```diff
@@ -0,0 +1,16 @@
+import { SettingsAdminSigningKeysTable } from '@/settings/admin-panel/security/components/SettingsAdminSigningKeysTable';
+import { t } from '@lingui/core/macro';
+import { H2Title } from 'twenty-ui/display';
+import { Section } from 'twenty-ui/layout';
+
+export const SettingsAdminSecurity = () => {
+  return (
+    <Section>
+      <H2Title
+        title={t`Signing Keys`}
```
FelixMalfait: Let's make it a section in health instead of a dedicated tab?

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-key-manager.service.ts
```diff
@@ -63,6 +63,46 @@ export class JwtKeyManagerService {
...
+        {
+          revokedAt: new Date(),
+          isCurrent: false,
+          privateKey: null,
+        },
+      );
+    }
+
+    await this.coreEntityCacheService.invalidate('signingKeyPublicKey', id);
+    this.currentSigningKeyPromise = null;
```
FelixMalfait, line 110: This only reset in the current pod, so it won't be revoked on other pods or workers until they restart (htey have their own JwtKeyManagerService instance). Not a big issue I guess?

FelixMalfait: Maybe a TTL of 1 hour or something this?

packages/twenty-front/src/modules/settings/admin-panel/security/components/SettingsAdminSigningKeysTable.tsx
```diff
@@ -0,0 +1,187 @@
...
+                    ariaLabel={t`Copy key ID`}
+                    onClick={() =>
+                      copyToClipboard(signingKey.id, t`Key ID copied`)
+                    }
+                  />
+                </TableCell>
+                <TableCell>
+                  {isDefined(signingKey.revokedAt) ? (
+                    <OverflowingTextWithTooltip
+                      text={beautifyExactDate(signingKey.revokedAt)}
```
FelixMalfait: Nit: I like "beautifyPastDateRelativeToNow" or something like that as the main text, and then the exact date non-simplifed on over (otherwise here you would just get "Today" and not the exact time)

FelixMalfait (approved):
Great!

packages/twenty-front/src/modules/settings/admin-panel/security/components/SettingsAdminSecurity.tsx
charlesBochet: ok!

PR #20611, opened 2026-05-15
docs(self-host): document ENCRYPTION_KEY, FALLBACK_ENCRYPTION_KEY and key rotation procedures

charlesBochet:
## Summary
- Documents the new at-rest encryption envelope (`ENCRYPTION_KEY` / `FALLBACK_ENCRYPTION_KEY`) introduced in v2.5+ and clarifies its relationship to the legacy `APP_SECRET`-as-encryption-key path.
- Adds a new dedicated **Key rotation** guide covering manual / Enterprise-cron JWT signing-key rotation, signing-key revocation, and the online `ENCRYPTION_KEY` rotation procedure (including the new \`secret-encryption:rotate\` CLI shipped in a follow-up PR).
- Updates the docker-compose quickstart to generate a dedicated \`ENCRYPTION_KEY\` from day 1.
- Mentions the v2.5+ enc:v2 backfill in the upgrade guide.

English-only — the localized mirrors will be picked up by i18n CI.

## Test plan
- [ ] Mintlify build passes locally / in CI
- [ ] Sidebar entry renders under **Self-Host → Key rotation**
- [ ] Internal links to /developers/self-host/capabilities/key-rotation resolve from setup.mdx, docker-compose.mdx and upgrade-guide.mdx

PR #20612, opened 2026-05-15
feat(server): Enterprise cron that rotates the current JWT signing key

charlesBochet:
## Summary
Adds a daily Enterprise-only cron that rotates the current ES256 JWT signing key once it has been current for `SIGNING_KEY_ROTATION_DAYS`. Manual rotation from the admin panel is unaffected.

### Behaviour
- `SIGNING_KEY_ROTATION_DAYS` is **opt-in**: when unset, the cron is a no-op.
- Rotation flips `isCurrent` and clears the previous key's `privateKey` in the same transaction, then inserts the new `isCurrent=true` row.
- The previous key's row is kept (`revokedAt` stays `null`) so its `publicKey` can keep verifying tokens it signed until they expire; only the encrypted `privateKey` is wiped since it can no longer be used to sign.
- **No auto-revocation** — revoking a key remains a manual admin action, reserved for leak / emergency response.
- The cron is also a no-op when `EnterprisePlanService.isValid()` is `false`.

### Wiring
- `JwtKeyManagerService.rotateCurrent()`
- `SigningKeyRotationService.rotateIfDue()` (reads `SIGNING_KEY_ROTATION_DAYS`, skips when unset)
- `RotateSigningKeysCronJob` (Enterprise-gated, rethrows on failure) registered in `JwtModule`
- `RotateSigningKeysCronCommand` registered with `cron:register:all`
- `ROTATE_SIGNING_KEYS_CRON_PATTERN = '15 3 * * *'` (daily, no-op until threshold)

Operator documentation lives in #20611 (docs PR).


packages/twenty-server/src/engine/core-modules/jwt/services/signing-key-rotation.service.ts
```diff
@@ -0,0 +1,65 @@
...
+import { JwtKeyManagerService } from 'src/engine/core-modules/jwt/services/jwt-key-manager.service';
+import { TwentyConfigService } from 'src/engine/core-modules/twenty-config/twenty-config.service';
+
+const ONE_DAY_MS = 24 * 60 * 60 * 1000;
+
+export type SigningKeyRotationResult = {
+  rotated: boolean;
+  previousId: string | null;
+  newId: string | null;
+};
```
prastoin, line 16: ***Polish:*** Assuming we always have a previousId on first rotation
```ts
export type SigningKeyRotationResult = {
  rotated: true;
  previousId: string; // First rota could be null ?
  newId: string;
} | {
  rotated: false;
  previousId: never;
  newId: never;
};;
```

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-key-manager.service.ts
```diff
@@ -78,6 +78,38 @@ export class JwtKeyManagerService {
...
+    const generated = this.generateEcP256KeyPair();
+    const newId = randomUUID();
+
+    await this.signingKeyRepository.manager.transaction(
+      async (entityManager) => {
+        const repository = entityManager.getRepository(SigningKeyEntity);
+
+        await repository.update(
+          { isCurrent: true },
+          { isCurrent: false, privateKey: null },
```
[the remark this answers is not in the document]
prastoin: +1

prastoin (commented):
Critical issue with rotation service erasing previous private key

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-key-manager.service.ts
```diff
@@ -78,6 +78,38 @@ export class JwtKeyManagerService {
...
+          ),
+          isCurrent: true,
+          revokedAt: null,
+        });
+      },
+    );
+
+    await this.coreEntityCacheService.invalidate('signingKeyPublicKey', newId);
+    this.currentSigningKeyPromise = null;
+    this.currentSigningKeyCachedAt = 0;
```
prastoin: Shouldn't this be now ?

packages/twenty-server/test/integration/graphql/suites/auth/jwt-key-rotation/rotate-signing-keys-cron-job.integration-spec.ts
```diff
@@ -0,0 +1,105 @@
...
+    const rotatedKid = decodeJwtCompleteOrThrow(rotatedApiKeyToken).header
+      .kid as string;
+
+    expect(isNonEmptyString(rotatedKid)).toBe(true);
+    expect(rotatedKid).not.toBe(initialKid);
+
+    const callWithPreviousToken = await findManyApplications({
+      accessToken: initialApiKeyToken,
+      expectToFail: false,
+    });
```
prastoin, line 92: should be failing unless I'm mistaken

packages/twenty-server/src/engine/core-modules/jwt/services/signing-key-rotation.service.ts
charlesBochet: it could it some really specific edge case where the currentKey is revoked from admin panel and the cron is triggered. This is edge case, I can also make the service throw but both way looks OK

packages/twenty-server/test/integration/graphql/suites/auth/jwt-key-rotation/rotate-signing-keys-cron-job.integration-spec.ts
charlesBochet: why? privateKey is nulled but we still have publicKey to verify tokens

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-key-manager.service.ts
charlesBochet: this is marking the cache as expired:

In upgcoming calls:
 const isLocalCacheExpired =
      Date.now() - this.currentSigningKeyCachedAt >
      CURRENT_SIGNING_KEY_LOCAL_TTL_MS;
      
Putting 0 is to say that nothing is cached.

If we put Date.now(), this means we would have to also have to compute the cache which is not the case here

packages/twenty-server/test/integration/graphql/suites/auth/jwt-key-rotation/rotate-signing-keys-cron-job.integration-spec.ts
prastoin: I meant regarding the previous bug introduced in the rotation erasing previous privateKEy

packages/twenty-server/src/engine/core-modules/jwt/services/jwt-key-manager.service.ts
prastoin: Might wanna create a dedicated method abstraction as it's quite implicit

charlesBochet: Ok I can do that but that's nitpick, this code is very local and not exposed so easy to refactor / maintain and not bug prone!

charlesBochet: For you, I'm doing it :D
done

prastoin (approved):
Good !

PR #20613, opened 2026-05-15
feat(server): idempotent CLI to rotate ENCRYPTION_KEY across enc:v2 rows

charlesBochet:
## Summary
Adds the \`secret-encryption:rotate\` CLI command, which re-encrypts every at-rest secret stored in an \`enc:v2:\` envelope under the current \`ENCRYPTION_KEY\`. The command is **online** and **resumable**: a SQL filter skips rows already on the current keyId, so interrupting it (Ctrl-C, container restart, …) and re-running picks up where it left off without re-rotating earlier rows.

### Sites covered (one handler each)
| Site | Table.column | Scope |
| --- | --- | --- |
| \`connected-account-tokens\` | \`connectedAccount.{accessToken, refreshToken}\` | workspace |
| \`application-variable\` | \`applicationVariable.value\` (isSecret only) | workspace |
| \`application-registration-variable\` | \`applicationRegistrationVariable.encryptedValue\` | instance |
| \`signing-key-private-keys\` | \`signingKey.privateKey\` | instance |
| \`sensitive-config-storage\` | \`keyValuePair.value\` (isSensitive + STRING configs) | instance |
| \`totp-secrets\` | \`twoFactorAuthenticationMethod.secret\` | workspace |

Each handler:
- Filters at SQL level on \`value LIKE 'enc:v2:%' AND value NOT LIKE 'enc:v2:<primaryKeyId>:%'\` to enforce idempotency without re-decrypting already-rotated rows.
- Uses cursor-based batching (default **200**, capped **5000**).
- Threads \`workspaceId\` into HKDF for workspace-scoped sites; runs instance-scoped for the rest.

### CLI flags
| Flag | Description |
| --- | --- |
| \`-s, --site <site>\` | Limit to a single site. |
| \`-b, --batch-size <n>\` | Override per-batch row count. |
| \`-d, --dry-run\` | Decrypt + re-encrypt in memory, skip the \`UPDATE\`. |

The runner logs progress via Nest \`Logger\` (per-site start, completion, final summary) and exits non-zero when any site reports \`errors > 0\`. \`FALLBACK_ENCRYPTION_KEY\` must be set to the previous \`ENCRYPTION_KEY\` during rotation; the runner warns when it is unset.

Operator documentation lives in #20611 (docs PR).


packages/twenty-server/src/database/commands/secret-encryption-rotation/types/secret-encryption-rotation-handler.type.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export type SecretEncryptionRotationSiteResult =
+  SecretEncryptionRotationOutcome & {
+    siteName: string;
+    remainingBefore: number;
+    durationMs: number;
+  };
+
+export type SecretEncryptionRotationHandler = {
+  readonly siteName: string;
```
prastoin: ***Polish:*** Could be strictly typed as a keyof of the entity

packages/twenty-server/src/database/commands/secret-encryption-rotation/services/secret-encryption-rotation-runner.service.ts
```diff
@@ -0,0 +1,183 @@
...
+  constructor(
+    private readonly environmentConfigDriver: EnvironmentConfigDriver,
+    connectedAccountTokensRotationHandler: ConnectedAccountTokensRotationHandler,
+    applicationVariableRotationHandler: ApplicationVariableRotationHandler,
+    applicationRegistrationVariableRotationHandler: ApplicationRegistrationVariableRotationHandler,
+    signingKeyPrivateKeysRotationHandler: SigningKeyPrivateKeysRotationHandler,
+    sensitiveConfigStorageRotationHandler: SensitiveConfigStorageRotationHandler,
+    totpSecretsRotationHandler: TotpSecretsRotationHandler,
+  ) {
+    this.handlers = [
```
prastoin: ***Polish:*** this could be a strictly typed map
***Extreme polish:*** We could use a signed typescript string that we could discover across all the typeorm entities and strictly requires the maps key to be defined for each metadata

In a nutshell by typing encrypedValue: SecretString, would raise typescript error if the handler is not implemented

packages/twenty-server/src/database/commands/secret-encryption-rotation/services/secret-encryption-rotation-runner.service.ts
```diff
@@ -0,0 +1,183 @@
...
+        `[${handler.siteName}] start: ${remainingBefore} row(s) need rotation`,
+      );
+
+      const { rotated, skipped, errors } = await handler.rotate({
+        currentEncryptionKeyId,
+        batchSize: options.batchSize,
+        dryRun: options.dryRun,
+      });
+
+      const durationMs = Date.now() - siteStartedAt;
```
prastoin: could use literal perf here

packages/twenty-server/src/database/commands/secret-encryption-rotation/handlers/connected-account-tokens-rotation.handler.ts
```diff
@@ -0,0 +1,66 @@
...
+  type SecretEncryptionRotationContext,
+  type SecretEncryptionRotationOutcome,
+} from 'src/database/commands/secret-encryption-rotation/interfaces/secret-encryption-rotation-handler.interface';
+import { SecretEncryptionColumnRotationService } from 'src/database/commands/secret-encryption-rotation/services/secret-encryption-column-rotation.service';
+import { ConnectedAccountEntity } from 'src/engine/metadata-modules/connected-account/entities/connected-account.entity';
+
+const ENCRYPTED_COLUMNS = ['accessToken', 'refreshToken'] as const;
+
+@Injectable()
+export class ConnectedAccountTokensRotationHandler extends SecretEncryptionRotationHandler {
```
prastoin: as discussed we could centralize in a single handler all the identical ones which are not the application variable and the sensitive config

packages/twenty-server/src/database/commands/secret-encryption-rotation/handlers/application-variable-rotation.handler.ts
```diff
@@ -0,0 +1,56 @@
...
+
+import {
+  SecretEncryptionRotationHandler,
+  type SecretEncryptionRotationContext,
+  type SecretEncryptionRotationOutcome,
+} from 'src/database/commands/secret-encryption-rotation/interfaces/secret-encryption-rotation-handler.interface';
+import { SecretEncryptionColumnRotationService } from 'src/database/commands/secret-encryption-rotation/services/secret-encryption-column-rotation.service';
+import { ApplicationVariableEntity } from 'src/engine/core-modules/application/application-variable/application-variable.entity';
+
+const ENCRYPTED_COLUMN = 'value';
```
prastoin: should be strictly typed

packages/twenty-server/test/integration/secret-encryption/secret-encryption-rotation.integration-spec.ts
```diff
@@ -0,0 +1,198 @@
...
+              applicationVariables {
+                key
+                value
+                isSecret
+              }
+            }
+          }
+        `,
+        variables: { id: applicationId },
+      });
```
prastoin: create testing util for all raw mutations here

packages/twenty-server/test/integration/secret-encryption/secret-encryption-rotation.integration-spec.ts
```diff
@@ -0,0 +1,198 @@
...
+  });
+
+  it('exposes the expected stable list of sites', () => {
+    expect(runner.listSiteNames()).toEqual([
+      'connected-account-tokens',
+      'application-variable',
+      'application-registration-variable',
+      'signing-key-private-keys',
+      'sensitive-config-storage',
+      'totp-secrets',
```
prastoin: Could be an application constant ( strictly typed inferred from signature :eyes: )

packages/twenty-server/src/database/commands/secret-encryption-rotation/types/secret-encryption-rotation-handler.type.ts
charlesBochet: made it an abstract class, I will let strict type for later for now

packages/twenty-server/test/integration/secret-encryption/utils/run-secret-encryption-rotation-command.util.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  return args;
+};
+
+export const runSecretEncryptionRotationCommand = async (
+  args: RotateArguments = {},
+): Promise<void> => {
+  await new Promise<void>((resolve, reject) => {
+    const child = spawn('node', [COMMAND_JS_PATH, ...buildArgs(args)], {
```
prastoin, line 38: ***Suggestion:*** I would directly import the command manually trigger it's run

prastoin: testing services instead of commands, allow better grain
```ts
  const runner = module.get(UpgradeSequenceRunnerService);
```

packages/twenty-server/src/database/commands/secret-encryption-rotation/handlers/sensitive-config-storage-rotation.handler.ts
```diff
@@ -0,0 +1,191 @@
...
+  }): Promise<SecretEncryptionRotationOutcome> {
+    const rawValue = row.value as unknown;
+
+    if (
+      typeof rawValue !== 'string' ||
+      !rawValue.startsWith(SECRET_ENCRYPTION_ENVELOPE_V2_PREFIX)
+    ) {
+      return { rotated: 0, skipped: 1, errors: 0 };
+    }
+
```
prastoin, line 124: Should this be an error ?

packages/twenty-server/src/database/commands/secret-encryption-rotation/handlers/column-rotation-site.handler.ts
```diff
@@ -0,0 +1,181 @@
...
+    const { encryptedColumn } = this.config;
+    const rowId = row.id;
+    const currentValue = row[encryptedColumn] as string | null | undefined;
+
+    if (
+      !isDefined(currentValue) ||
+      !currentValue.startsWith(SECRET_ENCRYPTION_ENVELOPE_V2_PREFIX)
+    ) {
+      return { rotated: 0, skipped: 1, errors: 0 };
+    }
```
prastoin, line 128: would return as error

packages/twenty-server/src/database/commands/secret-encryption-rotation/handlers/column-rotation-site.handler.ts
```diff
@@ -0,0 +1,181 @@
...
+    if (!isDefined(this.config.extraWhere)) {
+      return qb;
+    }
+
+    for (const [column, value] of Object.entries(this.config.extraWhere)) {
+      const parameterKey = `extra_${column}`;
+      qb.andWhere(`row.${column} = :${parameterKey}`, {
+        [parameterKey]: value,
+      });
+    }
```
prastoin, line 179: If we historically have in database previous rotation key relicas this would get swallowed here as failed to re-encrypt instead of failed to decrypt

Even though that should not occurs the setup exists

For example while rotating from A to C whereas A to B partially failed
enc:v2:A-***
enc:v2:B***
enc:v2:C-***

prastoin (approved):

packages/twenty-server/test/integration/secret-encryption/utils/run-secret-encryption-rotation-command.util.ts
charlesBochet: Discussed with you, I think this is actually better to trigger the whole command without mocking

packages/twenty-server/src/database/commands/secret-encryption-rotation/handlers/column-rotation-site.handler.ts
charlesBochet: I think it's fine but maybe we can return a clearer error message if we do not recognize the encryption key at all?

packages/twenty-server/src/database/commands/secret-encryption-rotation/handlers/column-rotation-site.handler.ts
charlesBochet: +1

packages/twenty-server/src/database/commands/secret-encryption-rotation/handlers/column-rotation-site.handler.ts
```diff
@@ -0,0 +1,181 @@
...
+      this.logger.warn(
+        `[${this.siteName}] failed to re-encrypt row ${rowId}: ${
+          error instanceof Error ? error.message : String(error)
+        }`,
+      );
+      return { rotated: 0, skipped: 0, errors: 1 };
+    }
+  }
+
+  private applyExtraWhere(
```
prastoin, line 167: not sure why it's required anymore

prastoin (approved):

PR #20617, opened 2026-05-15
fix(front): propagate full selection to non-headless front-component commands

clivemeister:
## Summary

Non-headless `defineFrontComponent({ command: { availabilityType: 'RECORD_SELECTION', ... } })` invocations from the action bar arrive at the component with `selectedRecordIds = []`, even when the user has N records selected. This patch carries the full set through the side-panel pipeline.

## Trace

`useCommandMenuItemClick` takes the `isFrontComponent` branch (not the `shouldMountCommand` headless branch) for these commands. That branch built `recordContext` with a single `recordId`, set only when exactly one record was selected. With N > 1 or N = 0 the field was `undefined`; `SidePanelFrontComponentPage` then derived `selectedRecordIds = [recordContext.recordId]` (length 1) or `undefined`.

The headless mount-command path via `buildHeadlessCommandContextApi` propagates the full selection correctly — only the side-panel pipeline lost it.

## Change

Carry `selectedRecordIds: string[]` end-to-end through the side-panel front-component pipeline:

- `useCommandMenuItemClick` — pass the full array (or `undefined` if empty)
- `useOpenFrontComponentInSidePanel` — accept the array on `recordContext`
- `viewableFrontComponentRecordContextComponentState` — update the atom type
- `SidePanelFrontComponentPage` — forward to `FrontComponentRenderer` (which already accepts `selectedRecordIds`)

A single-record selection is just `selectedRecordIds.length === 1`, so existing call sites keep working. `grep` confirms no remaining callers of the old `recordContext.recordId` shape.

## Test plan

- [x] Typecheck of the four touched files
- [x] Manual: define a non-headless front-component with `availabilityType: 'RECORD_SELECTION'`, tick N rows on a list view, click the pinned action — `useFrontComponentExecutionContext((c) => c.selectedRecordIds).length === N`
- [x] Single-record case still works (`.length === 1`)
- [x] Multiple invocations replace the atom cleanly


FelixMalfait:
Closing as it looks like you brought a lot of unrelated changes! Happy to review a short / well-documented PR, thanks!

PR #20622, opened 2026-05-16
fix(twenty-front): anchor body text color to theme var

### LINKED ISSUE #20607: Front components don't inherit theme colors (always black text in dark mode)
**Describe the bug**

Front components rendered via `defineFrontComponent` in a page layout tab or side panel always display black text, even when Twenty is in dark mode. The sandboxed Remote DOM environment does not inherit the host's theme colors, and there is no SDK API to detect the current theme or access the host's CSS custom properties.

**To Reproduce**

1. Create a simple front component with text:
```tsx
const MyComponent = () => <div>Hello World</div>;
export default defineFrontComponent({ ... });
```
2. Add it to a page layout tab or command menu
3. Switch Twenty to dark mode
4. The text "Hello World" remains black, invisible on the dark background

**Expected behavior**

Front components should either:
- Inherit the host's theme text color (e.g., via CSS `color: inherit` properly propagated through the sandbox boundary)
- Or have access to the current theme via a hook like `useTheme()` that returns `'light' | 'dark'` or color tokens
- Or have CSS custom properties available (e.g., `var(--text-color)`)

**Actual behavior**

All text renders in `#000` regardless of the active theme. Developers must hardcode colors, which breaks in dark mode.

**Environment (please complete the following information):**

- Twenty SDK version: 2.3.1
- Twenty server version: self-hosted (labo)
- Browser: tested in Chrome and Firefox

**Additional context**

This affects any app using `defineFrontComponent` with custom HTML/React rendering. Common workarounds (hardcoded `#333`, `#666`, `#999`) fail in dark mode. Developers cannot reliably build UI that works in both themes.

Suggested solutions:

1. **Expose a `useTheme()` hook** in `twenty-sdk/front-component` returning `{ mode: 'light' | 'dark', colors: { text: string, background: string, ... } }`
2. **Apply the host's text color** to the sandbox root element so `color: inherit` works naturally
3. **Expose CSS custom properties** from the host into the sandbox scope so developers can use `var(--text-color)`, `var(--background-primary)`, etc.

### LINKED ISSUE #20627: Bug(UI): Title of Side panel with multi record info does not follow theme colors
## Bug Description
The "Update records" title of side panel with multi records info is not following the dark theme color correctly.

Image

## Expected behavior

It should have a lighter color to distinguish itself from the background.

ehconitin:
Fixes #20607.
also fixes https://github.com/twentyhq/twenty/issues/20627

Front Components rendered via Remote DOM produced black-on-dark text in dark mode for any unstyled element. `body` already anchored `background` to a theme var; the matching `color` rule was missing, so unstyled subtrees fell through to browser default `#000`.

before - 

CleanShot 2026-05-16 at 16 31 49@2x

CleanShot 2026-05-16 at 16 32 02@2x

after - 

CleanShot 2026-05-16 at 16 30 50@2x

CleanShot 2026-05-16 at 16 31 07@2x

### DIFF
--- a/packages/twenty-front/src/index.css
+++ b/packages/twenty-front/src/index.css
@@ -4,6 +4,7 @@ body {
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
   background: var(--t-background-tertiary);
+  color: var(--t-font-color-primary);
 }
 
 html {


Weiko (approved):
LGTM

PR #20626, opened 2026-05-16
Fix(twenty-front): Slash command shows "No commands found" when no match

### LINKED ISSUE #20625: Bug(UI): BlockNote Editor slash command shows nothing when no matches are found
## Bug Description
In the BlockNote editor, when using the slash command feature, if no matching items are found, nothing is displayed, which looks weird.

https://github.com/user-attachments/assets/ba991138-bb07-4286-a76a-dacae326894f

## Expected behavior
It should show something like "Command not found".

git-init-priyanshu:
Fixes #20625

Screenshot:
Screenshot 2026-05-16 at 9 40 44 PM


etiennejouan (commented):
Hi @git-init-priyanshu, slash menu closing when no match is intended (notion also does it for example). Thanks for your contribution. Feel free to take already opened issues to be sure to work on real fix.

PR #20632, opened 2026-05-17
[Website] Partners directory

rashad:
## What this PR does

Adds the **Partners Marketplace** page to the Twenty marketing website (`/partners-marketplace`), built with Next.js App Router. The page fetches live partner data from the Twenty API and presents it in a responsive grid with an interactive filter bar.

## Partners grid

- Fetches partners from the `/s/partners` endpoint via a typed `getPartners()` server-side fetcher
- Responsive 1 → 2 → 3 column grid (mobile / tablet / desktop)
- Each card shows name, region eyebrow, intro text, chip rows (Regions / Languages / Deploys), and a Calendly CTA
- Stagger entrance animation (700ms cascade, respects `prefers-reduced-motion`)

## Filter bar

- Three facets: **Region**, **Language**, **Deployment** — multi-select chips
- **Selection model:** OR within a facet, AND across facets (e.g. `Europe OR US` AND `French`)
- Filter state lives in URL search params (`?regions=EUROPE,US&languages=FRENCH`) — filtered views are shareable and browser-back works correctly
- Client-side filtering — no server round-trip per interaction
- Result count ("Showing 3 of 8 partners") updates live with `aria-live="polite"`
- "Clear filters" button resets all facets in one URL update, only shown when filters are active
- Empty state ("No partners match your filters") replaces the grid when nothing matches
- 200ms opacity fade-out on card removal; initial stagger animation preserved on first load
- `prefers-reduced-motion: reduce` disables all transitions

## Architecture

- `page.tsx` stays a **Server Component** — fetches partners server-side, all partner HTML is in the initial response for SEO
- `<MarketplaceClient>` is the client boundary — owns filter state via `useFilterState()` (backed by `useSearchParams`)
- Canonical URL set in page metadata so `?regions=...` deep-links don't get indexed as duplicates
- `<Suspense>` wrapper around `MarketplaceClient` for Next.js 15 `useSearchParams` compliance
- No new npm dependencies

## Test coverage

31 tests across three suites:
- `filter-partners.test.ts` — pure filter logic (OR / AND semantics, empty results)
- `filter-url-helpers.test.ts` — URL param encode / decode / toggle / round-trip
- `use-filter-state.test.tsx` — hook behaviour with mocked `next/navigation`

## Screenshot

Screenshot 2026-05-17 at 15 01 54

## Test plan

- [ ] Navigate to `/partners-marketplace` — all live partners render
- [ ] Click a Region chip — URL updates with `?regions=...`, cards filter, count updates
- [ ] Click the same chip again — selection removed, all cards return
- [ ] Select chips from two different facets — AND behaviour narrows results correctly
- [ ] Trigger empty state (e.g. filter to a region with no partners) — empty state shown with "Clear filters" button
- [ ] Click "Clear filters" — all cards return, URL cleared
- [ ] Deep-link to `?regions=EUROPE&languages=FRENCH` — page loads with filters applied
- [ ] Browser back button restores previous filter state


charlesBochet:
@rashad Let's try to merge this one (we aim to merge PR as fast as possible, which is not always the case but this is the goal). PR should not stay as draft for more than 24hours. 

@rashad remove claude for your commits

@Bonapara could you take a look at the design
@mabdullahabaid could you review the code

Bonapara:
Made some comments on the associated discord channel

packages/twenty-website-new/.env.example
```diff
@@ -35,3 +35,7 @@ ENTERPRISE_JWT_PUBLIC_KEY=
 
 # Optional: short-lived validity token length in days (default 30)
 # ENTERPRISE_VALIDITY_TOKEN_DURATION_DAYS=
+
+# Twenty workspace the marketing site reads data from (server-side only)
+TWENTY_API_URL=
```
FelixMalfait: Twenty API as a variable and folder name feels a bit generic ; there could be other use-cases for other Twenty-related apis - maybe TWENTY_PARTNERS_API (same for folder etc)?

packages/twenty-website-new/src/lib/website-routing/types.ts
```diff
@@ -8,6 +8,7 @@ export type WebsiteRouteId =
   | 'product'
   | 'pricing'
   | 'partners'
+  | 'partnersMarketplace'
```
FelixMalfait: nit: marketplace feels like an internal name that shouldn't be expose, we can do something more userfriendly like /partners/list etc (maybe there are better options!)

packages/twenty-website-new/src/lib/twenty-api/partner-types.ts
```diff
@@ -0,0 +1,20 @@
+export type DeploymentExpertise = 'CLOUD' | 'SELF_HOST';
+
+export type ServedGeo = 'EUROPE' | 'US' | 'LATAM' | 'MENA' | 'APAC' | 'AFRICA';
+
+export type SpokenLanguage =
```
FelixMalfait: nit: in server/front we try to have 1 export = 1 file (accordingly named). Code quality is lower and conventions are more relaxed in website but we can probably do better still

(same for barrel files, extract styles files etc. if we were in twenty-front I would have strict comments on that but on website I'm not sure if we do this well so let's just check and do something acceptable even if not perfect)

packages/twenty-website-new/src/lib/twenty-api/partner-types.ts
```diff
@@ -0,0 +1,20 @@
+export type DeploymentExpertise = 'CLOUD' | 'SELF_HOST';
+
+export type ServedGeo = 'EUROPE' | 'US' | 'LATAM' | 'MENA' | 'APAC' | 'AFRICA';
```
FelixMalfait: nit: in twenty-front we derives types from the graphql api to keep only a single source of truth. Maybe soemthing we should explore here too (could also be a bad idea)

packages/twenty-website-new/src/lib/twenty-api/client.ts
```diff
@@ -0,0 +1,39 @@
+const readRequiredEnv = (name: 'TWENTY_API_URL' | 'TWENTY_API_KEY'): string => {
```
FelixMalfait: Feels like a global util / something that should be shared across all envs  being accessed and not done locally?

packages/twenty-website-new/src/lib/twenty-api/partner-types.ts
FelixMalfait: I'm not sure how it works already in the twenty-sdk when developping frontend for apps but I think in a way that's what we do already

packages/twenty-website-new/src/app/[locale]/partners-marketplace/filter-url-helpers.ts
```diff
@@ -0,0 +1,74 @@
...
+export const ALL_REGIONS: readonly ServedGeo[] = [
+  'EUROPE',
+  'US',
+  'LATAM',
+  'MENA',
+  'APAC',
+  'AFRICA',
+];
+
+export const ALL_LANGUAGES: readonly SpokenLanguage[] = [
```
Comment (FelixMalfait) [inline packages/twenty-website-new/src/app/[locale]/partners-marketplace/filter-url-helpers.ts]: saw this somewhere else? Should be a single source of truth (ideally derived from graphql api as mentioned, but ok if not possible)

packages/twenty-website-new/src/app/[locale]/partners-marketplace/components/MarketplaceGrid.tsx
```diff
@@ -0,0 +1,111 @@
...
+  partners: readonly MarketplacePartner[];
+};
+
+export function MarketplaceGrid({ partners }: MarketplaceGridProps) {
+  const [rendered, setRendered] = useState(partners);
+  const [exitingSlugs, setExitingSlugs] = useState<ReadonlySet<string>>(
+    new Set(),
+  );
+
+  useEffect(() => {
```
Comment (FelixMalfait) [inline packages/twenty-website-new/src/app/[locale]/partners-marketplace/components/MarketplaceGrid.tsx]: Let's avoid client-side rendered and useEffect as much as possible. Sometimes it's not possible but double check.  Here it's not obvious to me that we can't do without useEffect but maybe I'm missing something

FelixMalfait (commented):
Great work! Didn't check how it looked, just read the code quickly

FelixMalfait:
Business POV: focus on project budgets (1k? 10k? etc) ; and also introducing a sense of whether a partner has a low vs high hourly rate (positioning) without necessarily asking them the explicit rate. Also Region (APAC) not as relevant as country imo (China != Japan, etc.), timezone could maybe be a factor... To build trust having profile pictures would be great (we can eventually apply a common style if it looks odd), logo eventually (not as good to build trust/help selection)

mabdullahabaid:
You'd have to move these changes from `twenty-website-new` package `twenty-website` instead since we renamed the package last night - I can do that for you if you're not currently working on the branch, let me know if you need any help resolving conflicts and moving these changes (or anything else).

packages/twenty-website-new/.env.example
rashad: Scoped it to partners: `lib/twenty-api` → `lib/partners-api`, `TWENTY_API_URL`/`TWENTY_API_KEY` → `TWENTY_PARTNERS_API_URL`/`TWENTY_PARTNERS_API_KEY`, and `twentyFetch` → `partnersApiFetch`.

packages/twenty-website-new/src/lib/website-routing/types.ts
rashad: Renamed the route to `/partners/list`, nested under the existing `/partners` page. The Next.js folder is now `app/[locale]/partners/list` and the route id is `partnersList`. Left the internal `Marketplace*` component names since they're not user-facing.

packages/twenty-website-new/src/lib/twenty-api/partner-types.ts
rashad: Pulled the facet vocabulary into its own `partner-facets.ts` and kept each file single-concern. Left the lib `index.ts` barrel (the package's public surface) and the small `chip-styles.ts`/`chip-labels.ts` files as-is, which felt acceptable for the website — can go further if you'd prefer.

packages/twenty-website-new/src/lib/twenty-api/partner-types.ts
rashad: Looked into this. The Partner object and its select-field options (regions/languages/deployment) live in the workspace, not the repo, so there's no static type to import. The proper single-source route is `twenty-client-sdk`'s `./generate` (GenQL), which codegens a typed core client from the workspace's GraphQL schema — but that means moving off the REST `/s/partners` call onto the GraphQL core API plus a build-time codegen step against the workspace schema, which is bigger than this PR. For now I removed the in-repo duplication (the facet arrays now derive the union types — see `partner-facets.ts`) and will wire the SDK codegen in a dedicated follow-up.

packages/twenty-website-new/src/lib/twenty-api/client.ts
rashad: Made it explicitly local to partners: scoped to `lib/partners-api` and stopped exporting the fetch helper from the barrel, so the lib's public surface is just `getPartners` + types instead of a generic client. Also inlined the env reads and dropped the `readRequiredEnv` indirection.

packages/twenty-website-new/src/app/[locale]/partners-marketplace/filter-url-helpers.ts
Comment (rashad) [reply packages/twenty-website-new/src/app/[locale]/partners-marketplace/filter-url-helpers.ts]: Fixed — regions/languages/deployment were declared twice (union types + runtime arrays). There's now one `as const` array per facet in `partner-facets.ts`, the union types are derived from those arrays, and the filter helpers + FilterBar consume the same arrays.

packages/twenty-website-new/src/app/[locale]/partners-marketplace/components/MarketplaceGrid.tsx
Comment (rashad) [reply packages/twenty-website-new/src/app/[locale]/partners-marketplace/components/MarketplaceGrid.tsx]: Removed it. The `useEffect` only drove a 200ms exit fade, and it read the state it set back in its own dependency array. Filtering already produces the correct list upstream, so the grid now renders partners directly and no longer needs to be a client component. The staggered entrance animation is pure CSS (`animation-delay`) and is unaffected. The only thing dropped is the exit fade — filtered-out cards disappear immediately now; easy to bring back without the effect if it's worth it.

rashad:
Rebased onto the latest `main` and moved everything from `twenty-website-new` → `twenty-website` (@mabdullahabaid — done on this branch, nothing to migrate on your side). Rewrote the branch into clean, Claude-free commits. Addressed @FelixMalfait's review — specifics in each thread.

rashad:
@FelixMalfait on the business-POV inputs (project budgets, hourly-rate positioning, country vs region, timezone, profile pictures/logos): those will be addressed in the partners app PR rather than in this one.

charlesBochet:
Did a quick security pass with my favorite LLM:

`partner.calendarLink` is passed straight into `<a href>` without checking the URL scheme (`PartnerCard.tsx`). If anyone with edit access in the partners workspace ever puts a `javascript:` or `data:text/html,...` URL in there it becomes stored XSS on twenty.com.

Confirm issue and fix like:
```
const isSafeHttpUrl = (raw: string) => {
  try { return ['https:', 'http:'].includes(new URL(raw).protocol); }
  catch { return false; }
};
```

charlesBochet:
I'm going to merge it, let's iterate on it @Bonapara @rashad

charlesBochet (approved):

PR #20637, opened 2026-05-17
fix(kanban): preserve scroll on board re-init + propagate same-column reorders via SSE

ehconitin:
closes https://discord.com/channels/1130383047699738754/1504130730840821860

https://github.com/user-attachments/assets/d5833031-01c6-4e46-b699-c29c42435a53

## Summary

Fixes two related issues with the kanban (board view) collaboration experience:

1. **Scroll-to-top on every data change** — `triggerRecordBoardInitialQuery` always scrolled the board to the top, even when re-initializing for a single-record data change (SSE echo of your own mutation, a collaborator's update). Scroll reset only makes sense when the dataset itself changes (filter / sort / group).
2. **Same-column reorders by other users did not propagate** — the server's diff function stripped `FieldMetadataType.POSITION`, so position-only updates produced empty `updatedFields` and short-circuited event emission entirely. SSE clients never received them.

## What's in here

- **Frontend** — `useTriggerRecordBoardInitialQuery` now exposes a `triggerRecordBoardInitialQueryWithoutScrollReset` variant; data-driven re-inits in `RecordBoardDataChangedEffect` use it, while genuine filter / sort / group changes keep the scroll-resetting `triggerRecordBoardInitialQuery`. `getRecordBoardEffectsForUpdateInputs` classifies each update as `trigger-initial-query` / `reposition-records` / `none`. For position- or group-only changes we skip the re-query and reposition records in place in the store (`useRepositionRecordsOnBoard`), which avoids the flicker and preserves scroll.
- **Server** — removes `POSITION` from `objectRecordChangedValues`' strip list, so position-only updates emit a non-empty diff and flow through SSE. Position is now treated as a field like any other across all event consumers (SSE, webhooks, workflows, logic functions); a trigger with an explicit field filter still excludes it.


Weiko (commented):
Leaving this one to our SSE expert 😄 @thomtrp

packages/twenty-front/src/modules/object-record/record-board/components/RecordBoardDataChangedEffect.tsx
```diff
@@ -38,6 +40,12 @@ export const RecordBoardDataChangedEffect = () => {
 
   const { removeRecordsFromBoard } = useRemoveRecordsFromBoard();
 
+  const debouncedTriggerRecordBoardInitialQuery = useDebouncedCallback(
+    () => triggerRecordBoardInitialQuery({ preserveScroll: true }),
+    SSE_BOARD_DEBOUNCE_TIME_IN_MS_TO_AVOID_SSE_OWN_EVENTS_RACE_CONDITION,
+    { leading: false },
+  );
```
thomtrp: curious why do we need that debounce?

packages/twenty-server/src/engine/core-modules/event-emitter/utils/object-record-changed-values.ts
```diff
@@ -34,8 +34,7 @@ export const objectRecordChangedValues = (
       if (
         key === 'updatedAt' ||
         key === 'searchVector' ||
-        field?.type === FieldMetadataType.RELATION ||
-        field?.type === FieldMetadataType.POSITION
+        field?.type === FieldMetadataType.RELATION
```
thomtrp: why do we need SSE events for position ?

packages/twenty-server/src/engine/metadata-modules/webhook/utils/transform-event-batch-to-webhook-events.ts
```diff
@@ -27,6 +27,18 @@ export const transformEventBatchToWebhookEvents = ({
...
+        'updatedFields' in eventData.properties
+          ? eventData.properties.updatedFields
+          : undefined;
+
+      if (
+        eventUpdatedFields?.length === 1 &&
+        eventUpdatedFields[0] === 'position'
+      ) {
+        continue;
+      }
```
thomtrp: does not looks super clean TBH

packages/twenty-server/src/modules/workflow/workflow-trigger/automated-trigger/listeners/workflow-database-event-trigger.listener.ts
```diff
@@ -394,26 +394,32 @@ export class WorkflowDatabaseEventTriggerListener {
...
     }
 
     if (action === DatabaseEventAction.UPSERTED) {
       const settings = eventListener.settings as UpsertEventTriggerSettings;
       const upsertEventPayload = eventPayload as ObjectRecordUpsertEvent;
+      const updatedFields = upsertEventPayload?.properties?.updatedFields ?? [];
+
+      if (updatedFields.length === 1 && updatedFields[0] === 'position') {
+        return false;
+      }
```
thomtrp: same

thomtrp (commented):
I think I miss a bit of understanding there

packages/twenty-front/src/modules/object-record/record-board/components/RecordBoardDataChangedEffect.tsx
ehconitin: Without it, the board re-inits twice on every drag -- once from the optimistic update and once when the SSE echo of your own move arrives. 50ms dedupes them. Same as the table!

packages/twenty-server/src/engine/core-modules/event-emitter/utils/object-record-changed-values.ts
ehconitin: The board, like the table, needs to re-query when a record's position changes -- there is no separate sync channel today -- and adding one would dwarf the bug. 

[twenty#17489](https://github.com/twentyhq/twenty/pull/17489) established this pattern for the table -- this PR extends it to the board.

packages/twenty-server/src/engine/metadata-modules/webhook/utils/transform-event-batch-to-webhook-events.ts
ehconitin: Can DRY the common check into a util -- but the approach (filter at each consumer) stays the same no? Or are you pushing for it to be filtered higher up?

thomtrp:
I have the feeling we are solving the issue by some patches but I am not fully convinced:
- I see why you added a debounce + that preventScroll but I feel the actual issue is that a position change should not run triggerRecordBoardInitialQuery. We may need a light version that would only update record position
- I am not convinced by the sse fixes to start listening at position. I wonder if we should either accept that workflows and webhook listen at position changes, either SSE should stop using objectRecordChangedValues and use its own, since this is not the same changes that are looked

packages/twenty-front/src/modules/object-record/record-board/components/RecordBoardDataChangedEffect.tsx
bosiraphael: Debounce seems hacky here

packages/twenty-front/src/modules/object-record/record-board/components/RecordBoardDataChangedEffect.tsx
```diff
@@ -14,15 +15,16 @@ import { useAtomComponentFamilySelectorCallbackState } from '@/ui/utilities/stat
...
 import { useAtomComponentStateCallbackState } from '@/ui/utilities/state/jotai/hooks/useAtomComponentStateCallbackState';
 import { useCallback } from 'react';
-import { isDefined } from 'twenty-shared/utils';
+import { assertUnreachable, isDefined } from 'twenty-shared/utils';
 
 export const RecordBoardDataChangedEffect = () => {
   const store = useStore();
   const { objectMetadataItem } = useRecordIndexContextOrThrow();
-  const { triggerRecordBoardInitialQuery } =
+  const { triggerRecordBoardInitialQueryWithoutScrollReset } =
```
thomtrp: not a big fan of having two functions. If we have to keep shouldResetScroll as a param, let's do it

thomtrp (approved):
Much better, please keep triggerRecordBoardInitialQuery as an unique function

PR #20638, opened 2026-05-17
chore(server): remove unused `CommandLogger` from command module

Rpaudel379:
## Summary

This PR removes the unused `CommandLogger` implementation located at:

```
/commands/command-logger.ts
```

The Command application context is bootstrapped using `LoggerService` from:

```ts
import { LoggerService } from 'src/engine/core-modules/logger/logger.service';
...
const loggerService = app.get(LoggerService);
...
// Inject our logger
app.useLogger(loggerService);
...

```

So `CommandLogger` is not imported, injected, or referenced anywhere in the Command execution flow and is safe to remove.

## Note
There is another `CommandLogger` class at:
```
/database/commands/logger.ts
```

This one is only used within `database-command` module and is unrelated to the Command module logger being removed in this PR.


ijreilly (approved):
thank you for contributing!

PR #20643, opened 2026-05-17
fix(server): add missing return in catch blocks of metadata resolvers

### LINKED ISSUE #20205: missing return in catch blocks of updateOneField / deleteOneField causes silent null response on certain errors
Bug Description
In field-metadata.resolver.ts, the catch blocks for updateOneField (line 78) and deleteOneField (line 104) are missing a return before the exception handler call. If the handler doesn't throw (which can happen for WorkspaceMigrationBuilderException — it calls workspaceMigrationBuilderExceptionFormatter and returns void), the function falls through and returns undefined. GraphQL then serializes this as null with no error surfaced to the client.

The same issue exists in object-metadata.resolver.ts (lines 151, 174, 197) and index-metadata.resolver.ts (line 44).

Interestingly createOneField in the same file (line 56) correctly uses return fieldMetadataGraphqlApiExceptionHandler(...) — so the pattern is inconsistent within the same file.

Expected behavior
When an error is caught and the handler doesn't throw, the mutation should still terminate cleanly rather than silently returning null to the client. The catch blocks should return the handler call.

Technical inputs
Add return before fieldMetadataGraphqlApiExceptionHandler(...) at lines 78 and 104 in packages/twenty-server/src/engine/metadata-modules/field-metadata/field-metadata.resolver.ts
Add return before objectMetadataGraphqlApiExceptionHandler(...) at lines 151, 174, 197 in packages/twenty-server/src/engine/metadata-modules/object-metadata/object-metadata.resolver.ts
Add return at line 44 in packages/twenty-server/src/engine/metadata-modules/index-metadata/index-metadata.resolver.ts
The fix is a one-word change per line — just adding return before each handler call

shubham-0707:
## Summary
- Added missing `return` before exception handler calls in `updateOneField` and `deleteOneField` (`field-metadata.resolver.ts`)
- Added missing `return` before exception handler calls in `createOneObject`, `deleteOneObject`, and `updateOneObject` (`object-metadata.resolver.ts`)
- Without these, when the handler doesn't throw, the functions silently return `undefined` which GraphQL serializes as `null` with no error surfaced to the client

## Note
- `index-metadata.resolver.ts` (mentioned in the issue) already has `return [];` after the handler call, so no change was needed there
- The pattern is now consistent with `createOneField` which already had `return` correctly

Fixes #20205

### DIFF
--- a/packages/twenty-server/src/engine/metadata-modules/field-metadata/field-metadata.resolver.ts
+++ b/packages/twenty-server/src/engine/metadata-modules/field-metadata/field-metadata.resolver.ts
@@ -122,7 +122,7 @@ export class FieldMetadataResolver {
 
       return fromFlatFieldMetadataToFieldMetadataDto(flatFieldMetadata);
     } catch (error) {
-      fieldMetadataGraphqlApiExceptionHandler(error);
+      return fieldMetadataGraphqlApiExceptionHandler(error);
     }
   }
 
@@ -144,7 +144,7 @@ export class FieldMetadataResolver {
 
       return fromFlatFieldMetadataToFieldMetadataDto(flatFieldMetadata);
     } catch (error) {
-      fieldMetadataGraphqlApiExceptionHandler(error);
+      return fieldMetadataGraphqlApiExceptionHandler(error);
     }
   }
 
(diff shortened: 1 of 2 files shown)


prastoin (commented):
Hello there,
These are not oversight
The exception handler always throw
So here you're basically returning never

PR #20644, opened 2026-05-17
fix(server): handle network errors in RestApiService catch block

### LINKED ISSUE #20136: TypeError in RestApiService swallows internal network errors, resulting in silent 500 responses
## Bug Description

When the Twenty server is configured with a SERVER_URL that is inaccessible from inside the container (e.g., a Tailscale IP or an external Load Balancer IP without NAT loopback), requests to REST endpoints like /rest/metadata/objects fail with a 500 Internal Server Error.

However, the HTTP response returns an empty body (Content-Length: 0) and the application logs absolutely nothing. Debugging the server reveals that the underlying network error (e.g., ECONNREFUSED when the app tries to reach its own GraphQL endpoint) is being masked by a TypeError, which is then caught and silenced by NestJS's UnhandledExceptionFilter.

## Expected behavior

The internal HTTP client should safely handle network-level errors where an HTTP response object doesn't exist. The original network error (like ECONNREFUSED) should be logged to the server console to allow for proper infrastructure debugging, and a proper error response should be returned to the client instead of an empty body.

## Technical inputs

- The issue is in packages/twenty-server/src/engine/api/rest/rest-api.service.ts inside the catch block of RestApiService.call. (https://github.com/twentyhq/twenty/blob/51384bc0858c895bf14bc7a1838e8fcb87644269/packages/twenty-server/src/engine/api/rest/rest-api.service.ts#L44)

The code catches err from httpClient.post() and blindly throws new RestApiException(err.response.data.errors);

For network-level failures (timeout, connection refused), err.response is undefined.
This triggers a TypeError: Cannot read properties of undefined (reading 'data').

This TypeError crashes the execution, which the UnhandledExceptionFilter then catches, wiping the trace and returning the empty 500 status.

Easily fixable by checking if err.response?.data?.errors exists before parsing it, and falling back to throwing/logging the raw err.message if it is a pure network failure.

shubham-0707:
## Summary
- Added safe null check for `err.response?.data?.errors` in `RestApiService.call()` catch block
- When the internal HTTP client fails with a network-level error (ECONNREFUSED, timeout), `err.response` is `undefined` — accessing `.data.errors` on it throws a `TypeError` which gets silently swallowed, returning an empty 500
- Now falls back to throwing the raw error message for network failures instead of crashing

## Changes
- `packages/twenty-server/src/engine/api/rest/rest-api.service.ts`

Fixes #20136

### DIFF
--- a/packages/twenty-server/src/engine/api/rest/rest-api.service.ts
+++ b/packages/twenty-server/src/engine/api/rest/rest-api.service.ts
@@ -1,4 +1,4 @@
-import { Injectable, InternalServerErrorException } from '@nestjs/common';
+import { Injectable } from '@nestjs/common';
 
 import { type AxiosResponse } from 'axios';
 
@@ -46,9 +46,7 @@ export class RestApiService {
         throw new RestApiException(err.response.data.errors);
       }
 
-      throw new InternalServerErrorException(
-        err.message ?? 'Internal server error',
-      );
+      throw err;
     }
 
     if (isDefined(response.data.errors) && response.data.errors.length > 0) {
--- a/packages/twenty-server/src/engine/api/rest/rest-api.service.ts
+++ b/packages/twenty-server/src/engine/api/rest/rest-api.service.ts
@@ -41,7 +41,13 @@ export class RestApiService {
         },
       });
     } catch (err) {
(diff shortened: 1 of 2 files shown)


ijreilly (approved):
thank you for contributing !

PR #20641, opened 2026-05-17
Fix: currency icon and amount text misalignment

git-init-priyanshu:
Fixes #20640 

Screenshots:

image

image

PR #20646, opened 2026-05-17
fix(front): align currency icon vertically with amount text

### LINKED ISSUE #20640: Bug(UI): Currency icon not aligned properly
## Bug Description

The currency symbol and its text are not aligned.
In setttings -> data-model -> Object -> New field -> Currency:

Image

And in Table:
Image

In other places too which use `CurrencyDisplay.tsx` component.

## Expected behavior
The currency symbol and the amount text should be vertically aligned.

shubham-0707:
## Summary
- Replaced inline `<span>` wrapping the currency icon with a Linaria styled component using `display: flex` and `align-items: center`
- The icon was misaligned with the amount text in table views and settings because the inline span didn't vertically center the SVG icon

## Changes
- `packages/twenty-front/src/modules/ui/field/display/components/CurrencyDisplay.tsx`

Fixes #20640

### DIFF
--- a/packages/twenty-front/src/modules/ui/field/display/components/CurrencyDisplay.tsx
+++ b/packages/twenty-front/src/modules/ui/field/display/components/CurrencyDisplay.tsx
@@ -1,6 +1,9 @@
 import { useContext, useId, useState } from 'react';
 import { createPortal } from 'react-dom';
+import { styled } from '@linaria/react';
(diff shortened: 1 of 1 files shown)


git-init-priyanshu:
#20641

etiennejouan (approved):
Hi @shubham-0707, LGTM. thank you for your contribution.

PR #20670, opened 2026-05-18
fix(filters): make filter dispatcher own relation-target resolution

FelixMalfait:
## Summary

Two relation-traversal bugs surfaced post-merge of #20533, both rooted in the same architectural smell: the GraphQL filter dispatcher took a flat `fields: FieldShared[]` array and silently dropped any filter whose `relationTargetFieldMetadataId` wasn't in that array. Callers had to remember to pre-augment the list with relation targets — and 16+ call sites did not all know this.

This PR fixes both bugs and removes the smell.

### Bug 1 — Save as new view loses the relation target

`useCreateViewFromCurrentView` built the create-filter input without `relationTargetFieldMetadataId`. The saved view's filter persisted without the traversal — on reload the chip showed "Company contains 'air'" instead of "Company → Name contains 'air'". Discarded at save time, not at read time.

Fix: include `relationTargetFieldMetadataId` in the create input. (Commit 1.)

### Bug 2 — Workflow Search Records drops one-hop traversals

`FindRecordsWorkflowAction` built its fields list from `flatObjectMetadata.fieldIds` only (source object's fields). The shared dispatcher then couldn't resolve the relation target field on the related object and silently dropped the filter — a configured "People where Company → Name Contains 'Airbnb'" came through as `{ and: [] }`.

This was the same shape as bugs already fixed in 5 other call sites (chart filters, view filters, record table, etc.). The pattern was: caller forgets to augment fields → dispatcher silently drops the filter.

Fix (commit 2): change the dispatcher to take a `findFieldMetadataItemById: (id) => FieldShared | undefined` resolver callback. Both source-field and relation-target-field lookups go through the same resolver, so callers no longer need to know about the augmentation requirement. Frontend callers pass a workspace-wide resolver built from `flattenedFieldMetadataItemsSelector`; server callers wrap `findFlatEntityByIdInFlatEntityMaps` on `flatFieldMetadataMaps`. In both cases relation-target lookups just work, because the resolver can see fields on related objects.

## Why this matters

Before: "if you call the dispatcher, pre-augment your fields list with relation targets, or filters get silently dropped." An invariant only enforceable by code review, broken often enough to ship two user-visible bugs in one week.

After: the dispatcher resolves field ids itself. There's no list to forget to augment. The failure mode (filter silently dropped) becomes structurally impossible at the dispatcher boundary.

Net diff: 240 insertions, 319 deletions. Removed `augmentFieldsWithRelationTargets` (frontend) and the workflow whack-a-mole code (server).

## Test plan
- [ ] Save view: create an advanced filter using a one-hop relation traversal, click "Save as new view", reload, confirm the chip still reads "Source → Target operator value"
- [ ] Workflow: configure a Search Records action with a relation-traversal filter, run the workflow, confirm the filter is actually applied
- [ ] Dashboard chart: configure a chart with a relation-traversal filter, confirm the chart data respects it
- [ ] Record table, group-by, calendar, total count, footer aggregates: all continue to work with both plain and relation-traversal filters


packages/twenty-front/src/modules/command-menu-item/engine-command/utils/buildHeadlessCommandContextApi.ts
```diff
@@ -99,7 +100,7 @@ export const buildHeadlessCommandContextApi = ({
         contextStoreFilters: filters,
         contextStoreFilterGroups: filterGroups,
         objectMetadataItem,
-        flattenedFieldMetadataItems,
+        findFieldMetadataItemById: (id) => fieldMetadataItemByIdMap.get(id),
```
charlesBochet, line 103: non blocking: I'm not a fan of passing a callback here, why not passing the map direclty? I'm not sure it harms here (maybe as we not using useCallback)

packages/twenty-server/src/engine/metadata-modules/view/services/view-query-params.service.ts
```diff
@@ -107,39 +107,12 @@ export class ViewQueryParamsService {
...
-            color: 'color' in opt ? opt.color : undefined,
-            position: opt.position,
-          })),
-        };
-      })
-      .filter(isDefined);
-
     const filter = computeRecordGqlOperationFilter({
-      fields,
+      findFieldMetadataItemById: (id) =>
```
charlesBochet, line 111: non-blocking: not a big fan it be too, new pattern we don't have either. Is there a problem passing the all fields?

charlesBochet (approved):
I don't like the pattern and I would pass the map of object or of field directly (for JS is just pointers), I'm not sure to get why we got this complex by passing a callback

But approving if you want to merge it this way, this is not enough to block it

FelixMalfait:
Attempting to refactor in a stacked PR, we'll see if it works...

PR #20679, opened 2026-05-18
use declared outputSchema for logic-function steps

abdulrahmancodes:
When a logic function declares `workflowActionTriggerSettings.outputSchema`, use it as the step's initial output schema so downstream steps can pick variables without first running the Test tab. A successful test run still overrides the schema with the inferred shape, preserving "test wins" behavior. Falls back to the existing "Generate Function Output" LINK placeholder when no schema is declared (custom code steps, older functions).

https://github.com/user-attachments/assets/af9c45ed-d623-4234-be9f-46812fd06e2e


charlesBochet:
@thomtrp

packages/twenty-shared/src/logic-function/input-schema-to-output-schema.ts
```diff
@@ -0,0 +1,77 @@
...
+  return (LEAF_TYPES as string[]).includes(type);
+};
+
+const convertProperty = (
+  key: string,
+  property: InputSchemaProperty,
+): Leaf | Node => {
+  const label = property.label ?? key;
+
+  if (property.type === 'object') {
```
thomtrp, line 31: prefer isObject guard

thomtrp (approved):

PR #20681, opened 2026-05-18
fix(ai-chat) - upload files

### LINKED ISSUE #20437: Uploaded file to AI chat is not sent to AI
**Scenario:**
1. Log in
2. Open AI chat
3. Start new chat with AI and upload file
4. Send the message with file

**Actual:** File disappears from message and it's not sent to AI

**Expected:** File is sent to AI in message

https://github.com/twentyhq/private-issues/issues/474

etiennejouan:
closes https://github.com/twentyhq/twenty/issues/20437

bonus : persist file filename for UI display


packages/twenty-server/src/engine/metadata-modules/ai/ai-chat/services/agent-chat-streaming.service.ts
```diff
@@ -32,6 +32,11 @@ import { type StreamAgentChatJobData } from 'src/engine/metadata-modules/ai/ai-c
 import { AgentChatEventPublisherService } from 'src/engine/metadata-modules/ai/ai-chat/services/agent-chat-event-publisher.service';
 import { AgentChatService } from 'src/engine/metadata-modules/ai/ai-chat/services/agent-chat.service';
 
+export type FileAttachment = {
```
FelixMalfait: nit: one export per file

FelixMalfait (approved):

YurkaTim:
Hi!
If an unsupported file sent to a chat, error message appears and then it's impossible to continue the chat because only error message is reproduced.
"Failed to get response
'file part media type text/markdown' functionality not supported."
Actually it's also ridiculous that simple txt, md are unsupported but it's probably not related to the issue.

PR #20691, opened 2026-05-18
fix(front): display all fitting relation items in record table cells

### LINKED ISSUE #12039: On record table, relation fields, we only display 3 items even when the cell is larger
## Bug Description

Displayed cell by default
Image

On hover
Image

## Expected behavior

By default we should see as many items as the cell can display.

shubham-0707:
## Summary
- Always use `ExpandableList` for relation from-many field display, even when the cell is not focused
- Previously, unfocused cells rendered chips in a plain container with `overflow: hidden`, which only showed ~3 items regardless of available space
- Now the dynamic overflow detection in `ExpandableList` ensures as many chips as fit are visible, with a `+N` count shown on hover

## Test plan
- [ ] Open a record table with a relation field that has more than 3 linked records
- [ ] Verify that the cell displays as many chips as fit in its width
- [ ] Verify that hovering shows the `+N` chip count for hidden items
- [ ] Resize the column wider and verify more items become visible

Fixes #12039


charlesBochet:
Closing, see comment on https://github.com/twentyhq/twenty/pull/20696

PR #20703, opened 2026-05-19
Stop bundling twenty-ui react cjs runtime code

martmull:
For this front component using Avatar from `twenty-sdk/ui`

```typescript
import { defineFrontComponent } from 'twenty-sdk/define';
import { Avatar } from 'twenty-sdk/ui';

// React component - implement your UI here
const Component = () => {
  return (
    <div style={{ padding: '20px', fontFamily: 'sans-serif' }}>
      <Avatar avatarUrl={null} placeholder={'test'} />
      <h1>My new component!</h1>
      <p>This is your front component: fc-test</p>
    </div>
  );
};

export default defineFrontComponent({ ... });
```

## Before
image

## After

image

### DIFF
--- a/packages/twenty-ui/vite.config.ts
+++ b/packages/twenty-ui/vite.config.ts
@@ -125,9 +125,13 @@ export default defineConfig(({ command }) => {
         name: 'twenty-ui',
       },
       rollupOptions: {
-        external: Object.keys(packageJson.dependencies || {}).filter(
-          (dep) => !BUNDLED_DEPS.includes(dep),
-        ),
+        external: (id: string) => {
+          const deps = Object.keys(packageJson.dependencies || {}).filter(
+            (dep) => !BUNDLED_DEPS.includes(dep),
+          );
+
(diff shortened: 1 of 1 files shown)


packages/twenty-ui/vite.config.ts
```diff
@@ -125,9 +125,13 @@ export default defineConfig(({ command }) => {
         name: 'twenty-ui',
       },
       rollupOptions: {
-        external: Object.keys(packageJson.dependencies || {}).filter(
-          (dep) => !BUNDLED_DEPS.includes(dep),
-        ),
+        external: (id: string) => {
+          const deps = Object.keys(packageJson.dependencies || {}).filter(
```
Weiko: AFAIK, when you pass a function to external, Rollup calls that function once for every import specifier it encounters while walking the module graph. So here you are computing external deps N times. Could we move this at the top of defineConfig? @martmull

Weiko (approved):
LGTM

PR #20704, opened 2026-05-19
fix(ai) - add ai model preferences fallback

etiennejouan:
**Problem** 
AI_MODEL_PREFERENCES, JSON env var is not supported + IS_CONFIG_VARIABLES_IN_DB_ENABLED=false in twenty cloud server
-> No option to set AI_MODEL_PREFERENCES

**Solution** 
AI_MODEL_PREFERENCES supports three override sources beyond the hardcoded code defaults, in priority order:

- DB (IS_CONFIG_VARIABLES_IN_DB_ENABLED=true), the only writable source; admin-panel mutations persist here
- ENV not usable in Twenty Cloud, which does not handle JSON-format env vars
- **Introduced in this PR** --> File (AI_MODEL_PREFERENCES_STORAGE_PATH), a read-only startup fallback, the only viable override in Cloud/self-managed deployments where DB config is disabled and JSON env vars are unsupported.

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/twenty-config/config-variables.ts
+++ b/packages/twenty-server/src/engine/core-modules/twenty-config/config-variables.ts
@@ -44,7 +44,7 @@ import {
 } from 'src/engine/core-modules/twenty-config/twenty-config.exception';
 import { type AiModelPreferences } from 'src/engine/metadata-modules/ai/ai-models/types/ai-model-preferences.type';
 import { type AiProvidersConfig } from 'src/engine/metadata-modules/ai/ai-models/types/ai-providers-config.type';
-import { loadDefaultModelPreferences } from 'src/engine/metadata-modules/ai/ai-models/utils/load-default-model-preferences.util';
+import { DEFAULT_MODEL_PREFERENCES } from 'src/engine/metadata-modules/ai/ai-models/utils/load-default-model-preferences.util';
 
 export class ConfigVariables {
   @ConfigVariablesMetadata({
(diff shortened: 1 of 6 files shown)


packages/twenty-server/src/engine/metadata-modules/ai/ai-models/types/ai-model-preferences.schema.ts
```diff
@@ -0,0 +1,8 @@
+import { z } from 'zod';
+
+export const aiModelPreferencesSchema = z.object({
```
Weiko, line 3: Seems to be too permissive maybe?

packages/twenty-server/src/engine/metadata-modules/ai/ai-models/services/ai-model-preferences.service.ts
```diff
@@ -1,21 +1,64 @@
...
+
+    try {
+      this.filePreferences = await this.fetchPreferences(storagePath);
+      this.logger.log(
+        `Loaded AI model preferences from storage: ${storagePath}`,
+      );
+    } catch (error: unknown) {
+      const message = error instanceof Error ? error.message : String(error);
+
+      this.logger.warn(
```
Weiko: Do we really want a "silent" log here?

packages/twenty-server/src/engine/core-modules/twenty-config/config-variables.ts
```diff
@@ -1410,7 +1410,16 @@ export class ConfigVariables {
     type: ConfigVariableType.JSON,
   })
   @IsOptional()
-  AI_MODEL_PREFERENCES: AiModelPreferences = loadDefaultModelPreferences();
+  AI_MODEL_PREFERENCES: AiModelPreferences = DEFAULT_MODEL_PREFERENCES;
```
prastoin, line 1413: That derivates from the current default standard defined by the default ai catalog stored as json in source code directly

Weiko (commented):
Not a strong opinion on this and we probably do that in other places and this might be intended but if the DB row is the source of truth after update and someone later inspects config/ai-model-preferences.json expecting to see the current state, they'll see the old state. Again, probably expected 👍

packages/twenty-server/src/engine/metadata-modules/ai/ai-models/services/ai-model-preferences.service.ts
```diff
@@ -79,13 +121,27 @@ export class AiModelPreferencesService {
...
       prefs[key] = current.filter((id) => !idSet.has(id));
     }
 
+    await this.persistPreferences(prefs);
+  }
+
+  private async persistPreferences(prefs: AiModelPreferences): Promise<void> {
     await this.twentyConfigService.set('AI_MODEL_PREFERENCES', prefs);
+    this.filePreferences = null;
+  }
```
prastoin, line 127: bit redundant

prastoin (approved):

packages/twenty-server/src/engine/metadata-modules/ai/ai-models/services/ai-model-preferences.service.ts
etiennejouan: I first followed the default-ai-catalog service logic but throwing error seems better

PR #20708, opened 2026-05-19
refactor(server): drop logo select workaround in flat-application cache

charlesBochet:
## Summary

Replaces the temporary `select: { ... }` workaround in `WorkspaceFlatApplicationMapCacheService` (introduced by #20159) with a property-level `@WasIntroducedInUpgrade` decorator on `ApplicationEntity.logo`.

#20159's own description called itself out: *"This is a temporary fix for cross-version upgrade process, a better fix would be to expose an hasInstanceCommandBeenRun() util (and later a decorator)"*. The decorator now exists, courtesy of #20686.

## Root cause recap

`ApplicationEntity.logo` is added by `2-2-instance-command-fast-1777539664664-add-logo-to-application.ts`. The column is declared on the entity class, so before that instance command runs (i.e. on a cross-version upgrade from a 2.1 or older baseline), TypeORM's bare `repository.find()` emits `SELECT \"logo\" …` against a table that doesn't have the column yet → upgrade aborts. #20159 worked around this by listing every column **except** `logo` in an explicit `select`, with an `as unknown as FindOptionsSelect<ApplicationEntity>` cast.


Weiko (approved):
LGTM, thanks!

PR #20713, opened 2026-05-19
fix(twenty-front): prevent connected account row overflow on long status label

### LINKED ISSUE #20594: Email settings not visible
## Bug Description

The “Advanced Settings” option is not displayed under Email Settings:

You have to scroll to the right to see the three dots for Advanced Settings.

URL: /settings/accounts/configuration/

**See screenshot:**

Image

Image

neo773:
Reproducible on German language

Before

SCR-20260519-ofhi

After

SCR-20260519-oesa

/closes #20594

### DIFF
--- a/packages/twenty-front/src/modules/settings/accounts/components/SettingsAccountsConnectedAccountsRowRightContainer.tsx
+++ b/packages/twenty-front/src/modules/settings/accounts/components/SettingsAccountsConnectedAccountsRowRightContainer.tsx
@@ -11,6 +11,19 @@ const StyledRowRightContainer = styled.div`
   align-items: center;
   display: flex;
   gap: ${themeCssVariables.spacing[4]};
+  min-width: 0;
+  overflow: hidden;
+`;
+
+const StyledStatusWrapper = styled.div`
+  display: flex;
+  min-width: 0;
+  overflow: hidden;
+`;
+
+const StyledDropdownMenuWrapper = styled.div`
+  display: flex;
+  flex-shrink: 0;
 `;
 
 export const SettingsAccountsConnectedAccountsRowRightContainer = ({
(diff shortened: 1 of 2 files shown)


packages/twenty-front/src/modules/settings/accounts/components/SettingsAccountsConnectedAccountsRowRightContainer.tsx
```diff
@@ -11,6 +11,19 @@ const StyledRowRightContainer = styled.div`
   align-items: center;
   display: flex;
   gap: ${themeCssVariables.spacing[4]};
+  min-width: 0;
+  overflow: hidden;
+`;
+
+const StyledStatusWrapper = styled.div`
```
charlesBochet: we should use regular tooling / of fix UI component if needed but not re-implment overflow: hidden everytime

Here I feel we want the tag to have a maxWidth and to internally overflow? is that a new behavior on the app?

If yes, we should enhance our tags, add stories (screenshot in the PR), etc.

charlesBochet (changes requested):

packages/twenty-front/src/modules/settings/accounts/components/SettingsAccountsConnectedAccountsRowRightContainer.tsx
charlesBochet: Or maybe looking at your screenshots on the PR, we don't need hidden at all?

neo773: removed defensive overflow was an oversight

charlesBochet (approved):

PR #20720, opened 2026-05-19
Fix BUILDER_INTERNAL_SERVER_ERROR message

ijreilly:
The throw site was passing (code, message) to a constructor whose signature is (message, code), so exception.message ended up as the literal string "BUILDER_INTERNAL_SERVER_ERROR" and the real error.message was stored in exception.code where nothing reads it. Swapping the two args puts the real error message back into exception.message, which is the field Yoga's error handler copies into the GraphQL response's top-level message — and that's the field the CLI prints.


Weiko (approved):
LGTM

PR #20730, opened 2026-05-19
Drop legacy rolePermissionFlag.flag column + fallback logic

Weiko:
## Summary
- **New fast migration** `2-7-instance-command-fast-1779600000000-finalize-role-permission-flag-cutover.ts`:
  - `DROP CONSTRAINT IDX_ROLE_PERMISSION_FLAG_FLAG_ROLE_ID_UNIQUE`
  - `ALTER COLUMN permissionFlagId SET NOT NULL`
  - `DROP COLUMN flag`
  - `down()` repopulates `flag` from the catalog via `permissionFlagId` and restores the old unique.
- **Entity**: `RolePermissionFlagEntity` hides the `flag` column by using the new decorator + drops old `@Unique` decorator; `permissionFlagId` and the `permissionFlag` relation become non-nullable.
- **Deletes** the synthesizer `synthesize-flat-permission-flag-from-flag.util.ts` and every fallback branch that used it
(`from-role-permission-flag-entity-to-flat-role-permission-flag.util.ts`, `from-flat-role-permission-flag-to-role-permission-flag-dto.util.ts`, `permissions.service.ts`,
`workspace-roles-permissions-cache.service.ts`, `fromRoleEntityToRoleDto.util.ts`, `flat-role-permission-flag-validator.service.ts`,
`role-permission-flag.service.ts:getEffectiveUniversalIdentifier`).
- **Write path**: ~~drops `flag` from `CreateRolePermissionFlagInput`, the create util, and the application-manifest converter.~~
- **Metadata configs**: ~~removes `flag` from `all-entity-properties-configuration-by-metadata-name.constant.ts` (rolePermissionFlag block)~~ and flips `permissionFlag.isNullable` to
`false` in `all-many-to-one-metadata-relations.constant.ts`.

### Why the `flag` field stays declared in the entity

The decorator (`@WasRemovedInUpgrade`) is the right tool for the lifecycle marker, but it's a **reflect-metadata** runtime decorator — TypeScript can't see it at compile time. So while
the adapter ([`UpgradeAwareEntityMetadataAdapter`](packages/twenty-server/src/engine/twenty-orm/upgrade-aware/upgrade-aware-entity-metadata.adapter.ts)) now correctly flips
`isSelect`/`isInsert`/`isUpdate` to `false` once the drop migration's cursor is crossed, the *static* TypeScript types derived from `RolePermissionFlagEntity`
(`UniversalFlatRolePermissionFlag`, `FlatRolePermissionFlag`, `MetadataEntityPropertyConfiguration<'rolePermissionFlag'>`, etc.) still see `flag` as a required scalar property — because the entity declares it.

That means every producer of one of those derived types must include `flag`:
- `from-create-role-permission-flag-input-to-flat-role-permission-flag-to-create.util.ts` plumbs it through.
- `from-permission-flag-to-universal-flat-role-permission-flag.util.ts` (the application-manifest converter) sets `flag: permissionFlag.flag`.
- `all-entity-properties-configuration-by-metadata-name.constant.ts` has a `flag` entry under `rolePermissionFlag`.
- `CreateRolePermissionFlagInput` keeps the `flag` field.
- `RolePermissionFlagService.upsertPermissionFlags` passes `flag: permissionFlag.key as PermissionFlagType` to the create util.

Explored phantom-brand approach (`RemovedInUpgrade<T>` wrapper on the field type, key-filter mapped type applied inside `ScalarFlatEntity` / `UniversalFlatEntityFrom`) but previous commands could have `flag === undefined` (downcast from the brand since we can't compare with UpgradeMigrationName like we do with a decorator). That's a **silent-read** failure mode: compiles fine, comparisons against `flag` silently always-false, no error surfaces. Probably worth too much risk for what's a small amount of plumbing?

The eventual full deletion of `flag` (entity field included) is a future cleanup once we drop cross-upgrade support for versions ≤ 2.6

Note: Not sure if this PR (and even the decorator) is really needed in the end, seems we need to keep a lot of code in place to handle legacy. Maybe a simple noop [At]Deprecated is enough @charlesBochet (and a migration to set the column nullable if that was not the case before + remove associated constraints)


packages/twenty-server/src/engine/twenty-orm/upgrade-aware/upgrade-aware-entity-metadata.adapter.ts
```diff
@@ -277,9 +279,15 @@ export class UpgradeAwareEntityMetadataAdapter implements OnModuleInit {
       metadata.givenTableName = resolved.effectiveTableName;
     }
 
+    metadata.columns = [...snapshot.canonicalColumns];
+
     for (const column of metadata.columns) {
       this.applyColumnShape({ column, snapshot, resolved });
     }
+
+    metadata.columns = metadata.columns.filter(
```
Weiko, line 288: Needed so typeorm migration generator does not pick the hidden columns (and to fix the CI)

packages/twenty-server/src/database/commands/upgrade-version-command/2-7/2-7-instance-command-fast-1779600000000-finalize-role-permission-flag-cutover.ts
```diff
@@ -0,0 +1,69 @@
...
+       ALTER COLUMN "permissionFlagId" SET NOT NULL`,
+    );
+
+    await queryRunner.query(
+      `ALTER TABLE "core"."rolePermissionFlag" DROP COLUMN "flag"`,
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
```
Weiko, line 27: a bit verbose... We never run the down anyway

packages/twenty-server/src/database/commands/upgrade-version-command/2-7/2-7-instance-command-fast-1779600000000-finalize-role-permission-flag-cutover.ts
```diff
@@ -0,0 +1,69 @@
...
+      `ALTER TABLE "core"."rolePermissionFlag"
+       ADD CONSTRAINT "IDX_ROLE_PERMISSION_FLAG_FLAG_ROLE_ID_UNIQUE"
+       UNIQUE ("flag", "roleId")`,
+    );
+
+    await queryRunner.query(
+      `ALTER TABLE "core"."rolePermissionFlag"
+       ALTER COLUMN "permissionFlagId" DROP NOT NULL`,
+    );
+  }
```
prastoin, line 68: ***Polish:*** Not even sure if it's worth it, maybe a simpler best effort down implem?

packages/twenty-server/src/database/commands/upgrade-version-command/2-7/2-7-instance-command-fast-1779600000000-finalize-role-permission-flag-cutover.ts
```diff
@@ -0,0 +1,69 @@
...
+{
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "core"."rolePermissionFlag"
+       DROP CONSTRAINT IF EXISTS "IDX_ROLE_PERMISSION_FLAG_FLAG_ROLE_ID_UNIQUE"`,
+    );
+
+    await queryRunner.query(
+      `ALTER TABLE "core"."rolePermissionFlag"
+       ALTER COLUMN "permissionFlagId" SET NOT NULL`,
```
prastoin, line 18: ***Question:*** Is it really a fast ? will this run successfully on data ?
I guess this has been backfilled previously

packages/twenty-server/src/engine/metadata-modules/role-permission-flag/utils/from-flat-role-permission-flag-to-role-permission-flag-dto.util.ts
```diff
@@ -2,31 +2,28 @@ import { PermissionFlagType } from 'twenty-shared/constants';
...
-        applicationId: flatRolePermissionFlag.applicationId,
-        applicationUniversalIdentifier:
-          flatRolePermissionFlag.applicationUniversalIdentifier,
-      });
+  if (!isDefined(permissionFlag)) {
+    throw new PermissionsException(
+      `Permission flag ${flatRolePermissionFlag.permissionFlagUniversalIdentifier} not found`,
+      PermissionsExceptionCode.PERMISSION_NOT_FOUND,
+    );
+  }
```
prastoin, line 26: That's damn better

packages/twenty-server/src/engine/metadata-modules/role-permission-flag/role-permission-flag.service.ts
```diff
@@ -98,12 +95,10 @@ export class RolePermissionFlagService {
...
         .filter(isDefined),
     );
-    const getEffectiveUniversalIdentifier = (
-      pf: FlatRolePermissionFlag,
-    ): string =>
-      pf.permissionFlagUniversalIdentifier ?? SystemPermissionFlag[pf.flag];
     const existingSet = new Set(
-      currentRolePermissionFlagsForRole.map(getEffectiveUniversalIdentifier),
+      currentRolePermissionFlagsForRole.map(
+        (pf) => pf.permissionFlagUniversalIdentifier,
```
prastoin, line 100: ***Out of scope:*** naming abrv

packages/twenty-server/src/engine/metadata-modules/role/utils/fromRoleEntityToRoleDto.util.ts
```diff
@@ -24,8 +24,7 @@ export const fromRoleEntityToRoleDto = (role: RoleEntity): RoleDTO => {
     permissionFlags: role.rolePermissionFlags?.map((rolePermissionFlag) => ({
       id: rolePermissionFlag.id,
       roleId: rolePermissionFlag.roleId,
-      flag: (rolePermissionFlag.permissionFlag?.key ??
-        rolePermissionFlag.flag) as PermissionFlagType,
+      flag: rolePermissionFlag.permissionFlag.key as PermissionFlagType,
```
prastoin, line 27: ***Challenge:*** From my understanding the `permissionFlag` is `string | PermissionFLatType` ( which leads to `string` in the end ) now no ?

packages/twenty-server/src/engine/twenty-orm/upgrade-aware/upgrade-aware-entity-metadata.adapter.ts
```diff
@@ -277,9 +279,15 @@ export class UpgradeAwareEntityMetadataAdapter implements OnModuleInit {
...
+      (column) => !resolved.hiddenPropertyNames.has(column.propertyName),
+    );
```
prastoin, line 290: ***Remark:*** This feels weird to filter out the columns we've been apply column shape too 🤔 
Also each column has a mutation side effect, so if someone would change this code location it would involve a regression

prastoin (approved):
Neat !

packages/twenty-server/src/database/commands/upgrade-version-command/2-7/2-7-instance-command-fast-1779600000000-finalize-role-permission-flag-cutover.ts
Weiko: The slow from the previous version took care of that

prastoin: Why wasn't it applying the non nullable constraint ?

packages/twenty-server/src/engine/metadata-modules/role/utils/fromRoleEntityToRoleDto.util.ts
Weiko: Yes! Will do in next PR when we'll handle 3rd party apps permission flags

packages/twenty-server/src/database/commands/upgrade-version-command/2-7/2-7-instance-command-fast-1779600000000-finalize-role-permission-flag-cutover.ts
Weiko: true, we should have done that in the same migration!

packages/twenty-server/src/engine/twenty-orm/upgrade-aware/upgrade-aware-entity-metadata.adapter.ts
Weiko: Noted, I can fix that in my next PR (same outcome/runtime behavior but cleaner code)

PR #20733, opened 2026-05-19
Update create twenty app scaffolded front component

martmull:
- Replace inline SVG icons with proper Avatar and icon components (IconBox, IconHierarchy, IconLayout, IconSettingsAutomation) from twenty-sdk/ui in the scaffolded front component
  template
  - Strip trailing slashes from workspace/API URLs in both create-twenty-app CLI and twenty-sdk remote commands to prevent malformed requests
  - Fix the application settings link to navigate to #installed anchor
  - Bump twenty-sdk, twenty-client-sdk, and create-twenty-app versions to 2.6.0

image

### DIFF
--- a/packages/twenty-sdk/src/cli/commands/remote.ts
+++ b/packages/twenty-sdk/src/cli/commands/remote.ts
@@ -117,7 +117,7 @@ export const registerRemoteCommands = (program: Command): void => {
           return;
         }
 
-        let apiUrl = options.apiUrl;
+        let apiUrl = options.apiUrl?.replace(/\/+$/, '');
 
         if (!apiUrl) {
           const detectedUrl = await detectLocalServer();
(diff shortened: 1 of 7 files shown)


packages/twenty-sdk/src/cli/commands/remote.ts
```diff
@@ -117,7 +117,7 @@ export const registerRemoteCommands = (program: Command): void => {
           return;
         }
 
-        let apiUrl = options.apiUrl;
+        let apiUrl = options.apiUrl?.replace(/\/+$/, '');
```
charlesBochet: seems duplicated 3 times, should we introduce an util (sanitizeApiUrl) or do we already have it in twenty-shared btw? this looks to be a common usecase

charlesBochet (commented):
Left a comment

charlesBochet (approved):

PR #20743, opened 2026-05-19
fix: render PAGE_LAYOUT nav items with standard icon tile and compute…

martmull:
## Before

image

## after

image


Weiko (approved):
LGTM

PR #20752, opened 2026-05-20
Deprecate and backfill delete `ConnectedAccount` twenty standard object

prastoin:
# Introduction
Following connected account permissions refactor and encryption
Removing the old workspace schema twenty standard application connectedAccount objects and related standard fields and index

- a lot of deadcode
- instance command backfill cleaning the connected account object from workspaces


Weiko (approved):
LGTM! I guess the next step will be calendarChannel,messageChannel and messageFolder standard objects?

Issue #20658, opened 2026-05-18
Compose/Reply fails with "SMTP is not configured for connected account" when SMTP is configured (works with iCloud, fails with all-inkl)


smemon-pixel:
<html>
<body>
<p class="font-claude-response-body break-words whitespace-normal leading-[1.7]">When replying to an email from a Twenty contact, the compose action fails with:</p>
<blockquote class="ml-2 border-l-4 border-border-300/10 pl-4 text-text-300">
<p class="font-claude-response-body break-words whitespace-normal leading-[1.7]">⚠ <strong>SMTP is not configured for connected account <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">3599bd54-f0a5-4d24-bde5-0221c2e82747</code></strong></p>
</blockquote>
<p class="font-claude-response-body break-words whitespace-normal leading-[1.7]">…even though SMTP <strong>is</strong> configured for that connected account via <em>Settings → Emails → Connected Accounts → IMAP/SMTP/CalDAV</em>. The connection was saved successfully (no errors during setup or test) and the IMAP side of the same account is working — incoming mail syncs and is visible in the timeline.</p>
<p class="font-claude-response-body break-words whitespace-normal leading-[1.7]">The bug is <strong>provider-specific</strong>:</p>
<div class="overflow-x-auto w-full px-2 mb-6">
Provider | SMTP host | SMTP send
-- | -- | --
✅ iCloud | smtp.mail.me.com:587 (STARTTLS) | Works
❌ all-inkl.com (KAS) | w01e6f4f.kasserver.com:587 (STARTTLS) | Fails with the error above

</div>
<p class="font-claude-response-body break-words whitespace-normal leading-[1.7]">Both accounts are configured the same way (per-user IMAP/SMTP/CalDAV connected account, STARTTLS on 587, valid credentials confirmed via <code class="bg-text-200/5 border border-0.5 border-border-300 text-danger-000 whitespace-pre-wrap rounded-[0.4rem] px-1 py-px text-[0.9rem]">swaks</code>/Thunderbird against the same all-inkl server).</p>
<p class="font-claude-response-body break-words whitespace-normal leading-[1.7]"><strong>Screenshot:</strong>

<img width="728" height="380" alt="Image" src="https://github.com/user-attachments/assets/274238d2-2d20-4e8e-bdd4-d239ce7dfbfa" />

[...]

PR #20755, opened 2026-05-20
fix(email): resolve reply account from thread channel

Vinzz2303:
## Summary

Fixes the reply account resolution path so email replies use the connected account associated with the thread's message channel instead of blindly selecting the first connected account.

This targets the failure mode reported in #20658 where replying can surface `SMTP is not configured for connected account` even though the thread's actual IMAP/SMTP account has SMTP configured.

## Changes

- Query `myMessageChannels` alongside `myConnectedAccounts` in `useEmailThread`
- Resolve the latest message's `messageChannelId` to its `connectedAccountId`
- Return the matching connected account for reply context instead of `myConnectedAccounts[0]`
- Add a regression test proving the hook chooses the thread channel's account when the first account is different

## Validation

- `git diff --check`

I could not run the full frontend test locally in this workspace because dependency installation failed with `ENOSPC: no space left on device` while Yarn was cloning a dependency into the Windows temp/cache path. The code change is intentionally narrow and covered by the added hook regression test.


Vinzz2303:
Pushed a follow-up formatting pass on the touched files. Local validation now passes:

- `prettier --check` on `useEmailThread.ts` and `useEmailThread.test.ts`
- `git diff --check`

The PR is ready for review; CI has been re-triggered on the latest commit.

neo773 (approved):
LGTM

Vinzz2303:
Hi @neo773, thanks again for the review and merge. Could you confirm whether #20755 is associated with the Twenty IMAP Algora bounty, or if there is a preferred claim flow I should follow?

Vinzz2303:
Hi @neo773, just following up after #20755 was merged. Could you confirm whether this PR is eligible for the Twenty IMAP Algora bounty, or if there is a claim step I should complete on Algora?

neo773:
@Vinzz2303 
I think there’s a misunderstanding here. The bounty was already claimed a year ago there is no bounty for you to “claim”

Vinzz2303:
@neo773 Hi, thanks for clarifying. Sorry for the misunderstanding I thought the bounty was still active.

I understand there is no bounty to claim. If any part of my contribution is still useful to the project and there is a way to support contributors through sponsorship or a small appreciation, I’d really appreciate it. But no worries if that’s not possible.

Thanks for your time.

PR #20763, opened 2026-05-20
feat(sdk): error on incompatible view filter operand at sync time

ehconitin:
view filters with mismatched operand + field type now error at sync -- was silently failing before


packages/twenty-server/src/engine/workspace-manager/workspace-migration/workspace-migration-builder/validators/services/flat-view-filter-validator.service.ts
```diff
@@ -200,4 +229,33 @@ export class FlatViewFilterValidatorService {
...
+    fieldType,
+    subFieldName,
+  }: {
+    operand: ViewFilterOperand;
+    fieldType: FieldMetadataType;
+    subFieldName: string | null | undefined;
+  }) {
+    if (!(fieldType in FILTER_OPERANDS_MAP)) {
+      return undefined;
+    }
```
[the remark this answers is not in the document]
ehconitin: we dont do this anywhere else -- not sure if its deliberate. defer? @martmull  wdyt

packages/twenty-server/src/engine/metadata-modules/view-filter/dtos/inputs/create-view-filter.input.ts
```diff
@@ -27,7 +27,7 @@ export class CreateViewFilterInput {
 
   @IsOptional()
   @IsEnum(ViewFilterOperand)
-  @Field({ nullable: true, defaultValue: ViewFilterOperand.CONTAINS })
```
martmull, line 30: why removing defaultValue here?

martmull (commented):
do you think we can add some basic validation in `packages/twenty-sdk/src/sdk/define/views/define-view.ts` so we can reject wrong views before calling syncApplication endpoint? Also, i am wondering if we could leverage typescript and provide typing check on filter compatibility in packages/twenty-shared/src/application/viewManifestType.ts

packages/twenty-server/src/engine/metadata-modules/view-filter/dtos/inputs/create-view-filter.input.ts
ehconitin: no defaultValue can be right here because it depends on the referenced field's type :)

ehconitin:
failing qa - 

CleanShot 2026-05-21 at 21 30 15@2x

CleanShot 2026-05-21 at 21 26 39@2x


passing qa - 

CleanShot 2026-05-21 at 21 39 51@2x

CleanShot 2026-05-21 at 21 39 32@2x

ehconitin:
@martmull 

(a) done -- bad operands now fail when building the app, not just at sync --uses the same check as server side.

(b) holding off for now -- for typescript to catch it we'd need to change how filters point at fields? (today its just a UUID string with no type info) -- and no clear strategy on how to support it widely no?

martmull (approved):

PR #20765, opened 2026-05-20
Ci server custom jest reporter 

prastoin:
# Introduction
Only display failing unit tests trace in the ci-server server-test jobs
So it's possible to identify what unit test are failing without having to re run them locally

### DIFF
--- a/packages/twenty-server/jest.config.mjs
+++ b/packages/twenty-server/jest.config.mjs
@@ -9,7 +9,7 @@ const jestConfig = {
   prettierPath: null,
   // to enable logs, comment out the following line
   silent: true,
-  ...(isCI && { reporters: ['./ci-reporter.js'] }),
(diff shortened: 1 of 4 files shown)


packages/twenty-server/jest.config.mjs
```diff
@@ -1,12 +1,15 @@
 import { createRequire } from 'module';
 const require = createRequire(import.meta.url);
 
+const isCI = process.env.CI === 'true';
```
Weiko, line 4: Why not using the existing test env instead? I didn't know we had this CI env

prastoin: process.env.CI is standard env var mostly defined by most cicd providers directly

charlesBochet (approved):

charlesBochet (commented):
Does not look super robust but LGTM

prastoin:
Why's that @charlesBochet ?

PR #20769, opened 2026-05-20
fix(auth): clarify error when joining a non-active workspace

FelixMalfait:
## Summary

When an existing user accepts an invite into a workspace whose `activationStatus` is not `ACTIVE` (e.g. `SUSPENDED`, `INACTIVE`, `PENDING_CREATION`), the throw in `throwIfWorkspaceIsNotReadyForSignInUp` returns:

> User is not part of the workspace

The message describes the symptom (they aren't a member yet) instead of the cause (the workspace can't accept new members), which makes invitees assume their invite is broken when the real issue is the target workspace's state.

The sibling branch a few lines above — for brand-new users hitting the same non-ACTIVE workspace — already returns `"Workspace is not ready to welcome new members"`. This PR reuses the same message in the existing-user branch so both paths give a consistent, accurate explanation.

Single file, two string literals.

## Test plan

- [ ] Sign in via Google with an existing Twenty account, accepting an invite to a `SUSPENDED` workspace → confirm the new message is shown instead of "User is not part of the workspace".
- [ ] Confirm the happy path (sign-in to an `ACTIVE` workspace via invite) is unchanged — early-return on `ACTIVE` is untouched.

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/auth/services/sign-in-up.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/auth/services/sign-in-up.service.ts
@@ -231,26 +231,31 @@ export class SignInUpService {
   ) {
     if (workspace.activationStatus === WorkspaceActivationStatus.ACTIVE) return;
 
-    // Existing members can still sign in to a non-active workspace (e.g.
-    // to resolve billing on a SUSPENDED one). We only block adding new
-    // members.
-    if (user.userData.type === 'existingUser') {
-      const userWorkspaceExists =
-        await this.userWorkspaceService.checkUserWorkspaceExists(
-          user.userData.existingUser.id,
-          workspace.id,
-        );
-
-      if (userWorkspaceExists) return;
+    if (user.userData.type !== 'existingUser') {
+      throw new AuthException(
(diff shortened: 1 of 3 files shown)


packages/twenty-server/src/engine/core-modules/auth/services/sign-in-up.service.ts
```diff
@@ -231,31 +231,26 @@ export class SignInUpService {
...
 
-    if (user.userData.type !== 'existingUser') {
-      throw new AuthException(
-        'Workspace is not ready to welcome new members',
-        AuthExceptionCode.FORBIDDEN_EXCEPTION,
-        {
-          userFriendlyMessage: msg`Workspace is not ready to welcome new members`,
-        },
-      );
+    // Existing members can still sign in to a non-active workspace (e.g.
```
charlesBochet: I don't understand why we refactor the code. To me the only logical change is that the error message has change

charlesBochet (approved):
Approving but the fix looks convoluted. It could actually just be changing the error message, no need to re-arrange the logic which was made of early returns and is now made of nested if

PR #20779, opened 2026-05-20
Add editor mode for text field widgets

Bonapara:
## Summary

Tested ↓

CleanShot 2026-05-20 at 21 14 04@2x

- Enable `FieldDisplayMode.EDITOR` for plain `TEXT` field widgets while keeping `FIELD` as the default display mode.
- Add a plain multiline text editor renderer for `TEXT + EDITOR` field widgets with optimistic record-store/cache updates and debounced persistence.
- Reuse the shared `TextArea` component through a transparent, uncapped variant so the editor has no input chrome and lets the widget grow.
- Add unit coverage for text display-mode config and a Storybook scenario for a text field widget in editor mode.

## useEffect cleanup note

`FieldWidgetTextEditor` flushes the debounced persist callback in a `useEffect` cleanup:

```ts
useEffect(() => () => persistTextDebounced.flush(), [persistTextDebounced]);
```

This follows the existing debounced autosave cleanup pattern already used in `WorkflowEditActionHttpRequest`. It ensures pending text changes are persisted when the widget unmounts, while `onBlur` still flushes immediately for normal editor exits.

## Validation

- `npx nx test twenty-front --testPathPattern=page-layout`
- `npx nx typecheck twenty-front`
- `npx nx lint twenty-front`
- Browser check on `http://apple.localhost:3001/object/company/20202020-a305-41e7-8c72-ba44072a4c58` for transparent textarea, no internal max-height/scroll, equal padding, and widget growth.

Note: lint passes with two unrelated existing warnings in `NavigationDrawerItem.tsx` and `ConfigVariableEdit.tsx`.


packages/twenty-front/src/modules/page-layout/widgets/field/components/FieldWidgetTextEditor.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  const handleChange = useCallback(
+    (text: string) => {
+      if (isRecordFieldReadOnly === true) {
+        return;
+      }
+
+      setFieldValue(text);
+
+      modifyRecordFromCache({
```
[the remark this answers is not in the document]
Bonapara: RichTextFieldEditor does the same kind of local store/cache update before calling its debounced persist function. Since this PR was explicitly matching rich-text editor behavior, the text editor inherits that existing autosave tradeoff.

packages/twenty-front/src/modules/page-layout/widgets/field/components/FieldWidgetTextEditor.tsx
```diff
@@ -0,0 +1,118 @@
...
+    updateOneRecord({
+      objectNameSingular: objectMetadataItem.nameSingular,
+      idToUpdate: recordId,
+      updateOneRecordInput: {
+        [fieldName]: text,
+      },
+    });
+  }, 300);
+
+  useEffect(() => () => persistTextDebounced.flush(), [persistTextDebounced]);
```
charlesBochet, line 64: Let's avoid that

packages/twenty-front/src/modules/page-layout/widgets/field/components/__stories__/FieldWidget.stories.tsx
```diff
@@ -425,6 +425,102 @@ export const TextFieldWidget: Story = {
...
+        configurationType: WidgetConfigurationType.FIELD,
+        fieldMetadataId: nameField.id,
+        fieldDisplayMode: FieldDisplayMode.EDITOR,
+      },
+      createdAt: '2024-01-01T00:00:00Z',
+      updatedAt: '2024-01-01T00:00:00Z',
+      deletedAt: null,
+    };
+
+    setTestObjectMetadataItemsInMetadataStore(
```
charlesBochet: looks like we have a lot of duplicated code

packages/twenty-front/src/modules/page-layout/widgets/field/components/FieldWidgetTextEditor.tsx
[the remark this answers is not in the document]
charlesBochet: we should investigate why this is needed, it should not

packages/twenty-front/src/modules/ui/input/components/TextArea.tsx
```diff
@@ -11,19 +11,22 @@ import { themeCssVariables } from 'twenty-ui/theme-constants';
 
 const MAX_ROWS = 5;
 
+type TextAreaVariant = 'default' | 'transparent';
```
charlesBochet, line 12: we should add stories for them

packages/twenty-front/src/modules/ui/input/components/TextArea.tsx
```diff
@@ -11,19 +11,22 @@ import { themeCssVariables } from 'twenty-ui/theme-constants';
...
+
 export type TextAreaProps = {
   textAreaId: string;
   label?: string;
   disabled?: boolean;
   height?: number;
   minRows?: number;
-  maxRows?: number;
+  maxRows?: number | null;
```
charlesBochet: maxRows was already undefinable

packages/twenty-front/src/modules/page-layout/widgets/field/components/FieldWidgetTextEditor.tsx
charlesBochet: tech debt but hard to avoid
 
 keeping the `useEffect(() => () => persistTextDebounced.flush(), ...)` cleanup to match the existing autosave pattern already used in `WorkflowEditActionHttpRequest`, `WorkflowEditActionCreateRecord`, `WorkflowEditActionFormBuilder`, `WorkflowEditActionDeleteRecord`, `WorkflowEditActionUpdateRecord`, `WorkflowEditActionUpsertRecord`, `WorkflowEditActionEmailBase`, `WorkflowEditActionFormFiller`, `WorkflowEditActionFindRecords`, `SettingsSkillForm`, and `SettingsAgentForm` (11 sites). `onBlur` alone doesn't reliably catch unmount-without-blur cases like tab switches, so dropping the cleanup here would risk silent data loss. Happy to remove it later if we agree to rip out the pattern repo-wide.

charlesBochet (approved):

PR #20781, opened 2026-05-20
Fix: Added description below application variable

### LINKED ISSUE #20757: App Settings tab doesn't render description field from applicationVariables
Been building a Twenty app and ran into something that seems like a gap worth flagging.

In `defineApplication()`, each entry in `applicationVariables` accepts a `description` field:

```ts
applicationVariables: {
  FIELD_CREATION_ENABLED: {
    universalIdentifier: '...',
    value: 'true',
    description: 'When enabled, automatically creates custom fields when unknown keys appear in payloads.',
  },
}
```

The description is there in the type definition (`NonSecretApplicationVariable.description?: string`) but nothing renders it in the Settings tab UI. Users just see the raw variable key (e.g. `FIELD_CREATION_ENABLED`) and an input — no context about what the setting actually does.

This makes the Settings tab pretty hard to use for anyone who didn't write the app, since the variable names alone aren't self-explanatory.

**What I'd expect:** the description showing up as either helper text below the input or a small tooltip/info icon next to the label — similar to how most settings UIs handle this.

Had to work around it by using `settingsCustomTabFrontComponentUniversalIdentifier` to build a custom panel, which feels like more than should be necessary just to explain what a setting does.

Would be great to have this rendered natively so the built-in Settings tab is actually useful for end users.

git-init-priyanshu:
Fixes #20757 

Before: 
Screenshot 2026-05-21 at 12 51 25 AM

After:
Screenshot 2026-05-21 at 12 50 52 AM


packages/twenty-front/src/pages/settings/applications/tabs/SettingsApplicationDetailEnvironmentVariablesTable.tsx
```diff
@@ -38,24 +43,28 @@ export const SettingsApplicationDetailEnvironmentVariablesTable = ({
...
+                    }
+                    return val;
+                  }),
+                );
+                onUpdateDebounced({ ...editedEnvVariable, value: newValue });
+              }}
+              placeholder={t`Value`}
+              fullWidth
+            />
+            <StyledDescription>
```
etiennejouan: And if description is undefined ?

```
{isNonEmptyString(editedEnvVariable.description) && (
  <StyledDescription>
    {editedEnvVariable.description}
  </StyledDescription>
)}
```

etiennejouan (approved):
Hi @git-init-priyanshu, thank you so much for your contribution.

Add design changes.

on hover
Image
with description
Image
without
Image

git-init-priyanshu:
> Hi @git-init-priyanshu, thank you so much for your contribution.
> 
> Add design changes.
> 
> on hover Image with description Image without Image

Thank you @etiennejouan for the changes. You could have asked me to make those changes, I am happy to contribute.

PR #20794, opened 2026-05-21
Fix must wait 3 days to create app in twenty-apps

martmull:
as title

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -10,4 +10,8 @@ nodeLinker: node-modules
 
 npmMinimalAgeGate: 3d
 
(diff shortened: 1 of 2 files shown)


prastoin (changes requested):
What's the issue you're encountering ?

prastoin (commented):
Isn't it the default value ?
Looks like a yarnrc scope issue rather than a configuration one

prastoin:
Would rather whitelist our known repo in this yarnrc instead

prastoin (approved):
Could remove twenty-create-app
Approved !

PR #20797, opened 2026-05-21
Navigate to installed page after app install

martmull:
as title


FelixMalfait:
I feel like there's something to do with connected accounts. Should an app be able to express when var or connected accounts are mandatory? Should we redirect to the setting/config tab or highlight it? Maybe it can be in README for v1 but curious what you had in mind

martmull:
> I feel like there's something to do with connected accounts. Should an app be able to express when var or connected accounts are mandatory? Should we redirect to the setting/config tab or highlight it? Maybe it can be in README for v1 but curious what you had in mind

we will improve the overall experience with Thomas, I just open atomic PRs to improve it gradually, starting with quick wins

packages/twenty-front/src/pages/settings/applications/SettingsAvailableApplicationDetails.tsx
```diff
@@ -98,9 +100,20 @@ export const SettingsAvailableApplicationDetails = () => {
 
   const handleInstall = async () => {
     if (isDefined(detail)) {
-      await install({
+      const success = await install({
         universalIdentifier: detail.universalIdentifier,
       });
+
+      if (success) {
+        const { data } = await refetchApplication();
```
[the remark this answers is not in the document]
prastoin: cubic is right we're missing a catch block here, though it was already the case before

prastoin (approved):

packages/twenty-front/src/pages/settings/applications/SettingsAvailableApplicationDetails.tsx
```diff
@@ -98,9 +100,20 @@ export const SettingsAvailableApplicationDetails = () => {
 
   const handleInstall = async () => {
     if (isDefined(detail)) {
-      await install({
+      const success = await install({
         universalIdentifier: detail.universalIdentifier,
       });
+
+      if (success) {
```
prastoin: ***Polish:*** early return

packages/twenty-front/src/pages/settings/applications/SettingsAvailableApplicationDetails.tsx
```diff
@@ -43,14 +44,15 @@ export const SettingsAvailableApplicationDetails = () => {
...
+  const navigateSettings = useNavigateSettings();
   const { install, isInstalling } = useInstallMarketplaceApp();
   const { upgrade, isUpgrading } = useUpgradeApplication();
 
   const canInstallMarketplaceApps = useHasPermissionFlag(
     PermissionFlagType.MARKETPLACE_APPS,
   );
 
-  const { data: applicationData } = useQuery(
+  const { data: applicationData, refetch: refetchApplication } = useQuery(
```
prastoin: ***Question:*** Isn't this mutation returning the installed application ?

PR #20803, opened 2026-05-21
Upload application file resolver exception management and integration coverage

prastoin:
# Introduction
Earlier and better exception handling of the upload application file resolver
+ coverage

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/application/application-development/application-development.resolver.ts
+++ b/packages/twenty-server/src/engine/core-modules/application/application-development/application-development.resolver.ts
@@ -219,6 +219,19 @@ export class ApplicationDevelopmentResolver {
       );
     }
 
+    const application =
+      await this.applicationService.findByUniversalIdentifier({
+        universalIdentifier: applicationUniversalIdentifier,
+        workspaceId,
+      });
+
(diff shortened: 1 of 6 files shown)


packages/twenty-server/src/engine/core-modules/application/application-development/application-development.resolver.ts
```diff
@@ -219,6 +219,20 @@ export class ApplicationDevelopmentResolver {
       );
     }
 
+    const application = await this.applicationService.findByUniversalIdentifier(
```
charlesBochet, line 222: this looks a bit fragile, shouldn't we check this within fileStorageService ?

charlesBochet: or will the throw anyway and it's only to improve the message?

charlesBochet (approved):

packages/twenty-server/src/engine/core-modules/application/application-development/application-development.resolver.ts
prastoin: it was throwing a 500 deeper indeed

PR #20673, opened 2026-05-18
`[CONNECTED_ACCOUNT_BREAKING_CHANGE]` Encrypt `ConnectedAccount` `connectionParameters`

prastoin:
# Introduction
Prevent any cross user `connectedAccount` `connectionParamaters` leak
Also encrypt in db all `connectionParameters` password
Never return any password through `DTO` anymore
The settings now allow update mutation without providing the password in edition mode

Verified all `connectionParameters.password` interaction

## Integration tests
- Added more coverage for both failing and successful paths
- Introduced a new env var that allow bypass the provider connection test

## Legacy connected Account decryption support
Stop allowing non encrypted decryption on `accessToken` and `refreshToken`, only allow legacy decryption on refactored `connectionParameters`

## Upsert ownership
Completely got rid of the connected workspace schema context which is legacy
Also now a user can only upsert a connected account for him only..

## New UI
image
If in edition the password is by default disabled
It needs to be selected as being edited to be enabled

## Next
- Refactor tool permissions flag not to include connected accounts
- Remove the legacy connected standard object
- Refactor and improve connected account resolver auth

PR #20807, opened 2026-05-21
feat(server): rotate connectedAccount.connectionParameters via secret-encryption:rotate

charlesBochet:
## Summary

Adds a new `connected-account-connection-parameters` site to the `secret-encryption:rotate` CLI introduced in #20613, so the nested password envelopes inside `connectedAccount.connectionParameters` (IMAP / SMTP / CALDAV — encrypted at-rest in #20673) are re-encrypted under the current `ENCRYPTION_KEY` alongside every other at-rest secret site. Without this, rotating `ENCRYPTION_KEY` on a 2.7+ instance would silently leave IMAP / SMTP / CalDav passwords on the old key id.

### Why a new handler

A dedicated handler is required (rather than reusing `ColumnRotationSiteHandler`) because the envelope lives at `connectionParameters->'<PROTOCOL>'->>'password'`, not in the whole column, and up to three independent envelopes may need rotating per row. The handler:

- Uses the same cursor-based, idempotent, online pattern as the existing handlers, with a SQL predicate that skips rows where every non-null protocol password is already on the current key id.
- Threads \`workspaceId\` into HKDF, matching how \`EncryptConnectionParametersSlowInstanceCommand\` backfilled.
- Rebuilds only the protocols whose passwords are not yet current, so a partial mid-row failure cannot cause unnecessary re-encryption on resume.
- Guards the UPDATE with jsonb-level deep equality (\`IS NOT DISTINCT FROM CAST(:json AS jsonb)\`) so optimistic concurrency is unaffected by Postgres's internal jsonb key ordering vs. JSON.stringify ordering.
- Refuses to rotate plaintext passwords (counted as \`errors\`) — operators must finish the 2.7 slow instance command (\`EncryptConnectionParametersSlowInstanceCommand\`) before running rotation.

### Sites covered (now)

| Site | Location | Scope |
| --- | --- | --- |
| \`connected-account-access-token\` | \`connectedAccount.accessToken\` | workspace |
| \`connected-account-refresh-token\` | \`connectedAccount.refreshToken\` | workspace |
| **\`connected-account-connection-parameters\`** (new) | \`connectedAccount.connectionParameters.{IMAP,SMTP,CALDAV}.password\` | workspace |
| \`application-variable\` | \`applicationVariable.value\` (isSecret) | workspace |
| \`application-registration-variable\` | \`applicationRegistrationVariable.encryptedValue\` | instance |
| \`signing-key-private-key\` | \`signingKey.privateKey\` | instance |
| \`totp-secret\` | \`twoFactorAuthenticationMethod.secret\` | workspace |
| \`sensitive-config-storage\` | \`keyValuePair.value\` (sensitive STRING configs) | instance |

### DIFF
--- a/packages/twenty-server/src/database/commands/secret-encryption-rotation/constants/secret-encryption-rotation-site-name.constant.ts
+++ b/packages/twenty-server/src/database/commands/secret-encryption-rotation/constants/secret-encryption-rotation-site-name.constant.ts
@@ -1,6 +1,8 @@
 export const SECRET_ENCRYPTION_ROTATION_SITE_NAME = {
   CONNECTED_ACCOUNT_ACCESS_TOKEN: 'connected-account-access-token',
   CONNECTED_ACCOUNT_REFRESH_TOKEN: 'connected-account-refresh-token',
+  CONNECTED_ACCOUNT_CONNECTION_PARAMETERS:
+    'connected-account-connection-parameters',
   APPLICATION_VARIABLE: 'application-variable',
   APPLICATION_REGISTRATION_VARIABLE: 'application-registration-variable',
   SIGNING_KEY_PRIVATE_KEY: 'signing-key-private-key',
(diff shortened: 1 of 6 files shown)


charlesBochet:
Merging as this follows same pattern already carefully reviewed. I'll take feedback if needed tomorrow

prastoin (commented):
looks good to me ! nothing more than nitpicks

PR #20809, opened 2026-05-21
fix(ai-chat) - fix browser context injection

etiennejouan:
Move the browsing context out of the system prompt and injecting it directly into the last user message instead.
Previously (before this PR), browsing context change, update system prompt then break whole conversation history ... and caching. Now, browsing context is sent with last message only if changed.

"Benchmark" this PR vs main : 
- same conv with 3-4 turns - 60% -> 85% cache ratio || 0.31 credits -> 0.13

### DIFF
--- a/packages/twenty-front/src/modules/ai/hooks/useAgentChat.ts
+++ b/packages/twenty-front/src/modules/ai/hooks/useAgentChat.ts
@@ -46,7 +46,6 @@ export const useAgentChat = (
 
   const [, setPendingThreadIdAfterFirstSend] = useState<string | null>(null);
 
-
   const setAgentChatUploadedFiles = useSetAtomState(
     agentChatUploadedFilesState,
   );
--- a/packages/twenty-front/src/modules/ai/hooks/useAgentChat.ts
+++ b/packages/twenty-front/src/modules/ai/hooks/useAgentChat.ts
@@ -15,6 +15,7 @@ import { SEND_CHAT_MESSAGE } from '@/ai/graphql/mutations/sendChatMessage';
 import { STOP_AGENT_CHAT_STREAM } from '@/ai/graphql/mutations/stopAgentChatStream';
 import { useAgentChatModelId } from '@/ai/hooks/useAgentChatModelId';
 import { useGetBrowsingContext } from '@/ai/hooks/useBrowsingContext';
+import { agentChatLastSentBrowsingContextFamilyState } from '@/ai/states/agentChatLastSentBrowsingContextFamilyState';
 import { useOptimisticallyUnarchiveOnSend } from '@/ai/hooks/useOptimisticallyUnarchiveOnSend';
 import {
   AGENT_CHAT_NEW_THREAD_DRAFT_KEY,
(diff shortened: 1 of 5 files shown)


packages/twenty-front/src/modules/ai/hooks/useAgentChat.ts
```diff
@@ -151,13 +163,17 @@ export const useAgentChat = (
...
           messageId,
-          browsingContext: browsingContext ?? null,
+          browsingContext: browsingContextToSend ?? null,
           modelId: modelIdForRequest ?? undefined,
           fileAttachments:
             fileAttachments.length > 0 ? fileAttachments : undefined,
         },
       });
 
+      if (isDefined(browsingContextChanged)) {
```
Weiko: is it not a boolean? Should always be false/true right?

packages/twenty-front/src/modules/ai/hooks/useAgentChat.ts
```diff
@@ -151,13 +163,17 @@ export const useAgentChat = (
           threadId,
           text: contentToSend,
           messageId,
-          browsingContext: browsingContext ?? null,
+          browsingContext: browsingContextToSend ?? null,
```
Weiko: '?? null' seems redundant now since you do
```typescript
const browsingContextToSend = browsingContextChanged
      ? browsingContext
      : null
```

Weiko (approved):
LGTM!

packages/twenty-front/src/modules/ai/hooks/useAgentChat.ts
etiennejouan: Right, and you can be sure this one line is 100% me (not ai) generated... Rename constant

packages/twenty-front/src/modules/ai/hooks/useAgentChat.ts
```diff
@@ -96,6 +97,17 @@ export const useAgentChat = (
     }));
 
     const browsingContext = getBrowsingContext();
+    const lastSentBrowsingContextAtom =
+      agentChatLastSentBrowsingContextFamilyState.atomFamily(threadId);
+    const lastSentBrowsingContext = store.get(lastSentBrowsingContextAtom);
+    const isBrowsingContextChanged =
```
ijreilly, line 103: nit / `has...Changed` or `is...Different`

Weiko (approved):

packages/twenty-server/src/engine/metadata-modules/ai/ai-chat/services/chat-execution.service.ts
```diff
@@ -226,11 +222,22 @@ export class ChatExecutionService {
       }
     }
 
+    if (isDefined(browsingContext)) {
+      const contextString = this.buildContextFromBrowsingContext(
+        workspace,
+        browsingContext,
+      );
+
+      processedMessages = this.injectBrowsingContextIntoLastUserMessage(
```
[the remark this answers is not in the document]
ijreilly: not relevant because ce use processedMessages to compute the token-size right? @etiennejouan

ijreilly (approved):

packages/twenty-front/src/modules/ai/hooks/useAgentChat.ts
etiennejouan: merged :/ i'll do it

PR #20811, opened 2026-05-21
Fix workflow cron trigger cache stuck without TTL

ijreilly:
## Problem

The cron-trigger cache key (`module:workflow:workflow-cron-triggers`) can get stuck without a TTL, silently halting **all** cron-triggered workflows for a whole tenant until the key is manually deleted from Redis.

Repro path:

1. Cache miss → DB-scan branch runs.
2. Inner loop writes triggers via `hashSet` (creates the key, **no TTL yet**).
3. Worker crashes / OOMs / gets killed by a deploy between any `hashSet` and the trailing `expire(1h)` call.
4. Key now exists with TTL = `-1` and a partial set of fields.
5. Next tick: `hashGetValues` returns those fields → `cachedValues.length > 0` → **cache-hit branch** → `expire` is never called.
6. Key has no TTL, so it never auto-expires. The DB-scan branch never runs again. New / missing triggers are never picked up. Workflows go silent.

Observed in production: cache key with `TTL: no limit` and 121 fields. Deleting the key restored normal behaviour (next tick rebuilt with TTL ~3600).

## Fix

Make the write + TTL set **atomic** so the key can never be observable without a TTL.

- Add `CacheStorageService.hashSetWithExpire` that runs `HSET` + `PEXPIRE` in a single Lua `EVAL`, mirroring the existing `hashSetIfExists` pattern.
- Replace the `hashSet` + trailing `expire` pair in `WorkflowCronTriggerCronJob` with `hashSetWithExpire`, called per trigger.
- Drop the now-redundant `if (triggerCount > 0) expire(...)` block.

## Test plan

- [x] `workflow-cron-trigger-cron.job.spec.ts` updated; all 10 tests pass
- [x] `oxlint` + `prettier` clean on changed files
- [ ] Verify on the affected self-hosted instance: after this lands, deleting the key once more (or letting it expire) should produce a key whose TTL is always set and refreshed each hour

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/cache-storage/services/cache-storage.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/cache-storage/services/cache-storage.service.ts
@@ -342,6 +342,37 @@ end`;
     }) as Promise<number>;
   }
 
+  // Atomic HSET + PEXPIRE. Use when the caller must guarantee that the key
+  // is never observable without a TTL — otherwise a crash between a plain
+  // hashSet and a follow-up expire() can leave the key alive forever.
+  async hashSetWithExpire({
+    key,
+    field,
+    value,
+    ttlMs,
+  }: {
+    key: string;
+    field: string;
+    value: string;
+    ttlMs: Milliseconds;
+  }): Promise<number> {
+    if (!this.isRedisCache()) {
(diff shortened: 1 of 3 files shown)


ijreilly:
Reopened under correct branch name.

PR #20816, opened 2026-05-21
Fix/15800 sanitize optimistic record input

kennedydqz-del:
## Summary

  `computeOptimisticRecordFromInput` throws when the `recordInput` contains fields not present in the object's metadata.   The raw `recordInput` was passed directly to this function, while `sanitizeRecordInput` (which filters unknown         fields) was only applied for the GraphQL mutation.

  This caused an uncaught error: "Should never occur, encountered unknown fields name in objectMetadataItem task" when
  clicking "+ Add New" from relation fields on custom objects pointing to Tasks or Notes.

  ## Changes

  | File | Change |
  |------|--------|
  | `useCreateOneRecord.ts` | Use `sanitizedInput` instead of raw `recordInput` for optimistic cache computation |
  | `useCreateManyRecords.ts` | Use `sanitizedRecord` instead of raw `recordToCreate` for optimistic cache computation |
  | `jest.config.mjs` | Add `rxjs`/`uuid` to transform allowlist; map `twenty-ui`/`@ui` aliases to source so tests can
  run without pre-built twenty-ui |

  ## Test plan

  All 8 existing tests pass:
  ✓ should generate correct optimistic record if no relation field is present
  ✓ should generate correct optimistic record with actor field
  ✓ should generate correct optimistic record createdBy when recordInput contains id
  ✓ should generate correct optimistic record if relation field is present but cache is empty
  ✓ should generate correct optimistic record even if recordInput contains field __typename
  ✓ should generate correct optimistic record if relation field is present and cache is not empty
  ✓ should generate correct optimistic record if relation field is null and cache is empty
  ✓ should throw an error if recordInput contains fields unrelated to the current objectMetadata


prastoin (changes requested):
Hello there, auto close
Abusive workflow edition
Please scope your PR

PR #20819, opened 2026-05-21
fix(server): encrypt token post refresh

prastoin:
# Introduction
Jobs were refreshing token and returning them as plain text, resulting to underlying code flow failure as expecting encrypted tokens

## Next
We should define a strong typescript signature to avoid such things to happen again, or least have an explicit naming

### DIFF
--- a/packages/twenty-server/src/modules/connected-account/refresh-tokens-manager/services/connected-account-refresh-tokens.service.ts
+++ b/packages/twenty-server/src/modules/connected-account/refresh-tokens-manager/services/connected-account-refresh-tokens.service.ts
@@ -37,7 +37,6 @@ export class ConnectedAccountRefreshTokensService {
     private readonly connectedAccountRepository: Repository<ConnectedAccountEntity>,
   ) {}
 
-  // Returns encrypted tokens safe to store on in-memory entities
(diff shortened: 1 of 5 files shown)


Weiko (approved):
LGTM
I think we need to take some time, probably just renaming, to clarify what's encrypted and what's not

packages/twenty-server/src/modules/connected-account/refresh-tokens-manager/services/connected-account-refresh-tokens.service.ts
```diff
@@ -116,35 +100,34 @@ export class ConnectedAccountRefreshTokensService {
...
-      accessToken: connectedAccountTokens.accessToken,
-      refreshToken: connectedAccountTokens.refreshToken,
+      accessToken: plaintextTokens.accessToken,
+      refreshToken: plaintextTokens.refreshToken,
       workspaceId,
     });
 
-    const authContext = buildSystemAuthContext(workspaceId);
-
-    await this.globalWorkspaceOrmManager.executeInWorkspaceContext(async () => {
```
Weiko, line 136: good catch!

PR #20820, opened 2026-05-21
fix(server): auto-index target<X>Id join columns on polymorphic standard objects

### LINKED ISSUE #20726: performance issues on timelineActivity fetching
Twenty was returning QueryFailedError: Query read timeout to users when opening records that fan out timeline-activity lookups for custom models.
Investigation showed slow PostgreSQL queries on timelineActivity filtering by custom relation columns such as targetPartnerId together with deletedAt IS NULL. Our timelineActivity table is about 21.9M rows / 7.1 GB, so these lookups were falling back to sequential scans and taking 20-40s.

Example query shape:
```
SELECT "timelineActivity"."targetPartnerId", "timelineActivity"."id", "timelineActivity"."name"
FROM "workspace_xxxxx"."timelineActivity"
WHERE "targetPartnerId" IN ($1)
  AND "timelineActivity"."deletedAt" IS NULL
LIMIT 60;
```

From current Twenty upstream code, built-in timelineActivity relations such as Company, Person, and Opportunity get explicit BTREE indexes, but the custom-object creation flow only adds relation target fields on standard objects like timelineActivity and creates the custom object’s default searchVector GIN index. It does not generate a BTREE index for the new timelineActivity.target<CustomModel>Id column. In our workspace this left custom timeline relation columns unindexed.
We ruled out infra pressure as the primary cause. There were no pod restarts, node memory usage was around 29%, and pod memory stayed well below limits. The bottleneck was PG-side query latency on timelineActivity.
As an immediate fix we added:
```
CREATE INDEX CONCURRENTLY "IDX_timelineActivity_targetPartnerId_deletedAt"
ON workspace_xxxxxx."timelineActivity"
USING btree ("targetPartnerId", "deletedAt");
```

After that, the same lookup dropped from 20-41s to about 4.1 ms, and the timeout errors stopped. EXPLAIN ANALYZE switched from sequential scan to index scan on IDX_timelineActivity_targetPartnerId_deletedAt.
We then applied the same composite index pattern to the remaining custom models, all created with CREATE INDEX CONCURRENTLY.

For some of the models we've added partial index on _Deal(position) WHERE deletedAt IS NULL.

Environment: PostgreSQL 17.9.1, aarch64,  compiled by gcc (Ubuntu 13.3.0-6ubuntu2~24.04.1) 13.3.0, 64-bit. Extensions enabled. btree_gin 1.3, pg_buffercache 1.5, pg_prewarm 1.2, pg_trgm 1.6, plpgsql, 1.0, unaccent 1.1, uuid-ossp.

twenty deployed to k8s, images used,  twentycrm/twenty:v2.4.0, redis:7.4-alpine, all arm.

FelixMalfait:
Closes #20726

## The bug

`timelineActivity` (and the three other polymorphic standard objects — `attachment`, `noteTarget`, `taskTarget`) store relations as N nullable `target<X>Id` columns, one per related object. Each one is a join key queried as `WHERE target<X>Id IN (...) AND deletedAt IS NULL`.

For **built-in** related objects (Person, Company, Opportunity, …), each `target<X>Id` column gets a BTREE index, declared statically in `compute-{timelineActivity,attachment,noteTarget,taskTarget}-standard-flat-index-metadata.util.ts`.

For **custom** related objects, the same `target<CustomObject>Id` column was added — **without an index**. On a `timelineActivity` table at issue-reporter scale (~21.9M rows, 7.1 GB), this turned record loads into 20–40s sequential scans and produced `QueryFailedError: Query read timeout` for end users.

## Diagnosis

The morph/relation field generator (`generateMorphOrRelationFlatFieldMetadataPair`) already creates a BTREE index for the field that owns the join column and returns it alongside the field metadata pair. The two user-driven entry points (`fromRelationCreateFieldInput…`, `fromMorphRelationCreateFieldInput…`) correctly destructure and propagate that index.

But the **custom-object creation path** — `buildDefaultRelationFlatFieldMetadatasForCustomObject`, called when a user creates a new custom object — destructured only `{ flatFieldMetadatas }` and threw away `indexMetadatas`. So every `target<CustomObject>Id` column added to the four polymorphic standard objects has been shipping unindexed since custom morph relations went in.

## The fix

Three commits.

### 1. `fix(server): index target<CustomObject>Id columns on standard polymorphic objects`

13 lines across 2 files.

- `build-default-relation-flat-field-metadatas-for-custom-object.util.ts` — also destructure `indexMetadatas` from the pair generator and accumulate them into the returned record (new field `standardTargetFlatIndexMetadatas`).
- `from-create-object-input-to-flat-object-metadata-and-flat-field-metadatas-to-create.util.ts` — append the accumulated indexes to `flatIndexMetadataToCreate`. The migration pipeline at `object-metadata.service.ts:559–562` already passes `flatIndexMetadataToCreate` to the migration runner, so no further wiring is needed.

From now on, creating a custom object also creates the four BTREE indexes — one per polymorphic standard object's new `target<CustomObject>Id` column — atomically with the rest of the migration.

### 2. `feat(server): backfill workspace command for relation join column indexes`

For existing workspaces whose custom objects were created before the forward-fix.

`upgrade:2-8:backfill-relation-join-column-indexes` is a `@RegisteredWorkspaceCommand('2.8.0', 1798100000000)` matching the pattern from `2-7-workspace-command-…-drop-connected-account-standard-object.command.ts`.

Per workspace:
1. Load `flatObjectMetadataMaps`, `flatFieldMetadataMaps`, `flatIndexMaps` from the workspace cache.
2. Resolve the four polymorphic standard object IDs by `nameSingular` against `DEFAULT_RELATIONS_OBJECTS_STANDARD_IDS`.
3. Collect every field ID that's already covered by any existing index.
4. Filter `flatFieldMetadataMaps` to MORPH_RELATION fields on those four objects whose `settings.relationType === MANY_TO_ONE` (i.e. owns a join column) and whose ID isn't in the indexed set.
5. Generate a BTREE `UniversalFlatIndexMetadata` for each via `generateIndexForFlatFieldMetadata` (same helper the forward-fix uses).
6. Create the indexes in the workspace schema with **CONCURRENTLY** (see commit 3).
7. Submit the metadata through `WorkspaceMigrationValidateBuildAndRunService` so it lands in `indexMetadata` and the cache — same pipeline as a normal metadata change. The pipeline's own `CREATE INDEX IF NOT EXISTS` no-ops because the index already exists.

Properties:
- **Idempotent.** Re-running is a no-op once indexes exist.
- **Scoped.** Only the four polymorphic standard objects, only their MANY_TO_ONE morph relation fields, only those with no covering index.
- **Same code path as the forward-fix.** The backfill produces exactly the indexes the forward-fix would have created at custom-object creation time.
- **`--dry-run` supported** via the base `ActiveOrSuspendedWorkspaceCommandRunner`.

### 3. `feat(server): create index CONCURRENTLY in relation join column backfill`

Adds an opt-in `concurrently` flag to `WorkspaceSchemaIndexManagerService.createIndex` (threaded through `createIndexInWorkspaceSchema`). When `true`, emits `CREATE INDEX CONCURRENTLY IF NOT EXISTS …`. Defaults to `false` — every existing caller keeps the current transactional `CREATE INDEX` behavior.

The backfill command opts in. It creates a QueryRunner **without** `startTransaction()`, issues the CONCURRENTLY indexes one-by-one (each waits for the previous to finish), then submits the metadata through the normal migration pipeline whose own `CREATE INDEX IF NOT EXISTS` is now a no-op.

Why not flip the default for the helper:
- `CREATE INDEX CONCURRENTLY` cannot run inside a transaction — Postgres errors out. The migration pipeline calls `createIndex` from inside a transactional schema migration.
- CONCURRENTLY doesn't roll back with the transaction. If the surrounding migration fails, the index remains and you end up with metadata/schema drift.
- Failed CONCURRENTLY builds leave an INVALID index behind that needs manual `DROP`.
- UNIQUE indexes have different failure semantics under CONCURRENTLY (deferred, not immediate).

So CONCURRENTLY is opt-in, used only where it's the right tool (post-hoc backfills on populated tables).

## Decisions / tradeoffs

- **Single-column BTREE vs partial `WHERE deletedAt IS NULL` vs composite.** Twenty's queries always include `deletedAt IS NULL`. A partial index would be slightly better than a plain BTREE (smaller, no wasted seeks on soft-deleted rows). This PR ships single-column to match the existing built-in target index pattern, which already covers >95% of the available speedup (the 20s→4ms drop the reporter saw comes from having any index — composite/partial is a second-order effect). Switching all relation indexes to partial is a separate, broader change.
- **CONCURRENTLY operator caveat.** If a CONCURRENTLY build is interrupted (kill, connection drop, OOM), Postgres leaves the index as INVALID. We deliberately don't probe `pg_index` for invalid leftovers on every create — catalog-table queries can be slow at multi-tenant scale and the failure mode is rare. Recovery is manual: `DROP INDEX <name>` and re-run the backfill.
- **Forward-fix is not gated** behind a feature flag. The change is metadata-pipeline-internal; before, custom-object creation silently produced a degraded state. After, it produces the correct state. No new public API, no behavioural change for end users besides the indexes existing.

## Risk

- Forward-fix: changes only the metadata produced during custom-object creation. New objects get four extra `FlatIndexMetadata` rows and four extra `CREATE INDEX` statements during their creation migration. Tables are empty at that point so the index builds in microseconds.
- Helper change: API-compatible, default behavior unchanged. The new `concurrently` parameter is optional.
- Backfill: read-only state probe → CONCURRENTLY index creation (no write blocking) → metadata insert via the normal migration pipeline. Idempotent. Reverting is `DROP INDEX`.

## Test plan

- [ ] Verify forward-fix: create a custom object, confirm four new BTREE indexes appear on `timelineActivity`, `attachment`, `noteTarget`, `taskTarget` for the new `target<CustomObject>Id` columns, and that `flatIndexMaps` has matching entries.
- [ ] Verify backfill on a workspace that had custom objects created before the fix: run `--dry-run` first, confirm the expected indexes are listed; then run for real, confirm the indexes appear in pg (and as `indisvalid = true` in `pg_index`) and in `flatIndexMaps`. Re-run; confirm no-op.
- [ ] Verify backfill on a clean workspace: should log "no missing indexes" and exit.
- [ ] Verify CONCURRENTLY behavior under load: run backfill against a workspace with active writes on `timelineActivity`; confirm inserts/updates keep working during index build (no `ShareLock` waits in `pg_stat_activity`).
- [ ] On the affected reporter-scale workspace, confirm `EXPLAIN ANALYZE` switches from sequential scan to index scan and timeline activity timeouts go away.


packages/twenty-server/src/database/commands/upgrade-version-command/2-8/2-8-workspace-command-1798100000000-backfill-relation-join-column-indexes.command.ts
```diff
@@ -0,0 +1,225 @@
...
+          schemaName,
+          tableName,
+          index: {
+            name: universalFlatIndexMetadata.name,
+            columns: [joinColumnName],
+            isUnique: universalFlatIndexMetadata.isUnique,
+            type: universalFlatIndexMetadata.indexType,
+            where: universalFlatIndexMetadata.indexWhereClause ?? undefined,
+          },
+          concurrently: true,
```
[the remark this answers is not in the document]
FelixMalfait: Good catch. Fixed in 2e9171279b: when concurrently=true, WorkspaceSchemaIndexManagerService.createIndex now queries pg_index for an INVALID entry with the target name and DROP INDEX CONCURRENTLY IF EXISTS it before re-creating. Backfill is self-healing across interruptions; operators no longer need to manually clean up.

packages/twenty-server/src/engine/twenty-orm/workspace-schema-manager/services/workspace-schema-index-manager.service.ts
```diff
@@ -19,12 +19,22 @@ export class WorkspaceSchemaIndexManagerService {
     schemaName,
     tableName,
     index,
+    concurrently = false,
   }: {
     queryRunner: QueryRunner;
```
Weiko, line 28: https://www.postgresql.org/docs/current/sql-createindex.html
```
 Another difference is that a regular CREATE INDEX command can be performed within a transaction block, but CREATE INDEX CONCURRENTLY cannot.
```
We should check if the queryRunner has isTransactionActive true before doing concurrently logic and skip if it's the case but then you kind of lie on the contract if you set concurrently true. (and expect to be concurrently), maybe it should throw or have a dedicated method that creates its own query runner 🤔

packages/twenty-server/src/database/commands/upgrade-version-command/2-8/2-8-workspace-command-1798100000000-backfill-relation-join-column-indexes.command.ts
```diff
@@ -0,0 +1,225 @@
...
+              flatEntityToUpdate: [],
+            },
+          },
+          workspaceId,
+          applicationUniversalIdentifier:
+            twentyStandardFlatApplication.universalIdentifier,
+        },
+      );
+
+    if (validateAndBuildResult.status === 'fail') {
```
Weiko, line 211: If this fails and validateBuildAndRunWorkspaceMigration rolled back, you'll have N physical indexes in DB. Ideally those 2 entities should be created within the same transaction and index action handler of the migration runner should both create the index metadata and the physical index (but again we also know that concurrently does not work with transactions...)

Weiko: I guess it recovers fine with the "IF NOT EXISTS" part but we still might have unbound indexes 🤔

packages/twenty-server/src/database/commands/upgrade-version-command/2-8/2-8-workspace-command-1798100000000-backfill-relation-join-column-indexes.command.ts
```diff
@@ -0,0 +1,225 @@
...
+import { generateIndexForFlatFieldMetadata } from 'src/engine/metadata-modules/flat-field-metadata/utils/generate-index-for-flat-field-metadata.util';
+import { isMorphOrRelationFlatFieldMetadata } from 'src/engine/metadata-modules/flat-field-metadata/utils/is-morph-or-relation-flat-field-metadata.util';
+import { type FlatObjectMetadata } from 'src/engine/metadata-modules/flat-object-metadata/types/flat-object-metadata.type';
+import { WorkspaceSchemaManagerService } from 'src/engine/twenty-orm/workspace-schema-manager/workspace-schema-manager.service';
+import { WorkspaceCacheService } from 'src/engine/workspace-cache/services/workspace-cache.service';
+import { getWorkspaceSchemaContextForMigration } from 'src/engine/workspace-manager/workspace-migration/workspace-migration-runner/utils/get-workspace-schema-context-for-migration.util';
+import { WorkspaceMigrationValidateBuildAndRunService } from 'src/engine/workspace-manager/workspace-migration/services/workspace-migration-validate-build-and-run-service';
+
+const POLYMORPHIC_STANDARD_OBJECT_NAMES_SINGULAR: ReadonlySet<string> = new Set(
```
Weiko, line 21: Looks like we set IDS within a "name singular" set, is it only a naming issue or is the filter wrong?

packages/twenty-server/test/integration/metadata/suites/object-metadata/__snapshots__/failing-create-one-object-metadata-v2.integration-spec.ts.snap
```diff
@@ -2940,12 +3004,12 @@ exports[`Object metadata creation should fail v2 when namePlural has invalid cha
         },
       ],
     },
-    "message": "Validation failed for 17 fieldMetadatas, 1 objectMetadata, 1 view, 5 viewFields, 1 index",
+    "message": "Validation failed for 17 fieldMetadatas, 1 objectMetadata, 1 view, 5 viewFields, 5 indexs",
```
Weiko: nit: plural is "indexs" here for some reason

packages/twenty-server/src/engine/twenty-orm/workspace-schema-manager/services/workspace-schema-index-manager.service.ts
FelixMalfait: Good call. Fixed in 2a025fe599: createIndex now throws when concurrently=true but queryRunner.isTransactionActive. The contract is now "opt-in to concurrent, but only if you control the transaction context." The backfill command already creates its own QueryRunner without startTransaction, so it's good.

packages/twenty-server/src/database/commands/upgrade-version-command/2-8/2-8-workspace-command-1798100000000-backfill-relation-join-column-indexes.command.ts
FelixMalfait: Real concern but recoverable: if validateBuildAndRunWorkspaceMigration's metadata transaction rolls back after CONCURRENTLY succeeded, the physical indexes remain orphans. On re-run the backfill filter "!indexedFieldIds.has(field.id)" reads from flatIndexMaps (the metadata cache), so the orphan-bound fields are re-targeted, CONCURRENTLY no-ops via IF NOT EXISTS, and the metadata insert retries. Net: the unbound state only persists until the next backfill run. Not adding eager cleanup-on-failure for now — happy to add it if the recovery window is too long for your taste.

FelixMalfait: Right — re-runs re-bind because the filter is driven by metadata (flatIndexMaps), not pg state. The unbound window is bounded by "until next backfill run". Acceptable for a one-shot upgrade command IMO; let me know if you want explicit cleanup on metadata-failure.

packages/twenty-server/src/database/commands/upgrade-version-command/2-8/2-8-workspace-command-1798100000000-backfill-relation-join-column-indexes.command.ts
FelixMalfait: Naming issue, not a filter bug. DEFAULT_RELATIONS_OBJECTS_STANDARD_IDS in twenty-shared is misnamed — its values are nameSingular strings (`['timelineActivity', 'attachment', 'noteTarget', 'taskTarget']`), not IDs. My local Set is correctly typed as nameSingular values and the filter against flatObject.nameSingular matches them. Could rename the upstream constant as a separate cleanup; out of scope here.

packages/twenty-server/test/integration/metadata/suites/object-metadata/__snapshots__/failing-create-one-object-metadata-v2.integration-spec.ts.snap
FelixMalfait: Fixed in 2a025fe599: the validation-summary message handler now pluralizes "index" as "indexes" via an irregular-plurals map. Both affected snapshot files (this one + failing-create-one-field-metadata-morph-relation-v2) updated.

packages/twenty-server/src/engine/twenty-orm/workspace-schema-manager/services/workspace-schema-index-manager.service.ts
FelixMalfait: I guess it will fail no if you use CONCURRENTLY within a transaction // will be a code issue and throw a Sentry error? Not sure it deserves a try/catch because it's a a code error more than something expected to catch?

packages/twenty-server/src/engine/workspace-manager/workspace-migration/interceptors/utils/workspace-migration-builder-graphql-api-exception-handler.util.ts
```diff
@@ -1,6 +1,8 @@
 import { ALL_METADATA_NAME } from 'twenty-shared/metadata';
 import { isDefined } from 'twenty-shared/utils';
 
+import { plural } from 'pluralize';
```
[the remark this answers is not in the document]
FelixMalfait: Intentional. The maintainer asked to switch to the pluralize library specifically to avoid maintaining a local irregulars map; adding 'index → indexes' back as an override would defeat that. Both 'indexes' and 'indices' are correct English; pluralize gives the Latinate form, which is fine for an internal validation-summary string. Postgres uses both forms in its docs/catalogs (pg_indexes view, pg_index table), so there's no clear DB-correctness argument either way.

packages/twenty-server/src/engine/twenty-orm/workspace-schema-manager/services/workspace-schema-index-manager.service.ts
FelixMalfait: OK AI came before I did and agreed, let's keep its change :D

packages/twenty-server/src/engine/twenty-orm/workspace-schema-manager/services/workspace-schema-index-manager.service.ts
```diff
@@ -19,12 +19,28 @@ export class WorkspaceSchemaIndexManagerService {
...
     schemaName: string;
     tableName: string;
     index: WorkspaceSchemaIndexDefinition;
+    concurrently?: boolean;
   }): Promise<void> {
+    if (concurrently) {
+      if (queryRunner.isTransactionActive) {
+        throw new Error(
```
charlesBochet: let's throw custom exception and use lingue

FelixMalfait: See my comment here: https://github.com/twentyhq/twenty/pull/20820#discussion_r3287043117 ; for me this is a code error - not a custom exception that needs to be translated

packages/twenty-server/test/integration/metadata/suites/object-metadata/__snapshots__/failing-create-one-object-metadata-v2.integration-spec.ts.snap
FelixMalfait: (pre-existing)

packages/twenty-server/src/database/commands/upgrade-version-command/2-8/2-8-workspace-command-1798100000000-backfill-relation-join-column-indexes.command.ts
FelixMalfait: (pre-existing pattern)

packages/twenty-server/src/database/commands/upgrade-version-command/2-8/2-8-workspace-command-1798100000000-backfill-relation-join-column-indexes.command.ts
FelixMalfait: Agree with AI. For me it's a trade-off of using Concurrently and fine for a command. I think it's better than the first version I did which did not use concurrently but could freeze

packages/twenty-server/src/engine/twenty-orm/workspace-schema-manager/services/workspace-schema-index-manager.service.ts
FelixMalfait: Done in c24bcc32c6. Added a CONCURRENT_INDEX_CREATION_IN_TRANSACTION code to WorkspaceSchemaManagerExceptionCode and now throw WorkspaceSchemaManagerException — userFriendlyMessage falls through to STANDARD_ERROR_MESSAGE (this is a developer-contract violation, not user-facing copy).

packages/twenty-server/src/engine/twenty-orm/workspace-schema-manager/services/workspace-schema-index-manager.service.ts
```diff
@@ -62,6 +80,34 @@ export class WorkspaceSchemaIndexManagerService {
...
+  private async dropInvalidIndexIfExists({
+    queryRunner,
+    schemaName,
+    indexName,
+  }: {
+    queryRunner: QueryRunner;
+    schemaName: string;
+    indexName: string;
+  }): Promise<void> {
+    const invalidIndex = await queryRunner.query(
```
charlesBochet: this won't be pretty on prod I believe, this will likely be slow. 
1) I might be wrong, so let's run this query on prod reader to check the performance
2) can we avoid it? do we need this check? usually we just assume that our metadata is up-to-date with our postgres state.

FelixMalfait: Removed in 6ad832cf3c. You're right — the probe was defending against a rare failure mode (interrupted CONCURRENTLY → INVALID index) with a known manual recovery path (DROP + re-run). Catalog-table queries at multi-tenant scale aren't worth that cost. Kept only the transaction-active guard. Updated the PR description's CONCURRENTLY caveat bullet to match.

packages/twenty-server/src/engine/twenty-orm/workspace-schema-manager/services/workspace-schema-index-manager.service.ts
charlesBochet: yes, but our consensus is that we should always throw Custom Exception as this eases the handling later + we should add userFriendlyMessage as low as possible to avoid overhead in upper layer (very complex to put them later as the caller needs to udnerstand what went wrong in the underlying service)

FelixMalfait: Done in 3b4e1a1be5. The CONCURRENT_INDEX_CREATION_IN_TRANSACTION code now maps to a specific MessageDescriptor (`msg`Could not create the index because it must run outside a database transaction.``) in the schema-manager exception's switch, instead of falling through to STANDARD_ERROR_MESSAGE.

packages/twenty-server/src/engine/twenty-orm/workspace-schema-manager/services/workspace-schema-index-manager.service.ts
FelixMalfait: OK

packages/twenty-server/src/engine/twenty-orm/workspace-schema-manager/services/workspace-schema-index-manager.service.ts
FelixMalfait: Imo this shouldn't have been added at all, it's not our job to convert code errors into user-exceptions since code errors should not happen. But since it's added  and you seem to agree ok I'll update it!

FelixMalfait: I mean it does use isTransactionActive so it's a bit different scope / probably broader than the native error that would have been thrown, so I agree

PR #20821, opened 2026-05-21
Add empty operands to UUID filter type in workflow filter action

### LINKED ISSUE #20808: Search records node doesn't have all filters
In Search records node there are less filters than in Filter node, the goal is to have the same filters in Search records and in Filter nodes.

https://github.com/user-attachments/assets/7f7069a8-9c94-4b7a-b607-31ba3216dbb5

thomtrp:
Fixes https://github.com/twentyhq/twenty/issues/20808

## Summary
- Adds `IS_EMPTY` and `IS_NOT_EMPTY` operands to the UUID entry in `getStepFilterOperands`, aligning the workflow filter action with the find records (search) action which already includes these operands for ID-type fields.

## Test plan
- [ ] Open a workflow with a filter action, select an ID-type field, and verify the operand dropdown now includes "Is empty" and "Is not empty"
- [ ] Open a workflow with a find records action, select an ID-type field, and verify the operand dropdown is consistent with the filter action


Weiko (approved):
LGTM!
As mentioned, I believe we can refactor this, it seems it shares a lot of logic with our common filter/operand parsing

charlesBochet:
I picked this up because the Sentry/Cubic regression flags worried me and the util had no unit tests. I added snapshot tests for `getStepFilterOperands` and `getRecordFilterOperands` (55 cases total) on top of `main` first to lock in the baseline, then carried them over and ran them against your branch. The PR's refactor introduced 5 regressions in step filter (only 2 of which Sentry caught).

I pushed a follow-up commit (`f0c750c`) that keeps your "share the map" refactor — which I agree with — but actually aligns the shared map so both consumers stay correct.

## What the snapshot tests proved

49 of 55 snapshots are byte-identical to `main` baseline. The 6 intentional diffs:

### Step filter (workflow filter action) — vs `main`

| Field type | Before (main) | After | Notes |
|---|---|---|---|
| `UUID` | `[IS, IS_NOT]` | `[IS, IS_NOT, IS_EMPTY, IS_NOT_EMPTY]` | ✅ the PR's stated goal |
| `NUMERIC` | `[IS, IS_NOT, GTE, LTE, IS_EMPTY, IS_NOT_EMPTY]` | unchanged (now via `NUMBER` alias) | fixes the Sentry regression where it fell through to default and gained nonsensical `CONTAINS`/`DOES_NOT_CONTAIN` |
| `RATING` | `[IS, IS_NOT, IS_EMPTY, IS_NOT_EMPTY]` | `[IS, IS_NOT, GTE, LTE, IS_EMPTY, IS_NOT_EMPTY]` | ⚠️ gains `GTE`/`LTE` — ratings are numeric (1-5) so these are useful, but flagging as a UX change |
| `DATE_TIME` / `DATE` | `IS_RELATIVE` at position 7 | `IS_RELATIVE` at position 2 | ⚠️ dropdown reorder only, same operands — now matches record-filter UX |

### Record filter (object record filtering) — vs `main`

| Field type | Before (main) | After | Notes |
|---|---|---|---|
| `UUID` | `[IS]` | `[IS, IS_NOT, IS_EMPTY, IS_NOT_EMPTY]` | ⚠️ gains 3 operands for ID-type filters. The PR description mentioned the find-records action "already includes these for ID-type fields" — that wasn't actually true; with this commit it now is |
| `RATING` | `[IS, GTE, LTE, IS_EMPTY, IS_NOT_EMPTY]` | `[IS, IS_NOT, GTE, LTE, IS_EMPTY, IS_NOT_EMPTY]` | ⚠️ gains `IS_NOT` |

## Source diff

```diff
--- a/packages/twenty-front/src/modules/object-record/record-filter/utils/getRecordFilterOperands.ts
+++ b/packages/twenty-front/src/modules/object-record/record-filter/utils/getRecordFilterOperands.ts
@@ -112,6 +112,7 @@ export const FILTER_OPERANDS_MAP = {
   ],
   RATING: [
     RecordFilterOperand.IS,
+    RecordFilterOperand.IS_NOT,
     RecordFilterOperand.GREATER_THAN_OR_EQUAL,
     RecordFilterOperand.LESS_THAN_OR_EQUAL,
     ...emptyOperands,
@@ -139,7 +140,11 @@ export const FILTER_OPERANDS_MAP = {
   ],
   BOOLEAN: [RecordFilterOperand.IS],
   TS_VECTOR: [RecordFilterOperand.VECTOR_SEARCH],
-  UUID: [RecordFilterOperand.IS],
+  UUID: [
+    RecordFilterOperand.IS,
+    RecordFilterOperand.IS_NOT,
+    ...emptyOperands,
+  ],
 } as const satisfies FilterOperandMap;
```

```diff
--- a/packages/twenty-front/src/modules/workflow/workflow-steps/filters/utils/getStepFilterOperands.ts
+++ b/packages/twenty-front/src/modules/workflow/workflow-steps/filters/utils/getStepFilterOperands.ts
@@ -62,6 +62,8 @@ export const getStepFilterOperands = ({
     case 'UUID':
       return FILTER_OPERANDS_MAP.UUID;
+    case 'NUMERIC':
+      return FILTER_OPERANDS_MAP.NUMBER;
     case 'ACTOR': {
```

(`NUMERIC` can't live in the shared map because it's not in `FILTERABLE_FIELD_TYPES` — but its operands are identical to `NUMBER`, so aliasing is safe.)

## Decisions worth a sanity check

The ⚠️ rows above are intentional behavioral changes — let me know if any aren't desirable and I'll revert them with a local override (the snapshot tests will catch any further drift).

Anyway, feel free to revert my commit if you'd rather scope this PR more narrowly — happy either way. Thanks for the cleanup!

charlesBochet (approved):

charlesBochet:
"share the map" commit exposed dropdown options whose runtime support was incomplete in **both** the record-filter compiler and the workflow filter-action evaluator. Pushed `2e38e40` to close every gap end-to-end so all operands in the shared map actually work.

## What was broken (pre-fix audit)

Running the new operands through the full pipeline:

| Combo | Where it broke |
|---|---|
| **Record filter UUID + IS_NOT** | `turnRecordFilterIntoGqlOperationFilter.ts` (UUID switch — `default` throws) |
| **Record filter UUID + IS_EMPTY / IS_NOT_EMPTY** | `getEmptyRecordGqlOperationFilter.ts` (no UUID case — throws `UNSUPPORTED_EMPTY_FILTER_TYPE`) |
| **Record filter RATING + IS_NOT** | `turnRecordFilterIntoGqlOperationFilter.ts` (RATING switch — `default` throws) |
| **Step filter UUID + IS_EMPTY / IS_NOT_EMPTY** *(the PR's stated goal)* | `evaluate-filter-conditions.util.ts` → `evaluateUuidFilter` (`default` throws) |
| **Step filter RATING + GTE / LTE** | fell through to `evaluateDefaultFilter` which does `Number('RATING_3')` → `NaN` → comparisons always false |

Plus: `buildValueFromFilter` (the create-record-from-filter helper) would have failed type-check because `RecordFilterToRecordInputOperand<T>` is derived from `FILTER_OPERANDS_MAP[T]`.

## What this commit changes

**Frontend (`turnRecordFilterIntoGqlOperationFilter`)** — add cases mirroring the SELECT pattern:

```diff
 case 'RATING':
       switch (recordFilter.operand) {
         case RecordFilterOperand.IS: ...
+        case RecordFilterOperand.IS_NOT:
+          return {
+            not: {
+              [fieldMetadataItem.name]: {
+                eq: convertRatingToRatingValue(parseFloat(recordFilter.value)),
+              } as RatingFilter,
+            },
+          };

 case 'UUID': {
       switch (recordFilter.operand) {
         case RecordFilterOperand.IS: ...
+        case RecordFilterOperand.IS_NOT:
+          return {
+            not: {
+              [fieldMetadataItem.name]: { in: recordIds } as UUIDFilter,
+            },
+          };
```

**Frontend (`getEmptyRecordGqlOperationFilter`)** — add UUID branch alongside SELECT / NUMBER:

```diff
+    case 'UUID':
+      emptyRecordFilter = {
+        [correspondingField.name]: { is: 'NULL' } as UUIDFilter,
+      };
+      break;
```

**Frontend (`buildValueFromFilter`)** — extend UUID/RATING handlers; new operands return `undefined` since "is not / is empty / is not empty" don't seed a concrete value for create-record-from-filter.

**Backend (`evaluate-filter-conditions.util.ts`)** — extend UUID, add RATING:

```diff
 function evaluateUuidFilter(filter) {
   switch (filter.operand) {
     case ViewFilterOperand.IS: ...
     case ViewFilterOperand.IS_NOT: ...
+    case ViewFilterOperand.IS_EMPTY:
+      return !isNonEmptyString(filter.leftOperand);
+    case ViewFilterOperand.IS_NOT_EMPTY:
+      return isNonEmptyString(filter.leftOperand);
   }
 }

+// Rating values are stored as 'RATING_1'..'RATING_5'. Comparisons must
+// operate on the numeric rank (1..5), not Number() on the enum string
+// (which would be NaN), nor lexicographic order.
+function parseRatingRank(value): number | undefined { ... }
+
+function evaluateRatingFilter(filter) {
+  // IS / IS_NOT / GTE / LTE / IS_EMPTY / IS_NOT_EMPTY using parseRatingRank
+}
```

## Tests added

- `turnRecordFilterIntoGqlOperationFilter.test.ts`: UUID `IS_NOT` and RATING `IS_NOT` cases.
- `getEmptyRecordGqlOperationFilter.test.ts`: UUID `IS_EMPTY` case.
- `evaluate-filter-conditions.util.spec.ts`: UUID `IS_EMPTY`/`IS_NOT_EMPTY` and a full rating describe block (IS, IS_NOT, GTE, LTE, IS_EMPTY, IS_NOT_EMPTY, unparseable string, unsupported operand → throw).

## Verification

| Suite | Result |
|---|---|
| `twenty-shared` (all 1231 unit tests) | ✅ |
| `twenty-server` — `evaluate-filter-conditions.util.spec.ts` (80 tests) | ✅ |
| `twenty-front` — filter modules (90 tests incl. 55 snapshots) | ✅ |
| `twenty-front` — record-table (64 tests) | ✅ |
| `npx nx typecheck twenty-shared/twenty-front/twenty-server` | ✅ |
| `npx nx lint:diff-with-main` (all 3) | ✅ |

## Out of scope (deliberate)

- **Exposing RATING in the workflow step filter field picker.** `WorkflowStepFilterFieldSelect.NON_SELECTABLE_FIELD_TYPES` still excludes RATING — so although the backend now evaluates RATING step filters correctly, users still can't pick a rating field via the UI. That's a separate product decision (and would require wiring a rating star input into `WorkflowStepFilterValueInput` / `FormFieldInput`). Happy to do that in a follow-up PR if you want it; left as-is here to keep this PR scoped.

Let me know if you'd rather I revert the two commits and ship something smaller — e.g., just the original PR's intent (UUID + emptiness operands for step filter) using a local override. But given how shallow the runtime gaps turned out to be, fixing them properly seemed like the right move.

PR #20827, opened 2026-05-21
make connectedAccount clients self contained

neo773:
Replace OAuth2ClientManagerService with per provider each loading their own entity and resolving tokens internally

Removes the ugly spread pattern of sprinkling tokens everywhere, this caused downtime of messaging when we migrated to encrypted tokens


packages/twenty-server/src/modules/connected-account/oauth2-client-manager/drivers/microsoft/microsoft-oauth2-client.provider.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+    provider = module.get<MicrosoftOAuth2ClientProvider>(
+      MicrosoftOAuth2ClientProvider,
+    );
+  });
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe('getClient', () => {
```
charlesBochet, line 100: I'm not sure about the value of such test TBH
We should rather invest on:
- typescript devXP (discussed with Paul)
- integration test on messaging (either by mocking with msw or by mocking sdks layer)

charlesBochet (approved):
LGTM, left comments but that can be addressed later

charlesBochet:
Holding on merge as only the author @neo773 knows if this has been well tested (we don't have integration test on this area)

PR #20834, opened 2026-05-22
[Website] force-static releases page & move workspace bullet to pro self-host

mabdullahabaid:
Add `export const dynamic = 'force-static'` to the releases page to prevent runtime re-renders on Cloudflare Workers where fs is unavailable, which caused stale "Releases were not found" errors after ISR cache eviction. 

Also moves the "Up to 5 workspaces" bullet from the Organization plan to the Pro plan on the self-hosting pricing view.


prastoin (approved):
mmmhmm

charlesBochet (approved):

PR #20835, opened 2026-05-22
Download tarball instead of yarn install when installing application

martmull:
as title

Tested on Vexa public application install
- before -> 5.4s
- after -> 2.55s

Tested on Exa public application install
- before -> 5.5s
- after -> 2.24s

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/application/application-package/application-package-fetcher.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/application/application-package/application-package-fetcher.service.ts
@@ -13,30 +13,20 @@ import { type PackageJson } from 'type-fest';
 import { Repository } from 'typeorm';
 import { v4 } from 'uuid';
 
-import {
-  ApplicationRegistrationEntity
-} from 'src/engine/core-modules/application/application-registration/application-registration.entity';
-import {
-  ApplicationRegistrationSourceType
-} from 'src/engine/core-modules/application/application-registration/enums/application-registration-source-type.enum';
+import { ApplicationRegistrationEntity } from 'src/engine/core-modules/application/application-registration/application-registration.entity';
+import { ApplicationRegistrationSourceType } from 'src/engine/core-modules/application/application-registration/enums/application-registration-source-type.enum';
 import {
   ApplicationException,
   ApplicationExceptionCode,
 } from 'src/engine/core-modules/application/application.exception';
 import { ApplicationEntity } from 'src/engine/core-modules/application/application.entity';
-import {
-  assertValidNpmPackageName
-} from 'src/engine/core-modules/application/application-package/utils/assert-valid-npm-package-name.util';
-import {
-  extractTarballSecurely
-} from 'src/engine/core-modules/application/application-package/utils/extract-tarball-securely.util';
+import { assertValidNpmPackageName } from 'src/engine/core-modules/application/application-package/utils/assert-valid-npm-package-name.util';
+import { extractTarballSecurely } from 'src/engine/core-modules/application/application-package/utils/extract-tarball-securely.util';
 import { readJsonFileOrThrow } from 'src/engine/core-modules/application/application-package/utils/read-json-file.util';
 import { resolvePackageContentDir } from 'src/engine/core-modules/application/application-package/utils/tarball-utils';
 import { FileStorageService } from 'src/engine/core-modules/file-storage/file-storage.service';
 import { FileEntity } from 'src/engine/core-modules/file/entities/file.entity';
-import {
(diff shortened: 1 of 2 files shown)


packages/twenty-server/src/engine/core-modules/application/application-package/application-package-fetcher.service.ts
```diff
@@ -224,107 +231,76 @@ export class ApplicationPackageFetcherService implements OnModuleInit {
...
-      lines.push(`//${registryHost}/:_authToken=${config.authToken}`);
+    if (authToken) {
+      headers['Authorization'] = `Bearer ${authToken}`;
     }
 
-    if (lines.length > 0) {
-      await fs.writeFile(
-        join(config.workDir, '.npmrc'),
-        lines.join('\n') + '\n',
+    const response = await fetch(metadataUrl, {
```
prastoin: ***Suggestion:*** Please use standard fetcher

packages/twenty-server/src/engine/core-modules/application/application-package/application-package-fetcher.service.ts
```diff
@@ -224,107 +231,76 @@ export class ApplicationPackageFetcherService implements OnModuleInit {
...
-  ): Promise<void> {
-    const packageJson = {
-      name: 'twenty-app-resolver-workspace',
-      private: true,
-      dependencies: {
-        [packageName]: versionSpec,
-      },
+    const metadata = (await response.json()) as {
+      dist?: { tarball?: string };
     };
```
prastoin: ***Suggestion:*** validate using zod

prastoin (commented):
I'm sorry but I'm not sure to understand
We're now always publishing the external dependencies (`node_modules`) in the built tarball ?

If that's the case it's not a good practice as they could be spoofed
It also means publishing in the related registry your literal packages dependencies too which is not common

prastoin (commented):
Ah nevermind I got it
the previous yarn install was a hacky way to do a GET of the tarball on the registry
I prefer this implem too
But what was the initial motivation if it's not cleanliness ?

martmull:
> I prefer this implem too
> But what was the initial motivation if it's not cleanliness ?

it was too slow

packages/twenty-server/src/engine/core-modules/application/application-package/application-package-fetcher.service.ts
```diff
@@ -224,107 +231,76 @@ export class ApplicationPackageFetcherService implements OnModuleInit {
...
-        'stderr' in error &&
-        typeof (error as { stderr: unknown }).stderr === 'string'
-          ? (error as { stderr: string }).stderr
-          : undefined;
-
-      const message =
-        stderr ?? (error instanceof Error ? error.message : String(error));
+    if (authToken && isSameHost) {
+      headers['Authorization'] = `Bearer ${authToken}`;
+    }
```
prastoin, line 291: this seems quite fragile

prastoin: lets include the protocol too comparing origins

packages/twenty-server/src/engine/core-modules/application/application-package/application-package-fetcher.service.ts
```diff
@@ -224,105 +244,83 @@ export class ApplicationPackageFetcherService implements OnModuleInit {
...
+            },
+          }
+        : {}),
+    });
 
-  private async resolveLocalYarnPath(workDir: string): Promise<string> {
-    const yarnrcPath = join(workDir, '.yarnrc.yml');
-    const yarnrcContent = await fs.readFile(yarnrcPath, 'utf-8');
-    const match = yarnrcContent.match(/^yarnPath:\s*(.+)$/m);
+    let response;
```
prastoin, line 255: redundant let could handle response happy patch in try statement

packages/twenty-server/src/engine/core-modules/application/application-package/application-package-fetcher.service.ts
```diff
@@ -224,105 +244,83 @@ export class ApplicationPackageFetcherService implements OnModuleInit {
...
     };
 
-    await fs.writeFile(
-      join(workDir, 'package.json'),
-      JSON.stringify(packageJson, null, 2),
-    );
-  }
+    const httpClient = this.secureHttpClientService.getHttpClient({
+      timeout: RESOLUTION_TIMEOUT_MS,
+      ...(authToken
```
prastoin: ***Nitpick:*** isDefined

prastoin (approved):

PR #20836, opened 2026-05-22
Deprecate `messageChannel` `messageFolder` `calendarChannel` standard objects

prastoin:
# Introduction

Removing old standard objects `messageChannel` and `messageFolder` and `calendarChannel`


Weiko (approved):
LGTM

PR #20845, opened 2026-05-22
Fix permission flag removed flag property from diffing

### LINKED ISSUE #20843: App install fails updating `rolePermissionFlag`: manifest uses `flag` but server expects different property
Running on v2.7.3 with twenty-sdk@2.7

**Description**

Installing/syncing a Twenty app fails when the generated manifest includes role permission flags. The migration attempts to update `rolePermissionFlag` with a `flag` property, but the server metadata layer rejects it because `RolePermissionFlagEntity` does not contain a `flag` property.

**Error**

```json
{
  "message": "Migration action 'update' for 'rolePermissionFlag' failed",
  "extensions": {
    "code": "APPLICATION_INSTALLATION_FAILED",
    "action": {
      "type": "update",
      "metadataName": "rolePermissionFlag",
      "universalIdentifier": "98ece046-59ad-5f4c-a593-83adbe47f863",
      "update": {
        "flag": "UPLOAD_FILE"
      }
    },
    "errors": {
      "metadata": {
        "message": "Property \"flag\" was not found in \"RolePermissionFlagEntity\". Make sure your query is correct.",
        "code": "INTERNAL_SERVER_ERROR"
      }
    },
    "userFriendlyMessage": "Migration execution failed."
  }
}
```

**Expected Behavior**

The app should install successfully when role `permissionFlags` are declared through the SDK, for example:

```ts
permissionFlags: [
  PermissionFlag.UPLOAD_FILE,
  PermissionFlag.SEND_EMAIL_TOOL,
  PermissionFlag.HTTP_REQUEST_TOOL,
]
```

**Actual Behavior**

The generated migration tries to update `rolePermissionFlag` using:

```json
{
  "flag": "UPLOAD_FILE"
}
```

The server rejects this with:

```text
Property "flag" was not found in "RolePermissionFlagEntity"
```

**Likely Cause**

There appears to be a schema mismatch between the SDK-generated manifest/migration payload and the server-side `RolePermissionFlagEntity`. The SDK emits `flag`, while the server entity appears to expect another property name, likely `permissionFlag`.

**Workaround**

Post-process the generated manifest before install and rewrite role permission flag entries from:

```json
{ "flag": "UPLOAD_FILE" }
```

to:

```json
{ "permissionFlag": "UPLOAD_FILE" }
```

**Impact**

Apps that declare role permission flags cannot be installed or synced without manually modifying the generated manifest.

Weiko:
## Why it fixes the bug

RolePermissionFlagEntity.flag (role-permission-flag.entity.ts:54) is marked @WasRemovedInUpgrade for 2.7.0_FinalizeRolePermissionFlagCutoverFastInstanceCommand. 
After 2.7.0, the column is gone from real DBs and the metadata layer no longer accepts writes to it — but the diffing config still listed flag with toCompare: true. So when an SDK-generated manifest carried a flag value, computeUniversalFlatEntityPropertiesToCompareAndStringify (all-universal-flat-entity-properties-to-compare-and-stringify.constant.ts:55-69) included it in the comparison, the diff emitted { update: { flag: "UPLOAD_FILE" } }, and the metadata update failed with Property "flag" was not found in "RolePermissionFlagEntity".

Switching toCompare: false makes the diff skip flag; the only properties compared are now permissionFlagUniversalIdentifier and roleUniversalIdentifier, which is what the post-cutover
entity actually supports.

Fixes https://github.com/twentyhq/twenty/issues/20843

### DIFF
--- a/packages/twenty-server/src/engine/metadata-modules/flat-entity/constant/__tests__/__snapshots__/all-universal-flat-entity-properties-to-compare-and-stringify.constant.spec.ts.snap
+++ b/packages/twenty-server/src/engine/metadata-modules/flat-entity/constant/__tests__/__snapshots__/all-universal-flat-entity-properties-to-compare-and-stringify.constant.spec.ts.snap
@@ -258,7 +258,6 @@ exports[`ALL_UNIVERSAL_FLAT_ENTITY_PROPERTIES_TO_COMPARE_AND_STRINGIFY should ma
     "propertiesToCompare": [
       "permissionFlagUniversalIdentifier",
       "roleUniversalIdentifier",
-      "flag",
     ],
     "propertiesToStringify": [],
   },
--- a/packages/twenty-server/src/engine/metadata-modules/flat-entity/constant/all-entity-properties-configuration-by-metadata-name.constant.ts
+++ b/packages/twenty-server/src/engine/metadata-modules/flat-entity/constant/all-entity-properties-configuration-by-metadata-name.constant.ts
@@ -1198,7 +1198,7 @@ export const ALL_ENTITY_PROPERTIES_CONFIGURATION_BY_METADATA_NAME = {
       universalProperty: 'roleUniversalIdentifier',
     },
     flag: {
-      toCompare: true,
+      toCompare: false,
       toStringify: false,
       universalProperty: undefined,
     },


prastoin (approved):

prastoin (commented):
We might wanna set the flag as never in the entity and adapt the dynamic typing to get rid of them

Weiko:
Will do in the upcoming PR and try to simplify this part yes I agree!

PR #20847, opened 2026-05-22
[Breaking change] Prepare non-system permission flags

Weiko:
# Summary

Replaces the enum-keyed `permissionFlags: PermissionFlag[]` on roles with `permissionFlagUniversalIdentifiers: string[]`

This unlocks mixing system flags (`SystemPermissionFlag.*`) with app-defined flags in a role config.

This is a breaking change. Existing app source must switch to the new field.

# Breaking changes

- `RoleManifest.permissionFlags` removed. Use `RoleManifest.permissionFlagUniversalIdentifiers: string[]`.
- `RoleConfig.permissionFlags` removed (was `PermissionFlagType[]`). Use `RoleConfig.permissionFlagUniversalIdentifiers: string[]`.
- `PermissionFlagManifest` type removed from `twenty-shared/application`.
- `PermissionFlag` re-export removed from `twenty-sdk/define`. `SystemPermissionFlag` is re-exported in its place.
- Retargeting a permission flag between roles is now classified as delete + create instead of update

 ### Not in this PR
- definePermissionFlag SDK function and top-level Manifest.permissionFlags catalog (apps defining their own custom flags). Until those land, permissionFlagUniversalIdentifiers only accepts SystemPermissionFlag.* UUIDs; arbitrary UUIDs fail validation.


packages/twenty-front/src/pages/settings/applications/tabs/SettingsApplicationPermissionsTab.tsx
```diff
@@ -15,10 +15,21 @@ import {
...
 } from 'twenty-shared/application';
+import {
+  type PermissionFlagType,
+  SystemPermissionFlag,
+} from 'twenty-shared/constants';
 import { isDefined } from 'twenty-shared/utils';
 import { v4 as uuidv4 } from 'uuid';
 import { FieldMetadataType } from '~/generated-metadata/graphql';
 
+const SYSTEM_PERMISSION_FLAG_BY_UNIVERSAL_IDENTIFIER = Object.fromEntries(
```
Weiko, line 26: To update once the FE supports apps permission flags

packages/twenty-server/src/engine/core-modules/application/application-manifest/converters/from-permission-flag-to-universal-flat-role-permission-flag.util.ts
```diff
@@ -1,26 +1,45 @@
...
   return {
-    universalIdentifier: permissionFlag.universalIdentifier,
+    universalIdentifier,
     applicationUniversalIdentifier,
     roleUniversalIdentifier,
-    permissionFlagUniversalIdentifier:
-      SystemPermissionFlag[permissionFlag.flag],
-    flag: permissionFlag.flag,
+    permissionFlagUniversalIdentifier,
+    flag: resolvedFlag as PermissionFlagType,
```
Weiko, line 42: legacy deprecated field, to remove in next PR

packages/twenty-apps/examples/postcard/src/roles/default-function.role.ts
```diff
@@ -1,4 +1,4 @@
-import { PermissionFlag, defineRole } from 'twenty-sdk/define';
```
martmull, line 1: don't update existing applicaitons without updating their twenty-sdk version (you should wait for 2.7.0 publish)

packages/twenty-apps/internal/call-recording/src/roles/default-role.ts
martmull, line 1: same

packages/twenty-apps/examples/postcard/src/roles/default-function.role.ts
Weiko: Good point! @martmull

Weiko: Actually internal apps ignore the version within the package (except for internal/twenty-for-twenty which does `"twenty-sdk": "npm:twenty-sdk@2.0.0"` instead of `"twenty-sdk": "2.0.0"`. I guess this is expected so the CI always compare with the most recent version

martmull: i don't get why we have those "npm:twenty-sdk@2.0.0" version. This looks to be a useless alias

Image

Weiko: Ok, as discussed, only updated the ones used in the CI (post-card and rich-app (which is used in e2e))

Weiko: hello world is not using permission flags so nothing to update

martmull (approved):

PR #20825, opened 2026-05-21
Fix breaking change in install app command

martmull:
add backward compatibility for twenty-sdk install command

PR #20848, opened 2026-05-22
Add check for breaking api changes

martmull:
- update ci-breaking-changes.yaml so it check for api contrat breaks
- check fails properly when removing fix https://github.com/twentyhq/twenty/pull/20825
- check it turns green again when adding fix back


charlesBochet (approved):
Thank you!

PR #20861, opened 2026-05-22
fix(front): keep app variable cache in sync after update

nicolas-besnard:
Updating an application variable in Workspace / Applications / <App> / Settings persisted server-side but the Apollo cache kept the old value. Switching tabs unmounted the settings tab, and remounting reseeded the input from the stale cache — only a full refresh showed the new value.

Mutation now writes the new value into the ApplicationVariable entity via cache.modify, so FindOneApplication reflects the change immediately. Hook + table updated to pass the variable id through.

Adds a hook test that pre-seeds the cache and asserts the cached value after the mutation.

NOTE: saving the plain value in Apollo's cache might not be the best approach here


packages/twenty-front/src/pages/settings/applications/hooks/useUpdateOneApplicationVariable.ts
```diff
@@ -5,15 +5,27 @@ export const useUpdateOneApplicationVariable = () => {
...
     applicationId: string;
   }) => {
-    return await mutate({ variables: { key, value, applicationId } });
+    return await mutate({
+      variables: { key, value, applicationId },
+      update: (cache) => {
+        cache.modify({
+          id: cache.identify({ __typename: 'ApplicationVariable', id }),
+          fields: {
+            value: () => value,
```
martmull: We should not update if isSecret is true

nicolas-besnard: I agree but it means that we will get same issue gain with secret variable.

Another solution would be to return the encrypted value in `application-variable.service.ts` and make the `updateOneApplicationVariable` mutation return it. 

WDYT?

martmull: Yes, you can just to a `refetchQueries: [getOperationName(FIND_ONE_APPLICATION) ?? ''],` in that case
A little bit heavy, but ok as this won't change a lot

martmull (approved):

PR #20864, opened 2026-05-23
Add create record commands for each object in the command menu

Bonapara:
# Goal

Allow users to create records in less clicks. For instance, to create a task today you have to go on the "tasks object" and click the "add task" button. This feature allows to quickly add records from anywhere without losing your current focus

The ability to create tasks and notes faster is heavily requested by customers.

It will also be a great help for the **"record is created" workflow trigger** that was used to send an empty record to the workflow

## Summary

- Add global create-record command menu entries for manually creatable objects.
- Add the create-record side panel flow and preserve the existing current-index create behavior for null payloads.
- Harden CREATE_NEW_RECORD payload validation and add focused frontend/server coverage.

https://github.com/user-attachments/assets/b5d601ff-7d2e-4495-9d80-7690b6e6e633

## Validation

- `npx nx test twenty-front --testFile=src/modules/command-menu-item/engine-command/record/no-selection/components/__tests__/CreateNewRecordCommand.test.tsx --codeCoverage=false --runInBand`
- `npx nx test twenty-front --testFile=src/modules/side-panel/pages/create-record/components/__tests__/SidePanelCreateRecordPage.test.tsx --codeCoverage=false --runInBand`
- `npx nx test twenty-server --testFile=src/engine/workspace-manager/workspace-migration/workspace-migration-builder/validators/services/__tests__/flat-command-menu-item-validator.service.spec.ts --codeCoverage=false --runInBand`
- `npx nx lint:diff-with-main twenty-front`
- `npx nx lint:diff-with-main twenty-server`

## Note

The targeted backend integration spec could not run locally because the test Postgres setup is missing role `postgres` before assertions start.


packages/twenty-server/src/engine/metadata-modules/command-menu-item/constants/command-menu-item-engine-key-coherence-constraint-sql.constant.ts
```diff
@@ -0,0 +1,6 @@
...
+  'CHK_CMD_MENU_ITEM_ENGINE_KEY_COHERENCE';
+
+const CREATE_NEW_RECORD_PAYLOAD_SQL = `("payload" IS NULL OR (jsonb_typeof("payload") = 'object' AND "payload" ? 'objectMetadataItemId' AND jsonb_typeof("payload" -> 'objectMetadataItemId') = 'string' AND "payload" ->> 'objectMetadataItemId' <> '' AND NOT ("payload" ? 'path')))`;
+
+export const COMMAND_MENU_ITEM_ENGINE_KEY_COHERENCE_CONSTRAINT_SQL = `("engineComponentKey" = 'TRIGGER_WORKFLOW_VERSION' AND "workflowVersionId" IS NOT NULL AND "frontComponentId" IS NULL AND "payload" IS NULL) OR ("engineComponentKey" = 'FRONT_COMPONENT_RENDERER' AND "frontComponentId" IS NOT NULL AND "workflowVersionId" IS NULL AND "payload" IS NULL) OR ("engineComponentKey" = 'NAVIGATION' AND "payload" IS NOT NULL AND "workflowVersionId" IS NULL AND "frontComponentId" IS NULL) OR ("engineComponentKey" = 'CREATE_NEW_RECORD' AND "workflowVersionId" IS NULL AND "frontComponentId" IS NULL AND ${CREATE_NEW_RECORD_PAYLOAD_SQL}) OR ("engineComponentKey" NOT IN ('TRIGGER_WORKFLOW_VERSION', 'FRONT_COMPONENT_RENDERER', 'NAVIGATION', 'CREATE_NEW_RECORD') AND "workflowVersionId" IS NULL AND "frontComponentId" IS NULL AND "payload" IS NULL)`;
```
charlesBochet, line 6: Feel this is unrealistic, I have no experience with such big constraint, please check performances there

charlesBochet (changes requested):
This is a heavy one where the complexity mainly is in QAing the behavior + making sure upgrade command works as expected + making sure performance keep getting good.

Might be a bit ambitious @Bonapara unless you invest in testing.

Also, could you clarify PR description to be more product oriented, I'm not sure to fully understand the goal of this change

packages/twenty-server/src/engine/metadata-modules/command-menu-item/constants/command-menu-item-engine-key-coherence-constraint-sql.constant.ts
Bonapara: I checked this locally with synthetic `commandMenuItem` rows on Postgres 16.

Results:
- 100k rows: existing constraint validation `19.60ms`; new constraint validation with CREATE_NEW_RECORD object payloads `19.33ms`; inserting 10k CREATE_NEW_RECORD rows with the new constraint `21.08ms`.
- 1M rows: new constraint validation `230.26ms`; inserting 100k CREATE_NEW_RECORD rows with the new constraint `270.83ms`.

So I do not expect runtime read/search impact: this is a CHECK constraint, so it is evaluated on command-menu writes and during constraint validation, not on command-menu reads. Also, `commandMenuItem` should be low-cardinality in real workspaces compared with the benchmark sizes.

I agree the main thing to be careful about is upgrade behavior when replacing the constraint, especially locking/validation, more than expression CPU cost.

charlesBochet: Cardinality is way higher than that, can you test with and without the constraint with 10M rows in the table which is more realistic

Bonapara: Codex re-ran the benchmark locally on Postgres 16.6 with 10M-row temp tables, comparing the old constraint, the PR constraint, and inserts with/without the constraint. No Twenty data was touched.

Results:
- 10M mixed command-menu rows: old constraint validation `1.404s`; new PR constraint validation `1.321s`.
- After adding CREATE_NEW_RECORD object payload rows, 10.1M rows total: new PR constraint validation `1.364s`.
- Insert 100k CREATE_NEW_RECORD object-payload rows: `76ms` without the constraint vs `120ms` with the constraint.
- Worst case, 10M rows all CREATE_NEW_RECORD object payloads: PR constraint validation on 10.1M rows `5.179s`; warmed 100k insert `74ms` without the constraint vs `120ms` with the constraint.

Codex's read: the expression CPU cost looks acceptable even at 10M rows. The command-menu read path is unaffected because this is a CHECK constraint, so it is evaluated on writes and during validation, not reads.

The remaining thing worth improving is upgrade locking: adding the replacement constraint as `NOT VALID`, then running `VALIDATE CONSTRAINT`, would keep the heavier validation scan out of the strongest lock window.

charlesBochet:
Closing as staled, let's re-open if needed :)

FelixMalfait:
Will take over this when I have time (after improving the onboarding/settings most likely)

FelixMalfait:
@Weiko will take over

PR #20866, opened 2026-05-23
refactor(jwt): gate signing-key auto-rotation cron on SIGNING_KEY_ROTATION_DAYS

charlesBochet:
Only register the JWT signing-key rotation cron when `SIGNING_KEY_ROTATION_DAYS` is set, and move that variable to Advanced Settings.


prastoin (approved):

prastoin:
Only caveat I can think of is that we can't really see in the metric if a skipped job is not registered or just being skipped whereas if it was ending up as ended up as a noop from the job logic directly

But the tradeof is worth it versus filling the queue with no ops jobs !

PR #20868, opened 2026-05-23
Downscope code owners and cover more paths

prastoin:


charlesBochet:
@prastoin Pinging you, not sure this is the right direction

Weiko (commented):
Let's close for now and discuss next week

PR #20873, opened 2026-05-23
feat(logic-function): add LIVE / PREBUILT execution modes

charlesBochet:
### Why

1. Sending the code to the lambda (~1Mb usually) is heavy on network and results to a constant traffic of ~30Mb/s on AWS which results into TB of network data every month
2. eval(1MB of code) is not that fast, it's heavy on memory and CPU on lambda side

### High level

Adds two execution modes for logic functions, gated behind the new `IS_LOGIC_FUNCTION_PREBUILT_MODE_ENABLED` workspace feature flag (off everywhere by default):

- **LIVE** (current behavior, preserved bit-for-bit): the compiled bundle is read from object storage and shipped in every Lambda invoke payload. Used for fast iteration in the workflow editor / Settings test runs.
- **PREBUILT** (new): the bundle is installed onto the per-function Lambda alongside the unified executor, and invocations carry only `{ params, env, handlerName }` — saving JSON payload egress and warm-start `import()` cost on every call.

### Key design choices

- **Unified Lambda handler** (`constants/executor/index.mjs`) dispatches at runtime: `event.code` present ? LIVE (write to `/tmp`, dynamic import) : `import('./prebuilt-logic-function.mjs')`. Both code paths always coexist on the deployment package, so the same Lambda can serve either mode without redeploying.
- **Install runs inside the `validateBuildAndRun` migration pipeline**, not at execute time. `Create/UpdateLogicFunctionActionHandlerService` calls `driver.installPrebuiltBundle` when `executionMode` flips LIVE?PREBUILT or `checksum` changes while PREBUILT, gated on `isBuildUpToDate=true` and a fresh checksum.
- **Strict execute, no reconciliation**: `LogicFunctionExecutorService.execute` resolves `effectiveExecutionMode` (caller override > feature flag > entity column). For PREBUILT it asks the driver `getInstalledBundleChecksum` (Lambda `twenty:bundle-checksum` tag for AWS, sidecar file locally) and throws `LOGIC_FUNCTION_PREBUILT_BUNDLE_NOT_INSTALLED` on mismatch.
- **Feature flag gates every side effect**: with the flag off the executor forces LIVE, the action-handler install hooks bail before AWS, and workflow activation does not flip the mode. Rollback is just turning the flag off.

### Lifecycle

- New workflow CODE step ? `LIVE`, no install.
- Workflow activated ? build + activation flips `executionMode=PREBUILT` ? action-handler installs the bundle + sets the Lambda tag.
- Draft from active version ? duplicated logic function reset to `LIVE`.
- App install ? manifest converter sets `PREBUILT`, create-action handler installs.
- Test runs (`executeOneFromSource`, workflow editor) pass `executionMode=LIVE` explicitly.

### Observability

`[lambda-timing]` log lines now include `effectiveExecutionMode` and `payloadBytes`; the action handler logs `install_duration_ms` for each install.

## Test plan

- [x] `npx nx typecheck twenty-server` ? passes
- [x] `npx oxlint --type-aware` on all changed files ? 0 warnings, 0 errors
- [x] `npx nx test twenty-server` ? 588 suites / 5009 tests pass (no regressions vs main)
- [x] New unit suite `flat-logic-function-validator.service.spec.ts` ? 9/9
- [x] Existing `workflow-version-step-operations.workspace-service.spec.ts` ? 8/8 (verified the new token-based DI avoids a circular-import regression)
- [x] Snapshot for `ALL_UNIVERSAL_FLAT_ENTITY_PROPERTIES_TO_COMPARE_AND_STRINGIFY` updated to include `executionMode`
- [x] Integration suite `logic-function-execution.integration-spec.ts` extended to assert `executionMode=LIVE` on newly-created functions and continues to exercise the LIVE happy path
- [ ] Manual staging rollout: flip `IS_LOGIC_FUNCTION_PREBUILT_MODE_ENABLED` per workspace, observe `[lambda-timing]` `payloadBytes` drop + `install_duration_ms`, then ramp in prod.


packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/local.driver.ts
```diff
@@ -300,13 +317,19 @@ export class LocalDriver implements LogicFunctionDriver {
...
+      // LIVE: fetch the latest bundle from object storage on every invoke.
+      // PREBUILT: use the locally-cached bundle written during install.
+      const inMemoryBuiltHandlerPath = isPrebuilt
+        ? this.getInstalledBundlePath(flatLogicFunction)
+        : await this.logicFunctionResourceService.copyBuiltCodeInMemory({
+            workspaceId: flatLogicFunction.workspaceId,
+            applicationUniversalIdentifier,
+            builtHandlerPath: flatLogicFunction.builtHandlerPath,
+            inMemoryDestinationPath: sourceTemporaryDir,
+          });
```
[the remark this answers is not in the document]
charlesBochet: Fixed: the prebuilt bundle is copied into `sourceTemporaryDir` before execution (see `copyPrebuiltBundleIntoExecutionDir`), so Node walks up the same tree as the assembled `node_modules`.

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-executor/logic-function-executor.service.ts
```diff
@@ -111,6 +124,19 @@ export class LogicFunctionExecutorService {
...
+      flatLogicFunction,
+      callerOverride: executionMode,
+    });
+
+    if (effectiveExecutionMode === LogicFunctionExecutionMode.PREBUILT) {
+      await this.assertPrebuiltBundleInstalled({
+        driver,
+        flatLogicFunction,
+      });
+    }
```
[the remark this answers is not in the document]
charlesBochet: Per-invoke check is intentional. `GetFunction` cost is in the noise relative to a Lambda invoke, and trusting the DB would let drift silently route to the wrong bundle. Happy to add an in-memory TTL cache later if it shows up in latency budgets.

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/lambda.driver.ts
```diff
@@ -1258,4 +1286,101 @@ export class LambdaDriver implements LogicFunctionDriver {
...
+  async getInstalledBundleChecksum(
+    flatLogicFunction: FlatLogicFunction,
+  ): Promise<string | null> {
+    const lambdaExecutor = await this.getLambdaExecutor(flatLogicFunction);
+
+    if (!isDefined(lambdaExecutor)) {
+      return null;
+    }
+
+    return lambdaExecutor.Tags?.[LAMBDA_PREBUILT_BUNDLE_CHECKSUM_TAG] ?? null;
```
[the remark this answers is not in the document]
charlesBochet: Same as the executor thread, intentional for now. Will revisit with caching once we have real invocation metrics.

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-executor/logic-function-executor.service.ts
```diff
@@ -111,6 +124,19 @@ export class LogicFunctionExecutorService {
...
     const driver = this.logicFunctionDriverFactory.getCurrentDriver();
 
+    const effectiveExecutionMode = await this.resolveEffectiveExecutionMode({
+      workspaceId,
+      flatLogicFunction,
+      callerOverride: executionMode,
+    });
+
+    if (effectiveExecutionMode === LogicFunctionExecutionMode.PREBUILT) {
+      await this.assertPrebuiltBundleInstalled({
```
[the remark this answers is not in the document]
charlesBochet: Same as the Copilot threads on this same path, the per-invoke check is intentional. We can layer an in-memory cache later if metrics call for it.

packages/twenty-server/src/engine/core-modules/application/application-manifest/converters/from-logic-function-manifest-to-universal-flat-logic-function.util.ts
```diff
@@ -37,6 +40,7 @@ export const fromLogicFunctionManifestToUniversalFlatLogicFunction = ({
     workflowActionTriggerSettings:
       logicFunctionManifest.workflowActionTriggerSettings ?? null,
     isBuildUpToDate: true,
+    executionMode: LogicFunctionExecutionMode.PREBUILT,
```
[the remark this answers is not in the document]
charlesBochet: Mitigated by `623a1ba`: install always runs when `executionMode === PREBUILT && isBuildUpToDate && checksum != null`, regardless of the feature flag. The flag only gates the executor runtime path, so a flag flip can never land on a missing bundle.

packages/twenty-server/src/engine/workspace-manager/workspace-migration/workspace-migration-runner/action-handlers/logic-function/services/update-logic-function-action-handler.service.ts
```diff
@@ -94,4 +114,77 @@ export class UpdateLogicFunctionActionHandlerService extends WorkspaceMigrationR
...
+        UpdateLogicFunctionActionHandlerService.name,
+      );
+    } catch (error) {
+      this.logger.error(
+        `Failed to install prebuilt bundle for function ${existingLogicFunction.id} ` +
+          `after ${Date.now() - installStart}ms: ` +
+          `${error instanceof Error ? error.message : String(error)}`,
+        UpdateLogicFunctionActionHandlerService.name,
+        error instanceof Error ? error.stack : undefined,
+      );
```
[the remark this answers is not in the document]
charlesBochet: Not a bug here. The action handlers inherit `protected readonly logger: LoggerService` from the base handler, and that custom `LoggerService.error` signature is `error(message, category, ...optionalParams)`, so `(message, ClassName.name, stack)` is the documented order. The lambda driver uses the plain NestJS `Logger`, which is why its call sites look different.

prastoin:
@cubic-dev-ai review

packages/twenty-server/src/engine/core-modules/application/application-manifest/converters/from-logic-function-manifest-to-universal-flat-logic-function.util.ts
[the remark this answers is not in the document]
martmull: not sure on this one, PREBUILT in dev mode this can be an issue isn't it?

martmull (approved):
Looks good, just a command on logic function to flat default execution mode

packages/twenty-server/test/integration/graphql/suites/workflow/code-step-prebuilt-workflow.integration-spec.ts
```diff
@@ -0,0 +1,314 @@
...
+
+export const main = async (params: { items: Array<{ category: string; name: string }> }): Promise<object> => {
+  const grouped = groupBy(params.items, 'category');
+  return {
+    grouped,
+    categories: Object.keys(grouped),
+  };
+};`;
+
+describe('Code step workflow with PREBUILT logic function (e2e)', () => {
```
prastoin, line 27: ***Remark:*** interesting that's funny that you've went for the through workflow integration test
I feel like this would be cool to keep long term
Also we could cover the logic function execution through apps test context

packages/twenty-server/src/modules/workflow/workflow-builder/workflow-version-step/code-step/services/code-step-build.service.ts
```diff
@@ -114,4 +117,94 @@ export class CodeStepBuildService {
...
+        : undefined;
+
+      if (!isDefined(applicationUniversalIdentifier)) {
+        throw new WorkflowTriggerException(
+          `Logic function '${logicFunctionId}' references applicationId='${flatLogicFunction.applicationId ?? 'null'}' that did not resolve to an application`,
+          WorkflowTriggerExceptionCode.INTERNAL_ERROR,
+        );
+      }
+
+      await this.logicFunctionFromSourceHelperService.updateOneFromMetadata({
```
prastoin, line 199: ***Remark:*** out of scope, below method cache invalidation is redundant as the validate build and run already handles it
```ts
    const { flatLogicFunctionMaps: recomputedFlatLogicFunctionMaps } =
      await this.flatEntityMapsCacheService.getOrRecomputeManyOrAllFlatEntityMaps(
        {
          workspaceId,
          flatMapsKeys: ['flatLogicFunctionMaps'],
        },
      );
```

packages/twenty-server/src/modules/workflow/workflow-trigger/workspace-services/workflow-trigger.workspace-service.ts
```diff
@@ -136,6 +136,11 @@ export class WorkflowTriggerWorkspaceService {
           steps: workflowVersion.steps ?? [],
         });
 
+        await this.codeStepBuildService.switchCodeStepLogicFunctionsToPrebuilt({
+          workspaceId,
+          steps: workflowVersion.steps ?? [],
+        });
```
prastoin, line 142: ***remark:*** at first sight seems quite weird for both method calls not to be merged but I got your idea, caller could dynamically call the `switchCodeStepLogicFunctionsToPrebuilt`, that's pure nitpicking

packages/twenty-server/src/modules/workflow/workflow-builder/workflow-version-step/code-step/utils/extract-code-step-logic-function-ids-from-workflow-steps.util.ts
```diff
@@ -0,0 +1,51 @@
...
+type WorkflowStepForLogicFunctionIdExtraction =
+  | CodeStepForLogicFunctionIdExtraction
+  | NonCodeStepForLogicFunctionIdExtraction;
+
+const isCodeStepForLogicFunction = (
+  step: WorkflowStepForLogicFunctionIdExtraction,
+): step is CodeStepForLogicFunctionIdExtraction =>
+  step.type === WorkflowActionType.CODE;
+
+export const extractCodeStepLogicFunctionIdsFromWorkflowSteps = (
```
prastoin, line 37: ***Bug:*** Shouldn't we use a set here ? what if a workflow has two code node that shares the same logic function id ?

packages/twenty-server/src/modules/workflow/workflow-builder/workflow-version-step/code-step/services/code-step-build.service.ts
```diff
@@ -114,4 +117,94 @@ export class CodeStepBuildService {
...
+
+    const { flatLogicFunctionMaps, flatApplicationMaps } =
+      await this.workspaceManyOrAllFlatEntityMapsCacheService.getOrRecomputeManyOrAllFlatEntityMaps(
+        {
+          workspaceId,
+          flatMapsKeys: ['flatLogicFunctionMaps', 'flatApplicationMaps'],
+        },
+      );
+
+    for (const logicFunctionId of logicFunctionIds) {
```
prastoin, line 153: ***Suggestion:*** Could create an update many to avoid looping here and perform batch update
```ts
      await this.logicFunctionFromSourceHelperService.updateOneFromMetadata({
```

packages/twenty-server/src/engine/workspace-manager/workspace-migration/workspace-migration-runner/action-handlers/workspace-schema-migration-runner-action-handlers.module.ts
```diff
@@ -100,6 +101,7 @@ import { UpdateConnectionProviderActionHandlerService } from 'src/engine/workspa
     TypeOrmModule.forFeature([ApplicationEntity]),
     WorkspaceSchemaManagerModule,
     SecretEncryptionModule,
+    FeatureFlagModule,
```
prastoin: ***Self reminder:*** unless I'm mistaken the feature flag should be accessible from the additional cache maps, double check why this is added to the DI

prastoin (commented):
Reviewed 1/3 of the PR ( continuing after lunch )

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/lambda.driver.ts
```diff
@@ -1287,4 +1302,122 @@ export class LambdaDriver implements LogicFunctionDriver {
...
+  }
```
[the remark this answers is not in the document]
prastoin: yep we need to verify roles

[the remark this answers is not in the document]
prastoin: We could also have a local cache in order to avoid having to perform a GET everytime

packages/twenty-server/src/modules/workflow/workflow-builder/workflow-version-step/code-step/services/code-step-build.service.ts
charlesBochet: I can do that as a follow up

packages/twenty-server/src/modules/workflow/workflow-builder/workflow-version-step/code-step/utils/extract-code-step-logic-function-ids-from-workflow-steps.util.ts
charlesBochet: Done, switched to a Set.

packages/twenty-server/src/engine/workspace-manager/workspace-migration/workspace-migration-runner/action-handlers/workspace-schema-migration-runner-action-handlers.module.ts
charlesBochet: Right, dropped.

packages/twenty-server/src/modules/workflow/workflow-builder/workflow-version-step/code-step/services/code-step-build.service.ts
charlesBochet: will follow up in a separate PR

packages/twenty-server/src/modules/workflow/workflow-trigger/workspace-services/workflow-trigger.workspace-service.ts
charlesBochet: keeping it split for now

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/drivers/lambda.driver.ts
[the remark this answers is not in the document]
charlesBochet: Filed twentyhq/twenty-infra#699 to grant lambda:ListTags and lambda:TagResource. Holding off on a local cache for now.

prastoin (approved):

Issue #1009, opened 2023-07-31
Discuss `useRef` usage


thaisguigon:
https://react.dev/reference/react/useRef

> `useRef` is a React Hook that lets you reference a value that’s not needed for rendering.

As discussed with @magrinj and @emilienchvt, we would like to reassess how we can use this hook for better performance. Right now it is mostly only used for manipulating the DOM with HTML element refs but it could be useful in other cases where we want to keep reference to a value without triggering re-renders.

magrinj:
I completely agree. While `useRef` is often used for referencing mutable DOM elements, its utility goes far beyond that. As pointed out in the React documentation, it can be used to maintain a reference to a value that does not affect the render output. This can include values from a setInterval operation, timers, or other data sources that need to persist across multiple render cycles.

The major advantage here is that changes to these references do not trigger component re-renders, which can significantly improve performance in scenarios where the reference is updated frequently but the view does not need to be refreshed.

Beyond this, I think `useRef` could be beneficial in scenarios involving complex calculations or tasks that we wouldn't want to repeat unnecessarily due to re-renders. For example, we might store the results of such computations in a ref, ensuring that we only perform them once and use the stored result thereafter.

In essence, we can look at `useRef` as a mechanism for storing "memory" or "state" that doesn't affect the render output but might be important for the logic of our components. This can open the door to more performant and efficient components. Looking forward to exploring this further with the team!

magrinj:
cc @charlesBochet @lucasbordeau

charlesBochet:
My opinion is that we should avoid it except when using it to manipulate DOM reference.
To me:
- useRef is very similar to useState. If we do need to do complex computations useCallback (+ useState) can achieve the same output.
- they are rare cases where we need it (I think setInterval is one if you want to be able to clearInterval)

As this is introducing a second way to do things we can do with setState with the same complexity (I personnaly find that useRef API complexifies code reading but this is subjective), I would strongly advocate to have one way of doing things. This will help keeping the codebase uniform, and prevent discussions on either we should use useRef or useState in "THIS" specific cases (which will likely never ends are they are the same). Keeping clear coding conventions is a very important winning condition of this project.

In a nutshell, I would really push to keep only using useState + useCallback on the project, except if there is no easy way to do so. Do you have a particular example into mind?

@lucasbordeau has refactored the whole table component (which is by far the most complex we have), there are no more re-render on table, it's performant and no useRef have been used

magrinj:
[...]

PR #20876, opened 2026-05-24
chore(security): bump protobufjs and esbuild to clear CVEs

charlesBochet:
## Summary

A self-hoster reported that Trivy blocks the `twentycrm/twenty:v2.7.x` image on three fixed-critical CVEs. The reachable risk is low (none of the vulnerable code paths are exposed to attacker-controlled input in our deployment), but the findings are real and easy to clear by bumping the affected dependencies in their owning workspaces.

> **Follow-up:** This PR clears two of the three CVEs but leaves CVE-2025-68121 unaddressed because `esbuild@0.27.x` is built with Go 1.23.8, which is past Go's support window. The remaining fix is in **#20902** (bumps esbuild to `^0.28.0`, built with Go 1.26.1).

### CVE-2026-41242 — `protobufjs` < 7.5.5

Pulled transitively into the production image via `@opentelemetry/sdk-node`, `@opentelemetry/auto-instrumentations-node`, and `@grpc/grpc-js` → `@grpc/proto-loader`. Lockfile was on 7.5.3; this matches dismissed dependabot alert #1009 (Critical 9.4).

**Fix:** add `protobufjs: ^7.5.5` as a direct dep of `twenty-server` (the workspace that exercises it via the OpenTelemetry gRPC exporters) and run `yarn dedupe protobufjs` to collapse the residual transitive 7.5.3 copy. Resolves to 7.6.0.

### CVE-2024-24790 — Go stdlib (net/netip IPv6 parsing), fixed in Go 1.21.11 / 1.22.4

Present in the Go-built `bin/esbuild` shipped by `@esbuild/<platform>` packages. Two paths put esbuild into the production image:

1. `twenty-client-sdk` declares `esbuild` as a runtime dep (used by its `./generate` entry point).
2. `twenty-server` had `@lingui/vite-plugin` in `dependencies`, which pulls `@lingui/cli` as a runtime sub-dep, which bundles `esbuild@0.21.5` nested under `node_modules/@lingui/cli/node_modules/esbuild/`.

**Fix:**
- Bump `twenty-client-sdk`'s `esbuild` from `^0.25.0` to `^0.27.3` (built with Go 1.23.8 — patched for CVE-2024-24790).
- Move `@lingui/vite-plugin` from `dependencies` to `devDependencies` in `twenty-server`. The plugin is not imported by any source file — it was misclassified.

### CVE-2025-68121 — Go stdlib (crypto/tls cert validation bypass), CVSS 10.0 / Critical — addressed in follow-up #20902

`esbuild@0.27.x` is built with Go 1.23.8, which is past Go's support window. The Go security team only patched 1.24.13 / 1.25.7 / 1.26.0-rc.3+ for this issue, so 0.27.x still ships a vulnerable Go toolchain. PR #20902 bumps `esbuild` to `^0.28.0` (built with Go 1.26.1) to close this. esbuild does not actually use `crypto/tls` at runtime, so this is not exploitable — the fix is solely to clear the Trivy finding.

### Verification

Ran `yarn workspaces focus --production twenty twenty-server twenty-emails twenty-shared twenty-client-sdk` (the same command the Dockerfile uses) and inventoried the resulting `node_modules`. After all three changes:

- `node_modules/esbuild/` → **0.27.7 only** (Go 1.23.8 — patches CVE-2024-24790)
- `node_modules/protobufjs/` → **7.6.0 only** (patches CVE-2026-41242)

No nested copies of either package remain in the production install. CVE-2025-68121 still requires #20902.

### Follow-up worth tracking separately

`esbuild` should arguably not be in `twenty-client-sdk`'s `dependencies` at all — only the `./generate` entry point uses it, and the server never imports that entry. Moving it to optional `peerDependencies` would stop shipping a Go binary into the production image entirely. Out of scope for this PR.

### DIFF
--- a/packages/twenty-client-sdk/package.json
+++ b/packages/twenty-client-sdk/package.json
@@ -42,7 +42,7 @@
   "dependencies": {
     "@genql/cli": "^3.0.3",
     "@genql/runtime": "^2.10.0",
-    "esbuild": "^0.25.0",
+    "esbuild": "^0.27.3",
     "graphql": "^16.8.1"
   },
   "devDependencies": {
--- a/packages/twenty-server/package.json
+++ b/packages/twenty-server/package.json
@@ -168,6 +168,7 @@
     "planer": "1.2.0",
     "pluralize": "8.0.0",
     "postal-mime": "^2.6.1",
+    "protobufjs": "^7.5.5",
     "psl": "^1.9.0",
     "react": "18.3.1",
     "react-dom": "18.3.1",
--- a/packages/twenty-server/package.json
+++ b/packages/twenty-server/package.json
@@ -50,7 +50,6 @@
     "@lingui/core": "^5.1.2",
     "@lingui/format-po": "5.1.2",
     "@lingui/react": "5.1.2",
-    "@lingui/vite-plugin": "5.1.2",
     "@microsoft/microsoft-graph-client": "3.0.7",
     "@microsoft/microsoft-graph-types": "^2.40.0",
     "@nestjs/axios": "3.1.2",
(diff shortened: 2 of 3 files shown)


Weiko (commented):
Can we merge this @charlesBochet ?

FelixMalfait (approved):

Issue #20877, opened 2026-05-24
Twenty AI fails with OpenAI Zero Data Retention: "Items are not persisted for Zero Data Retention organizations"


mfamularopsyc:
## Summary

Twenty AI fails when using an OpenAI API key from an organization/project with Zero Data Retention enabled.

The AI chat returns:
Failed to get response
Item with id 'fc_0ca7d25a52784c2f016a1316f336988197a4a249c10a0b7f27' not found. Items are not persisted for Zero Data Retention organizations. Remove this item from your input and try again.

### Environment
 - Twenty version: v2.7.3
 - Deployment: self-hosted Docker
 - OpenAI provider: OpenAI API key
 - OpenAI organization/project: Zero Data Retention enabled
 - Redis: local Docker Redis
 - Postgres: AWS RDS PostgreSQL

### Steps to reproduce
 - Self-host Twenty.
 - Configure Twenty AI with an OpenAI API key from a Zero Data Retention organization/project.
 - Open AI chat in Twenty.
 - Send a prompt.
 - The request fails with the error below.

### Actual behavior
Twenty returns:
Failed to get response
Item with id 'fc_...' not found. Items are not persisted for Zero Data Retention organizations. Remove this item from your input and try again.

## Expected behavior
Twenty should either:
 - Support OpenAI Zero Data Retention organizations by using a stateless request flow that does not depend on persisted OpenAI item IDs; or
 - Detect this OpenAI error and show a clear message that OpenAI Zero Data Retention is currently unsupported.

## Notes
The error appears to come from OpenAI because Zero Data Retention organizations do not persist items. It seems Twenty may be referencing a previously returned OpenAI item/tool-call ID in a later request.

Could Twenty AI avoid sending persisted item references when using OpenAI, or provide a configuration option for ZDR-compatible/stateless OpenAI requests?

mfamularopsyc:
I opened a PR for this here: #20880

It adds an opt-in `OPENAI_ZERO_DATA_RETENTION_ENABLED` setting and sanitizes replayed OpenAI chat history when enabled so ZDR organizations do not receive persisted OpenAI/tool item references.

FelixMalfait:
Related:
https://github.com/vercel/ai/issues/7901
https://github.com/vercel/ai/issues/10060
https://github.com/vercel/ai/issues/9119

[...]

PR #20880, opened 2026-05-25
Fix OpenAI ZDR chat history replay

mfamularopsyc:
## Summary

Adds an opt-in OpenAI Zero Data Retention compatibility mode for AI chat history replay.

When `OPENAI_ZERO_DATA_RETENTION_ENABLED=true` and the selected SDK package is `@ai-sdk/openai`, replayed chat history is sanitized before model-message conversion. Prior tool results/errors are preserved as plain text context, while provider-specific persisted tool/item IDs are removed so ZDR organizations do not receive references to non-persisted OpenAI items.

The setting defaults to `false`, so existing OpenAI and non-OpenAI behavior is unchanged unless explicitly enabled.

## Tests

Passed:
- `node_modules/.bin/tsgo -p packages/twenty-server/tsconfig.json`
- Focused AI ZDR Jest spec: 3 tests passed
- Full `twenty-server` Jest suite: passed
- Changed-file lint: passed
- Changed-file format: passed
- Full server format check: passed
- `git diff --check`

Could not verify locally:
- `yarn nx typecheck twenty-server` failed before project execution with `Failed to start plugin worker`.
- Full server lint fails on existing unrelated files outside this change.

Related to issue #20877

### DIFF
--- a/packages/twenty-server/.env.example
+++ b/packages/twenty-server/.env.example
@@ -87,6 +87,7 @@ FRONTEND_URL=http://localhost:3001
 # ———————— AI ————————
 # API keys for built-in providers (also editable from Admin Panel > Config Variables):
 # OPENAI_API_KEY=
+# OPENAI_ZERO_DATA_RETENTION_ENABLED=false
 # ANTHROPIC_API_KEY=
 # GOOGLE_API_KEY=
 # XAI_API_KEY=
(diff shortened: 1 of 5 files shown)


FelixMalfait:
Thanks, I'm not sure this is the right fix, will comment in issue!

Issue #20879, opened 2026-05-24
[Security/Medium]: Missing path traversal validation in public-assets endpoint


sulthonzh:
## Description
The `public-assets/:workspaceId/:applicationId/*path` endpoint in `FileController` does not validate the filepath parameter before using it, creating a potential path traversal vulnerability.

## Steps to Reproduce
1. Navigate to `/public-assets/{workspaceId}/{applicationId}/../../sensitive-file`
2. The controller joins path segments without validation:
   ```typescript
   // file.controller.ts:49
   const filepath = join(...req.params.path);
   ```
3. This filepath is passed to database query and storage operations

## Impact
- Database query runs with unvalidated filepath before storage validation
- Could allow attackers to enumerate files across different workspaces/applications
- While `assertStoragePathIsSafe` at storage layer blocks `../`, the database query could succeed if malicious paths exist
- Defense-in-depth issue - bypass possible if storage driver changes or validation is disabled
- Public assets are accessible without authentication (@UseGuards(PublicEndpointGuard, NoPermissionGuard))

## Suggested Fix
Add validation in the controller before database query:

```typescript
// In file.controller.ts, after line 48
import { assertStoragePathIsSafe } from 'src/engine/core-modules/file-storage/utils/assert-storage-path-is-safe.util';

// Validate filepath before database query
assertStoragePathIsSafe(filepath);

// OR sanitize at controller level
const normalized = path.normalize(filepath);
if (normalized.split(path.sep).includes('..')) {
  throw new FileException(
    'Invalid filepath: path traversal detected',
    FileExceptionCode.ACCESS_DENIED,
  );
}
```

## Environment
- Version: 0.2.1
- OS: macOS

## References
- Vulnerable file: packages/twenty-server/src/engine/core-modules/file/controllers/file.controller.ts
- Route: GET /public-assets/:workspaceId/:applicationId/*path

prastoin:
Hello there,
Thanks for your report !
Unless I'm mistaken nothing exploitable here, but the concern is legit
`fileStorageService` operations are validated and sanitized (including writes). The DB query is also workspaceId + applicationId scoped, and it's an exact-match on the path column rather than a filesystem traversal, there's nothing to walk up from at db lvl

That said, it's a legit code smell, raising a PR shortly to fail earlier and remove the ambiguity. Thanks for flagging!

prastoin:
Also please prioritize opening security advisories ( CVE ) when it comes to raise security concerns 🙏 @sulthonzh

PR #20881, opened 2026-05-25
Early return in public assets

prastoin:
# Introduction
Related https://github.com/twentyhq/twenty/issues/20879

More abstracted response error and cleaner integrity check before performing any in database search
Nothing critical patched here

Also added integration coverage to the related endpoint

Fixed the stream on error throw that would have been bubbling up into node process

## Next
Once this has been approved will re-apply to all the existing prone file.getBy* methods and controllers endpoints

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/file/controllers/file.controller.ts
+++ b/packages/twenty-server/src/engine/core-modules/file/controllers/file.controller.ts
@@ -18,7 +18,6 @@ import {
   FileStorageExceptionCode,
 } from 'src/engine/core-modules/file-storage/interfaces/file-storage-exception';
 import { isSafeRelativePath } from 'src/engine/core-modules/file-storage/utils/is-safe-relative-path.util';
-
 import {
   FileException,
   FileExceptionCode,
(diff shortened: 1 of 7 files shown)


packages/twenty-server/src/engine/core-modules/file/controllers/file.controller.ts
```diff
@@ -48,40 +51,55 @@ export class FileController {
...
+    setFileResponseHeaders(res, mimeType);
 
+    stream.on('error', () => {
       throw new FileException(
-        `Error retrieving file: ${error.message}`,
+        'Error streaming file from storage',
         FileExceptionCode.INTERNAL_SERVER_ERROR,
       );
-    }
+    });
```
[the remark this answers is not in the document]
prastoin: indeed pre-existing but will fix !

packages/twenty-server/src/engine/core-modules/file/controllers/file.controller.ts
```diff
@@ -48,39 +52,60 @@ export class FileController {
...
+      .getFileStreamByPath({
         workspaceId,
         applicationId,
         fileFolder: FileFolder.PublicAsset,
         filepath,
-      });
-
-      setFileResponseHeaders(res, mimeType);
+      })
+      .catch((error) => {
```
Weiko, line 74: Do we want all exceptions thrown by getFileStreamByPath to be re-thrown as 500?

Weiko (approved):
LGTM

packages/twenty-server/src/engine/core-modules/file/controllers/file.controller.ts
prastoin: I feel like we prefer by abstract on non authenticated endpoints
Whereas before we would have been leaking typeorm raw query

PR #20882, opened 2026-05-25
fix(billing) - skip redundant cap writes

etiennejouan:
fixes https://sonarly.com/issue/40207

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/billing/services/billing-usage-cap.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/billing/services/billing-usage-cap.service.ts
@@ -112,10 +112,6 @@ export class BillingUsageCapService {
 
     const [billingSubscriptionItem] = billingSubscriptionItems;
 
-    if (billingSubscriptionItem.hasReachedCurrentPeriodCap === hasReachedCap) {
-      return;
-    }
-
     await this.billingSubscriptionItemRepository.update(
       { id: billingSubscriptionItem.id },
       { hasReachedCurrentPeriodCap: hasReachedCap },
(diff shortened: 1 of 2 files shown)


packages/twenty-server/src/engine/core-modules/billing/services/billing-usage-cap.service.ts
```diff
@@ -110,8 +110,14 @@ export class BillingUsageCapService {
       );
     }
 
+    const [billingSubscriptionItem] = billingSubscriptionItems;
+
+    if (billingSubscriptionItem.hasReachedCurrentPeriodCap === hasReachedCap) {
+      return;
```
prastoin: ***Question:*** shouldn't we log ?

packages/twenty-server/src/engine/core-modules/billing/services/billing-usage.service.ts
```diff
@@ -317,7 +317,11 @@ export class BillingUsageService {
         -usedCredits,
       );
 
-    if (decrementedAvailableCredits <= 0) {
+    const previousAvailableCredits = decrementedAvailableCredits + usedCredits;
```
prastoin: ***Question:*** why not using `availableCredits` ?

prastoin (approved):

packages/twenty-server/src/engine/core-modules/billing/services/billing-usage-cap.service.ts
etiennejouan: Removed, you're right it should not happen

PR #20888, opened 2026-05-25
Set OpenAI Responses store false for AI chat and agents

mfamularopsyc:
## Summary

This PR sets `openai.store = false` for Twenty's `@ai-sdk/openai` AI calls.

This follows the approach discussed in #20877: instead of adding a new Twenty-specific Zero Data Retention config variable, OpenAI Responses calls no longer rely on OpenAI-stored response/item references. This should help Zero Data Retention organizations and may also avoid stale persisted-item replay errors for non-ZDR OpenAI users.

Changes included:

- Adds a shared OpenAI provider-options helper that merges `openai.store = false` for `@ai-sdk/openai` models.
- Applies the helper to AI chat `streamText` calls.
- Applies the helper to workflow/agent `generateText` calls.
- Preserves OpenAI encrypted reasoning metadata through DB/UI message mappers so reasoning context can be replayed without stored OpenAI item references.
- Does not add a new env/config variable.

Related to issue #20877.

## Behavior / Tradeoffs

This changes OpenAI Responses behavior for all Twenty OpenAI users, not only ZDR users.

The intended benefit is that Twenty no longer depends on OpenAI-stored response/item references. The main tradeoff is reduced provider-side item-reference reuse for non-ZDR OpenAI users.

To reduce the impact for reasoning models, this PR preserves `providerMetadata.openai.reasoningEncryptedContent` through message persistence/replay so reasoning context can still be provided without stored OpenAI item references.

## Tests

- Focused server Jest tests for OpenAI provider-options merging and reasoning metadata mapping.
- Focused frontend Jest test for reasoning metadata mapping.
- `oxlint` and `oxfmt --check` on changed files.
- `git diff --check`.


etiennejouan (commented):
Hi @mfamularopsyc, thanks for your contribution.  

Could you explain me why do you need to add provider in part ?

Concerning provider options, I would prefer to gather everything in same util. Can be done in the cache one (to rename then).

Final question, have you test with an openAI ZDR key ?

mfamularopsyc:
Hi @etiennejouan, thanks for the review.

The provider metadata on message parts is needed to preserve OpenAI-specific metadata returned by the AI SDK, especially on reasoning parts. With OpenAI Responses and `store: false`, we cannot rely on OpenAI-side stored response state between turns. The AI SDK can attach OpenAI metadata such as encrypted reasoning content / item ids to reasoning parts, so I added the mapping to round-trip that metadata through DB <-> UI message conversion instead of dropping it.

I also addressed your provider options comment: I moved the OpenAI `store: false` handling into the existing cache provider-options area and renamed it to a broader `provider-options.util.ts`, so cache options and OpenAI provider options are now gathered in the same util.


And yes, I tested with an OpenAI ZDR key.

  Direct Responses API checks:
  - `store: false` request succeeded and returned `"store": false`.
  - Retrieving that response ID returned `Response with id ... not found`.
  - Control request with `store: true` also returned `"store": false`.
  - Retrieving that response ID also returned `Response with id ... not found`.

  Twenty app-level check:
  - Ran Twenty locally with the ZDR key as `OPENAI_API_KEY`.
  - Triggered a native OpenAI AI chat/agent flow.
  - The request completed successfully and created a People record, with no ZDR/storage error.

image

mfamularopsyc:
Fixed in the latest commit by typing the provider options utility `sdkPackage` parameters as `AiSdkPackage` from `twenty-shared/ai` instead of `string`.

mfamularopsyc:
Resolved the merge conflicts with latest `main`.

The resolution keeps the new upstream provider options behavior while preserving this PR’s ZDR fix:
  - chat execution now passes `nativeBinding.providerOptions` through `getCallLevelProviderOptions(...)`
  - agent execution now merges `nativeBinding.providerOptions` with `getReasoningProviderOptions(...)`, then passes the result through `getCallLevelProviderOptions(...)`
  - `getCallLevelProviderOptions(...)` still applies OpenAI `store: false`
  - `sdkPackage` parameters are typed as `AiSdkPackage` from `twenty-shared/ai`

I also added/kept test coverage for merging existing provider options with call-level OpenAI options.

  Validation:
  - `provider-options.util.spec.ts`: 9 tests passed
  - Prettier check passed for the touched files
  - Retested Twenty locally with an OpenAI ZDR key; the AI chat/agent flow completed successfully with no ZDR/storage error

etiennejouan:
@mfamularopsyc trying to test but don't manage to setup an OpenAI ZDR key 🤔

mfamularopsyc:
> @mfamularopsyc trying to test but don't manage to setup an OpenAI ZDR key 🤔

I had to contact OpenAI directly (https://openai.com/contact-sales/), request the key, explain why I needed it (because we are a medical group), and fill out some paperwork to get it.

etiennejouan (dismissed):

etiennejouan (approved):

PR #20889, opened 2026-05-25
Refactor and centralize file mimeType integrity check and sanitization

prastoin:
# Introduction
closes https://github.com/twentyhq/private-issues/issues/484

This PR refactors the writeFile API to never expect to be passed a mimetype, its extract is done programmatically low level so any callers will pass through
Same for the file sanitization

## IANA override
Disclaimer for consistency we existing behavior we wanted to always have `application/typescript`
- should we rather consider fallbacking to octect-steam instead ?
- Any pulbic assets that has .ts will now also fallback to `application/typescript` instead of the official IANA

## Integration
Added coverage


packages/twenty-server/src/engine/core-modules/file-storage/utils/prepare-file-for-storage-or-throw.util.ts
```diff
@@ -0,0 +1,44 @@
...
+//
+// 1. Routes the source through extractFileInfoOrThrow (magic-byte sniffing,
+//    TWENTY_MIME_POLICY, then mrmime) to derive the authoritative mime and
+//    extension. Throws FileStorageException(INVALID_EXTENSION) when bytes
+//    contradict a binary path extension.
+// 2. Sanitizes the source content (currently DOMPurify on SVG; no-op for
+//    anything else). Sanitization is enforced here so callers cannot skip it.
+//
+// Returning both the sanitized source and the resolved mime keeps the
+// extract pass single-shot — FileStorageService consumes both directly.
```
etiennejouan: nit : do we need these comments ?

etiennejouan (approved):
LGTM 👏

packages/twenty-server/src/engine/core-modules/file-storage/utils/prepare-file-for-storage-or-throw.util.ts
prastoin: totally not

packages/twenty-server/src/engine/core-modules/file/utils/extract-file-info-or-throw.utils.ts
```diff
@@ -40,6 +41,15 @@ export const extractFileInfo = async ({
   let mimeType: string = 'application/octet-stream';
 
   if (isNonEmptyString(ext)) {
+    // Twenty policy wins over IANA/mrmime for the (small) set of extensions
```
Weiko: Does not seem to be the case? Why don't we do that check earlier then? (see L32 that can return earlier)

PR #20896, opened 2026-05-25
fix: separate initial timeline loading from fetchMore loading state

### LINKED ISSUE #20742: Bug: While fetching more timeline activities it always scrolls to the top.
## Bug Description
In the timeline activity inside the side panel, when we scroll down it fetches more data and it displays a skeleton and after fetching finishes the scroll position always jumps back to the top. Because of this, we have to scroll to the bottom again to load more data.

https://github.com/user-attachments/assets/571fda92-c372-4f5c-9f99-265187a1f1d3

## Expected behavior
Other activities like Emails does it bit differently. The skeleton would load only for initial data and for every subsequent fetches it would just show "Loading more..." text instead.

## Technical inputs
I have also noticed that for every activity it uses the same `<SkeletonLoader />` component which looks great for timeline but not as much for other activities. If we want I can create a seperate Skeleton component for each activity.

DeviSriSaiCharan:
Fixes: #20742 

# Issue:
In the timeline activity inside the side panel, when we scroll down it fetches more data and it displays a skeleton and after fetching finishes the scroll position always jumps back to the top. Because of this, we have to scroll to the bottom again to load more data.

https://github.com/user-attachments/assets/40d99df7-bdfb-4351-bc4f-baec2a035f13

## Root Cause
`useTimelineActivities` exposed a single `loading` state from `useFindManyRecords`, which became true for both:
- Initial timeline fetch
- Pagination / fetchMore requests

`loadingTimelineActivities` becomes true whenever a network request is triggered, including pagination requests where timeline records are already available.

Because of this, the UI could not distinguish between the first query loading state and subsequent fetchMore loading states.

## Fix
Added a separate firstQueryLoading state to detect only the first timeline request.

The first query is identified by checking:

- the request is still loading
- and no timeline activities have been loaded yet

Once activities are already available, any future loading state is treated as pagination/loading more records instead of initial loading.

This allows the UI to correctly handle:

- Skeleton loaders for first load
- Infinite scroll loaders for pagination
- Empty states after loading finishes

https://github.com/user-attachments/assets/48e8e078-82e2-43d8-823f-2f71e4f4f6e1


etiennejouan (approved):
Hi @DeviSriSaiCharan, thank you for your first contribution ! Looks good to me and works well 👏

PR #20902, opened 2026-05-26
chore(security): bump esbuild to ^0.28.0 to clear CVE-2025-68121 (CVSS 10.0)

charlesBochet:
## Summary

Follow-up to #20876. That PR bumped `esbuild` to `^0.27.3` to address the Go-stdlib CVEs the self-hoster reported, but only one of the two Go CVEs is actually fixed at that level. This PR closes the remaining gap.

### Why 0.27.3 wasn't enough

`esbuild` ships a Go-built binary inside the `@esbuild/<platform>` packages. The vulnerability lives in the bundled Go toolchain, not in any JavaScript. Verified by reading the Go `buildinfo` section from `node_modules/@esbuild/<platform>/bin/esbuild`:

- `esbuild@0.27.7` → built with **Go 1.23.8**
- `esbuild@0.28.0` → built with **Go 1.26.1**

CVE-2024-24790 (IPv6 zone parsing) is fixed in Go 1.21.11 / 1.22.4, so 0.27.x covers it.

**CVE-2025-68121** (crypto/tls cert validation bypass via TLS session resumption, **CVSS 10.0 / Critical** per [NVD](https://nvd.nist.gov/vuln/detail/cve-2025-68121)) is fixed only in Go 1.24.13, 1.25.7, and 1.26.0-rc.3+. Go 1.23.x is past Go's support window and will not receive this fix. So `esbuild@0.27.x` still ships a Go binary that Trivy correctly flags as vulnerable.

### Reachable risk in Twenty

Low. `esbuild` does not use `crypto/tls` at runtime — it reads files, parses, transforms, and writes. The vulnerable code path is dead code inside the binary, present but never executed. The scan finding is what we are clearing, not an exploitation risk.

### Fix

Bump `twenty-client-sdk`'s `esbuild` from `^0.27.3` to `^0.28.0` (resolves to 0.28.0, built with Go 1.26.1).

### Verification

Ran `yarn workspaces focus --production twenty twenty-server twenty-emails twenty-shared twenty-client-sdk` (the same install the Dockerfile uses) and confirmed:

- `node_modules/esbuild/` resolves to `esbuild@0.28.0` (single copy)
- The bundled `node_modules/@esbuild/<platform>/bin/esbuild` binary reports `go1.26.1` in its `buildinfo`

## Test plan

- [x] `nx build twenty-client-sdk` passes (esbuild's `build()` API is stable across 0.27 → 0.28)
- [x] Production focus install shows Go 1.26.1 in the shipped binary
- [ ] CI green
- [ ] Re-run Trivy against the resulting image; confirm CVE-2025-68121 no longer appears

### DIFF
--- a/packages/twenty-client-sdk/package.json
+++ b/packages/twenty-client-sdk/package.json
@@ -42,7 +42,7 @@
   "dependencies": {
     "@genql/cli": "^3.0.3",
     "@genql/runtime": "^2.10.0",
-    "esbuild": "^0.27.3",
+    "esbuild": "^0.28.0",
     "graphql": "^16.8.1"
   },
   "devDependencies": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -5945,9 +5945,9 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@esbuild/aix-ppc64@npm:0.27.7":
-  version: 0.27.7
-  resolution: "@esbuild/aix-ppc64@npm:0.27.7"
+"@esbuild/aix-ppc64@npm:0.28.0":
+  version: 0.28.0
+  resolution: "@esbuild/aix-ppc64@npm:0.28.0"
   conditions: os=aix & cpu=ppc64
   languageName: node
   linkType: hard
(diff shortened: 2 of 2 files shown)


Weiko (approved):
LGTM

PR #20904, opened 2026-05-26
feat(twenty-partners): partnerContent catalog + TFT import improvements

rashad:
## Summary

Two related threads for the internal `twenty-partners` app:

1. **Redesign `partnerQuote` → `partnerContent`.** The object was mis-modeled as a sales/pre-invoice doc (`amount`, opportunity link). In TFT it's actually a marketing-content catalog — customer quotes, case studies, partner quotes, logos — moving through a production lifecycle. This renames it in place and reshapes it to mirror TFT's `CustomerContent`.
2. **Import tooling improvements** to the TFT importer + multi-env workflow.

## Changes

**Schema (`partnerContent`)**
- Rename `partnerQuote` → `partnerContent` (object, view, nav, relation fields, identifiers).
- Add `contentType` MULTI_SELECT `[CUSTOMER_QUOTE, CASE_STUDY, PARTNER_QUOTE, LOGO]` and `interview` LINKS.
- Add `customerCompany` / `customerPerson` relations; keep `partner`; drop the `opportunity` link (TFT has none).
- Drop `amount`; rename the FILES field `quoteFile` → `documents` (`attachments` is a reserved morph-relation name).

**Importer (`import-from-tft.ts`)**
- Import the full content catalog (all types), not just `PARTNER_QUOTE`.
- Map TFT `partnerTimezone` → `region`, default `languagesSpoken=[ENGLISH]`, and set `deploymentExpertise=[SELF_HOST]` when scope includes `HOSTING_ENVIRONMENT`.
- Filter to partner-relevant records only: opportunities linked to a partner (20 of 164), content linked to a partner (10 of 22). Drops general sales-pipeline / customer-only noise.
- Dedupe companies by **normalized domain** (Twenty's unique key), not just name — fixes duplicate-entry crashes when the same company arrives under different names.
- Progress logging throughout.

**Tooling**
- `purge-soft-deleted` script (soft-deleted rows block re-imports via unique constraints).
- Multi-env script variants (`*:prod`) selected via `ENV_FILE`.

## Testing

Verified on a local Twenty instance and on `partner.twenty.com`:
- 122 partners, 20 partner-linked opportunities, 10 partner-linked content (all types), 229 domain-deduped companies.
- Schema confirmed via metadata introspection; `yarn twenty typecheck` clean.

## Notes

- Renaming an installed object isn't a pure in-place migration on a server that already had `partnerQuote` — the working path is `uninstall → deploy → install` (safe here: prod had no data).


packages/twenty-apps/internal/twenty-partners/package.json
```diff
@@ -16,8 +16,13 @@
     "test": "vitest run",
     "test:watch": "vitest",
     "seed": "tsx src/scripts/seed.ts",
+    "seed:prod": "ENV_FILE=.env.prod tsx src/scripts/seed.ts",
+    "purge": "tsx src/scripts/purge-soft-deleted.ts",
+    "purge:prod": "ENV_FILE=.env.prod tsx src/scripts/purge-soft-deleted.ts",
```
prastoin, line 21: ***Nitpick:*** Usually `:prod` command should be run on a built project which means targetting `js` files from `dist` `build` folder

packages/twenty-apps/internal/twenty-partners/src/scripts/purge-soft-deleted.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  for (const obj of OBJECTS) {
+    // destroy<Object>s is the bulk variant; capitalise + already-plural names.
+    const mutationName = `destroy${obj.charAt(0).toUpperCase()}${obj.slice(1)}`;
+    const data = await gql(url, key, `mutation { ${mutationName}(filter: { deletedAt: { is: NOT_NULL } }) { id } }`);
+    const destroyed = data[mutationName]?.length ?? 0;
+    console.log(`[purge] ${obj}: destroyed ${destroyed} soft-deleted`);
+  }
+
+  console.log('[purge] done');
```
prastoin, line 53: ***Question:*** Shouldn't this be a logic function ? Is the use case only local here ?

packages/twenty-apps/internal/twenty-partners/package.json
prastoin: if the related script was a logic function we could still run them from the CLI as from the UI using command menu action item for instance

prastoin (commented):
Hey there, sorry for my lack of context
Left few questions regarding the scripts nature
I'm not sure if the scriptings has been introduce as for dev only by choice or if they should have been logic functions ?

packages/twenty-apps/internal/twenty-partners/src/scripts/purge-soft-deleted.ts
rashad: No worries at all — happy to give the context!

These three (`seed`, `import-from-tft`, `purge-soft-deleted`) are intentionally dev/ops scripts for now. `import-from-tft` is really a one-shot ETL: it seeds this workspace from the live TFT data (`twentyfortwenty.twenty.com`), reading there with a separate API key and writing here. Since the source credentials are operator secrets that shouldn't live in the partners workspace runtime — and `purge` hard-destroys rows to clear soft-deleted ghosts that block re-imports, which I'd rather not expose as a one-click UI action — they fit better as CLI tooling an operator runs during setup. They also lean on dry-run + console preflight output, which is pretty CLI-shaped.

You're right that logic functions are the natural home for this longer term, though. It's on the partners backlog to explore exactly that: logic functions for a two-way sync between the TFT and partners workspaces, once this becomes recurring rather than a one-time seed.

packages/twenty-apps/internal/twenty-partners/package.json
rashad: Fair point on the usual `:prod`-runs-built-JS convention. In this case the scripts aren't part of the built app bundle — they're run via `tsx` on source by an operator (same as the local `seed`/`import:dryrun` variants), so there's no `dist/` artifact to target and the `:prod` variant just swaps the env file. If/when they graduate to logic functions (that two-way TFT↔partners sync we want to explore), they'd build and deploy with the app and become invokable from both the CLI and the command menu, exactly as you're describing.

prastoin (approved):
Looks good to me thanks for the explanation !

It could be risky to have a production API_KEY locally in corresponding .env, that's not catastrophic at all just to keep in mind
Also `.gitignore` is configured to prevent committing local .env production API KEY so RAS

```
# env files (can opt-in for committing if needed)
.env*
```

PR #20909, opened 2026-05-26
Fix app logo init

martmull:
- display application registration or application logo in all application related setting pages
- works with raw url and path to public asset application folder
- works will all publish types (private, public/npm, and sync)
- fix wrong type display in settings developer tab

## After
image
image


prastoin (commented):
Please resolve the cubic comments first 🙏 
CI also seems to be red

PR #20913, opened 2026-05-26
[Website] Generate release notes manifest at build time

mabdullahabaid:
Removed the releases page’s runtime dependency on `fs` and `process.cwd()` by introducing a build-time manifest generator: release notes still live as markdown under `src/content/releases`, but a new script now parses their frontmatter/content, validates that each note has a release, title, and preview image (and that the image actually exists), sorts the notes, and emits a typed `generated-release-notes.ts` file that the app imports at runtime. 

Updated the releases loader to return that generated data, changed the menu releases preview and release JSON-LD to use explicit typed fields (`title`, `previewImage`) instead of scraping markdown with regex at runtime, wired the generator into Nx so it runs automatically before `dev`, `build`, and `typecheck`, and fixed two stale image references in the release MDX files that the new validation exposed.


mabdullahabaid:
> Changes were made to package.json, but not to yarn.lock - Perhaps you need to run `yarn install`?

Only a script was added, no dependency changes.

packages/twenty-website/src/lib/releases/load-local-release-notes.ts
```diff
@@ -1,79 +1,6 @@
-import fs from 'fs';
-import path from 'path';
-
-import matter from 'gray-matter';
-
-import type { LocalReleaseNote } from '@/lib/releases/types';
-import { compareSemanticVersions } from '@/lib/releases/compare-semantic-versions';
```
FelixMalfait, line 7: is compareSemanticVersions still used?

PR #20915, opened 2026-05-26
Fix page layout widget tab moves

ehconitin:
Fixes widget moves between page layout tabs by making pageLayoutTabId part of the flat-entity diff, so the save mutation no longer silently drops the new tab assignment.

https://discord.com/channels/1130383047699738754/1508737039128985680


packages/twenty-server/src/engine/metadata-modules/flat-entity/constant/all-entity-properties-configuration-by-metadata-name.constant.ts
```diff
@@ -930,7 +930,7 @@ export const ALL_ENTITY_PROPERTIES_CONFIGURATION_BY_METADATA_NAME = {
       universalProperty: undefined,
     },
     pageLayoutTabId: {
-      toCompare: false,
+      toCompare: true,
       toStringify: false,
       universalProperty: 'pageLayoutTabUniversalIdentifier',
       isOverridable: true,
```
prastoin, line 936: We should add new validation that checks if the new page layout tab id exists, in the update validator

Weiko, line 936: Weird that isOverridable was true but toCompare was false 🤔 (we can probably improve typing for that)
Can you check if overrides work properly? (by moving a standard widget to another tab)

prastoin: Indeed nice catch I think it takes precedence

ehconitin: done! validator checks the effective tab id now (override-or-base), unit tests cover both paths + integration test covers the override happy path.

ehconitin: overrides does work properly!

Weiko (approved):
LGTM 👍

PR #20916, opened 2026-05-26
fix(billing) - fix orphaned stripe subs 2/2

etiennejouan:
Fixes https://sonarly.com/issue/40688

Should have been included in https://github.com/twentyhq/twenty/commit/0edd8d400c646cd6a40ff0fea5342a0f61645d5e


prastoin (commented):
As discussed sentry relica will be grabbed in a future other job run

prastoin (approved):

packages/twenty-server/src/engine/core-modules/user/services/user.service.ts
```diff
@@ -322,7 +322,8 @@ export class UserService extends TypeOrmQueryService<UserEntity> {
     const userWorkspaceId = userWorkspace.id;
 
     if (workspaceMembers.length === 1) {
-      await this.workspaceService.deleteWorkspace(workspaceId);
+      await this.workspaceService.suspendWorkspace(workspaceId);
```
Weiko, line 325: Do we want to run those operations in a transaction? Could be overkill for now, just something to keep in mind

Weiko (approved):
LGTM

PR #20919, opened 2026-05-26
Basic app logo fixes

martmull:
as title, took the quick win fixes from https://github.com/twentyhq/twenty/pull/20909/changes#diff-3367344412b2f44f0273d8019c1bc36396198244b9558d02921b135f62522baaR180 and leave the main fix for later as it requires an architectural update

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/application/application-development/application-development.resolver.ts
+++ b/packages/twenty-server/src/engine/core-modules/application/application-development/application-development.resolver.ts
@@ -303,11 +303,11 @@ export class ApplicationDevelopmentResolver {
         `${serverUrl}/public-assets/${workspaceId}/${applicationId}/${filePath}`,
     );
 
-    await this.applicationRegistrationService.updateFromManifest(
+    await this.applicationRegistrationService.updateFromManifest({
       applicationRegistrationId,
-      manifestWithResolvedUrls,
-      ApplicationRegistrationSourceType.LOCAL,
-    );
+      manifest: manifestWithResolvedUrls,
+      sourceType: ApplicationRegistrationSourceType.LOCAL,
+    });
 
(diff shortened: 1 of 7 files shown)


packages/twenty-server/src/engine/core-modules/application/application-development/application-development.resolver.ts
```diff
@@ -306,6 +306,7 @@ export class ApplicationDevelopmentResolver {
     await this.applicationRegistrationService.updateFromManifest(
       applicationRegistrationId,
       manifestWithResolvedUrls,
+      ApplicationRegistrationSourceType.LOCAL,
```
prastoin: ***Nitpick:*** record's arg ( hasn't been introduced in this PR but would gain in readability especially in this review context )

prastoin: everything that comes from the development resolver is LOCAL ?
I'm confused why the source type would change on a synchronization ?
I feel like we should have two code paths:
- Sync resulting in creation ( introduces SourceType )
- Sync resulting in update ( immutable )

martmull: you can start by deploying so sourceType will be tarball, then you sync so sourceType will be LOCAL

prastoin: From my understanding you should stick to the initial app installation funnel
If installed as a tarball from registry should keep being upgraded as is
If installed manually same
I might be missing something !

martmull: i think we should stick with latest installation mean. otherwise displayed information is false

prastoin (approved):

PR #20920, opened 2026-05-26
Fix/restore channel association scalar field metadata

neo773:
[#20836](https://github.com/twentyhq/twenty/pull/20836) dropped the channel objects but even though calendarChannelId/messageChannelId/messageFolderId already existed in compute standard flat field, there was never an upgrade command to readd them on the surviving association objects

so existing workspaces lost the field metadata (columns survived) and import workers throw 
```Error: Unknown error importing calendar events for calendar channel <REDACTED> in workspace <REDACTED>: Query runner already released. Cannot run queries anymore.```

This PR adds that command

### DIFF
--- a/packages/twenty-server/src/database/commands/upgrade-version-command/2-8/2-8-workspace-command-1798100020000-restore-channel-association-scalar-field-metadata.command.ts
+++ b/packages/twenty-server/src/database/commands/upgrade-version-command/2-8/2-8-workspace-command-1798100020000-restore-channel-association-scalar-field-metadata.command.ts
@@ -1,0 +1,259 @@
+import { Command } from 'nest-commander';
+
+import { InjectRepository } from '@nestjs/typeorm';
+import { FieldMetadataType } from 'twenty-shared/types';
+import { isDefined } from 'twenty-shared/utils';
+import { Repository } from 'typeorm';
+import { type QueryDeepPartialEntity } from 'typeorm/query-builder/QueryPartialEntity';
+import { v4 as uuidv4 } from 'uuid';
+
+import { ActiveOrSuspendedWorkspaceCommandRunner } from 'src/database/commands/command-runners/active-or-suspended-workspace.command-runner';
+import { WorkspaceIteratorService } from 'src/database/commands/command-runners/workspace-iterator.service';
+import { type RunOnWorkspaceArgs } from 'src/database/commands/command-runners/workspace.command-runner';
+import { ApplicationService } from 'src/engine/core-modules/application/application.service';
+import { RegisteredWorkspaceCommand } from 'src/engine/core-modules/upgrade/decorators/registered-workspace-command.decorator';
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by prastoin — APPROVED** [review]
Thanks !

File: packages/twenty-server/src/database/commands/upgrade-version-command/2-8/2-8-workspace-command-1798100020000-restore-channel-association-scalar-field-metadata.command.ts [hunk 1]
```diff
@@ -0,0 +1,282 @@
...
+  {
+    objectUniversalIdentifier: '20202020-a1b0-40b0-8ab0-5b6c7d8e9f0a',
+    fieldUniversalIdentifier: 'b3369d31-3856-4a7a-b007-ee353918127c',
+    fieldName: 'messageFolderId',
+    label: 'Message Folder',
+    description: 'Message Folder',
+    icon: 'IconFolder',
+    isNullable: false,
+  },
+];
```
prastoin, line 71: ***Nitpick:*** could have been derivated from the twenty standard app declaration

Lets not care as you've already tested locally on a eq reproduction

packages/twenty-server/src/database/commands/upgrade-version-command/2-8/2-8-workspace-command-1798100020000-restore-channel-association-scalar-field-metadata.command.ts
```diff
@@ -0,0 +1,282 @@
...
+      );
+
+    const fieldMetadataRowsWithApplication = fieldMetadataRowsToInsert.map(
+      (row) => ({
+        ...row,
+        applicationId: twentyStandardFlatApplication.id,
+      }),
+    );
+
+    await this.fieldMetadataRepository.insert(fieldMetadataRowsWithApplication);
```
prastoin, line 221: ***nitpick:*** we should use the validate build and run here

Lets not care as you've already tested locally on a eq reproduction

packages/twenty-server/src/database/commands/upgrade-version-command/2-8/2-8-workspace-command-1798100020000-restore-channel-association-scalar-field-metadata.command.ts
```diff
@@ -0,0 +1,282 @@
...
+
+    // The flat-maps / ORM entity metadata cache is not version-stamped, so bumping
+    // the metadata version alone leaves running workers serving stale entity metadata
+    // (missing the restored fields). Flush the whole workspace cache so they recompute
+    // from the updated metadata, then bump the version.
+    await this.workspaceCacheService.flush(
+      workspaceId,
+      Object.keys(WORKSPACE_CACHE_KEYS_V2) as WorkspaceCacheKeyName[],
```
prastoin: ***Blocker:*** Why would we need to flush the whole cache here ? and not only the field metadata related ones ?

PR #20921, opened 2026-05-26
Various fixes - some AI findings

Weiko:
App permissions tab:
- The fallback uuidv4() for a marketplace field was generated twice, so id and universalIdentifier could diverge; it's now computed once and reused as it seemed to be the intention (even though I don't really think it's a good idea)
- Renamed buildobjectMetadataItemsFromMarketplaceApp → buildObjectMetadataItemsFromMarketplaceApp to follow camelCase.

Morph relation validation:
- Fixed the user-facing message "At least one relation is require" → "...is required"
- Typos in the related test descriptions (Morh → Morph, samefield → same field) and their snapshots.

Docs
- The UUID field-type row in views.mdx only listed IS; updated to the full set supported by FILTER_OPERANDS_MAP (IS, IS_NOT, IS_EMPTY, IS_NOT_EMPTY).

### DIFF
--- a/packages/twenty-front/src/pages/settings/applications/tabs/SettingsApplicationPermissionsTab.tsx
+++ b/packages/twenty-front/src/pages/settings/applications/tabs/SettingsApplicationPermissionsTab.tsx
@@ -148,10 +148,11 @@ const buildFieldMetadataItemFromMarketplaceField = (
   field: ObjectFieldManifest,
 ): FieldMetadataItem => {
   const now = new Date().toISOString();
+  const universalIdentifier = field.universalIdentifier ?? uuidv4();
 
   return {
-    id: field.universalIdentifier ?? uuidv4(),
-    universalIdentifier: field.universalIdentifier ?? uuidv4(),
+    id: universalIdentifier,
+    universalIdentifier,
     name: field.name,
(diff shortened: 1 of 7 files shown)


packages/twenty-front/src/pages/settings/applications/tabs/SettingsApplicationPermissionsTab.tsx
```diff
@@ -148,10 +148,11 @@ const buildFieldMetadataItemFromMarketplaceField = (
   field: ObjectFieldManifest,
 ): FieldMetadataItem => {
   const now = new Date().toISOString();
+  const universalIdentifier = field.universalIdentifier ?? uuidv4();
 
   return {
-    id: field.universalIdentifier ?? uuidv4(),
-    universalIdentifier: field.universalIdentifier ?? uuidv4(),
+    id: universalIdentifier,
```
Weiko, line 154: Looks like we are still doing this pattern in the codebase, I don't really like having the same id for both and don't really see the point tbh. Does not seem to persist and is only frontend though 🤔

thomtrp (approved):

FelixMalfait (approved):

PR #20922, opened 2026-05-26
Fix workflow step output schema not reflecting user-renamed step name

### LINKED ISSUE #20906: use custom names in workflow step selection
In my workflow i have multiple search records steps, i do name them differently for easier understanding of what is happening in workflow. 
But in step section function all of them are just "search records", rather the names i've filled in and i basically have to guess which of the, is the record i'm looking for 

Image

thomtrp:
## Summary
- Fixes https://github.com/twentyhq/twenty/issues/20906
- Moves `markStepForRecomputation` from `useUpdateStep` into the lower-level `useUpdateWorkflowVersionStep` hook, so **every** caller that updates a step also triggers output schema recomputation.
- Previously, renaming a step via the side panel title input called `useUpdateWorkflowVersionStep` directly (bypassing `useUpdateStep`), so the variable picker kept showing the initial default name (e.g. "Create Record") instead of the user's custom name.

## Test plan
- [x] Rename a workflow step via the side panel title input
- [x] Verify the variable picker dropdown shows the updated name
- [x] Verify variable tags/chips in subsequent steps reflect the updated name
- [x] Verify that updating step settings (e.g. changing object type) still refreshes the output schema correctly

### DIFF
--- a/packages/twenty-front/src/modules/workflow/workflow-steps/hooks/__tests__/useUpdateStep.test.ts
+++ b/packages/twenty-front/src/modules/workflow/workflow-steps/hooks/__tests__/useUpdateStep.test.ts
@@ -1,10 +1,8 @@
 import { useUpdateStep } from '@/workflow/workflow-steps/hooks/useUpdateStep';
 import { act, renderHook } from '@testing-library/react';
-import { type WorkflowAction } from '~/generated/graphql';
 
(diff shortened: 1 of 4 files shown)


martmull (approved):

Weiko (approved):
LGTM

PR #20927, opened 2026-05-26
Fix cross version upgrade for 2.8

prastoin:


packages/twenty-server/src/engine/metadata-modules/flat-index-metadata/services/workspace-flat-index-map-cache.service.ts
```diff
@@ -52,6 +52,7 @@ export class WorkspaceFlatIndexMapCacheService extends WorkspaceCacheProvider<
               indexMetadataId: true,
               fieldMetadataId: true,
               order: true,
+              subFieldName: true,
```
Weiko, line 55: Not sure why we do a manual select here, seems we are listing all the columns or am I missing something? This could be error prone in the future

Weiko (approved):
LGTM, small comment

packages/twenty-server/src/engine/metadata-modules/flat-index-metadata/services/workspace-flat-index-map-cache.service.ts
prastoin: yep indeed

PR #20928, opened 2026-05-26
tt-fix-logic-function-error-handling

thomtrp:
## Summary
- **Fix `callWithTimeout` timer leak**: the `setTimeout` was never cleared when the callback resolved first, leaving orphaned timers (up to 15 minutes) in the Node.js event loop. Now uses `try/finally` with `clearTimeout`.
- **Properly classify timeout errors**: introduced `ExecutionTimedOutError` so the Lambda driver can throw `LOGIC_FUNCTION_EXECUTION_TIMEOUT` instead of `LOGIC_FUNCTION_EXECUTION_FAILED` — users now see "Function execution timed out" instead of the generic "Function execution failed."

## Test plan
- [x] Execute a logic function normally and verify it still works
- [x] Trigger a logic function timeout and verify the error message says "Function execution timed out" (not "Function execution failed")
- [x] Verify that a deleted logic function invocation logs a specific "was deleted" warning
- [x] Verify no orphaned timers remain after logic function execution completes


packages/twenty-server/src/engine/core-modules/logic-function/logic-function-drivers/utils/call-with-timeout.ts
```diff
@@ -1,14 +1,24 @@
...
 }): Promise<T> => {
-  return Promise.race([
-    callback(),
-    new Promise<T>((_, reject) =>
-      setTimeout(() => reject(new Error('Execution timed out')), timeoutMs),
-    ),
-  ]);
+  let timer: ReturnType<typeof setTimeout>;
+
+  const timeoutPromise = new Promise<T>((_, reject) => {
```
Weiko: explore AbortSignal.timeout?

Weiko (approved):
LGTM

martmull (approved):

PR #20933, opened 2026-05-26
Fix index cache computation

prastoin:
## Introduction
Should rely on custom typeorm entity loader layer that inspects the upgradeMigration that has bene run to dynamically request existing col only


Weiko (approved):
LGTM

PR #20935, opened 2026-05-26
fix(role): rebind API keys + agents before deleting their role

FelixMalfait:
## Customer-reported bug

A customer hit this when using the AI chat:

```json
{
  "message": "API key 760d4822-da40-4b3f-9031-40563d7ed6c9 has no role assigned",
  "extensions": {
    "code": "INTERNAL_SERVER_ERROR",
    "userFriendlyMessage": "This API key has no role assigned."
  }
}
```

Their integration authenticates via API key. Somewhere along the way, the role bound to that API key was deleted, leaving the API key authenticated but role-less. Any request that hits a permission check (`getRoleIdForApiKeyId`) blows up.

## Root cause

In `RoleService.deleteManyRoles`, the pre-deletion cleanup (`assignDefaultRoleToMembersWithRoleToDelete`) only rebinds **user workspaces** to the workspace default role. API keys and agents pointing at the role are ignored. Because `RoleTargetEntity.role` declares `onDelete: 'CASCADE'`, the FK then drops the role_target rows for those API keys / agents — but the API keys themselves stay in `api_key`, now orphaned in `apiKeyRoleMap`.

A previous read-side workaround ([2767ddac44](https://github.com/twentyhq/twenty/commit/2767ddac44) — make the `role` ResolveField nullable) handled the API-key-details page, but did not address the write paths (`getRoleIdForApiKeyId`).

## Fix

- Rename `assignDefaultRoleToMembersWithRoleToDelete` → `rebindTargetsOfRoleToDeleteToDefaultRole` and extend it to rebind API keys (via `ApiKeyRoleService.assignRoleToApiKey`) and agents (via `AiAgentRoleService.assignRoleToAgent`) in the same step, before the role is deleted.
- If the workspace default role doesn't satisfy `canBeAssignedToApiKeys` / `canBeAssignedToAgents`, the inner `assignRoleTo*` validation throws. We catch that and rethrow as a `PermissionsException` with a role-deletion-context message and two new codes — `ROLE_CANNOT_BE_ASSIGNED_TO_API_KEYS` / `ROLE_CANNOT_BE_ASSIGNED_TO_AGENTS` — so the admin sees a clear "reassign these first" prompt rather than a confusing inner error.

## Scope / non-goals

- **Already-orphaned API keys are not auto-healed.** The customer still needs to reassign a role to their existing orphan API key via the UI (Settings > API Keys > [the key] > role). A separate cleanup command for existing orphans is a follow-up.
- I did not investigate *why* the customer's session was authenticated via API key in the AI chat — that may be their integration setup. Worth confirming with them separately.

## Test plan

- [ ] Workspace with default role `Admin` (which has `canBeAssignedToApiKeys: true`): create an API key with a custom role, delete the custom role → API key is rebound to Admin, requests keep working.
- [ ] Workspace with default role `Member` (default, has `canBeAssignedToApiKeys: false`): create an API key with a custom role, delete the custom role → role deletion fails with the new `ROLE_CANNOT_BE_ASSIGNED_TO_API_KEYS` error explaining the admin must reassign first. API key + custom role are both unchanged.
- [ ] Same two scenarios for agents (`canBeAssignedToAgents`).
- [ ] Existing user-workspace rebind behavior is unchanged.
- [ ] Role deletion with no dependent API keys / agents still works.

### DIFF
--- a/packages/twenty-server/src/engine/metadata-modules/role/role.service.ts
+++ b/packages/twenty-server/src/engine/metadata-modules/role/role.service.ts
@@ -461,11 +461,10 @@ export class RoleService {
           error instanceof ApiKeyException &&
           error.code === ApiKeyExceptionCode.ROLE_CANNOT_BE_ASSIGNED_TO_API_KEYS
         ) {
-          throw this.toRoleDeleteRebindException({
+          throw this.toApiKeyRebindException({
             error,
             roleLabel,
-            targetKind: 'apiKey',
-            targetCount: apiKeysToRebind.length,
+            apiKeyCount: apiKeysToRebind.length,
           });
         }
         throw error;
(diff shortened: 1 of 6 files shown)


packages/twenty-server/src/engine/metadata-modules/role/role.service.ts
```diff
@@ -406,13 +411,19 @@ export class RoleService {
     });
   }
 
+  // Rebinds role_targets (users, API keys, agents) to the workspace default
```
FelixMalfait: Remove this comment

FelixMalfait: Done in 76dfed9bee — both comment blocks removed (the one here on `rebindTargetsOfRoleToDeleteToDefaultRole` and the matching one on `toRoleDeleteRebindException` below).

packages/twenty-server/src/engine/metadata-modules/role/role.service.ts
```diff
@@ -427,5 +438,82 @@ export class RoleService {
...
+          workspaceId,
+        });
+      } catch (error) {
+        throw this.toRoleDeleteRebindException({
+          error,
+          roleLabel,
+          targetKind: 'apiKey',
+          targetCount: apiKeysToRebind.length,
+        });
+      }
```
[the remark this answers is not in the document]
FelixMalfait: Good catch — fixed in 76dfed9bee. The catch blocks now narrow to `ApiKeyException` with `ROLE_CANNOT_BE_ASSIGNED_TO_API_KEYS` and `AiException` with `ROLE_CANNOT_BE_ASSIGNED_TO_AGENTS`; any other failure (migration builder errors, DB timeouts, etc.) propagates unchanged.

packages/twenty-server/src/engine/metadata-modules/role/role.service.ts
```diff
@@ -427,5 +442,91 @@ export class RoleService {
...
+
+    const apiKeysToRebind =
+      await this.apiKeyRoleService.getApiKeysAssignedToRole(
+        roleId,
+        workspaceId,
+      );
+
+    for (const apiKey of apiKeysToRebind) {
+      try {
+        await this.apiKeyRoleService.assignRoleToApiKey({
```
Weiko, line 454: Should we batch?

Weiko (approved):
LGTM

packages/twenty-server/src/engine/metadata-modules/role/role.service.ts
FelixMalfait: Not worth it imo

packages/twenty-server/src/engine/metadata-modules/role/role.service.ts
```diff
@@ -427,5 +442,91 @@ export class RoleService {
...
+    error: unknown;
+    roleLabel: string;
+    targetKind: 'apiKey' | 'agent';
+    targetCount: number;
+  }): Error {
+    const targetLabel = targetKind === 'apiKey' ? 'API key' : 'agent';
+    const innerMessage = error instanceof Error ? error.message : String(error);
+
+    return new PermissionsException(
+      `Cannot delete role "${roleLabel}": ${targetCount} ${targetLabel}(s) depend on it and the workspace default role cannot take them over (${innerMessage}). Reassign these ${targetLabel}(s) to another role first.`,
```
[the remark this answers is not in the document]
FelixMalfait: Fixed in 71e1d8b693. Dropped the `(${innerMessage})` interpolation; the developer-facing message is now just `Cannot delete role "<label>": the workspace default role cannot be assigned to <apiKey|agent>s.`, and the client copy comes from `userFriendlyMessage`.

PR #20947, opened 2026-05-27
fix(website) - fix getPartners

etiennejouan:
**Context**
Fix empty marketplace at /partners/list: page was being statically prerendered with getPartners() returning [] (build-time fetch failure), then served from OpenNext's R2 cache forever — so the partners API was never actually called in production.

**Change**
Wrap getPartners in unstable_cache with revalidate: 300, matching the existing fetchGithubStarCount / fetchDiscordMemberCount ISR pattern. After deploy, the cached empty result expires within 5 minutes and the worker refetches from the partners API at runtime (where the env vars actually exist), populating the page. Also drops the now-redundant cache: 'no-store' from partnersApiFetch.


rashad (approved):
works locally lt's deploy to see if that fixes the issue

Issue #20942, opened 2026-05-27
Remove custom file field maximum files limit and improve UX on attached files creation / edition / deletion


n2ojim:
## Scope & Context

We use Twenty CRM to manage records that require attaching multiple files (photos, documents, scans, media, etc.) to individual records via custom FILES fields. Currently, Twenty enforces a hard limit of maximum 10 attached files per FILES field, which is far too restrictive for many business use cases across industries (property management, construction, legal, medical, event planning, etc.).

Additionally, the current file attachment UX suffers from critical usability issues: files are displayed by their random internal File IDs instead of human-readable names, renaming is cumbersome, and there are no quick actions (inline delete, clear gallery view). This severely limits the practical usage of the FILES field type and prevents building useful workflows on top of attached files (e.g., image processing, document generation).

## Current behavior

**1. FILES field hard limit at 10**

In Settings > Data Model, when editing a custom FILES field and trying to set "Max values" above 10, the UI rejects the input with the validation error:
```
maxNumberOfValues must be defined in settings and be a number greater than 0 and less than or equal to 10
```

<img width="1083" height="299" alt="Image" src="https://github.com/user-attachments/assets/3dbab1db-e4a4-4c49-8cc2-594ff7e63271" />

**2. File IDs displayed instead of names**

In the record detail view, custom FILES fields display attached files using their internal File UUIDs (e.g., `f47ac10b-58cc-4372-a567-0e02b2c3d479`) instead of the original uploaded file name (e.g., `contract-final.pdf`, `kitchen-photo.jpg`). These UUIDs are meaningless to end users and make the search/filter box in the field completely useless, as no user can search for or recognize a random ID.

<img width="375" height="247" alt="Image" src="https://github.com/user-attachments/assets/3ec5d7d5-2060-433e-8626-976d7c8e7da4" />

**3. No inline delete action**

There is no quick way to remove an attached file directly from the custom field. Users must navigate elsewhere or use a multi-step process to detach a file. This slows down day-to-day operations significantly when managing galleries or document sets. Without any preview, we cannot recognize what file is with the UUID only.

<img width="318" height="224" alt="Image" src="https://github.com/user-attachments/assets/f0004251-76e5-41a1-844f-a7babd7c6e4b" />

**4. Cramped, text-only file display**

Files are rendered in a very compact, plain-text list format that provides no visual overview. For image files, this makes it impossible to quickly identify which file is which without clicking through each one individually.

[...]

PR #20950, opened 2026-05-27
feat: raise FILES field max number of values from 10 to 60

FelixMalfait:
## Summary

Raises the artificial hardcoded ceiling on `maxNumberOfValues` for custom FILES fields from `10` to `60` so users can attach more files per record.

- Bumped `FILES_FIELD_MAX_NUMBER_OF_VALUES` constant in `twenty-shared` from `10` to `60`
- Updated validator unit test (inline snapshots + "exceeds max" case)
- Updated create/update files-field metadata integration tests and Jest snapshots

The frontend Zod schema only enforces a `min`, so no frontend changes are required — the backend constant is the single source of truth for the upper bound.

Refs #20942

🤖 Generated with [Claude Code](https://claude.ai/code)


etiennejouan (approved):
LGTM

PR #20953, opened 2026-05-27
feat(twenty-orm): introduce WorkspaceScopedRepository for core/metadata workspace-scoped entities

FelixMalfait:
## Summary

Adds a third tenancy enforcement layer for entities that live in shared schemas (`core`, `metadata`) and carry a `workspaceId` column — previously the only safeguard at this layer was developer discipline (remembering to put `workspaceId` in every WHERE clause).

### The three layers, after this PR

| Layer | Scope | How it's enforced |
|---|---|---|
| 1. Workspace data | per-workspace schema (companies, people, custom objects) | `twentyORMManager.getRepository(workspace, E)` — physical isolation (own data source) |
| 2. Metadata | shared `metadata` schema (objectMetadata, fieldMetadata, views, roles…) | Flat-entity-maps cache — workspace-scoped in-memory map, lookups by id within it |
| 3. Core (new) | shared `core` schema (agent threads/turns/messages, app tokens, etc.) | `WorkspaceScopedRepository<T>` — `workspaceId` is a required positional argument on every read/write |

### The wrapper (`packages/twenty-server/src/engine/twenty-orm/workspace-scoped-repository/`)
- `WorkspaceScopedRepository<T extends WorkspaceScopedEntity>` — wraps a TypeORM `Repository<T>`, requires `workspaceId` on every `find`/`findOne`/`findOneOrFail`/`update`/`delete`/`softDelete`/`insert`/`save`/`count` call, merging it into the WHERE or stamping it on the entity. `createQueryBuilder` is an explicit escape hatch (caller scopes manually).
- Provided via Nest DI with `@InjectWorkspaceScopedRepository(EntityClass)` and the `provideWorkspaceScopedRepository(EntityClass)` provider factory.
- 19 unit tests cover the merge behavior, override-on-conflict, and the array-where (OR) case.

### Lint enforcement (`packages/twenty-oxlint-rules/rules/prefer-workspace-scoped-repository.ts`)
- New `twenty/prefer-workspace-scoped-repository` rule (level: **error**).
- Blacklist of entity names: raw `@InjectRepository(E)` is rejected if `E` is on the list.
- Initial list: `AgentTurnEntity`, `AgentMessageEntity`, `AgentMessagePartEntity`, `AgentChatThreadEntity`, `AgentTurnEvaluationEntity`, `AgentEntity`.
- Designed to grow over time as more consumers are migrated.
- 5 rule tests.

### Migration in this PR
All consumers of the six blacklisted entities, including:
- AI agent / chat / monitor resolvers, services, and jobs
- `AgentService`, `AiAgentRoleService`, `AiAgentWorkflowAction`, `ApplicationService`, `WorkspaceFlatAgentMapCacheService`
- Admin-panel chat (migrated where the lookup is workspace-known; one documented `eslint-disable` on the threadId-discovery lookup that necessarily precedes the `allowImpersonation` permission check)
- `AiAgentRoleService` unit spec updated to mock the scoped wrapper

## Future work (deliberately not in this PR)

A standalone audit identified ~14 additional `core`/`metadata` entities with `workspaceId` that currently use raw `@InjectRepository` and could be added to the blacklist. Notable candidates: `UserWorkspaceEntity` (42 sites), `AppTokenEntity` (10), `FileEntity` (7), `BillingCustomerEntity`/`BillingSubscriptionEntity` (~22 combined). Each should be its own PR — the migration is mechanical but the surface is wide.

## Test plan
- [x] `npx nx lint twenty-server` — 0 warnings, 0 errors
- [x] `npx jest workspace-scoped-repository` — 19/19 pass
- [x] `npx nx test twenty-oxlint-rules` — 215/215 pass
- [x] `npx jest src/engine/metadata-modules/ai` — 44/44 pass
- [ ] Manual smoke: end-to-end AI agent chat send/receive (reviewer)
- [ ] Manual smoke: AI agent monitor — list turns, run evaluation (reviewer)
- [ ] Manual smoke: admin-panel chat thread inspection (reviewer)

### DIFF
--- a/packages/twenty-oxlint-rules/rules/prefer-workspace-scoped-repository.ts
+++ b/packages/twenty-oxlint-rules/rules/prefer-workspace-scoped-repository.ts
@@ -2,20 +2,15 @@ import { defineRule } from '@oxlint/plugins';
 
 export const RULE_NAME = 'prefer-workspace-scoped-repository';
 
-// Entities allowed to be injected via raw @InjectRepository. Everything
-// else matching @InjectRepository(*Entity) is required to go through
-// @InjectWorkspaceScopedRepository so workspaceId is enforced on every
-// read/write. New entities default to enforcement: add to this set only
-// when raw access is genuinely the right shape (no workspaceId column,
-// pivot/global table, or workspaceId-discovered-from-the-row patterns).
-const EXCLUSIONS = new Set<string>([
-  // Workspace itself / pivot / nullable workspaceId by design
+// Entities that legitimately do not fit the workspace-scoped wrapper:
+// the workspace itself, pivots, nullable-workspaceId rows, and global
+// tables with no workspaceId column.
+const STRUCTURAL_EXEMPTIONS = new Set<string>([
   'WorkspaceEntity',
   'UserWorkspaceEntity',
   'AppTokenEntity',
   'ApplicationRegistrationEntity',
 
-  // Global tables with no workspaceId column
   'ApplicationVariableEntity',
   'BillingMeterEntity',
   'BillingPriceEntity',
@@ -30,10 +25,14 @@ const EXCLUSIONS = new Set<string>([
   'SigningKeyEntity',
   'UserEntity',
   'WorkspaceSSOIdentityProviderEntity',
+]);
 
-  // Workspace-scoped entities currently outside the wrapper. Removing an
-  // entry forces every raw @InjectRepository site for it to migrate to
-  // the wrapper or carry an explicit eslint-disable.
+// Workspace-scoped entities still injected raw at one or more call
+// sites. Routed here to keep the rule enforceable without forcing a
+// repo-wide migration in a single change. Removing an entry forces
+// every raw @InjectRepository site for it to switch to the wrapper or
+// carry an explicit eslint-disable.
+const WORKSPACE_SCOPED_EXEMPTIONS = new Set<string>([
   'ApplicationEntity',
   'ApplicationRegistrationVariableEntity',
   'CalendarChannelEntity',
@@ -65,6 +64,15 @@ const EXCLUSIONS = new Set<string>([
   'ViewGroupEntity',
   'ViewSortEntity',
   'WebhookEntity',
+]);
+
+// Entities allowed to be injected via raw @InjectRepository. Every
+// other @InjectRepository(*Entity) site must go through
+// @InjectWorkspaceScopedRepository so workspaceId is enforced on every
+// read/write. New entities default to enforcement.
+const EXCLUSIONS = new Set<string>([
+  ...STRUCTURAL_EXEMPTIONS,
+  ...WORKSPACE_SCOPED_EXEMPTIONS,
 ]);
 
 const matchInjectRepositoryEntity = (decorator: any): string | null => {
--- a/packages/twenty-oxlint-rules/rules/prefer-workspace-scoped-repository.ts
+++ b/packages/twenty-oxlint-rules/rules/prefer-workspace-scoped-repository.ts
@@ -14,6 +14,12 @@ const BLACKLIST = new Set<string>([
   'AgentChatThreadEntity',
   'AgentTurnEvaluationEntity',
   'AgentEntity',
+  'TwoFactorAuthenticationMethodEntity',
+  'ApiKeyEntity',
+  'FeatureFlagEntity',
+  'ApprovedAccessDomainEntity',
+  'EmailingDomainEntity',
+  'PublicDomainEntity',
 ]);
 
 const isInjectRepositoryDecoratorForBlacklistedEntity = (
--- a/packages/twenty-oxlint-rules/rules/prefer-workspace-scoped-repository.ts
+++ b/packages/twenty-oxlint-rules/rules/prefer-workspace-scoped-repository.ts
@@ -2,9 +2,7 @@ import { defineRule } from '@oxlint/plugins';
 
 export const RULE_NAME = 'prefer-workspace-scoped-repository';
 
-// Entities that legitimately do not fit the workspace-scoped wrapper:
-// the workspace itself, pivots, nullable-workspaceId rows, and global
-// tables with no workspaceId column.
+// Entities that legitimately do not carry a workspaceId column.
 const STRUCTURAL_EXEMPTIONS = new Set<string>([
   'WorkspaceEntity',
   'UserWorkspaceEntity',
@@ -27,11 +25,7 @@ const STRUCTURAL_EXEMPTIONS = new Set<string>([
   'WorkspaceSSOIdentityProviderEntity',
 ]);
 
-// Workspace-scoped entities still injected raw at one or more call
-// sites. Routed here to keep the rule enforceable without forcing a
-// repo-wide migration in a single change. Removing an entry forces
-// every raw @InjectRepository site for it to switch to the wrapper or
-// carry an explicit eslint-disable.
+// Workspace-scoped entities exempted from the wrapper at the call site.
 const WORKSPACE_SCOPED_EXEMPTIONS = new Set<string>([
   'ApplicationEntity',
   'ApplicationRegistrationVariableEntity',
@@ -66,10 +60,7 @@ const WORKSPACE_SCOPED_EXEMPTIONS = new Set<string>([
   'WebhookEntity',
 ]);
 
-// Entities allowed to be injected via raw @InjectRepository. Every
-// other @InjectRepository(*Entity) site must go through
-// @InjectWorkspaceScopedRepository so workspaceId is enforced on every
-// read/write. New entities default to enforcement.
+// Everything else must use @InjectWorkspaceScopedRepository.
 const EXCLUSIONS = new Set<string>([
   ...STRUCTURAL_EXEMPTIONS,
   ...WORKSPACE_SCOPED_EXEMPTIONS,
@@ -108,12 +99,12 @@ export const rule = defineRule({
     type: 'problem',
     docs: {
       description:
-        'Disallow raw @InjectRepository for workspace-scoped entities. Entities carrying a workspaceId must be injected via @InjectWorkspaceScopedRepository so the workspaceId guard is enforced on every read/write. Entities that do not fit the wrapper are listed in the EXCLUSIONS set in this rule.',
+        'Disallow raw @InjectRepository for workspace-scoped entities. Use @InjectWorkspaceScopedRepository so workspaceId is enforced on every read/write.',
     },
     schema: [],
     messages: {
       preferWorkspaceScopedRepository:
-        "{{entityName}} must be injected via @InjectWorkspaceScopedRepository({{entityName}}) returning WorkspaceScopedRepository<{{entityName}}>. Raw @InjectRepository bypasses the workspaceId guard that prevents cross-workspace data access. If {{entityName}} genuinely does not fit (no workspaceId column, cross-tenant lookup, etc.), either add it to the EXCLUSIONS set in this rule or suppress with `// eslint-disable-next-line twenty/prefer-workspace-scoped-repository` and a short reason.",
+        'Use @InjectWorkspaceScopedRepository({{entityName}}) instead of raw @InjectRepository so the workspaceId guard is enforced. If {{entityName}} genuinely does not fit, add it to EXCLUSIONS or suppress with `// eslint-disable-next-line twenty/prefer-workspace-scoped-repository` and a short reason.',
     },
   },
   create: (context) => {
@@ -124,10 +115,6 @@ export const rule = defineRule({
         }
 
         for (const param of node.value.params ?? []) {
-          // Parameter decorators live on `.decorators` for both
-          // TSParameterProperty (e.g. `private readonly foo`) and plain
-          // parameters (e.g. `foo: Foo`). Checking both covers shorthand
-          // and explicit-assignment constructor forms.
           for (const decorator of param.decorators ?? []) {
             const entityName = matchInjectRepositoryEntity(decorator);
 
--- a/packages/twenty-oxlint-rules/rules/prefer-workspace-scoped-repository.ts
+++ b/packages/twenty-oxlint-rules/rules/prefer-workspace-scoped-repository.ts
@@ -2,56 +2,71 @@ import { defineRule } from '@oxlint/plugins';
 
 export const RULE_NAME = 'prefer-workspace-scoped-repository';
 
-// Entities that MUST be accessed through WorkspaceScopedRepository
-// (@InjectWorkspaceScopedRepository) — using the raw TypeORM repository
-// (@InjectRepository) on any of these is a lint error.
-//
-// Each entry is a core- or metadata-schema entity carrying a
-// workspaceId column. The list expresses intent: every such entity
-// belongs here. Entities still in active migration keep their
-// existing raw call sites suppressed with `// eslint-disable-next-line
-// twenty/prefer-workspace-scoped-repository` plus a
-// `TODO(workspace-scoped)` note, so the blacklist captures the
-// eventual goal rather than just what's already been refactored.
-//
-// Note on naming: "blacklist" here means "entities forbidden from raw
-// @InjectRepository". It is *not* a list of entities the rule opts
-// into checking — the rule's enforcement scope is the entire codebase,
-// and this set is what triggers the error.
-const BLACKLIST = new Set<string>([
-  'AgentTurnEntity',
-  'AgentMessageEntity',
-  'AgentMessagePartEntity',
-  'AgentChatThreadEntity',
-  'AgentTurnEvaluationEntity',
-  'AgentEntity',
-  'TwoFactorAuthenticationMethodEntity',
-  'ApiKeyEntity',
-  'FeatureFlagEntity',
-  'ApprovedAccessDomainEntity',
-  'EmailingDomainEntity',
-  'PublicDomainEntity',
-  'FileEntity',
-  'BillingCustomerEntity',
-  'BillingSubscriptionEntity',
-  'BillingEntitlementEntity',
+// Entities allowed to be injected via raw @InjectRepository. Everything
+// else matching @InjectRepository(*Entity) is required to go through
+// @InjectWorkspaceScopedRepository so workspaceId is enforced on every
+// read/write. New entities default to enforcement: add to this set only
+// when raw access is genuinely the right shape (no workspaceId column,
+// pivot/global table, or workspaceId-discovered-from-the-row patterns).
+const EXCLUSIONS = new Set<string>([
+  // Workspace itself / pivot / nullable workspaceId by design
+  'WorkspaceEntity',
+  'UserWorkspaceEntity',
+  'AppTokenEntity',
+  'ApplicationRegistrationEntity',
+
+  // Global tables with no workspaceId column
+  'ApplicationVariableEntity',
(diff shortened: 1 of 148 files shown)


packages/twenty-oxlint-rules/rules/prefer-workspace-scoped-repository.ts
```diff
@@ -0,0 +1,120 @@
...
+//
+// Each entry is a core- or metadata-schema entity carrying a
+// workspaceId column. The list expresses intent: every such entity
+// belongs here. Entities still in active migration keep their
+// existing raw call sites suppressed with `// eslint-disable-next-line
+// twenty/prefer-workspace-scoped-repository` plus a
+// `TODO(workspace-scoped)` note, so the blacklist captures the
+// eventual goal rather than just what's already been refactored.
+//
+// Note on naming: "blacklist" here means "entities forbidden from raw
```
FelixMalfait: What if people add new entiies in the future? They won't think about updating this rule. That's why I said "blacklist", it meant "every ***Entity except eventually those in that list". Please update the logic

packages/twenty-server/src/engine/core-modules/admin-panel/services/admin-panel-chat.service.ts
```diff
@@ -7,18 +7,30 @@ import { type AdminChatMessageDTO } from 'src/engine/core-modules/admin-panel/dt
...
+  InjectWorkspaceScopedRepository,
+  WorkspaceScopedRepository,
+} from 'src/engine/twenty-orm/workspace-scoped-repository/workspace-scoped-repository';
 
 @Injectable()
 export class AdminPanelChatService {
   constructor(
     @InjectRepository(WorkspaceEntity)
     private readonly workspaceRepository: Repository<WorkspaceEntity>,
+    // Admin support flow: the only reads here are for an authenticated
```
FelixMalfait: Shorter comment to justify the eslint-disable

packages/twenty-server/src/engine/core-modules/approved-access-domain/services/approved-access-domain.service.ts
```diff
@@ -198,8 +208,10 @@ export class ApprovedAccessDomainService {
       );
     }
 
+    // Cross-workspace lookup by id is gated by the signed validation token:
```
FelixMalfait: Remove comment

packages/twenty-server/src/engine/core-modules/approved-access-domain/services/approved-access-domain.service.ts
```diff
@@ -281,30 +291,29 @@ export class ApprovedAccessDomainService {
...
-      },
-    });
+    return this.approvedAccessDomainRepository.find(workspace.id);
   }
 
   async findValidatedApprovedAccessDomainWithWorkspacesAndSSOIdentityProvidersDomain(
     domain: string,
   ) {
-    return await this.approvedAccessDomainRepository.find({
+    // Cross-workspace SSO discovery: given a user's email domain at sign-in
```
FelixMalfait: Remove comment

packages/twenty-server/src/engine/core-modules/billing/services/billing-portal.workspace-service.ts
```diff
@@ -38,10 +42,15 @@ export class BillingPortalWorkspaceService {
     private readonly stripeBillingPortalService: StripeBillingPortalService,
     private readonly workspaceDomainsService: WorkspaceDomainsService,
     private readonly billingSubscriptionService: BillingSubscriptionService,
+    // Only used in computeBillingPortalSessionURLOrThrow which already
+    // scopes by workspace.id and a status filter — could be migrated
+    // independently when that method is split out. Keeping raw here to
+    // limit blast radius of this commit.
+    // eslint-disable-next-line twenty/prefer-workspace-scoped-repository
```
FelixMalfait: Please migrate it and remove this comment

packages/twenty-server/src/engine/core-modules/billing/services/billing-subscription-update.service.ts
```diff
@@ -57,6 +57,10 @@ export class BillingSubscriptionUpdateService {
     private readonly billingPriceRepository: Repository<BillingPriceEntity>,
     @InjectRepository(BillingSubscriptionItemEntity)
     private readonly billingSubscriptionItemRepository: Repository<BillingSubscriptionItemEntity>,
+    // Lookup is by subscription id only; threading workspaceId through
+    // updateSubscription is a larger refactor since callers come from
+    // multiple service layers.
```
FelixMalfait: Solve this cleanly and remove this comment

packages/twenty-server/src/engine/core-modules/billing/services/billing-subscription.service.ts
```diff
@@ -44,17 +48,23 @@ export class BillingSubscriptionService {
...
     private readonly billingPlanService: BillingPlanService,
-    @InjectRepository(BillingEntitlementEntity)
-    private readonly billingEntitlementRepository: Repository<BillingEntitlementEntity>,
+    @InjectWorkspaceScopedRepository(BillingEntitlementEntity)
+    private readonly billingEntitlementRepository: WorkspaceScopedRepository<BillingEntitlementEntity>,
+    // Mixed use: some calls scope by workspaceId, but
+    // getCurrentBillingSubscription accepts either workspaceId or
+    // stripeCustomerId in its criteria, and the conflict-upsert keys
+    // on stripeSubscriptionId. Keeping raw to avoid splitting the
+    // public surface in this PR; future cleanup could dual-inject.
```
FelixMalfait: Let's dual inject, we want this PR to be cleaned. Remove ANY comment in this PR that refers to the current process/changes or to things left to do (do then). Comments should refer to an absolute sense and makes sense to someone outside of the context of this PR. They should also be kept short. Review ALL comments in this PR.

packages/twenty-server/src/engine/core-modules/file-storage/file-storage.service.ts
```diff
@@ -138,8 +142,14 @@ export class FileStorageService {
     const applicationRepository = queryRunner
       ? queryRunner.manager.getRepository(ApplicationEntity)
       : this.applicationRepository;
+    // When called within a transaction we need the QueryRunner-bound
+    // FileEntity repository instead of the injected one; wrap it with
+    // the same workspace-scoped facade so the call sites below stay
+    // identical.
```
FelixMalfait: Remove comment

FelixMalfait: No memory leak  / aren't we creating more new repositiories than before here?

packages/twenty-server/src/engine/twenty-orm/workspace-scoped-repository/workspace-scoped-repository.ts
```diff
@@ -0,0 +1,206 @@
...
+// The constraint is enforced both at compile time (the generic bound below)
+// and at runtime (every read/write merges workspaceId into the criteria).
+export type WorkspaceScopedEntity = ObjectLiteral & { workspaceId: string };
+
+// Wraps a TypeORM Repository so that every operation is scoped to a
+// caller-supplied workspaceId. Built to make cross-workspace IDOR bugs
+// impossible-by-default for `core` and `metadata` schema entities that
+// carry a workspaceId column. Workspace-data schema entities should keep
+// using WorkspaceRepository / twentyORMManager.
+export class WorkspaceScopedRepository<T extends WorkspaceScopedEntity> {
```
FelixMalfait, line 21: 1 export = 1 file, respect the codebase naming/file/folder conventions

packages/twenty-oxlint-rules/rules/prefer-workspace-scoped-repository.ts
FelixMalfait: Inverted to a real blacklist in fa6d07c61c — fires for every `*Entity` not in `EXCLUSIONS`. New entities default to enforcement.

packages/twenty-server/src/engine/core-modules/admin-panel/services/admin-panel-chat.service.ts
FelixMalfait: Shortened to 2 lines in e812b542ac.

packages/twenty-server/src/engine/core-modules/approved-access-domain/services/approved-access-domain.service.ts
FelixMalfait: Removed in e812b542ac.

packages/twenty-server/src/engine/core-modules/approved-access-domain/services/approved-access-domain.service.ts
FelixMalfait: Removed in e812b542ac.

packages/twenty-server/src/engine/core-modules/billing/services/billing-portal.workspace-service.ts
FelixMalfait: Migrated to `@InjectWorkspaceScopedRepository` in e812b542ac; the call site now uses the scoped `findOne(workspaceId, …)`.

packages/twenty-server/src/engine/core-modules/billing/services/billing-subscription-update.service.ts
FelixMalfait: Threaded `workspaceId` through `updateSubscription(workspaceId, subscriptionId, …)` in e812b542ac so the lookup goes through the scoped repo.

packages/twenty-server/src/engine/core-modules/billing/services/billing-subscription.service.ts
FelixMalfait: Dual-injected in e812b542ac. Scoped repo for every workspaceId-bearing call; the unscoped one is used only when `criteria.workspaceId` is absent (Stripe webhook by `stripeCustomerId`).

packages/twenty-server/src/engine/core-modules/file-storage/file-storage.service.ts
FelixMalfait: Removed in e812b542ac.

FelixMalfait: No leak: TypeORM caches `manager.getRepository(Entity)` per EntityManager, so we only allocate a single small stateless wrapper per `writeFile` call (GCd at scope exit). Replaced the inline `new` with a `withManager(manager)` helper on the wrapper for clarity (e812b542ac); unit test added in fa6d07c61c.

packages/twenty-server/src/engine/twenty-orm/workspace-scoped-repository/workspace-scoped-repository.ts
FelixMalfait: Split into 5 files in e812b542ac: workspace-scoped-entity.type.ts, get-workspace-scoped-repository-token.util.ts, inject-workspace-scoped-repository.decorator.ts, provide-workspace-scoped-repository.ts, and workspace-scoped-repository.ts (class only). All 84 import sites updated.

packages/twenty-server/src/engine/twenty-orm/workspace-scoped-repository/workspace-scoped-repository.ts
```diff
@@ -0,0 +1,176 @@
...
+  // Returns a new wrapper bound to the given EntityManager. Use inside
+  // a QueryRunner transaction so reads/writes participate in the
+  // transaction while still being workspace-scoped.
+  withManager(manager: EntityManager): WorkspaceScopedRepository<T> {
+    return new WorkspaceScopedRepository<T>(
+      manager.getRepository(this.repository.target),
+    );
+  }
+
+  private mergeWorkspaceIdIntoWhere(
```
FelixMalfait, line 168: Can you please check the impact on performance and Indexes? I guess workspaceId should usually appear early in the clause no since it's a great first key for indexes?

packages/twenty-oxlint-rules/rules/prefer-workspace-scoped-repository.ts
```diff
@@ -0,0 +1,138 @@
...
+  'MessageFolderEntity',
+  'RolePermissionFlagEntity',
+  'SigningKeyEntity',
+  'UserEntity',
+  'WorkspaceSSOIdentityProviderEntity',
+
+  // Workspace-scoped entities currently outside the wrapper. Removing an
+  // entry forces every raw @InjectRepository site for it to migrate to
+  // the wrapper or carry an explicit eslint-disable.
+  'ApplicationEntity',
```
Weiko, line 30: Small nit: This sublist could live in a dedicated ENTITIES_TO_MIGRATE to distinguish clearly what's left to migrate. Not a strong opinion on this, the comment is probably enough

packages/twenty-server/src/engine/twenty-orm/workspace-scoped-repository/__tests__/workspace-scoped-repository.spec.ts
```diff
@@ -0,0 +1,298 @@
...
+
+      expect(repository.findOne).toHaveBeenCalledWith({
+        where: [
+          { id: 'a', workspaceId: WORKSPACE_ID },
+          { status: 'queued', workspaceId: WORKSPACE_ID },
+        ],
+      });
+    });
+
+    it('overrides any caller-supplied workspaceId with the scoped one', async () => {
```
Weiko: I feel like this could be error prone since it's well hidden. Not sure what we want to do 🤔

packages/twenty-server/src/engine/core-modules/two-factor-authentication/two-factor-authentication.resolver.ts
```diff
@@ -146,19 +146,16 @@ export class TwoFactorAuthenticationResolver {
       );
     }
 
-    if (
-      twoFactorMethod.userWorkspace.userId !== user.id ||
-      twoFactorMethod.userWorkspace.workspaceId !== workspace.id
```
Weiko, line 151: is this intentional?

packages/twenty-server/src/engine/twenty-orm/workspace-scoped-repository/workspace-scoped-repository.ts
FelixMalfait: Good catch. Done in bafe65c67e: `workspaceId` is now emitted as the **first** key in the merged WHERE/criteria (and on every clause of an array-of-OR). Caller-supplied `workspaceId` is stripped first so the scoped value always wins. \n\nPostgres' planner picks the index regardless of WHERE clause order, so this is intent/clarity (and matches how the composite indexes `(workspaceId, …)` are declared) rather than raw perf — but it also makes the security boundary obvious in query logs. Added a unit test that asserts `Object.keys(where)[0] === 'workspaceId'` so the order can't regress silently.

packages/twenty-oxlint-rules/rules/prefer-workspace-scoped-repository.ts
```diff
@@ -0,0 +1,138 @@
...
+  },
+  create: (context) => {
+    return {
+      MethodDefinition: (node: any) => {
+        if (node.kind !== 'constructor') {
+          return;
+        }
+
+        for (const param of node.value.params ?? []) {
+          if (param.type !== 'TSParameterProperty') {
```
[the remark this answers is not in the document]
FelixMalfait: Valid finding — done in bafe65c67e. Dropped the `TSParameterProperty`-only check so the rule iterates `param.decorators` for any constructor parameter shape (parameter-property shorthand AND plain `(@InjectRepository(X) foo: Repository<X>)` followed by explicit `this.foo = foo;`). Added a rule-tester case covering the regression.

packages/twenty-server/src/engine/core-modules/two-factor-authentication/two-factor-authentication.resolver.ts
FelixMalfait: I guess this check is not useful since we know for sure now it belongs to the workspace already

packages/twenty-server/src/engine/twenty-orm/workspace-scoped-repository/workspace-scoped-repository.ts
```diff
@@ -0,0 +1,194 @@
...
+// Wraps a TypeORM Repository so every operation is scoped to a
+// caller-supplied workspaceId. Makes cross-workspace IDOR bugs
+// impossible-by-default for core- and metadata-schema entities that
+// carry a workspaceId column. Workspace-data schema entities continue
+// to use WorkspaceRepository / twentyORMManager.
+export class WorkspaceScopedRepository<T extends WorkspaceScopedEntity> {
+  constructor(private readonly repository: Repository<T>) {}
+
+  findOne(workspaceId: string, options: FindOneOptions<T>): Promise<T | null> {
+    return this.repository.findOne({
```
Weiko, line 27: As mentioned, we should throw if workspaceId is undefined during runtime. If we do that we are fixing another big issue we had where typeorm normal behavior is to drop the column from the WHERE clause which would have returned all entities (which is what your PR is also trying to fix)

packages/twenty-server/src/engine/twenty-orm/workspace-scoped-repository/workspace-scoped-repository.ts
FelixMalfait: Done in bafe65c67e (and locked in with a Object.keys assertion test). `workspaceId` is now prepended to the merged WHERE/criteria and we strip any caller-supplied value first. Postgres's planner picks the index regardless of clause order so this is intent/clarity (matches the composite `(workspaceId, …)` indexes) rather than raw perf — and it makes the security boundary obvious in query logs.

packages/twenty-oxlint-rules/rules/prefer-workspace-scoped-repository.ts
FelixMalfait: Done in f0942332d9. Split into two named sets: `STRUCTURAL_EXEMPTIONS` (workspace itself / pivots / nullable / global tables) and `WORKSPACE_SCOPED_EXEMPTIONS` (workspace-scoped but still injected raw). The two categories are now obvious at a glance.

packages/twenty-server/src/engine/twenty-orm/workspace-scoped-repository/__tests__/workspace-scoped-repository.spec.ts
FelixMalfait: Good call. Switched to **throw** rather than silent override in a4ee053bff. Any caller-supplied `workspaceId` in the WHERE / criteria now raises a descriptive Error, and tests cover the single-clause + array-of-OR + update-criteria paths.

packages/twenty-server/src/engine/twenty-orm/workspace-scoped-repository/workspace-scoped-repository.ts
FelixMalfait: Done in f0942332d9. Every public method on the wrapper now calls `assertWorkspaceId` which throws on `undefined` / `null` / `''`. Spec covers each method with a parameterised test. As you said this closes the silently-dropped-from-WHERE TypeORM behaviour too.

PR #20958, opened 2026-05-27
fix(front): ignore IME composition Enter in input hotkeys

### LINKED ISSUE #20954: Confirming an IME (Japanese/CJK) composition with Enter submits the input instead of only confirming the conversion
## Bug Description

When using an IME (Input Method Editor) for CJK languages — Japanese, Chinese, Korean — typing shows conversion candidates that are confirmed ("commit" / 確定) by pressing Enter. In Twenty, that confirmation Enter is **also** consumed by the input's Enter handler, so the field is submitted/saved before the user has finished composing their text.

In practice this makes it very hard to type CJK text into any input that submits on Enter (record cell editing, record title, text/links/emails fields, the command menu, etc.), as well as a few inputs that handle Enter directly (workspace creation name, new API key name, blocklist input).

**Steps to reproduce**

1. Set the OS input source to a Japanese (or other CJK) IME.
2. Open a record (e.g. People / Companies) and start editing the **Name** field, or open the record-creation input.
3. Type romaji so that conversion candidates appear (the text is still composing / underlined, the IME candidate window is shown).
4. Press **Enter** to confirm the conversion.
5. **Observed:** the field is submitted/saved immediately on that Enter, before composition is finished — the user cannot confirm the conversion without also triggering submit.

**Demo (GIF):** typing `てすと` → IME candidate window is open → a single Enter both confirms the conversion **and** saves the field (`テスト` is immediately committed to the record and the list):

Image

## Expected behavior

Pressing Enter **while an IME composition is in progress** should only confirm the conversion and must not trigger submit / escape / tab hotkeys. Only an Enter pressed on already-committed text (no composition in progress) should submit. This is the standard behavior expected by all CJK input users (and also benefits accented / dead-key input on Latin layouts).

## Technical inputs

Root cause: the frontend never inspects `KeyboardEvent.isComposing` (nor the legacy `keyCode === 229`). `isComposing` is currently not used anywhere in `twenty-front`.

- Enter / Escape / Tab hotkeys for inputs are centralized in `useHotkeysOnFocusedElement` (`react-hotkeys-hook@^4.4.4`), whose callback receives the `KeyboardEvent` but fires regardless of composition state. `useRegisterInputEvents` routes `onEnter` / `onEscape` / `onTab` through it for ~13 input components.
- `react-hotkeys-hook` v4 does not handle IME composition on its own, so a guard must be added explicitly.
- A few components handle Enter directly via `onKeyDown`: `CreateWorkspace.tsx`, `SettingsDevelopersApiKeysNew.tsx`, `SettingsAccountsBlocklistInput.tsx`.

Suggested fix: short-circuit when a composition is in progress, e.g. in `useHotkeysOnFocusedElement`:

```ts
if (keyboardEvent.isComposing || keyboardEvent.keyCode === 229) {
  return;
}
```

plus the same guard in the direct `onKeyDown` handlers. This has **no effect on non-IME (Latin) typing**, since `isComposing` is only true during an active composition.

**Environment**

- Twenty: v2.8.3 (self-hosted, official `twentycrm/twenty:latest` Docker image; also reproducible on current `main` — `isComposing` is unused in `twenty-front`)
- Browser: Chrome
- OS: macOS with Japanese IME

I have a fix for this and would be happy to open a PR.

fyujiwara:
## What

Pressing Enter to confirm an IME (CJK) composition no longer submits the input. The Enter / Escape / Tab handlers now ignore key events fired while a composition is in progress (`isComposing`, or the legacy `keyCode === 229`).

Fixes #20954

## Why

`isComposing` was not checked anywhere in `twenty-front`, so the Enter that confirms a Japanese / Chinese / Korean conversion was also consumed as a submit / escape / tab hotkey — making it very hard to type CJK text into any input that submits on Enter.

## Changes

- `useHotkeysOnFocusedElement` — central guard; covers every input wired through `useRegisterInputEvents` (~13 components) and all hotkeys routed through this hook.
- Direct `onKeyDown` Enter handlers: `CreateWorkspace`, `SettingsDevelopersApiKeysNew`, `SettingsAccountsBlocklistInput`.

## Notes

- No effect on non-IME (Latin) typing — `isComposing` is only true during an active composition. It also improves accented / dead-key input on Latin layouts.
- `react-hotkeys-hook@4` does not handle IME composition on its own, so the guard is explicit.

## Testing

Manually verified with a Japanese IME on Chrome (macOS) against the v2.8.3 self-hosted image: romaji + Enter now only confirms the conversion; a second Enter on committed text submits as expected. The GIF in #20954 shows the original buggy behavior.

### DIFF
--- a/packages/twenty-front/src/modules/settings/accounts/components/SettingsAccountsBlocklistInput.tsx
+++ b/packages/twenty-front/src/modules/settings/accounts/components/SettingsAccountsBlocklistInput.tsx
@@ -74,6 +74,9 @@ export const SettingsAccountsBlocklistInput = ({
   });
 
   const handleKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
+    if (e.nativeEvent.isComposing || e.keyCode === 229) {
+      return;
+    }
     if (e.key === Key.Enter) {
       submit();
     }
--- a/packages/twenty-front/src/modules/ui/utilities/hotkey/hooks/useHotkeysOnFocusedElement.ts
+++ b/packages/twenty-front/src/modules/ui/utilities/hotkey/hooks/useHotkeysOnFocusedElement.ts
@@ -48,6 +48,10 @@ export const useHotkeysOnFocusedElement = ({
   return useHotkeys(
     keys,
     (keyboardEvent, hotkeysEvent) => {
+      if (keyboardEvent.isComposing || keyboardEvent.keyCode === 229) {
+        return;
+      }
+
       callScopedHotkeyCallback({
         keyboardEvent,
         hotkeysEvent,
--- a/packages/twenty-front/src/pages/onboarding/CreateWorkspace.tsx
+++ b/packages/twenty-front/src/pages/onboarding/CreateWorkspace.tsx
@@ -141,6 +141,9 @@ export const CreateWorkspace = () => {
   );
 
   const handleKeyDown = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.nativeEvent.isComposing || event.keyCode === 229) {
+      return;
+    }
     if (event.key === Key.Enter) {
       event.preventDefault();
       handleSubmit(onSubmit)();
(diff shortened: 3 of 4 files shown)


bosiraphael (approved):
LGTM, thank you for your contribution

PR #20256, opened 2026-05-04
feat(sdk): add defineCommandMenuItem 

FelixMalfait:
## Summary

- Add `defineCommandMenuItem` and `definePageLayoutWidget` as standalone SDK defines, mirroring the existing `definePageLayoutTab` pattern. Both entities can still be declared nested inside their parent (`defineFrontComponent.command` / `definePageLayout.tabs[].widgets[]`).
- Add `CommandMenuItem` and `PageLayoutWidget` to the `SyncableEntity` enum and the dev-mode UI labels.
- Wire the SDK manifest-build to extract the two new defines into top-level `commandMenuItems` / `pageLayoutWidgets` arrays on the manifest, and the server aggregator to consume them through the existing flat-entity converters.
- On the server, expose `Application.commandMenuItems` (relation + DTO + service hydration in `findOneApplication`).
- On the front, list command menu items in the application content tab and add a dedicated detail page with a settings tab, mirroring how `frontComponents` are surfaced.
- Add `twenty add` templates and Vitest unit tests for both new defines.
- Document the standalone-vs-nested pattern in `packages/twenty-sdk/README.md`.

### Why

Until now, command menu items could only be declared as the nested `command:` field on `defineFrontComponent` — there was no way to register a command menu item from a separate file or from another package. The `SyncableEntity` enum had 12 values, while the server already synced 18 (including `commandMenuItem` and `pageLayoutWidget`). The same gap existed for `pageLayoutWidget`, which had no top-level define despite being synced server-side. This PR closes both gaps and aligns the SDK surface with what the server actually accepts.

The standalone defines coexist with the nested form — pick one per entity, never both with the same `universalIdentifier` (the manifest aggregator will throw on duplicates). The README now documents this.

## Test plan

- [x] `npx nx typecheck twenty-sdk` / `twenty-server` / `twenty-front`
- [x] `npx nx lint:diff-with-main twenty-front` / `twenty-server`
- [x] `npx nx lint twenty-sdk` / `twenty-shared`
- [x] New unit tests: `define-command-menu-item.spec.ts`, `define-page-layout-widget.spec.ts`
- [x] Existing manifest extract config tests still pass
- [ ] Codegen `npx nx run twenty-front:graphql:generate --configuration=metadata` should be re-run after merge — the generated `graphql.ts` was patched manually to include `commandMenuItems` on `Application` and the `FindOneApplication` document.
[...]

PR #20960, opened 2026-05-27
fix(sdk): restore nested defineFrontComponent  (create command menu items

Weiko:
## What

`defineFrontComponent({ command: {...} })` silently stopped creating a `CommandMenuItem`. The front component is synced, but its command never appears in the command menu (Cmd+K).

#20256 ("add defineCommandMenuItem") was meant to let nested and standalone forms coexist, but it removed the nested path entirely: `command` was dropped from
`FrontComponentConfig`/`FrontComponentManifest`, its validation in `defineFrontComponent`, and the manifest-build extraction — without a replacement, deprecation, or warning. Apps using
 the nested form get a front component with no command menu item.

  ## Fix

  Re-introduce the nested `command`, but **lower it into the top-level `commandMenuItems` array at build time** instead of restoring the old "store on manifest + expand server-side" path:

  - `twenty-shared`: re-add `FrontComponentCommandManifest` (+ barrel export).
  - `twenty-sdk`: re-add `FrontComponentCommandConfig` and `command?` on `FrontComponentConfig`; re-add nested-command validation in `defineFrontComponent`.
  - `manifest-build`: when a front component has a `command`, push it into `commandMenuItems` with `frontComponentUniversalIdentifier` set to the component's id.

### DIFF
--- a/packages/twenty-sdk/src/cli/utilities/build/manifest/manifest-build.ts
+++ b/packages/twenty-sdk/src/cli/utilities/build/manifest/manifest-build.ts
@@ -51,7 +51,7 @@ import {
   getInputSchemaFromSourceCode,
   jsonSchemaToInputSchema,
 } from 'twenty-shared/logic-function';
-import { assertUnreachable } from 'twenty-shared/utils';
+import { assertUnreachable, isDefined } from 'twenty-shared/utils';
 
 const loadSources = async (appPath: string): Promise<string[]> => {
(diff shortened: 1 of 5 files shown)


martmull (approved):

Weiko:
Closing as https://github.com/twentyhq/twenty/pull/20256 was actually the correct path and the breaking change was actually intended

PR #20961, opened 2026-05-27
Fix focus in front components inputs

### LINKED ISSUE #20714: Front-component renderer doesn't push focus context — host global hotkeys (g+key, /, ?) fire while typing in worker-rendered inputs
## Summary

When a user focuses an `<input>` inside a front-component (Web Worker / Remote DOM sandbox) and types, Twenty's host-side global hotkeys — specifically the `g + <key>` go-to-record sequences registered via `useGoToHotkeys` — still fire. Typing the letters `g`, `c`, `o`, `p`, etc. into a free-text filter chip input navigates away to the corresponding object index page instead of inserting the character. Standalone `/`, `?` and similar single-letter shortcuts have the same problem.

This makes any front-component that contains a free-text input effectively unusable for typed entry.

## Reproduction

1. Define a `defineFrontComponent({...})` that renders a plain `<input type="text" />`.
2. Mount it as a `FRONT_COMPONENT` widget on any page-layout tab.
3. Focus the input and type `gc`. Twenty navigates to the Companies index page.
4. Typing `/` opens the command menu. Typing `?` opens keyboard help.

Reproduced against both a self-hosted local Twenty (localhost) and a hosted instance.

## Diagnosis

`useGoToHotkeys` is explicitly configured to fire even when typing in form / contentEditable elements:

```ts
// packages/twenty-front/src/modules/ui/utilities/hotkey/hooks/useGoToHotkeys.ts
useGlobalHotkeysSequence(
  'g',
  key,
  () => { /* navigate */ },
  {
    enableOnContentEditable: true,
    enableOnFormTags: true,
    preventDefault: true,
  },
  ...
);
```

The gate that disables this is `currentGlobalHotkeysConfig.enableGlobalHotkeysConflictingWithKeyboard`, checked inside `useGlobalHotkeysCallback`:

```ts
// packages/twenty-front/src/modules/ui/utilities/hotkey/hooks/useGlobalHotkeysCallback.ts
if (
  !containsModifier &&
  !currentGlobalHotkeysConfig.enableGlobalHotkeysConflictingWithKeyboard
) {
  return;
}
```

Twenty's own input components flip this flag to `false` on focus — `TextArea.tsx`, `TitleInput.tsx`, `FormFieldInputInnerContainer.tsx`, `CommandMenuItemNumberInput.tsx`, `useModal.tsx`, and several others all set `enableGlobalHotkeysConflictingWithKeyboard: false` via the focus stack.

**The `FrontComponentRenderer` does not push any focus context.** I scanned both `packages/twenty-front-component-renderer/src/host/components/FrontComponentRenderer.tsx` and `packages/twenty-front/src/modules/front-components/components/FrontComponentRenderer.tsx` (and the wrapping `FrontComponentRendererProvider`) for hotkey / focus / keyboard references — zero matches.

There is also no API on the exported `frontComponentHostCommunicationApi` for the worker to manipulate the hotkey gate or focus stack:

```ts
// packages/twenty-front-component-renderer/src/types/FrontComponentHostCommunicationApi.ts
// Exposes only: navigate, requestAccessTokenRefresh, openSidePanelPage,
// openCommandConfirmationModal, unmountFrontComponent, enqueueSnackbar,
// closeSidePanel, updateProgress.
```

So when focus enters a worker-rendered input, the Jotai atom still reports `enableGlobalHotkeysConflictingWithKeyboard: true`, and the host's document-level keydown listener fires the `g + key` callback before the worker even sees the event.

## Why worker-side mitigations don't work

I tried:

1. Bubble-phase `onKeyDown` on the front-component root that calls `e.stopPropagation()` + `e.nativeEvent.stopImmediatePropagation()` when the target is a form element → no effect.
2. The same handler in capture phase → no effect (and would block typing if it did, since capture fires before the input).

These fail because by the time the worker dispatches the synthetic event, Twenty's host listener has already fired and the navigation has already executed (the host handler itself calls `stopImmediatePropagation` per `useGlobalHotkeysCallback`). The worker is too late in the chain.

## Suggested fix

`FrontComponentRenderer` should push `enableGlobalHotkeysConflictingWithKeyboard: false` (and likely also `enableGlobalHotkeysWithModifiers: false`) onto the focus stack when an `<input>` / `<textarea>` / contentEditable element inside the rendered worker tree is focused, and pop it on blur — mirroring the pattern used by `TextArea.tsx` / `FormFieldInputInnerContainer.tsx`.

If detecting focus inside the proxied tree is fragile, an alternative is to push the focus context whenever a front-component widget is mounted (broader scope, but front-components with no typeable inputs are rare in practice).

A third option is to expose the focus-stack API on `frontComponentHostCommunicationApi` so app authors can opt in explicitly per component.

## Versions

- Platform: v2.4.0 (latest release as of 2026-05-19)
- twenty-sdk: 2.5.0
- twenty-client-sdk: 2.5.0

## Related

- The `useScopedHotkeys` / focus-stack mechanism is the established pattern in the codebase for this exact concern (PR #6322).
- Reproducer dashboard available on request.

bosiraphael:
Fixes https://github.com/twentyhq/twenty/issues/20714

Fixes keyboard hotkey conflicts when typing inside `<input>` / `<textarea>` elements rendered by Front Components. Editable fields rendered through the component renderer now properly push/pop a focus item onto Twenty's focus stack, disabling global keyboard hotkeys while the user is typing.

## Before

https://github.com/user-attachments/assets/2003c2cb-2698-480f-aedf-bb2f30396572

## After

https://github.com/user-attachments/assets/2c7c6cb0-ecd7-4557-a77b-4d1f264345f0


martmull (approved):

Weiko (approved):
LGTM

PR #20965, opened 2026-05-27
Fix: Prevent hotkey/Enter handling during IME composition

### LINKED ISSUE #20954: Confirming an IME (Japanese/CJK) composition with Enter submits the input instead of only confirming the conversion

bosiraphael:
Fixes https://github.com/twentyhq/twenty/issues/20954
Prevents keyboard handlers from firing while users are composing characters with an IME (e.g. Japanese, Chinese, Korean input). When typing with an IME, pressing Enter to confirm a candidate would previously trigger submit/hotkey actions unintentionally.

Added isComposing / keyCode === 229 guards in:
- useHotkeysOnFocusedElement — global hotkey hook
- SettingsAccountsBlocklistInput — blocklist entry submit
- CreateWorkspace — onboarding form submit
- SettingsDevelopersApiKeysNew — API key name submit

## Video QA

https://github.com/user-attachments/assets/8fc385e1-bdda-4e34-899e-a248df9bfdd7

### DIFF
--- a/packages/twenty-front/src/modules/settings/accounts/components/SettingsAccountsBlocklistInput.tsx
+++ b/packages/twenty-front/src/modules/settings/accounts/components/SettingsAccountsBlocklistInput.tsx
@@ -74,6 +74,9 @@ export const SettingsAccountsBlocklistInput = ({
   });
 
   const handleKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
+    if (e.nativeEvent.isComposing || e.keyCode === 229) {
+      return;
+    }
     if (e.key === Key.Enter) {
       submit();
     }
--- a/packages/twenty-front/src/modules/ui/utilities/hotkey/hooks/useHotkeysOnFocusedElement.ts
+++ b/packages/twenty-front/src/modules/ui/utilities/hotkey/hooks/useHotkeysOnFocusedElement.ts
@@ -48,6 +48,10 @@ export const useHotkeysOnFocusedElement = ({
   return useHotkeys(
     keys,
     (keyboardEvent, hotkeysEvent) => {
+      if (keyboardEvent.isComposing || keyboardEvent.keyCode === 229) {
+        return;
+      }
+
       callScopedHotkeyCallback({
         keyboardEvent,
         hotkeysEvent,
--- a/packages/twenty-front/src/pages/onboarding/CreateWorkspace.tsx
+++ b/packages/twenty-front/src/pages/onboarding/CreateWorkspace.tsx
@@ -141,6 +141,9 @@ export const CreateWorkspace = () => {
   );
 
   const handleKeyDown = (event: React.KeyboardEvent<HTMLInputElement>) => {
+    if (event.nativeEvent.isComposing || event.keyCode === 229) {
+      return;
+    }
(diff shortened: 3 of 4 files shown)


bosiraphael:
Closing as duplicate of https://github.com/twentyhq/twenty/pull/20958

PR #20967, opened 2026-05-27
refactor(website): replace axios with native fetch

FelixMalfait:
## Summary

Follow-up to #20966 (Stripe fetch client). axios's default Node http adapter on workerd has the same TLS hang the Stripe SDK had — it just doesn't surface today because all three call sites are wrapped in \`unstable_cache(revalidate: 3600)\` and the cache is populated at build time, so misses are rare and the failure mode is a silent \`null\` to the layout.

This swaps the three remaining axios calls in \`twenty-website\` for native \`fetch\` and removes axios from \`packages/twenty-website/package.json\`. The package is still used by other workspaces, so yarn.lock keeps the other resolution.

Touched call sites:
- \`src/lib/releases/fetch-latest-release-tag.ts\` (GitHub releases — runs at build time, cosmetic)
- \`src/lib/community/fetch-github-star-count.ts\` (GitHub star count in menu)
- \`src/lib/community/fetch-discord-member-count.ts\` (Discord member count in menu)

## Test plan

- [ ] After merge + deploy: confirm GitHub star + Discord member counts render in the site menu (non-zero, formatted)
- [ ] Confirm \`/releases\` shows the latest tag-gated visible release notes
- [ ] No \`axios\` in worker bundle (\`grep axios .open-next/worker.js\` should be empty)

### DIFF
--- a/packages/twenty-website/package.json
+++ b/packages/twenty-website/package.json
@@ -24,7 +24,6 @@
     "@lottiefiles/dotlottie-react": "^0.18.10",
     "@tabler/icons-react": "^3.41.1",
     "@wyw-in-js/babel-preset": "^0.8.1",
-    "axios": "^1.14.0",
     "framer-motion": "^11.18.0",
     "gray-matter": "^4.0.3",
     "next": "16.1.7",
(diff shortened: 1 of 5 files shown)


FelixMalfait:
@mabdullahabaid fyi

PR #20981, opened 2026-05-27
[Apps] Inject caller permission context into logic function execution

Weiko:
## What & why

Logic functions already receive an env-injected "API caller context" (API URL, app access token). This PR extends that so a function can also know **what the calling role is allowed to do** — object permissions, field permissions, permission flags, and the role's capability booleans — **without making an API call**. 
The context is computed server-side at execution time and injected as a new env variable, then read synchronously via a small SDK helper.

## How it works

- The executor resolves the **caller's `roleId`** (never assuming a user — works for users, API keys, application API keys, and agents, all of which carry a role via `role-target`) and builds a `LogicFunctionPermissionContext`, serialized into a new `TWENTY_PERMISSION_CONTEXT` env var alongside the existing token vars.
- Objects, fields, and permission flags are keyed by `universalIdentifier` (stable across renames). Role capability booleans live at the root.
- The context is built **entirely from the workspace cache** (`rolesPermissions`, `flatRoleMaps`, `flatRolePermissionFlagMaps`, `flatPermissionFlagMaps`), consistent with the rest of the permission system — no extra DB query. Permission-flag semantics mirror `getUserWorkspacePermissions` (settings flags gated by `canUpdateAllSettings`, tool flags by `canAccessAllTools`, plus explicit per-flag grants), and now naturally cover app-defined custom flags.

## Caller-path wiring

Every path that triggers a logic function now resolves and forwards `roleId`:

- **Route trigger** — resolves the role from the request's auth context via the shared `buildWorkspaceAuthContext` + `resolveRolePermissionConfig` guards.
- **Trigger job** — `roleId` added to the job payload.
- **Workflow action** — derived from the run's `rolePermissionConfig`.
- **Tool executor** — uses the already-resolved `context.roleId`.
- **Dev/test execution** (`executeOneLogicFunction`) — resolves the caller's role from `@AuthUserWorkspaceId`.

Background triggers with no caller (cron, system db-events) get an empty default context.

## SDK

New `getPermissionContext()` helper in `twenty-sdk/logic-function`:

```ts
import { getPermissionContext, ObjectPermissionLevel, FieldPermissionLevel } from 'twenty-sdk/logic-function';

const perms = getPermissionContext();
perms.hasObjectPermission(PERSON_UID, ObjectPermissionLevel.READ);
perms.hasFieldPermission(PERSON_UID, NAME_FIELD_UID, FieldPermissionLevel.READ);
perms.hasPermissionFlag(WORKFLOWS_FLAG_UID);
perms.canReadAllObjectRecords;
```

## Refactor

Extracted the raw→WorkspaceAuthContext conversion out of WorkspaceAuthContextMiddleware into a reusable buildWorkspaceAuthContext util, so the route trigger resolves the caller role through the existing auth guards instead of hand-rolled field checks.

## Testing

- Sample function permission-context-demo.function.ts (auth-required route) demonstrating the SDK helper.
- Integration test asserting TWENTY_PERMISSION_CONTEXT is injected and that the caller's WORKFLOWS flag (guaranteed by the mutation's guard) resolves true, keyed by its
universalIdentifier.


packages/twenty-apps/internal/self-hosting/src/logic-functions/permission-context-demo.function.ts
martmull, line 1: Why editing this app?

packages/twenty-sdk/src/sdk/logic-function/permission-context/get-permission-context.ts
```diff
@@ -0,0 +1,113 @@
...
+
+const EMPTY_PERMISSION_CONTEXT: LogicFunctionPermissionContext = {
+  canReadAllObjectRecords: false,
+  canUpdateAllObjectRecords: false,
+  canSoftDeleteAllObjectRecords: false,
+  canDestroyAllObjectRecords: false,
+  canUpdateAllSettings: false,
+  canAccessAllTools: false,
+  permissionFlags: {},
+  objectsPermissions: {},
```
martmull, line 16: Add field permissions for consistency?

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-executor/logic-function-executor.service.ts
```diff
@@ -269,6 +284,86 @@ export class LogicFunctionExecutorService {
...
+        canUpdate: objectPermissions.canUpdateObjectRecords,
+        canSoftDelete: objectPermissions.canSoftDeleteObjectRecords,
+        canDestroy: objectPermissions.canDestroyObjectRecords,
+        restrictedFields,
+      };
+    }
+
+    return {
+      ...capabilities,
+      permissionFlags,
```
martmull: What about field permissions?

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-executor/logic-function-executor.service.ts
```diff
@@ -269,6 +284,86 @@ export class LogicFunctionExecutorService {
     };
   }
 
+  private async buildPermissionContext({
```
martmull: Too complicated to stay in this service, can you move it in permission service?

packages/twenty-apps/internal/self-hosting/src/logic-functions/permission-context-demo.function.ts
Weiko: This goes through the CI I think? so I thought it would at least do typecheck, I can remove

packages/twenty-sdk/src/sdk/logic-function/permission-context/get-permission-context.ts
Weiko: Field permissions is within objectPermission as "restrictedFields", this is consistent with the rest of the codebase. People should use the new helps anyway so I think this is fine

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-executor/logic-function-executor.service.ts
Weiko: same

packages/twenty-server/src/engine/core-modules/logic-function/logic-function-executor/logic-function-executor.service.ts
Weiko: True

Weiko:
Will reopen later this week

PR #20984, opened 2026-05-27
fix(twenty-front): resolve SettingsDataModelFieldEditFormValues type TODO

Dakshx07:
### Description
Resolves a pending `//TODO: fix this type` comment inside the settings data-model folder.
Replaced the generic and unsafe `& any` type with a strict, type-safe interface extension that handles the optional `relation` data-model payload properly.

This improves compilation safety and aligns with twenty-front's core principle of strict TypeScript usage.


ijreilly:
Thank you for your contribution!
Typecheck is failing on this PR, the type change is likely missing something.
Please do the appropriate changes, then run `npx nx run twenty-front:typecheck` locally and re-open a PR only once it passes.  thank you !

PR #20985, opened 2026-05-27
fix: disable go-to hotkeys when focus is inside input or textarea

### LINKED ISSUE #20894: keyboard shortcuts edge case
Some of the keyboard shortcuts trigger switching page in the middle of typing. 
Specifically G+N opens Notes page, while writing "assignment" in workflow name. 
I guess if input is in focus short cuts should not be handled at all

steps to reproduce:
1) click create new workflow
2) enter "assignment" as workflow name

dilanmelvin:
Fixes #20894

Global navigation shortcuts (G+N, G+C, G+S, etc.) were firing while
the user was typing in an input or textarea, causing unexpected page
navigation mid-typing.

Reproducer: create a new workflow → type "assignment" as the name →
G+N triggers and navigates to the Notes page.

## Root cause

`useGoToHotkeys` passed `enableOnFormTags: true` and
`enableOnContentEditable: true` to `react-hotkeys-hook`, which tells
the library to fire even when focus is inside a form element.

## Fix

Set both flags to `false`. Two-line change.
Added regression tests for input and textarea focus scenarios.


dilanmelvin:
The CI failure in `front-sb-test` is unrelated to this PR.

The failing test is `DateFieldDisplay > Performance` which checks that 
the component renders in < 0.1ms. It measured 0.10014ms — a difference 
of 0.00014ms caused by CI machine load, not by this change.

This PR only modifies `useGoToHotkeys.ts` and its test file. The 
DateFieldDisplay component is not touched anywhere in this diff.

dilanmelvin:
Hi team! This fixes #20894 — global go-to shortcuts (G+N, G+C, etc.) were firing while typing in input fields because enableOnFormTags and enableOnContentEditable were both set to true. Changed both to false in useGoToHotkeys.ts. Added regression tests for input and textarea focus. The front-sb-test CI failure is an unrelated DateFieldDisplay performance flake (0.10014ms vs 0.1ms threshold). Happy to make any changes if needed!

Weiko (commented):
Hi @dilanmelvin, thanks for your contribution.
This should have been fixed by https://github.com/twentyhq/twenty/pull/21029 already
(@bosiraphael wdyt?)

bosiraphael:
Hello @Weiko, yes indeed, this should be solved. Closing then

dilanmelvin:
Thanks for the clarification! I must have missed the earlier fix in #21029. Appreciate you taking a look and confirming it's already resolved. I'll close this on my end. Thanks everyone! 🙌

PR #20988, opened 2026-05-28
feat(settings): add cancel and done buttons to select field bulk edit (#20986)

### LINKED ISSUE #20986: [feat]: add Cancel and Done buttons to select field bulk edit
## Scope & Context
During reproducing #20840 , I found that on Settings -> Data Models -> Objects -> Opportunities -> Stage (select field values), bulk edit for options has no way to confirm or discard edits inside bulk mode. Users must save or cancel the whole field settings form.

## Current behavior
On select/multi-select field settings (e.g. Stage), choosing "Bulk edit" from the options menu opens a text area. After editing lines, there are no done/save or cancel options for bulk editing only.

Image

## Expected behavior
In bulk edit, show Cancel and Done at the bottom of the values card (same footer area as Add option in single-edit mode).
- Done - apply textarea -> options and return to single edit.
- Cancel - discard bulk changes and return to options as before bulk edit started.
- Typing should not commit until Done (or equivalent, e.g. single edit commits like Done).

Image

## Technical inputs
We have to just update `SettingsDataModelFieldSelectForm.tsx`.

awf-cpu:
This PR adds Cancel and Done buttons to the bottom of the select/multi-select field options bulk edit mode. It defers updating the form options until Done is clicked, and cancels/discards any changes when Cancel is clicked. Closes #20986.

### DIFF
--- a/packages/twenty-front/src/modules/settings/data-model/fields/forms/select/components/SettingsDataModelFieldSelectForm.tsx
+++ b/packages/twenty-front/src/modules/settings/data-model/fields/forms/select/components/SettingsDataModelFieldSelectForm.tsx
@@ -395,11 +395,6 @@ export const SettingsDataModelFieldSelectForm = ({
                                   setOriginalOptions(options);
                                   setIsBulkInputMode(true);
                                 } else {
-                                  const nextOptions = convertBulkTextToOptions(
-                                    bulkInputText,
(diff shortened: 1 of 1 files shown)


FelixMalfait:
Thanks! This is not an issue / a design that was validated it's just a user suggestion, I think we'd want to do this differently

PR #21004, opened 2026-05-28
feat(settings): add cancel and done buttons to select field bulk edit (#20986)

awf-cpu:
This PR adds Cancel and Done buttons to the bottom of the select/multi-select field options bulk edit mode. It defers updating the form options until Done is clicked, and cancels/discards any changes if Cancel is clicked.


Weiko (commented):
@awf-cpu can you please add a screenshot BEFORE/AFTER, it's generally preferred after FE changes

awf-cpu:
### Before
before

### After
after

FelixMalfait:
Thanks but this is not something we're looking to add / it doesn't fit our designs

PR #21007, opened 2026-05-28
fix: refresh admin panel feature flags after toggling

FelixMalfait:
## Summary

- Add `refetchQueries` to the `updateWorkspaceFeatureFlag` mutation in the admin workspace detail page so the toggle reflects the new value after toggling.
- Rename `useFeatureFlagState` → `useAdminUpdateFeatureFlag` since the hook lives under `admin-panel` and is only consumed by the admin workspace detail page.

## Bug

In the admin panel, toggling a feature flag for a workspace other than the admin's own workspace sent the backend mutation successfully, but the toggle in the UI remained unchanged.

The displayed value is derived from:
```tsx
const currentWorkspaceValue =
  currentWorkspace?.id === workspaceId
    ? currentWorkspace?.featureFlags?.find((f) => f.key === flag.key)?.value
    : undefined;
const displayedValue = currentWorkspaceValue ?? flag.value;
```

When viewing a different workspace, `currentWorkspaceValue` is `undefined` so the toggle reads `flag.value` from the `WORKSPACE_LOOKUP_ADMIN_PANEL` query. That query was never refetched after the mutation, so the displayed value stayed stale.

The existing optimistic Jotai update on `currentWorkspaceState` still runs — it is needed so the rest of the app (anything consuming `useIsFeatureEnabled`) reacts immediately when an admin toggles a flag on their own workspace.

## Test plan

- [ ] Open the admin panel → pick a workspace that is not your own → Feature Flags tab → toggle a flag → toggle visually flips after the mutation completes.
- [ ] Same flow on your own workspace → toggle flips, and any UI gated on that flag also reacts.
- [ ] If the mutation fails, the toggle reverts (existing `onError` rollback path).

### DIFF
--- a/packages/twenty-front/src/modules/settings/admin-panel/hooks/useAdminUpdateFeatureFlag.ts
+++ b/packages/twenty-front/src/modules/settings/admin-panel/hooks/useAdminUpdateFeatureFlag.ts
@@ -3,7 +3,7 @@ import { useAtomState } from '@/ui/utilities/state/jotai/hooks/useAtomState';
 import { isDefined } from 'twenty-shared/utils';
 import { type FeatureFlagKey } from '~/generated-admin/graphql';
 
-export const useFeatureFlagState = () => {
+export const useAdminUpdateFeatureFlag = () => {
   const [currentWorkspace, setCurrentWorkspace] = useAtomState(
     currentWorkspaceState,
   );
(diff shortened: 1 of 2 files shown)


utara-ux (commented):
/usr/bin/tar -xf /home/runner/work/_temp/bc97dc30-19cc-40f8-a2e9-292585a82f05/cache.tzst -P -C /home/runner/work/twenty/twenty --use-compress-program unzstd

PR #21009, opened 2026-05-28
fix(front): surface widget render errors via ErrorBoundary onError

clivemeister:
## Summary

The page-layout widget `ErrorBoundary` in `WidgetCardShell` renders a generic
"Invalid Configuration" fallback whenever a widget renderer throws. Because
there is no `onError` handler, the underlying error is swallowed — unrelated
widget types (fields, notes, front-component, etc.) all surface the same chip
with no telemetry, which makes render failures hard to triage.

This adds an `onError` handler that forwards the caught error to `console.error`
and to Sentry (when available) with the widget's `id`, `type`, and
`configurationType` as extra context. It reuses the same dynamic-import Sentry
pattern already used by `AppErrorBoundary` and `CommandMenuItemErrorBoundary`,
so it degrades gracefully to a console log when Sentry is not configured. The
fallback UI is unchanged.

## Test plan

- [ ] `npx nx typecheck twenty-front` passes
- [ ] `npx nx lint:diff-with-main twenty-front` passes
- [ ] When a widget renderer throws, the "Invalid Configuration" chip still
      renders and the error now appears in the browser console / Sentry


charlesBochet:
Replaced the onError/Sentry logging approach with surfacing the actual error message in the existing tooltip on hover, no need for extra console noise or Sentry dependency

charlesBochet (approved):
Thank you!

PR #21011, opened 2026-05-28
fix: reload stale clients on any older app version, not just major

clivemeister:
## Context

The GraphQL error handler emits an `APP_VERSION_MISMATCH` error (surfaced on the client as a "your app version is out of date, please refresh" prompt) when a client's app version is behind the server's. Today that comparison only fires when the **major** version differs:

```ts
const frontEndMajor = semver.parse(frontEndAppVersion)?.major;
const backendMajor = semver.parse(backendAppVersion)?.major;
if (isDefined(frontEndMajor) && isDefined(backendMajor) && frontEndMajor < backendMajor) { ... }
```

Twenty ships schema changes in minor and patch releases too. A tab left open across a minor/patch deploy keeps sending GraphQL documents built against the previous schema. If a field was removed or renamed, those operations fail with opaque field-level errors and the user never gets the refresh prompt — a hard reload is the only recovery.

## What this does

- **Server:** fires the mismatch whenever the client version is strictly older than the server (`semver.lt`), regardless of which version component changed.
- **Client:** `onAppVersionMismatch` now reloads the page once (in addition to the existing snackbar) to pull the current `index.html` and hashed assets, guarded by a short `sessionStorage` window so a still-stale reload can't cause a refresh loop.

## Notes

- No behavior change for clients on the same or newer version.
- Relies on the existing `x-app-version` header and `APP_VERSION` config that already drive this check.


charlesBochet:
Dropped the client-side auto-reload, the existing snackbar already prompts the user to refresh. Kept the server-side semver.lt change which only fires when a query has already failed validation, so it just turns a cryptic field error into a clear refresh prompt and has no impact on stale clients whose queries still validate

charlesBochet (dismissed):
Thank you! I have removed the change on frontend, I feel it's overkill, let's just only report version mismatch (including full semver)

charlesBochet (approved):

PR #21017, opened 2026-05-28
Fix admin pannel server variable config tab

martmull:
## Before

image

## After

image


packages/twenty-server/src/engine/core-modules/admin-panel/admin-panel.resolver.ts
```diff
@@ -725,6 +727,16 @@ export class AdminPanelResolver {
     );
   }
 
+  @UseGuards(AdminPanelGuard)
+  @Mutation(() => ApplicationRegistrationVariableEntity)
```
Weiko: This should return a DTO, not an Entity otherwise we might expose things we don't want. DTO is there to control which field is exposed.

packages/twenty-front/src/modules/settings/application-registrations/components/ApplicationRegistrationConfigVariableEditForm.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  return (
+    <ConfigVariableEdit
+      title={variable.key}
+      description={variable.description}
+      input={
+        <TextInput
+          value={value}
+          placeholder={
+            !isEditing ? (variable.value ?? t`set-config-value`) : ''
```
Weiko: t`set-config-value` < Not sure how this one will translate tbh. Probably fine but maybe we can find a better placeholder 🤔

packages/twenty-front/src/pages/settings/admin-panel/SettingsAdminApplicationRegistrationConfigVariableDetail.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  const [updateVariable] = useMutation(
+    UpdateAdminApplicationRegistrationVariableDocument,
+    {
+      client: apolloAdminClient,
+      refetchQueries: [FindAdminApplicationRegistrationVariablesDocument],
+    },
+  );
+
+  if (!variable || !registration) {
```
Weiko: We probably want to distinguish "loading" (registration) from "not found" (variable)?

packages/twenty-front/src/generated-admin/graphql.ts
```diff
@@ -195,6 +195,18 @@ export enum ApplicationRegistrationSourceType {
   TARBALL = 'TARBALL'
 }
 
+export type ApplicationRegistrationVariable = {
```
Weiko: Why do we have both this one and ApplicationRegistrationVariableDto?

Weiko: Ok, because you return the Entity in the resolver. This shouldn't be there after you fix my other comment

Weiko (commented):
Left some comments

Weiko (approved):

PR #21018, opened 2026-05-28
fix(contact-creation): enrich missing names on auto-created contacts

FelixMalfait:
## Summary

Three related fixes to the auto-creation of People records from calendar events and email messages, all centred on the data-quality problem of contacts being created with missing or malformed names.

### 1. Enrich names on existing contacts (commit 1)

Previously: when an email or calendar import matched an existing Person by email, the existing record was left untouched — even if the new source carried a better name.

This is the root cause of contacts like `"Félix"` (no last name) sticking around forever: the `To:`/`Cc:` headers of outbound emails rarely include a display name, and Google Calendar only returns `displayName` for attendees already in the organizer's address book. So the first sighting often creates a Person as `{firstName: "felix", lastName: ""}`, and a later inbound `From: "Félix Malfait" <felix@twenty.com>` — which would have produced the right name — gets silently dropped because the Person already exists.

The new `computePeopleToEnrichNames` bucket and `CreatePersonService.enrichPeopleNames` method fill in missing `firstName`/`lastName` fields from the new parsed name, with conservative rules:
- Only enrich when the existing Person's `createdBy.source` is `CALENDAR` or `EMAIL` — `MANUAL`, `IMPORT`, `API`, `WORKFLOW`, etc. are never touched.
- Only fill empty fields. Non-empty `firstName`/`lastName` are never overwritten.
- Soft-deleted contacts continue to be handled by the existing restore path.

### 2. Handle multi-comma "Last, First, Suffix" display names (commit 2)

The comma-inverted swap in the parser previously required *exactly* one comma. Names like `"Smith, Jane, Jr."`, `"O'Brien, Mary, MD"` or `"Doe, John, Patrick"` fell through to the space-split fallback, which stored the comma in `firstName` (e.g. `"Smith,"`) and produced garbled records (the avatar shows a single "B" and the name reads `"Barbey, Julien"` because the entire string lives in `firstName`).

The regex now splits on the first comma and treats the remainder as the first name, collapsing any further commas to spaces. Single-comma behaviour is unchanged.

### 3. Perf: skip the parser when an existing record is already populated (commit 3)

`computePeopleToEnrichNames` runs on every cron-driven email/calendar import batch. The first version called the display-name parser for every matched existing person, even when both `firstName` and `lastName` were already set — i.e. the steady-state case after the initial enrichment pass.

Reordered so the cheap "both fields populated" check short-circuits before any parsing happens. Same behaviour, fewer parser calls on the hot path.

## Test plan

- [x] 8 new unit tests for the enrichment bucket: empty `lastName` enrichment, both `EMAIL` and `CALENDAR` sources, non-overwrite of non-empty fields, skip on `MANUAL`/`IMPORT`, skip when the new source also has no last name, skip for soft-deleted, fill `firstName` while preserving `lastName`, handle null `name` field
- [x] 5 new parser tests for multi-comma forms: `"Last, First, Suffix"`, credential suffixes (`MD`), three-token forms, whitespace around inner commas, `:GROUP` tag interaction
- [x] 1 new parser test on the single-comma path covering multi-word first names (`"Smith, Mary Jane"`)
- [x] All 15 existing parser tests still pass
- [x] All 116 tests in `contact-creation-manager` pass
- [x] `npx oxlint --type-aware` + `npx oxfmt --check` on changed files
- [ ] Manual: trigger a fresh contact creation from an outbound email with no display name, then a subsequent inbound email from the same address with a full display name, and confirm the Person's last name gets populated

### DIFF
--- a/packages/twenty-server/src/modules/contact-creation-manager/services/create-company-and-contact.service.ts
+++ b/packages/twenty-server/src/modules/contact-creation-manager/services/create-company-and-contact.service.ts
@@ -350,8 +350,10 @@ export class CreateCompanyAndPersonService {
       { existingPerson: PersonWorkspaceEntity }
     >,
   ): { personId: string; name: FullNameMetadata }[] {
-    const peopleToEnrichNames: { personId: string; name: FullNameMetadata }[] =
-      [];
+    const peopleToEnrichNamesByPersonId = new Map<
+      string,
+      { personId: string; name: FullNameMetadata }
+    >();
 
     for (const contact of uniqueContacts) {
       const existingPerson = shouldCreateOrRestorePeopleByHandleMap.get(
@@ -359,6 +361,14 @@ export class CreateCompanyAndPersonService {
       )?.existingPerson;
 
       if (!isDefined(existingPerson) || !isNull(existingPerson.deletedAt)) {
+        continue;
+      }
+
+      // A single Person can match multiple contacts in the same batch when
+      // its primaryEmail and additionalEmails both appear in the import.
+      // Keep the first enrichment we compute so updateMany doesn't issue
+      // two UPDATEs against the same row with order-dependent winners.
+      if (peopleToEnrichNamesByPersonId.has(existingPerson.id)) {
         continue;
       }
 
@@ -398,7 +408,7 @@ export class CreateCompanyAndPersonService {
         continue;
       }
 
-      peopleToEnrichNames.push({
+      peopleToEnrichNamesByPersonId.set(existingPerson.id, {
         personId: existingPerson.id,
         name: {
           firstName: shouldEnrichFirstName
@@ -409,7 +419,7 @@ export class CreateCompanyAndPersonService {
       });
     }
 
-    return peopleToEnrichNames;
+    return Array.from(peopleToEnrichNamesByPersonId.values());
   }
 
   formatPeopleToCreateFromContacts({
--- a/packages/twenty-server/src/modules/contact-creation-manager/services/create-company-and-contact.service.ts
+++ b/packages/twenty-server/src/modules/contact-creation-manager/services/create-company-and-contact.service.ts
@@ -339,10 +339,10 @@ export class CreateCompanyAndPersonService {
     };
   }
 
-  // Only enrich names on contacts that the auto-creation pipeline itself
-  // created (CALENDAR or EMAIL). Manually created, imported, API-created and
-  // other curated sources stay untouched. We only fill in fields that are
-  // currently empty — never overwrite existing data.
+  // Stages per-personId name enrichments for existing People auto-created via
+  // CALENDAR or EMAIL. Empty fields are filled from new sources (first
+  // non-empty value wins across multiple contacts mapping to the same Person);
+  // populated fields are never overwritten.
   private computePeopleToEnrichNames(
     uniqueContacts: Contact[],
     shouldCreateOrRestorePeopleByHandleMap: Map<
@@ -350,11 +350,6 @@ export class CreateCompanyAndPersonService {
       { existingPerson: PersonWorkspaceEntity }
     >,
   ): { personId: string; name: FullNameMetadata }[] {
-    // Stage enrichments by personId. A single Person can match more than one
-    // contact in the same batch when its primaryEmail and additionalEmails
-    // both appear in the import — we merge field-by-field so a partial first
-    // match (e.g. only a firstName) doesn't lock out a later contact from
-    // filling the still-empty lastName.
     const enrichmentByPersonId = new Map<
       string,
       { firstName: string; lastName: string }
@@ -369,11 +364,8 @@ export class CreateCompanyAndPersonService {
         continue;
       }
 
-      // Soft-deleted matches are about to be restored in this same batch (see
-      // contactsThatNeedPersonRestore above), so enrich them too — restore
-      // runs before enrich in createCompaniesAndPeople, which means the row
-      // is no longer soft-deleted by the time the enrichment UPDATE fires.
-
+      // Soft-deleted matches are restored earlier in the same job, so the
+      // enrichment UPDATE runs against an un-deleted row.
       const existingSource = existingPerson.createdBy?.source;
 
       if (
@@ -383,8 +375,6 @@ export class CreateCompanyAndPersonService {
         continue;
       }
 
-      // Take staged values if we've already touched this Person in this batch,
-      // otherwise start from what the DB already has.
       const staged = enrichmentByPersonId.get(existingPerson.id);
       const currentFirstName =
         staged?.firstName ?? existingPerson.name?.firstName ?? '';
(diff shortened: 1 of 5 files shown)


packages/twenty-server/src/modules/contact-creation-manager/services/create-person.service.ts
```diff
@@ -89,6 +90,42 @@ export class CreatePersonService {
...
+        const personRepository =
+          await this.globalWorkspaceOrmManager.getRepository(
+            workspaceId,
+            PersonWorkspaceEntity,
+            {
+              shouldBypassPermissionChecks: true,
+            },
+          );
+
+        const enrichedPeople = await personRepository.updateMany(
```
Weiko, line 114: possible race condition if persons were updated in the meantime but we don't really have strong locking/optimistic strategy currently (or at least easy to use that wouldn't be over-engineered)

packages/twenty-server/src/modules/contact-creation-manager/services/create-company-and-contact.service.ts
```diff
@@ -321,11 +333,77 @@ export class CreateCompanyAndPersonService {
...
+        !isNonEmptyString(existingFirstName) &&
+        isNonEmptyString(parsedFirstName);
+      const shouldEnrichLastName =
+        !isNonEmptyString(existingLastName) && isNonEmptyString(parsedLastName);
+
+      if (!shouldEnrichFirstName && !shouldEnrichLastName) {
+        continue;
+      }
+
+      peopleToEnrichNames.push({
```
Weiko: is it possible that we end up there with 2 contacts sharing the same personId? Then the updateMany later will take only depending on the ordering? Probably fine 🤔

Weiko (approved):
LGTM

packages/twenty-server/src/modules/contact-creation-manager/services/create-company-and-contact.service.ts
```diff
@@ -321,11 +333,85 @@ export class CreateCompanyAndPersonService {
...
+      ) {
+        continue;
+      }
+
+      const existingFirstName = existingPerson.name?.firstName ?? '';
+      const existingLastName = existingPerson.name?.lastName ?? '';
+      const existingFirstNameIsEmpty = !isNonEmptyString(existingFirstName);
+      const existingLastNameIsEmpty = !isNonEmptyString(existingLastName);
+
+      // Skip the parser when the existing record is already fully populated.
```
FelixMalfait: Remove comment

packages/twenty-server/src/modules/contact-creation-manager/services/create-company-and-contact.service.ts
FelixMalfait: Good catch — this is a real bug. A Person with both `primaryEmail` and `additionalEmails` can match two contacts in the same batch (different display names), so we'd push two entries with the same `personId` and `updateMany` would do two UPDATEs with the last write winning. Fixed in 81e9820 by deduping into a `Map<personId, ...>` keyed by personId — first computed enrichment wins per Person, with a test covering the case.

packages/twenty-server/src/modules/contact-creation-manager/services/create-person.service.ts
FelixMalfait: Acknowledged — the SELECT-then-UPDATE window means a concurrent manual edit between the cron's read of `alreadyCreatedPeople` and our `updateMany` could be overwritten. Closing it properly would need a conditional UPDATE (`WHERE name->>'firstName' = '' OR IS NULL`) which `updateMany` doesn't currently support, or per-row `update()` calls — both feel over-engineered for the rarity. Leaving as-is in line with your note; happy to come back to it if it becomes an issue.

packages/twenty-server/src/modules/contact-creation-manager/services/create-company-and-contact.service.ts
```diff
@@ -321,11 +333,95 @@ export class CreateCompanyAndPersonService {
...
+
+      if (!isDefined(existingPerson) || !isNull(existingPerson.deletedAt)) {
+        continue;
+      }
+
+      // A single Person can match multiple contacts in the same batch when
+      // its primaryEmail and additionalEmails both appear in the import.
+      // Keep the first enrichment we compute so updateMany doesn't issue
+      // two UPDATEs against the same row with order-dependent winners.
+      if (peopleToEnrichNamesByPersonId.has(existingPerson.id)) {
```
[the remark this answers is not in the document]
FelixMalfait: Good catch — fixed in 61de500. Switched from "first wins" dedupe to per-field merge: each contact mapping to the same Person can fill any field still empty in the staged value, but never overwrites a populated one. Still emits at most one UPDATE per Person per batch, and added a test (`should merge partial enrichments across contacts mapping to the same Person`) covering the exact scenario.

packages/twenty-server/src/modules/contact-creation-manager/services/create-company-and-contact.service.ts
```diff
@@ -321,11 +333,101 @@ export class CreateCompanyAndPersonService {
...
+
+    for (const contact of uniqueContacts) {
+      const existingPerson = shouldCreateOrRestorePeopleByHandleMap.get(
+        contact.handle.toLowerCase(),
+      )?.existingPerson;
+
+      if (!isDefined(existingPerson) || !isNull(existingPerson.deletedAt)) {
+        continue;
+      }
+
```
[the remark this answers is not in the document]
FelixMalfait: Good catch — fixed in 4a07b0d. Dropped the soft-deleted filter in `computePeopleToEnrichNames`: every soft-deleted match in this batch is already going through `restorePeople` (which only touches `deletedAt`/`companyId`, not the name fields), and restore runs before enrich in `createCompaniesAndPeople` — so by the time the enrichment UPDATE fires, the row is no longer soft-deleted. The existing test for this case now asserts the soft-deleted contact gets enriched on restore.

packages/twenty-server/src/modules/contact-creation-manager/services/create-person.service.ts
```diff
@@ -89,6 +90,42 @@ export class CreatePersonService {
...
+          peopleToEnrich.map(({ personId, name }) => ({
+            criteria: personId,
+            partialEntity: { name },
+          })),
+          undefined,
+          ['id'],
+        );
```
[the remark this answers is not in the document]
FelixMalfait: Same concern Weiko flagged earlier on this PR (resolved [thread](https://github.com/twentyhq/twenty/pull/21018#discussion_r3318313393)) — acknowledged, not fixing. Closing the window properly needs a conditional UPDATE (`WHERE name->>'firstName' IS NULL/empty`) which the workspace ORM's `updateMany` doesn't currently support, or per-row `update()` calls. Both feel over-engineered given the small window between SELECT and UPDATE in the cron job and the conservative scope (only fills empty fields on auto-created records). Happy to revisit if it turns into a real-world issue.

PR #21024, opened 2026-05-28
Fix playwright CI

Weiko:
## Context
The Install Playwright step ran npx playwright install with no arguments, which downloads all browsers (Chromium + Firefox + WebKit + ffmpeg, ~500MB+) on every run with no caching.

Fix:
- Install Chromium only — npx playwright install chromium instead of all browsers.
- Cache the browser binaries — actions/cache on ~/.cache/ms-playwright, keyed on the resolved Playwright version (v4-playwright-browsers-${{ runner.os }}-<version>). On a cache hit the install step is skipped entirely; the cache invalidates automatically when the Playwright version bumps.


prastoin (approved):
Could use twenty save cache composite actions, that are redundantly safe for external contrib fork cache interaction

PR #21025, opened 2026-05-28
Allow kanban cross-column drag when sorting is enabled

Val4evr:
## Summary

This PR allows kanban cards to be dragged across columns while sorting is enabled.

Previously, any board drag while a sort was active opened the “Remove sorting?” modal. That makes sense for same-column reordering, because manual reorder conflicts with the active sort. But for cross-column moves, the user is changing the grouped field, not trying to manually reorder the destination column.

With this change:

- Same-column drag with sorting enabled still opens the existing remove-sorting modal.
- Cross-column drag with sorting enabled updates only the group field.
- The destination column keeps using the active sort to determine where the card appears.
- Unsorted board drag behavior continues to update `position` as before.

## Why

On sorted kanban boards, moving a card to another column is a valid workflow even though manual reordering is not. The previous guard blocked both cases because it only checked whether sorting was active, not whether the card stayed inside the same column.

## Implementation

The drop behavior now distinguishes between:

- sorted same-column drops, which remain blocked
- sorted cross-column drops, which are allowed without a position update
- unsorted drops, which keep the existing position-update behavior

A small helper captures that decision and has focused unit coverage.

## Validation

- Manually verified sorted cross-column drag persists after refresh.
- Manually verified sorted same-column drag still opens the remove-sorting modal.
- Manually verified unsorted same-column drag still reorders cards.
- Manually verified unsorted cross-column drag still moves cards.
- Ran focused Jest coverage for the sorted board drop decision.
- Ran formatting and oxlint checks on touched frontend files.
- Ran `twenty-front` typecheck.
- Ran `twenty-front` production build.


FelixMalfait:
This is something users have complained about multiple times. I haven't tested your code but I agree on the product direction you describe, thanks for raising a fix.

I'll let the code review to someone in our team. @thomtrp @ijreilly maybe? Thanks!

martmull:
@Bonapara can you validate the functionality? Code looks ok to me

Bonapara:
@Val4evr Can you join a video recording to your PR?

FelixMalfait:
@Bonapara it's a good change, I feel like the description is more clear than what a video would bring (it's just about showing vs not showing the error, we wouldn't see much more than what's described?, @Val4evr did a great job listing the why/validation)

valeriyproklov:
> @Val4evr Can you join a video recording to your PR?

Sorry for the wait @Bonapara 

https://github.com/user-attachments/assets/30ab02d4-0588-4c0a-8fab-875edc1b24c7

FelixMalfait:
@Bonapara merging because this has been opened for long without feedback from your side. For me the direction is correct.

PR #21030, opened 2026-05-28
Fix AI permission gating: use Ask AI for chat UI, AI Settings for admin endpoints

### LINKED ISSUE #20662: Non-admin users with "Ask AI" enabled still cannot access AI chats without broader AI settings permission
## Description

On a self-hosted Twenty `v2.5.1` deployment, non-admin users cannot reliably use AI chats even when their role has the `Ask AI` capability enabled.

In our case, AI chat access appears to remain blocked unless the same role is also granted the broader workspace AI settings permission (`AI` / `AI_SETTINGS`, "Create and configure AI agents"). This makes least-privilege role setup difficult, because a user who should only be allowed to use AI chat may also need a more administrative AI permission to see or use the feature.

We also observed repeated `THREAD_NOT_FOUND` GraphQL/runtime errors when AI chat-related queries fire for such non-admin users.

As of **May 18, 2026**, I checked the public issue tracker and did not find a clearly matching open issue for this exact non-admin `Ask AI` / `AI_SETTINGS` behavior on newer versions.

## Environment

- Twenty version: `v2.5.1`
- Deployment type: self-hosted via Docker Compose
- Role model: non-admin role with granular permissions
- AI is configured and available for admins
- Non-admin test users are in the same workspace

## Steps to Reproduce

1. Configure AI correctly so that AI chats work for an admin user.
2. Create or edit a non-admin role.
3. Enable the `Ask AI` action for that role.
4. Do **not** enable the broader workspace AI settings permission (`AI` / `AI_SETTINGS` / "Create and configure AI agents").
5. Assign that role to a non-admin user.
6. Log in as that user and try to access AI chat entry points / Ask AI flow.

## Expected Behavior

A user with the `Ask AI` permission should be able to access and use AI chats without requiring broader AI administration/settings permissions.

## Actual Behavior

One or more of the following happens for the non-admin user:

- AI chat entry points are missing from the UI
- AI chat navigation is not available
- AI-related queries still fire, but the user cannot actually use the chat flow
- backend/runtime logs show repeated `THREAD_NOT_FOUND` errors related to AI chat/thread state

## Notes

This may be related to permission gating around AI chat navigation and AI settings.

Relevant upstream history that may be related:

- `v1.22.0` release notes mention: `Fix AI chat threads query firing for users without AI permissions` (`#19507`)
- `v2.2.0` release notes mention: `Gate AI chat navigation entries by AI_SETTINGS permission` (`#20089`)

From a product/permissions perspective, the current behavior is confusing because `Ask AI` appears insufficient on its own for non-admin users.

## Impact

This makes AI chat usage unreliable for role-scoped workspace members and weakens least-privilege permission design, because exposing AI chat may require granting broader AI configuration access.

## Questions

- Is this current behavior expected?
- Should `Ask AI` alone be enough to access AI chats?
- If `AI_SETTINGS` is intentionally required, should that be documented more explicitly in the permissions model/UI?
- If not expected, should AI chat queries and navigation be fully gated so non-admin users do not hit broken states or `THREAD_NOT_FOUND` errors?

FelixMalfait:
## Summary

Closes #20662.

Two AI permission flags exist:
- **`AI`** (label "Ask AI") — user-facing: chat with AI agents, use AI features
- **`AI_SETTINGS`** (label "AI") — admin: create and configure AI agents

After auditing every use of these flags I found:

### Frontend — chat UI gated by the admin permission (user-facing bug from the issue)
A user granted only `Ask AI` could not see chat tabs, the "new chat" button (desktop & mobile), or the chat content pane; thread initialization was also skipped, leaving the chat in a half-initialized state and producing intermittent `THREAD_NOT_FOUND` errors. Switched these to `AI`:
- `MainNavigationDrawerTabsRow.tsx`
- `MainNavigationDrawer.tsx`
- `MobileNavigationBar.tsx`
- `AgentChatThreadInitializationEffect.tsx`

### Backend — admin-only resolvers gated by the user permission (privilege escalation)
Two resolvers had a class-level guard of `AI`, letting any user with the user-facing flag reach admin endpoints (skill CRUD, eval runs). Switched the class-level guards to `AI_SETTINGS`:
- `SkillResolver` — create/update/delete/activate/deactivate skills
- `AgentTurnResolver` — read turns, run/grade evaluations

### Left as-is (already correct)
- `AgentResolver` — class-level `AI` for reads (workflow editors and admin pages both need them), mutation-level `AI_SETTINGS` overrides for writes
- `AgentChatResolver` & `AgentChatSubscriptionResolver` — already `AI`
- `AiGenerateTextController` — already `AI`
- Workspace AI config fields in `workspace.service.ts` — already `AI_SETTINGS`

## Test plan

- [ ] As a user with `Ask AI` only (no `AI_SETTINGS`): chat tabs, "new chat" button, and chat history pane are visible on desktop + mobile; sending a message works; no `THREAD_NOT_FOUND` errors
- [ ] As a user with `AI_SETTINGS` but no `Ask AI`: chat UI is hidden
- [ ] As a user with `Ask AI` only: calling `skills` / `createSkill` / `agentTurns` / `runEvaluationInput` via GraphQL returns permission denied
- [ ] As an admin (`AI_SETTINGS`): skill settings and agent eval pages still work

### DIFF
--- a/packages/twenty-front/src/modules/ai/components/AgentChatThreadInitializationEffect.tsx
+++ b/packages/twenty-front/src/modules/ai/components/AgentChatThreadInitializationEffect.tsx
@@ -73,13 +73,7 @@ export const AgentChatThreadInitializationEffect = () => {
         replaceDraft('agentChatThreads', result.data.chatThreads);
         applyChanges();
       });
-  }, [
-    storeEntry.status,
-    hasAiPermission,
-    client,
-    replaceDraft,
-    applyChanges,
-  ]);
+  }, [storeEntry.status, hasAiPermission, client, replaceDraft, applyChanges]);
 
   useEffect(() => {
     setAgentChatThreadsLoading(
--- a/packages/twenty-front/src/modules/ai/components/AgentChatThreadInitializationEffect.tsx
+++ b/packages/twenty-front/src/modules/ai/components/AgentChatThreadInitializationEffect.tsx
@@ -31,9 +31,7 @@ import {
 export const AgentChatThreadInitializationEffect = () => {
   const client = useApolloClient();
   const { replaceDraft, applyChanges } = useUpdateMetadataStoreDraft();
-  const hasAiSettingsPermission = useHasPermissionFlag(
-    PermissionFlagType.AI_SETTINGS,
-  );
+  const hasAiPermission = useHasPermissionFlag(PermissionFlagType.AI);
 
   const currentAiChatThread = useAtomStateValue(currentAiChatThreadState);
   const setCurrentAiChatThread = useSetAtomState(currentAiChatThreadState);
@@ -58,7 +56,7 @@ export const AgentChatThreadInitializationEffect = () => {
     useAtomState(hasInitializedAgentChatThreadsState);
 
   useEffect(() => {
-    if (storeEntry.status !== 'empty' || !hasAiSettingsPermission) {
+    if (storeEntry.status !== 'empty' || !hasAiPermission) {
       return;
     }
 
@@ -77,7 +75,7 @@ export const AgentChatThreadInitializationEffect = () => {
       });
   }, [
     storeEntry.status,
-    hasAiSettingsPermission,
+    hasAiPermission,
     client,
     replaceDraft,
     applyChanges,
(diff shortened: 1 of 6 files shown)


etiennejouan:
LGTM !

PR #21037, opened 2026-05-29
fix(ai): route xAI search through Responses API as native tools

ehconitin:
xAI deprecated Live Search, so the `searchParameters` provider option now returns 410. This routes all xAI models through the Responses API and binds web/X search as native agent tools, matching how Anthropic/OpenAI expose search.

- xAI provider now uses `provider.responses()` — its `webSearch()`/`xSearch()` tools only run against the Responses endpoint, not chat completions
- web/X search migrated from the `provider-option` variant to `sdk-tool` (`web_search`/`x_search`); deleted the dead `searchParameters` path, the `provider-option` variant, and `providerOptions` on `NativeModelBinding`
- dropped a dead `rolePermissionConfig` param on `getAgentRoleId`, left over from #20331

### DIFF
--- a/packages/twenty-server/src/engine/metadata-modules/ai/ai-models/services/ai-model-config.service.ts
+++ b/packages/twenty-server/src/engine/metadata-modules/ai/ai-models/services/ai-model-config.service.ts
@@ -16,7 +16,6 @@ import {
   RegisteredAiModel,
 } from 'src/engine/metadata-modules/ai/ai-models/services/ai-model-registry.service';
 import { SdkProviderFactoryService } from 'src/engine/metadata-modules/ai/ai-models/services/sdk-provider-factory.service';
-import { type NativeModelBinding } from 'src/engine/metadata-modules/ai/ai-models/types/native-model-binding.type';
 import { type NativeModelToolOptions } from 'src/engine/metadata-modules/ai/ai-models/types/native-model-tool-options.type';
 import { getNativeModelToolsForSdkPackage } from 'src/engine/metadata-modules/ai/ai-models/utils/get-native-model-tools-for-sdk-package.util';
 
@@ -26,15 +25,6 @@ export class AiModelConfigService {
     private readonly aiModelRegistryService: AiModelRegistryService,
     private readonly sdkProviderFactory: SdkProviderFactoryService,
   ) {}
-
-  getNativeModelBinding(
-    model: RegisteredAiModel,
-    options: NativeModelToolOptions = {},
-  ): NativeModelBinding {
-    return {
-      tools: this.getNativeModelTools(model, options),
-    };
-  }
 
   getReasoningProviderOptions(model: RegisteredAiModel): ProviderOptions {
     switch (model.sdkPackage) {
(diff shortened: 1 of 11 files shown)


packages/twenty-server/src/engine/metadata-modules/ai/ai-models/services/sdk-provider-factory.service.ts
```diff
@@ -121,6 +125,19 @@ export class SdkProviderFactoryService {
     };
   }
 
+  private buildXaiProvider(config: AiProviderConfig): AiSdkProviderInstance {
```
FelixMalfait, line 128: @claude why couldn't we use the buildStandardProvider? It's not clear for me on first read why the author introduced buildXaiProvider and didn't reuse buildStandardProvider

packages/twenty-server/src/engine/metadata-modules/ai/ai-models/types/native-model-binding.type.ts
```diff
@@ -1,7 +1,5 @@
-import { type ProviderOptions } from '@ai-sdk/provider-utils';
 import { type ToolSet } from 'ai';
 
 export type NativeModelBinding = {
```
FelixMalfait: This type feels superfluous ; can't we just use ToolSet if it's just that?

packages/twenty-server/src/engine/metadata-modules/ai/ai-models/services/ai-model-config.service.ts
```diff
@@ -56,112 +53,66 @@ export class AiModelConfigService {
...
-          tools[webSearchTool.directToolName] =
-            anthropicProvider.tools.webSearch_20250305();
+        if (options.webSearch === true && isDefined(nativeTools.webSearch)) {
+          const anthropicProvider =
+            this.sdkProviderFactory.getRawAnthropicProvider(providerName);
+
+          if (isDefined(anthropicProvider)) {
+            tools[nativeTools.webSearch.directToolName] =
+              anthropicProvider.tools.webSearch_20250305();
+          }
```
FelixMalfait, line 62: @claude after this changes? Is websearch still enabled by default in the ai chat for compatible models? I think that was the case before, just want to double check

packages/twenty-server/src/engine/metadata-modules/ai/ai-models/services/sdk-provider-factory.service.ts
FelixMalfait: @claude just for my personal knowledge can you please explain the differences between response and chat api. Which one is more used and recommended by providers? Can you lookup through the code and see if we make a good use of it and don't use the wrong api for the given use-case anywhere?

packages/twenty-server/src/engine/metadata-modules/ai/ai-models/services/ai-model-config.service.ts
FelixMalfait: @claude after this changes is websearch still enabled by default in the ai chat for compatible models? I think that was the case before, just want to double check

packages/twenty-server/src/engine/metadata-modules/ai/ai-models/types/native-model-binding.type.ts
FelixMalfait: @claude fix this, make sure you lint/typecheck and push to the branch directly

PR #21038, opened 2026-05-29
i18n(ar-SA): complete translations, fix corruption, normalize terminology

ahmadoqsrawi:
## What this does

Completes the Arabic (`ar-SA`) translations in `twenty-front` and the server-side `twenty-server/i18n` locale files, and fixes a few bugs that were rendering as broken text in the Arabic UI.

## Before

| File | Empty entries | Total |
|---|---|---|
| `twenty-front/src/locales/ar-SA.po` | 679 (~21%) | 3,181 |
| `twenty-server/.../i18n/locales/ar-SA.po` | 451 (~29%) | 1,532 |

A handful of entries also had corrupted msgstr values that were being rendered to users as literal escape characters. The most visible one was `Settings`, stored as `\ا\ل\إ\ع\د\ا\د\ا\ت`, which appeared in the sidebar with a backslash between every Arabic letter. 18 other entries had `msgstr "\\\\"` instead of a real translation, including Privacy Policy, Profile, Roles, Security, Search, Results, Remove, Releases, Regenerate Key, Send invite email, and a few others.

A separate group of bugs lived in entries that *were* translated but had small breakage:
- `: Future`, `: NotEmpty`, `: NotNull`, `: Past` filter operand labels dropped the `: ` prefix, gluing the field name straight to the operand in the chip.
- `: NotNull` left the English word "null" inside the Arabic string.
- `" billed annually"` had lost its leading space and ran together with the price on the trial card.

On the server side, several standard object and widget names were empty, so even with the Arabic locale active, the sidebar and admin screens showed English for People, Opportunities, Workflows, Person, Opportunity, Task, and ~25 others.

## What changed

1. Filled every empty Arabic msgstr in both PO files. Worked in topical batches (billing, roles, workflows, AI agents, dashboard, admin panel, apps, data model, settings, misc) and reviewed each batch before applying.
2. Fixed the corruption bugs above.
3. Fixed the mechanical bugs in already-translated entries (operand labels, "billed annually", "NotNull").
4. Filled in the missing standard object and widget names on the server side.

## Terminology pass

Same English term now maps to the same Arabic word across both files. Highlights:

| EN | AR | Notes |
|---|---|---|
| object | عنصر | was `كائن` in 142 places |
| widget | ودجت | 5 spots used `عنصر واجهة` |
| feature flags | علامات الميزات | a few used `أعلام`, which means banners |
| tool (AI agent) | أداة | |
| prompt (AI) | الموجّه | |
| credits (SaaS quota) | الأرصدة | |
| token (AI/LLM) | رمز | |
| agent (AI) | وكيل | |
| Enterprise (tier name) | Enterprise | kept as English |

All 18 audited glossary terms verified as having a single Arabic translation across both PO files after the sweep.

## Arabic plural handling

Where the English source uses `{count, plural, one {...} other {...}}`, the Arabic translation expands to the full CLDR set of forms (`zero`, `one`, `two`, `few`, `many`, `other`). Arabic genuinely uses these forms in everyday speech (for example, the form for "2 items" is different from "3 items" is different from "11 items"), so the runtime gets to pick the right arm for the actual count instead of always falling back to `other`.

## Numbers

| File | Empty before | Empty after | Total |
|---|---|---|---|
| `twenty-front/src/locales/ar-SA.po` | 679 | **0** | 3,181 |
| `twenty-server/.../i18n/locales/ar-SA.po` | 451 | **0** | 1,532 |

Diff: +1,346 / -1,293 across the two files.

## Validation

- Every Lingui placeholder (`{0}`, `{fieldName}`, ICU plural arms, `<0>...</0>` React component slots) was checked to be present in both `msgid` and `msgstr`.
- Glossary consistency was re-scanned after the bulk apply.
- Spot-checked the result in the Arabic-direction (RTL) flow on a self-hosted instance for filter chips, sidebar object names, and the billing copy that the mechanical bugs touched.


Weiko (commented):
Hi @ahmadoqsrawi,
Translations are handled asynchronously via Crowdin.
If you want to contribute, you need to do that on https://twenty.crowdin.com/ and not in the repo, otherwise your changes would be overwritten by Crowdin.

ahmadoqsrawi:
Thanks for the heads-up, I'll redo this through twenty.crowdin.com so it doesn't get overwritten.

PR #21042, opened 2026-05-29
Fix latency spike on application lookup

martmull:
Fixes https://discord.com/channels/1130383047699738754/1509645089062781058

Caching application entities to improve authentication latency


packages/twenty-server/src/engine/core-modules/application/application-entity-cache-provider.service.ts
```diff
@@ -0,0 +1,31 @@
...
+import { CoreEntityCache } from 'src/engine/core-entity-cache/decorators/core-entity-cache.decorator';
+import { type FlatApplication } from 'src/engine/core-modules/application/types/flat-application.type';
+import { InjectRepository } from '@nestjs/typeorm';
+import { Repository } from 'typeorm';
+import { ApplicationEntity } from 'src/engine/core-modules/application/application.entity';
+import { fromApplicationEntityToFlatApplication } from 'src/engine/core-modules/application/utils/from-application-entity-to-flat-application.util';
+
+@Injectable()
+@CoreEntityCache('applicationEntity')
+export class ApplicationEntityCacheProviderService extends CoreEntityCacheProvider<FlatApplication> {
```
Weiko: There is already a cache for application entity. The cache key is flatApplicationMaps and the service is WorkspaceFlatApplicationMapCacheService. Unless I'm missing something I don't think you need to add one @martmull

martmull: oh ok

Weiko (approved):
LGTM

PR #21049, opened 2026-05-29
Fix: Not able to add multiple handles to Blocklist in settings/accounts

### LINKED ISSUE #21031: Can't add multiple domains to blocklist
**Scenario:**
1. Log in
2. Go to Settings > Accounts
3. Type few domains separated with comma (e.g. `@apple.com, @test.com`)
4. Save changes

**Actual:** Multiple domains can't be saved despite placeholder suggesting it's possible to save multiple domains at the same time

**Expected:** Multiple domains are separated and saved

git-init-priyanshu:
Fixes: #21031

### Root cause:
This feature was never fully implemented even though the placeholder text suggested it was supported. It could only update one handle at a time.

### Fix
Fixed the zod validation to validate each handle separately.
Used `useCreateManyRecords` to update multiple handles at the same time.

### Before:
Screenshot 2026-05-29 at 3 25 12 PM

### After:

https://github.com/user-attachments/assets/69354129-422a-41de-baf7-fa5a28f01f3f


FelixMalfait:
Hey @git-init-priyanshu thanks! Could you please update the PR description to better describe the root cause. When was this bug introduced? Was it always like that? Why? Maybe also  add a video of the before/after or something like that. Also link to the issue it fixes. Thanks a lot!

git-init-priyanshu:
> Hey @git-init-priyanshu thanks! Could you please update the PR description to better describe the root cause. When was this bug introduced? Was it always like that? Why? Maybe also add a video of the before/after or something like that. Also link to the issue it fixes. Thanks a lot!

Sorry about that, it is updated now.

packages/twenty-front/src/modules/settings/accounts/components/SettingsAccountsBlocklistInput.tsx
```diff
@@ -21,8 +21,14 @@ const StyledLinkContainer = styled.div`
   margin-right: ${themeCssVariables.spacing[2]};
 `;
 
+const parseHandles = (value: string): string[] =>
+  value
+    .split(',')
+    .map((handle) => handle.trim())
+    .filter((handle) => !!handle);
```
etiennejouan: nit: isNonEmptyString

etiennejouan: @claude please

etiennejouan (dismissed):
LGTM ! Thanks @git-init-priyanshu

PR #21052, opened 2026-05-29
fix: show relation field changes in the timeline

### LINKED ISSUE #20970: Timeline activities don't display the before/after value for relation field changes
## Bug Description

When updating a record, the timeline correctly shows a diff for most field types (text, address, date, etc.), but for **RELATION** fields (e.g. changing `person.company`) the timeline either:
- shows nothing at all (if the relation was the *only* changed field — the entire activity gets filtered out), or
- shows the other field changes from the same update *without* showing the company change.

So changing the company on a Person does not produce a "Company: Acme → Initech" diff line the way changing the address produces an address diff. The relation change is effectively invisible in the audit trail.

## Expected behavior

When `person.company` changes from Acme to Initech, the timeline entry on the Person record should show something like:

> Felix updated Company: **Acme** → **Initech** on John Doe

…the same way other field types render in `EventFieldDiff`. The same applies to any MANY_TO_ONE relation (opportunity.pointOfContact, task.assignee, etc.).

## Technical inputs

Root cause is in [`object-record-changed-values.ts:34-41`](https://github.com/twentyhq/twenty/blob/main/packages/twenty-server/src/engine/core-modules/event-emitter/utils/object-record-changed-values.ts#L34-L41):

```ts
if (
  key === 'updatedAt' ||
  key === 'searchVector' ||
  field?.type === FieldMetadataType.RELATION ||
  field?.type === FieldMetadataType.POSITION
) {
  return acc;
}
```

RELATION fields are explicitly skipped from the diff payload. POSITION was added with the same skip in #18777 with the explanation *"Position were not properly displayed because we never implemented a display for this"* — RELATION appears to have been skipped from the very first commit (`d145684966` "New Timeline") for the same underlying reason: there was no renderer wired up for relation diffs.

There is also a key-mismatch issue downstream — the formatted records carry the join-column name (`companyId`), but the frontend's `fieldMetadataItemMap` in [`EventFieldDiffContainer.tsx`](https://github.com/twentyhq/twenty/blob/main/packages/twenty-front/src/modules/activities/timeline-activities/rows/main-object/components/EventFieldDiffContainer.tsx#L21-L27) is indexed by field metadata *name* (`company`). So even if the backend stopped skipping the field, the frontend would currently throw `Cannot find field metadata item for field name companyId`.

### What needs to happen

1. **Backend** — stop skipping RELATION in `objectRecordChangedValues`, at least for MANY_TO_ONE relations. The diff payload should record both the before/after foreign-key values **and** enough cached info to render a human-readable label (probably the related record's label-identifier value at the time of the change, so the timeline stays meaningful even if the related record is later renamed or deleted). The pattern is similar to `linkedRecordCachedName` already used for `linked-note` / `linked-task` events.
2. **Frontend** — update `EventFieldDiffContainer` / `EventFieldDiff` so a diff entry keyed by the join column (`companyId`) resolves to the relation field metadata (`company`), and render the before/after using the cached related-record name from the payload (or by fetching the records if still around).
3. **Tests** — extend `object-record-changed-values.spec.ts` (which currently only covers scalar + POSITION skip) with a RELATION case once the behavior changes.

### Relevant files

- `packages/twenty-server/src/engine/core-modules/event-emitter/utils/object-record-changed-values.ts` — the explicit skip
- `packages/twenty-server/src/engine/twenty-orm/utils/format-twenty-orm-event-to-database-batch-event.util.ts` — calls the above and persists `properties.diff`
- `packages/twenty-server/src/modules/timeline/repositories/timeline-activity.repository.ts` — filters out empty diffs (so today, a relation-only change produces no row at all)
- `packages/twenty-front/src/modules/activities/timeline-activities/rows/main-object/components/EventRowMainObjectUpdated.tsx`
- `packages/twenty-front/src/modules/activities/timeline-activities/rows/main-object/components/EventFieldDiffContainer.tsx`
- `packages/twenty-front/src/modules/activities/timeline-activities/rows/main-object/components/EventFieldDiff.tsx`

### Reproduction

1. Open any Person record that has a Company set.
2. Change the Company to a different one (and don't touch any other field).
3. Look at the Person's timeline → no entry is created.
4. Repeat but also edit the Person's name in the same save → only the name change appears in the diff; the company change is missing.

parshipcy:
## Summary
- Fixes #20970 
- When we changed only a relation field (e.g. company on a person):
  - The diff builder skipped all RELATION fields -> empty diff -> no timeline row.
  - If we change company and something else, only the scalar field appeared; the company change was missing.
  - Even with a diff, the UI validated keys by field name (`company`) while join columns (`companyId`) could be filtered out.

## Solution
- For `MANY_TO_ONE`, compare join column values (`companyId`) and store the diff under the relation field name (`company`):
```
"company": {
  "before": { "id": "<old-id>" },
  "after": { "id": "<new-id>" }
}
```
- Frontend
  - `filterOutInvalidTimelineActivities`: resolve diff keys by field name or join column via `findFieldMetadataItemByDiffKey`, so relation diffs are not stripped.
  - `EventRelationFieldDiffValues`: resolve related record labels by id; show only the new value in the row (same pattern as other fields: Company -> airSlate).
  - Tooltip (relations only): on hover, show before -> after with readable names (e.g. Microsoft -> Apple). Scalar and composite fields (e.g. Updated by) are unchanged and do not get this tooltip.
  - `EventFieldDiff`: route RELATION diffs to the relation renderer; all other field types keep the existing FieldDisplay behavior.

### What you’ll see
On a person (or opportunity) timeline after changing company:
```
You updated Company → airSlate
(tooltip: Microsoft → Apple)
```

## Test plan
- Change only company on a person -> timeline shows a company update with names.
- Change company and name in one save -> both appear in the diff.
- Clear company -> row shows Empty; tooltip reflects previous -> empty if applicable.
- Same we can do for the Opportunities also
- Scalar / ACTOR fields (e.g. Updated by) - no new tooltip; display unchanged.

### Before
Screenshot 2026-05-29 155123
<br><br>
Screenshot 2026-05-29 155149
<br><br>

### After
Screenshot 2026-06-01 154539
Screenshot 2026-06-01 154555
Screenshot 2026-06-01 154639
Screenshot 2026-06-01 154654
Screenshot 2026-06-01 154802

### DIFF
--- a/packages/twenty-front/src/modules/activities/timeline-activities/rows/main-object/components/EventFieldDiff.tsx
+++ b/packages/twenty-front/src/modules/activities/timeline-activities/rows/main-object/components/EventFieldDiff.tsx
@@ -2,12 +2,8 @@ import { styled } from '@linaria/react';
 
 import { EventFieldDiffLabel } from '@/activities/timeline-activities/rows/main-object/components/EventFieldDiffLabel';
 import { EventFieldDiffValue } from '@/activities/timeline-activities/rows/main-object/components/EventFieldDiffValue';
-import {
-  EventFieldDiffValueEffect
-} from '@/activities/timeline-activities/rows/main-object/components/EventFieldDiffValueEffect';
-import {
-  EventRelationFieldDiffValues
-} from '@/activities/timeline-activities/rows/main-object/components/EventRelationFieldDiffValues';
+import { EventFieldDiffValueEffect } from '@/activities/timeline-activities/rows/main-object/components/EventFieldDiffValueEffect';
+import { EventRelationFieldDiffValues } from '@/activities/timeline-activities/rows/main-object/components/EventRelationFieldDiffValues';
 import { isRelationFieldChangeValue } from '@/activities/timeline-activities/utils/relationFieldChangeValue';
 import { type FieldMetadataItem } from '@/object-metadata/types/FieldMetadataItem';
 import { type EnrichedObjectMetadataItem } from '@/object-metadata/types/EnrichedObjectMetadataItem';
--- a/packages/twenty-front/src/modules/activities/timeline-activities/rows/main-object/components/EventFieldDiff.tsx
+++ b/packages/twenty-front/src/modules/activities/timeline-activities/rows/main-object/components/EventFieldDiff.tsx
@@ -2,8 +2,12 @@ import { styled } from '@linaria/react';
 
 import { EventFieldDiffLabel } from '@/activities/timeline-activities/rows/main-object/components/EventFieldDiffLabel';
 import { EventFieldDiffValue } from '@/activities/timeline-activities/rows/main-object/components/EventFieldDiffValue';
-import { EventFieldDiffValueEffect } from '@/activities/timeline-activities/rows/main-object/components/EventFieldDiffValueEffect';
-import { EventRelationFieldDiffValues } from '@/activities/timeline-activities/rows/main-object/components/EventRelationFieldDiffValues';
+import {
+  EventFieldDiffValueEffect
+} from '@/activities/timeline-activities/rows/main-object/components/EventFieldDiffValueEffect';
+import {
+  EventRelationFieldDiffValues
+} from '@/activities/timeline-activities/rows/main-object/components/EventRelationFieldDiffValues';
 import { isRelationFieldChangeValue } from '@/activities/timeline-activities/utils/relationFieldChangeValue';
 import { type FieldMetadataItem } from '@/object-metadata/types/FieldMetadataItem';
 import { type EnrichedObjectMetadataItem } from '@/object-metadata/types/EnrichedObjectMetadataItem';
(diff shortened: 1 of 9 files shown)


martmull:
@Bonapara do we want to display the initial relation value? It looks like this will be the only field type where we show the value before -> after

If yes, i am not sure about this double arrow design: `FIELD NAME -> InitialValue -> finalValue` can you validate @Bonapara?

image

Bonapara:
Let's keep it like before and add the "before" value in a tooltip like that:

image

martmull (dismissed):

martmull:
image

martmull:
Sorry @parshipcy i think i have erased your changes

martmull (dismissed):

Bonapara:
@martmull can you check the arrow color?

https://www.figma.com/design/xt8O9mFeLl46C5InWwoMrN/Twenty?node-id=32530-139853&t=vCDd4iIPA3SEJq9I-11

martmull:
> i think some of my commits were reverted?

Yes I have pushed force (after a rebase) and vanished your commit. Can we merge like this? Your submission will still be taken into account

parshipcy:
> Yes I have pushed force (after a rebase) and vanished your commit. Can we merge like this? Your submission will still be taken into account

Yes sure, no problem

martmull:
image

martmull (approved):

PR #21057, opened 2026-05-29
Docs update: Calling a logic function from a front component

bosiraphael:
Documents how a headless front component calls a server-side logic function over HTTP via the /s/ route, so AI agents have a clear reference for implementing this pattern.


Weiko (approved):
LGTM
@FelixMalfait seems we are encouraging people to build the API url (with the /s) themselves right now. I think we should have some wrapping somewhere in case we want to implement scoped app routes 🤔

martmull (approved):

PR #21061, opened 2026-05-29
fix: SSE event stream reconnection after idle connection death

thomtrp:
The SSE event stream could silently die from network partitions, NAT table flushes, browser tab throttling, or server restarts. When this happened:
1. The `error` callback only called `captureException` — no reconnection was triggered
2. The `complete` callback was `() => {}` — a cleanly terminated stream left the client permanently broken
3. No mechanism existed to detect a silently dead connection where no FIN/RST was received

## Summary

- **Fix `error`/`complete` callbacks**: The `graphql-sse` subscription's `error` callback only reported to Sentry, and `complete` was a no-op. Both now set `shouldDestroyEventStreamState = true` to trigger the destroy-recreate lifecycle, ensuring detected transport failures and clean stream terminations lead to automatic reconnection.
- **Add server-side keepalive**: The existing heartbeat timer now runs every 30s (instead of 6min) and publishes empty events through the Redis pub/sub channel in addition to refreshing the Redis TTL (throttled to ~6min). Unlike GraphQL Yoga's opaque SSE comment pings, these are real subscription events that flow through the client's `next`/`message` handlers.
- **Add client-side keepalive monitor (`SSEKeepAliveEffect`)**: Tracks the timestamp of the last received event. If no event arrives within 90 seconds (3x the keepalive interval), it clears query listeners and triggers a stream destroy-recreate cycle.

## Test plan

- [x] Start the app, verify SSE events flow normally (workflow runs update in real-time)
- [x] Leave the app idle for >90 seconds, then trigger a workflow run — verify the stream auto-reconnects and events are delivered
- [x] Kill the server, restart it, verify the frontend recovers its event stream
- [x] Verify keepalive events (empty `objectRecordEventsWithQueryIds`/`metadataEvents`) appear in browser network tab every ~30s
- [x] Verify no regressions in SSE-dependent features (record updates, metadata changes, workflow run visualization)

### DIFF
--- a/packages/twenty-front/src/modules/sse-db-event/components/SSEKeepAliveEffect.tsx
+++ b/packages/twenty-front/src/modules/sse-db-event/components/SSEKeepAliveEffect.tsx
@@ -1,0 +1,43 @@
+import { SSE_LIVENESS_CHECK_INTERVAL_IN_MS } from '@/sse-db-event/constants/SseLivenessCheckIntervalInMs';
+import { SSE_LIVENESS_TIMEOUT_IN_MS } from '@/sse-db-event/constants/SseLivenessTimeoutInMs';
+import { activeQueryListenersState } from '@/sse-db-event/states/activeQueryListenersState';
+import { lastSseEventReceivedTimestampState } from '@/sse-db-event/states/lastSseEventReceivedTimestampState';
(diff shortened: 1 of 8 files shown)


martmull (approved):
lgtm nice

PR #21068, opened 2026-05-29
fix(twenty-front): enable text selection for display-mode fields

### LINKED ISSUE #21056: Read-only/display field values are not reliably selectable or copyable
When a record field is rendered in display mode, the visible scalar value can be difficult or impossible to select with the cursor. This is especially noticeable for email-like values, phone numbers, dates, IDs, and other short text values shown in record field widgets.

This makes it hard to copy field values without entering edit mode. For users who only have read access, entering edit mode may not be available at all.

**Expected behavior**
Users should be able to copy visible scalar field values without needing edit mode.

Possible UX directions:

1. Make display-mode field values selectable by default.
2. Scope selectable text to record detail / read-only field widgets.
3. Add a copy affordance for scalar field values.

I initially explored a copy button, but that may add UI clutter and can depend on clipboard API behavior. Native text selection may be the simpler and more expected interaction.

**Actual behavior**
Some display-mode field values inherit `user-select: none` through inline-cell wrappers or nested display components, so dragging over the visible value does not reliably select text.

**Reproduction**
1. Open a record detail side panel.
2. Find a scalar field value such as an email address, phone number, date, or system field.
3. Try to select the visible value with the cursor.
4. In some cases, the text cannot be selected, and there is no alternate copy affordance.

**Why this matters**
Copying field values is a common read-only workflow. It should not require edit permission, and it should not require switching the field into edit mode.

**Related context**
#4599 covered adding a copy button to floating/edit inputs. This issue is related, but specifically concerns display-mode/read-only field values and native text selection without entering edit mode.

dev-kp-eloper:
## Description
This PR resolves a usability issue where scalar field values (emails, phone numbers, dates, IDs, text, etc.) rendered in display-mode or read-only mode in the record detail side panel could not be highlighted, selected, or copied natively with the cursor.

## Root Cause
Both `RecordInlineCellContainer` and `RecordInlineCellHoveredPortalContent` wrapper elements had `user-select: none;` hardcoded in their styled-component definitions. This styling propagated down to all nested display widgets, locking their content and preventing native text selection.

## Changes
- Updated `StyledInlineCellBaseContainer` in `RecordInlineCellContainer.tsx` to use `user-select: text;` instead of `none;`.
- Updated `StyledInlineCellBaseContainer` in `RecordInlineCellHoveredPortalContent.tsx` to use `user-select: text;` instead of `none;`.

These changes restore natural browser text selection capabilities for record detail widgets without altering interactive edit-mode behaviors.

## Verification
- Verified styling changes.
- Tested locally to ensure that text highlighting and copy-pasting function correctly when dragging over read-only fields.

Closes #21056

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-inline-cell/components/RecordInlineCellContainer.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-inline-cell/components/RecordInlineCellContainer.tsx
@@ -61,7 +61,7 @@ const StyledInlineCellBaseContainer = styled.div<{ readonly: boolean }>`
   display: flex;
   gap: ${themeCssVariables.spacing[1]};
   height: fit-content;
-  user-select: none;
+  user-select: text;
   width: 100%;
 `;
 
--- a/packages/twenty-front/src/modules/object-record/record-inline-cell/components/RecordInlineCellContainer.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-inline-cell/components/RecordInlineCellContainer.tsx
@@ -45,6 +45,7 @@ const StyledValueContainer = styled.div<{ readonly: boolean }>`
   display: flex;
   min-width: 0;
   position: relative;
+  user-select: text;
   width: 100%;
 `;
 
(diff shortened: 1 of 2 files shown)


charlesBochet (approved):
LGTM, made a minor fix, thank you!

PR #21075, opened 2026-05-30
fix(workflow): hide empty option for non-nullable select fields

### LINKED ISSUE #20139: Bug: For Non nullable fields (SELECT, MULTISELECT), do not allow workflow update / create / upsert to set "no selected"

brendanerofeev:
## Summary

Closes #20139.

In a workflow **Create / Update / Upsert Record** action, the field input for a `SELECT` field always offered a **"No value"** option. Selecting it clears the field — even when that field is **not nullable** (for example an opportunity `Stage`). The maintainer asked in the issue to *"not allow to put a 'No stage' if the field is not nullable"*.

### Root cause

`FormSelectFieldInput` unconditionally builds a `defaultEmptyOption` and passes it as the `<Select>`'s `emptyOption`, regardless of the field's nullability:

```tsx
const defaultEmptyOption = {
  label: label ? t`No ${label}` : t`No value`,
  value: '',
  icon: IconCircleOff,
};
// ...
<Select ... emptyOption={defaultEmptyOption} />
```

The nullability is already available on the field metadata (`FieldSelectMetadata.isNullable`), and the regular (non-workflow) record input already gates its clear affordance on it in `SelectFieldInput.tsx` (`fieldDefinition.metadata.isNullable && canSelectEmpty ? handleClearField : undefined`). The workflow form input just wasn't using it.

### Fix

- Add an optional `isNullable` prop to `FormSelectFieldInput` and only build the empty option when the field is nullable (`isNullable !== false`). `emptyOption` is already optional on `<Select>`, so omitting it cleanly removes the choice.
- Pass `isNullable={field.metadata?.isNullable}` from `FormFieldInput`'s `isFieldSelect` branch.

The prop is optional and defaults to the previous behaviour, so the other `FormSelectFieldInput` call sites (country, calling code, currency, cron, etc.) are unaffected.

Scope is intentionally limited to single `SELECT`: `FieldMultiSelectMetadata` has no `isNullable` field and the multi-select input has no equivalent empty option, so it is left untouched.

## Test plan

- [x] `npx oxlint --type-aware` — passes on the changed files
- [x] `npx oxfmt --check` — passes
- [x] `tsc --noEmit` — no type errors in the changed files
- [x] Added a `NonNullable` Storybook interaction test asserting the "No value" option is not rendered when `isNullable={false}`
- [ ] Manual: in a workflow Update Record step on Opportunity, the **Stage** field no longer shows a "No value" option; a nullable select still does


brendanerofeev:
This PR is currently open, not draft, and mergeable, but it looks blocked by the external triage automation rather than by a visible code review item.

The `Automated Review` status is failing with:

> Does not follow Twenty technical standards (1 issue)

However, the linked `twentyhq/ci-privileged` workflow logs are private, so I cannot see the specific standards finding to address it. Could someone please share the concrete issue from that run, or clear/re-run the external triage if it is stale?

For context, the visible diff is limited to passing select-field nullability into `FormSelectFieldInput`, hiding the empty option for non-nullable selects, and adding a Storybook regression story.

charlesBochet (approved):
Thank you @brendanerofeev , made a few fixes

PR #21080, opened 2026-05-30
fix: filtered view resetting to unfiltered list on navigation

### LINKED ISSUE #21079: [bug]: Saved view lost when switching sections
## Bug Description
Actually, i was looking into #20840 but then I found a separate issue that made it hard to tell whether filters were actually persisted.

### Steps
1. Open Opportunities (or any object).
2. Add a filter (e.g. Stage) and use Save as new view.
3. Confirm the URL updates with the new viewId and the list is filtered.
4. Navigate to another section (e.g. People).
5. Return to the Opportunities section.
6. You will notice that the URL reverts to default and list showing the unfiltered data.

https://github.com/user-attachments/assets/5d1b0d78-750a-430a-b08b-c214d52f63d9

## Expected behavior
Return to the same saved view with filters still applied (same viewId in the URL).

## Actual
The URL reverts to the default “All …” (index) view, and the list shows the unfiltered data. The filter looked “lost” even after Save as new view

parshipcy:
## Summary
- Resolves #21079 
- Object navigation links no longer force the default index view, which had no saved filters.
- Returning to an object after “Save as new view” now opens the last visited (filtered) view instead of the unfiltered default list.

## Test plan
1. Add a filter -> save as new view -> list is filtered
2. Navigate away and back -> filters still applied (not reset to unfiltered)

### Before
https://github.com/user-attachments/assets/25326339-a3a1-4171-89cc-5149e254982e

### After
https://github.com/user-attachments/assets/b4600043-fc7a-4670-9c68-23daa6c31ec8


parshipcy:
@Bonapara @martmull can you please take a look when you have time?

Bonapara:
From a product perspective it would be ok for me for object navigation items to display the last seen view

parshipcy:
> From a product perspective it would be ok for me for object navigation items to display the last seen view

Yes that's the issue - object navigation was always forcing the default index view, so if a user created/selected a filtered saved view and navigated away, coming back reset them to the unfiltered list.

charlesBochet (approved):
Thank you!

PR #21082, opened 2026-05-31
[Website] i18n module, page-local sections, translatable copy

mabdullahabaid:
**i18n** — collapsed the ~22 scattered i18n files into a single module and turned on Spanish alongside French.

**Sections** — dropped the old compound pattern (`Section.Root`, `Section.Heading`, …). Reusable layout shells moved to `src/templates/`, atomic bits stay in `design-system/`, and each page now owns its copy in local `_components` blocks instead of pulling it out of shared sections. Data files hold arrays only, no prose.

**Copy → `<Trans>`** — A lot of headings were split across several `<HeadingPart>`s just for font styling, which meant each piece was a separate translation string. A translator got "Build your Enterprise CRM" and "at AI Speed" as two unrelated strings and had no way to reorder them for their language. Those are now single `<Trans>` units with placeholders. Same idea for the old `\n` + `white-space: pre-line` line-break trick: replaced with a small `ResponsiveLineBreak` element so the break is doesn't quietly rot, and did a dead-code pass.

The de-fragmentation changes the message IDs, so around 60 strings will fall back to English in fr/es until Crowdin re-syncs.


packages/twenty-website/src/sections/Footer/components/Footer.tsx
```diff
@@ -0,0 +1,21 @@
+import { FOOTER_DATA } from '@/sections/Footer/data';
+
+import { Bottom } from './Bottom';
+import { Logo } from './Logo';
+import { Nav } from './Nav';
+import { Root } from './Root';
+
+// The footer is the same composition on every page, so it owns its data and
+// child composition; the root layout just renders <Footer />.
```
FelixMalfait: Remove comments like this one, they are falsely useful

packages/twenty-website/src/app/[locale]/(home)/_components/HomeStepperSection.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export function HomeStepperSection() {
+  const steps: HomeStepperStepType[] = [
+    {
+      heading: (
+        <Heading size="lg" weight="light">
+          <Trans>
+            <HeadingPart fontFamily="serif">
+              Begin with production-grade
+            </HeadingPart>{' '}
```
Comment (FelixMalfait) [inline packages/twenty-website/src/app/[locale]/(home)/_components/HomeStepperSection.tsx]: Let's get rid of all those {' '} and find something more elegant, they shouldn't be necessary and will pollute translations (I'm not a fan of including HTML within trans which will already make the translations look ugly but OK for now, but let's at least keep this html as minimal as possible :))

FelixMalfait (approved):

packages/twenty-website/src/app/[locale]/(home)/_components/HomeStepperSection.tsx
Comment (mabdullahabaid) [reply packages/twenty-website/src/app/[locale]/(home)/_components/HomeStepperSection.tsx]: Removed {' '} and used CSS ::after to solve!

packages/twenty-website/src/app/[locale]/product/_components/ProductThreeCards.tsx
```diff
@@ -0,0 +1,27 @@
...
+    <ThreeCardsSection scheme="light">
+      <ThreeCardsIntro align="left" maxWidthMd={600}>
+        <Eyebrow>
+          <HeadingPart fontFamily="sans">
+            <Trans>Stop settling for trade-offs.</Trans>
+          </HeadingPart>
+        </Eyebrow>
+        <Heading size="lg" weight="light">
+          <Trans>
+            <HeadingPart fontFamily="serif">A modern CRM with</HeadingPart>
```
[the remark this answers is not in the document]
Comment (mabdullahabaid) [reply packages/twenty-website/src/app/[locale]/product/_components/ProductThreeCards.tsx:19]: Added space via CSS ::after already.

packages/twenty-website/src/app/[locale]/why-twenty/_components/WhyTwentyMeaningEditorial.tsx
```diff
@@ -0,0 +1,42 @@
...
+        crossX: '120px',
+        crossY: '0px',
+        lineColor: theme.colors.secondary.border[10],
+      }}
+      introAlign="right"
+      eyebrowColorScheme="secondary"
+      eyebrow={i18n._(msg`What this means`)}
+      heading={
+        <Trans>
+          <HeadingPart fontFamily="serif">Differentiation now</HeadingPart>
```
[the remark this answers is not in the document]
Comment (mabdullahabaid) [reply packages/twenty-website/src/app/[locale]/why-twenty/_components/WhyTwentyMeaningEditorial.tsx:25]: Added space via CSS ::after already.

PR #21084, opened 2026-05-31
fix: reorder table columns against visible fields only (#20940)

### LINKED ISSUE #20940: Bug: "Move Left" & "Move Right" buttons to rearrange columns in table view are unreliable.
## Bug Description

Reordering columns in the default table view sometimes works, sometimes doesn't

https://github.com/user-attachments/assets/8bd052c8-48d8-4664-99f9-0d44f1f1e84e

## Technical inputs

Clicking "Move Right/Left" seems to cycle it's position through several invisible columns (maybe the position is being calculated against all fields, not just the visible ones.)

singhharsh1708:
### Problem
The "Move Left" / "Move Right" actions in the table column header menu were unreliable. Clicking them often produced no visible change, or appeared to move the column an inconsistent number of positions.

### Root cause
`useMoveRecordField` computed the swap target from **all** record fields (`currentRecordFieldsComponentState`) sorted by position — including hidden and non-readable columns. As a result, a move frequently swapped positions with an *invisible* neighbor, leaving the visible column order unchanged.

This was also inconsistent with the drag-and-drop reorder path (`useReorderVisibleRecordFields`), which already operates only on the visible field set, and with the dropdown's own Move enable/disable logic, which is based on `visibleRecordFields`.

### Fix
`useMoveRecordField` now sources the neighbor from `visibleRecordFieldsComponentSelector` — the same selector that drives the table display and the Move menu items (`isVisible && isReadable && isActive`, sorted by position). The real `position` values are still swapped, so hidden columns keep their positions and only the visible order changes.

### Tests
Added `useMoveRecordField.test.tsx`, which seeds real object metadata with a hidden column interleaved between visible ones (by position) and asserts that the visible selector reorders correctly after a move. The test fails against the previous implementation and passes with this change.

### How to verify
1. Open any table view.
2. Open a column header menu and click "Move Right" / "Move Left".
3. The column now moves reliably by one visible position each click, regardless of hidden columns.


charlesBochet (dismissed):
Thank you @singhharsh1708 :)

charlesBochet (approved):
Thank you @singhharsh1708 :)

PR #21001, opened 2026-05-28
`EncryptedString` `PlaintextString` branded string types

prastoin:
## Summary

closes https://github.com/twentyhq/core-team-issues/issues/2464

Introduces compile-time branded types to distinguish encrypted ciphertext from plaintext strings, preventing mix-ups like the one fixed in #20819 — but at the type level rather in addition to the one existing at runtime.

### Branded string primitives

- Created `EncryptedString` and `PlaintextString` as hard nominal brands using `z.string().brand(...)`, making them non-assignable to each other or to raw `string`
- Created `isEncryptedString` type predicate to narrow `string` to `EncryptedString` based on the `enc:v2:` envelope prefix
- Retyped `SecretEncryptionService`: `encryptVersioned` accepts `PlaintextString`, `decryptVersioned` returns `PlaintextString`

### Entity typing

- Typed encrypted columns across entities: `SigningKeyEntity.privateKey`, `TwoFactorAuthenticationMethodEntity.secret`, `ApplicationRegistrationVariableEntity.encryptedValue`, `ApplicationVariableEntity.value`
- Parameterized JSONB types for connected account connection parameters (`ImapSmtpCaldavParams<Pwd>`) with reusable aliases `EncryptedImapSmtpCaldavParams` / `DecryptedImapSmtpCaldavParams`
- Typed DTOs (`CreateApplicationRegistrationVariableInput`, `UpdateApplicationRegistrationVariablePayload`, `UpdateApplicationVariableEntityInput`) with `PlaintextString`

### ApplicationVariable always-encrypt uniformization

- Retyped `ApplicationVariableEntity.value` to `EncryptedString | ''` — all values are now encrypted regardless of `isSecret`
- Updated `ApplicationVariableEntityService` to always encrypt on write and always decrypt on read
- Simplified `UpdateApplicationVariableActionHandlerService` by removing conditional encrypt/decrypt-on-isSecret-toggle logic
- Added slow instance command (`2.9.0`) to backfill-encrypt existing `isSecret=false` plaintext rows and tighten the `CHECK` constraint

### ConfigStorageService refactor

- Split `convertAndSecureValue` (which used `any`) into two well-typed methods: `convertAndDecrypt` and `convertAndEncrypt`
- Introduced `isSensitiveStringValue` type predicate to narrow values before encryption/decryption

### What's next
- Typeorm entity derivation to strictly type sitemap configuration as code + handler logic for encryption rotation
- https://github.com/twentyhq/core-team-issues/issues/2465

PR #21085, opened 2026-05-31
Strictly type encryption rotation key site maps constants through entity type derivation

prastoin:
# Introduction
Followup https://github.com/twentyhq/twenty/pull/21001

Now that the typeorm entities provide grains over their `encryptedString` value, we can strictly type the sitemaps of the encrypted string to rotate in case of encryption key rotation and also the integration tests tests cases


packages/twenty-server/src/engine/core-modules/secret-encryption/branded-strings/contains-encrypted-string.type.ts
```diff
@@ -0,0 +1,13 @@
+import { type EncryptedString } from 'src/engine/core-modules/secret-encryption/branded-strings/encrypted-string.type';
+
+export type ContainsEncryptedString<T> = T extends EncryptedString
```
Weiko, line 3: Can we make this nested type a bit more readable? Maybe split it a bit?
```ts
type AnyPropertyContainsEncryptedString<T> = true extends {
  [K in keyof T]: ContainsEncryptedString<NonNullable<T[K]>>;
}[keyof T]
  ? true
  : false;

export type ContainsEncryptedString<T> = T extends EncryptedString
  ? true
  : T extends ReadonlyArray<infer U>
    ? ContainsEncryptedString<U>
    : T extends object
      ? AnyPropertyContainsEncryptedString<T>
      : false;
```

Weiko (approved):

PR #21098, opened 2026-06-01
fix(twenty-sdk): implement and export real helper family at runtime in front-component

### LINKED ISSUE #21094: isDefined imported from twenty-sdk/front-component is null at runtime (crashes custom front-components)
## Bug Description

A custom React FRONT_COMPONENT (built with the Twenty SDK) that imports `isDefined` from `twenty-sdk/front-component` crashes at runtime in the browser. As soon as `isDefined(...)` is called during render, the component throws. This is what showed on the record page:

`FrontComponent error: Uncaught TypeError: Xs is not a function`

(`Xs` is the minified name of `isDefined` in the built bundle.)

`useRecordId` imported from the same entry point works fine, so the import path itself resolves. Replacing `isDefined(x)` with a plain `x != null` removes the crash entirely.

Observed on a self-hosted v2.8.3 instance.

## Expected behavior

`isDefined`, since it is exported from the public SDK front-component entry point, should be a callable function at runtime in a front-component (the same way `useRecordId` is) — or it should not be exported from that entry point at all if it is not meant for runtime use.

## Technical inputs

Reporting as a new contributor (not a developer); I found this with Claude Code and traced the source against a clone that may be behind latest `main`, so paths/lines may differ — please verify.

- `isDefined` appears to be a build-time TYPE STUB: defined as `export const isDefined = null as unknown as (value: unknown) => boolean;` in `front-component-api/conditional-availability/conditional-availability-variables.ts` (intended for static conditional-availability expressions).
- It is re-exported through the public SDK / front-component entry, so importing it into a front-component yields `null`. Calling `null` produces "is not a function".
- The front-component esbuild build bundles it as `null` (externals list is empty); the real `isDefined` in `twenty-shared/utils` isn't reachable from a built front-component. `useRecordId` works because it is a real implementation.
- Repro: `defineFrontComponent` -> `import { isDefined } from 'twenty-sdk/front-component'` -> call `isDefined(x)` in render -> sync/install -> open the record -> TypeError.
- Possible fixes: stop re-exporting the stub from the front-component entry, OR ship a runtime-safe util export, OR mark stub-only exports clearly in types/JSDoc.

Dakshx07:
### Related Issues
Closes #21094 

### Problem
Custom React `FRONT_COMPONENT` widgets built using the Twenty SDK crash in the browser at runtime when they attempt to import and use utility helpers like `isDefined` or `isNonEmptyString` from `twenty-sdk/front-component` (e.g. producing `TypeError: isDefined is not a function` / `FrontComponent error: Uncaught TypeError: Xs is not a function`).

### Root Cause
In `packages/twenty-sdk/src/sdk/front-component/conditional-availability/conditional-availability-variables.ts`, the callable-helper family (`isDefined`, `isNonEmptyString`, `includes`, `every`, `some`, `none`, etc.) was exported as dummy null stubs intended only for static conditional availability compilation analysis:
```typescript
export const isDefined = null as unknown as (value: unknown) => boolean;
export const isNonEmptyString = null as unknown as (value: unknown) => boolean;

### DIFF
--- a/packages/twenty-sdk/src/sdk/front-component/conditional-availability/conditional-availability-variables.ts
+++ b/packages/twenty-sdk/src/sdk/front-component/conditional-availability/conditional-availability-variables.ts
@@ -30,63 +30,83 @@ export const canImpersonate =
 export const canAccessFullAdminPanel =
   null as unknown as CommandMenuContextApi['canAccessFullAdminPanel'];
 
+import { isDefined as realIsDefined, safeGetNestedProperty } from 'twenty-shared/utils';
+import { isNonEmptyString as realIsNonEmptyString, isNonEmptyArray } from '@sniptt/guards';
+
 export const objectMetadataItem =
   null as unknown as CommandMenuContextApi['objectMetadataItem'];
 
 export const objectMetadataLabel =
   null as unknown as CommandMenuContextApi['objectMetadataLabel'];
 
-export const isDefined = null as unknown as (value: unknown) => boolean;
-export const isNonEmptyString = null as unknown as (value: unknown) => boolean;
-export const includes = null as unknown as (
-  array: unknown,
-  value: unknown,
-) => boolean;
-export const every = null as unknown as (
-  array: unknown,
-  prop: string,
-) => boolean;
-export const everyDefined = null as unknown as (
-  array: unknown,
-  prop: string,
-) => boolean;
-export const everyEquals = null as unknown as (
-  array: unknown,
-  prop: string,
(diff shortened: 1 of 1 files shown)


bosiraphael (commented):
Hello @Dakshx07. Thank you for your contribution, but I have to close this PR since it relies on a misunderstanding of what these functions are supposed to be used for. We will think about a way to make this clearer, maybe with a linter rule.
See https://github.com/twentyhq/twenty/issues/21094#issuecomment-4591042207 for context

PR #21106, opened 2026-06-01
fix(ai): correct RICH_TEXT and MORPH_RELATION record filter operators

richroberts-prog:
## Problem

The AI find-records tool generates filter schemas via `generateFieldFilterZodSchema`. `RICH_TEXT` currently shares the `TEXT` case, so the agent is told it can use scalar text operators (`like`/`ilike`/`startsWith`/`endsWith`/`eq`/…) directly on a rich-text field.

But `RICH_TEXT` is a **composite** type (`markdown` + `blocknote` sub-fields, see `rich-text.composite-type.ts`). Applying a scalar operator to the composite root throws at query time:

```
ERROR [FindRecordsService] Failed to find records: Object person doesn't have any "ilike" field.
ERROR [FindRecordsService] Failed to find records: Sub field "ilike" not found for composite type: RICH_TEXT
```

`FindRecordsService` catches and returns `success: false`, so the agent retries mid-turn — burning latency/tokens — and can **never** search rich-text body content (note bodies, `about`, etc.).

## Fix

Give `RICH_TEXT` its own case in the filter-schema generator that exposes the `markdown` and `blocknote` sub-fields, each carrying the text operators — mirroring the existing composite patterns for `EMAILS` (`primaryEmail`), `PHONES` (`primaryPhoneNumber`), `LINKS` (`primaryLinkUrl`), `FULL_NAME`, and `ADDRESS`.

So the agent now emits:

```jsonc
{ "noteBody": { "markdown": { "ilike": "%onboarding%" } } }   // valid composite sub-field filter
```

instead of:

```jsonc
{ "noteBody": { "ilike": "%onboarding%" } }                   // throws on composite root
```

This both **stops the throw** and **makes rich-text content actually searchable** (the original intent). `TEXT` keeps its existing root-level scalar operators unchanged.

## Test

Added `__tests__/field-filters.zod-schema.spec.ts`:
- `RICH_TEXT` routes pattern operators onto `markdown` / `blocknote`
- root-level scalar operators on `RICH_TEXT` are no longer accepted
- `TEXT` root-level operators unchanged

## Notes

- No DB/schema migration; render/tool-schema layer only.
- Reproduced against `twentycrm/twenty:latest`; the faulting code is unchanged on `main` as of this PR.


richroberts-prog:
Expanded scope: also fixes `MORPH_RELATION` fields (e.g. `noteTarget.targetPerson`), which had no case in `generateFieldFilterZodSchema` and fell through to the text default — advertising `like`/`ilike` to the agent. The query runner resolves them as relations and throws `Object <target> doesn't have any "ilike" field`, and the agent couldn't filter by the relation at all. Now keyed as `${name}Id` with the relation id filter (eq/neq/in/is), like MANY_TO_ONE relations. Surfaced via the same "Tom Foreman's notes" repro (fetching a person's notes goes through `noteTarget.targetPersonId`).

packages/twenty-server/src/engine/core-modules/record-crud/zod-schemas/field-filters.zod-schema.ts
```diff
@@ -53,6 +52,45 @@ export const generateFieldFilterZodSchema = (
...
+        .optional();
+
+      return z
+        .object({
+          markdown: richTextSubFieldFilter.describe(
+            'Filter by the markdown text content (use this to search rich-text body text)',
+          ),
+          blocknote: richTextSubFieldFilter.describe(
+            'Filter by the raw BlockNote JSON representation',
+          ),
```
etiennejouan: nit: blocknote subfield is useless, can be removed

packages/twenty-server/src/engine/core-modules/record-crud/zod-schemas/field-filters.zod-schema.ts
```diff
@@ -498,6 +536,34 @@ export const generateFieldFilterZodSchema = (
...
+            .array(z.string().uuid())
+            .optional()
+            .describe('Related record ID in array of values'),
+          is: NullCheckEnum.optional().describe(
+            'Check for missing or empty related record. Use "NULL" to find records with no relation, "NOT_NULL" for records with a relation',
+          ),
+        })
+        .optional()
+        .describe(`Filter by ${fieldName} (morph relation field)`);
+    }
```
etiennejouan: It should be treated identically to the RELATION MANY_TO_ONE case

packages/twenty-server/src/engine/core-modules/record-crud/zod-schemas/record-filter.zod-schema.ts
```diff
@@ -36,11 +36,15 @@ export const generateRecordFilterSchema = (
...
+    // MANY_TO_ONE relations and morph relations (each morph target also has an
+    // `${name}Id` join column).
+    const usesRelationIdKey =
+      (isFieldMetadataEntityOfType(field, FieldMetadataType.RELATION) &&
+        field.settings?.relationType === RelationType.MANY_TO_ONE) ||
+      isFieldMetadataEntityOfType(field, FieldMetadataType.MORPH_RELATION);
 
-    filterShape[isManyToOneRelationField ? `${field.name}Id` : field.name] =
+    filterShape[usesRelationIdKey ? `${field.name}Id` : field.name] =
       fieldFilter;
```
etiennejouan:
```
const isManyToOneRelationField =
      (isFieldMetadataEntityOfType(field, FieldMetadataType.RELATION) || isFieldMetadataEntityOfType(field, FieldMetadataType.MORPH_RELATION)) &&
      field.settings?.relationType === RelationType.MANY_TO_ONE;

    filterShape[isManyToOneRelationField ? `${field.name}Id` : field.name] =
      fieldFilter;
```

etiennejouan (commented):
Hi @richroberts-prog, thank you for your contribution, good catch ! Left some comments.

francozeta:
I took a quick pass at the review comments here and prepared a small patch on top of this branch:

https://github.com/francozeta/twenty/tree/fix/help-ai-richtext-filters

It removes the `blocknote` rich-text subfield, makes `MORPH_RELATION` share the same handling path as many-to-one relations, and updates the tests accordingly. Happy to adjust or drop it if you already have this covered.

etiennejouan (approved):

PR #21072, opened 2026-05-30
feat(settings): discovery hero rollout + ephemeral playground token

FelixMalfait:
## Summary

Two intertwined streams of work:

### UI — discovery hero pattern, settings shell, AI/API redesign
- **Generalize `SettingsDiscoveryHeroCard`** and use it on Layout, Data Model, Apps, AI, API/Webhooks, Members. Drops 4 per-page wrapper files (`SettingsObjectCoverImage`, `SettingsLayoutCoverImage`, `SettingsLayoutCustomizeVideoModal`, `SettingsDataModelVisualizeVideoModal`). Each page now supplies cover src, modal id, and tab list.
- **Modal**: swap `<video>` placeholder for the Vimeo iframe pattern from `twenty-docs`, per-tab `vimeoId`. Drop the parallel border-bottom on the header (TabList draws its own baseline) and the grey background behind the video. Note: Vimeo's embed allowlist applies — the iframes load with the correct URL on `localhost` but the player itself requires the video owner to allow the dev/staging domains in Vimeo settings.
- **AI page** rebuilt into a Cockpit pattern (Overview / Models / Skills / Tools / Usage). New `SettingsAiOverviewTab` with default Smart/Fast pickers, at-a-glance stats, and an MCP signpost that deep-links to `/settings/api-webhooks#mcp`. System Prompt link moved under Models. Advanced tab removed.
- **API & Webhooks** now has 4 tabs (Playground / MCP / API Keys / Webhooks). Hero card above tabs. Playground tab inverted to "Core API" / "Metadata API" sections, each containing REST + GraphQL cards — schema is the meaningful axis, protocol is secondary. Hash deep-link sync delegated to the shared `TabListFromUrlOptionalEffect`.
- **Settings shell**: unified drawer outer padding (kill `isSettings` branch), extract `CollapsibleNavigationDrawerSection`, add `iconColor` on settings nav items, fix Exit Settings button alignment, 880px content cap.

### Backend — strategy C: ephemeral playground token
The legacy paste-your-API-key flow is replaced by an on-demand short-lived token scoped to the calling user's permissions. No shared "Playground" API key to manage or revoke.

- New `JwtTokenTypeEnum.PLAYGROUND`. `PlaygroundTokenJwtPayload = Omit<AccessTokenJwtPayload, 'type' | impersonation fields>` so any future ACCESS claim flows through automatically.
[...]

PR #21107, opened 2026-06-01
fix: return proper FORBIDDEN GraphQL errors from ApiKeyResolver

Weiko:
## Context
CI is broken on main, regression introduced in https://github.com/twentyhq/twenty/pull/21072

Guard-rejected ApiKey mutations returned malformed GraphQL responses. RequireAccessTokenGuard (and SettingsPermissionGuard) throw plain AuthException/PermissionException classes, which are not GraphQLErrors. ApiKeyResolver had no @UseFilters, so these exceptions were never translated, they surfaced as request-level errors with no data key (data: undefined) and a non-FORBIDDEN code, instead of data: null + FORBIDDEN.

This broke the `createApiKey › should reject a non-ACCESS token even with API key permission` integration test (expect(res.body.data).toBeNull() received undefined). The sibling generateApiKeyToken test passed only because it lives on AuthResolver, which already declares these filters.

## Fix
Add the standard exception filters to ApiKeyResolver, matching the idiom used by other guard-protected resolvers
```ts
@UseFilters(AuthGraphqlApiExceptionFilter, PermissionsGraphqlApiExceptionFilter)
```

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/api-key/api-key.resolver.ts
+++ b/packages/twenty-server/src/engine/core-modules/api-key/api-key.resolver.ts
@@ -1,4 +1,4 @@
-import { UseGuards } from '@nestjs/common';
+import { UseFilters, UseGuards } from '@nestjs/common';
 import { Args, Mutation, Parent, Query, ResolveField } from '@nestjs/graphql';
 
 import { type QueryDeepPartialEntity } from 'typeorm/query-builder/QueryPartialEntity';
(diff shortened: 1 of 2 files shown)


martmull (approved):

prastoin (approved):

prastoin:
force merging as we have an egg and chicken issue with:
- playwright infinite installation loop
- broken storybook test #21072
- broken integration test #21072

PR #21109, opened 2026-06-01
fix(ai): correct find-records tool description (top-level filter fields)

richroberts-prog:
## Problem

The AI `find_<object>` tool builds its input schema with `generateFindToolInputSchema`, which **spreads field filters at the args root** (alongside `limit`/`offset`/`orderBy`/`and`/`or`/`not`). `tool-executor.service.ts` then maps the raw model args to a filter with:

```ts
const { limit, offset, orderBy, ...filter } = args;
```

The zod schema is only used to generate the JSON schema *shown* to the model (`z.toJSONSchema(...)`) — it is **never used to validate the args coming back**. So when the model emits a bare operator where a field name belongs, e.g. `{ ilike: "Foreman" }`, it passes straight to the query runner, which throws and burns a retry mid-turn:

```
ERROR [FindRecordsService] Failed to find records: Object person doesn't have any "ilike" field.
ERROR [FindRecordsService] Failed to find records: Object person doesn't have any "eq" field.
```

Two contributing faults:
1. **The tool description actively misleads the model** — it says ``use filter: { id: { eq: "record-id" } }``, a `filter` wrapper the root-spread schema doesn't have, inviting the malformed shape.
2. **No server-side validation** — invalid root keys reach the query runner instead of being rejected against the advertised contract.

## Fix

1. **`FindRecordsService` prunes invalid filter keys before querying.** Using the same filter shape the tool schema advertises (`generateRecordFilterSchema(...).filterShape`), it drops any key that is neither a real field nor a logical operator (`and`/`or`/`not`), recursing through `and`/`or`/`not`. A model that sends `{ ilike: "Foreman" }` now gets a valid (empty) filter rather than an exception. Extracted as a pure, unit-tested util `pruneFilterToAllowedKeys`.
2. **Corrected the `find_<object>` tool description** to describe the real top-level-field shape and explicitly warn against a `filter` wrapper and bare root operators.

## Test

`__tests__/prune-filter-to-allowed-keys.util.spec.ts` covers: valid filters untouched, bare root operators dropped, valid siblings preserved, `and`/`or`/`not` recursion, and non-object input.

## Notes

- Defensive for all `FindRecordsService` callers; `find_one` (`{ id: { eq } }`) and workflow find-records pass valid filters and are unaffected.
- Companion to #21106 (RICH_TEXT composite filters). Both surfaced from the same `"Tom Foreman's notes"` AI-chat repro; this PR addresses the root-level-operator half.


richroberts-prog:
**Scope reduced.** The earlier version validated the find filter against the tool schema inside the shared `FindRecordsService` — but that service is also used by the find-records **workflow action**, which supports one-hop relation-traversal filters the AI tool schema doesn't model. That stripped legitimate workflow filters and broke `find-records-relation-traversal-workflow.integration-spec`. 

This PR is now just the **find-tool description correction** (the part that actually addressed the root-level-operator confusion): the description told the model to use a `filter: { … }` wrapper that doesn't exist in the root-spread schema. The RICH_TEXT / MORPH_RELATION schema-correctness fixes live in #21106.

charlesBochet (approved):

PR #21113, opened 2026-06-01
Bump playwright

prastoin:
Playwright installation is infinite looping in the ci
seems like to be a global outage


etiennejouan (approved):

prastoin:
force merging as we have an egg and chicken issue with:
- playwright infinite installation loop
- broken storybook test #21072
- broken integration test #21072

PR #21115, opened 2026-06-01
feat: seed private onboarding kanban with nested account-owner filter

FelixMalfait:
## What & why

New users land in a CRM that feels empty and miss powerful features (kanban views, advanced/nested filters, private views). This seeds one example view that puts several of those concepts in front of the workspace creator on day one.

Every new workspace's creator gets a **private "My Pipeline" kanban**: Opportunities grouped by stage, filtered by the nested advanced rule **`company.accountOwner = me`**. They immediately see a kanban, a nested relation filter, the "Me" variable, and a private (per-user) view.

## Changes

**Backend — seed the view (`twenty-server`)**
- New `SeedOnboardingPrivateViewService` creates, for the first user at workspace activation:
  - a `KANBAN` view on Opportunity grouped by `stage`, `visibility: UNLISTED` + `createdByUserWorkspaceId` (private to the creator), SUM(amount) aggregate, mirroring the standard "By Stage" card fields;
  - a nested view filter `opportunity.company → company.accountOwner = me`.
- Wired into `activateWorkspace` after prefill, **best-effort** (try/catch) so a seed failure never blocks activation.
- `prefill-companies` now sets `accountOwner` to the workspace creator on the example companies, so the board isn't empty on day one (mirrors how example opportunities already set `owner`).

**Frontend — make that filter user-authorable (`twenty-front`)**
- A many-to-one relation can now be selected as an advanced-filter **leaf** (previously excluded). It filters by the relation's foreign key — `company.accountOwnerId = X`, a single hop the backend already resolves — not a multi-hop traversal.
- The record-value picker resolves its options from the **leaf** relation's target object (e.g. WorkspaceMember + the "Me" pin), not the source relation's.

## Design notes
- **"Private view" = `UNLISTED` + `createdByUserWorkspaceId`.** `isViewVisibleToUser` returns true only for the creator, so no new visibility model is needed.
- **FK leaf, not a second hop.** `company.accountOwner = me` compiles to `{ company: { accountOwnerId: { in: [me] } } }` — a scalar compare on the already-joined table. The genuinely multi-hop form (`company.accountOwner.<field>`) stays rejected by the one-hop depth cap.
- **Resilient.** ViewFilter cascades on field deletion; deleting `accountOwner` drops the filter and the board simply shows all opportunities (no broken view).

## Testing
- `turnRecordFilterIntoGqlOperationFilter` unit tests: added relation-leaf cases (incl. `= me`) asserting the FK form — 59/59.
- `SeedOnboardingPrivateViewService` unit test: asserts seeded view/fields/filter shapes — 3/3.
- Updated `workspace.service.spec` for the new dependency — 86/86.
- typecheck + lint green across twenty-front, twenty-server, twenty-shared.

## Not yet done
- End-to-end run (activate a fresh workspace and eyeball the seeded view) — recommended before merge.

### DIFF
--- a/packages/twenty-front/src/modules/object-record/advanced-filter/components/AdvancedFilterRelationTargetFieldSelectMenu.tsx
+++ b/packages/twenty-front/src/modules/object-record/advanced-filter/components/AdvancedFilterRelationTargetFieldSelectMenu.tsx
@@ -62,11 +62,6 @@ export const AdvancedFilterRelationTargetFieldSelectMenu = ({
       ? sourceFieldMetadataItem.relation.targetObjectMetadata.id
       : null;
 
-  // A many-to-one relation can be selected as a filter leaf: it filters by the
-  // relation's foreign key (e.g. company.accountOwnerId = X), a single-hop
-  // equality the backend resolves on the already-joined table. Selecting one
-  // does not drill further into the related object, so no multi-hop traversal
-  // is composed here.
   const { filterableFieldMetadataItems: relationTargetFields } =
     useFilterableFieldMetadataItems(targetObjectMetadataId ?? '');
 
--- a/packages/twenty-front/src/modules/object-record/advanced-filter/components/AdvancedFilterRelationTargetFieldSelectMenu.tsx
+++ b/packages/twenty-front/src/modules/object-record/advanced-filter/components/AdvancedFilterRelationTargetFieldSelectMenu.tsx
@@ -62,16 +62,13 @@ export const AdvancedFilterRelationTargetFieldSelectMenu = ({
       ? sourceFieldMetadataItem.relation.targetObjectMetadata.id
       : null;
 
-  const { filterableFieldMetadataItems: allTargetFields } =
+  // A many-to-one relation can be selected as a filter leaf: it filters by the
+  // relation's foreign key (e.g. company.accountOwnerId = X), a single-hop
+  // equality the backend resolves on the already-joined table. Selecting one
+  // does not drill further into the related object, so no multi-hop traversal
+  // is composed here.
+  const { filterableFieldMetadataItems: relationTargetFields } =
     useFilterableFieldMetadataItems(targetObjectMetadataId ?? '');
-
-  // The backend supports a single hop only. Exclude many-to-one relations
-  // from the target list so the user can't compose multi-hop traversals
-  // (e.g. Person → Company → ParentCompany) that the dispatcher would
-  // collapse back to a filter-by-id on the intermediate relation.
-  const relationTargetFields = allTargetFields.filter(
-    (field) => !isManyToOneRelationField(field),
-  );
 
   if (
     !isDefined(sourceFieldMetadataItem) ||
--- a/packages/twenty-server/src/engine/core-modules/workspace/services/workspace.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/workspace/services/workspace.service.ts
@@ -65,6 +65,7 @@ import { WorkspaceCacheStorageService } from 'src/engine/workspace-cache-storage
 import { getWorkspaceSchemaName } from 'src/engine/workspace-datasource/utils/get-workspace-schema-name.util';
 import { WorkspaceDataSourceService } from 'src/engine/workspace-datasource/workspace-datasource.service';
 import { PrefillLogicFunctionService } from 'src/engine/workspace-manager/standard-objects-prefill-data/services/prefill-logic-function.service';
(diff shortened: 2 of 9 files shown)


packages/twenty-front/src/modules/object-record/advanced-filter/components/AdvancedFilterRelationTargetFieldSelectMenu.tsx
```diff
@@ -62,17 +62,14 @@ export const AdvancedFilterRelationTargetFieldSelectMenu = ({
       ? sourceFieldMetadataItem.relation.targetObjectMetadata.id
       : null;
 
-  const { filterableFieldMetadataItems: allTargetFields } =
+  // A many-to-one relation can be selected as a filter leaf: it filters by the
```
FelixMalfait: Remove this comment

packages/twenty-server/src/engine/core-modules/workspace/services/workspace.service.ts
```diff
@@ -371,6 +373,26 @@ export class WorkspaceService extends TypeOrmQueryService<WorkspaceEntity> {
       schemaName: getWorkspaceSchemaName(workspace.id),
     });
 
+    // Seed a private onboarding kanban for the workspace creator. Best-effort:
```
FelixMalfait: There is already a place where we handle view seeds (something like "prefill views"), why couldn't we handle it there?

packages/twenty-server/src/engine/workspace-manager/standard-objects-prefill-data/services/seed-onboarding-private-view.service.ts
```diff
@@ -0,0 +1,135 @@
...
+import { ViewService } from 'src/engine/metadata-modules/view/services/view.service';
+
+const SEEDED_VIEW_CARD_FIELD_SIZE = 150;
+
+// Seeds a private "My Pipeline" kanban for the workspace creator: opportunities
+// grouped by stage, filtered to the accounts they own
+// (company.accountOwner = me). It introduces new users to kanban views, nested
+// relation filters and private (per-user) views through a single example view.
+@Injectable()
+export class SeedOnboardingPrivateViewService {
```
FelixMalfait, line 24: Why couldn't we re-use/adapt the existing view seeding service?

FelixMalfait:
Thanks all — pushed `69ee0bd6b0` addressing the feedback.

### @FelixMalfait
- **"Handle it where we handle view seeds (prefill views)"** — moved. Seeding now runs inside `prefillCreatedWorkspaceRecords` (after the record transaction, next to `prefillWorkflowCommandMenuItems`) instead of a standalone `activateWorkspace` step.
- **"Re-use/adapt the existing view seeding service"** — I looked hard at hosting this in the `twenty-standard-application` flat-metadata system (what seeds the standard "By Stage" kanban). It structurally can't host *this* view, for two reasons:
  1. **Per-user owner.** A private view needs `createdByUserWorkspaceId` = the workspace creator. Standard views are keyed by a static `universalIdentifier` with values identical across every workspace, `createdByUserWorkspaceId` is a `toCompare` field, and the sync (`synchronizeTwentyStandardApplicationOrThrow({ workspaceId })`) isn't even passed a `userId`. There's nowhere to put a per-creator value in the shared catalog.
  2. **Managed ⇒ resurrected.** Standard views are permanently managed: delete one and the next sync recreates it and reverts edits. An onboarding view should be create-once.

  The nested "me" filter itself *is* expressible there — but the owner + one-time semantics aren't, so it stays a runtime `ViewService` call (now minimal and folded into the prefill flow). Happy to revisit if you had a different service in mind.
- **"Remove this comment"** — done.

### sentry — `prefill-companies` returns `workspaceMember_id`
False positive: `.select('id')` aliases to `id`, not `workspaceMember_id` (the `alias_` prefix only appears for entity-relative selects like `.select('workspaceMember.id')`), and the sibling `prefill-opportunities.util.ts` already relies on this exact pattern for `ownerId`. I made it explicit anyway — `.select('id', 'id')` — so there's zero ambiguity.

### cubic
- **Arbitrary workspace member** — fixed: `ORDER BY "createdAt" ASC` so the account owner is deterministically the creator (the only member at activation time).
- **Frontend fallback loads the wrong object** — fixed: the source-field fallback now applies only when there is no relation-target leaf; a leaf that's set but unresolvable throws (like the adjacent guard) rather than loading the source object's records.

All green: typecheck + lint (front/server/shared), `workspace.service.spec`, and the seed-service + relation-leaf unit tests.

FelixMalfait:
Splitting this up. Reopening as two focused PRs: (1) enable many-to-one relations as advanced-filter leaves (the frontend change, self-contained and reviewable on its own), and (2) the onboarding view seeding, separately. This PR is superseded by #1.

PR #21116, opened 2026-06-01
Front component s3 redirect

prastoin:
# Introduction
Unload the server of the file stream when possible
Also fix inconsistent pipeline exception management

Needs to highly be QA, not sure how the cors will behave here

### DIFF
--- a/packages/twenty-server/src/engine/metadata-modules/front-component/front-component.service.ts
+++ b/packages/twenty-server/src/engine/metadata-modules/front-component/front-component.service.ts
@@ -1,13 +1,14 @@
 import { Injectable } from '@nestjs/common';
 
-import { Readable } from 'stream';
-
 import { FileFolder } from 'twenty-shared/types';
 import { isDefined } from 'twenty-shared/utils';
 
 import { ApplicationService } from 'src/engine/core-modules/application/application.service';
 import { type FlatApplication } from 'src/engine/core-modules/application/types/flat-application.type';
 import { FileStorageService } from 'src/engine/core-modules/file-storage/file-storage.service';
+import { type FileResponse } from 'src/engine/core-modules/file/types/file-response.type';
+import { getContentDisposition } from 'src/engine/core-modules/file/utils/get-content-disposition.utils';
+import { TwentyConfigService } from 'src/engine/core-modules/twenty-config/twenty-config.service';
 import { WorkspaceManyOrAllFlatEntityMapsCacheService } from 'src/engine/metadata-modules/flat-entity/services/workspace-many-or-all-flat-entity-maps-cache.service';
 import { findFlatEntityByIdInFlatEntityMapsOrThrow } from 'src/engine/metadata-modules/flat-entity/utils/find-flat-entity-by-id-in-flat-entity-maps-or-throw.util';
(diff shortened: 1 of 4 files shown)


packages/twenty-server/src/engine/metadata-modules/front-component/front-component.service.ts
```diff
@@ -315,11 +317,29 @@ export class FrontComponentService {
...
+      expiresInSeconds: this.twentyConfigService.get(
+        'STORAGE_S3_PRESIGNED_URL_EXPIRES_IN',
+      ),
+      responseContentType: mimeType,
+      responseContentDisposition: getContentDisposition(mimeType),
     });
+
+    if (presignedUrl) {
+      return { type: 'redirect', presignedUrl };
+    }
```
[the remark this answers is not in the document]
prastoin: overkill imo

packages/twenty-server/src/engine/metadata-modules/front-component/front-component.service.ts
```diff
@@ -315,11 +317,29 @@ export class FrontComponentService {
...
       resourcePath: frontComponent.builtComponentPath,
+    };
+
+    const presignedUrl = await this.fileStorageService.getPresignedUrl({
+      ...resourceIdentifier,
+      expiresInSeconds: this.twentyConfigService.get(
+        'STORAGE_S3_PRESIGNED_URL_EXPIRES_IN',
+      ),
+      responseContentType: mimeType,
+      responseContentDisposition: getContentDisposition(mimeType),
```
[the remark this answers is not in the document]
prastoin: Wow that's a nice catch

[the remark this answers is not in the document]
prastoin: Not applicable for front components but accurate for the public assets though 🤔

packages/twenty-server/src/engine/metadata-modules/front-component/front-component.service.ts
```diff
@@ -315,11 +317,29 @@ export class FrontComponentService {
...
     });
+
+    if (presignedUrl) {
+      return { type: 'redirect', presignedUrl };
```
charlesBochet, line 338: why do we need a type + still supporting streaming?

prastoin: Thats incase the presigned isn't enabled on your twenty instance

prastoin: Seems to be an already common pattern across the codebase

charlesBochet: mmh OK!
Not a fan of these "non-deterministic" patterns where we fallback without knowing why

I would prefer
```
if (isPresignedEnabled) { presign path, early return }
stream()
```

rather than
```
tryToPresign()
if works { early return}
stream()
```

But if that's the pattern elsewhere LGTM, approving as this is not a big deal

charlesBochet (approved):

packages/twenty-server/src/engine/metadata-modules/front-component/front-component.service.ts
prastoin: Agreed

PR #21117, opened 2026-06-01
Fix front component pointer/mouse event coordinates

### LINKED ISSUE #21000: FrontComponent sandbox: pointer event handlers fire, but all position fields (clientX/Y, pageX/Y, movementX/Y…) are stripped to undefined
## Describe the bug

In a `defineFrontComponent` widget, React's pointer event handlers (`onPointerDown`, `onPointerMove`, `onPointerUp`, `onPointerCancel`, `onPointerLeave`) **are invoked** by the Worker bridge — but every position field of the event object is `undefined`. The same goes for `e.nativeEvent` and `e.currentTarget.getBoundingClientRect()`.

Net result: any drag-based interaction (signature pad, free-hand drawing, custom slider, lasso select, panning) is impossible inside an FC, because the handler receives `pointerdown` / `pointermove` but has no idea **where** the pointer is.

This is a regression-vs-expectation in 2.x. In ≤ 2.4 the move-type events were not even forwarded (the React props bag only contained `aria-hidden, style, onClick, children`). 2.8 forwards the handler invocations — which suggested drawing should now work — but the payload is empty.

## To Reproduce

Minimal repro FC:

```tsx
import React, { useState } from 'react';
import { defineFrontComponent } from 'twenty-sdk/define';

const Probe = () => {
  const [info, setInfo] = useState('— move the pointer over the box —');
  const dump = (e: React.PointerEvent<SVGSVGElement>) => {
    const a: any = e;
    setInfo(
      `type=${e.type} c=${a.clientX},${a.clientY} ` +
      `p=${a.pageX},${a.pageY} s=${a.screenX},${a.screenY} ` +
      `m=${a.movementX},${a.movementY} x=${a.x},${a.y} ` +
      `nativeEvent=${typeof a.nativeEvent} ` +
      `rect=${!!e.currentTarget?.getBoundingClientRect?.()}`,
    );
  };
  return (
    <svg width={480} height={120} style={{ background: '#fff', touchAction: 'none' }}
         onPointerDown={dump} onPointerMove={dump} onPointerUp={dump}>
      <text x={4} y={14} fontSize={11} fill="red" fontFamily="monospace">{info}</text>
    </svg>
  );
};

export default defineFrontComponent({
  universalIdentifier: '<some-uuid>',
  name: 'probe-pointer-event-fields',
  component: Probe,
});
```

1. `yarn twenty app:publish --private -r <remote> && yarn twenty app:install -r <remote>`
2. Drop the widget on any record page layout.
3. Drag a mouse (or touch on mobile) across the SVG.

### Actual output rendered inside the SVG

```
type=pointermove c=undefined,undefined p=undefined,undefined s=undefined,undefined m=undefined,undefined x=undefined,undefined nativeEvent=undefined rect=false
```

### Expected output

```
type=pointermove c=712,544 p=712,1158 s=712,544 m=3,1 x=712,544 nativeEvent=object rect=true
```

(Or at least one of the position fields populated so the FC can map pointer position to its own coordinate system.)

## Additional diagnostics

Reading the props directly off the rendered SVG element (from the host DOM, via DevTools console):

```js
const svg = document.querySelector('svg[width="480"]');
const k = Object.keys(svg).find(x => x.startsWith('__reactProps'));
Object.keys(svg[k]);
// → ["aria-hidden","width","height","style",
//    "onPointerDown","onPointerMove","onPointerUp","onPointerCancel","onPointerLeave",
//    "children"]
```

`getBoundingClientRect()` on the SVG **from the host DOM** returns a correct rect (`{left:608, top:502, width:481, height:133}`), so the element itself is real. The discrepancy lives in what the Worker bridge serializes onto the synthetic event handed to the FC handler.

Confirmed by counting handler invocations: a 350px-wide drag with the system mouse produces ~5 `pointermove` invocations inside the FC (so events flow), but each one reports `clientX = undefined`.

## Workarounds tried (all dead-ends)

| Source of coordinates | Result |
|---|---|
| `e.clientX/Y` | `undefined` |
| `e.pageX/Y` | `undefined` |
| `e.screenX/Y` | `undefined` |
| `e.movementX/Y` | `undefined` |
| `e.x / e.y` | `undefined` |
| `e.nativeEvent.offsetX/Y` | `e.nativeEvent` is `undefined` |
| `e.currentTarget.getBoundingClientRect()` | returns null/0 inside the FC |
| `setPointerCapture` | does not exist on the stub element |

`console.log` from inside the FC does **not** propagate to the host devtools console either, so the only practical way to inspect the event is to render its fields back into the DOM (text element in the same SVG).

## Impact

- **Hand-drawn signature widgets**: impossible.
- **Free-hand drawing / annotation**: impossible.
- **Custom sliders, dual-thumb range, color pickers**: impossible.
- **Drag-to-reorder lists, kanban-like drag**: impossible.
- **Drawing on a calendar / time axis**: impossible.

This is a fairly significant restriction for "do anything with an FC". In our HVAC CRM we have a hand-over confirmation widget (Předávka) which would benefit from a real signature pad on tablets; currently we fall back to a text-based confirmation stamp (cursive name + timestamp).

## Suggested fix area

The Worker-bridge event serializer (somewhere in `packages/twenty-front` that mirrors host React events into the FC Worker) should include the pointer position fields. Minimum:

- `clientX`, `clientY` (single-touch mouse / pen)
- `pageX`, `pageY` (for full-page coords)
- `movementX`, `movementY` (deltas for relative drawing without rect)
- `pointerType` (already present? worth confirming)
- `button`, `buttons` (already present? worth confirming)
- `pressure` (handy for variable-width signature)

Even just `clientX/Y` would unlock drawing if `getBoundingClientRect()` is also fixed on the SVG ref (or if the bridge exposes the host-DOM rect for the element).

## Environment

- Twenty server: **2.8.x** (self-hosted on Elestio)
- `twenty-sdk` / `twenty-client-sdk`: **2.8.0**
- Browser: tested on Chromium 12x desktop + iOS Safari (touch)
- App build flow: `yarn twenty app:publish --private` → server tarball, `yarn twenty app:install`
- For prior context: same probe in 1.14.217 / Twenty 2.4 — pointer handlers were not even forwarded; the regression-of-expectation is that 2.8 forwards the handler but not the data.

bosiraphael:
Fixes https://github.com/twentyhq/twenty/issues/21000

Front-component event handlers read standard event fields (event.clientX, event.offsetX, …), but these were always undefined. On the remote side, serialized event data was passed only as the CustomEvent's detail — and CustomEvent ignores every constructor option except detail, so the values lived at event.detail.clientX and never on the event object itself.

- Added `applySerializedEventProperties`,  to copy a curated allowlist of event-level keys onto the event. Element/target state (value, checked, files, scroll, media props) stays in `applySerializedEventTargetProperties`, applied to this (the dispatch element = event.target).
- Added x/y to `SerializedEventData` and to host-side serialization in `createHtmlHostWrapper`.
- Added an `svg-pointer `story + `createHtmlTagPointerStory`

Note: Also pinned @types/react to v18 so the renderer stops dragging in React 19 types and breaking typecheck.


packages/twenty-front/src/modules/settings/playground/components/__stories__/PlaygroundSetupForm.stories.tsx
```diff
@@ -2,6 +2,7 @@ import { PlaygroundSetupForm } from '@/settings/playground/components/Playground
 import { SnackBarDecorator } from '~/testing/decorators/SnackBarDecorator';
 import { type Meta, type StoryObj } from '@storybook/react-vite';
 import { ComponentDecorator, RouterDecorator } from 'twenty-ui/testing';
+import { SnackBarDecorator } from '~/testing/decorators/SnackBarDecorator';
```
Weiko: duplicate import it seems 👀

Weiko (approved):
LGTM @bosiraphael

PR #21118, opened 2026-06-01
fix(messaging): fix reply-quotation stripping that emptied email bodies

neo773:
some synced messages were stored with empty bodies, others with the entire reply thread re-quoted, planer was stripping entirely quoted forwards down to nothing and not trimming inline reply history at all

switched plaintext quote stripping to `email-reply-parser`, falling back to the full text when it strips everything so forwards don't end up blank. kept planer for the html path, and normalized body whitespac


packages/twenty-server/src/modules/messaging/message-import-manager/drivers/imap/services/imap-message-text-extractor.service.ts
```diff
@@ -1,20 +1,21 @@
...
 export class ImapMessageTextExtractorService {
   private readonly convertHtmlToText = createHtmlToTextConverter();
 
   extractTextWithoutReplyQuotations(parsed: ParsedEmail): string {
     if (parsed.text) {
-      const extractedText = planer.extractFrom(parsed.text, 'text/plain');
-
-      return safeDecodeURIComponent(extractedText);
+      return normalizeMessageText(
+        safeDecodeURIComponent(extractTextWithoutReplyQuotations(parsed.text)),
```
prastoin: That's starting to make a lot of nesting here, even if redundant lets consider creating intermediate const

prastoin (commented):
Confused why outlook/microsoft provider is not directly impacted here as the gmail and imap services

prastoin (approved):

packages/twenty-server/src/modules/messaging/message-import-manager/drivers/imap/services/imap-message-text-extractor.service.ts
neo773: Microsoft missed my mind somehow with it being in the picture, I wasn't happy so I refactored it to a utility it's now one consistent call across all drivers

packages/twenty-server/src/modules/messaging/message-import-manager/utils/extract-message-body-text.util.ts
```diff
@@ -0,0 +1,28 @@
...
+  text,
+  html,
+}: {
+  text?: string | null;
+  html?: string | null;
+}): string => {
+  const candidate = isNonEmptyString(text)
+    ? text
+    : isNonEmptyString(html)
+      ? createHtmlToTextConverter()(html)
```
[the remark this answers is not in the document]
neo773: intentional we cant do singleton usage, it's a multi tenant environment

neo773:
Not sure why CI is being flaky

PR #21120, opened 2026-06-01
fix: lowercase OAuth handle to prevent duplicate connected accounts

neo773:
Google and Microsoft connect flows stored the provider-returned email verbatim, so a re-auth with different casing (Sam@ vs sam@) missed the existing-account lookup and created a duplicate. Normalize the handle to lowercase at extraction, matching the SSO controller.

Note: not doing backfill for now

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/auth/controllers/microsoft-apis-auth.controller.ts
+++ b/packages/twenty-server/src/engine/core-modules/auth/controllers/microsoft-apis-auth.controller.ts
@@ -103,7 +103,7 @@ export class MicrosoftAPIsAuthController {
         );
       }
 
-      const handle = emails[0].value;
+      const handle = emails[0].value.toLowerCase();
 
       const connectedAccountId =
         await this.microsoftAPIsService.refreshMicrosoftRefreshToken({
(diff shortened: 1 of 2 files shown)


packages/twenty-server/src/engine/core-modules/auth/controllers/microsoft-apis-auth.controller.ts
```diff
@@ -103,7 +103,7 @@ export class MicrosoftAPIsAuthController {
         );
       }
 
-      const handle = emails[0].value;
+      const handle = emails[0].value.toLowerCase();
```
FelixMalfait, line 106: @claude No equivalent to change in IMAP?

FelixMalfait (approved):

packages/twenty-server/src/engine/core-modules/auth/controllers/microsoft-apis-auth.controller.ts
FelixMalfait: Let's do it then, no @neo773?

charlesBochet: Merging this one, but let's open a follow up PR @neo773

neo773: Will address

PR #21139, opened 2026-06-02
fix(front): recompute ExpandableList visible chips on resize

### LINKED ISSUE #12039: On record table, relation fields, we only display 3 items even when the cell is larger

dilanmelvin:
## Summary

Relation field cells in the record table render their chips through
`ExpandableList`, which measured how many chips fit only once (during the
children ref pass) and cached the cutoff. It recomputed on item-count and
hover changes, but never when the cell's available width changed — so a cell
measured while narrow stayed stuck on that count even after the column grew
wider. This is the "only ~3 items shown even when the cell is larger" bug.

This PR adds a `ResizeObserver` on the outer container that resets the first
hidden child index whenever the available width changes, so the list reveals
as many chips as fit (and re-trims when narrowed). The outer container is
observed because its width tracks the available width independently of how
many chips are currently rendered, which avoids a
measure → trim → shrink → re-measure feedback loop. The observer is cleaned
up on unmount.

## Test plan

- [x] Added a Storybook interaction test (`RecomputesVisibleChipsOnResize`)
      that renders the list in a narrow container, widens it, and asserts more
      chips become visible.
- [x] Verified the test fails without the fix and passes with it.
- [x] `oxfmt` and `oxlint` pass on the changed files.
- Manual: open a record table with a to-many relation field that has several
  linked records, widen the column, and confirm more chips appear.

Fixes #12039


charlesBochet (approved):
Thank you @dilanmelvin :)

dilanmelvin:
@charlesBochet You're welcome, and thank you for taking the time to review my contribution.

PR #21142, opened 2026-06-02
Provide additional logsobservability  to workflow runs (per node)

ijreilly:
Surfaces per-step "Logs" tabs in the workflow run side panel so users can see what each step actually did (model + tokens + tool calls for AI, console output for serverless functions, request/response for HTTP, recipients/body for Email).

ai_agent_without_websearch

## Storage

- New `stepLogs` JSONB column on the `workflowRun` workspace entity, typed as `Record<string, WorkflowRunStepLog>` (keyed by step id).
- Schema lives in `twenty-shared`: `workflowRunStepLogSchema` with a discriminated `details.type` union for `AI_AGENT | CODE | HTTP_REQUEST | EMAIL` — frontends and backends consume the same Zod-inferred type.
- Field is added to existing workspaces via a workspace upgrade command (`2-9 add-workflow-run-step-logs-field`); the standard-object metadata declares it for new workspaces.
- Writes happen atomically per step in `WorkflowRunStepLogWorkspaceService.setStepLog` using `jsonb_set`. That lets concurrent steps in the same run write their own keys without contending with the existing lock around `workflowRun.state`.
- Per-step payload is hard-capped at 256 KB; anything larger is dropped with a `logger.warn`, so a pathological tool call can never bloat a row. See below for more information.

## How logs are produced

**Aalmost everything was already being collected; this PR mostly persists and renders it.**

- **AI agent** — `AgentAsyncExecutorService` already tracked token usage, model id, native web-search count, and the AI SDK's `steps[]`. We map those into the log via `mapAiStepsToToolCallLogs` (`searchVector` stripped from record outputs, per-call input/output capped at 32/64 KB, max 200 tool calls per step). The only new measurement is a wall-clock `durationMs` taken around `executeAgent`, and we now fold native web-search cost into the displayed `totalCostInDollars` (it was already billed, just not shown).
- **Code / serverless function** — reuses the `console.log` output the function runner already returns (`logsByLevel`); `build-code-step-log.util` only repackages it.
- **HTTP request** — built from the action's existing input/output via `build-http-request-step-log.util`. No new signals collected.
- **Email (send / draft)** — added `sanitizedHtmlBody` + `plainTextBody` to the existing tool outputs (a small additive change), then `build-email-step-log.util` consumes them.

No additional AI inference or external calls are made for logging — the cost is a small CPU overhead per step plus the JSONB write.

## Security

The log surface intentionally shows whatever the workflow touched, which made redaction and sanitization the main design concern.

- **HTTP — secrets in headers**: existing `SENSITIVE_HEADER_NAMES` set (Authorization, Cookie, …) replaced with `[redacted]` in both request and response.
- **HTTP — secrets in URLs**: `SENSITIVE_URL_PARAM_NAMES` (e.g. `api_key`, `token`, `access_token`) replaced in the query string via `URL`-based parsing.
- **HTTP — secrets in bodies**: `SENSITIVE_BODY_KEY_REGEX` deep-walks JSON request/response bodies (object input or stringified JSON) and redacts matching keys. Applied to the `error` field too, since transport-layer errors sometimes embed structured payloads.
- **Email — XSS risk in body preview**: tool outputs now expose a server-side `sanitizedHtmlBody`; the log builder prefers it over the raw user-authored `input.body`, with `plainTextBody` as a second fallback. The original raw body is only used if sanitization didn't happen (e.g. tool failed before composing).
- **AI — internal/noisy data**: `searchVector` (Postgres tsvector strings) is stripped from record outputs returned by Twenty tools to avoid leaking internal full-text-search payloads.
- **DB bloat / runaway agents**: 256 KB per-step cap + 32 KB / 64 KB per-tool-call input/output cap + 200 tool calls per step.

logic_function

## More details on Log size & truncation

Logs are stored in `workflowRun.stepLogs` (JSONB), keyed by `stepId`.

### Per-step cap

Each step's log is hard-capped at **256 KB** (`MAX_STEP_LOG_BYTES` in `WorkflowRunStepLogWorkspaceService.setStepLog`).

For ~99% of workflows this is roomy — typical real-world sizes:
- Code / serverless function: 1–20 KB
- HTTP request: 5–70 KB
- Email: 5–30 KB
- AI agent (a handful of tool calls): 5–50 KB

### Two layers of bounding

1. **Per-field truncation** in each builder (before writing):
   - **Code**: ≤ 500 entries, ≤ 4 KB per message, ≤ 8 KB stack trace
   - **HTTP**: ≤ 32 KB per body (request + response), UTF-8 byte-aware
   - **Email**: ≤ 8 KB body preview, UTF-8 byte-aware
   - **AI agent**: ≤ 32 KB tool input, ≤ 64 KB tool output, ≤ 200 tool calls/step

2. **Global per-step safety net** at write time: if the assembled `stepLog` still exceeds 256 KB, the write is **dropped entirely** with a `logger.warn`. The workflow itself keeps running unaffected.

### What this means in practice

- **Safe**: workflow execution, step results, downstream steps — never blocked by log size.
- **Safe**: iterators (each iteration overwrites the previous log for that `stepId`, so they can't accumulate).
- **Safe**: step retries (same `stepId` is overwritten, not appended).
- **Possible**: an AI agent step with many large tool outputs (e.g., 50+ heavy `web_search` calls) can exceed 256 KB → the **entire** step's log is dropped, side panel shows "No logs were recorded for this step". The user has no explicit signal that the log was dropped due to size (only server-side warn).
- **Possible** (theoretical): a workflow with hundreds of distinct steps could push the row toward Postgres's internal ~256 MB jsonb limit. Beyond that, individual `jsonb_set` writes would error and be swallowed by the action's try/catch — workflow still completes.

### Possible future hardening (not in this PR)

- Replace "drop entire log" with a stub that preserves the summary card (cost, duration, status) and marks `truncated.reason = 'size_cap'`.
- Surface size-drops in the UI (similar to the existing `<StyledTruncatedNotice>`).
- Emit a metric so dropped logs are observable in dashboards.


packages/twenty-front/src/modules/workflow/workflow-run/observability/WorkflowRunStepLogsAiAgentDetail.tsx
```diff
@@ -0,0 +1,220 @@
...
+    return `$${dollars.toFixed(4)}`;
+  }
+
+  return `$${formatNumber(dollars, { decimals: 2 })}`;
+};
+
+const formatTokenCount = (tokens: number): string =>
+  formatNumber(tokens, { abbreviate: true, decimals: 1 });
+
+export const WorkflowRunStepLogsAiAgentDetail = ({
```
thomtrp, line 88: should not be in component folder. Should be in workflow-actions/ associated action, or if there is a specific folder for runs, in there. Same for other components

packages/twenty-front/src/modules/workflow/workflow-steps/components/WorkflowRunStepLogsEmailDetail.tsx
```diff
@@ -0,0 +1,219 @@
...
+  padding: ${themeCssVariables.spacing[3]};
+  word-break: break-word;
+`;
+
+type EmailDetails = Extract<WorkflowRunStepLog['details'], { type: 'EMAIL' }>;
+
+// HTML detection is intentionally permissive — the body is sanitized
+// upstream by `composeEmail`, so the trade-off is "render as HTML when it
+// looks like HTML, otherwise show the raw text". Body content lives in the
+// same trust boundary as workflow inputs.
```
thomtrp: please remove

packages/twenty-front/src/modules/workflow/workflow-steps/components/workflowRunStepLogsFormatters.ts
```diff
@@ -0,0 +1,28 @@
+// Shared formatters and constants for the `WorkflowRunStepLogs*` family of
+// components. Kept in a plain `.ts` file so it can be imported from both
+// component files and the sibling `workflowRunStepLogsStyles.ts` without
+// pulling in JSX.
+
+// SF Mono ships with macOS and is preferred for code-like content; the
+// stack falls back to Monaco/Consolas-style faces on other platforms.
```
thomtrp: same, let's not keep AI comments

packages/twenty-front/src/modules/workflow/workflow-steps/components/workflowRunStepLogsFormatters.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const formatDuration = (ms: number): string => {
+  if (ms < 1000) {
+    return `${ms}ms`;
+  }
+
+  return `${(ms / 1000).toFixed(1)}s`;
+};
+
+export const formatBytes = (bytes: number): string => {
```
thomtrp: should rather be 2 utils and a const than a component

packages/twenty-front/src/modules/workflow/workflow-run/observability/WorkflowRunStepLogsHttpRequestDetail.tsx
```diff
@@ -0,0 +1,256 @@
...
+  StyledSection,
+  StyledSectionTitle,
+  StyledStatusBadge,
+  StyledSummaryCard,
+  StyledSummaryHeader,
+  StyledTitle,
+} from '@/workflow/workflow-steps/components/workflowRunStepLogsStyles';
+
+// HTTP-specific badges and containers — kept local because none of them are
+// reused outside the HTTP step log view.
```
thomtrp: same

packages/twenty-front/src/modules/workflow/workflow-steps/components/WorkflowRunStepLogsEmailDetail.tsx
```diff
@@ -0,0 +1,219 @@
...
+            <StyledBodyContainer>
+              {looksLikeHtml(details.bodyPreview) ? (
+                <div
+                  dangerouslySetInnerHTML={{ __html: details.bodyPreview }}
+                />
+              ) : (
+                <pre style={{ margin: 0, whiteSpace: 'pre-wrap' }}>
+                  {details.bodyPreview}
+                </pre>
+              )}
```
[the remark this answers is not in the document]
ijreilly: ### Fix applied

WorkflowRunStepLogsEmailDetail.tsx:
Removed the looksLikeHtml regex heuristic and the <div dangerouslySetInnerHTML={{ __html: details.bodyPreview }} /> branch entirely.
details.bodyPreview is now always rendered inside a < pre > (new StyledBodyPre, replacing the previous inline style), so React's automatic text-node escaping applies unconditionally.
Added a comment at the styled component explaining why this is text-only (server fallback to raw input.body when the email tool fails before sanitization → can't trust the string at the dangerouslySetInnerHTML boundary).

### Why text and not a sanitizer
The whole frontend currently has zero dangerouslySetInnerHTML usages (this was the only one) and no DOMPurify/sanitize-html dependency. Adding a sanitizer just for one log view would be a significant surface increase.
This is an inspection view, not a preview. Showing raw HTML tags as escaped text is actually more useful for debugging — reviewers can see exactly what the workflow attempted to send (broken markup, failed variable substitutions, etc.).
React text escaping is bulletproof; sanitizer bypasses ship in the wild every year.

packages/twenty-server/src/modules/workflow/workflow-executor/workflow-actions/tool-executor-workflow-action.ts
```diff
@@ -100,13 +110,105 @@ export class ToolExecutorWorkflowAction implements WorkflowAction {
...
+      await this.persistEmailStepLog({
+        workflowRunId: runInfo.workflowRunId,
+        workspaceId: runInfo.workspaceId,
+        stepId: currentStepId,
+        mode: step.type === WorkflowActionType.SEND_EMAIL ? 'SEND' : 'DRAFT',
+        input: toolInput as WorkflowSendEmailActionInput,
+        output: toolOutput,
+        durationMs,
+      });
+    }
```
thomtrp: not a big fan of doing if-else everywhere, we already have one above. We should have another place in the flow to persist the logs, or do a switch

packages/twenty-server/src/modules/workflow/workflow-executor/workflow-actions/tool-executor-workflow-action.ts
```diff
@@ -100,13 +110,105 @@ export class ToolExecutorWorkflowAction implements WorkflowAction {
...
+
     return {
       result: toolOutput.result as object,
       error: toolOutput.error,
     };
   }
+
+  private async persistHttpRequestStepLog({
```
thomtrp: step specific function should not live there. Maybe tool executor needs to be split

packages/twenty-front/src/modules/workflow/workflow-steps/components/WorkflowRunStepLogsEmailDetail.tsx
```diff
@@ -0,0 +1,219 @@
...
+        </StyledRecipientsCard>
+      </StyledSection>
+
+      <StyledSection>
+        <StyledSectionTitle>{t`Body`}</StyledSectionTitle>
+        {isDefined(details.bodyPreview) && details.bodyPreview.length > 0 ? (
+          <>
+            <StyledBodyContainer>
+              {looksLikeHtml(details.bodyPreview) ? (
+                <div
```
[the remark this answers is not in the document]
ijreilly: fixed

packages/twenty-server/src/modules/workflow/workflow-executor/workflow-actions/http-request/utils/build-http-request-step-log.util.ts
```diff
@@ -0,0 +1,261 @@
...
+
+  if (bodyBytes <= MAX_BODY_BYTES) {
+    return { body: redacted, bodyBytes, bodyTruncated: false };
+  }
+
+  return {
+    body: `${redacted.slice(0, MAX_BODY_BYTES)}${TRUNCATION_SENTINEL}`,
+    bodyBytes,
+    bodyTruncated: true,
+  };
```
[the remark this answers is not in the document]
ijreilly: fixed

packages/twenty-server/src/engine/metadata-modules/ai/ai-agent-execution/utils/map-ai-steps-to-tool-call-logs.util.ts
```diff
@@ -0,0 +1,135 @@
...
+  } catch {
+    return TRUNCATION_SENTINEL;
+  }
+
+  if (serialized.length <= maxBytes) {
+    return value;
+  }
+
+  return `${serialized.slice(0, maxBytes)}${TRUNCATION_SENTINEL}`;
+};
```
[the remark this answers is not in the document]
ijreilly: fixed

packages/twenty-server/src/modules/workflow/workflow-executor/workflow-actions/http-request/utils/build-http-request-step-log.util.ts
```diff
@@ -0,0 +1,261 @@
...
+  return redacted;
+};
+
+const redactUrl = (rawUrl: string): string => {
+  let parsed: URL;
+
+  try {
+    parsed = new URL(rawUrl);
+  } catch {
+    return rawUrl;
```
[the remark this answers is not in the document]
ijreilly: fixed

packages/twenty-server/src/modules/workflow/workflow-executor/workflow-actions/code/code.workflow-action.ts
```diff
@@ -52,10 +58,44 @@ export class CodeWorkflowAction implements WorkflowAction {
...
+    workspaceId,
+    stepId,
+    result,
+  }: {
+    workflowRunId: string;
+    workspaceId: string;
+    stepId: string;
+    result: LogicFunctionExecuteResult;
+  }): Promise<void> {
+    try {
```
thomtrp, line 86: try catch logic seems duplicated everywhere. Should be common

packages/twenty-server/src/modules/workflow/workflow-executor/workflow-actions/ai-agent/ai-agent.workflow-action.ts
```diff
@@ -73,27 +80,103 @@ export class AiAgentWorkflowAction implements WorkflowAction {
...
+  }): Promise<void> {
+    if (!executionResult.modelId) {
+      return;
+    }
+
+    const toolCalls = executionResult.steps
+      ? mapAiStepsToToolCallLogs(executionResult.steps)
+      : [];
+
+    const details: AiAgentStepLogDetails = {
```
thomtrp: why not a builder? like buildCodeStepLog() / buildEmailStepLog()

packages/twenty-server/src/modules/workflow/workflow-executor/workflow-actions/tool-executor-workflow-action.ts
thomtrp: Each tool type already has its own Tool implementation (HttpTool, SendEmailTool, DraftEmailTool), its own build*StepLog util, and its own input type. Yet they're all funneled through one class that switches on type. This should be split into separate WorkflowAction implementations

packages/twenty-server/src/modules/workflow/workflow-executor/workflow-actions/tool-executor-workflow-action.ts
ijreilly: I split it !

packages/twenty-front/src/modules/workflow/workflow-run/observability/WorkflowRunStepLogsAiAgentDetail.tsx
```diff
@@ -0,0 +1,220 @@
...
+  StyledMetric,
+  StyledMetricLabel,
+  StyledMetricsRow,
+  StyledMetricValue,
+  StyledSection,
+  StyledSectionTitle,
+  StyledSummaryCard,
+  StyledSummaryHeader,
+  StyledTitle,
+} from '@/workflow/workflow-run/observability/workflowRunStepLogsStyles';
```
ijreilly, line 30: Image

packages/twenty-front/src/modules/workflow/workflow-run/observability/workflowRunStepLogsFormatters.ts
```diff
@@ -0,0 +1,19 @@
+export const formatDuration = (ms: number): string => {
+  if (ms < 1000) {
+    return `${ms}ms`;
+  }
+
+  return `${(ms / 1000).toFixed(1)}s`;
+};
+
+export const formatBytes = (bytes: number): string => {
```
thomtrp, line 9: should be utils

packages/twenty-front/src/modules/workflow/workflow-run/observability/WorkflowRunStepLogsAiAgentDetail.tsx
```diff
@@ -0,0 +1,220 @@
...
+  text-align: right;
+`;
+
+const StyledToolList = styled.div`
+  display: flex;
+  flex-direction: column;
+  gap: ${themeCssVariables.spacing[1]};
+`;
+
+const formatCost = (dollars: number): string => {
```
FelixMalfait, line 73: nit: this probably exist somewhere else for sure, same for the tokencount

packages/twenty-front/src/modules/workflow/workflow-run/observability/workflowRunStepLogsFormatters.ts
FelixMalfait, line 9: 2 export in 1 file ; should probably be shared utils

packages/twenty-server/src/engine/core-modules/tool/tools/email-tool/draft-email-tool.ts
```diff
@@ -57,6 +57,8 @@ export class DraftEmailTool implements Tool {
           ccRecipients: data.recipients.cc,
           bccRecipients: data.recipients.bcc,
           subject: data.sanitizedSubject,
+          sanitizedHtmlBody: data.sanitizedHtmlBody,
```
FelixMalfait, line 60: We try to have smaller tool response side to overloading the chat/mcp context. This will double a field that's already very large so it's probably not the right way to handle this
Do we need the plain text body too?

packages/twenty-server/src/engine/metadata-modules/ai/ai-agent-execution/services/agent-async-executor.service.ts
```diff
@@ -257,69 +260,83 @@ export class AgentAsyncExecutorService {
...
-        );
+        result = structuredResult.output as object;
       }
 
+      const resolvedModelId = registeredModel.modelId;
+      const tokenCostInDollars = this.aiBillingService.calculateCost(
+        resolvedModelId,
+        { usage: accumulatedUsage, cacheCreationTokens },
+      );
+      const totalCostInDollars =
```
FelixMalfait, line 323: It feels odd to store both dollar + credits. I think so far we only stored one and conversion was done at the very last step (converted for cloud user / not converted to credit for self-hosting users ; the cloud user only speak in credits)

packages/twenty-server/src/engine/metadata-modules/ai/ai-agent-execution/utils/map-ai-steps-to-tool-call-logs.util.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  const { value: truncatedValue, truncated } = truncateStringToUtf8ByteBudget(
+    serialized,
+    maxBytes,
+  );
+
+  return truncated ? truncatedValue : value;
+};
+
+export type MapAiStepsToToolCallLogsOptions = {
```
FelixMalfait, line 69: 1 export per file

thomtrp (commented):
Much better! 💪 

The architecture could still be improved IMO 🙈  

We now have a folder workflow-run/observability, but we have workflow-run related components in workflow-actions folder already. I would avoid that workflow-run/observability folder. Instead, I would put:
- logs related to a specific step, in the dedicated step folder in workflow-actions/step-name
- common logs related components in workflow-actions/components
- utils and styles should go in workflow-actions/utils

packages/twenty-server/src/engine/metadata-modules/ai/ai-agent-execution/utils/__tests__/map-ai-steps-to-tool-call-logs.util.spec.ts
```diff
@@ -0,0 +1,226 @@
...
+    ];
+
+    const result = mapAiStepsToToolCallLogs(steps);
+
+    expect(result).toHaveLength(1);
+    expect(result[0].state).toBe('error');
+    expect(result[0].errorMessage).toContain('Validation failed');
+  });
+
+  it('truncates oversized tool input and output', () => {
```
FelixMalfait, line 71: Not sure if we really need all this complexity / added code ; it feels like it's a protection that should be built not in the context of logs but more at the tool level (what Etienne is working). The cost of storing a million characters in postgres is many many orders of magnitude below the token cost for a million characters so the first to feel an outsized tool call would be end-user (and since we charge it becomes a "rich problem" for us because we can afford DB). 
Not saying we shouldn't do anything but this is typical LLM over-reaching imo in the sense that it's not something a human developer would have done ; LLMs are lazy when it's about doing useful deep refactoring but not lazy enough when it's about writing small side-transformation scripts like that :)

packages/twenty-shared/src/workflow/schemas/workflow-run-step-log-schema.ts
```diff
@@ -0,0 +1,125 @@
...
+});
+
+// We intentionally keep the runtime schema permissive: the column is a
+// JSONB blob written by the server and the consumers don't validate
+// individual `details` shapes. The strict per-step type (with the
+// discriminated `details` union) lives in `WorkflowRunStepLog` and is
+// applied at the boundaries that *produce* logs (server-side writers).
+// Tighter zod parsing here would collapse the discriminated union to `{}`
+// when inferred through `z.record`, breaking front-end indexing.
+export const workflowRunStepLogsSchema = z.record(z.string(), z.unknown());
```
FelixMalfait, line 125: One export per file?

packages/twenty-server/src/engine/core-modules/tool/tools/email-tool/draft-email-tool.ts
FelixMalfait: I didn't understand at first you added logs for the send email node too. It now makes more sense that you want to show both html + non-html (still think the sanitization belongs elsewhere?). 

The problem is tools belong to AI chat + workflow. 
It feels like @etiennejouan is going into a direction where the output of a tool depends on the context it is being called from. AI chat would get limited responses by default + TOON format ; workflows would get JSON + long format by default ; MCP could maybe parametize to get both depending on user preference (e.g. a script wants json, a mcp in chat might not need it)

packages/twenty-server/src/utils/truncate-string-to-utf8-byte-budget.util.ts
```diff
@@ -0,0 +1,31 @@
...
+export type Utf8TruncationResult = {
+  value: string;
+  originalBytes: number;
+  truncated: boolean;
+};
+
+export const utf8ByteLengthOf = (value: string): number =>
+  Buffer.byteLength(value, 'utf8');
+
+export const truncateStringToUtf8ByteBudget = (
```
FelixMalfait, line 12: Maybe also over-engineered vs simple truncation? to challenge, I could be wrong

packages/twenty-server/src/modules/workflow/workflow-executor/workflow-actions/tool-backed/tool-backed.workflow-action.ts
```diff
@@ -0,0 +1,114 @@
...
+import { type WorkflowAction } from 'src/modules/workflow/workflow-executor/workflow-actions/types/workflow-action.type';
+import { WorkflowRunStepLogWorkspaceService } from 'src/modules/workflow/workflow-runner/workflow-run/workflow-run-step-log.workspace-service';
+
+type BuildStepLogArgs<TInput> = {
+  input: TInput;
+  output: ToolOutput;
+  durationMs: number;
+};
+
+export abstract class ToolBackedWorkflowAction<
```
FelixMalfait, line 22: Not sure if this abstraction is really necessary since it's only used twice, feels like a service or something like that could be more suited than using class inheritance which adds complexity

FelixMalfait: Ok I see this is a port of something similar so maybe we want to keep it to avoid tackling a problem that already was there. Maybe comment is not relevant. But let's just challenge it quickly

packages/twenty-server/src/engine/core-modules/application-logs/utils/strip-ansi-escapes.util.ts
```diff
@@ -0,0 +1,6 @@
+const ANSI_CSI_REGEX = /\u001B\[[0-?]*[ -/]*[@-~]/g;
+
+const ANSI_OSC_REGEX = /\u001B\][^\u0007\u001B]*(?:\u0007|\u001B\\)/g;
+
+export const stripAnsiEscapes = (value: string): string =>
```
FelixMalfait, line 5: Feels a bit too local to be solved here now? We probably want to do a treatment for all logs surfaces? (e.g. logs page etc)

packages/twenty-server/src/engine/core-modules/application-logs/utils/parse-application-log-lines.ts
```diff
@@ -18,14 +19,14 @@ export const parseApplicationLogLines = (rawLogs: string): ParsedLogLine[] => {
       return {
         timestamp: new Date(match[1]),
         level: match[2],
-        message: match[3],
+        message: stripAnsiEscapes(match[3]),
       };
     }
```
FelixMalfait, line 25: unrelated tip. In claude code, always run /simplify on your PR, helps remove the slop (not that there is much here, but it's good to end the PR with that + challenge the complexity and idiomaticity with one or two prompts)

FelixMalfait (approved):
Great work! Made some comments but I think they are not a blocker for merge you can deploy to prod tomorrow behind flag and address my comments in a followup PR

PR #21143, opened 2026-06-02
fix(ai-chat) - fix breaking cache because of signed files

etiennejouan:
Each message that contains a file attachment gets a different URL on each new turn. 
The AI SDK passes these URLs into the model messages, so from the model's perspective every previous file-containing message has changed → no cache hit is possible for any conversation containing file attachments, regardless of how many turns ago the file was shared.


etiennejouan:
The model loses the ability to directly inspect the file content on turns after the first. Whether this is a real problem depends on the use case:

Common case (text + one image, multi-turn Q&A): Not a problem, the model extracts meaning from the image in turn 1 and reasons from its own response in subsequent turns.
Problematic case: User attaches an image and then 3 turns later asks "zoom in on the top-left corner", the model can only see [attached: image.png].
Fixing it properly requires to enable agent to access a file via Id. To be done

PR #21144, opened 2026-06-02
refactor(twenty-server): collapse workspace migration build orchestrator

charlesBochet:
## Why

`WorkspaceMigrationBuildOrchestratorService.buildWorkspaceMigration` was 1117 lines of 30 near-identical copy-pasted blocks (one per metadata entity) plus a 165-line ordered `actions: [...]` literal. Adding a new entity meant copying ~30 lines and hoping the boilerplate stayed in sync; the actual per-entity execution order was buried in the noise.

## Gain

- Orchestrator: **1117 → 388 lines**
- Net diff: **-558 lines**
- Adding a new entity is now a single line in the registry
- Execution order is visible at a glance

### DIFF
--- a/packages/twenty-server/src/engine/workspace-manager/workspace-migration/services/workspace-migration-build-orchestrator.service.ts
+++ b/packages/twenty-server/src/engine/workspace-manager/workspace-migration/services/workspace-migration-build-orchestrator.service.ts
@@ -56,9 +56,6 @@ import { WorkspaceMigrationViewActionsBuilderService } from 'src/engine/workspac
 import { WorkspaceMigrationWebhookActionsBuilderService } from 'src/engine/workspace-manager/workspace-migration/workspace-migration-builder/builders/webhook/workspace-migration-webhook-actions-builder.service';
 import { WorkspaceEntityMigrationBuilderService } from 'src/engine/workspace-manager/workspace-migration/workspace-migration-builder/services/workspace-entity-migration-builder.service';
 
-// One step of the orchestrator pipeline. Each task is bound at construction
-// time to a specific (metadataName, builderService) pair so its generic type
-// parameter is preserved. The orchestrator just runs them sequentially.
 type EntityActionsBuilderTask = {
   metadataName: AllMetadataName;
   run: (context: EntityActionsBuilderRunContext) => Promise<void>;
@@ -95,10 +92,6 @@ const createEntityActionsBuilderTask = <T extends AllMetadataName>(
       return;
     }
 
-    // Dynamic lookup of `fromToAllFlatEntityMaps[flatEntityMapsKey]` returns
-    // the union over all metadata names; the cast narrows it back to the
-    // bound `T` of this task. This is sound because the registration tied
-    // `metadataName` and `builderService` together at construction time.
     const result = await builderService.validateAndBuild({
       additionalCacheDataMaps,
       buildOptions,
@@ -111,11 +104,6 @@ const createEntityActionsBuilderTask = <T extends AllMetadataName>(
     if (result.status === 'fail') {
       orchestratorFailureReport[metadataName].push(...result.errors);
     } else {
-      // TS sees `orchestratorActionsReport[T]` as the intersection of all
-      // per-metadata-name action records (since indexing a mapped type by a
-      // generic union is invariant) and rejects an assignment from a single
-      // concrete value. Narrowing the report once to the bound `T` makes the
-      // write type-safe under the per-task generic.
       const reportForCurrentMetadata = orchestratorActionsReport as Record<
         T,
         MetadataUniversalWorkspaceMigrationActionsRecord<T>
@@ -128,10 +116,6 @@ const createEntityActionsBuilderTask = <T extends AllMetadataName>(
 
 @Injectable()
 export class WorkspaceMigrationBuildOrchestratorService {
-  // The execution order matters: builders mutate the shared optimistic
-  // flat-entity maps as they run, and later builders depend on those
-  // mutations (e.g. role-bound entities expect roles to already be in the
-  // maps, view children expect their parent view to exist, ...).
   private readonly entityActionsBuilderTasksInExecutionOrder: ReadonlyArray<EntityActionsBuilderTask>;
 
   constructor(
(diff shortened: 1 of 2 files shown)


packages/twenty-server/src/engine/workspace-manager/workspace-migration/services/workspace-migration-build-orchestrator.service.ts
```diff
@@ -42,43 +52,222 @@ import { WorkspaceMigrationViewFilterActionsBuilderService } from 'src/engine/wo
...
+    additionalCacheDataMaps,
+    buildOptions,
+    fromToAllFlatEntityMaps,
+    optimisticAllFlatEntityMaps,
+    orchestratorActionsReport,
+    orchestratorFailureReport,
+    workspaceId,
+  }) => {
+    const flatEntityMapsKey = getMetadataFlatEntityMapsKey(metadataName);
+    const fromToFlatEntityMaps = fromToAllFlatEntityMaps[flatEntityMapsKey];
```
prastoin: would force cast here instead of two
```ts
      from: fromToFlatEntityMaps.from as MetadataUniversalFlatEntityMaps<T>,Add a comment on  line R98Add diff commentMarkdown input:  edit mode selected.WritePreviewAdd a suggestionHeadingBoldItalicQuoteCodeLinkUnordered listNumbered listTask listMentionReferenceMore Formatting tools items 4Saved repliesAdd FilesPaste, drop, or click to add filesCancelCommentStart a review
      to: fromToFlatEntityMaps.to as MetadataUniversalFlatEntityMaps<T>,
```

packages/twenty-server/src/engine/workspace-manager/workspace-migration/services/workspace-migration-build-orchestrator.service.ts
```diff
@@ -42,43 +52,222 @@ import { WorkspaceMigrationViewFilterActionsBuilderService } from 'src/engine/wo
...
+      from: fromToFlatEntityMaps.from as MetadataUniversalFlatEntityMaps<T>,
+      to: fromToFlatEntityMaps.to as MetadataUniversalFlatEntityMaps<T>,
+      workspaceId,
+    });
+
+    if (result.status === 'fail') {
+      orchestratorFailureReport[metadataName].push(...result.errors);
+    } else {
+      orchestratorActionsReport[metadataName] =
+        result.actions as OrchestratorActionsReport[T];
```
prastoin: Unless I'm mistaken force cast should not be necessary here

prastoin (approved):

prastoin (commented):
Note: this refactor depends on the fact that we're updating the optimisticCache through mutation

packages/twenty-server/src/engine/workspace-manager/workspace-migration/services/workspace-migration-build-orchestrator.service.ts
```diff
@@ -42,43 +52,222 @@ import { WorkspaceMigrationViewFilterActionsBuilderService } from 'src/engine/wo
...
+        workspaceMigrationWebhookActionsBuilderService,
+      ),
+      createEntityActionsBuilderTask(
+        ALL_METADATA_NAME.applicationVariable,
+        workspaceMigrationApplicationVariableActionsBuilderService,
+      ),
+      createEntityActionsBuilderTask(
+        ALL_METADATA_NAME.connectionProvider,
+        workspaceMigrationConnectionProviderActionsBuilderService,
+      ),
```
prastoin, line 286: I'm afraid this might be some how rigid at some point 🤔 
Also the constructor params order declaration is really important, we should add a comment about not interfering with its intergrity

packages/twenty-server/src/engine/workspace-manager/workspace-migration/services/workspace-migration-build-orchestrator.service.ts
```diff
@@ -143,778 +332,18 @@ export class WorkspaceMigrationBuildOrchestratorService {
...
-          {
-            additionalCacheDataMaps,
-            from: fromFlatConnectionProviderMaps,
-            to: toFlatConnectionProviderMaps,
-            buildOptions,
-            dependencyOptimisticFlatEntityMaps: optimisticAllFlatEntityMaps,
-            workspaceId,
-          },
-        );
+    const runContext: EntityActionsBuilderRunContext = {
```
Weiko, line 353: Nit: Is this variable useful for readability?

Weiko (approved):

packages/twenty-server/src/engine/workspace-manager/workspace-migration/services/workspace-migration-build-orchestrator.service.ts
charlesBochet: Done in 9e0cc95. Introduced a `FromToMetadataUniversalFlatEntityMaps<T>` type alias and cast once at the indexed-access site.

packages/twenty-server/src/engine/workspace-manager/workspace-migration/services/workspace-migration-build-orchestrator.service.ts
charlesBochet: Tried removing the cast — CI typecheck failed with:

```
error TS2322: Type 'MetadataUniversalWorkspaceMigrationActionsRecord<T>' is not assignable to type 'OrchestratorActionsReport[T]'.
  Type ... is not assignable to type 'MetadataUniversalWorkspaceMigrationActionsRecord<"agent"> & MetadataUniversalWorkspaceMigrationActionsRecord<"applicationVariable"> & ... 28 more ... & MetadataUniversalWorkspaceMigrationActionsRecord<...>'.
```

This is TS mapped-type invariance on write: indexing `OrchestratorActionsReport` with a generic key `T` widens the expected value to the intersection of all 30 variants. Restored the cast in ce59b65 with an inline comment so it does not look gratuitous.

packages/twenty-server/src/engine/workspace-manager/workspace-migration/services/workspace-migration-build-orchestrator.service.ts
charlesBochet: Two parts:

**On the params-order concern**: NestJS DI resolves by **type**, not by position, so the constructor parameter order is irrelevant — what matters is the order of items in `entityActionsBuilderTasksInExecutionOrder`. Added a comment on the array in 9e0cc95 documenting this (and the optimistic-cache mutation dependency you flagged at the top of the review).

**On rigidity**: it is deliberate. The execution order is a load-bearing invariant of the orchestrator — each builder mutates the shared optimistic cache and downstream entities depend on upstream ones being processed first. Auto-registration via decorators or per-builder ordering metadata would trade this explicitness for indirection. Happy to revisit if you have a concrete pain point in mind.

charlesBochet: For now it looks like we don't need flexibility at this layer

packages/twenty-server/src/engine/workspace-manager/workspace-migration/services/workspace-migration-build-orchestrator.service.ts
charlesBochet: checked, it's needed!

PR #21145, opened 2026-06-02
Allow functional iframes in front components while blocking sandbox escapes

### LINKED ISSUE #19899: FrontComponent <iframe> is force-sandboxed with sandbox="", which blocks JS, forms, popups and same-origin
## Description

When rendering an `<iframe>` from a custom front component (Twenty App via the SDK), Twenty hard-codes `sandbox=""` on every iframe element via `FORCED_PROPS_BY_TAG` in `createHtmlHostWrapper.ts`. As a result, any `sandbox` attribute passed by the app developer is silently overridden, and the embedded page is sandboxed in the most restrictive possible way: **no scripts, no forms, no popups, no same-origin, nothing**.

This effectively makes `<iframe>` unusable in front components for any non-trivial embedding (analytics dashboards, OAuth flows, custom mini-apps, payment widgets, etc.).

## Reproduction

1. Create a front component (Twenty App) that renders an iframe pointing to a page that runs JavaScript:

```tsx
import { defineFrontComponent } from 'twenty-sdk';

export const MyComponent = defineFrontComponent({
  name: 'my-iframe-component',
  Component: () => (
    <iframe
      src="https://example.com/my-page-with-js"
      sandbox="allow-scripts allow-same-origin"
      style={{ width: '100%', height: '400px' }}
    />
  ),
});
```

2. Build & install the app into a Twenty workspace, render the component on any page.
3. Open DevTools → Elements panel and inspect the rendered iframe.

**Observed**:

```html
<iframe src="https://example.com/my-page-with-js" sandbox="" style="..."></iframe>
```

The `sandbox` attribute provided in JSX is overridden to an empty string. JavaScript inside the iframe does not execute.

**Expected**: the `sandbox` attribute provided by the app developer should be respected, with sensible defaults applied only when the developer does not explicitly set one.

## Root cause

[`packages/twenty-front-component-renderer/src/host/utils/createHtmlHostWrapper.ts`](https://github.com/twentyhq/twenty/blob/main/packages/twenty-front-component-renderer/src/host/utils/createHtmlHostWrapper.ts) (around line 170 on `main`):

```ts
const FORCED_PROPS_BY_TAG: Record<string, Record<string, unknown>> = {
  iframe: { sandbox: '' },
};

export const createHtmlHostWrapper = (htmlTag: string) => {
  // ...
  return ({ children, ...props }: WrapperProps) =>
    React.createElement(
      htmlTag,
      { ...filterProps(props), ...forcedProps }, // ← forcedProps wins
      isVoid ? undefined : children,
    );
};
```

`forcedProps` is spread **after** the user's props, so any user-supplied `sandbox` is overwritten. The default value `''` is also the most restrictive possible sandbox value (HTML spec: an empty `sandbox` attribute applies *all* restrictions).

## Impact

This blocks several real-world Twenty App use cases:

- Embedding analytics dashboards (Superset, Metabase, Grafana — same family as #16200 but for front components, not the widget iframe).
- Embedding OAuth / auth flows.
- Embedding any custom mini-app that needs JS to function.
- Our own use case: an Avito (Russian classifieds) chat panel embedded as a front component on Opportunity pages — we had to monkey-patch the production bundle to remove this default before the iframe became usable.

## Proposed fix

Two reasonable directions, happy to send a PR for whichever the maintainers prefer.

### Option A — respect user-provided `sandbox`, keep a sensible default

```ts
const DEFAULT_PROPS_BY_TAG: Record<string, Record<string, unknown>> = {
  iframe: { sandbox: 'allow-scripts allow-forms allow-popups' },
};

return ({ children, ...props }: WrapperProps) =>
  React.createElement(
    htmlTag,
    { ...defaultProps, ...filterProps(props) }, // ← user props win
    isVoid ? undefined : children,
  );
```

Apps that don't set `sandbox` get a safe default (scripts allowed, no `same-origin` so the embed cannot read host cookies); apps that do set `sandbox` opt in to whatever they need.

### Option B — keep current "force" semantics but fix the value

If the goal is genuinely to **prevent** apps from relaxing sandboxing, then at least the current default should not be `''` (which is the *strictest* mode and breaks even basic interactive content). A value like `'allow-scripts'` would still keep apps from accessing host cookies but allow JS to run.

This option is much more restrictive than what most app authors will want; Option A is preferred.

## Additional context

- Discovered on `twentycrm/twenty:latest` (current release as of April 2026; server `package.json` reports `0.2.1`).
- Confirmed in source on `main` at commit `3eeaebb0`.
- Related (different code path, already fixed): #16200 — widget iframe was missing `allow-same-origin`. Our case is the front-component iframe, which is a separate render path.

Happy to send a PR — just let me know which direction you prefer.

bosiraphael:
Fixes https://github.com/twentyhq/twenty/issues/19899

Front component iframes were previously forced to `sandbox=""`, which fully locks them down: no scripts, no forms, no popups. That broke any legitimate embedded content (maps, widgets, embeds) developers tried to render.

But we can't just trust the app-provided sandbox value either: tokens like allow-same-origin or allow-top-navigation would let a malicious embed escape the sandbox and hijack the host Twenty tab.

- Add `sanitizeIframeSandbox`, which keeps the iframe useful while enforcing security:
applies a safe default (allow-scripts allow-forms allow-popups) when no sandbox is set
always forces allow-scripts so embeds work
- strips dangerous tokens (`allow-same-origin`, all `allow-top-navigation`*, `allow-popups-to-escape-sandbox`), case-insensitively
- Wire it into `createHtmlHostWrapper` so every `iframe` rendered by a front component is sanitized.
- Add unit tests for the sanitizer and Storybook interaction tests asserting dangerous sandboxes are stripped.


packages/twenty-front-component-renderer/src/__stories__/shared/test-utils/matchers/expectSandboxSanitized.ts
```diff
@@ -0,0 +1,47 @@
...
+  'allow-top-navigation-to-custom-protocols',
+  'allow-popups-to-escape-sandbox',
+];
+
+type ExpectSandboxSanitizedParams = {
+  canvas: Canvas;
+  timeout?: number;
+};
+
+export const expectSandboxSanitized = async ({
```
charlesBochet: naming, this should be reflecting that this is for storybook and Iframe

charlesBochet (approved):
Approving! I feel an allow list (instead of a deny list) would be more "safe" regarding sandbox keywords but LGTM

Weiko (approved):

PR #21147, opened 2026-06-02
feat: allow many-to-one relations as advanced filter leaves

FelixMalfait:
## What

Lets a many-to-one relation be selected as the **leaf** of an advanced (nested) filter. Previously the nested-field submenu excluded relations, so you could filter `Opportunities WHERE company.Name contains X` but not `Opportunities WHERE company.accountOwner = me`.

## How it works

Selecting a relation leaf filters by its **foreign key** — `company.accountOwnerId = X` — a single hop the backend already resolves on the joined table (`{ company: { accountOwnerId: { in: [...] } } }`). It is **not** a multi-hop traversal: filtering on a *scalar field of* the related record (e.g. `company.accountOwner.name`) stays excluded, since that needs a second join the backend caps at one hop.

Two changes:
- **`AdvancedFilterRelationTargetFieldSelectMenu`** — stop excluding many-to-one relations from the nested-field submenu.
- **`ObjectFilterDropdownRecordSelect`** — resolve the record picker's object from the *leaf* relation's target (e.g. WorkspaceMember, including the "Me" pin) rather than the source relation's object. The source-field fallback applies only when there is no leaf.

## Testing
- Added `turnRecordFilterIntoRecordGqlOperationFilter` unit cases asserting a relation leaf (and `= me`) compiles to the FK form — 59/59.
- typecheck + lint green (twenty-front, twenty-shared).

Seeding an onboarding view that uses this filter will follow in a separate PR.


packages/twenty-front/src/modules/object-record/object-filter-dropdown/components/ObjectFilterDropdownRecordSelect.tsx
```diff
@@ -79,8 +89,32 @@ export const ObjectFilterDropdownRecordSelect = ({
     throw new Error('fieldMetadataItemUsedInFilterDropdown is not defined');
   }
 
+  // For a nested relation filter (e.g. company → accountOwner) the records to
+  // choose from belong to the leaf relation's target object (WorkspaceMember),
+  // not the source relation's object (Company). Only a direct relation filter
+  // (no leaf) falls back to the source field; a leaf that is set but cannot be
+  // resolved must not load records from the source object.
```
FelixMalfait: Make comment shorter (2-lines max)

packages/twenty-shared/src/utils/filter/__tests__/turnRecordFilterIntoGqlOperationFilter.test.ts
```diff
@@ -1038,5 +1044,61 @@ describe('turnRecordFilterIntoRecordGqlOperationFilter', () => {
...
+      });
+
+      expect(result).toEqual({
+        company: {
+          accountOwnerId: { in: ['550e8400-e29b-41d4-a716-446655440000'] },
+        },
+      });
+    });
+
+    // The seeded "My Pipeline" onboarding view (company.accountOwner = me)
```
FelixMalfait: Not accurate anymore? is this from the previous PR?

packages/twenty-front/src/modules/object-record/object-filter-dropdown/components/ObjectFilterDropdownRecordSelect.tsx
```diff
@@ -79,8 +89,32 @@ export const ObjectFilterDropdownRecordSelect = ({
...
+
+  const effectiveFieldMetadataItem = isDefined(
+    relationTargetFieldMetadataIdUsedInDropdown,
+  )
+    ? relationTargetFieldMetadataItem
+    : fieldMetadataItemUsedInFilterDropdown;
+
+  if (!isDefined(effectiveFieldMetadataItem)) {
+    throw new Error('effectiveFieldMetadataItem is not defined');
+  }
```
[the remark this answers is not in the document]
FelixMalfait: Fixed in a26ec03 — the picker no longer throws on an unresolved leaf, it renders nothing instead. A bare `return null` here would break rules-of-hooks (the record-picker hooks run after), so I split the hook-bearing picker into an inner component and the wrapper bails cleanly.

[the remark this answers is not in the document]
FelixMalfait: Fixed in a26ec03 — returns null gracefully instead of throwing. Since the picker hooks run after this point, a direct `return null` tripped rules-of-hooks, so I extracted them into an inner component and the wrapper bails without a conditional-hook violation.

packages/twenty-front/src/modules/object-record/object-filter-dropdown/components/ObjectFilterDropdownRecordSelect.tsx
FelixMalfait: Trimmed to 2 lines in a26ec03.

packages/twenty-shared/src/utils/filter/__tests__/turnRecordFilterIntoGqlOperationFilter.test.ts
FelixMalfait: Right — that was leftover from the combined PR. Removed the seeded-view reference and reworded it to just describe the `= me` → FK behaviour (a26ec03).

PR #21150, opened 2026-06-02
refactor(server): merge duplicate `TypeOrmModule.forFeature` calls in `MessagingMessageCleanerModule`

Rpaudel379:
Combined two separate `TypeOrmModule.forFeature()` calls into one. Both registered entities on the default data source, so no behavioral change.  Repositories for WorkspaceEntity and MessageChannelEntity remain injectable as before.
Cleaner imports, one less redundant call.


prastoin (approved):
Please try focusing on picking existing issue 🙏 
Still thanks for your contribution !

PR #21158, opened 2026-06-02
[Call Recording] Add standard object

ehconitin:
Adds **Call Recording** as a first-class standard object (Twenty's flat-metadata
standard-object system), with a hidden junction to calendar events and a backfill
command for existing workspaces. Everything is gated behind the
`IS_CALL_RECORDING_ENABLED` feature flag.

### What's included
- **`CallRecording`**: audio/video files, transcript, status, recording policy,
  timing, external bot/recording ids. Label identifier is `meetingOccurrenceKey`.
- **`CallRecordingCalendarEventAssociation`**: hidden junction linking a recording
  to a calendar event (dedupes one bot to many subscribers of the same meeting).
- Full metadata graph via the flat-metadata builders: fields, indexes, views,
  view fields/groups, record page layout, and navigation items.
- **Metadata-only reverse relation** on `CalendarEvent`: present in standard
  metadata, omitted from the TS entity class to avoid expanding recursive
  nested-insert types.
- **Upgrade command (2.9.0)** backfilling active/suspended workspaces:
  - Creates the full graph; idempotent (skips when it already exists).
  - Moves a colliding custom `callRecording` object aside to `callRecordingOld`
    (numeric suffix if that name is also taken).
  - Navigation items (commands)  are flag-gated by `universalIdentifier`, so a custom object
    reusing the name is never gated.

### QA
Run locally against existing workspaces (with and without a name collision) and a
freshly created workspace:
- [x] Backfill, collision: custom `callRecording` renamed to `callRecordingOld`;
  standard graph created.
- [x] Backfill, no collision: standard graph created; unrelated custom object untouched.
- [x] Idempotent: re-run is a no-op, with no duplicate metadata and counts unchanged.
- [x] New workspace via `init()` produces an identical graph to the backfill
  (`universalIdentifier` set-diff = 0).
- [x] Label identifier (`meetingOccurrenceKey`) holds position 0 in non-widget views.
- [x] Nav items gated behind the feature flag; collision-renamed object's nav
  expression re-pointed to its new name.
- [x] Unit tests cover collision name resolution and nav-gating logic.


prastoin:
Hey there please ping me once this is ready to be reviewed !

packages/twenty-server/src/modules/call-recording/standard-objects/call-recording-calendar-event-association.workspace-entity.ts
```diff
@@ -0,0 +1,21 @@
...
+import { type EntityRelation } from 'src/engine/workspace-manager/workspace-migration/types/entity-relation.interface';
+import { type CallRecordingWorkspaceEntity } from 'src/modules/call-recording/standard-objects/call-recording.workspace-entity';
+import { type CalendarEventWorkspaceEntity } from 'src/modules/calendar/common/standard-objects/calendar-event.workspace-entity';
+
+const EVENT_EXTERNAL_ID_FIELD_NAME = 'eventExternalId';
+
+export const SEARCH_FIELDS_FOR_CALL_RECORDING_CALENDAR_EVENT_ASSOCIATION: FieldTypeAndNameMetadata[] =
+  [{ name: EVENT_EXTERNAL_ID_FIELD_NAME, type: FieldMetadataType.TEXT }];
+
+export class CallRecordingCalendarEventAssociationWorkspaceEntity extends BaseWorkspaceEntity {
```
FelixMalfait: Do we really want to do a many to many? I feel like it's okay to have separated call recording from calendar event which gives us some flexibility (e.g. if two recording happen in a row we can keep one attached and the other is orphaned but we don't have to erase it entirely), but I feel like many to many is overkill

packages/twenty-server/src/engine/workspace-manager/twenty-standard-application/utils/field-metadata/compute-call-recording-standard-flat-field-metadata.util.ts
```diff
@@ -0,0 +1,508 @@
...
+          color: 'gray',
+        },
+      ],
+    },
+    standardObjectMetadataRelatedEntityIds,
+    dependencyFlatEntityMaps,
+    twentyStandardApplicationId,
+    now,
+  }),
+  recordingPolicy: createStandardFieldFlatMetadata({
```
FelixMalfait: I don't get this, isn't it a calendar event thing rather than call recording?

packages/twenty-server/src/engine/workspace-manager/twenty-standard-application/utils/field-metadata/compute-call-recording-standard-flat-field-metadata.util.ts
```diff
@@ -0,0 +1,508 @@
...
+    standardObjectMetadataRelatedEntityIds,
+    dependencyFlatEntityMaps,
+    twentyStandardApplicationId,
+    now,
+  }),
+  meetingOccurrenceKey: createStandardFieldFlatMetadata({
+    objectName,
+    workspaceId,
+    context: {
+      fieldName: 'meetingOccurrenceKey',
```
FelixMalfait: I don't get it, is this the external id?

packages/twenty-server/src/engine/workspace-manager/twenty-standard-application/utils/field-metadata/compute-call-recording-standard-flat-field-metadata.util.ts
```diff
@@ -0,0 +1,508 @@
...
+          color: 'gray',
+        },
+      ],
+    },
+    standardObjectMetadataRelatedEntityIds,
+    dependencyFlatEntityMaps,
+    twentyStandardApplicationId,
+    now,
+  }),
+  sourceApplicationId: createStandardFieldFlatMetadata({
```
FelixMalfait: No need for "source" imo, on objectMetadata is just applicationId for example?

packages/twenty-server/src/engine/workspace-manager/twenty-standard-application/utils/field-metadata/compute-call-recording-standard-flat-field-metadata.util.ts
```diff
@@ -0,0 +1,508 @@
...
+    standardObjectMetadataRelatedEntityIds,
+    dependencyFlatEntityMaps,
+    twentyStandardApplicationId,
+    now,
+  }),
+  failureReason: createStandardFieldFlatMetadata({
+    objectName,
+    workspaceId,
+    context: {
+      fieldName: 'failureReason',
```
FelixMalfait: This won't be very nice for end-users to see but why not... I feel like ideally we would prefer to let them know though other means!

FelixMalfait:
Did you check the Fireflies app to see if it will be easy to adapt it and push the data in there?

ehconitin:
chatted on Discord!

will be simplifying the schema!

packages/twenty-server/src/database/commands/upgrade-version-command/2-10/utils/get-standard-flat-entities-to-create-or-throw.util.ts
```diff
@@ -0,0 +1,62 @@
...
+        existingFlatEntityMaps.byUniversalIdentifier[universalIdentifier],
+      )
+    ) {
+      return [];
+    }
+
+    return [standardFlatEntity];
+  });
+
+export const getExistingOrStandardFlatEntityOrThrow = <
```
prastoin, line 36: ***Question:*** why do you need this here ?

packages/twenty-server/src/database/commands/upgrade-version-command/2-10/utils/call-recording-name-collision.util.ts
```diff
@@ -0,0 +1,204 @@
...
+        flatObjectMetadata.nameSingular,
+        flatObjectMetadata.namePlural,
+      ]),
+    ...additionalTakenNames,
+  ]);
+
+  for (let attempt = 0; attempt < MAX_OLD_NAME_ATTEMPTS; attempt++) {
+    const discriminator = attempt === 0 ? '' : `${attempt + 1}`;
+    const nameSingular = `${CALL_RECORDING_OLD_NAME_SINGULAR}${discriminator}`;
+    const namePlural = `${CALL_RECORDING_OLD_NAME_PLURAL}${discriminator}`;
```
prastoin, line 67: seems a bit overkill but ok !

packages/twenty-server/src/database/commands/upgrade-version-command/2-10/utils/get-standard-flat-entities-to-create-or-throw.util.ts
```diff
@@ -0,0 +1,62 @@
+import { isDefined } from 'twenty-shared/utils';
+
+import { type SyncableFlatEntity } from 'src/engine/metadata-modules/flat-entity/types/flat-entity-from.type';
+import { type FlatEntityMaps } from 'src/engine/metadata-modules/flat-entity/types/flat-entity-maps.type';
+
+export const getStandardFlatEntitiesToCreateOrThrow = <
```
prastoin, line 6: could have got away of this if consumed one degree lowest `validateBuildAndRunWorkspaceMigrationFromTo`

prastoin (approved):
Haven't tested ! Looks pretty strong
Left few comments

packages/twenty-server/src/database/commands/upgrade-version-command/2-10/2-10-workspace-command-1799000055000-sync-call-recording-standard-objects.command.ts
prastoin: ***bug:*** Unless i'm mistaken that's wrong it should be the entity to be renamed applicationUniversalIdentifier
Also they could be from different app technically.. so we might do each rename atomically

prastoin (changes requested):

packages/twenty-server/src/database/commands/upgrade-version-command/2-10/2-10-workspace-command-1799000055000-sync-call-recording-standard-objects.command.ts
```diff
@@ -0,0 +1,410 @@
...
+    if (
+      objectMetadataRenameUpdates.length > 0 ||
+      fieldMetadataRenameUpdates.length > 0
+    ) {
+      const renameResult =
+        await this.workspaceMigrationValidateBuildAndRunService.validateBuildAndRunWorkspaceMigration(
+          {
+            isSystemBuild: true,
+            applicationUniversalIdentifier:
+              workspaceCustomFlatApplication.universalIdentifier,
```
prastoin: you're assuming it's the custom, it might technically be any app
I feel like that's acceptable today but would not in a year because apps will be everywhere
At some point we might standardize this behavior with email fowarding etc

prastoin (approved):

packages/twenty-server/src/database/commands/upgrade-version-command/2-10/2-10-workspace-command-1799000055000-sync-call-recording-standard-objects.command.ts
ehconitin: okay!

PR #21160, opened 2026-06-02
feat: allow apps to add view fields to existing views (defineViewField)

### LINKED ISSUE #2468: Add defineViewField in twenty-sdk to extend existing view
Required to create a last-contacted-at field on Person object for example

martmull:
## Summary

Lets a Twenty application add **view fields (columns) to an existing view it does not own** — including standard views like the People index view — without redeclaring/owning that view. This mirrors the existing, working pattern by which an app adds a custom field to a standard object via `defineField` + `objectUniversalIdentifier`.

The asymmetry being removed was purely in the manifest schema: `ViewFieldManifest` only existed *nested* inside `ViewManifest.fields[]`, so adding a view field forced declaring a `ViewManifest` — which the sync treats as a view the app creates and owns, and rejects when the UID is a standard view's. Validation, persistence, the FK aggregator machinery, and uninstall cleanup were already generic and cross-app-safe, so no engine changes were needed.

### Changes
- **twenty-shared:** new top-level `StandaloneViewFieldManifest` (`ViewFieldManifest & { viewUniversalIdentifier }`), `Manifest.viewFields`, and a `SyncableEntity.ViewField` member.
- **twenty-sdk:** `defineViewField` (validates `universalIdentifier` + `viewUniversalIdentifier` + `fieldMetadataUniversalIdentifier`), CLI manifest assembly of a top-level `viewFields` list, and `dev:add viewField` scaffolding.
- **twenty-server:** one top-level loop over `manifest.viewFields` that reuses the existing `fromViewFieldManifestToUniversalFlatViewField` converter (already parameterized by `viewUniversalIdentifier`). No validator/persistence/aggregator changes.

### Notes for maintainers
- Confirm the `Manifest.viewFields` optionality convention — implemented as a **required** array to mirror `fields`/`views`.
- Two different apps adding a column for the same field to the same view conflicts on the existing unique `(fieldMetadataId, viewId)` partial index; the existing `flat-view-field-validator` duplicate check surfaces this as a structured validation error.
- `dev:add viewField` scaffolding is included (was optional in the plan).

## Test Plan
- [x] `twenty-shared` typecheck
- [x] `twenty-sdk` 364 unit tests + `buildManifest` assembly test (rich-app fixture) + typecheck + prettier
- [x] `twenty-server` typecheck + `lint:diff-with-main`
- [x] **Server integration suite** `successful-manifest-update-view-field.integration-spec.ts` (4/4):
  - standalone view field attaches to the standard `allPeople` view without recreating it (sync succeeds, no `INVALID_VIEW_DATA`/`ENTITY_ALREADY_EXISTS`)
  - uninstall removes the contributed column while the standard view + its columns remain intact
  - duplicate `(view, field)` rejected with `METADATA_VALIDATION_FAILED`
  - unknown target view rejected
- [x] Sibling `successful-manifest-update-field.integration-spec.ts` still green (no harness regression)

### DIFF
--- a/packages/twenty-apps/fixtures/rich-app/src/navigation-menu-items/post-cards.navigation-menu-item.ts
+++ b/packages/twenty-apps/fixtures/rich-app/src/navigation-menu-items/post-cards.navigation-menu-item.ts
@@ -3,7 +3,7 @@ import { NavigationMenuItemType } from 'twenty-shared/types';
 import { POST_CARD_UNIVERSAL_IDENTIFIER } from '../objects/post-card.object';
 
 export default defineNavigationMenuItem({
-  universalIdentifier: 'c1a2b3c4-0001-4a7b-8c9d-0e1f2a3b4c5d',
+  universalIdentifier: 'e8031eca-d6ea-4a4b-b828-38227dba896a',
   position: 0,
   type: NavigationMenuItemType.OBJECT,
   targetObjectUniversalIdentifier: POST_CARD_UNIVERSAL_IDENTIFIER,
--- a/packages/twenty-apps/fixtures/rich-app/src/viewFields/post-card-number-view-field.ts
+++ b/packages/twenty-apps/fixtures/rich-app/src/viewFields/post-card-number-view-field.ts
@@ -1,9 +1,10 @@
 import { defineViewField } from 'twenty-sdk/define';
+import { ALL_POST_CARDS_VIEW_ID } from '../views/all-post-cards.view';
 
 export default defineViewField({
   fieldMetadataUniversalIdentifier: '7b57bd63-5a4c-46ca-9d52-42c8f02d1df6',
   position: 5,
   universalIdentifier: 'cd582d11-ea21-4dc3-b9c1-0298ce3b6b54',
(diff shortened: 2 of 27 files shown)


prastoin (commented):
Hey just hijacking to precise that this will work only for the twenty standard app views

martmull:
> Hey just hijacking to precise that this will work only for the twenty standard app views

why?

prastoin:
> > Hey just hijacking to precise that this will work only for the twenty standard app views
> 
> why?

The validate build and run contextualises the intersection of the currently processed applicationId + twenty standard applicationId
If you try creating an integration test that tries to create a view field on another app view ( which is not twenty standard ) it should fail and view not found

Not sure of the use case here btw

I mean I'm ok with the fact to allow creating atomic view fields across the sdk but unsure of the plus value of creating view field on strangers views

martmull:
> e that this will work only for the twenty standard app views

oh ok i did not read that this way. Yes indeed this should work only for standard views. I will check a guard is implemented to make this certain

prastoin (commented):
Could you please implement an integration test that tries to create a view field from app B on an app A view ? it should be considered as failing

charlesBochet (approved):

PR #21161, opened 2026-06-02
fix: Edit Layout keeping the command menu open

### LINKED ISSUE #2460: Edit layout button should close the command menu
The edit layout button should not keep the command menu open. It currently looks like the edit layout command is meant to pin command menus, but it is not.

https://github.com/user-attachments/assets/d1f915a9-92b7-4287-b4ff-20e2ccacc942

parshipcy:
## Summary
- Fixes https://github.com/twentyhq/core-team-issues/issues/2460
- Engine/headless commands always skipped closing the menu (`closeSidePanelOnCommandMenuListExecution: false`), even when the item was not pinned. Edit Layout is `isPinned: false`, so the menu should close like other list-only actions.

## Approach
- Option 1 (I chose this one): Derive close behavior from `item.isPinned` -> pinned commands keep the menu open; non-pinned ones close it.
- Option 2 (not chosen): remove the engine command override totally and use the default close behavior for all commands.

Option 1 is more targeted: it fixed Edit Layout without changing pinned commands (e.g. Export progress in the menu list). Option 2 is simpler but widens the blast radius to every engine command clicked from the side panel list.

### Before
https://github.com/user-attachments/assets/70b8dc75-af00-4917-81a1-646381f571d5

### After
https://github.com/user-attachments/assets/f733d696-9330-4e6c-993b-8a8133c53e0d


Bonapara:
thanks! at 0:03 there is a flash and "edit layout" is disappearing from the menu. We should find a way to make the animation smoother

image

image

Bonapara:
@Weiko do you have advice on how to make the animation smoother?

parshipcy:
@Bonapara I found a solution for this. Should I raise a separate PR for it or include the fix here?

https://github.com/user-attachments/assets/32bf9ae6-06ce-4ba1-b3fb-920e1c75b242

could you check if it looks okay now?

parshipcy:
@Weiko @Bonapara So basically i tried to add a “wait until the panel is fully closed” step in the code:
1. Start closing the panel
2. wait until close animation finish  <- this part is newly added
3. Then run the command (mount + layout bar)

Bonapara:
let's wait for Weiko input. @parshipcy can you push the code in the PR?

charlesBochet (approved):
Thank you @parshipcy :)

PR #21168, opened 2026-06-03
fix: workspace root redirect to use last visited object

### LINKED ISSUE #21166: Bug: Home redirect always lands on first object alphabetically, ignores last-visited
### Description
Going to the workspace root (`/`) always redirects me to the same object: the first one alphabetically by `nameSingular`. It doesn't matter which object I visited last, and clearing browser cache/localStorage makes no difference. In my workspace a custom object called "Activities" (`nameSingular: "activity"`) always wins because "a" sorts first.

### Steps to reproduce
1. Have multiple objects, with one that sorts first alphabetically by `nameSingular`.
2. Visit a different object's index page, e.g. Companies.
3. Go to the workspace root `/`.
4. Expected: redirect to the last-visited object (Companies). Actual: redirect to the alphabetically-first object.

### Root cause
`useDefaultHomePagePath` uses `lastVisitedObjectMetadataItem` if it's set, otherwise it falls back to the first of `readableNonSystemObjectMetadataItems`, which is sorted with `a.nameSingular.localeCompare(b.nameSingular)`. The last-visited branch never fires, for two reasons.

First, `lastVisitedObjectMetadataItemIdState` is created via `createAtomState({ useLocalStorage: true })`, which calls `atomWithStorage(key, default, undefined, undefined)`, so `getOnInit` is unset and defaults to `false` in jotai 2.x. The atom is only ever accessed imperatively through `store.get`/`store.set` and is never subscribed via a hook, so it never mounts or hydrates. On a fresh load `store.get(...)` returns the default `null` and the code falls through to the alphabetical fallback. Note that the `useSessionStorage` and `useCookieStorage` branches in the same factory do pass `{ getOnInit: true }`, only the `useLocalStorage` branch doesn't.

Second, `defaultHomePagePath` is a `useMemo` that reads the atom imperatively and doesn't list it as a dependency, so it isn't reactive to in-session updates either.

### Suggested fix
Pass `{ getOnInit: true }` for the `useLocalStorage` branch in `createAtomState`, or for this atom specifically, so `store.get` returns the persisted value. Optionally have `useDefaultHomePagePath` subscribe to the atom (e.g. `useAtomValue`) and add it to the memo deps so it stays reactive. Worth adding a regression test that sets a last-visited id and asserts the redirect resolves to that object, since the current tests only cover the fallback.

### Environment
- Twenty `main` @ 6b5a956 (2026-06-02)
- jotai ^2.17.1

parshipcy:
## Summary
- Fixes #21166
- Visiting `/` always redirected to the alphabetically first object (e.g. Activities / Companies) instead of the last visited object.
- So the solution is: Reading last visited object from `localStorage` on page load and react when it changes, so `/` sends us back to the object we were on instead of always picking the first one alphabetically.

## Testing
- Open People (`/objects/people`), navigate to `/` -> lands on People, not the first object alphabetically
- Hard refresh on `/` after visiting another object -> still redirects to last visited
- Fresh workspace with no last visited -> still falls back to first readable non-system object (alphabetical)
- Unit tests

### Before:
https://github.com/user-attachments/assets/c3e0f125-fad1-4aca-abab-9995aeea2f13

### After:
https://github.com/user-attachments/assets/6d2f6b02-332d-4e4e-9f7c-6d4b489301ae

### DIFF
--- a/packages/twenty-front/src/modules/navigation/hooks/__tests__/useDefaultHomePagePath.test.ts
+++ b/packages/twenty-front/src/modules/navigation/hooks/__tests__/useDefaultHomePagePath.test.ts
@@ -2,9 +2,10 @@ import { currentUserState } from '@/auth/states/currentUserState';
 import { currentUserWorkspaceState } from '@/auth/states/currentUserWorkspaceState';
 import { metadataStoreState } from '@/metadata-store/states/metadataStoreState';
 import { useDefaultHomePagePath } from '@/navigation/hooks/useDefaultHomePagePath';
+import { lastVisitedObjectMetadataItemIdState } from '@/navigation/states/lastVisitedObjectMetadataItemIdState';
 import { AggregateOperations } from '@/object-record/record-table/constants/AggregateOperations';
 import { useSetAtomState } from '@/ui/utilities/state/jotai/hooks/useSetAtomState';
-import { jotaiStore } from '@/ui/utilities/state/jotai/jotaiStore';
+import { jotaiStore, resetJotaiStore } from '@/ui/utilities/state/jotai/jotaiStore';
 import { renderHook, waitFor } from '@testing-library/react';
 import { Provider as JotaiProvider } from 'jotai';
 import { createElement, useEffect, type ReactNode } from 'react';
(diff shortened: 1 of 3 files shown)


parshipcy:
**Do not merge** - waiting for product requirements

prastoin:
Hello there ! Thanks for your contribution
Lets reopen this when we've landed on a product direction on the related issue

PR #21169, opened 2026-06-03
feat(messaging): append the sender's Gmail signature to outbound email

sherifkozman:
Tracking issue: #21181


prastoin (changes requested):
Hey there @sherifkozman, thanks for your suggestion and contribution !
Please prefer opening a feature request issue rather than opening a PR out of the blue so this can be discussed in prior
The underneath need seems legit to me @neo773 what do you think ?

neo773:
@prastoin 
I think this is more of a product question rather than code
Do we want to inherit provider signatures or let user make their own in Twenty in Email composer

cc
@Bonapara

prastoin:
Yeah completely wasn't talking about this pr content

prastoin:
Let's open a discussion on GitHub

sherifkozman:
Reopened as https://github.com/twentyhq/twenty/pull/21182 with the CI fix (regenerated the `FeatureFlagKey` GraphQL codegen) and tracked via #21181 per the suggestion above.

PR #21172, opened 2026-06-03
Fix Unzipped size must be smaller than 250Mb error

martmull:
move dev dependencies into devDependencies

### DIFF
--- a/packages/twenty-sdk/package.json
+++ b/packages/twenty-sdk/package.json
@@ -70,14 +70,11 @@
   },
   "license": "AGPL-3.0",
   "dependencies": {
-    "@genql/cli": "^3.0.3",
-    "@genql/runtime": "^2.10.0",
     "@sniptt/guards": "^0.2.0",
     "axios": "^1.13.5",
     "chalk": "^5.3.0",
     "chokidar": "^4.0.0",
     "commander": "^12.0.0",
-    "dotenv": "^16.4.0",
     "esbuild": "^0.25.0",
     "graphql": "^16.8.1",
     "graphql-sse": "^2.5.4",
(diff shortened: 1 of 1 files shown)


prastoin (changes requested):
Genql runtime should be necessary imo ?

prastoin:
Or they could be removed as its twenty-client-sdk scope now

prastoin (commented):
Same why removing dotenv zod ?

prastoin (commented):
Uuid too

martmull:
we don't use dotenv and zod in twenty-sdk, uuid is not removed, just updated without comma in the end

packages/twenty-sdk/package.json
```diff
@@ -90,12 +88,10 @@
...
     "twenty-client-sdk": "workspace:*",
     "typescript": "^5.9.3",
-    "uuid": "^13.0.0",
-    "vite": "^7.0.0",
-    "vite-tsconfig-paths": "^4.2.1",
-    "zod": "^4.1.11"
+    "uuid": "^13.0.0"
   },
   "devDependencies": {
+    "@genql/cli": "^3.0.3",
```
prastoin, line 94: could be removed too I think

packages/twenty-sdk/package.json
```diff
@@ -90,12 +88,10 @@
     "tinyglobby": "^0.2.15",
     "twenty-client-sdk": "workspace:*",
     "typescript": "^5.9.3",
-    "uuid": "^13.0.0",
-    "vite": "^7.0.0",
-    "vite-tsconfig-paths": "^4.2.1",
-    "zod": "^4.1.11"
```
prastoin, line 96: should be restored imo ?

prastoin (commented):
Removed packages are tiny
I don't understand how this package json could be the 250MO root cause here

prastoin (approved):
This won't be retro effective on existing sdk instances
We should consider allow declaring a package.json per logic function

PR #21173, opened 2026-06-03
[WIP] Feat/marketing emails

neo773:
Marketing/campaign emails on top of the emailing-domain (SES) feature: send a broadcast to a hand-picked list, with per-customer-domain unsubscribe links and opt-out-only **unsubscribe topics**.

## Model

Standard objects (workspace schema, flat-metadata):
- `messageCampaign` — a campaign send (subject, body template, from address, status, list, optional unsubscribe topic).
- `messageList` + `messageListMember` — the hand-picked audience (person ↔ list join). A campaign's recipients are its list's members; everyone is sendable unless suppressed.

Core entities (`core` schema, workspace-scoped — readable by the public unsubscribe flow without a workspace context):
- `unsubscribeTopic` — an opt-out-only category (name, description, visibility). There is no opt-in subscription state.
- `messageSuppression` — the single consent store: a row with `unsubscribeTopicId` NULL is a global block; a row with an `unsubscribeTopicId` and reason `UNSUBSCRIBE` is a per-topic opt-out. Two partial unique indexes dedupe global vs per-topic rows (Postgres treats NULLs as distinct).
- `emailingDomain` — the workspace's SES sending domain, auto-provisioned when an email channel is added (and cleaned up when its last channel is removed), with verification status + DNS records.

Campaign messages reuse the existing `message` / `messageThread` / `messageParticipant` model — one outbound `message` per recipient with a `deliveryStatus` state machine.

## Sending

- `sendMessageCampaign` resolves the audience **under the caller's permissions**, creates the campaign, and enqueues a single fan-out job (the request never materializes per-recipient rows or jobs).
- The fan-out job materializes one QUEUED message per recipient (deterministic ids → idempotent re-runs, reconciles crash-orphaned rows) and fans out per-recipient send jobs carrying **only ids**.
- Each send job renders per-recipient `{{variable}}` merge fields and sends via `EmailingDomainSenderService`, which applies suppression (global + per-topic) and the unsubscribe footer/headers. Suppressed recipients are recorded `SKIPPED`.
- The campaign finalizes `SENT`, or `SENT_WITH_ERRORS` if any recipient terminally failed.
- `previewMessageCampaignAudience` returns a pre-send breakdown (total / without-email / duplicate / globally-unsubscribed / topic-unsubscribed / sendable), shown as a hint under the composer pickers.

## Unsubscribe

- Encrypted (AES-256-GCM) token carrying workspaceId, address, optional `unsubscribeTopicId`, `issuedAt`, and a `preview` flag.
- One-click POST (RFC 8058) + `mailto:` — topic-scoped when the token carries a topic, global otherwise.
- Preferences page: a checkbox per visible topic (checked = still receiving); submitting creates per-topic opt-outs for unchecked topics and lifts re-checked ones (UNSUBSCRIBE only — never `BOUNCE`/`COMPLAINT`, never a global block).
- A **Preview** action in settings opens the live page via a preview-claim token; opt-out POSTs are no-ops for preview tokens, so previewing never mutates state.
- SES webhooks: inbound unsubscribe + outbound bounce/complaint → suppression (race-safe against at-least-once delivery, with reason escalation that never downgrades).
- Per-customer unsubscribe hostname (Cloudflare DNS); sends are gated on it being active, except in LOG/demo mode.

## Architecture

Campaign orchestration, suppression, the sender, the unsubscribe controller, and the SES webhook handlers live in `src/modules/emailing` + `src/modules/messaging-webhooks` (the workspace-feature layer). `core-modules/emailing-domain` keeps the SES driver, domain provisioning, the `unsubscribeTopic` / `messageSuppression` core entities, and the unsubscribe token/hostname plumbing. Domain creation is validated (`CreateEmailingDomainInput` — domain-format regex, lowercased) before any value reaches SES or the unsubscribe hostname.

## Frontend

- Campaign composer side panel (from / list / unsubscribe topic / subject / body) with a live audience-preview hint.
- Email settings: email channels each showing their auto-provisioned sending domain in a single section (status + DNS records + a "Check verification" action), plus an **Unsubscribe Topics** section to create/manage topics and preview the recipient page. A demo-mode banner is shown when the LOG driver is active.


packages/twenty-server/src/utils/get-domain-name-by-email.ts
```diff
@@ -2,7 +2,10 @@ import { msg } from '@lingui/core/macro';
 import { isNonEmptyString } from '@sniptt/guards';
 
 import { UserInputError } from 'src/engine/core-modules/graphql/utils/graphql-errors.util';
+import { getDomainFromEmail } from 'src/utils/get-domain-from-email';
 
+// Asserting variant of getDomainFromEmail: throws a UserInputError when the
```
FelixMalfait: (comment for AI to fix)
The naming isn't clear, if it's the assertive variable then the name should reflect that and we shouldn't need a comment.  Do we need both?

packages/twenty-server/src/utils/get-domain-from-email.ts
```diff
@@ -0,0 +1,13 @@
+// The domain of an email address: everything after the last '@'. Splitting on
+// the last '@' (not the first) matters because a quoted local part may itself
+// contain one — the domain of "a@b"@example.com is example.com. Returns
+// undefined when the address has no '@', and preserves the original case.
+export const getDomainFromEmail = (email: string): string | undefined => {
```
FelixMalfait, line 1: Remove the comment in this file

packages/twenty-server/src/modules/messaging/common/query-hooks/message/apply-messages-visibility-restrictions.service.ts
```diff
@@ -90,7 +89,8 @@ export class ApplyMessagesVisibilityRestrictionsService {
             .filter(isDefined);
 
           if (messageChannels.length === 0) {
-            throw new NotFoundError('Associated message channels not found');
+            messages.splice(i, 1);
```
FelixMalfait, line 92: Can you explain this change? (for AI to reply). I don't understand the rationale

packages/twenty-server/src/modules/emailing/standard-objects/message-topic.workspace-entity.ts
```diff
@@ -0,0 +1,25 @@
...
+import { BaseWorkspaceEntity } from 'src/engine/twenty-orm/base.workspace-entity';
+import { type FieldTypeAndNameMetadata } from 'src/engine/workspace-manager/utils/get-ts-vector-column-expression.util';
+import { type EntityRelation } from 'src/engine/workspace-manager/workspace-migration/types/entity-relation.interface';
+import { type MessageCampaignWorkspaceEntity } from 'src/modules/emailing/standard-objects/message-campaign.workspace-entity';
+import { type MessageSuppressionWorkspaceEntity } from 'src/modules/emailing/standard-objects/message-suppression.workspace-entity';
+import { type MessageTopicSubscriptionWorkspaceEntity } from 'src/modules/emailing/standard-objects/message-topic-subscription.workspace-entity';
+
+const NAME_FIELD_NAME = 'name';
+
+export const SEARCH_FIELDS_FOR_MESSAGE_TOPIC: FieldTypeAndNameMetadata[] = [
```
FelixMalfait: 1 file = 1 export usually, do we need these two exports? Do similar workspace-entity files have two export? Unless it's the norm/majority then fix it

packages/twenty-server/src/modules/emailing/standard-objects/message-topic-subscription.workspace-entity.ts
```diff
@@ -0,0 +1,19 @@
+import { BaseWorkspaceEntity } from 'src/engine/twenty-orm/base.workspace-entity';
+import { type EntityRelation } from 'src/engine/workspace-manager/workspace-migration/types/entity-relation.interface';
+import { type MessageTopicWorkspaceEntity } from 'src/modules/emailing/standard-objects/message-topic.workspace-entity';
+import { type PersonWorkspaceEntity } from 'src/modules/person/standard-objects/person.workspace-entity';
+
+export class MessageTopicSubscriptionWorkspaceEntity extends BaseWorkspaceEntity {
+  // Denormalized copy of the topic name so the subscription is identifiable
+  // (its labelIdentifier) — a relation can't be a labelIdentifier in Twenty.
+  topicName: string | null;
```
FelixMalfait: I don't really understand why we need topicName and the comment above? This looks like  MessageTopicSubscription is a pivot table, we need to leverage JunctionRelation, I don't think it needs to have a human-understandable identifier really, no? feels like it could be out of sync / cause more issues than it solves

packages/twenty-server/src/modules/emailing/standard-objects/message-topic-subscription.workspace-entity.ts
```diff
@@ -0,0 +1,19 @@
...
+  status: string;
+  subscribedAt: Date | null;
+  unsubscribedAt: Date | null;
+  source: string;
```
FelixMalfait: What is source? I'd advocate to keep the data model not too big here, it's not obvious to me what would be the options for source but maybe it will be as I read this PR

packages/twenty-server/src/modules/emailing/standard-objects/message-suppression.workspace-entity.ts
```diff
@@ -0,0 +1,21 @@
...
+import { type FieldTypeAndNameMetadata } from 'src/engine/workspace-manager/utils/get-ts-vector-column-expression.util';
+import { type MessageTopicWorkspaceEntity } from 'src/modules/emailing/standard-objects/message-topic.workspace-entity';
+
+const EMAIL_ADDRESS_FIELD_NAME = 'emailAddress';
+
+export const SEARCH_FIELDS_FOR_MESSAGE_SUPPRESSION: FieldTypeAndNameMetadata[] =
+  [{ name: EMAIL_ADDRESS_FIELD_NAME, type: FieldMetadataType.TEXT }];
+
+export class MessageSuppressionWorkspaceEntity extends BaseWorkspaceEntity {
+  emailAddress: string;
```
FelixMalfait: Do we leverage the EMAIL field type? It should be an EMAILS field (with max email=1), this will help for a good UX for UI tables etc

packages/twenty-front/src/modules/activities/emails/hooks/useCampaignComposerState.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  const { sendMessageCampaign, loading } = useSendMessageCampaign();
+
+  const canSend =
+    messageTopicId !== null &&
+    fromAddress.trim().length > 0 &&
+    subject.trim().length > 0 &&
+    !loading;
+
+  const handleSend = useCallback(async () => {
```
FelixMalfait: Do we need useCallback? check

packages/twenty-server/src/utils/get-domain-name-by-email.ts
FelixMalfait: Renamed to `getDomainFromEmailOrThrow` (matches the codebase's `OrThrow` convention) and dropped the comment. Both variants are used: the soft one returns `undefined` for filtering paths (blocklist checks, internal-participant filtering), the throwing one raises a translated `UserInputError` where the email is user input (sign-up, invites, subdomain derivation). Fixed in e874ab99fd.

packages/twenty-server/src/utils/get-domain-from-email.ts
FelixMalfait: Removed in e874ab99fd.

packages/twenty-server/src/modules/messaging/common/query-hooks/message/apply-messages-visibility-restrictions.service.ts
FelixMalfait: Campaign sends materialize `message` records on a workspace-owned email-group channel. When that channel later disappears (emailing domain removed, account cleanup) the messages are left with no resolvable channel, and the old code threw `NotFoundError` — one orphaned record 500'd the whole message query for everyone in the workspace. The splice gives those messages the same treatment as messages whose channels the requester can't see (the existing fall-through at the bottom of this loop): they're dropped from the result instead of failing the request. Throwing made sense when a channel-less message could only mean corrupted data; with email-group sends it's a reachable state in a normal lifecycle.

packages/twenty-server/src/modules/emailing/standard-objects/message-topic.workspace-entity.ts
FelixMalfait: Checked: 28 of the 34 standard-object entity files pair the entity class with a `SEARCH_FIELDS_FOR_X` export (person, company, note, task…), so this follows the majority pattern. Leaving as is.

packages/twenty-server/src/modules/emailing/standard-objects/message-topic-subscription.workspace-entity.ts
FelixMalfait: Agreed — removed `topicName` and its denormalization in the subscription service. The join now only carries the two FKs plus status/timestamps, and its `labelIdentifierFieldMetadataName` is `id`, same as the other pivot (`messageListMember`). Done in e874ab99fd.

packages/twenty-server/src/modules/emailing/standard-objects/message-topic-subscription.workspace-entity.ts
FelixMalfait: It was a SELECT (FORM / IMPORT / API / MANUAL) recording how the subscription was created, but nothing ever passed anything other than the MANUAL default — removed the field and its type in e874ab99fd.

packages/twenty-server/src/modules/emailing/standard-objects/message-suppression.workspace-entity.ts
FelixMalfait: Switched to EMAILS with `settings: { maxNumberOfValues: 1 }`, same shape as `person.emails` (e874ab99fd). One consequence: EMAILS can't be a label identifier (`LABEL_IDENTIFIER_FIELD_METADATA_TYPES` only allows TEXT / FULL_NAME / UUID), so the object's labelIdentifier moved to `id`. The (emailAddress, topic) unique index binds to `emailAddressPrimaryEmail` via the composite's unique-constraint subfield.

packages/twenty-front/src/modules/activities/emails/hooks/useCampaignComposerState.ts
FelixMalfait: Removed it — the memoization bought nothing here: its dependency list includes every piece of composer state, so the identity changed on each keystroke anyway, and the hotkey binding re-registers regardless because the page rebuilt its wrapper every render. `handleSend` is now a plain function guarded by `canSend` (which also blocks re-entry while the send mutation is in flight), passed straight to `useHotkeysOnFocusedElement`.

packages/twenty-server/src/modules/emailing/services/message-campaign.service.ts
```diff
@@ -0,0 +1,602 @@
...
+        where: { id: messageId },
+      });
+
+      // QUEUED is a first attempt, FAILED a previous attempt of this same job
+      // being retried by the queue. Everything else (SENT, BOUNCED, COMPLAINED)
+      // is terminal: never send twice.
+      if (
+        message === null ||
+        (message.deliveryStatus !== CAMPAIGN_MESSAGE_DELIVERY_STATUS.QUEUED &&
+          message.deliveryStatus !== CAMPAIGN_MESSAGE_DELIVERY_STATUS.FAILED)
```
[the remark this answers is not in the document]
FelixMalfait: Fixed in 25cf6192e7 — the catch now wraps only the provider call. Once SES accepts, a bookkeeping failure (SENT update, association update) still fails the job, but the persisted SENT status makes the retry hit the terminal-status guard and no-op instead of resending. The remaining at-least-once window is the provider call itself timing out after acceptance, which no client-side structure can remove.

packages/twenty-server/src/engine/core-modules/emailing-domain/services/unsubscribe-token.service.ts
```diff
@@ -0,0 +1,56 @@
...
+    return Buffer.from(
+      this.secretEncryptionService.encryptVersioned(
+        JSON.stringify(payload) as PlaintextString,
+      ),
+    ).toString('base64url');
+  }
+
+  verify(token: string): UnsubscribeTokenPayload | null {
+    try {
+      const decrypted = this.secretEncryptionService.decryptVersioned(
```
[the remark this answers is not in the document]
FelixMalfait: No compatibility shim needed: the unsubscribe flow (token service, controller, List-Unsubscribe headers) ships entirely in this PR and the feature is dark behind IS_EMAIL_GROUP_ENABLED — no tokens of the old format have been issued outside preview environments.

packages/twenty-server/src/engine/workspace-manager/twenty-standard-application/utils/object-metadata/create-standard-flat-object-metadata.util.ts
```diff
@@ -239,6 +239,190 @@ export const STANDARD_FLAT_OBJECT_METADATA_BUILDERS_BY_OBJECT_NAME = {
...
+        namePlural: 'messageSuppressions',
+        labelSingular: i18nLabel(msg`Email Suppression`),
+        labelPlural: i18nLabel(msg`Email Suppressions`),
+        description: i18nLabel(
+          msg`Email addresses blocked from receiving emails after a bounce, complaint, or unsubscribe`,
+        ),
+        icon: 'IconMailOff',
+        isSystem: true,
+        isUIReadOnly: true,
+        labelIdentifierFieldMetadataName: 'id',
```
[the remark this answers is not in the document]
FelixMalfait: Known tradeoff, made deliberately: EMAILS is not an allowed label-identifier type (LABEL_IDENTIFIER_FIELD_METADATA_TYPES is TEXT / FULL_NAME / UUID), and the review asked for the EMAILS type on emailAddress for table UX. messageSuppression is an isSystem log object with no browsable views, and the address is the first column wherever rows do surface — same pattern as the messageListMember pivot.

packages/twenty-server/src/modules/emailing/services/message-topic-subscription.service.ts
```diff
@@ -0,0 +1,347 @@
...
+            { shouldBypassPermissionChecks: true },
+          );
+
+        const topics = await topicRepository.find({
+          where: {
+            id: In(subscriptions.map((subscription) => subscription.topicId)),
+          },
+        });
+
+        return topics.map((topic) => ({
```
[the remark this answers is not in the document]
FelixMalfait: Intended. A hard topic delete cascades the subscription rows (onDelete CASCADE), so the only "orphans" are subscriptions to a soft-deleted topic — and offering those on the public preferences page would let people toggle a topic that can no longer send (campaign sends require resolving the topic). The old behavior only appeared to handle this by showing the stale denormalized topicName, which is exactly what was removed per review.

packages/twenty-server/src/engine/workspace-manager/twenty-standard-application/utils/field-metadata/compute-message-suppression-standard-flat-field-metadata.util.ts
```diff
@@ -0,0 +1,355 @@
...
+  emailAddress: createStandardFieldFlatMetadata({
+    objectName,
+    workspaceId,
+    context: {
+      fieldName: 'emailAddress',
+      type: FieldMetadataType.EMAILS,
+      label: i18nLabel(msg`Email address`),
+      description: i18nLabel(msg`The suppressed email address`),
+      icon: 'IconMail',
+      isNullable: true,
```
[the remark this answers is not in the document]
FelixMalfait: The flag doesn't buy the enforcement it suggests: composite subcolumns are nullable whenever the property is isRequired: false (generate-column-definitions: `parentFlatFieldMetadata.isNullable || !compositeProperty.isRequired`), and both EMAILS properties are isRequired: false — so the columns are nullable either way, exactly like person.emails (also isNullable: true). Rows are written only by MessageSuppressionService, which normalizes and always sets the address; the field is isUIReadOnly on an isSystem object, so there's no user write path to guard.

packages/twenty-server/src/modules/emailing/services/__tests__/emailing-domain-sender.service.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+            listUnsubscribedAddresses.map((address) => address.toLowerCase()),
+          ),
+        ),
+    } as unknown as MessageTopicSubscriptionService;
+    const unsubscribeTokenService = {
+      sign: jest.fn().mockReturnValue('signed-token'),
+    } as unknown as UnsubscribeTokenService;
+    const twentyConfigService = {
+      get: jest.fn().mockReturnValue(EmailingDomainDriver.AWS_SES),
+    } as unknown as TwentyConfigService;
```
charlesBochet: code smell + remove these tests they are useless anyway (we mock too many things to make them reliable and maintainable)

PR #21179, opened 2026-06-03
Fix lambda error

martmull:
- move twenty-sdk from dependencies to devDependencies
- add documentation about breaking
- add warning about moving the package to dev dependencies

### DIFF
--- a/packages/twenty-sdk/src/cli/utilities/build/manifest/utils/validate-package-json-dependencies.ts
+++ b/packages/twenty-sdk/src/cli/utilities/build/manifest/utils/validate-package-json-dependencies.ts
@@ -1,0 +1,30 @@
+import { pathExists, readJson } from '@/cli/utilities/file/fs-utils';
+import path from 'path';
+import { isDefined } from 'twenty-shared/utils';
+
+type PackageJsonDependencies = {
+  dependencies?: Record<string, string>;
+};
+
+// twenty-sdk ships the CLI and build/scaffolding tooling used only at dev and
+// build time — it is never imported by the published app's runtime. Keeping it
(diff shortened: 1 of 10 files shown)


prastoin (commented):
Please elaborate on the motivations behind
```
move twenty-sdk from dependencies to devDependencies
```

Unless I'm mistaken from our earlier discussion you seemed to say that twenty-sdk was required at runtime by the logic function
In the end it's only used at build time for the manifest craft ?

packages/twenty-sdk/src/cli/utilities/build/manifest/utils/validate-package-json-dependencies.ts
```diff
@@ -0,0 +1,30 @@
...
+    return [];
+  }
+
+  const packageJson = await readJson<PackageJsonDependencies>(packageJsonPath);
+
+  if (isDefined(packageJson.dependencies?.['twenty-sdk'])) {
+    return [
+      '"twenty-sdk" is listed under "dependencies" in package.json. It is a build-time only tool and should be moved to "devDependencies".',
+    ];
+  }
```
prastoin, line 27: this should be documented in next packages release notes

prastoin (commented):
In what would this be resolving the application settings issue ?

martmull:
@prastoin all the required twenty sdk runtime dependencies are bundled. So we don’t need twenty sdk in production.

prastoin:
> @prastoin all the required twenty sdk runtime dependencies are bundled. So we don’t need twenty sdk in production.

From what I can remember it's getting bundled because it's in the dependencies and not the dev-dep
I'm double checking the vite config

prastoin (approved):

PR #21180, opened 2026-06-03
feat(settings): add Logs as a dedicated tab in General settings

FelixMalfait:
## What & why

The audit-log viewer lived as a full-screen page reachable only via a "View Logs" button buried in the **Security** tab. This surfaces it as the **third tab in General settings** (`General | Security | Logs`), consistent with the other tabs.

## Changes

- **Relocated** the event-logs module `pages/settings/security/event-logs/` → `modules/settings/event-logs/` and render it as tab content instead of a `FullScreenContainer` page. Dropped `SettingsPath.EventLogs`, its route, and the fullscreen handling in favor of the `general#logs` hash tab.
- **Security tab:** removed the "View Logs" entry; kept the log-retention setting there.
- **In-tab gating** (shown to users with the Security permission): Enterprise upgrade card when not entitled, a clear "ClickHouse not configured" placeholder otherwise (derived from client config), and the query is skipped when disabled. Replaces a bespoke error component that string-matched error messages with the shared `SettingsEmptyPlaceholder` / `SettingsEnterpriseFeatureGateCard`.
- **Layout:** boxed content column with the table selector + filters grouped in a `Card` and the results table below, matching settings conventions. Kept the existing fixed filters (page/event name, member, period) rather than recreating the record-view filter chips (those are tightly coupled to record/view context).

Frontend + `twenty-shared` only — no changes to the log query or data.

## Test plan

- [x] `npx nx typecheck twenty-front` and `npx nx lint twenty-front` pass
- [x] Settings → General shows three tabs; Logs is the third; breadcrumb stays "Workspace / General"
- [x] With Enterprise + ClickHouse: table selector, filters, refresh, and the paginated table work
- [x] Non-Enterprise: Enterprise upgrade card shown; no failing query fires
- [ ] Enterprise without ClickHouse: shows the "ClickHouse not configured" placeholder
- [ ] Security tab still shows the log-retention setting and the "View Logs" button is gone
- [ ] A user without the Security permission sees neither the Security nor Logs tab


packages/twenty-docker/twenty/Dockerfile
```diff
@@ -203,7 +203,24 @@ RUN tar -C / -Jxpf /tmp/s6-overlay-noarch.tar.xz \
 RUN apk add --no-cache \
     postgresql18 postgresql18-contrib \
     redis \
-    curl jq su-exec
+    curl jq su-exec gcompat
+
+# ClickHouse single static binary — backs application logs (and the other
```
FelixMalfait: Remove comment

packages/twenty-docker/twenty/Dockerfile
```diff
@@ -252,10 +269,11 @@ ENV PG_DATABASE_URL=postgres://twenty:twenty@localhost:5432/default \
     DISABLE_CRON_JOBS_REGISTRATION=true \
     IS_BILLING_ENABLED=false \
     SIGN_IN_PREFILLED=true \
-    APPLICATION_LOG_DRIVER=CONSOLE
+    CLICKHOUSE_URL=http://localhost:8123/twenty \
+    APPLICATION_LOG_DRIVER=CLICKHOUSE
```
FelixMalfait: Is this going to break the CLI? I know we have a way to tail application logs in the twenty sdk/cli right? Not sure how it could work with cloud honestly? Did it work? Are we breaking it?

packages/twenty-front/src/modules/settings/event-logs/components/EventLogFilters.tsx
```diff
@@ -33,6 +34,21 @@ const StyledFilterItem = styled.div`
...
+const StyledPeriodItem = styled.div`
+  flex: 1.5;
+  min-width: 280px;
+`;
+
+const StyledPeriodRow = styled.div`
+  display: flex;
+  gap: ${themeCssVariables.spacing[2]};
+
+  > * {
```
FelixMalfait: We don't really like this kind of pattern, can we do something more targeted?

packages/twenty-server/src/engine/core-modules/event-logs/event-logs.service.ts
```diff
@@ -171,14 +173,30 @@ export class EventLogsService {
...
+    table: EventLogTable,
+  ): Promise<void> {
     if (!this.clickHouseService.getMainClient()) {
       throw new EventLogsException(
         'Audit logs require ClickHouse to be configured. Please set the CLICKHOUSE_URL environment variable.',
         EventLogsExceptionCode.CLICKHOUSE_NOT_CONFIGURED,
       );
     }
 
+    // Application logs are available without an Enterprise license; every other
```
FelixMalfait: Remove comment

packages/twenty-docker/twenty/Dockerfile
FelixMalfait: Done — removed the ClickHouse comments in the Dockerfile.

packages/twenty-docker/twenty/Dockerfile
FelixMalfait: Checked: this does not break the CLI. dev:function:logs tails via a GraphQL subscription that publishes the execution result.logs directly, independent of APPLICATION_LOG_DRIVER; cloud stays DISABLED (the default), so it is unaffected. Keeping CLICKHOUSE so the dev-image Logs tab shows application logs. Separately, I noticed the subscription publish runs unconditionally on every execution in prod — proposing a follow-up to gate it on active subscribers (kept out of this PR).

packages/twenty-front/src/modules/settings/event-logs/components/EventLogFilters.tsx
FelixMalfait: Done — replaced the > * child selector with explicit StyledPeriodField wrappers around each date input.

packages/twenty-server/src/engine/core-modules/event-logs/event-logs.service.ts
FelixMalfait: Done — removed the comment.

packages/twenty-front/src/modules/settings/event-logs/components/SettingsLogs.tsx
```diff
@@ -0,0 +1,190 @@
...
+                onChange={handleTableChange}
+              />
+            </StyledSelectorGrow>
+            <IconButton
+              Icon={IconRefresh}
+              variant="secondary"
+              size="medium"
+              ariaLabel={t`Refresh`}
+              onClick={() => refetch()}
+            />
```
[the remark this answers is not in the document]
FelixMalfait: Fixed in 6f5bfdf: the Refresh button now only calls `refetch()` when `canQuery` is true, so it no longer triggers the skipped query (NO_ENTITLEMENT) for non-enterprise tables. (In the stacked follow-up #21197 the Refresh button is replaced by a pause/resume control on a live-by-default view, so it goes away entirely.)

packages/twenty-front/src/modules/settings/event-logs/components/SettingsLogs.tsx
```diff
@@ -0,0 +1,190 @@
...
+                value={selectedTable}
+                onChange={handleTableChange}
+              />
+            </StyledSelectorGrow>
+            <IconButton
+              Icon={IconRefresh}
+              variant="secondary"
+              size="medium"
+              ariaLabel={t`Refresh`}
+              onClick={() => refetch()}
```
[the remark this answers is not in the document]
FelixMalfait: Fixed in 6f5bfdf — `refetch()` is now guarded on `canQuery`, so Refresh no longer hits the skipped query for non-enterprise log tables.

packages/twenty-shared/src/types/SettingsPath.ts
```diff
@@ -67,8 +67,8 @@ export enum SettingsPath {
   WebhookDetail = 'api-webhooks/webhooks/:webhookId',
   Integrations = 'integrations',
   Security = 'general#security',
+  Logs = 'general#logs',
```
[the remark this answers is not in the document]
FelixMalfait: `security/event-logs` was a hidden modal route nested under the Security tab (not a primary/bookmarkable destination) — this PR consolidates it into the dedicated Logs tab. Skipping a redirect for that transient internal route; happy to add one if there is evidence of external deep links.

.github/workflows/ci-test-docker-compose.yaml
```diff
@@ -27,11 +27,12 @@ jobs:
...
-      - name: Login to Docker Hub
-        uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3.7.0
-        with:
-          username: ${{ vars.DOCKERHUB_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_PASSWORD }}
+      # Pull base images through Google's Docker Hub mirror — avoids Docker Hub
+      # rate limits and needs no credentials (this repo is public).
+      - name: Configure Docker Hub mirror
+        run: |
+          echo '{"registry-mirrors":["https://mirror.gcr.io"]}' | sudo tee /etc/docker/daemon.json
```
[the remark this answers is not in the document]
FelixMalfait: `registry-mirrors` routes all docker.io pulls through Google`s `mirror.gcr.io` pull-through cache (far higher limits), so pulls no longer depend on authenticated Docker Hub. This repo is public, so a Docker Hub token cannot be stored here anyway — the mirror is the intended replacement for the removed credentials.

PR #21186, opened 2026-06-03
Twenty new ui README

bosiraphael:
Generated a twenty new ui implementation plan README after the discussion with @charlesBochet


Weiko (commented):
Interesting! Letting @charlesBochet reviewing since I don't have all the context

charlesBochet (approved):

PR #21187, opened 2026-06-03
Fix axios related dependabot alerts generated against root yarn.lock

mabdullahabaid:
Fixes the following Dependabot alerts: https://github.com/twentyhq/twenty/security/dependabot?q=is%3Aopen+package%3Aaxios+manifest%3Ayarn.lock+has%3Apatch

Upgraded the referenced version in root yarn.lock. Creating a separate PR for the nested ones to keep the updates isolated (e.g. /seed-dependencies/yarn.lock).


packages/twenty-server/src/utils/image.ts
```diff
@@ -52,7 +52,7 @@ export const getImageBufferFromUrl = async (
 
     const contentType = response.headers['content-type'];
 
-    if (contentType && !contentType.startsWith('image/')) {
+    if (typeof contentType === 'string' && !contentType.startsWith('image/')) {
```
mabdullahabaid: Axios 1.16 widened the TypeScript type of header values. Indexing response.headers[...] used to resolve to string, but now resolves to the union string | number | true | string[] | AxiosHeaders. TypeScript won't let us call .startsWith() on that union (only string has it), so npx nx typecheck twenty-server fails with TS2339.

The `typeof contentType === 'string'` guard just narrows the union back to string so the compiler is satisfied

FelixMalfait: I think we have utils for that instead, can you check? I know there are many typeof already because AI loves it but we should prefer the utils

FelixMalfait (approved):
Made a small comment

packages/twenty-server/src/utils/image.ts
mabdullahabaid: Replaced with `isNonEmptyString`

PR #21188, opened 2026-06-03
perf(ai) - per-tool schema resolution in resolveSchemas

etiennejouan:
learn_tools(["find_many_companies"]) goes from ~170ms → ~2ms (85x faster, measured with 40 objects / empty fields — real workspaces would see even bigger savings). Previously it generated schemas for all ~250 tools and discarded 249; 

now it generates exactly the requested one(s).

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/tool-provider/providers/database-tool.provider.ts
+++ b/packages/twenty-server/src/engine/core-modules/tool-provider/providers/database-tool.provider.ts
@@ -9,8 +9,8 @@ import { canObjectBeManagedByAutomation } from 'twenty-shared/workflow';
 import { z } from 'zod';
 
 import { type GenerateDescriptorOptions } from 'src/engine/core-modules/tool-provider/interfaces/generate-descriptor-options.type';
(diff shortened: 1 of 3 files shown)


packages/twenty-server/src/engine/core-modules/tool-provider/providers/database-tool.provider.ts
```diff
@@ -293,6 +307,23 @@ export class DatabaseToolProvider implements ToolProvider {
     return descriptors;
   }
 
+  private hasMatchingTool(
+    toolNames: Set<string>,
+    snakeSingular: string,
+    snakePlural: string,
+  ): boolean {
+    return (
+      toolNames.has(`find_${snakePlural}`) ||
```
Weiko, line 316: @etiennejouan didn't we rename this into find_many?

Weiko (approved):
Nice! @etiennejouan 👏

packages/twenty-server/src/engine/core-modules/tool-provider/providers/database-tool.provider.ts
etiennejouan: Yes ! but still not merged.. can't wait to solve all coming conflicts !

PR #21193, opened 2026-06-03
fix(metadata): resolve junction targets order-independently during mgration

neo773:
A junction relation points at a target field on the join object that another action may create later (two junctions into the same join reference each other). The builder validator now also looks up the target in the to be created set, and the runner mints every field id up front so the target resolves regardless of action order, the same way relation pairs are already handled.


packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/validators/utils/validate-junction-target-settings.util.ts
```diff
@@ -58,10 +60,17 @@ export const validateJunctionTargetSettings = ({
...
+    (isDefined(remainingFlatFieldMetadataMaps)
+      ? findFlatEntityByUniversalIdentifier({
+          universalIdentifier: junctionTargetFieldUniversalIdentifier,
+          flatEntityMaps: remainingFlatFieldMetadataMaps,
+        })
+      : undefined) ??
+    findFlatEntityByUniversalIdentifier({
+      universalIdentifier: junctionTargetFieldUniversalIdentifier,
+      flatEntityMaps: flatFieldMetadataMaps,
+    });
```
prastoin, line 73: good

packages/twenty-server/src/engine/workspace-manager/workspace-migration/workspace-migration-runner/action-handlers/field/services/utils/build-field-id-to-be-created-in-migration-map.util.ts
```diff
@@ -0,0 +1,40 @@
...
+      )
+    ) {
+      fieldIdByUniversalIdentifier.set(
+        action.relatedUniversalFlatFieldMetadata.universalIdentifier,
+        action.relatedFieldId ?? v4(),
+      );
+    }
+  }
+
+  return fieldIdByUniversalIdentifier;
```
prastoin: I feel like this is duplicated, this should already exists somewhere

packages/twenty-server/src/engine/workspace-manager/workspace-migration/workspace-migration-runner/types/workspace-migration-action-runner-args.type.ts
```diff
@@ -15,6 +15,10 @@ export type WorkspaceMigrationActionRunnerArgs<
   allFlatEntityMaps: AllFlatEntityMaps;
   workspaceId: string;
   flatApplication: FlatApplication;
+  allFieldIdToBeCreatedInMigrationByUniversalIdentifierMap?: Map<
```
prastoin: seems too late to me, should be builder responsability

prastoin (commented):
Hey there, I got your point but
Do you have any existing github issue / sentry / sonarly related issue I can rely on ?

packages/twenty-server/src/engine/workspace-manager/workspace-migration/workspace-migration-runner/types/workspace-migration-action-runner-args.type.ts
prastoin: I feel like we should integrate in existing `enrichCreateWorkspaceMigrationActionsWithIds`

neo773:
@prastoin 
Hi, it's a prerequisite for PR https://github.com/twentyhq/twenty/pull/21173 it introduces several workspace entities that use Junction

packages/twenty-server/src/engine/workspace-manager/workspace-migration/workspace-migration-runner/types/workspace-migration-action-runner-args.type.ts
neo773: @prastoin 
Pushed a new change and tested it end to end with the AWS SES PR
Could you take a look when you get some time? Thanks

packages/twenty-server/src/engine/workspace-manager/workspace-migration/workspace-migration-runner/action-handlers/field/services/create-field-action-handler.service.ts
```diff
@@ -47,24 +46,21 @@ export class CreateFieldActionHandlerService extends WorkspaceMigrationRunnerAct
...
-        action.relatedFieldId ?? v4(),
+    const {
+      action,
+      allFlatEntityMaps,
+      allFieldIdToBeCreatedInMigrationByUniversalIdentifierMap,
+    } = context;
+
+    if (!isDefined(allFieldIdToBeCreatedInMigrationByUniversalIdentifierMap)) {
+      throw new Error(
+        'Missing field id map for create field action transpilation',
```
prastoin: why making the id required ?

neo773: addressed in new commit

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/validators/utils/validate-junction-target-settings.util.ts
```diff
@@ -31,6 +32,7 @@ const createError = (
 export const validateJunctionTargetSettings = ({
   universalFlatFieldMetadata,
   flatFieldMetadataMaps,
+  remainingFlatFieldMetadataMaps,
```
FelixMalfait, line 35: nit: I'm not sure I understand what this means "remaining"? I did get the more explicit "allFieldIdToBeCreatedxxx" for example

FelixMalfait (approved):

FelixMalfait:
As discussed let's maybe try to convert taskTarget/noteTarget so that it's tested? If we do this let's create an issue for a future migration command but let's not do the migration command for legacy objects here since it wasn't our goal

PR #21198, opened 2026-06-03
feat(i18n): add Serbian Latin (sr-Latn) locale

svezda:
## What

Adds **Serbian Latin (`sr-Latn`)** as a new locale, alongside the existing Serbian Cyrillic (`sr-Cyrl`).

Serbia uses both scripts officially and interchangeably, and a large share of users prefer the Latin script. Because Serbian Cyrillic→Latin is a deterministic, lossless 1:1 transliteration (the standard *gajica/vukovica* mapping), the new catalog is generated directly from the existing human-translated `sr-Cyrl` strings — so `sr-Latn` ships fully translated from day one rather than as empty stubs.

## How

**Translations** — `sr-Latn.po` + compiled `generated/sr-Latn.ts` for `twenty-front`, `twenty-emails` and `twenty-server`, transliterated 1:1 from their `sr-Cyrl` counterparts (ђ→đ, ж→ž, љ→lj, њ→nj, ћ→ć, ч→č, џ→dž, ш→š). Only `msgstr` values change — English `msgid`s, ICU placeholders (`{name}`) and markup (`<0>…</0>`) are preserved byte-for-byte.

**Registration**
- `APP_LOCALES` — add `sr-Latn` (single source of truth; every package's `lingui.config` derives its locale list from `Object.values(APP_LOCALES)`)
- `getDateFnsLocale` — map to `date-fns/locale/sr-Latn`
- `LocalePicker` — add the **Serbian (Latin)** option
- emails `i18n.utils.ts` & server `i18n.service.ts` — register the catalog in the `Record<keyof typeof APP_LOCALES, Messages>` maps

## Validation

Performed locally on the generated files:
- **0** residual Cyrillic code points in any `sr-Latn` file
- compiled `.ts` catalogs parse and have **key sets identical** to `sr-Cyrl` (front 3247, emails 43, server 1538)
- `msgid` sets are **byte-identical** between the Cyrillic and Latin `.po` files (only `msgstr` + the 3 language-header lines differ)

## Notes

This is a mechanical transliteration of the existing human-reviewed Cyrillic catalog, intended as a high-quality starting point. Once `sr-Latn` is enabled on the Crowdin project, future updates can flow through the normal translation-sync workflow.


Weiko:
Hi @svezda, 
Translations are handled asynchronously via Crowdin.
If you want to contribute you need to do that on https://twenty.crowdin.com/ otherwise those changes would be overwritten by Crowdin.

PR #21209, opened 2026-06-04
fix(front): settings skeleton, app-detail header & empty favorites

FelixMalfait:
Three small post-redesign UI fixes. Each is an independent commit, so they can be split into separate PRs if preferred.

## 1. Settings loading skeleton — match the rounded-card layout

The redesign (#21131) moved settings chrome into a rounded card (`SettingsPageLayout`: bordered header with breadcrumb + centered title, optional secondary bar, 760px body), but `SettingsSkeletonLoader` still rendered the old flat `PageHeader` + `PageBody` — so pages painted as a full-width flat bar then snapped into the card.

- `SettingsSkeletonLoader` now reproduces the card and **reuses the real `SettingsPageHeader` + `SettingsPageContainer`**, so the frame aligns by construction; the card CSS is replicated (not `SettingsPageLayout`) to avoid the layout's side effects (hotkeys, side panel, info banner).
- It's **composed with `SettingsSectionSkeletonLoader`** so the loading body is identical whether or not chrome is present. Rule: no chrome on screen yet → full-page skeleton; chrome already on screen → body-only `SettingsSectionSkeletonLoader` (the admin Enterprise tab now uses it, matching its sibling tabs). A short comment on each component documents this.

## 2. Application detail header — pass a plain title

`SettingsApplicationDetails` / `SettingsAvailableApplicationDetails` passed a custom `SettingsApplicationDetailTitle` (avatar + name + multi-line description, fixed width) into `SettingsPageLayout`'s **centered single-line title slot**, which broke the header. They now pass the app's display name like every other page. The available-app "unlisted" notice moves into the body as a reusable `InlineBanner`; the now-unused `SettingsApplicationDetailTitle` is removed.

## 3. Navigation — hide Favorites when empty

Always rendering the Favorites section (#21087) left a stray "Favorites" title above Workspace for users with no favorites. It now renders only when at least one favorite exists (redundant per-child guards dropped). Note: the "+ add favorite" entry point therefore appears once you have ≥1 favorite; the first favorite is created from a record/view as before.

## Verification
- `nx typecheck twenty-front` ✅ · `oxlint` + `oxfmt --check` on changed files ✅
- i18n catalogs intentionally untouched — handled by the repo's separate i18n pipeline.


packages/twenty-front/src/modules/settings/components/SettingsSkeletonLoader.tsx
```diff
@@ -1,54 +1,83 @@
...
+import { SettingsSectionSkeletonLoader } from '@/settings/components/SettingsSectionSkeletonLoader';
+import { SettingsPageHeader } from '@/settings/components/layout/SettingsPageHeader';
+import { useIsMobile } from '@/ui/utilities/responsive/hooks/useIsMobile';
 import { styled } from '@linaria/react';
 import { useContext } from 'react';
 import Skeleton, { SkeletonTheme } from 'react-loading-skeleton';
 import { ThemeContext, themeCssVariables } from 'twenty-ui/theme-constants';
 
-const StyledContainer = styled.div`
+// Full-page settings loading state: the rounded-card chrome (header + body)
```
FelixMalfait: Remove this comment

FelixMalfait: Done in 6fb5a1568c. I kept the short rule note on `SettingsSectionSkeletonLoader` (the body-only counterpart) so the when-to-use-which is documented in one place — happy to drop that too if you'd rather have none.

FelixMalfait: Done — removed all comments in 406713c72a (including the note on `SettingsSectionSkeletonLoader`).

PR #21214, opened 2026-06-04
Set 200 code for post requests

martmull:
as title, nestJS use to set 201 for post requests but some services (like google) requests 200 response code

See https://discord.com/channels/1130383047699738754/1511054250971758642/1511785364027609099 for context

### DIFF
--- a/packages/twenty-server/src/engine/metadata-modules/route-trigger/route-trigger.controller.ts
+++ b/packages/twenty-server/src/engine/metadata-modules/route-trigger/route-trigger.controller.ts
@@ -2,6 +2,8 @@ import {
   Controller,
   Delete,
   Get,
+  HttpCode,
+  HttpStatus,
   Patch,
   Post,
   Put,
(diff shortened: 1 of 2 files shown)


packages/twenty-server/src/engine/metadata-modules/route-trigger/route-trigger.controller.ts
```diff
@@ -25,6 +27,7 @@ export class RouteTriggerController {
   constructor(private readonly routeTriggerService: RouteTriggerService) {}
 
   @Get('*path')
+  @HttpCode(HttpStatus.OK)
```
Weiko, line 30: I'm surprised we have to do that everywhere and not only for POST, I thought 201 would be for creation only.
I'm also concerned about the fact that it won't be iso with the rest of the REST api, right? 🤔

Weiko (approved):
LGTM

packages/twenty-server/src/engine/metadata-modules/route-trigger/route-trigger.controller.ts
martmull: we don't have to, i just did it for clarity and consistency

martmull: i think it will be iso isn't it?

martmull: https://docs.nestjs.com/controllers
image

PR #21215, opened 2026-06-04
feat(ci): simplify visual regression dispatch to twenty-ui only

charlesBochet:
## Summary

- Only trigger visual regression on `CI UI` workflow (drop `CI Front`)
- Remove tarball re-packaging step — `ci-privileged` now downloads the artifact directly via GitHub API
- Remove `mode`/`project` parameters from the dispatch payload (hardcoded to twenty-ui in ci-privileged)
- Pass `run_id` of the triggering CI UI workflow so ci-privileged can fetch the correct artifact

## Context

Part of the fast visual regression CI initiative. The `ci-privileged` workflow has been simplified to only handle `twenty-ui` screenshots uploaded directly to Argos.

## Test plan

- [x] Full E2E verified on production: screenshots → Argos build → diff results → PR comment

### DIFF
--- a/.github/workflows/visual-regression-dispatch.yaml
+++ b/.github/workflows/visual-regression-dispatch.yaml
@@ -1,12 +1,12 @@
 name: Visual Regression Dispatch
 
-# Uses workflow_run to dispatch visual regression to ci-privileged.
-# This runs in the context of the base repo (not the fork), so it has
-# access to secrets — making it work for external contributor PRs.
+# Dispatches visual regression processing to ci-privileged after CI completes.
+# Runs in the context of the base repo (not the fork) so it has access to secrets,
+# making it work for external contributor PRs.
 
 on:
   workflow_run:
-    workflows: ['CI Front', 'CI UI']
+    workflows: ['CI UI']
     types: [completed]
 
 permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/visual-regression-dispatch.yaml
```diff
@@ -107,45 +104,23 @@ jobs:
...
             -f event_type=visual-regression \
             -f "client_payload[pr_number]=$PR_NUMBER" \
-            -f "client_payload[run_id]=$RUN_ID" \
+            -f "client_payload[run_id]=$WORKFLOW_RUN_ID" \
             -f "client_payload[repo]=$REPOSITORY" \
             -f "client_payload[project]=$PROJECT" \
             -f "client_payload[branch]=$BRANCH" \
-            -f "client_payload[commit]=$COMMIT" \
-            -f "client_payload[mode]=$MODE"
+            -f "client_payload[commit]=$COMMIT"
```
prastoin, line 126: The leas, the better IMO

prastoin: I usually prefer recomputing the metadata of the other side after only passing PR number

prastoin (approved):

.github/workflows/visual-regression-dispatch.yaml
charlesBochet: I think it can be useful to run on main too so commit seems more accurate

Weiko (approved):

PR #21216, opened 2026-06-04
Add more control on http trigger

### LINKED ISSUE #2472: Add more control on HTTP trigger
Ability for dev to set
- code response
- content-type
- maybe some other response caracteristics

martmull:
add "new Response" utils to define response code or content type of http route triggered logic function responses

follow up of https://github.com/twentyhq/twenty/pull/21214

### DIFF
--- a/packages/twenty-server/src/engine/metadata-modules/route-trigger/route-trigger.controller.ts
+++ b/packages/twenty-server/src/engine/metadata-modules/route-trigger/route-trigger.controller.ts
@@ -2,23 +2,26 @@ import {
   Controller,
   Delete,
   Get,
-  HttpCode,
-  HttpStatus,
   Patch,
   Post,
   Put,
   Req,
+  Res,
   UseFilters,
   UseGuards,
 } from '@nestjs/common';
 
-import { Request } from 'express';
+import { Request, Response } from 'express';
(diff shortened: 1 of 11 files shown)


packages/twenty-server/src/engine/metadata-modules/route-trigger/route-trigger.controller.ts
```diff
@@ -27,47 +30,94 @@ export class RouteTriggerController {
...
+  }
+
+  private sendResponse(
+    response: Response,
+    { statusCode, headers, body }: RouteTriggerResponse,
+  ) {
+    response.status(statusCode);
+
+    for (const [key, value] of Object.entries(headers)) {
+      response.setHeader(key, value);
```
Weiko: I feel like an allowlist would be safer and we should avoid copying the whole header (e.g CORS related headers, Cookies, XSS related issues... Content-Type is probably fine but would need an allowlist in itself as well 🤔)

martmull: ok, claude suggest this also, but i tough leaving this free would improve the devX

Weiko (approved):

PR #21233, opened 2026-06-04
fix(ai) - optim

etiennejouan:
1. tool-registry.service.ts, Pass precomputed catalog to resolveSchemas()

resolveSchemas() now accepts an optional precomputedCatalog parameter. Both getToolsByName() and getToolInfo() pass the catalog they already fetched, eliminating a redundant getCatalog() rebuild inside resolveSchemas().

2. database-tool.provider.ts, Skip field lookup when schemas=false

When building the catalog index (includeSchemas=false), getFlatFieldsFromFlatObjectMetadata() is no longer called for each of the 25 objects. The hasGroupByToolInputSchema() check is also skipped, group_by tools are always included in the index, with the real eligibility check deferred to learn_tools time.

--> 100/150ms gain on learn/execute_tool execution


packages/twenty-server/src/engine/core-modules/tool-provider/services/tool-registry.service.ts
```diff
@@ -50,16 +47,15 @@ export class ToolRegistryService {
     return results.flat();
   }
 
-  // On-demand schema generation for specific tools
   async resolveSchemas(
     toolNames: string[],
     context: ToolProviderContext,
+    precomputedCatalog?: ToolIndexEntry[],
```
prastoin: optional never consumed ?

prastoin (approved):

packages/twenty-server/src/engine/core-modules/tool-provider/services/tool-registry.service.ts
prastoin: record's arg

Weiko (approved):

PR #21234, opened 2026-06-04
Fix `getApplicationSubAllFlatEntityMaps` to prune unrelated appIds universal identifier aggregators

prastoin:
# Introduction
Atm when computing the `fromAllFlatEntityMaps` we're retrieving all the applicationIds related metadata entities to build a flat entity maps scoped to them ( atm always the applicationId + twenty-standard application id ) only inter app dependency we manage for the moment

A flat entity contains universal identifier aggregator to its related entities

The issue was that the `getApplicationSubAllFlatEntityMaps` wasn't pruning the aggregator by app

Now added a new process phase after the initial one that will check that all the aggregators contains universal identifiers that has been retrieve from the appId + appId standard intersection

## TDD test
Created a very human readable ( that's a joke ) test


thomtrp (approved):

prastoin:
FYI @martmull

packages/twenty-server/src/engine/metadata-modules/flat-entity/utils/prune-dangling-foreign-key-aggregators-in-all-flat-entity-maps-through-mutation.util.ts
```diff
@@ -0,0 +1,121 @@
...
+          childUniversalIdentifiers: new Set(
+            Object.keys(
+              looseAllFlatEntityMaps[
+                getMetadataFlatEntityMapsKey(relation.metadataName)
+              ].byUniversalIdentifier,
+            ),
+          ),
+        }));
+
+      if (aggregatorsWithChildIdentifiers.length === 0) {
```
Weiko, line 99: nit: could resolve the parent map first and `continue` early when it's empty, so the Set construction only runs when there's actually something to process?

Weiko (approved):
LGTM + small nit

PR #21235, opened 2026-06-04
fix(ai) - optimize metadata CRUD tools

etiennejouan:
Reduces output tokens for all 13 metadata tools by (~49%) based on production sampling data.

GET tools (field + object metadata)

System fields are now returned as compact {id, name, type} instead of the full ~20-key payload (opt-in includeFullSystemFields to get full payload). System objects are similarly compacted to {id, nameSingular, namePlural}.
Internal fields the agent never uses (searchVector, deletedAt, position, updatedBy) are excluded entirely.
workspaceId and applicationId are hoisted into a response envelope instead of being repeated on every record.
Null/default-false properties are stripped from custom field and object payloads (e.g. options: null, settings: null, isUIReadOnly: false).

CUD tools (create/update/delete)

Create and update field tools now return {id, name, type, label} instead of the full DTO.
Create and update object tools now return {id, nameSingular, labelSingular} instead of the full DTO.
Delete tools return {id, success: true} instead of the full DTO of the deleted entity.
Validation errors are grouped by message — e.g. 10 fields failing the same check produce one line with all names instead of 10 identical lines.

Learn schemas (all tools)

UUID pattern regex stripped from JSON schemas (keeps format: "uuid").
$schema and additionalProperties: false stripped from all generated schemas.
All Zod .describe() annotations and tool descriptions shortened.
Skill & tool description updates:

All references to the removed list_object_metadata_items tool replaced with get_object_metadata / get_field_metadata across skill instructions, dashboard tools, view filter/sort tools, and MCP server instructions.


packages/twenty-server/src/engine/metadata-modules/field-metadata/tools/field-metadata-tools.factory.ts
```diff
@@ -4,162 +4,133 @@ import { type ToolSet } from 'ai';
...
 
+import { compactMetadataOutput } from 'src/engine/core-modules/tool-provider/utils/compact-metadata-output.util';
 import { formatValidationErrors } from 'src/engine/core-modules/tool-provider/utils/format-validation-errors.util';
 import { FieldMetadataService } from 'src/engine/metadata-modules/field-metadata/services/field-metadata.service';
-import { fromFlatFieldMetadataToFieldMetadataDto } from 'src/engine/metadata-modules/flat-field-metadata/utils/from-flat-field-metadata-to-field-metadata-dto.util';
 import { WorkspaceMigrationBuilderException } from 'src/engine/workspace-manager/workspace-migration/exceptions/workspace-migration-builder-exception';
 
+const EXCLUDED_FIELD_NAMES = new Set([
+  'searchVector',
+  'deletedAt',
```
ijreilly: I think I would move deletedAt to FIELD_STRIP_WHEN_NULLISH instead

packages/twenty-server/src/engine/metadata-modules/field-metadata/tools/field-metadata-tools.factory.ts
```diff
@@ -170,14 +141,15 @@ export class FieldMetadataToolsFactory {
     return {
       get_field_metadata: {
         description:
-          'Find fields metadata. Retrieve information about the fields of objects in the workspace data model.',
+          'Retrieve field metadata. Returns an array of fields. System fields are compact {id, name, type} by default; set includeFullSystemFields=true for full payload. Internal fields (searchVector, deletedAt, position, updatedBy) are excluded.',
```
ijreilly: nit but I would rename to get_fields_metadata as I think get_field_metadata introduces ambiguity about whether we are receiving an object or an array

packages/twenty-server/src/engine/api/mcp/utils/build-mcp-server-instructions.util.ts
```diff
@@ -26,6 +26,8 @@ export const buildMcpServerInstructions = (
     `  ACTION:           http_request | send_email | draft_email | navigate_app | code_interpreter | search_help_center`,
     `  WORKFLOW:         create_complete_workflow | create/update/delete_workflow_version_step | activate/deactivate_workflow_version`,
     `  METADATA:         get/create/update/delete_object_metadata | get/create/update/delete_field_metadata`,
+    `                     Both GET tools return system items as compact summaries by default; set includeFullSystemObjects / includeFullSystemFields=true for full payload`,
+    `                     Both wrap results in {workspaceId, applicationId, ...}`,
```
ijreilly: I'm also unsure what "wraps in { ... }" mean

ijreilly (approved):
overall looks good ! 

What I think would help reviewing this type or PR is to have one detailed example in the description, like the same question asked twice to the agent, before and after the changes, with the detail of the tools they call + the answer of the tools + the token count

packages/twenty-server/src/engine/api/mcp/utils/build-mcp-server-instructions.util.ts
etiennejouan: good catch, logic removed about this output shape, don't remove it in prompt here

PR #21240, opened 2026-06-04
2467 app feedback from an app developer

### LINKED ISSUE #2467: App feedback from an app developer
Image

https://discord.com/channels/1130383047699738754/1509539714443509920

martmull:
# Improve local app-development DX: visible diffs, diagnosable
  errors, and resilient sync

  ## Summary

  Addresses feedback from an app developer building on the SDK: the
  local
  iteration loop felt heavy and opaque, and when local metadata
  drifted, the only
  way back to a clean state was often a full local reset. This PR
  makes dev sync
  **legible** (you can see what changed and what failed by name),
  **safe under
  concurrency** (parallel syncs no longer interleave and corrupt
  metadata), and
  **self-healing** (a sync that re-declares an already-existing entity
  reconciles
  instead of failing).

  > Context: an app developer reported that errors like
  `ENTITY_ALREADY_EXISTS` on
  > a pageLayout identifier and `Migration action 'create' for
  'fieldMetadata'
  > failed` forced uninstall/reinstall cycles or a full
  `docker:reset`, and that
  > it was hard to tell whether a failure came from their code, the
  manifest, the
  > metadata diff/sync, or local instance state.

  ## What's changed

  ### 1. Surface the metadata diff in dev sync output (`twenty-sdk`)
  The server already returned the applied `actions` from
  `syncApplication`, but the
  CLI discarded them and printed a bare `✓ Synced`. It now renders the
  diff:

  Metadata changes: 2 created, 1 updated, 1 deleted
    created objectMetadata rocket
    created fieldMetadata timelineActivities
    updated fieldMetadata
    deleted pageLayout
  ✓ Synced

  - New pure util `formatSyncActionsSummary` (grouped counts +
  per-action detail,
    capped with an `…and N more` line, ignores malformed entries).
  - Wired into the sync orchestrator step.

  ### 2. Name the failing entity in migration errors (`twenty-server`)
  `WorkspaceMigrationRunnerException` (EXECUTION_FAILED) now includes
  the failing
  entity's `universalIdentifier`:

  > `Migration action 'create' for 'fieldMetadata'
  (universalIdentifier: 2020…) failed`

  Resolved defensively (`create` → `flatEntity?.universalIdentifier`,
  `update`/`delete` → `universalIdentifier`); the clause is omitted
  when absent so
  the error path never throws.

  ### 3. Serialize dev sync per workspace with a cache lock
  (`twenty-server`)
  Concurrent `syncApplication` calls against the same workspace (e.g.
  multiple
  agents iterating on an app) could interleave their metadata
  migrations and read a
  stale workspace flat-entity cache, leaving metadata partially
  applied. The
  manifest sync is now wrapped in a per-workspace cache lock
  (`app-sync:<workspaceId>`), mirroring the lock already used by the
  application
  install path. The rate-limit throttle stays outside the lock.

  ### 4. Reconcile already-existing entities during manifest sync
  (`twenty-server`)
  When the workspace already contains an entity that the manifest
  re-declares but
  that's missing from the application's metadata slice (stale cache or
  drifted
  ownership from a previous partial sync), the builder planned a
  `create` that
  collided with the existing entity and failed the whole sync. We now
  fold those
  existing entities back into the `from` maps before building the
  diff, so the
  builder treats them as updates/no-ops instead of creates. Scoped to
  the
  application manifest sync path; the shared metadata engine keeps its
  strict
  already-exists guard for UI/boot paths. The slice is re-pruned after
  folding to
  keep one-to-many relation references consistent.

  ### 5. Logic functions sync in LIVE execution mode
  `fromLogicFunctionManifestToUniversalFlatLogicFunction` now sets
  `executionMode: LIVE` (was `PREBUILT`) so synced logic functions run
  in live
  execution mode during development.

  ### 6. More actionable manifest build warnings (`twenty-sdk`)
  Dropped the noisy generic "No object/logic function/front component
  defined"
  warnings in favor of a targeted warning when an agent has no
  `responseFormat`
  defined.

  ## Testing

  already-exists guard for UI/boot paths. The slice is re-pruned after folding to
  keep one-to-many relation references consistent.

  ### 5. Logic functions sync in LIVE execution mode
  `fromLogicFunctionManifestToUniversalFlatLogicFunction` now sets
  `executionMode: LIVE` (was `PREBUILT`) so synced logic functions run in live
  execution mode during development.

  ### 6. More actionable manifest build warnings (`twenty-sdk`)
  Dropped the noisy generic "No object/logic function/front component defined"
  warnings in favor of a targeted warning when an agent has no `responseFormat`
  defined.

  ## Testing

  - **Unit (added):**
    - `format-sync-actions-summary` (5), `sync-application-orchestrator-step` (2)
    - `application-development.resolver` lock behavior (2)
    - `workspace-migration-runner.exception` message (2)
    - `reconcile-existing-entities-into-application-from` (3)
    - `manifest-validate` warnings (updated)
  - All implemented test-first (red → green observed).
  - `twenty-server` typecheck clean; oxlint (type-aware) + oxfmt clean on changed files.
  - Existing `format-upgrade-error-for-storage` snapshots unchanged.

  ## Follow-ups / not in scope

  - **Integration coverage for the reconcile (#4 above is unit-tested only):** a
    DB-level test that seeds a drifted workspace (entity under a different app id)
    and asserts `syncApplication` reconciles instead of throwing — recommended
    before relying on this for relation-heavy manifests.
  - A `dev --dry-run` that previews the diff before applying (the server already
    returns the actions needed to build it).
  - Docs: a "which command when" guide + recovery ladder
    (re-sync → uninstall/reinstall → reset as last resort).

  ★ Insight ─────────────────────────────────────
  - The three server-side fixes form a deliberate defense-in-depth ladder: #3 (lock) prevents the drift at the cause, #4 (reconcile) heals drift that already exists, and #2 (named errors)
  makes whatever still fails diagnosable. Framing them this way in the PR helps reviewers see they're complementary, not redundant.
  - I called your two commits out explicitly (sections 5–6) rather than folding them into mine, so the reviewer/author attribution stays honest and the "live execution mode" + "warnings"
  changes don't look like side effects of the DX work.
  ─────────────────────────────────────────────────


packages/twenty-server/src/engine/core-modules/application/application-manifest/converters/from-logic-function-manifest-to-universal-flat-logic-function.util.ts
```diff
@@ -40,7 +40,7 @@ export const fromLogicFunctionManifestToUniversalFlatLogicFunction = ({
     workflowActionTriggerSettings:
       logicFunctionManifest.workflowActionTriggerSettings ?? null,
     isBuildUpToDate: true,
-    executionMode: LogicFunctionExecutionMode.PREBUILT,
+    executionMode: LogicFunctionExecutionMode.LIVE,
```
martmull, line 43: @charlesBochet can you validate this change please?

prastoin: From my understanding should depends on if production app installation or development app sync

charlesBochet:
@martmull could you please split this into multiple PRs? also PR desc is hard to read!

martmull:
Split into smaller, focused PRs for easier review:
- #21249 — surface metadata diff + name failing migration actions
- #21250 — serialize dev sync per workspace (cache lock)
- #21251 — dry-run preview for dev sync

These three are stacked (#21249 ← #21250 ← #21251). The remaining work from this branch (sync error hints, flatEntity labels, dev-mode summary UI, docs) stays on `2467-app-feedback-from-an-app-developer` and can be opened as follow-up PRs.

PR #21249, opened 2026-06-05
feat(app-dev): surface metadata diff in dev sync and name failing migration actions

martmull:
Split out of #21240.

- Render the applied metadata changes (created/updated/deleted + identifiers) in the dev sync output instead of a bare `✓ Synced`.
- Include the failing entity's `universalIdentifier` in `WorkspaceMigrationRunnerException` messages so conflicts are diagnosable.

image

image

image

PR #21250, opened 2026-06-05
fix(app-dev): serialize dev sync per workspace with a cache lock

martmull:
Split out of #21240. Stacked on #21249 (review/merge that first).

Concurrent `syncApplication` calls on the same workspace could interleave their metadata migrations and leave metadata partially applied. Wrap the manifest sync in a per-workspace cache lock (`app-sync:<workspaceId>`), mirroring the install path. The rate-limit throttle stays outside the lock.


charlesBochet (approved):
LGTM

PR #21253, opened 2026-06-05
fix: use canonical oxlint rule id in lint-disable directives

FelixMalfait:
## What

Many `oxlint-disable` / `eslint-disable` directives across the repo carry a corrupted rule id — `@typescripttypescript/<rule>` — most likely a find-and-replace accident that mangled the eslint-era `@typescript-eslint/` prefix.

oxlint matches disable directives **loosely by rule name**, so these still suppress in practice (not a silent no-op), but the id is malformed and misleading.

## Change

Replace them with the **canonical oxlint id** `typescript/<rule>` — matching the plugin name and rule keys declared in `.oxlintrc.json` — **127 files, 262 directives**:

| rule | count |
| --- | ----- |
| `typescript/no-explicit-any` | 250 |
| `typescript/ban-ts-comment` | 6 |
| `typescript/no-misused-promises` | 4 |
| `typescript/no-empty-object-type` | 2 |

- `twenty-server`: 122 files
- `twenty-front`: 5 files

Comment-only — no code or runtime changes.

## Verification

`oxlint --type-aware -c .oxlintrc.json` reports **0 warnings / 0 errors** for both `twenty-server` and `twenty-front`. Every changed line is exactly the id correction inside a disable directive (262 insertions / 262 deletions, no collateral edits).

> Addresses the cubic review, which flagged that the canonical oxlint id is `typescript/...` (no `@`). Worth noting the original `@typescripttypescript/` was not actually a silent no-op — oxlint matches these directives loosely by rule name — but `typescript/` is the correct, config-aligned id.

### DIFF
--- a/packages/twenty-front/src/loading/components/__stories__/Loading.stories.tsx
+++ b/packages/twenty-front/src/loading/components/__stories__/Loading.stories.tsx
@@ -30,7 +30,7 @@ export default meta;
 export type Story = StoryObj<typeof RecordIndexPage>;
 
 export const Default: Story = {
-  // oxlint-disable-next-line @typescripttypescript/ban-ts-comment
+  // oxlint-disable-next-line typescript/ban-ts-comment
   // @ts-ignore
   decorators: [LoadingDecorator, PageDecorator],
   play: async ({ canvasElement }) => {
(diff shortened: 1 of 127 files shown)


Weiko (approved):
LGTM

FelixMalfait:
Thanks @cubic-dev-ai — updated all the directives to the canonical `typescript/<rule>` id (no `@`), matching the plugin name and rule keys in `.oxlintrc.json`. Force-pushed in `0f8d6a6833`.

One clarification for the record: the original `@typescripttypescript/` prefix was **not** actually a silent no-op. oxlint matches `*-disable` directives loosely by rule name, so every prefix (including `@typescript-eslint/`, `@typescript/`, and even a bogus one) suppresses `no-explicit-any` — I verified this with a standalone oxlint 1.53.0 repro. So this is a correctness/canonical-id cleanup rather than a behavioral fix, but `typescript/` is the right id and matches the config.

PR #21254, opened 2026-06-05
Add People Data Labs enrichment logic functions

bosiraphael:
## Add People Data Labs enrichment logic functions

Implements the **enrichment logic functions** that call the PDL API and map responses onto standard + `pdl*` fields.

- **Logic functions** — shared core powers bulk workflow actions (`enrich-person` / `enrich-company`) and single-record AI tools. Each guards a TTL, matches via identifier, fills standard fields only when empty, always writes `pdl*` fields, and normalizes SELECT values against field options.
- **Data model** — PDL employer now links via the **standard `company` relation** (find-or-create, fill-only-if-empty); removed the dedicated `pdlCurrentCompany` relation, six `pdlJobCompany*` scalars, and the `pdl*` indexes.

**Not functional yet:** the post-install workflow seeding (`seed-enrichment-workflow.ts`) is left in place but blocked — it needs workflow-builder mutations that the CRUD-only app SDK schema doesn't expose.

**Deferred:** metering/billing and auto-enrichment triggers.

https://github.com/user-attachments/assets/a570a484-ae79-4c57-a274-e068fdab78e4

https://github.com/user-attachments/assets/6edfb5e1-cada-4892-87e0-b485eb359692


Weiko (approved):
Why removing indexes for pdlLastEnrichedAt? @bosiraphael 
LGTM

bosiraphael:
@Weiko The index is no longer needed after a change in strategy

PR #21255, opened 2026-06-05
[draft] messaging: add integration tests

neo773:


charlesBochet:
As discussed together:
- let's use the API as much as possible
- let's not use sync driver as this is not real execution code path

Closing, let's re-open on next push for a new review!

PR #21256, opened 2026-06-05
fix: guard against undefined logicFunctionId when destroying workflow CODE steps

thomtrp:
## Summary
- Adds `isDefined` guard in `handleLogicFunctionSubEntities` to skip CODE steps with undefined `logicFunctionId` instead of crashing
- Adds same guard in `runWorkflowVersionStepDeletionSideEffects` for consistency
- Rejects CODE steps in `create_complete_workflow` AI tool at runtime to prevent creating workflows with missing logic functions in the first place

Fixes `"Logic function with id undefined not found"` INTERNAL_SERVER_ERROR when destroying workflows whose CODE steps were created via `create_complete_workflow` without a proper logic function.

## Test plan
- [x] Destroy a workflow that has a CODE step with undefined logicFunctionId → should succeed silently
- [x] Try creating a workflow with a CODE step via `create_complete_workflow` tool → should return error message
- [x] Normal workflow destroy with valid CODE steps still deletes the logic function

### DIFF
--- a/packages/twenty-server/src/modules/workflow/workflow-tools/tools/create-complete-workflow.tool.ts
+++ b/packages/twenty-server/src/modules/workflow/workflow-tools/tools/create-complete-workflow.tool.ts
@@ -116,6 +116,15 @@ This is the most efficient way for AI to create workflows as it handles all the 
     activate?: boolean;
   }) => {
     try {
+      const codeSteps = parameters.steps.filter(
+        (step) => step.type === ('CODE' as string),
+      );
+
+      if (codeSteps.length > 0) {
+        throw new Error(
+          'CODE steps cannot be created via create_complete_workflow because it does not create the underlying logic function. Use create_workflow_version_step instead.',
+        );
+      }
       const workflowId = await createWorkflow({
         deps,
         context,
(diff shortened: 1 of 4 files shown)


packages/twenty-server/src/modules/workflow/workflow-tools/tools/create-complete-workflow.tool.ts
```diff
@@ -116,6 +116,15 @@ This is the most efficient way for AI to create workflows as it handles all the
     activate?: boolean;
   }) => {
     try {
+      const codeSteps = parameters.steps.filter(
+        (step) => step.type === ('CODE' as string),
+      );
+
+      if (codeSteps.length > 0) {
+        throw new Error(
```
Weiko: Should we throw custom workflow errors? This will throw an INTERNAL with no context @thomtrp

Weiko (approved):
LGTM

packages/twenty-server/src/modules/workflow/workflow-tools/tools/create-complete-workflow.tool.ts
thomtrp: sure 👍

PR #21269, opened 2026-06-05
fix: allow password reset when user is already authenticated

### LINKED ISSUE #21248: You do not have permission to perform this action while trying to reset password
A user with member access can't login, and tries to reset their password. You keep on receiving **You do not have permission to perform this action.** when clicking on the password reset link. Tried both admin and user roles as well as impersonating a user.

Pantkartik:
## Summary

Fixes #21248 — "You do not have permission to perform this action" when clicking a password reset link while already logged in.

## Root Cause

When a logged-in user follows a password reset link, their browser sends the existing `Authorization: Bearer <access_token>` header with every GraphQL request.

The `GraphQLHydrateRequestFromTokenMiddleware` runs **before** any resolver guards and calls `validateTokenByRequest`. If the token validation throws an `AuthException` (e.g. `FORBIDDEN_EXCEPTION` due to workspace mismatch, stale/invalid token, etc.), the middleware:
1. Catches the error
2. **Writes the error response directly** (`writeGraphqlResponseOnExceptionCaught`)
3. **Returns without calling `next()`** — the request never reaches the resolver

This means public endpoints like `validatePasswordResetToken` and `updatePasswordViaResetToken` — which are decorated with `@UseGuards(PublicEndpointGuard, NoPermissionGuard)` — returned `"You do not have permission to perform this action"` even though they should be freely accessible.

## Fix

In `hydrateGraphqlRequest` inside `middleware.service.ts`, catch `AuthException` errors during token validation and **continue gracefully** (only setting the locale), rather than propagating the error upward to the middleware's error handler.

Authentication requirements are enforced by the **guards declared on each resolver** (`WorkspaceAuthGuard`, `UserAuthGuard`, etc.), not by the hydration middleware. Non-`AuthException` errors (e.g. unexpected infrastructure failures) are still re-thrown.

### `packages/twenty-server/src/engine/middlewares/middleware.service.ts`
- Wrapped `validateTokenByRequest` call in `hydrateGraphqlRequest` in a `try/catch`
- On `AuthException`: logs a warning, sets the locale from headers, and returns (allowing the request to proceed to guards)
- On other exceptions: re-throws as before

## Reproduction Steps

1. Log in as any workspace member
2. Trigger a password reset email (or use an existing password reset link)
3. Click the reset link **while still logged in** in the same browser
4. **Before fix**: received "You do not have permission to perform this action"
5. **After fix**: the password reset form loads and works correctly


charlesBochet:
Hi @Pantkartik, closing this one, I don't think the original issue is about that + approach looks wrong + CI is red

Pantkartik:
Fine i might would have closed this yess about the approach ,will surely look into it thanks @charlesBochet

PR #21270, opened 2026-06-05
fix: surface proper errors for People create/delete constraint violations

### LINKED ISSUE #21119: Can't create or delete People records
```
{
    "data": {
        "createPerson": null
    },
    "errors": [
        {
            "message": "Data validation error.",
            "extensions": {
                "code": "INTERNAL_SERVER_ERROR"
            }
        }
    ]
}
```

https://github.com/twentyhq/private-issues/issues/492

Pantkartik:
## Summary

Fixes #21119 — `createPerson` / `deletePerson` mutations fail with a generic `INTERNAL_SERVER_ERROR: "Data validation error."` instead of a meaningful error, blocking core CRM record management.

## Root Cause

The `computeTwentyORMException` function in `twenty-orm` contains a catch-all block that matches **every known Postgres error code** via `Object.values(POSTGRESQL_ERROR_CODES).includes(errorCode)` and discards all error detail, throwing:

```ts
throw new PostgresException('Data validation error.', errorCode);
```

This `PostgresException` is then converted by the GraphQL error handler into `INTERNAL_SERVER_ERROR`, masking the real constraint violation. Common mutations like `createPerson` that hit:
- **`NOT_NULL_VIOLATION` (23502)** — a required field is missing
- **`FOREIGN_KEY_VIOLATION` (23503)** — referenced record missing or deletion blocked by a FK
- **`RESTRICT_VIOLATION` (23001)** — record deletion blocked by a referencing row

...all silently surface as the same opaque `"Data validation error."` with `INTERNAL_SERVER_ERROR`.

Already handled correctly before the catch-all:
- `UNIQUE_VIOLATION` → delegates to `handleDuplicateKeyError` ✅
- `INVALID_TEXT_REPRESENTATION` → `TwentyORMException(INVALID_INPUT)` ✅
- Query read timeout → `TwentyORMException(QUERY_READ_TIMEOUT)` ✅

## Fix

Add explicit handling **before** the catch-all for the four most common data-integrity constraint errors, converting them to `TwentyORMException(INVALID_INPUT)` with a clear user-facing message. The GraphQL error handler then returns `BAD_USER_INPUT` (400) instead of `INTERNAL_SERVER_ERROR` (500).

### `packages/twenty-server/src/engine/twenty-orm/error-handling/compute-twenty-orm-exception.ts`

Added specific handling for:
| Postgres Code | Constant | User-facing message |
|---|---|---|
| `23502` | `NOT_NULL_VIOLATION` | "A required field is missing. Please provide all required values and try again." |
| `23503` | `FOREIGN_KEY_VIOLATION` | "This operation references a record that does not exist or cannot be modified due to existing relationships." |
| `23001` | `RESTRICT_VIOLATION` | "This record cannot be deleted because it is still referenced by other records." |

## Before / After

**Before:**
```json
{
  "data": { "createPerson": null },
  "errors": [{
    "message": "Data validation error.",
    "extensions": { "code": "INTERNAL_SERVER_ERROR" }
  }]
}
```

**After (e.g. NOT_NULL_VIOLATION):**
```json
{
  "data": { "createPerson": null },
  "errors": [{
    "message": "A required field is missing. Please provide all required values and try again.",
    "extensions": { "code": "BAD_USER_INPUT" }
  }]
}
```

### DIFF
--- a/packages/twenty-server/src/engine/twenty-orm/error-handling/compute-twenty-orm-exception.ts
+++ b/packages/twenty-server/src/engine/twenty-orm/error-handling/compute-twenty-orm-exception.ts
@@ -23,7 +23,6 @@ const CONSTRAINT_VIOLATION_MESSAGES: Record<string, MessageDescriptor> = {
   [POSTGRESQL_ERROR_CODES.NOT_NULL_VIOLATION]: msg`A required field is missing. Please provide all required values and try again.`,
   [POSTGRESQL_ERROR_CODES.FOREIGN_KEY_VIOLATION]: msg`This operation references a record that does not exist or cannot be modified due to existing relationships.`,
   [POSTGRESQL_ERROR_CODES.RESTRICT_VIOLATION]: msg`This record cannot be deleted because it is still referenced by other records.`,
-  [POSTGRESQL_ERROR_CODES.CHECK_VIOLATION]: msg`One or more field values are invalid. Please check your input and try again.`,
 };
 
 export const computeTwentyORMException = async (
--- a/packages/twenty-server/src/engine/twenty-orm/error-handling/compute-twenty-orm-exception.ts
+++ b/packages/twenty-server/src/engine/twenty-orm/error-handling/compute-twenty-orm-exception.ts
@@ -1,3 +1,4 @@
+import { type MessageDescriptor } from '@lingui/core';
 import { msg } from '@lingui/core/macro';
 import { isDefined } from 'twenty-shared/utils';
 import { QueryFailedError } from 'typeorm';
@@ -17,6 +18,13 @@ import {
 interface QueryFailedErrorWithCode extends QueryFailedError {
   code?: string;
 }
+
+const CONSTRAINT_VIOLATION_MESSAGES: Record<string, MessageDescriptor> = {
+  [POSTGRESQL_ERROR_CODES.NOT_NULL_VIOLATION]: msg`A required field is missing. Please provide all required values and try again.`,
+  [POSTGRESQL_ERROR_CODES.FOREIGN_KEY_VIOLATION]: msg`This operation references a record that does not exist or cannot be modified due to existing relationships.`,
+  [POSTGRESQL_ERROR_CODES.RESTRICT_VIOLATION]: msg`This record cannot be deleted because it is still referenced by other records.`,
+  [POSTGRESQL_ERROR_CODES.CHECK_VIOLATION]: msg`One or more field values are invalid. Please check your input and try again.`,
+};
 
 export const computeTwentyORMException = async (
   error: Error,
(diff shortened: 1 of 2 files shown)


prastoin (commented):
Hello there ! Thanks for your contribution
What do you think @etiennejouan ?
Can't remember if we had something similar here ?

prastoin (dismissed):
Haven't see @charlesBochet pushes ok for me

Pantkartik:
Could you please merge the changes it's a long while since not entertained @prastoin

packages/twenty-server/src/engine/twenty-orm/error-handling/compute-twenty-orm-exception.ts
```diff
@@ -18,6 +19,13 @@ interface QueryFailedErrorWithCode extends QueryFailedError {
   code?: string;
 }
 
+const CONSTRAINT_VIOLATION_MESSAGES: Record<string, MessageDescriptor> = {
+  [POSTGRESQL_ERROR_CODES.NOT_NULL_VIOLATION]: msg`A required field is missing. Please provide all required values and try again.`,
+  [POSTGRESQL_ERROR_CODES.FOREIGN_KEY_VIOLATION]: msg`This operation references a record that does not exist or cannot be modified due to existing relationships.`,
+  [POSTGRESQL_ERROR_CODES.RESTRICT_VIOLATION]: msg`This record cannot be deleted because it is still referenced by other records.`,
+  [POSTGRESQL_ERROR_CODES.CHECK_VIOLATION]: msg`One or more field values are invalid. Please check your input and try again.`,
```
etiennejouan: I think we prefer 500 for this one because invalid value should be catch earlier. then it would help us fix valdiation layer

etiennejouan:
> Hello there ! Thanks for your contribution What do you think @etiennejouan ? Can't remember if we had something similar here ?

As explain (and clearly ! 👏 ) in PR description, we did it for specific errors (the ones, postgres only is the most efficient layer to check/throw). Concerning not null, wonder if we should not intercept it earlier, same for validation. Nitpick.

And last nitpick, wonder if error messages are clear enough. I mean, it's better than current ones. But is it helping enough to troubleshoot ?

Pantkartik:
@etiennejouan Thanks for the feedback. I agree that NOT_NULL_VIOLATION and CHECK_VIOLATION could ideally be intercepted earlier in the validation layer. My intention was to improve the user-facing error when these database-level constraints are reached. I'm happy to adjust the handling if you'd prefer keeping some of these cases as INTERNAL_SERVER_ERROR to help identify validation gaps.

Pantkartik:
@charlesBochet @etiennejouan @prastoin could you please review and merge the pr

etiennejouan (commented):
Still thinking CHECK_VIOLATION error should not throw 400 as it should be catch earlier in validation layer. Then I would prefer current 500, having sentry bug and fixing them than silenting them.
Concerning other ones an the whole PR, looks good to me.

packages/twenty-server/src/engine/twenty-orm/error-handling/compute-twenty-orm-exception.ts
etiennejouan: @claude remove this CHECK_VIOLATION key please

etiennejouan (dismissed):

etiennejouan (approved):

PR #21280, opened 2026-06-06
feat(front): delete selected records with Delete/Backspace in table view

### LINKED ISSUE #21275: 'DEL' Keyboard Shortcut for Deleting Selected Records
## Scope & Context
To improve productivity and data management speed, we want to allow users to delete records directly using their keyboard when managing table views, mirroring standard OS and spreadsheet behaviors.

## Current behavior
When users select one or multiple records in a table view using the checkboxes, they must use the mouse to click the "Delete" button that appears in the bulk action bar. Pressing the `DEL` (or `Backspace` on Mac) key on the keyboard does nothing.

## Expected behavior
When one or multiple rows are selected in a table view, pressing the `DEL` key (or `Backspace` on macOS) should instantly trigger the deletion process for the selected records.
To prevent accidental deletions, pressing the key should open the standard deletion confirmation modal (maybe this should be a setting to activate the confirmation modal) or immediately soft-delete them while displaying an "Undo" toast notification.

## Technical inputs
- Add a keyboard event listener (handling `Delete` and `Backspace` keys) at the table/view component level.
- Ensure the listener only triggers when records are actively selected in the table and the user is not currently typing inside an input field or search bar.
- Hook the key press event into the existing bulk delete handler function used by the UI's bulk action bar.
- Ensure the focus management remains logical after the rows are removed (e.g., focus shifts back to the table container).

abhinavapasok:
## What / Why

Closes #21275.

When one or more rows are selected in a record table, pressing **Delete** (or **Backspace** on macOS) now deletes the selected records — mirroring standard spreadsheet behavior. Previously deletion required clicking the **Delete** button in the bulk action bar.

## How

Follows the existing `RecordTableBodyEscapeHotkeyEffect` pattern:

- **`useDeleteSelectedRecords`** (new hook, `record-table/hooks/`) — reads the current selection via `selectedRowIdsComponentSelector`, is gated on `objectPermissions.canSoftDeleteObjectRecords`, resets the row selection and deletes via the existing `useDeleteManyRecords`.
- **`RecordTableBodyDeleteHotkeyEffect`** (new) — binds `Delete` / `Backspace` through `useHotkeysOnFocusedElement`, scoped to `PageFocusId.RecordIndex`. Because it is scoped to the record-index focus, it does **not** fire while a cell editor or search input is focused, so Backspace still edits text there. Guarded by `isAtLeastOneTableRowSelectedSelector` so it is a no-op with an empty selection.
- Registered in `RecordTableBodyEffectsWrapper` next to the Escape effect, so it applies to both grouped and non-grouped tables.

`preventDefault` is enabled so Backspace doesn't trigger browser navigation.

## Behavior notes

- Matches the current bulk Delete button (soft delete). A confirmation modal / opt-in setting — floated in the issue to guard against accidental deletes — would be a natural follow-up but is intentionally left out here to keep the change minimal and consistent with existing behavior.
- Respects object permissions: does nothing when the user lacks soft-delete permission.

## Testing

- `nx lint:diff-with-main twenty-front` — passes (oxlint + oxfmt)
- `nx typecheck twenty-front` — passes


charlesBochet:
Thanks a lot for this @abhinavapasok

After discussing it, we don't want to ship instant Delete/Backspace deletion without a confirmation step. The risk of accidental data loss is too high here, especially with Backspace which people hit reflexively, and we haven't settled on the right product behavior yet (confirmation modal vs an opt-in setting vs an undo toast).

Since that's a product decision we still need to make, we're going to close this for now rather than leave it hanging. This isn't a reflection on your work, and we'd be glad to revisit once we've decided on the UX. Really appreciate the contribution.

PR #21287, opened 2026-06-06
fix(front): reject backslash paths in isValidReturnToPath (open-redirect hardening)

joeltco:
## Summary

`isValidReturnToPath` validates the post-login `returnTo` path and already rejects protocol-relative `//` paths — but not the backslash variant. Browsers normalize `\` to `/`, so `/\evil.com` resolves like `//evil.com` (a protocol-relative, external URL) while still passing the existing `//` check:

```ts
isValidReturnToPath("/\\evil.com"); // returns true today; should be false
```

This hardens the open-redirect guard by rejecting any path containing a backslash, so a `returnTo` can only ever be a same-site absolute path.

## Changes
- `isValidReturnToPath`: reject paths containing `\`.
- Added tests for backslash-tricked paths.

Framed as defense-in-depth — the validator should reject this class regardless of how each consumer performs the redirect.

### DIFF
--- a/packages/twenty-front/src/modules/auth/utils/isValidReturnToPath.ts
+++ b/packages/twenty-front/src/modules/auth/utils/isValidReturnToPath.ts
@@ -16,15 +16,7 @@ export const isValidReturnToPath = (path: string): boolean => {
     return false;
   }
 
-  if (!path.startsWith('/') || path.startsWith('//')) {
-    return false;
-  }
-
-  // Browsers normalize "\" to "/", so "/\evil.com" resolves like the
-  // protocol-relative "//evil.com" (an external URL) even though it passes the
-  // "//" check above. Reject backslashes so a returnTo can only ever be a
-  // same-site absolute path.
-  if (path.includes('\\')) {
+  if (!path.startsWith('/') || path.startsWith('//') || path.includes('\\')) {
     return false;
   }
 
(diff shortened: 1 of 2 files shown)


charlesBochet (approved):
Thank you!

packages/twenty-front/src/modules/auth/utils/isValidReturnToPath.ts
```diff
@@ -16,7 +16,7 @@ export const isValidReturnToPath = (path: string): boolean => {
     return false;
   }
 
-  if (!path.startsWith('/') || path.startsWith('//')) {
```
prastoin, line 19: @claude shouldn't we use the `isSafeRelativePath` here ?

joeltco: @prastoin good question — I looked into reusing it, but `isSafeRelativePath` isn't a fit here:

- It lives in `twenty-server` and imports Node's `path` (`isAbsolute`/`normalize`/`sep`), so it isn't usable from `twenty-front` browser code.
- More importantly the semantics are opposite: it rejects absolute paths via `isAbsolute()`, but a valid `returnTo` *must* start with `/` — so `isSafeRelativePath('/settings')` returns `false` and would reject every legitimate redirect target.
- The two solve different problems: that util guards filesystem traversal (`..`, null bytes) for uploads, whereas this guards open-redirect (`//`, `\`, and the excluded onboarding prefixes). Only the backslash check overlaps.

If we want to avoid duplicating the `//` / `\` logic, I'd be happy to extract a small `isSafeRedirectPath` helper into `twenty-shared` — but pointing this at the file-storage util would reintroduce the redirect cases it doesn't cover. Happy to do that follow-up if you'd like.

PR #21288, opened 2026-06-06
Fix email composer

neo773:
Re-adds #21177 which we had to revert in #21284.

The original PR broke workflow send/draft email steps because workflow actions never passed any user identity to the connected account lookup, so user-visibility accounts always came back as not found.

Fixed by resolving the acting user in workflow runs (run creator for manual runs, workflow creator for webhook/cron) and feeding that into the existing access check.

Also threw a clearer error when the account exists but belongs to someone else instead of pretending it doesn't exist.

Ships with a 2.11 upgrade command that reassigns workflow creators to the email account owner for existing workflows where they diverge (API/system created workflows have no creator to act as), otherwise those would keep failing the access check even with this fix.

### DIFF
--- a/packages/twenty-client-sdk/src/metadata/generated/schema.graphql
+++ b/packages/twenty-client-sdk/src/metadata/generated/schema.graphql
@@ -3007,8 +3007,8 @@ type Query {
   ): IndexConnection!
   findManyAgents: [Agent!]!
   findOneAgent(input: AgentIdInput!): Agent!
-  myConnectedAccounts: [ConnectedAccountPublicDTO!]!
   getRoles: [Role!]!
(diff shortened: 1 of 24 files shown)


FelixMalfait:
For me the workflow escalation problem is a big issue that needs to be tackled with a global picture. The conclusion I had come to is that for each workflow we should assign a role that is created on the fly (1 role = 1 workflow), when someone edits this workflow then the attached role is updated, that way he can see the workflow but he cannot edit an existing workflow to escalate permissions. This also solves the issue when a user is offboarded, his workflows don't stop working.

neo773:
Closing this PR to reduce noise in that case.
That would be the longterm fix and needs to be thought of and addressed in a separate PR.

PR #21289, opened 2026-06-06
Add Ota readiness contract and matrix coverage

bobaikato:
## Summary
- add an `ota.yaml` contract that models Twenty's proven readiness lanes: `verify`, `native-dev`, `integration`, and `docker-selfhost`
- add a cross-OS Ota matrix that validates the contract, exercises native and container dry-runs, and runs representative proof lanes where the contract claims them
- add a short contributor-facing Ota path to the local setup docs
- keep aggregate task sequencing explicit with `after_success` follow-ups instead of hiding it in shell `&&` chaining

## Why
Twenty is a serious monorepo with different operational surfaces for CI verification, local contributor development, service-backed integration, and Docker self-hosting.

This PR makes those lanes explicit in one readiness contract instead of leaving them scattered across scripts, env setup, and contributor knowledge. It keeps the distinctions between those paths intact rather than flattening them into one generic app story.

## Scope notes
- the contract intentionally keeps runtime surfaces and workflow paths separate
- host-platform truth stays explicit instead of being hand-waved away
- the matrix only proves the lanes this contract actually claims
- aggregate checks use structured success follow-ups where that sequencing is part of the task graph

## Validation
- green matrix run: https://github.com/bobaikato/twenty/actions/runs/27097038819


FelixMalfait:
Thank you! We are not looking to add this for now

PR #21292, opened 2026-06-07
fix: wrong record count on deleted and normal records

### LINKED ISSUE #11977: Wrong number of records in deleted record details

parshipcy:
## Summary
- Resolves #11977 
- When looking into the deleted records from People tab (or any object list), the record detail header showing 0/(total records) instead of the correct position among deleted records only, e.g. 1/3 or 3/7. So, this PR makes the count match what users see in the deleted-records list.
- Also normal records showing `0/N` in the header when opened from a list view (e.g. `0/48` -> `2/48`).

## Approach
I tried to keep the change small and avoid extra server requests:
- when a user came from a deleted-records view, we tell our existing queries to include soft-deleted records.
- for the position number, we use the record list the user already had open (from the index view they came from) instead of apollo cache, which didn’t include records, especially deleted ones, but also normal records.
- normal list behavior is not changed on the server side.

## Test plan
- Open people/company, delete a record
- Use the side menu -> “see deleted records”
- open a deleted record’s details
- confirm the header showing the correct position and total (e.g. 1/2, not 0/100)
- for normal list: open People (normal list, not deleted) -> click a record -> open full page -> confirm header shows correct position and total (e.g. `2/48`, not `0/48`)

### Before:
Screenshot 2026-06-07 135204
Screenshot 2026-06-07 141735

### After:
Screenshot 2026-06-07 134946
Screenshot 2026-06-07 135045
Screenshot 2026-06-07 141642


parshipcy:
fix for deleted records: https://github.com/twentyhq/twenty/pull/21292/changes/58f64bedf61e6bbd968c416335d4956dbda3da76
fix for normal records: https://github.com/twentyhq/twenty/pull/21292/changes/7e7f2ddf67fc2d8cf1d8f8c5d46f78cf3bd05e2b

charlesBochet (dismissed):
Thank you! I've cleaned a bit more

charlesBochet (approved):

PR #21293, opened 2026-06-07
fix(front): keep relation field record tables scoped to the host record

rcshetty3:
## Problem

When a relation field is added to a record page as a record **table** widget
(Page Layouts → a `FIELD` widget with `fieldDisplayMode: TABLE` and a `viewId`),
the table renders the **global** list of the related object instead of only the
records related to the current record.

Steps to reproduce:
1. On a Company record page layout, add a to-many relation field (e.g.
   `Opportunities`) as a widget and set its display mode to **Table** with a view
   (so it shows columns).
2. Open a Company record.
3. The Opportunities table lists *all* opportunities in the workspace, not just
   the ones linked to that company.

Note: when the same relation widget has **no** `viewId`, it is correctly scoped
to the record — but then it can't render custom columns. So custom columns and
relation-scoping were effectively mutually exclusive.

## Root cause

`FieldWidgetRelationTable` renders the related records through
`RecordTableWidgetRendererContent` using the widget's `viewId`. That path loads
the view's filters and fetches the related object's records, but **never applies
the relation filter** that constrains the table to the host record. With a
`viewId` present, the table therefore shows the whole object.

The relation filter itself already exists elsewhere —
`RecordDetailRelationSection` builds
``{ `${inverseRelationFieldName}Id`: { in: [recordId] } }`` for its aggregate.
It just isn't applied on the table path.

## Fix

- Add a pure helper `getRelationTableFilter()` that builds the host-relation
  filter for a to-many relation field (morph-aware, mirroring
  `RecordDetailRelationSection`).
- `FieldWidgetRelationTable` computes this filter and passes it down via the
  existing `RecordFilterValueDependenciesContext` (new optional
  `relationTableFilter`).
- `useFindManyRecordIndexTableParams` (rows) and
  `useAggregateRecordsForRecordTableColumnFooter` (footer aggregates) AND this
  filter into their queries.

The filter is scoped to the relation-table instance through the context and
defaults to `undefined`, so **every other table (record index, kanban,
dashboards, …) is unaffected** — `combineFilters` / object spread treat the
absent filter as a no-op. No backend changes.

## Tests

- New unit tests for `getRelationTableFilter` (to-many → foreign-key filter;
  to-one → none; unresolved relation type / field → none; morph relation;
  missing morph target names → none).
- `nx typecheck twenty-front`, `nx lint twenty-front`, and the new
  `nx test twenty-front` suite pass locally.

## Screenshots

Same record (a "Centre" with 0 related theory allocations and 34 related
orders), same page-layout (relation fields shown as Table widgets with a view).

**Before** — with a `viewId`, the relation tables show the *global* lists: the
Theory Allocations table is full of allocations belonging to *other* records,
and Collateral Orders shows 60 (the whole object's first page) instead of 34.

**After** — the same tables are scoped to the record: Theory Allocations is
empty (this record has none) and Collateral Orders shows exactly its 34 orders,
with the view's columns (Status / Total Value / Date).

## Verification

Verified on a self-hosted instance running the equivalent change (the four
touched files are byte-identical on `main` and the latest release tag): a
relation table widget with a `viewId` now shows only the host record's related
rows **with** the view's columns, the footer aggregates match the visible rows,
and the global record index is unchanged. Confirmed across records with
different related-record counts (e.g. a record with 34 related orders shows 34;
a record with 1 shows 1; records with 0 show an empty table).


charlesBochet:
LGTM, thank you!

charlesBochet:
actually, I could not reproduce the issue, I misunderstood the behavior you were trying to fix. I'm reverting it in the PR listed above

PR #21294, opened 2026-06-07
feat: inline image thumbnails and legacy-label fallback for FILES field chips

mvanhorn:
## Summary

Custom FILES field chips now show an inline image thumbnail for image attachments and fall back to the legacy label when an attachment predates filename storage. This covers two of the UX complaints n2ojim collected in #20942: image files were indistinguishable from other attachments, and older attachments rendered with an empty chip label. The 10-file cap from the same issue already shipped in #20950; the gallery/grid layout and hover-delete affordances are deliberately left for follow-ups per the maintainer's cost notes on the thread.

## Why this matters

#20942 is founder-tagged UX feedback on the new custom FILES field: once a record carries more than a couple of attachments, users scan chips visually, and a thumbnail answers "which one is the screenshot" without opening anything. The fallback keeps old records readable instead of showing blank chips. Changes stay inside `FileChip.tsx` and follow the existing file-display patterns; Storybook stories cover both behaviors.

## Testing

Added 9 Storybook stories: image attachment (thumbnail), non-image (icon unchanged), missing filename (legacy fallback label), long names, and combinations. Targeted typecheck of the changed files surfaced no errors; the monorepo's CI lint/build covers the rest.

Refs #20942


etiennejouan (commented):
Hi @mvanhorn, thanks for your contribution and for the tests, great!

It works great but I would take this logic down into the FileIcon comp to enjoy thumbnail in all other components (AttachmentRow, files picker, ...). wdyt ?

etiennejouan (dismissed):

Weiko (approved):

mvanhorn:
Thanks @Weiko, glad the inline image thumbnails landed with the legacy-label fallback for older attachments.

PR #21299, opened 2026-06-07
fix: use correct userWorkspaceId for navigation menu comparisons

### LINKED ISSUE #21291: [Twenty-front][Bug]: NavigationMenuItem.userWorkspaceId is set to workspaceMember.id
## Bug Description
I was working on this PR #21290 and found a bug:
Right now in `ViewPickerOptionDropdown.tsx` we are calculating `isFavorite` by comparing the `userWorkspaceId` from `navigationMenuItems` with `currentWorkspaceMemberId`.
```
  const isFavorite = navigationMenuItems.find(
    (item) =>
      item.viewId === view.id &&
      item.userWorkspaceId === currentWorkspaceMemberId,
  );
```
Which will never be same because they come from two different entities. `userWorkspaceId` is an FK to `UserWorkspace.id` of `core` schema, whereas `currentWorkspaceMemberId` is the id of the `WorkspaceMember` table of the `workspace_...` schema. So the equality check is never true and isFavorite is permanently false.

## Expected behavior

`isFavorite` should be `true` when the current user has already added that view to their favorites, and `false` otherwise.

## Technical inputs

The two IDs being compared live in different tables and are not interchangeable:
- `NavigationMenuItem.userWorkspaceId`  ->  foreign key to `core.userWorkspace.id`
- `currentWorkspaceMemberId` from `useNavigationMenuItemsData()`  ->  id of the `workspaceMember` row in the workspace schema

`UserWorkspace` and `WorkspaceMember` are 1-to-1, but their `id` values are different UUIDs minted in different tables, so the `===` check will never pass.

### Other places using the same pattern:
1. **`AddToFavoritesSingleRecordCommand.tsx`** (line 44)  -  when the user adds a record to favorites from the command menu, it writes `userWorkspaceId: currentWorkspaceMemberId` into the new `NavigationMenuItem`. Same broken FK on insert.

2. **`useNavigationMenuItemEditController.ts`** (lines 63 - 65, 88, 118) - pulls `currentWorkspaceMemberId` from the same hook and aliases it as `targetUserWorkspaceId`, then writes it into `userWorkspaceId` on both `createItem` and `updateItem` paths. Affects the whole favorites/navigation-menu edit flow, not just views.

**Common source:** `useNavigationMenuItemsData.ts` (line 19) returns `currentWorkspaceMember?.id` under the name `currentWorkspaceMemberId`. Every consumer above forwards it into a `userWorkspaceId` slot.

Manibharadwaj:
## What does this PR do?

Fixes a bug where `NavigationMenuItem.userWorkspaceId` was being compared/set to `WorkspaceMember.id` instead of the correct `UserWorkspace.id`, causing the favorites functionality to not work correctly.

Fixes #21291

## Problem

The `isFavorite` check in `ViewPickerOptionDropdown` and `createManyNavigationMenuItems` calls in multiple files were using `currentWorkspaceMemberId` (which is `WorkspaceMember.id` from the `workspace_*` schema) instead of the correct `UserWorkspace.id` (from the `core` schema).

This caused:
- `isFavorite` to always return `false` for user favorites
- Navigation menu items to be created with incorrect `userWorkspaceId`

## Root Cause

In `useNavigationMenuItemsData.ts`:
- `currentWorkspaceMemberId` was derived from `currentWorkspaceMember?.id` (WorkspaceMember.id)
- But `NavigationMenuItem.userWorkspaceId` expects a `UserWorkspace.id`
- These are two different entities from different schemas (core vs workspace)

## Solution

1. Added `currentUserWorkspaceId` to the `useNavigationMenuItemsData` hook return type
2. `currentUserWorkspaceId` is derived from `currentWorkspaceMember?.userWorkspaceId` 
3. Updated all comparisons and assignments to use `currentUserWorkspaceId` when dealing with `userWorkspaceId`

## Files Changed

- `packages/twenty-front/src/modules/navigation-menu-item/display/hooks/useNavigationMenuItemsData.ts` - Added `currentUserWorkspaceId` to return type
- `packages/twenty-front/src/modules/views/view-picker/components/ViewPickerOptionDropdown.tsx` - Fixed `isFavorite` check and `createManyNavigationMenuItems` call
- `packages/twenty-front/src/modules/command-menu-item/engine-command/record/single-record/components/AddToFavoritesSingleRecordCommand.tsx` - Fixed `createManyNavigationMenuItems` call
- `packages/twenty-front/src/modules/navigation-menu-item/edit/hooks/useNavigationMenuItemEditController.ts` - Fixed `targetUserWorkspaceId` assignment

## Testing

- No existing tests directly cover the `useNavigationMenuItemsData` hook
- The fix is a simple type/field correction that should not affect other components
- CI will verify TypeScript compilation and linting

## Checklist

- [x] I have read the [CONTRIBUTING.md](https://github.com/twentyhq/twenty/blob/main/.github/CONTRIBUTING.md) file
- [x] Changes are tested locally (TypeScript compilation)
- [x] Commit message follows repository conventions
- [x] PR is linked to the relevant issue (#21291)

### DIFF
--- a/packages/twenty-front/src/modules/command-menu-item/engine-command/record/single-record/components/AddToFavoritesSingleRecordCommand.tsx
+++ b/packages/twenty-front/src/modules/command-menu-item/engine-command/record/single-record/components/AddToFavoritesSingleRecordCommand.tsx
@@ -18,7 +18,7 @@ export const AddToFavoritesSingleRecordCommand = () => {
   }
 
   const { createManyNavigationMenuItems } = useCreateManyNavigationMenuItems();
-  const { navigationMenuItems, currentWorkspaceMemberId } =
+  const { navigationMenuItems, currentUserWorkspaceId } =
     useNavigationMenuItemsData();
 
   const handleExecute = () => {
@@ -41,7 +41,7 @@ export const AddToFavoritesSingleRecordCommand = () => {
         type: NavigationMenuItemType.RECORD,
         targetRecordId: selectedRecord.id,
         targetObjectMetadataId: objectMetadataItem.id,
-        userWorkspaceId: currentWorkspaceMemberId,
+        userWorkspaceId: currentUserWorkspaceId,
         position: maxPosition + 1,
       },
     ]);
--- a/packages/twenty-front/src/modules/navigation-menu-item/display/hooks/useNavigationMenuItemsData.ts
+++ b/packages/twenty-front/src/modules/navigation-menu-item/display/hooks/useNavigationMenuItemsData.ts
@@ -42,6 +42,7 @@ export const useNavigationMenuItemsData = (): NavigationMenuItemsData => {
     navigationMenuItems: userNavigationMenuItems,
     workspaceNavigationMenuItems,
     currentWorkspaceMemberId,
-    currentUserWorkspaceId: currentWorkspaceMember?.userWorkspaceId ?? undefined,
+    currentUserWorkspaceId:
+      currentWorkspaceMember?.userWorkspaceId ?? undefined,
   };
 };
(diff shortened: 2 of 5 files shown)


etiennejouan (approved):
Hi @Manibharadwaj thanks for your contribution and congrats for your first PR 👏

PR #21301, opened 2026-06-07
fix: match relation field filters in optimistic & RLS record matchers

### LINKED ISSUE #21345: [Bug] Relation field in a view filter throws "Not implemented yet" on every record mutation
## What

Putting a **relation field** (e.g. a self-relation like People "Referred By") into a view filter as `is (not) empty` makes every create/update/delete on that object throw:

```
Not implemented yet, use UUID filter instead on the corresponding "<field>Id" field
```

Repro:
1. Add a relation field on People (e.g. self-relation "Referred By").
2. Create a view with a filter on that relation field set to "is not empty".
3. Edit any Person. The optimistic update throws and the record list breaks.

CleanShot 2026-06-08 at 06 44 01@2x

CleanShot 2026-06-08 at 06 45 34@2x

## Why

On each mutation the Apollo optimistic effect (`triggerUpdateRecordOptimisticEffect`) re-evaluates all active view filters against the changed record via `isRecordMatchingFilter`. That matcher (and its server-side twin `isRecordMatchingRLSRowLevelPermissionPredicate`) only implements the relation **join column** branch (`<field>Id`) and `throw`s for the relation field object itself (`<field>`). The UI stores the filter keyed on the relation object even when "> Id" is chosen, so it hits the unimplemented branch. Users can't use a relation in a view filter at all.

## Fix

In the `RELATION` / `MORPH_RELATION` case of both matchers, when the filter key is the relation object (not the join column), match against the related record id instead of throwing:

- `packages/twenty-front/src/modules/object-record/record-filter/utils/isRecordMatchingFilter.ts`
- `packages/twenty-server/src/engine/twenty-orm/utils/is-record-matching-rls-row-level-permission-predicate.util.ts`

```ts
return isMatchingUUIDFilter({
  uuidFilter: filterValue as UUIDFilter,
  value: record[filterKey]?.id ?? null,
});
```

## Test plan

- [ ] View filter on a relation field (`is not empty` / `is empty` / `in`) matches by related id, no throw
- [ ] Editing a record while such a filter is active succeeds
- [ ] Existing join-column (`<field>Id`) filter path still works

josephj:
Closes #21345.

## What

It should be caused by the GraphQL optimistic query.

`isRecordMatchingFilter` (front, Apollo optimistic cache) and `isRecordMatchingRLSRowLevelPermissionPredicate` (server, RLS) now handle a view filter that targets a **relation field object** (e.g. an "is (not) empty" filter on a relation) by matching against the related record id, instead of throwing.

CleanShot 2026-06-08 at 06 44 01@2x

CleanShot 2026-06-08 at 06 45 34@2x

## Why

Both matchers only implemented the relation **join column** branch (`fooId`) and threw `Not implemented yet, use UUID filter instead on the corresponding "fooId" field` for the relation field itself (`foo`). In practice the UI still stores relation filters keyed on the relation object, so any view with such a filter made every create/update/delete on that object throw: the optimistic effect re-evaluates all active view filters against the changed record and hits the unimplemented branch.

Repro: add a self-relation field on People (e.g. "Referred By"), put it in a view filter as "is not empty", then edit any Person. The optimistic update throws.

## Behaviour change

| Scenario | Before | After |
|---|---|---|
| View filter on relation object (`referredBy is not empty`), then edit a record | Throws `Not implemented yet...` | Record matched by related id; update succeeds |
| Filter on relation join column (`referredById`) | Worked | Unchanged |

## Test plan

```bash
cd packages/twenty-front && npx jest isRecordMatchingFilter
cd packages/twenty-server && npx jest is-record-matching-rls-row-level-permission-predicate
```

- [x] Front: relation `is empty` / `is not empty` / `in` match by related id; join-column path still passes (20/20)
- [x] Server: relation `is empty` / `is not empty` match by related id (9/9)
- [x] `lint:diff-with-main` clean on both packages


josephj:
It seems fixed.

packages/twenty-front/src/modules/object-record/record-filter/utils/isRecordMatchingFilter.ts
```diff
@@ -433,9 +433,10 @@ export const isRecordMatchingFilter = ({
           });
         }
 
-        throw new Error(
-          `Not implemented yet, use UUID filter instead on the corresponding "${filterKey}Id" field`,
-        );
+        return isMatchingUUIDFilter({
+          uuidFilter: filterValue as UUIDFilter,
+          value: record[filterKey]?.id ?? null,
+        });
```
josephj, line 440: Filtering on the relation field itself rather than its join column (e.g. an "is (not) empty" view filter). Match against the related record id so optimistic cache updates don't throw on a relation filter.

FelixMalfait (approved):
Thanks!

PR #21308, opened 2026-06-08
feat(twenty-front): generalize the page primary/secondary bars (flat redesign)

FelixMalfait:
Replaces #21279 and #21282 with one clean PR from `main`.

Generalizes the settings primary-bar / secondary-bar card chrome to the record index, record show and standalone pages via a shared `PageCardLayout` + `PageCardHeader` (the side panel sits as a sibling of the content card), and applies the new flat design direction: square corners on the card, side panel and loading skeletons.

Iterating toward the new design (Figma node 102282-221623); the confirmed direction and the explicit "remove rounded corners" change are in, remaining designer specifics to follow.


packages/twenty-front/src/loading/components/RightPanelSkeletonLoader.tsx
```diff
@@ -1,14 +1,27 @@
...
   flex-direction: column;
   width: 100%;
 `;
 
-export const RightPanelSkeletonLoader = () => (
-  <StyledRightPanelContainer>
-    <PageContentSkeletonLoader />
-  </StyledRightPanelContainer>
-);
+export const RightPanelSkeletonLoader = () => {
```
FelixMalfait: Not sure if we need a RightPanelSkeletonLoader, is this really used? I feel like it could be delete entirely? And if we keep it I don't see a reason to keep a distinction between right panel in settings page and outside of setting pages

FelixMalfait: Good call — removed it. `UserOrMetadataLoader` now renders the shared `PageContentSkeletonLoader` directly, so the boot skeleton no longer branches on settings-vs-non-settings and matches the lazy-route fallback that renders right after. (b1aa1d29eb)

packages/twenty-front/src/pages/page-layout/StandalonePageLayoutPage.tsx
```diff
@@ -19,27 +19,28 @@ export const StandalonePageLayoutPage = () => {
...
               targetRecordIdentifier: undefined,
               layoutType: PageLayoutType.STANDALONE_PAGE,
               isInSidePanel: false,
             }}
           >
-            <MainContainerLayoutWithSidePanel>
+            <MainContainerLayout>
               <PageLayoutRenderer pageLayoutId={pageLayoutId} />
-            </MainContainerLayoutWithSidePanel>
+            </MainContainerLayout>
```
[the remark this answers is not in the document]
FelixMalfait: Fixed in d91a0e4045 — dropped `MainContainerLayout`, which nested `PageBody` (noisy background) and `PagePanel` (its own border) inside the flat card, producing a card-within-a-card. The standalone page-layout now renders directly in a plain `overflow-y: auto` container that inherits the card's `primary` background, matching how `RecordShowPage` renders the same dashboard content. The now-orphaned `MainContainerLayout` was deleted.

PR #21309, opened 2026-06-08
security: bump vulnerable direct dependencies (axios, next, vitest, qs, dompurify, …)

charlesBochet:
## What

Within-major version bumps of **direct** dependencies to clear a large batch of Dependabot alerts that are breaching (or near) their SLA. No major-version changes — all stay within the current major, so risk is low.

| Package | From → To | Clears |
|---|---|---|
| `axios` | ^1.13.5 → ^1.16.0 | ReDoS, Proxy-Auth leak, proto-pollution gadgets, NO_PROXY bypass, resource DoS (56 alerts) |
| `next` | 16.1.7 → ^16.2.6 | DoS, middleware/proxy bypass, SSRF, cache poisoning, XSS (32 alerts) |
| `vitest` | 4.0.18 → ^4.1.0 | **CRITICAL** — UI server arbitrary file read/exec (#1421) |
| `qs` | ^6.11.2 → ^6.15.2 | `qs.stringify` DoS |
| `dompurify` | 3.3.3 → ^3.4.0 | proto-pollution XSS + FORBID_TAGS / SAFE_FOR_TEMPLATES bypasses |
| `@nestjs/core` | 11.1.16 → ^11.1.18 | improper output neutralization / injection |
| `nodemailer` | 8.0.4 → 8.0.10 | SMTP command injection via CRLF (bumped via root `resolutions`) |
| `path-to-regexp` | ^8.2.0 → ^8.4.0 | ReDoS via multiple wildcards |
| `file-type` | ^21.3.1 → ^21.3.2 | ZIP decompression-bomb DoS |
| `@opentelemetry/exporter-prometheus` | ^0.211.0 → ^0.217.0 | exporter process crash via malformed HTTP request (#1183/#1184) |

## Notes
- Added a `next` root **resolution** so the dev-only `@react-email/preview-server` copy (hard-pinned at `16.0.10`) is also pulled up to the patched `16.2.x` line — otherwise that copy keeps the Next.js alerts open.
- `@opentelemetry/exporter-prometheus` 0.217 pulled `@opentelemetry/sdk-metrics` to 2.7.1 (compatible); `@opentelemetry/api` stays pinned at 1.9.1.
- **Transitive-only** vulnerable packages (undici, tmp, ws, brace-expansion, …) are handled in a **separate PR** per the split-by-group plan.
- Breaking major bumps (electron, uuid, serialize-javascript) and migrations (Apollo Server 3→4, simplemde) are intentionally **out of scope** here.

PR #21310, opened 2026-06-08
security: bump vulnerable transitive dependencies (postcss, fast-xml-parser, react-router, brace-expansion, …)

charlesBochet:
## What

Companion to the direct-dependency security PR (#21309). Bumps **transitive** vulnerable packages via `yarn up -R` (re-resolve to newest within existing ranges). **Lockfile-only — no root `package.json` / resolutions changes.**

### Cleared
| Package | Fix |
|---|---|
| `brace-expansion` | numeric-range / zero-step DoS |
| `follow-redirects` | auth header leak to cross-domain |
| `diff` | ReDoS in parse/applyPatch |
| `@protobufjs/utf8` | overlong UTF-8 decoding |
| `lodash` | — |

## Out of scope

Vulnerable copies that are **exact-pinned by third-party parents**, so they can only be fixed once those parents ship patched versions — they're not direct deps of any workspace, can't be added per-package, and a sibling dependency can't override a parent's exact pin (only a root resolution could, which we're intentionally not adding):

- `postcss` ← `next`, `styled-components`
- `fast-xml-parser` ← `@aws-sdk/xml-builder`
- `undici`, `ws`, `tmp`, `picomatch`, `webpack-dev-server`, `ip-address`, `unhead`, `yeoman-environment`, `@tootallnate/once`, `ajv`, `react-router` (need breaking major bumps or coordinated multi-package updates; mostly dev/build tooling, not shipped runtime)

> Split from the direct-deps PR (#21309) per the agreed split-by-group plan.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -16391,14 +16391,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@protobufjs/utf8@npm:^1.1.0":
-  version: 1.1.0
-  resolution: "@protobufjs/utf8@npm:1.1.0"
-  checksum: 10c0/a3fe31fe3fa29aa3349e2e04ee13dc170cc6af7c23d92ad49e3eeaf79b9766264544d3da824dba93b7855bd6a2982fb40032ef40693da98a136d835752beb487
-  languageName: node
-  linkType: hard
-
-"@protobufjs/utf8@npm:^1.1.1":
+"@protobufjs/utf8@npm:^1.1.0, @protobufjs/utf8@npm:^1.1.1":
   version: 1.1.1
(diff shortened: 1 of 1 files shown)


charlesBochet:
CI fix: the `react-router` 6.30.4 resolution desynced `react-router` from `react-router-dom`, creating two `@remix-run/router` copies → front typecheck error (`AppRouter.tsx` TS2322). Dropped that resolution; `react-router` is back to the working single `6.30.3` and twenty-front typecheck passes locally. The open-redirect there needs a coordinated react-router + react-router-dom bump (follow-up). postcss + fast-xml-parser resolutions kept.

martmull (approved):

PR #21311, opened 2026-06-08
Fix various graphs bugs

ijreilly:
Some bugs fixed in this PR
1. From UI any field could be chosen to group the query by it, while for instance, RAW_JSON type (eg workflowRun.state) is not supported by PostgreSQL to group a query by. Fix: removed it from the "group by" fields options in FE + in BE -->
2. The BE check existed (isFlatFieldMetadataSupportedInGroupBy) but the signature was malformed: it expected`{ fieldMetadataType, fieldMetadataName, fieldMetadataIsSystem }` while every caller passes a flat field metadata object with type/name/isSystem. So the check is mis-wired — at runtime the destructured props are undefined, making it always return true (validation bypassed). Fixed this.
3. Group by does not work with Morph relations if their direction is ONE_TO_MANY. Added that constraint.
4. Group by with morph relations were broken even for MANY_TO_ONE, because a morph is stored as one field per target (polymorphicOwnerRocket, polymorphicOwnerSurveyResult…), each with its own join column, but the frontend collapsed them into a single polymorphicOwner field — so the backend tried to resolve a non-existent polymorphicOwnerId. Fix: Frontend: added a target picker so you choose the specific morph target (then its sub-field), storing the real per-target field id. Backend: fixed validate-relation-subfield to use the per-target field's own relationTargetObjectMetadataId instead of the multi-target resolver that returned null.
5. (improvement) When an error occured in the query, the graph showed "No data". Updated it to "error". (screenshot 1)
6. When a field used as a filter on a graph is deleted, it is not deleted as a graph filter (which is ok because it would involve parsing all the graph's configuration json to find whether a field is referenced; there is no foreign key), which prevented from further modifying the graph's filters. Fixed this + add an indicator that the filter is can/should be removed (see screenshot 2)
7. "Ambiguous column name" PG error occurs when ordering by "creation date" of a related field, because both objects have createdAt field. Fixed it by adding table alias as prefix.
8. (improvement) While working on #5 I did not understand why we could directly do `"objectMetadataNameSingular"."columnName" `while I expected that for custom objects it would have to be `_objectMetadataNameSingular`. that's simply because we use an alias from the beginning. To add clarity, within groupBy code I replaced `objectMetadataNameSingular` with `objectAlias` everywhere it is indeed inherited from us using objectAlias.

Screenshot 2026-06-08 at 12 01 45

Screenshot 2026-06-08 at 11 53 55


packages/twenty-shared/src/utils/fieldMetadata/isFieldMetadataSupportedInGroupBy.ts
```diff
@@ -0,0 +1,10 @@
+import { FIELD_METADATA_TYPES_NOT_SUPPORTED_IN_GROUP_BY } from '@/constants';
+import { FieldMetadataType } from '@/types';
+
+export const isFieldMetadataSupportedInGroupBy = ({
+  type,
+}: {
+  type: FieldMetadataType;
+}): boolean => {
+  return !FIELD_METADATA_TYPES_NOT_SUPPORTED_IN_GROUP_BY.has(type);
+};
```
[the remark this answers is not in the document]
ijreilly: agreed, we should use the same util between BE and FE

packages/twenty-front/src/modules/side-panel/pages/page-layout/components/dropdown-content/ChartGroupByFieldSelectionMorphRelationFieldView.tsx
```diff
@@ -0,0 +1,151 @@
...
+                  selected={false}
+                  focused={selectedItemId === target.perTargetFieldId}
+                  LeftIcon={
+                    isDefined(target.icon) ? getIcon(target.icon) : undefined
+                  }
+                  hasSubMenu
+                  onClick={() => {
+                    setSelectedTarget(target);
+                  }}
+                />
```
bosiraphael: Since selected is always false, we don't need `MenuItemSelect`
```suggestion
                 <MenuItem
                  text={target.label}
                  focused={selectedItemId === target.perTargetFieldId}
                  LeftIcon={
                    isDefined(target.icon) ? getIcon(target.icon) : undefined
                  }
                  hasSubMenu
                  onClick={() => {
                    setSelectedTarget(target);
                  }}
                />
```

packages/twenty-shared/src/utils/fieldMetadata/shouldExcludeFieldFromAgentToolSchema.ts
```diff
@@ -0,0 +1,24 @@
...
+  isSystem,
+  excludeId = true,
+  additionalExcludedFieldNames = [],
+}: {
+  fieldName: string;
+  isSystem: boolean;
+  excludeId?: boolean;
+  additionalExcludedFieldNames?: string[];
+}): boolean => {
+  const excludedFieldNames = [
```
bosiraphael, line 14: nit: Could extract those in a const file?

packages/twenty-front/src/modules/side-panel/pages/page-layout/components/dropdown-content/ChartGroupByFieldSelectionMorphRelationFieldView.tsx
```diff
@@ -0,0 +1,151 @@
...
+      )
+      .map((morphRelation) => {
+        const targetObjectMetadataItem = objectMetadataItems.find(
+          (item) =>
+            item.nameSingular ===
+            morphRelation.targetObjectMetadata.nameSingular,
+        );
+
+        return {
+          perTargetFieldId: morphRelation.sourceFieldMetadata.id,
```
bosiraphael, line 74: Morph group-by shows a blank field name in the settings panel

This saves the per-target morph field id, but the front-end only keeps one collapsed morph field in objectMetadataItem.fields, so for the other target the id can't be found and the field name renders blank (useChartSettingsValues.ts:78). The chart still works. Fix: also resolve the id via morphRelations[].sourceFieldMetadata.id.

To reproduce with the seeded data (seeded data):

- Make a bar/pie chart on the Pet object.
- Group by → Polymorphic Owner → pick the Survey Result target → pick any field (e.g. Name).
- The "Data on display" row in the settings panel is blank, even though the chart renders. (Picking the Rocket target instead shows the name correctly)

bosiraphael (approved):
LGTM

PR #21313, opened 2026-06-08
feat(twenty-partners): notify Discord on new partner application

rashad:
Adds an `on-partner-application-created` logic function triggered on the `partner.created` database event. When the website application form creates a new Partner, it posts a rich embed to a Discord channel (applicant, company, country, languages, partner scope, skills) with a deep link to the record.

## How it works
- Fires only on genuine form submissions — discriminates via `createdBy.source === 'APPLICATION'`, which excludes seed/import (`API`) and manual UI (`MANUAL`) creation.
- Runs out-of-band on the worker (database event trigger), so it adds **no latency** to the applicant's submission, and the linked Person already exists by the time it runs.
- Best-effort: a Discord failure never fails the trigger (wrapped in `try/catch`, 8s timeout).

## Configuration (per workspace — Settings → Apps → Twenty Partners → Variables)
- `DISCORD_WEBHOOK_URL` (secret) — the incoming webhook URL. **The feature is a no-op when unset.**
- `PARTNER_APP_FRONTEND_URL` — workspace front-end base URL for the record deep link (e.g. `https://partners.twenty.com`).

## Notes
- New logic function + two application variables; version bumped to **0.4.0** (minor).
- Unit tests cover the source-guard branches, the on/off switch, the embed contents/ordering, and best-effort failure handling.
- The website and the existing `submit-partner-application` handler are untouched.


packages/twenty-apps/internal/twenty-partners/src/logic-functions/on-partner-application-created.logic-function.ts
```diff
@@ -0,0 +1,182 @@
...
+          }
+        : null,
+    };
+
+    const embed = buildApplicationEmbed(partnerForEmbed, frontendUrl);
+
+    const controller = new AbortController();
+    const timeout = setTimeout(() => controller.abort(), DISCORD_TIMEOUT_MS);
+    try {
+      await fetch(webhookUrl, {
```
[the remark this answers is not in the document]
rashad: Good catch — fixed in 803ac6b726. `fetch` resolves on HTTP error statuses (it only rejects on network errors), so a dead webhook (404), rejected payload (400) or rate limit (429) was being reported as delivered. The handler now checks `response.ok`, returns `notified: false` and warns on non-2xx, and a regression test covers it.

packages/twenty-apps/internal/twenty-partners/src/logic-functions/on-partner-application-created.logic-function.ts
```diff
@@ -0,0 +1,191 @@
...
+
+// The event 'after' hydrates the actor composite as a NESTED object
+// (after.createdBy.source), verified from the live partner.created payload.
+type PartnerCreatedAfter = {
+  id: string;
+  createdBy?: { source?: string | null } | null;
+};
+
+export const handler = async (
+  payload: DatabaseEventPayload,
```
martmull:
```suggestion
import { CoreSchema } from 'twenty-client-sdk/core';
...
  payload: DatabaseEventPayload<ObjectRecordCreateEvent<CoreSchema.Partner>>,
```

Partner typing should be importable from `twenty-client-sdk/generated`

packages/twenty-apps/internal/twenty-partners/src/logic-functions/on-partner-application-created.logic-function.ts
```diff
@@ -0,0 +1,191 @@
...
+): Promise<Record<string, unknown>> => {
+  const after = (payload?.properties as { after?: PartnerCreatedAfter } | undefined)?.after;
```
martmull: with payload `DatabaseEventPayload<ObjectRecordCreateEvent<Partner>>,` you won't need this

packages/twenty-apps/internal/twenty-partners/src/application-config.ts
```diff
@@ -17,5 +17,17 @@ export default defineApplication({
         'Shared secret required in the X-Application-Secret header on POST /partner-applications. Must match the website route\'s PARTNER_APPLICATION_SECRET env var. Set per-workspace in Settings → Apps → Twenty Partners → Variables.',
       isSecret: true,
     },
+    DISCORD_WEBHOOK_URL: {
+      universalIdentifier: '7056c98a-e7e1-4dba-8a40-b578f30b3479',
+      description:
+        'Discord incoming webhook URL. When set, a notification is posted to this channel each time the application form creates a new Partner. Leave empty to disable. Set per-workspace in Settings → Apps → Twenty Partners → Variables.',
+      isSecret: true,
```
martmull, line 24: why isSecret true?

martmull (commented):
Nice code, some comment though

packages/twenty-apps/internal/twenty-partners/src/logic-functions/on-partner-application-created.logic-function.ts
rashad: Done in a4cb6be25b — the handler is now `payload: DatabaseEventPayload<ObjectRecordCreateEvent<CoreSchema.Partner>>`, so `properties.after` is typed and the cast is gone. One note: `twenty-client-sdk/generated` is not an exported subpath in the installed SDK (2.4.0) — `require.resolve` fails on it — so I pulled `Partner` from `CoreSchema` in `twenty-client-sdk/core`, matching the existing `submit-partner-application` handler. Typechecks cleanly (`ActorSourceEnum` is a string union, so the `createdBy.source` guard stays valid).

packages/twenty-apps/internal/twenty-partners/src/logic-functions/on-partner-application-created.logic-function.ts
rashad: Removed in a4cb6be25b — the typed payload makes the hand-rolled `PartnerCreatedAfter` unnecessary.

packages/twenty-apps/internal/twenty-partners/src/application-config.ts
rashad: A Discord webhook URL is effectively a write credential — anyone with the full URL can post to the channel, and Discord recommends treating them as secrets. `isSecret: true` stores it encrypted and masks it in Settings, consistent with `PARTNER_APPLICATION_SECRET`. The one tradeoff is you cannot read it back in the UI, which is fine for a paste-once value. Happy to flip it to `false` if you would rather it be visible for ops — your call.

martmull: ok, you can use secret header signature to secure the endpoint 
https://docs.discord.com/developers/events/webhook-events#setting-up-an-endpoint

rashad: Really appreciate you digging into the security angle here 🙏 — and you'\''re right that signature verification is the correct tool for that doc'\''s scenario. The only nuance is the direction of the webhook.

That page covers **Webhook Events** — *outgoing* webhooks, where **Discord sends events to an endpoint we host** and we verify each request via the Ed25519 `X-Signature-Ed25519` / `X-Signature-Timestamp` headers (401 on a bad signature, 204 to the initial PING).

This feature is the inverse: an **incoming webhook**, where *we* POST a message **to** a Discord channel (`discord.com/api/webhooks/<id>/<token>`). There'\''s no endpoint of ours in the loop — nothing receives a signed request from Discord — so there'\''s no signature to verify. For this direction, Discord'\''s model is that the token in the webhook URL *is* the credential (anyone with the full URL can post to the channel), which is why I kept `isSecret: true`: it'\''s stored encrypted and masked in Settings, consistent with `PARTNER_APPLICATION_SECRET`.

If we ever add the reverse (Twenty *subscribing* to Discord events), the signature-verification flow from that doc would be exactly the right approach — happy to pick that up as a follow-up if it'\''d be useful. And of course glad to switch this to `isSecret: false` if you'\''d rather it be visible in Settings for ops — just say the word. 🙂

martmull (approved):

PR #21319, opened 2026-06-08
fix: prevent deletion of standard fields via API

thomtrp:
## Summary
- Adds a server-side guard that rejects deletion of standard fields (`isCustom: false`) in the `deleteOneField` path
- The UI already prevents this, but the API had no enforcement, allowing standard fields to be deleted via direct GraphQL calls

Without this guard, deleting a standard field like `jobTitle` cascades to drop dependent generated columns (e.g. `searchVector`), leaving the object in a broken state where all subsequent queries fail with "Data validation error."

## Test plan
- [x] Call `deleteOneField` with a standard field ID → should return `FIELD_MUTATION_NOT_ALLOWED` error
- [x] Call `deleteOneField` with a custom field ID → should succeed as before
- [x] UI deactivation of standard fields still works (deactivate != delete)

### DIFF
--- a/packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-delete-field-input-to-flat-field-metadatas-to-delete.util.ts
+++ b/packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-delete-field-input-to-flat-field-metadatas-to-delete.util.ts
@@ -54,7 +54,10 @@ export const fromDeleteFieldInputToFlatFieldMetadatasToDelete = ({
     );
   }
 
-  if (belongsToTwentyStandardApp(flatFieldMetadataToDelete) || flatFieldMetadataToDelete.isSystem) {
+  if (
+    belongsToTwentyStandardApp(flatFieldMetadataToDelete) ||
+    flatFieldMetadataToDelete.isSystem
+  ) {
     throw new FieldMetadataException(
       `Cannot delete standard field "${flatFieldMetadataToDelete.name}"`,
       FieldMetadataExceptionCode.FIELD_MUTATION_NOT_ALLOWED,
--- a/packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-delete-field-input-to-flat-field-metadatas-to-delete.util.ts
+++ b/packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-delete-field-input-to-flat-field-metadatas-to-delete.util.ts
@@ -53,6 +53,13 @@ export const fromDeleteFieldInputToFlatFieldMetadatasToDelete = ({
     );
   }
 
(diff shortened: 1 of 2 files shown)


packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-delete-field-input-to-flat-field-metadatas-to-delete.util.ts
```diff
@@ -53,6 +53,13 @@ export const fromDeleteFieldInputToFlatFieldMetadatasToDelete = ({
     );
   }
 
+  if (!flatFieldMetadataToDelete.isCustom) {
```
charlesBochet: Let's wait to merge Coco's PR first warn him to take this into account

yarn.lock
```diff
@@ -45690,15 +45690,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"nanoid@npm:^3.3.12":
```
charlesBochet: rebase on fresh main, not needed anymore

charlesBochet (changes requested):

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-delete-field-input-to-flat-field-metadatas-to-delete.util.ts
Weiko: @thomtrp depends on what you want. If you need to check if it's from the standard app, use belongsToTwentyStandardApp. If you actually want to check if it's system, use isSystem. isCustom can be replaced case by case now

charlesBochet (approved):

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/from-delete-field-input-to-flat-field-metadatas-to-delete.util.ts
```diff
@@ -53,6 +54,16 @@ export const fromDeleteFieldInputToFlatFieldMetadatasToDelete = ({
...
 
+  if (
+    belongsToTwentyStandardApp(flatFieldMetadataToDelete) ||
+    flatFieldMetadataToDelete.isSystem
+  ) {
+    throw new FieldMetadataException(
+      `Cannot delete standard field "${flatFieldMetadataToDelete.name}"`,
+      FieldMetadataExceptionCode.FIELD_MUTATION_NOT_ALLOWED,
+    );
+  }
```
prastoin, line 65: ***Self reminder:*** shoudln't be done here
By default the validate build and run should throw if you attempt to remove a field that is not owned by the appUniversalIdentifier used contextually for the run

PR #21321, opened 2026-06-08
feat(server): in-app server-level admin management (#19785)

### LINKED ISSUE #19785: Admin panel access: cache prevents granting canAccessFullAdminPanel via DB, needs in-app management
## Problem

When granting `canAccessFullAdminPanel` to a user by directly updating the `core.user` table in the database, the change has no effect because the `CoreEntityCacheService` caches the user entity in both an in-memory local cache (30-minute TTL) and Redis (no TTL). The cache is only invalidated on user soft-delete, never on user updates.

This means:
- Self-hosters who follow the [documented SQL command](https://twenty.com/developers/self-host/capabilities/troubleshooting) to grant admin access (`UPDATE core."user" SET "canAccessFullAdminPanel" = TRUE`) will see no effect until they manually flush Redis and restart the server.
- There is no in-app way to grant or revoke server-level admin rights (`canAccessFullAdminPanel`, `canImpersonate`).
- Debugging this is time-consuming since the frontend correctly reflects the DB value (fetched fresh), but the backend guards use the stale cached value.

### Root cause

`AdminPanelGuard` reads `request.user.canAccessFullAdminPanel`, which comes from `CoreEntityCacheService.get('user', userId)`. This cache is never invalidated when user fields are updated — only on soft-delete (`user.service.ts`).

### 1. Add in-app admin management UI

Add a section in the admin panel (behind `ServerLevelImpersonateGuard` / `canImpersonate`) to:
- List all server-level admins (users with `canAccessFullAdminPanel = true`)
- Grant/revoke `canAccessFullAdminPanel` for any user
- Grant/revoke `canImpersonate` for any user
- Properly invalidate the user entity cache after any update

### 2. Security measures

When modifying server-level admin rights:
- **Email notification to all existing admins** whenever a user is granted or revoked `canAccessFullAdminPanel` or `canImpersonate`, so rogue grants are immediately visible.
- **Audit log entry** for every admin rights change (who granted, to whom, when).
- **Prevent self-revocation** of the last admin to avoid lockout.
- **Require re-authentication** (password/SSO) before granting admin rights to prevent session hijacking escalation.

### 3. Fix cache invalidation

As a defense-in-depth measure, also invalidate the user entity cache whenever the user entity is updated (not just on soft-delete). This ensures direct DB changes propagate correctly after a Redis flush + server restart, and prevents similar stale-cache bugs for other user fields.

## Impact

- **Self-hosters** cannot grant admin panel access without Redis access + server restart
- **Debugging cost** is high — the frontend shows the correct value while the backend rejects requests
- **Security gap** — direct DB changes to admin rights are unaudited and have no notification mechanism

FelixMalfait:
## Closes #19785

In-app management of **server-level admin rights** (`canAccessFullAdminPanel`, `canImpersonate`) so self-hosters no longer need raw SQL + a Redis flush + restart to grant access.

> **Draft** — feature complete; `/code-review` + `/security-review` run and addressed.

### Background
`AdminPanelGuard` / `ServerLevelImpersonateGuard` read `request.user.{canAccessFullAdminPanel,canImpersonate}`, hydrated each request from `CoreEntityCacheService.get('user', …)` (local 30-min + Redis no-TTL). The cache was only invalidated on soft-delete, so a raw `UPDATE core."user"` never took effect. The **first** signup auto-gets both flags; every subsequent admin previously needed raw SQL.

### UX
- **Admin Panel → General → Administrators**: a read-only overview of every user with server-level access; each row links to that user's admin page.
- **Find anyone** via the user search (Recent Users) — available to full admins and impersonators — then open their **admin user page**.
- On the user page, an **"Administrator access"** card (gated on `canAccessFullAdminPanel`) has two toggles — *Full admin panel access* and *Impersonation* — that work for **any** user (a user with no access shows both off). Mirrors how **Impersonate** already works (find user → user page → act). Each change opens a confirm dialog with a **2FA code** field; the last full admin's toggle is disabled.

### Backend / security
- **Cache fix** — invalidate the user entity cache on committed user updates (not just soft-delete) so privilege changes propagate (~100 ms, cluster-wide) with no restart.
- `getServerAdmins` query + `updateServerAdminAccess` mutation (any `targetUserId`), gated on `canAccessFullAdminPanel`.
- `NoImpersonationGuard` on both — an impersonated full-admin session can't be used to escalate an impersonator.
- Fresh **2FA TOTP step-up** (enrolled+verified method **and** a fresh code; genuine 2FA errors surface; dev-skip on trusted `NODE_ENV`).
- **Last-admin lockout** in a transaction with a pessimistic row lock (no TOCTOU).
- **Email-to-all-admins + affected user** (rendered once per locale), structured log, audit event-log emit.
- **Authorization**: the read-only `userLookupAdminPanel` + `adminPanelRecentUsers` lookups now accept `canAccessFullAdminPanel OR canImpersonate` (new `AdminPanelOrImpersonateGuard`), so a full admin without impersonate can still find users to manage. Workspace/impersonation queries stay impersonate-gated.

### Reviews
- `/code-review` (max effort): 3 security findings (impersonation-escalation sink, lockout TOCTOU, step-up accepting PENDING 2FA) — **all fixed**. `/simplify`: applied. `/security-review`: **no high/medium vulnerabilities**.

### Follow-ups (not in this PR)
- Unit tests for `AdminPanelServerAdminService` + a frontend test.
- Point the self-host troubleshooting docs at the new UI.
- OTP retry UX: `ConfirmationModal` closes on confirm, so a wrong code needs a reopen (kept to reuse the existing modal; no new pattern).

### Notes for reviewers
- `generated-admin/graphql.ts` entries were hand-added to match codegen output (admin codegen needs a running server); re-run `nx graphql:generate twenty-front --configuration=admin` to confirm parity.
- First-admin bootstrap (first signup) is unchanged.


packages/twenty-front/src/modules/settings/admin-panel/components/SettingsAdminServerAdminAccess.tsx
```diff
@@ -0,0 +1,271 @@
...
+    openModal(SERVER_ADMIN_ACCESS_CONFIRMATION_MODAL_ID);
+  };
+
+  const handleConfirm = async () => {
+    if (pendingChange === null) {
+      return;
+    }
+
+    try {
+      // Both flags can be sent in one mutation, so granting full access is a
```
FelixMalfait: Remove comment

packages/twenty-front/src/modules/settings/admin-panel/components/SettingsAdminServerAdminAccess.tsx
```diff
@@ -0,0 +1,271 @@
...
+    client: apolloAdminClient,
+  });
+
+  const [updateServerAdminAccess] = useMutation(
+    UpdateServerAdminAccessDocument,
+    { client: apolloAdminClient },
+  );
+
+  const serverAdmins = data?.getServerAdmins ?? [];
+  // A user with no access simply isn't in getServerAdmins, so both flags
```
FelixMalfait: Remove comment

packages/twenty-server/src/engine/core-modules/admin-panel/dtos/update-server-admin-access.input.ts
```diff
@@ -0,0 +1,35 @@
...
+  @IsOptional()
+  @IsBoolean()
+  canAccessFullAdminPanel?: boolean;
+
+  @Field({ nullable: true })
+  @IsOptional()
+  @IsBoolean()
+  canImpersonate?: boolean;
+
+  // Fresh TOTP step-up code, required outside development to authorize the change.
```
FelixMalfait: Remove comment

packages/twenty-server/src/engine/guards/admin-panel-or-impersonate.guard.ts
```diff
@@ -0,0 +1,17 @@
+import { type CanActivate, type ExecutionContext } from '@nestjs/common';
+import { GqlExecutionContext } from '@nestjs/graphql';
+
+// Read-only admin-panel lookups (user/recent-users search) are available to
```
FelixMalfait, line 4: I'm wonder if we should rather do canAccessAdminPanel as something that supersedes canImpersonate. I don't feel like canAccessAdminPanel is properly designed. Let's look at options. Maybe it's too much for this PR though, let's balance all options

packages/twenty-server/src/engine/core-modules/user/services/user.service.ts
```diff
@@ -446,9 +446,15 @@ export class UserService extends TypeOrmQueryService<UserEntity> {
 
     user.isEmailVerified = true;
 
-    return queryRunner
+    const savedUser = queryRunner
       ? await queryRunner.manager.save(UserEntity, user)
       : await this.userRepository.save(user);
+
+    // Invalidate the cached user entity so auth guards and strategies read the
```
FelixMalfait: Remove comment

packages/twenty-server/src/engine/core-modules/event-logs/emit/events/workspace-event/server-admin/server-admin-access-changed.ts
```diff
@@ -0,0 +1,25 @@
...
+  event: z.literal(SERVER_ADMIN_ACCESS_CHANGED_EVENT),
+  properties: z.strictObject({
+    targetUserId: z.string(),
+    canAccessFullAdminPanel: z.boolean(),
+    canImpersonate: z.boolean(),
+    message: z.string().optional(),
+  }),
+});
+
+export type ServerAdminAccessChangedTrackEvent = z.infer<
```
FelixMalfait, line 18: 1 export per file (unless other similar event viles already have this pattern, but I don't like the 3 exports)

packages/twenty-server/src/engine/core-modules/admin-panel/admin-panel.resolver.ts
```diff
@@ -159,6 +169,34 @@ export class AdminPanelResolver {
     return this.adminStatisticsService.getTopWorkspaces(searchTerm);
   }
 
+  @UseGuards(AdminPanelGuard, NoImpersonationGuard)
+  @Query(() => [ServerAdminDTO])
+  async getServerAdmins(): Promise<ServerAdminDTO[]> {
+    return this.adminServerAdminService.getServerAdmins();
+  }
+
+  // NoImpersonationGuard prevents granting/revoking server admin rights from
```
FelixMalfait: Remove comment

packages/twenty-server/src/engine/core-modules/admin-panel/services/admin-panel-server-admin.service.ts
```diff
@@ -0,0 +1,335 @@
...
+        throw new UserInputError('Invalid two-factor authentication code.');
+      }
+
+      // Surface genuine 2FA errors (e.g. a corrupted secret) instead of
+      // masking them as a configuration problem.
+      throw error;
+    }
+  }
+
+  // Notify every full admin (and the affected user) so unexpected privilege
```
FelixMalfait: Remove comment

packages/twenty-server/src/engine/core-modules/admin-panel/services/admin-panel-server-admin.service.ts
```diff
@@ -0,0 +1,335 @@
...
+      );
+    } catch (error) {
+      if (
+        error instanceof TwoFactorAuthenticationException &&
+        error.code === TwoFactorAuthenticationExceptionCode.INVALID_OTP
+      ) {
+        throw new UserInputError('Invalid two-factor authentication code.');
+      }
+
+      // Surface genuine 2FA errors (e.g. a corrupted secret) instead of
```
FelixMalfait: Remove comment

packages/twenty-server/src/engine/core-modules/admin-panel/services/admin-panel-server-admin.service.ts
```diff
@@ -0,0 +1,335 @@
...
+  }: {
+    actorUserId: string;
+    actorWorkspaceId: string;
+    actorUserWorkspaceId: string;
+    otp?: string;
+  }): Promise<void> {
+    const isDevelopment =
+      this.twentyConfigService.get('NODE_ENV') === NodeEnvironment.DEVELOPMENT;
+
+    if (isDevelopment) {
```
FelixMalfait, line 159: Looks a bit risky to disable this in dev?

FelixMalfait: I'd remove this?

packages/twenty-server/src/engine/core-modules/admin-panel/services/admin-panel-server-admin.service.ts
```diff
@@ -0,0 +1,335 @@
...
+      actorWorkspaceId,
+      targetUser,
+    });
+
+    await this.notifyAdministrators({ actor, targetUser });
+
+    return this.toServerAdminDTO(targetUser);
+  }
+
+  // Granting or revoking server-level admin rights is a high-privilege action,
```
FelixMalfait: Remove comment

packages/twenty-server/src/engine/core-modules/admin-panel/services/admin-panel-server-admin.service.ts
```diff
@@ -0,0 +1,335 @@
...
+          );
+        }
+
+        await manager.save(UserEntity, targetUser);
+      });
+    } else {
+      await this.userRepository.save(targetUser);
+    }
+
+    // Auth guards and the JWT strategy read the cached user entity, so the
```
FelixMalfait: Remove comment

packages/twenty-server/src/engine/core-modules/admin-panel/services/admin-panel-server-admin.service.ts
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    const isRevokingFullAdmin =
+      targetUser.canAccessFullAdminPanel === true &&
+      nextCanAccessFullAdminPanel === false;
+
+    targetUser.canAccessFullAdminPanel = nextCanAccessFullAdminPanel;
+    targetUser.canImpersonate = nextCanImpersonate;
+
+    if (isRevokingFullAdmin) {
```
FelixMalfait: Prefer early returns to nested ifs or if/else structures

packages/twenty-emails/src/emails/server-admin-access-changed.email.tsx
```diff
@@ -0,0 +1,72 @@
...
+    ? enabledLabel
+    : disabledLabel;
+  const impersonateStatus = canImpersonate ? enabledLabel : disabledLabel;
+
+  return (
+    <BaseEmail locale={locale}>
+      <Title value={i18n._('Server administrator access changed')} />
+      <MainText>
+        <Trans
+          id="{actorName} updated server administrator access for {targetName} ({targetEmail})."
```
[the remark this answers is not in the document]
FelixMalfait: Fixed in 70185997a8 — replaced the space-containing `id`s with stable keys (`serverAdminAccessChanged.*`) and moved the human text to the `message` prop, so output is unchanged.

packages/twenty-emails/src/emails/server-admin-access-changed.email.tsx
```diff
@@ -0,0 +1,72 @@
...
+          values={{ actorName, targetName, targetEmail }}
+        />
+        <br />
+        <br />
+        <Trans
+          id="Full admin panel access: {fullAdminStatus}"
```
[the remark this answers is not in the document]
FelixMalfait: Fixed in 70185997a8 — all four `<Trans>` blocks now use space-free `id` keys + `message`.

packages/twenty-emails/src/emails/server-admin-access-changed.email.tsx
```diff
@@ -0,0 +1,72 @@
...
+          values={{ fullAdminStatus }}
+        />
+        <br />
+        <Trans
+          id="Impersonation: {impersonateStatus}"
```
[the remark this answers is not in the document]
FelixMalfait: Fixed in 70185997a8 — this `<Trans>` now uses a space-free `id` + `message`.

packages/twenty-emails/src/emails/server-admin-access-changed.email.tsx
```diff
@@ -0,0 +1,72 @@
...
+          values={{ impersonateStatus }}
+        />
+        <br />
+        <br />
+        <Trans id="If you did not expect this change, review your server administrators immediately." />
```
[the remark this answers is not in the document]
FelixMalfait: Fixed in 70185997a8 — warning `<Trans>` now uses a space-free `id` + `message`.

packages/twenty-server/src/engine/core-modules/user/services/user.service.ts
```diff
@@ -446,9 +446,15 @@ export class UserService extends TypeOrmQueryService<UserEntity> {
...
+    // fresh isEmailVerified value instead of a stale cached one.
+    await this.coreEntityCacheService.invalidate('user', userId);
```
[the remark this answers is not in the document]
FelixMalfait: Fixed in 70185997a8 — cache invalidation is now skipped inside the queryRunner transaction; it only runs on the committed (non-queryRunner) path.

packages/twenty-front/src/modules/settings/admin-panel/components/SettingsAdminServerAdmins.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  return (
+    <Section>
+      <H2Title
+        title={t`Administrators`}
+        description={t`Users with server-level access. Open a user to grant or revoke access; use the search below to find anyone.`}
+      />
+      {loading ? (
+        <SettingsSectionSkeletonLoader />
+      ) : serverAdmins.length === 0 ? (
```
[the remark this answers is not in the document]
FelixMalfait: Fixed in 70185997a8 — added an error branch that shows a load-failure message instead of the empty state.

packages/twenty-front/src/modules/settings/admin-panel/components/SettingsAdminServerAdminAccess.tsx
```diff
@@ -0,0 +1,203 @@
...
+      });
+      await refetch();
+      enqueueSuccessSnackBar({
+        message: t`Server administrator access updated.`,
+      });
+    } catch (error) {
+      enqueueErrorSnackBar({
+        message:
+          error instanceof Error
+            ? error.message
```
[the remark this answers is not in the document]
FelixMalfait: Fixed in 70185997a8 — now shows a generic, non-technical failure message instead of `error.message`.

packages/twenty-server/src/engine/core-modules/admin-panel/services/admin-panel-server-admin.service.ts
```diff
@@ -0,0 +1,335 @@
...
+
+      for (const recipient of recipientsById.values()) {
+        const locale = recipient.locale || SOURCE_LOCALE;
+        const localeRecipients = recipientsByLocale.get(locale) ?? [];
+
+        localeRecipients.push(recipient);
+        recipientsByLocale.set(locale, localeRecipients);
+      }
+
+      await Promise.allSettled(
```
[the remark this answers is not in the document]
FelixMalfait: Fixed in 70185997a8 — `Promise.allSettled` rejections are now counted and logged per locale instead of being swallowed.

packages/twenty-front/src/modules/settings/admin-panel/components/SettingsAdminServerAdminAccess.tsx
FelixMalfait: Removed.

packages/twenty-front/src/modules/settings/admin-panel/components/SettingsAdminServerAdminAccess.tsx
FelixMalfait: Removed.

packages/twenty-server/src/engine/core-modules/admin-panel/dtos/update-server-admin-access.input.ts
FelixMalfait: Removed.

packages/twenty-server/src/engine/core-modules/user/services/user.service.ts
FelixMalfait: Removed.

packages/twenty-server/src/engine/core-modules/admin-panel/admin-panel.resolver.ts
FelixMalfait: Removed.

packages/twenty-server/src/engine/core-modules/admin-panel/services/admin-panel-server-admin.service.ts
FelixMalfait: Removed.

packages/twenty-server/src/engine/core-modules/admin-panel/services/admin-panel-server-admin.service.ts
FelixMalfait: Removed.

packages/twenty-server/src/engine/core-modules/admin-panel/services/admin-panel-server-admin.service.ts
FelixMalfait: Removed.

packages/twenty-server/src/engine/core-modules/admin-panel/services/admin-panel-server-admin.service.ts
FelixMalfait: Removed.

packages/twenty-server/src/engine/core-modules/admin-panel/services/admin-panel-server-admin.service.ts
FelixMalfait: Refactored — `updateServerAdminAccess` now always saves inside one transaction, with the last-admin lock under a single `if (isRevokingFullAdmin)` guard instead of an if/else.

packages/twenty-server/src/engine/core-modules/event-logs/emit/events/workspace-event/server-admin/server-admin-access-changed.ts
FelixMalfait: Kept to match the sibling `impersonation/impersonation.ts`, which uses the same 3 exports (event const, zod schema, inferred type) + `registerEvent` — it's the established pattern in this folder.

packages/twenty-server/src/engine/guards/admin-panel-or-impersonate.guard.ts
FelixMalfait: Good call-out — this is the orthogonal-vs-hierarchy question. Options:

- **(a) keep orthogonal** (today): supports a "can configure the server but cannot impersonate/snoop users" least-privilege role.
- **(b) `canAccessFullAdminPanel` supersedes `canImpersonate`** (full admin ⇒ impersonation): simpler mental model, but loses that least-privilege split.
- **(c) one ordered server-role** (none/support/admin): cleanest long-term, biggest change.

I lean to **deferring the redesign out of this PR** as you suggested — this guard is only a *read-only* widening (full admins can use the user lookup/search), not a change to what the flags mean. Happy to take the role redesign as a follow-up. Leaving this open for your call.

packages/twenty-server/src/engine/core-modules/admin-panel/services/admin-panel-server-admin.service.ts
FelixMalfait: It mirrors the impersonation step-up, which does the same `NODE_ENV === 'development'` bypass (`impersonation.service.ts`). It only skips when `NODE_ENV` is development (local) — staging/prod still require the fresh 2FA. Trade-off without it: granting admin locally needs a verified 2FA method + a live TOTP code, which complicates local testing/demos. Happy to drop the bypass for always-on 2FA if you prefer — your call.

FelixMalfait:
@claude fix CI and push to this branch the fix directly
---
Nx read the output from the cache instead of running the command for 2 out of 3 tasks.

Error: SDK metadata client changes detected. Please run 'npx nx run twenty-client-sdk:generate-metadata-client' and commit the changes.

The following SDK metadata client changes were detected: 
...

packages/twenty-server/src/engine/core-modules/admin-panel/services/admin-panel-server-admin.service.ts
```diff
@@ -0,0 +1,325 @@
...
+        'Enable two-factor authentication before managing server administrators.',
+      );
+    }
+
+    try {
+      await this.twoFactorAuthenticationService.verifyTwoFactorAuthenticationMethodForAuthenticatedUser(
+        actorUserId,
+        otp,
+        actorWorkspaceId,
+      );
```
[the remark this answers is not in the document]
FelixMalfait: It's fine like that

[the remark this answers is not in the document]
FelixMalfait: Fixed in 2e65fa8d12 — the step-up now verifies the code against *any* workspace where the actor has a verified 2FA method (2FA is per-userWorkspace, but server-admin access is global), instead of only the current workspace. A wrong code falls through to the next; the final failure is surfaced via the resolver's TwoFactorAuthenticationExceptionFilter.

packages/twenty-server/src/engine/core-modules/admin-panel/services/admin-panel-server-admin.service.ts
```diff
@@ -0,0 +1,335 @@
...
+        'Enable two-factor authentication before managing server administrators.',
+      );
+    }
+
+    for (const [
+      index,
+      workspaceId,
+    ] of workspaceIdsWithVerifiedTwoFactor.entries()) {
+      try {
+        await this.twoFactorAuthenticationService.verifyTwoFactorAuthenticationMethodForAuthenticatedUser(
```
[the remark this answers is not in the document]
FelixMalfait: Fixed in 5709d577fe — reverted the multi-workspace loop. The step-up now verifies against the actor's current workspace only (one attempt, like every other 2FA check in the app), so a single request can't make one guess per workspace. If 2FA isn't set up in the current workspace it returns a clear, actionable error instead.

packages/twenty-server/src/engine/core-modules/admin-panel/services/admin-panel-server-admin.service.ts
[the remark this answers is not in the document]
FelixMalfait: Follow-up: the cross-workspace loop was reverted in 5709d577fe because it weakened 2FA brute-force resistance (one guess per workspace). Took your *alternative* suggestion instead — verify against the current workspace and surface a clear error if 2FA isn't set up there.

packages/twenty-front/src/modules/settings/admin-panel/components/SettingsAdminServerAdminAccess.tsx
```diff
@@ -0,0 +1,264 @@
...
+        },
+      });
+      await refetch();
+      enqueueSuccessSnackBar({
+        message: t`Server administrator access updated.`,
+      });
+    } catch {
+      enqueueErrorSnackBar({
+        message: t`Failed to update server administrator access.`,
+      });
```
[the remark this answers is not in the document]
FelixMalfait: Fixed in ff89bc9d69 — the catch now passes GraphQL errors to enqueueErrorSnackBar (via CombinedGraphQLErrors.is), so the specific backend message is shown ("Invalid verification code", "last server administrator", "enable 2FA"…); the generic fallback only applies to non-GraphQL errors.

FelixMalfait:
@claude fix the user friendly message, it shows "an error occured" to the end user. It should show "Enable two-factor authentication in your current workspace to manage server administrators" for example. Don't confuse the message and friendly message. Also the "admin panel" and "impersonation" chips don't look nice, try to find something nicer that's already used.
Fix this and push to the branch directly.

PR #21326, opened 2026-06-08
security: bump uuid 9 → 11 (server, shared, front)

charlesBochet:
Clears the `uuid` "missing buffer bounds check in v3/v5/v6" advisory — patched in **11.1.1**. Bumps `twenty-server`, `twenty-shared`, `twenty-front` from 9 → `^11.1.1`.

### Why 11 and not 13
uuid **11.1.x still ships a CommonJS build**, so jest loads it with **no config changes**. uuid went **ESM-only at v12+**, which would otherwise force `transformIgnorePatterns` workarounds across the jest projects (and broke server/integration/storybook CI on the earlier 13 attempt). 11.1.1 is the actual patched version, so this is the minimal fix.

### Changes
- `uuid` → `^11.1.1` in the three workspaces (lockfile regenerated under hardened mode)
- one test (`useCreateManyRecords.test.tsx`): pin the mocked `v4` to its string-returning overload — uuid's types declare a `Uint8Array` overload that `jest.mocked` resolves to (present in v11 too, unrelated to ESM).

All usages are named imports, so no source migration. typecheck passes (server/shared/front); affected specs pass. **No jest config changes.**

### DIFF
--- a/packages/twenty-front/package.json
+++ b/packages/twenty-front/package.json
@@ -142,7 +142,7 @@
     "twenty-ui": "workspace:*",
     "type-fest": "4.10.1",
     "use-debounce": "^10.0.0",
-    "uuid": "^9.0.0",
+    "uuid": "^11.1.1",
     "xlsx-ugnis": "^0.19.3",
     "zod": "^4.1.11"
   },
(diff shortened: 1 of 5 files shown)


bosiraphael (approved):
LGTM

PR #20890, opened 2026-05-25
Deprecate dummy enterprise key 1/2

ijreilly:
Remove usage of hasValidEnterpriseKey in FE (replaced by hasValidSignedEnterpriseKey)

To avoid breaking change at deploy time, we will wait until after this has been deployed in prod, to remove hasValidEnterpriseKey in the BE.

PR #21328, opened 2026-06-08
Deprecate dummy enterprise key 2/2

ijreilly:
Following [1/2](https://github.com/twentyhq/twenty/pull/20890)

Now that all usages of hasValidEnterpriseKey has been removed in prod and deployed, we can safely remove it altogether.


Weiko (approved):
LGTM

PR #21331, opened 2026-06-08
perf(ai): lazy-load agent chat runtime so it doesn't fetch/diff threads until opened

Weiko:
## Problem

On workspaces with a sizeable AI chat history, the whole app was freezing during navigation, including Settings (one navigation click measured ~6.5s).

## Root cause

`AgentChatProvider` is mounted app-wide in `AppRouterProviders`, so its effects run on every page. 
On every render it would:
1. auto-select the most recently active thread (`AgentChatThreadInitializationEffect`),
2. fetch that thread's **full message history** (`AgentChatMessagesFetchEffect`),
3. run `AgentChatStreamingPartsDiffSyncEffect` → `updateStreamingPartsWithDiff`, which loops over every message doing `isDeeplyEqual(existing, incoming)` + `structuredClone`.

A large thread would produce multi-second freeze on every interaction, app-wide. (Confirmed via a Chrome CPU profile)

 ## Fix

Don't run the agent-chat **message runtime** until the chat is actually opened.

## Note
There is still room for improvement, opening AI chats would still be very slow.

### DIFF
--- a/packages/twenty-front/src/modules/ai/components/AgentChatRuntimeEffects.tsx
+++ b/packages/twenty-front/src/modules/ai/components/AgentChatRuntimeEffects.tsx
@@ -1,0 +1,44 @@
+import { AgentChatMessagesFetchEffect } from '@/ai/components/AgentChatMessagesFetchEffect';
+import { AgentChatSessionStartTimeEffect } from '@/ai/components/AgentChatSessionStartTimeEffect';
+import { AgentChatStreamSubscriptionEffect } from '@/ai/components/AgentChatStreamSubscriptionEffect';
+import { AgentChatStreamingAutoScrollEffect } from '@/ai/components/AgentChatStreamingAutoScrollEffect';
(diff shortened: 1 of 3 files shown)


martmull (approved):

packages/twenty-front/src/modules/ai/components/AgentChatRuntimeEffects.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const [hasAgentChatBeenOpened, setHasAgentChatBeenOpened] = useAtomState(
+    hasAgentChatBeenOpenedState,
+  );
+
+  const isAgentChatOpen =
+    isSidePanelOpened && sidePanelPage === SidePanelPages.AskAI;
+
+  useEffect(() => {
+    if (isAgentChatOpen && !hasAgentChatBeenOpened) {
+      setHasAgentChatBeenOpened(true);
```
charlesBochet, line 27: a imperative approach would be better (when we open the chat by a hook we also set this state) but LGTM

PR #21332, opened 2026-06-08
chore: remove accidentally committed screenshots

parshipcy:
I think these screenshots were accidentally committed at the repo root in #21033. They are not referenced anywhere in the codebase, so removing them to keep the repo clean.


charlesBochet (approved):
Thank you!

PR #21340, opened 2026-06-08
fix(sdk): bump inquirer ^10 -> ^14 to drop external-editor/tmp from apps

charlesBochet:
The **central fix** for the `tmp` Dependabot alerts in `packages/twenty-apps/*` — so apps don't each need a per-app `resolutions` entry.

### Root cause
Every app depends on `twenty-sdk`, whose inquirer chain pulls the vulnerable `tmp`:
```
twenty-sdk → inquirer ^10 → @inquirer/prompts 7.x → @inquirer/editor 4.x → external-editor → tmp@0.0.33
```
`@inquirer/editor 5.x` dropped `external-editor` (and thus old `tmp`), and it's only reached via `@inquirer/prompts 8.x`, which requires **inquirer ≥ 13.4.3**. So `^12` isn't enough — bump to **`^14`** (latest):
```
inquirer 14 → @inquirer/prompts 8.5.2 → @inquirer/editor 5.2.2   (no external-editor)
```

### Verified
- The SDK uses the classic `inquirer.prompt([...])` API (uninstall / add / remote commands) — **typechecks cleanly under inquirer 14**.
- After the bump, the SDK's subtree resolves `@inquirer/editor@5.2.2` (the lingering `external-editor` in this repo's lockfile is from *other* consumers — `nx`/`zapier` — handled separately).

### Propagation
Fixes it **once** for every app using the SDK, with no per-app `package.json` additions. Existing `twenty-apps/*` clear their `tmp` alert once a new `twenty-sdk` is published and they bump to it; newly-scaffolded apps are clean immediately.

(The root-workspace `tmp` from `nx`/`zapier` is handled by twenty#21338.)

### DIFF
--- a/packages/twenty-sdk/package.json
+++ b/packages/twenty-sdk/package.json
@@ -93,7 +93,6 @@
   "devDependencies": {
     "@genql/cli": "^3.0.3",
     "@prettier/sync": "^0.5.2",
-    "@types/inquirer": "^9.0.0",
     "@types/node": "^24.0.0",
     "@types/react": "^19.0.0",
     "@types/react-dom": "^19.0.0",
--- a/packages/twenty-sdk/package.json
+++ b/packages/twenty-sdk/package.json
@@ -80,7 +80,7 @@
     "graphql": "^16.8.1",
     "graphql-sse": "^2.5.4",
     "ink": "^6.8.0",
-    "inquirer": "^10.0.0",
+    "inquirer": "^14.0.0",
     "jsonc-parser": "^3.2.0",
     "preact": "^10.28.3",
     "react": "^19.0.0",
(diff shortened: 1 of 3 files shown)


charlesBochet:
Addressed both review findings in 9e80d75:

- **P1 (sentry/cubic, HIGH):** `type: 'list'` → `type: 'select'` in `remote/index.ts` — confirmed real. inquirer 14 (`@inquirer/prompts`) removed the `list` alias; typecheck passed but it throws `UnknownPromptTypeError` at runtime.
- **P2 (cubic):** removed stale `@types/inquirer@^9` — inquirer 14 ships its own types (typecheck still green without it).

**Runtime-verified** against inquirer 14.0.2 by driving the prompt engine directly: `select`/`confirm`/`input`/`password` all resolve correctly; `list` throws `UnknownPromptTypeError: Prompt type "list" is not registered` — exactly the predicted crash.

PR #21347, opened 2026-06-08
fix(auth): additional workspace and identity validation in auth flows

FelixMalfait:
## What

Adds validation across three auth flows so that a session or reset link is consistently scoped to a workspace the authenticated principal belongs to, and to a verified identity.

- **Access token** (`jwt.auth.strategy.ts`): when resolving the request's user context, the token's `userWorkspaceId` must belong to the token's `workspaceId` — the same check the application-token path already performs.
- **OIDC** (`oidc.auth.strategy.ts`): reject sign-in when the identity provider explicitly reports `email_verified: false`.
- **Password reset** (`reset-password.service.ts`): a supplied `workspaceId` is only used when the user is a member of it; otherwise it falls back to the user's own first password-auth-enabled workspace.

## Tests

- `jwt.auth.strategy.spec.ts`: rejects an access token whose user workspace belongs to a different workspace than the token; existing mocks updated to carry the cached `workspaceId`.
- `oidc.auth.strategy.spec.ts` (new): rejects unverified email; accepts verified and absent-claim cases.
- `reset-password.service.spec.ts`: falls back when the supplied `workspaceId` is not one the user belongs to.

`tsgo`, `oxlint` and `oxfmt` all clean on the changed files.

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/auth/services/reset-password.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/auth/services/reset-password.service.ts
@@ -62,9 +62,10 @@ export class ResetPasswordService {
       }),
     );
 
-    const targetWorkspaceId =
-      workspaceId ??
-      (await this.findFirstPasswordAuthEnabledWorkspaceIdOrThrow(user.id));
+    const targetWorkspaceId = await this.resolveTargetWorkspaceId(
+      user.id,
+      workspaceId,
+    );
 
     const expiresIn = this.twentyConfigService.get(
       'PASSWORD_RESET_TOKEN_EXPIRES_IN',
@@ -122,6 +123,31 @@ export class ResetPasswordService {
       passwordResetToken: plainResetToken,
       passwordResetTokenExpiresAt: expiresAt,
     };
+  }
+
+  private async resolveTargetWorkspaceId(
+    userId: string,
+    workspaceId?: string,
+  ): Promise<string> {
+    if (!isDefined(workspaceId)) {
+      return this.findFirstPasswordAuthEnabledWorkspaceIdOrThrow(userId);
+    }
+
+    const requestedWorkspace = await this.workspaceRepository.findOne({
+      where: {
+        id: workspaceId,
+        isPasswordAuthEnabled: true,
+        workspaceUsers: {
+          user: {
+            id: userId,
+          },
+        },
+      },
+    });
+
+    return isDefined(requestedWorkspace)
+      ? requestedWorkspace.id
+      : this.findFirstPasswordAuthEnabledWorkspaceIdOrThrow(userId);
   }
 
   async sendEmailPasswordResetLink({
--- a/packages/twenty-server/src/engine/core-modules/auth/strategies/jwt.auth.strategy.spec.ts
+++ b/packages/twenty-server/src/engine/core-modules/auth/strategies/jwt.auth.strategy.spec.ts
@@ -332,6 +332,61 @@ describe('JwtAuthStrategy', () => {
       expect(user.user?.lastName).toBe('lastNameDefault');
       expect(user.userWorkspaceId).toBe(validUserWorkspaceId);
     });
+
+    it('should reject when the user workspace belongs to a different workspace than the token', async () => {
+      const validUserId = 'valid-user-id';
+      const validUserWorkspaceId = randomUUID();
+      const tokenWorkspaceId = randomUUID();
+      const otherWorkspaceId = randomUUID();
+
+      const payload = {
+        sub: validUserId,
+        type: JwtTokenTypeEnum.ACCESS,
+        userWorkspaceId: validUserWorkspaceId,
+        workspaceId: tokenWorkspaceId,
+      };
+
+      const mockWorkspace = new WorkspaceEntity();
+
+      mockWorkspace.id = tokenWorkspaceId;
+      workspaceStore[tokenWorkspaceId] = mockWorkspace;
+      userStore[validUserId] = { id: validUserId, lastName: 'lastNameDefault' };
+
+      coreEntityCacheService.get.mockImplementation(
+        async (keyName: string, entityId: string) => {
+          if (keyName === 'workspaceEntity') {
+            return workspaceStore[entityId] ?? null;
+          }
+
+          if (keyName === 'user') {
+            return userStore[entityId] ?? null;
+          }
+
+          if (keyName === 'userWorkspaceEntity') {
+            return {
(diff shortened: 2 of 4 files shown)


packages/twenty-server/src/engine/core-modules/auth/strategies/oidc.auth.strategy.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+    extractState: jest.fn().mockReturnValue({
+      identityProviderId: 'identity-provider-id',
+      workspaceInviteHash: undefined,
+    }),
+  });
+
+  const invokeValidate = (context: ReturnType<typeof buildContext>) => {
+    const done = jest.fn();
+
+    return OIDCAuthStrategy.prototype.validate
```
FelixMalfait: Invoking `validate` through the prototype here is deliberate: `openid-client`'s `Strategy` base constructor throws `client must be an instance of openid-client Client`, so the strategy can't be instantiated with a mock in a unit test. Calling `validate` with a mock `this` (providing `client.userinfo` and `extractState`) lets us exercise the `email_verified` branch in isolation without a live IdP.

packages/twenty-server/src/engine/core-modules/auth/strategies/oidc.auth.strategy.ts
```diff
@@ -94,6 +94,15 @@ export class OIDCAuthStrategy extends PassportStrategy(
 
       const userinfo = await this.client.userinfo(tokenset);
 
+      if (userinfo.email_verified === false) {
```
FelixMalfait: Not sure this is a good idea, this could break! Are we sure all providers pass email_verified?!

packages/twenty-server/src/engine/core-modules/auth/services/reset-password.service.ts
```diff
@@ -124,6 +125,30 @@ export class ResetPasswordService {
...
+          user: {
+            id: userId,
+          },
+        },
+      },
+    });
+
+    return isDefined(requestedWorkspace)
+      ? requestedWorkspace.id
+      : this.findFirstPasswordAuthEnabledWorkspaceIdOrThrow(userId);
```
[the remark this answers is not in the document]
FelixMalfait: Fixed in ffb0bff5ba — `resolveTargetWorkspaceId` now includes `isPasswordAuthEnabled: true` in the workspace lookup, so a supplied workspaceId is only used when it's a password-auth-enabled workspace the user belongs to; otherwise it falls back to `findFirstPasswordAuthEnabledWorkspaceIdOrThrow`. This aligns it with the fallback's intent.

packages/twenty-server/src/engine/core-modules/auth/strategies/oidc.auth.strategy.ts
```diff
@@ -94,6 +94,15 @@ export class OIDCAuthStrategy extends PassportStrategy(
...
+        return done(
+          new AuthException(
+            'Email not verified by identity provider',
+            AuthExceptionCode.INVALID_DATA,
```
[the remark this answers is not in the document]
FelixMalfait: Fixed in ffb0bff5ba — switched the unverified-email rejection to `AuthExceptionCode.EMAIL_NOT_VERIFIED` (which has its own exception-filter mapping) instead of `INVALID_DATA`.

packages/twenty-server/src/engine/core-modules/auth/services/reset-password.service.ts
```diff
@@ -124,6 +125,31 @@ export class ResetPasswordService {
...
 
+  private async resolveTargetWorkspaceId(
+    userId: string,
+    workspaceId?: string,
+  ): Promise<string> {
+    if (!isDefined(workspaceId)) {
+      return this.findFirstPasswordAuthEnabledWorkspaceIdOrThrow(userId);
+    }
+
+    const requestedWorkspace = await this.workspaceRepository.findOne({
```
Weiko, line 136: Should we use cache?

FelixMalfait: This mirrors the adjacent `findFirstPasswordAuthEnabledWorkspaceIdOrThrow` fallback (same direct `workspaceRepository.findOne` + `isPasswordAuthEnabled`), and it's a cold path (a password-reset request), so I kept it uncached for consistency — and to read fresh `isPasswordAuthEnabled`/membership state for the security check. It's also a membership + flag join rather than a by-id fetch, so a workspace-by-id cache wouldn't cover it directly (and this service doesn't currently inject a cache). Happy to switch to a cached lookup if you'd prefer.

FelixMalfait: Sorry AI reply, will lookup m

packages/twenty-server/src/engine/core-modules/auth/strategies/oidc.auth.strategy.ts
FelixMalfait: Dropped in 5624b60b14 — removed the `email_verified` check and its test entirely. The explicit-`false` rejection risked breaking non-conformant OIDC providers, and post-#21346 the IdP is already workspace-bound so the value was marginal. PR now keeps just the access-token workspace binding + password-reset workspace scoping.

packages/twenty-server/src/engine/core-modules/auth/services/reset-password.service.ts
FelixMalfait: I hesitated but I think we can leave like that to be extra paranoid (no cache on low usage endpoint + needs the freshest possible data), honestly we could argue both so happy to fix if you have a strong preference!

packages/twenty-server/src/engine/core-modules/auth/strategies/jwt.auth.strategy.spec.ts
```diff
@@ -332,6 +332,61 @@ describe('JwtAuthStrategy', () => {
       expect(user.user?.lastName).toBe('lastNameDefault');
       expect(user.userWorkspaceId).toBe(validUserWorkspaceId);
     });
+
+    it('should reject when the user workspace belongs to a different workspace than the token', async () => {
```
FelixMalfait, line 336: I now it's not really used but if that's really the case let's do a PR to remove all @charlesBochet? Wdyt?

FelixMalfait:
Raising another PR to fix the broken tests (broken on main)

PR #21357, opened 2026-06-09
Remove default command

martmull:
fix conflict with not existing commands

### DIFF
--- a/packages/twenty-sdk/src/cli/commands/dev/index.ts
+++ b/packages/twenty-sdk/src/cli/commands/dev/index.ts
@@ -55,7 +55,7 @@ export const registerDevCommands = (program: Command): void => {
   };
 
   program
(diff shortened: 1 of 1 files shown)


packages/twenty-sdk/src/cli/commands/dev/index.ts
```diff
@@ -55,7 +55,7 @@ export const registerDevCommands = (program: Command): void => {
   };
 
   program
-    .command('dev [appPath]', { isDefault: true })
+    .command('dev [appPath]')
     .description('Build and sync local changes (default command)')
```
Weiko: probably need to remove `(default command)` here then

Weiko (approved):
LGTM

charlesBochet (approved):

PR #21359, opened 2026-06-09
feat(logic-function): add buildStatus enum, deprecate isBuildUpToDate

charlesBochet:
## Why

Logic-function build state is tracked today by a single boolean `isBuildUpToDate`, which conflates **two** distinct stages:

1. the code is **built** from source into an artifact (file storage + checksum), and
2. the built artifact is **deployed** into the executor (e.g. the Lambda gets the code embedded — `installPrebuiltBundle`).

We need to tell those apart so we can move stage 2 (the slow, external executor deploy) out of the workspace-migration DB transaction and run it asynchronously. Today that deploy runs *inside* the migration transaction, so a deploy touching N functions serializes N Lambda updates under one open transaction — which times out the request gateway and can kill the connection mid-transaction (surfaced recently as a `Migration action 'update' for 'logicFunction' failed` / `Query runner already released` incident).

This is **PR 1 of 2** and is intentionally **behavior-preserving**: it introduces the state model and switches the read path, but does **not** change where the deploy happens yet.

## What

`LogicFunctionBuildStatus`:

| Value | Code built? | Deployed to executor? | Runs as |
|---|---|---|---|
| `NOT_BUILT` | no | — | not runnable |
| `CODE_BUILT` | yes | not yet | LIVE *(reserved for PR2)* |
| `READY` | yes | yes (or LIVE mode) | PREBUILT / LIVE |
| `DEPLOY_FAILED` | yes | failed | LIVE *(reserved for PR2)* |

- New enum + `buildStatus` column on `LogicFunctionEntity` (default `READY`). `isBuildUpToDate` kept and **mirrored**, marked `@deprecated` (removed in a follow-up once call-sites migrate).
- **Behavior-preserving:** every site that wrote `isBuildUpToDate` now also writes `buildStatus` (`false → NOT_BUILT`, `true → READY`). `CODE_BUILT`/`DEPLOY_FAILED` are defined but **never written** in this PR.
- Read path switched to `buildStatus === READY`. Invoke-time mode resolution extracted into `resolveEffectiveLogicFunctionExecutionMode`, which falls back to **LIVE** for `CODE_BUILT`/`DEPLOY_FAILED` (dormant until PR 2) and blocks only `NOT_BUILT` — identical outcomes to today for the only states that occur here.
- Schema via a **fast instance command** in `2-11/` (TypeORM migrations are frozen), creating the enum type, adding the column (default `READY`), and backfilling from `isBuildUpToDate`.
- Registered `buildStatus` in the flat-entity property config + editable properties. **GraphQL DTO intentionally untouched** to avoid schema-snapshot churn.

## Follow-up (PR 2)
Move `installPrebuiltBundle` out of the migration transaction onto the `logicFunctionQueue`: the migration commits with `buildStatus = CODE_BUILT`, an async job deploys and flips to `READY` (or `DEPLOY_FAILED`), functions run LIVE in the meantime, and a reconcile sweep retries stragglers. Also: stop masking the underlying error in `WorkspaceMigrationRunnerService` and guard the rollback when the query runner is already released.

## Testing
- `npx nx typecheck twenty-server` — clean.
- Server jest suite not run locally — relying on CI. Reviewers: please check any exhaustive flat-property snapshot/spec for `logicFunction`.

### DIFF
--- a/packages/twenty-server/src/database/commands/upgrade-version-command/2-11/2-11-instance-command-fast-1799000040000-add-logic-function-build-status.ts
+++ b/packages/twenty-server/src/database/commands/upgrade-version-command/2-11/2-11-instance-command-fast-1799000040000-add-logic-function-build-status.ts
@@ -1,0 +1,30 @@
+import { type QueryRunner } from 'typeorm';
+
+import { RegisteredInstanceCommand } from 'src/engine/core-modules/upgrade/decorators/registered-instance-command.decorator';
+import { type FastInstanceCommand } from 'src/engine/core-modules/upgrade/interfaces/fast-instance-command.interface';
+
+@RegisteredInstanceCommand('2.11.0', 1799000040000)
+export class AddLogicFunctionBuildStatusFastInstanceCommand
+  implements FastInstanceCommand
+{
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE TYPE "core"."logicFunction_buildstatus_enum" AS ENUM('NOT_BUILT', 'CODE_BUILT', 'READY', 'DEPLOY_FAILED')`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "core"."logicFunction" ADD "buildStatus" "core"."logicFunction_buildstatus_enum" NOT NULL DEFAULT 'READY'`,
+    );
+    await queryRunner.query(
+      `UPDATE "core"."logicFunction" SET "buildStatus" = CASE WHEN "isBuildUpToDate" THEN 'READY' ELSE 'NOT_BUILT' END`,
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "core"."logicFunction" DROP COLUMN "buildStatus"`,
+    );
+    await queryRunner.query(
+      `DROP TYPE "core"."logicFunction_buildstatus_enum"`,
+    );
+  }
+}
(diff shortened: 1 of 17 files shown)


packages/twenty-server/src/database/commands/upgrade-version-command/2-11/2-11-instance-command-fast-1799000040000-add-logic-function-build-status.ts
```diff
@@ -0,0 +1,30 @@
+import { type QueryRunner } from 'typeorm';
+
+import { RegisteredInstanceCommand } from 'src/engine/core-modules/upgrade/decorators/registered-instance-command.decorator';
+import { type FastInstanceCommand } from 'src/engine/core-modules/upgrade/interfaces/fast-instance-command.interface';
+
+@RegisteredInstanceCommand('2.11.0', 1799000040000)
```
Weiko, line 6: 2.12?

packages/twenty-server/src/engine/metadata-modules/logic-function/logic-function.entity.ts
```diff
@@ -65,9 +72,18 @@ export class LogicFunctionEntity
...
   @Column({ nullable: false, type: 'boolean', default: true })
   isBuildUpToDate: boolean;
 
+  @Column({
+    type: 'enum',
+    enum: LogicFunctionBuildStatus,
+    default: LogicFunctionBuildStatus.READY,
+    nullable: false,
+  })
+  buildStatus: LogicFunctionBuildStatus;
```
Weiko, line 85: Missing upgrade decorator?

charlesBochet:
Closing, this does not seem to be the root cause, can be tackled later

PR #21362, opened 2026-06-09
fix: gracefully handle missing logic functions during workflow destroy

thomtrp:
## Summary
- Wraps `deleteOneWithSource` calls in `.catch()` during workflow/step destruction so that a missing logic function (valid UUID but already deleted) no longer crashes the entire destroy operation
- Adds a `Logger` to `WorkflowVersionStepOperationsWorkspaceService` for the warning
- Fixes test mock to return a resolved Promise and use a valid UUID

## Context
When a CODE step references a `logicFunctionId` that is a valid UUID but the logic function no longer exists (e.g. deleted by a previous operation or orphaned), the destroy fails with "Logic function with id X not found". This blocks users from cleaning up workflows.

## Test plan
- [x] Destroy a workflow with CODE steps whose logic functions already exist → succeeds as before
- [ ] Destroy a workflow with CODE steps referencing a deleted/non-existent logic function → succeeds with a warning log instead of crashing


Weiko (approved):
@thomtrp should we update the faulty step at some point? (probably not in this flow)

PR #21370, opened 2026-06-09
feat(twenty-server): allow shouldHideEmptyGroups in app view manifest

FelixMalfait:
## Context

The view **Hide empty groups** setting (`shouldHideEmptyGroups`) can be toggled in the UI, is persisted on the `View` entity, exposed in the `CreateView`/`UpdateView` GraphQL inputs, and tracked by the flat-view sync machinery — but it could **not** be set from an app's view manifest.

Root cause: the field postdates the manifest plumbing (added in #16385, Dec 2025). Two spots were never updated to thread it through:
- `ViewManifest` didn't declare the field.
- `fromViewManifestToUniversalFlatView` hardcoded `shouldHideEmptyGroups: false`.

Ref: twentyhq/core-team-issues#414

## Changes

- Add optional `shouldHideEmptyGroups?: boolean` to `ViewManifest`.
- Read it in the converter (`?? false`), mirroring the existing `isCompact` handling.
- Cover it in the converter unit test (default + explicit value).

No migration or schema change — the column already exists, and downstream sync (`FLAT_VIEW_EDITABLE_PROPERTIES` + the universal-flat compare type) already handles it.

## Test

- `npx jest from-view-manifest-to-universal-flat-view` → 5 passed
- `tsgo -p tsconfig.json` (twenty-server) → no new errors
- oxlint + oxfmt clean

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/application/application-manifest/converters/__tests__/from-view-manifest-to-universal-flat-view.util.spec.ts
+++ b/packages/twenty-server/src/engine/core-modules/application/application-manifest/converters/__tests__/from-view-manifest-to-universal-flat-view.util.spec.ts
@@ -35,6 +35,7 @@ describe('fromViewManifestToUniversalFlatView', () => {
     expect(result.icon).toBe('IconList');
     expect(result.position).toBe(0);
     expect(result.isCompact).toBe(false);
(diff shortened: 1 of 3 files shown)


Weiko (approved):
LGTM

Issue #1434, opened 2023-09-04
Bug: Updated table columns are not being persisted


thaisguigon:
### Context

A table's modified column should be persisted on "Update view" button click. Currently clicking on the "Update view" button after resizing or changing a column's visibility causes an ApolloError.

### Expected behavior

**Scenario: Persisting a resized table column**

**GIVEN** I am on the Companies/People page
**WHEN** I resize a table column
**AND** I click on the "Update view" button in the "Filter and Sort" bar
**AND** I reload the page
**THEN** the table column has kept its new size

**Scenario: Persisting a table column's visibility**

**GIVEN** I am on the Companies/People page
**WHEN** I change a table column's visibility
**AND** I click on the "Update view" button in the "Filter and Sort" bar
**AND** I reload the page
**THEN** the table column has kept its new visibility

PR #21372, opened 2026-06-09
security: force shell-quote >= 1.8.4 (GHSA-w7jw-789q-3m8p, critical)

charlesBochet:
## What

`shell-quote <= 1.8.3` is affected by [GHSA-w7jw-789q-3m8p](https://github.com/advisories/GHSA-w7jw-789q-3m8p) / CVE-2026-9277 (**critical**): `quote()` backslash-escapes `.op` characters with `/(.)/g`, which doesn't match line terminators (`\n`, `\r`, U+2028/2029). A line terminator in an object token's `.op` value passes through unescaped, and POSIX shells treat a literal `\n` as a command separator — enabling shell command injection in callers that pass attacker-influenced object tokens to `quote()`. First patched in **1.8.4**.

This is Dependabot alert #1434 on the root `yarn.lock`.

## How

The root lockfile resolved two vulnerable versions:

- `1.8.1` — via the `^1.6.1` / `^1.7.3` / `^1.8.1` ranges
- `1.8.3` — **hard-pinned** by `concurrently@9.2.1` (used in `twenty-companion`)

`yarn up -R shell-quote` only re-resolves the ranged dependents; the exact `1.8.3` pin from `concurrently` stays. So I added a `shell-quote: "^1.8.4"` entry to root `resolutions`, matching the existing `tmp` / `chokidar` / `tar` security overrides. Every consumer now resolves to the patched `1.8.4`.

## Scope

- `package.json`: +1 resolution line.
- `yarn.lock`: two vulnerable entries collapse to a single `shell-quote@1.8.4`.
- `1.8.4` is a semver-compatible patch over `1.8.3`; latest `concurrently` (10.x) already depends on `1.8.4`.
- Verified no `shell-quote <= 1.8.3` remains in any lockfile across the repo.

### DIFF
--- a/packages/twenty-companion/package.json
+++ b/packages/twenty-companion/package.json
@@ -46,7 +46,7 @@
     "@timfish/forge-externals-plugin": "^0.2.1",
     "axios": "^1.16.0",
     "codemirror": "^6.0.1",
-    "concurrently": "^9.2.1",
+    "concurrently": "^10.0.0",
     "dotenv": "^16.5.0",
     "easymde": "^2.20.0",
     "electron-squirrel-startup": "^1.0.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -30126,6 +30126,13 @@ __metadata:
   languageName: node
   linkType: hard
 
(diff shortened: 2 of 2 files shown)


bosiraphael (approved):
LGTM

PR #21389, opened 2026-06-10
fix(twenty-front): new layout fast-follows — settings drawer, loading & command menu

FelixMalfait:
Second batch of new-layout fast-follows (master: twentyhq/core-team-issues#2478). All changes verified live against a running workspace.

## Settings drawer & header
- **twentyhq/core-team-issues#2489** — sidebar icons render as plain 16px icons, no background tiles.
- **twentyhq/core-team-issues#2488** — Advanced toggle spans the full drawer width; yellow dot removed.
- **twentyhq/core-team-issues#2497** — page title stays centered in the settings header (breadcrumb stays left).
- **twentyhq/core-team-issues#2490** — Exit Settings control aligned to the workspace switcher (24px, matching padding/gap).
- **twentyhq/core-team-issues#2499** — 2px vertical gap restored between collapsible drawer section items.
- **twentyhq/core-team-issues#2491** — settings drawer rhythm now matches the main app (28px items, 2px gaps, 28px section headers).
- **twentyhq/core-team-issues#2492** — Home/Chat tab switch no longer flickers: both tab subtrees stay mounted (a shared `NavigationDrawerTabbedContent` toggles visibility instead of remounting + flashing the chat skeleton).

## Loading states
- **twentyhq/core-team-issues#2486** — metadata loading shows an empty body (no dense skeleton rows).
- **twentyhq/core-team-issues#2487** — settings table keeps its layout while loading, with the shimmer localized to the first row's first cell.

## Command menu & navigation
- **twentyhq/core-team-issues#2501** — navigation section header height matches the nav item rhythm (28px).
- **twentyhq/core-team-issues#2502 (part 1)** — the page side-panel toggle stays as the dots glyph while the command menu is open, instead of morphing into a second close control.

## New-field flow
- **twentyhq/core-team-issues#2494** — the new-field stepper moved from a breadcrumb dropdown into a centered secondary wizard bar (back chevron + Save on the configure step); breadcrumb stays clean and the object label is the centered title.

## Descoped (substantive bugs already fixed)
- **twentyhq/core-team-issues#2500** — command-menu highlight right gutter: the menu-item base measures full-width, so it's likely a scrollbar gutter on the list, not the shared component. Left for a focused follow-up.
- **twentyhq/core-team-issues#2502 part 2** — moving the command-menu close from left to right is cosmetic (the duplicate-control bug is fixed by part 1) and would touch the shared `SidePanelTopBar` used by search/AI panels.

## Verification
typecheck (tsgo) + oxlint + oxfmt green for all changed files; each change DOM-measured / screenshotted in the running app.


packages/twenty-front/src/modules/navigation/components/NavigationDrawerTabbedContent.tsx
```diff
@@ -0,0 +1,41 @@
...
+type NavigationDrawerTabbedContentProps = {
+  showAiChatContent: boolean;
+  navigationContent: ReactNode;
+};
+
+const StyledTabContent = styled.div<{ isHidden: boolean }>`
+  display: ${({ isHidden }) => (isHidden ? 'none' : 'contents')};
+`;
+
+// Both tabs stay mounted and toggle visibility so switching Home <-> Chat does
```
FelixMalfait: Remove comment

packages/twenty-front/src/modules/navigation/components/NavigationDrawerTabbedContent.tsx
```diff
@@ -0,0 +1,41 @@
...
+// not remount either subtree (which flashed the chat skeleton on every switch).
+// The chat tab mounts lazily on first open, then stays mounted.
+export const NavigationDrawerTabbedContent = ({
+  showAiChatContent,
+  navigationContent,
+}: NavigationDrawerTabbedContentProps) => {
+  const [hasOpenedAiChat, setHasOpenedAiChat] = useState(showAiChatContent);
+
+  useEffect(() => {
```
FelixMalfait: Do we need this useEffect?

packages/twenty-front/src/modules/ui/layout/page/components/PageCardHeader.tsx
```diff
@@ -51,6 +53,16 @@ const StyledTitle = styled.div`
   min-width: 0;
 `;
 
+// Title is centered against the full header width while the breadcrumb stays left-aligned.
```
FelixMalfait: Remove comment

packages/twenty-front/src/modules/navigation/components/NavigationDrawerTabbedContent.tsx
FelixMalfait: Done — removed the comment.

packages/twenty-front/src/modules/navigation/components/NavigationDrawerTabbedContent.tsx
FelixMalfait: Good call — removed it. `hasOpenedAiChat` is now derived during render (`if (showAiChatContent && !hasOpenedAiChat) setHasOpenedAiChat(true)`), which is React's documented alternative to an effect for adjusting state from a prop. Keeps the lazy-mount + keep-alive without the effect.

packages/twenty-front/src/modules/ui/layout/page/components/PageCardHeader.tsx
FelixMalfait: Done — removed.

packages/twenty-front/src/modules/navigation/components/NavigationDrawerTabbedContent.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { type ReactNode, useState } from 'react';
+
+type NavigationDrawerTabbedContentProps = {
+  showAiChatContent: boolean;
+  navigationContent: ReactNode;
+};
+
+const StyledTabContent = styled.div<{ isHidden: boolean }>`
+  display: ${({ isHidden }) => (isHidden ? 'none' : 'contents')};
+`;
```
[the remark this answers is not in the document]
FelixMalfait: Fixed by removing the dynamic `display`: `StyledTabContent` now has a static `display: contents` with a static `&[data-hidden='true'] { display: none }` rule and toggles via a `data-hidden` attribute. No CSS variable is used for `display`, so the inactive tab is correctly hidden (keep-alive behavior unchanged).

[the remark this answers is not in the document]
FelixMalfait: Correction to my earlier reply — this is a **false positive**. I tested it in Chrome: a rule `.x { display: var(--d) }` computes to `display: none` when `--d: none` and `display: contents` when `--d: contents`. CSS variables **do** work for the `display` property, so Linaria's compiled `display: var(...)` hides the inactive tab correctly. Reverted to the original dynamic value (4f04bb0389) — the static-CSS change wasn't needed.

PR #21403, opened 2026-06-10
Support morph relations in workflow record nodes

ijreilly:
## Support morph (polymorphic) relations in workflow record nodes

Morph relations (e.g. a polymorphic `Owner` on `Pet` targeting `Person` or `Company`) were not selectable in the workflow **Create / Update / Upsert Record** nodes. This PR adds full support for setting them.

### What changed

**Frontend**
- `shouldDisplayFormField`: allow `MORPH_RELATION` (many-to-one) so morph fields appear in record forms.
- New `FormMorphRelationToOneFieldInput`: a polymorphic record picker across the morph's target objects, storing a self-describing value `{ targetObjectMetadataId, id }`.
- Wired the morph branch into `FormFieldInput`.

**Backend**
- New `formatWorkflowRecordMorphRelationFields` util: resolves the form value (stored under the base field name, e.g. `owner`) into the correct per-target join column (`ownerCompanyId`), nulling siblings to keep exactly one target referenced.
- Wired into the create / update / upsert workflow actions (update also expands `fieldsToUpdate` to the concrete join columns).

### Permissions handling
- The picker's search is scoped to only the morph targets the user can read (`canReadObjectRecords`), so it no longer breaks when a target object is inaccessible.
- If an existing value points to an object the user can't read, the field shows the reused **"Not shared"** lock display instead of an empty field, while remaining editable when other targets are readable.

### Notes
- No data schema / migration changes — reuses the existing per-target morph columns and stores the selection in the existing workflow step JSON settings.

Screenshot 2026-06-10 at 14 57 40

Also handles the case where the selected record is not readable
image


packages/twenty-server/src/engine/core-modules/search/search.resolver.ts
```diff
@@ -15,9 +15,14 @@ import { AuthWorkspace } from 'src/engine/decorators/auth/auth-workspace.decorat
 import { CustomPermissionGuard } from 'src/engine/guards/custom-permission.guard';
 import { WorkspaceAuthGuard } from 'src/engine/guards/workspace-auth.guard';
 import { WorkspaceManyOrAllFlatEntityMapsCacheService } from 'src/engine/metadata-modules/flat-entity/services/workspace-many-or-all-flat-entity-maps-cache.service';
+import { PermissionsGraphqlApiExceptionFilter } from 'src/engine/metadata-modules/permissions/utils/permissions-graphql-api-exception.filter';
 
 @CoreResolver()
-@UseFilters(SearchApiExceptionFilter, PreventNestToAutoLogGraphqlErrorsFilter)
+@UseFilters(
+  SearchApiExceptionFilter,
+  PermissionsGraphqlApiExceptionFilter,
```
ijreilly, line 23: so that permission errors in search do not end up at 500 anymore

packages/twenty-front/src/modules/object-record/record-field/ui/components/FormFieldInput.tsx
```diff
@@ -241,6 +246,15 @@ export const FormFieldInput = ({
...
       readonly={readonly}
     />
+  ) : isFieldMorphRelationManyToOne(field) ? (
+    <FormMorphRelationToOneFieldInput
+      label={field.label}
+      morphRelations={field.metadata.morphRelations}
+      defaultValue={defaultValue as FormMorphRelationToOneValue}
+      onClear={onClear}
+      onChange={onChange}
+      readonly={readonly}
```
thomtrp, line 257: missing VariablePicker, I guess we need variables for those fields?

packages/twenty-front/src/modules/object-record/record-field/ui/form-types/components/FormMorphRelationToOneFieldInput.tsx
```diff
@@ -0,0 +1,252 @@
...
+  label?: string;
+  morphRelations: FieldMetadataItemRelation[];
+  defaultValue?: FormMorphRelationToOneValue;
+  onChange: (value: JsonValue) => void;
+  onClear?: () => void;
+  readonly?: boolean;
+  testId?: string;
+};
+
+export const FormMorphRelationToOneFieldInput = ({
```
thomtrp, line 55: I don't understand why that component is so complicated compared to FormRelationToOneFieldInput. I think we should rely more on FormSingleRecordPicker and extend it if needed. Unless I am mistaken and this is not possible?

packages/twenty-server/src/modules/workflow/workflow-executor/utils/format-workflow-record-morph-relation-fields.util.ts
```diff
@@ -0,0 +1,139 @@
...
+    return {
+      targetObjectMetadataId: record.targetObjectMetadataId,
+      id: record.id,
+    };
+  }
+
+  return null;
+};
+
+export const formatWorkflowRecordMorphRelationFields = (
```
thomtrp: I think we probably only need one util, formatWorkflowRecordRelationFields. We call both utils together in actions. We should probably merge and test

ijreilly:
new changes

**Morph relation input refactor** (FormMorphRelationToOneFieldInput.tsx, FormFieldInput.tsx, FormSingleRecordPicker.tsx): the morph field now reuses the shared FormSingleRecordPicker instead of a hand-rolled dropdown, supports workflow variables, and handles unreadable records ("Not shared") and the no-readable-objects case. Also includes my two UI fixes: readonly box now fills full height, and the "Not shared" pill gets proper left inset.

**Clear/remove fix** (WorkflowEditActionCreateRecord/UpdateRecord/UpsertRecord.tsx): wired onClear into the form fields so the chip's cross actually removes the selected record/variable.

**Variable-picker scoping** (VariablePickerComponent.ts, WorkflowVariablePicker.tsx, WorkflowVariablesDropdown.tsx, WorkflowVariablesDropdownStepItems.tsx): added an optional objectNameSingularsToSelect filter so the picker only lets you pick records of the expected object type(s).

**Server side** (format-workflow-record-morph-relation-fields.util.ts): formatting logic for morph relation fields when executing workflow record actions.

packages/twenty-front/src/modules/object-record/record-field/ui/form-types/components/FormMorphRelationToOneFieldInput.tsx
ijreilly: right, refactored this

packages/twenty-server/src/modules/workflow/workflow-executor/utils/format-workflow-record-morph-relation-fields.util.ts
ijreilly: which other util should we remove ?

thomtrp: We should merge formatWorkflowRecordRelationFields and formatWorkflowRecordRelationFields together if possible. I don't see why we need both separated. We call both in a row anyway. And we should add a test on the final util

packages/twenty-front/src/modules/object-record/record-field/ui/form-types/types/VariablePickerComponent.ts
```diff
@@ -5,4 +5,7 @@ export type VariablePickerComponent = React.FC<{
   onVariableSelect: (variableName: string) => void;
   shouldDisplayRecordObjects?: boolean;
   shouldDisplayRecordFields?: boolean;
+  // When provided, only records belonging to these objects can be picked.
+  // Used by relation fields to restrict the picker to the expected target objects.
+  objectNameSingularsToSelect?: string[];
```
thomtrp, line 8: don't think we need that comment, this is clear enough

packages/twenty-server/src/modules/workflow/workflow-executor/workflow-actions/record-crud/update-record.workflow-action.ts
```diff
@@ -82,18 +83,32 @@ export class UpdateRecordWorkflowAction implements WorkflowAction {
...
+
     const filteredObjectRecord = filterValidFieldsInRecord(
       formattedObjectRecord,
       objectMetadataInfo.flatObjectMetadata,
       objectMetadataInfo.flatFieldMetadataMaps,
     );
 
-    const filteredFieldsToUpdate = workflowActionInput.fieldsToUpdate?.filter(
+    // A morph relation field selected for update maps to several per-target join
+    // columns, so expand the user-facing field name into the concrete columns.
```
thomtrp: please remove comment

thomtrp (approved):
LGTM, except for utils (`formatWorkflowRecordRelationFields` - `formatWorkflowRecordRelationFields`) that should be merged together + comments to remove.

Great work 🔥

PR #1305, opened 2023-08-25
Add hotkeys to modals

Weiko:
## Context
Adding hotkeys parameters to modals, for example Esc key will now close the modal

## Test

https://github.com/twentyhq/twenty/assets/1834158/767d3c6b-d2a4-491c-b758-612d8902d924

PR #1304, opened 2023-08-25
Scroll behavior part 2

charlesBochet:
This complete the work on scrollbars:
1)  simplify the API: 
`import { ScrollWrapper } from '@/ui/utilities/scroll/components/ScrollWrapper';`

```
 <ScrollWrapper>
...
</ScrollWrapper>
```

```
const StyledCustomScrollWrapper = styled(ScrollWrapper)`
  flex-direction: column;
`;

```
You can also customize it using styled component if needed:

2) I've made sure to have no undesirable re-renders.

3) Apply it on Show pages and fix issue on Table Layout

https://github.com/twentyhq/twenty/assets/12035771/1d8a5ed7-70f6-4205-8d3d-d98699e60bb9

PR #1287, opened 2023-08-22
remove arrows for number inputs

srikarsams:
Closes #1001

Issue #1278, opened 2023-08-18
New timeline design


Bonapara:
The new timeline will offer a more compact design

Elements that should be logged into the new timeline:
- A note is created
- A note is edited

- A task is created
- A task is completed

- An opportunity is created
- An opportunity has been moved to another step
- An opportunity fields have been edited (Toggle to show which fields)

- An email has been received

- A person has been associated to a company record

More logs coming soon.

![Image](https://github.com/twentyhq/twenty/assets/19412894/898d4ffb-a04a-4c85-b244-2b7dd97638e4)

P.S: Have a look on how the timeline is made on Figma, I only used Flexbox to make it easy to customize.

![Image](https://github.com/twentyhq/twenty/assets/19412894/f1778147-089c-4371-806d-9f58f8593a36)

https://www.figma.com/file/xt8O9mFeLl46C5InWwoMrN/Twenty?type=design&node-id=6163-81155&mode=design&t=Og4rTLwXiXSccbH4-4

abhip167:
I am happy to take this.

charlesBochet:
Putting this on-hold as we are waiting for the bases of the new backend to be there

FelixMalfait:
Closing as there is a similar, more detailed, issue + Brendan has done the frontend work

PR #21409, opened 2026-06-10
security: clear 8 Dependabot alerts via transitive/parent bumps (no resolutions)

charlesBochet:
Clears 8 Dependabot alerts via in-range transitive/parent bumps and one dead-dependency removal. **No `resolutions` overrides** were used — every fix is a real version bump within existing semver ranges or a parent upgrade.

### Root `yarn.lock`
- **react-router** 6.30.3 → 6.30.4 (open redirect via protocol-relative URL) — pulled through react-router-dom, ranges unchanged — alert #1382
- **yaml** 2.8.1 → 2.9.0 (stack overflow on deeply nested collections) — alert #734
- **uuid** `^13.0.0` → 13.0.2 in twenty-sdk + create-twenty-app (buffer bounds check) — alert #1164
- **ip-address** `^9.0.5` dropped by bumping **socks** 2.8.3 → 2.8.9 (now depends on `ip-address ^10.1.1`, which is unaffected) — alert #1171

### `seed-dependencies` lockfile
- **uuid** `^10.0.0` → `^11.1.1` (direct dep; removed now-redundant `@types/uuid` since uuid v11 ships its own types) — alert #1287
- **ip-address** `^9.0.5` dropped via the same socks bump — alert #1170

### `twenty-for-twenty` lockfile
- **resend** bumped to 6.12.4 (`^6.12.0` range kept), which drops its `svix@1.90.0 → uuid@^10` transitive chain — alert #1278

### `twenty-companion`
- Removed the unused **simplemde** dependency. The note editor loads SimpleMDE from a CDN `<script>` tag and never imports the npm package; `easymde` (its maintained fork) is already a dependency — alert #690

### Not addressed here
The remaining alerts can't be closed without `resolutions` overrides (deliberately avoided in this PR) or a larger migration:
- **qs** (#1305, #1304), **lodash** (#824 high / #823 / #385), **ws** (#1238), **postcss** (#1061) — vulnerable copies are pinned exact / bundled by parents (express, body-parser, @nestjs/*, next, styled-components, zapier) with no in-range patch.
- **webpack-dev-server** (#1237/#692/#691) — pinned by `@electron-forge/plugin-webpack` (still on v4); dev-tooling only.
- **uuid <11.1.1** (#1289) — spread across `^3`/`^8`/`^9` transitive ranges; reaching v11 is a breaking jump.
- **apollo-server-core** (#735/#736) — requires an Apollo Server 3 → 4 migration.

### DIFF
--- a/packages/create-twenty-app/package.json
+++ b/packages/create-twenty-app/package.json
@@ -35,7 +35,7 @@
     "lodash.kebabcase": "^4.1.1",
     "lodash.startcase": "^4.4.0",
     "twenty-sdk": "workspace:*",
-    "uuid": "^13.0.0"
+    "uuid": "^13.0.2"
   },
   "devDependencies": {
     "@swc/core": "^1.15.11",
--- a/packages/twenty-apps/internal/twenty-for-twenty/yarn.lock
+++ b/packages/twenty-apps/internal/twenty-for-twenty/yarn.lock
@@ -2838,17 +2838,17 @@ __metadata:
   linkType: hard
 
 "resend@npm:^6.12.0":
-  version: 6.12.0
-  resolution: "resend@npm:6.12.0"
+  version: 6.12.4
+  resolution: "resend@npm:6.12.4"
   dependencies:
     postal-mime: "npm:2.7.4"
-    svix: "npm:1.90.0"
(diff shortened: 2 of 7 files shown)


bosiraphael (approved):
LGTM

PR #824, opened 2023-07-22
Fix typos/language in docs

willkrakow:

PR #823, opened 2023-07-22
[807] fix: people page icon

js-ashwin:
## Changes
changed icon for people page.
Closes #807 

Screenshot 2023-07-22 at 4 26 31 PM

PR #21411, opened 2026-06-10
security: force lodash 4.18.1 to close code-injection CVE (#824) + prototype pollution

charlesBochet:
Adds a single `lodash` → **4.18.1** resolution to close the remaining lodash CVEs.

### Why a resolution
The dependency *union* already resolved to 4.18.1, but vulnerable copies remained because parents pin older versions that have no in-range patch:
- `4.17.21` pinned exact by `@nestjs/config@3.3.0`, `zapier-platform-*`, and `verdaccio`
- `4.17.23` via `~4.17.x` ranges (`@graphql-tools`, socket.io)

4.18.0 is the first patched release, so the only way to force the parent-pinned copies up is an override. lodash 4.18.1 is an API-compatible security release; after the resolution **every** `lodash` copy resolves to 4.18.1 and none below it remains.

This does **not** touch the separate `lodash.*` single-function packages (`lodash.merge`, `lodash.camelcase`, …), which are unaffected by these advisories.

### Closes
- **#824 — `_.template` code injection (HIGH)**
- #823 — prototype pollution in `_.unset` / `_.omit`
- #385 — prototype pollution in `_.unset` / `_.omit`

Verified with `yarn install --immutable` (passes the hardened-mode 3-day age gate; 4.18.1 was published 2026-04-01).

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "graphql": "16.8.1",
     "type-fest": "4.10.1",
     "typescript": "5.9.3",
+    "lodash": "4.18.1",
     "nodemailer": "8.0.10",
     "graphql-redis-subscriptions/ioredis": "^5.6.0",
     "@lingui/core": "5.1.2",
--- a/yarn.lock
+++ b/yarn.lock
@@ -41843,24 +41843,10 @@ __metadata:
   languageName: node
   linkType: hard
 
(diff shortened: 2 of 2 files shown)


charlesBochet:
Superseded by #21414, which closes the same lodash alerts (#824/#823/#385) by upgrading the parent packages (graphql-codegen 5→6, zapier-platform-core 15→19, spectral-functions) instead of using a `resolutions` override. Closing.

PR #21416, opened 2026-06-10
fix(server): register Lingui message compiler to stop "Uncompiled message detected" log flood

### LINKED ISSUE #21415: Server floods logs with "Uncompiled message detected" from Lingui on every t`…` call
## Bug Description

On a running `twenty-server`, the logs are continuously flooded with Lingui warnings:

```
Uncompiled message detected! Message:
> Could not find flat entity in maps
Please compile your catalog first.
That means you use raw catalog or your catalog doesn't have a translation for the message and fallback was used.
ICU features such as interpolation and plurals will not work properly for that message.
```

This fires on essentially every server-side `` t`…` `` call (exception messages, etc.). On a hosted deployment it was severe enough to saturate the platform's log rate limit (hundreds/sec, thousands of messages dropped per second).

Root cause: server-side `` t`…` `` macro calls resolve against Lingui's **global `i18n` singleton**, but `I18nService` only ever configures its own per-locale instances (`setupI18n()` in `i18nInstancesMap`) — it never loads a catalog into, or registers a messages compiler on, the global singleton. When `@lingui/core`'s `_()` encounters a message string with **no `_messageCompiler` set**, it logs "Uncompiled message detected" and falls back to the raw string. The warning is emitted **unconditionally** — it is not gated by `NODE_ENV` (verified in `@lingui/core` 5.1.2).

## Expected behavior

No "Uncompiled message detected" spam in server logs, and ICU interpolation/plurals work on these messages.

## Technical inputs

- File: `packages/twenty-server/src/engine/core-modules/i18n/i18n.service.ts`
- Register `compileMessage` via `setMessagesCompiler` on the global `i18n` (used by the `t` macro) and on each per-locale instance:

```ts
import { i18n, setupI18n /* … */ } from '@lingui/core';
import { compileMessage } from '@lingui/message-utils/compileMessage';
// …
i18n.setMessagesCompiler(compileMessage);            // global, used by t`…`
// and per instance:
localeI18n.setMessagesCompiler(compileMessage);
```

- `@lingui/message-utils` is already a transitive **runtime** dependency of `@lingui/core` (`@lingui/core@5.1.2 → @lingui/message-utils@^5.1.2`), so no new dependency is introduced.
- PR ready to submit alongside this issue.

machinagod:
## Summary

Fixes #21415.

Server-side `` t`…` `` macro calls (e.g. `FlatEntityMapsException`) resolve against Lingui's global `i18n` singleton, which `I18nService` never loads a catalog into and never registers a messages compiler on. With no `_messageCompiler` set, every such lookup logs `Uncompiled message detected!` and falls back to the raw string — flooding server logs (enough to hit hosting log rate limits) and disabling ICU interpolation on those messages. The warning is emitted unconditionally by `@lingui/core` (not `NODE_ENV`-gated).

## Changes

Register `compileMessage` via `setMessagesCompiler` in `I18nService.loadTranslations()`:
- on the global `i18n` singleton (used by the `t` macro), and
- on each per-locale instance.

`@lingui/message-utils` is already a transitive **runtime** dependency of `@lingui/core` (`@lingui/core@5.1.2 → @lingui/message-utils@^5.1.2`), so no new dependency is added.

## Testing

Deployed on a real instance: before, the server emitted hundreds of `Uncompiled message detected` lines/sec (saturating the log rate limit); after, steady-state shows `0`, and the worker (which shares the code path) likewise shows `0`. App behaviour is unchanged — messages already rendered via fallback; this silences the warning and enables ICU on the affected messages.


FelixMalfait (approved):

FelixMalfait:
Thank you!

PR #736, opened 2023-07-18
Add readme pictures

Bonapara:
In order to update the readme, I'm first adding picture materials to the repository

PR #21418, opened 2026-06-10
security: drop end-of-life apollo-server-core (#735, #736)

charlesBochet:
Closes the `apollo-server-core` alerts (**#735**, **#736**) by **removing the dependency** — no Apollo migration, no resolution.

### Why these were flagged "no patch available"
`apollo-server-core` is **Apollo Server v3, which is end-of-life** (per its npm deprecation notice). No patched release of this package will ever exist — the CVE fix lives only in the renamed `@apollo/server` v4 package.

### Why we can just drop it
twenty-server **doesn't use Apollo Server** — its GraphQL runtime is **GraphQL Yoga** (`YogaDriver`). `apollo-server-core` was imported for one thing only: the `gql` template tag in **6 integration test files**. `gql` from `graphql-tag` is identical (apollo-server-core merely re-exports it), `graphql-tag` is **already a direct dependency**, and **15 other twenty-server tests already import `gql` from it**.

### Change
- Swapped `import { gql } from 'apollo-server-core'` → `import { gql } from 'graphql-tag'` in the 6 test files.
- Removed `apollo-server-core` from `packages/twenty-server/package.json`.
- Result: `apollo-server-core` (and its transitive surface) is gone from `yarn.lock` entirely.

### Verification
- `yarn install --immutable` ✓
- No `apollo-server-core` references remain in source or lockfile
- Integration tests (which exercise the swapped `gql` imports) run in CI

### DIFF
--- a/packages/twenty-server/package.json
+++ b/packages/twenty-server/package.json
@@ -77,7 +77,6 @@
     "@sniptt/guards": "0.2.0",
     "addressparser": "1.0.1",
     "ai": "6.0.97",
-    "apollo-server-core": "3.13.0",
     "archiver": "7.0.1",
     "axios": "^1.16.0",
     "axios-retry": "^4.5.0",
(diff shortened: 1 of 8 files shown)


Weiko (approved):
LGTM

PR #21422, opened 2026-06-10
feat(workflow) - Add validation layer

etiennejouan:
Add workflow validation framework and consolidate output schema types/search logic into twenty-shared

This PR introduces a comprehensive workflow validation system that catches configuration errors at build-time, and consolidates the fragmented output-schema type definitions and variable-search logic from the front-end into twenty-shared

**Workflow validation** — A new system that checks workflows for errors before activation: graph connectivity (unreachable steps, dangling references), step parameter schemas (via Zod), variable references (typos, wrong step order), and workspace metadata (non-existent objects). Returns structured errors/warnings with "did you mean?" suggestions. Runs automatically after create_complete_workflow and update_workflow_version_step, and is also available as a standalone validate_workflow tool.

**Output schema consolidation** — Moves all output schema types and the variable-search logic from scattered front-end files into twenty-shared, replacing ~800 lines of duplicated per-schema-type code with a single unified searchVariableInOutputSchema dispatcher.

To do : 
- validation on CODE and AGENT step


packages/twenty-shared/src/workflow/validation/utils/__tests__/validate-workflow-variable-references.util.test.ts
```diff
@@ -0,0 +1,147 @@
+import {
```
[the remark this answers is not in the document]
etiennejouan: Good catch — added a regression test ("should not flag a self-reference as not-upstream") that references a step's own output via `{{s1.name}}` and asserts `VARIABLE_NOT_UPSTREAM` is not emitted, covering the `isSelfReference` branch.

packages/twenty-shared/src/workflow/validation/utils/__tests__/validate-workflow-variable-references.util.test.ts
```diff
@@ -0,0 +1,147 @@
...
+          type: 'CODE',
+          settings: { input: { value: '{{s1.unknownField}}' } },
+        },
+      ],
+    };
+
+    expect(getCodes(workflow)).toContain('VARIABLE_PATH_NOT_FOUND');
+  });
+
+  it('should resolve a valid trigger reference against the trigger output schema', () => {
```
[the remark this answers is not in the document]
etiennejouan: Added a test ("should flag an invalid path in the trigger output schema") that references `{{trigger.unknownField}}` against the trigger output schema and asserts `VARIABLE_PATH_NOT_FOUND`, balancing the existing valid-trigger-path coverage.

packages/twenty-shared/src/workflow/workflow-schema/utils/__tests__/resolve-variable-path-in-output-schema.test.ts
```diff
@@ -0,0 +1,206 @@
...
+    it('should resolve a field under "first"', () => {
+      expect(
+        resolveVariablePathInOutputSchema({
+          schema: findRecordsSchema,
+          propertyPath: ['first', 'title'],
+        }).found,
+      ).toBe(true);
+    });
+
+    it('should resolve the terminal "first" node', () => {
```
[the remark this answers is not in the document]
etiennejouan: I don't think this is a duplicate. The two tests live in different `describe` blocks and exercise different schemas:

- `find records schema` › "should resolve the terminal \"first\" node" uses `findRecordsSchema` (where `first` is a non-leaf `RECORD` node) with path `['first']`, and asserts the resolved label is `'First Task'`. This is a regression test for the Find-Records terminal-`first` fix.
- `code step schema that mimics find records keys` › "should resolve the terminal \"first\" leaf" uses `codeStepSchema` (where `first` is a plain leaf) to verify the generic-map path, ensuring a CODE step that happens to have a `first` key isn't misrouted through Find-Records logic.

The other test in the same block ("should resolve a field under \"first\"") uses the nested path `['first','title']`, not the terminal `['first']`. Keeping the terminal-node test as-is.

packages/twenty-shared/src/workflow/validation/utils/get-step-outgoing-step-ids.util.ts
```diff
@@ -0,0 +1,53 @@
...
+): string[] => {
+  const outgoingStepIds = new Set<string>(step.nextStepIds ?? []);
+
+  const input = getStepInput(step);
+
+  if (!isDefined(input)) {
+    return [...outgoingStepIds];
+  }
+
+  if (step.type === WORKFLOW_VALIDATION_STEP_TYPES.IF_ELSE) {
```
thomtrp: you need a real guard here. Check isIfElseAction

packages/twenty-shared/src/workflow/validation/utils/get-step-outgoing-step-ids.util.ts
```diff
@@ -0,0 +1,53 @@
...
+    const branches = (input as Partial<IfElseStepInput>).branches ?? [];
```
thomtrp: you won't need that casting if you have a guard

packages/twenty-shared/src/workflow/validation/utils/get-step-outgoing-step-ids.util.ts
```diff
@@ -0,0 +1,53 @@
...
+    for (const branch of branches) {
+      for (const nextStepId of branch?.nextStepIds ?? []) {
+        outgoingStepIds.add(nextStepId);
+      }
+    }
+  }
+
+  if (step.type === WORKFLOW_VALIDATION_STEP_TYPES.ITERATOR) {
+    const initialLoopStepIds =
+      (input as Partial<IteratorStepInput>).initialLoopStepIds ?? [];
```
thomtrp: same. Iterator guard and no casting

packages/twenty-shared/src/workflow/validation/utils/get-step-outgoing-step-ids.util.ts
```diff
@@ -0,0 +1,53 @@
+import { isDefined } from '@/utils';
+import { WORKFLOW_VALIDATION_STEP_TYPES } from '@/workflow/validation/constants/workflow-validation-step-types';
+import {
+  type IfElseStepInput,
+  type IteratorStepInput,
+  type ValidatableWorkflowStep,
+} from '@/workflow/validation/types/workflow-validation.type';
```
thomtrp: we already have WorkflowAction enum as a source of truth. Curious why we need new types?

packages/twenty-shared/src/workflow/workflow-schema/utils/collect-output-schema-paths.ts
```diff
@@ -0,0 +1,30 @@
+import { isObject } from '@sniptt/guards';
+
+import { type BaseOutputSchemaV2 } from '@/workflow/workflow-schema/types/base-output-schema.type';
+
+export const collectOutputSchemaPaths = (
```
thomtrp, line 5: nit: it seems that file names should use kebab cases. But i think that was there before

thomtrp (approved):
LGTM ✅

PR #21430, opened 2026-06-11
fix(twenty-shared): promote formatToShortNumber to next unit on rounding carry

devteamaegis:
## What's broken

`formatToShortNumber` returns the wrong unit at magnitude boundaries:

```
999999     => "1000k"   (expected "1m")
999999999  => "1000m"   (expected "1b")
999.95     => "1000"     (expected "1k")
```

It picks the unit (k/m/b) from the unrounded value, then `.toFixed(1)` rounds the quotient up to `1000.0`, which is never promoted to the next unit. This surfaces on aggregate numeric displays (record counts, currency sums).

## Why it happens

Unit selection uses the unrounded amount (`< 1_000_000` → k-branch), but the displayed figure is `(amount/1000).toFixed(1)`, which rounds. `999999/1000 = 999.999` → `"1000.0"` → `"1000k"`. The rounding carry crosses the boundary after the unit is already chosen.

## Fix

Pick the unit from a threshold table, round once, and if the rounded quotient is `1000.0`, emit `1` of the next unit up.

## Test

Added a case asserting `999999 → "1m"`, `999999999 → "1b"`, `999.95 → "1k"`, `-999999 → "-1m"`. All 11 existing cases still pass.

### DIFF
--- a/packages/twenty-shared/src/utils/__tests__/formatToShortNumber.test.ts
+++ b/packages/twenty-shared/src/utils/__tests__/formatToShortNumber.test.ts
@@ -30,6 +30,13 @@ describe('formatToShortNumber', () => {
     expect(formatToShortNumber(0)).toBe('0');
   });
 
+  it('promotes to the next unit when rounding carries over to 1000', () => {
+    expect(formatToShortNumber(999999)).toBe('1m');
+    expect(formatToShortNumber(999999999)).toBe('1b');
+    expect(formatToShortNumber(999.95)).toBe('1k');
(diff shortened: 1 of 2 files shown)


etiennejouan:
Hi @devteamaegis, just find your PR while I've just merged an other [one](https://github.com/twentyhq/twenty/pull/21591) fixing same issue. Yours is older, should have priority review. Sorry for that.
Thanks for your contribution. For coming contributions, please address opened issues, it's simpler for reviewer to then spot the oldest PR.

PR #21431, opened 2026-06-11
fix(smtp): respect user encryption settings for nodemailer transport

### LINKED ISSUE #21300: IMAP/SMTP connection: SMTP transport ignores the Encryption (SSL/TLS) setting
Connecting a custom mail account via **Settings → Accounts → Connect via IMAP/SMTP**: the **SMTP Encryption** selection ("SSL/TLS" vs "None") has no effect on the SMTP connection. The IMAP side honors the equivalent setting; the SMTP side does not.

### Where

In `packages/twenty-server/src/engine/core-modules/imap-smtp-caldav-connection/services/imap-smtp-caldav-connection.service.ts`:

- `testImapConnection` builds the IMAP client with `secure: params.secure ?? true`.
- `testSmtpConnection` builds the nodemailer transport with `{ host, port, auth, tls }` — **no `secure` field**.

With `secure` omitted, nodemailer derives it from the port (`true` only for `465`). So an **implicit-TLS** SMTP server on any non-465 port, with "SSL/TLS" selected in the UI, is contacted in **plaintext** and fails:

```
SMTP connection failed: Greeting never received
```

The identical host/credentials over IMAP (with "SSL/TLS") connect fine, which makes the inconsistency clear: the SMTP encryption dropdown is effectively ignored.

### Suggested fix

Pass the chosen encryption through to nodemailer, mirroring the IMAP path, e.g.:

```ts
secure: params.secure ?? (port === 465)
```

### Environment

- Twenty v2.9.2, self-hosted (Docker)

dankgarlic1:
The testSmtpConnection function was failing to pass the secure parameter from the UI down to the nodemailer.createTransport configuration object. Because the parameter was dropped, Nodemailer defaulted to secure: false for port 587, ignoring the user's explicit request for SSL/TLS encryption.

This PR mirrors the logic found in testImapConnection and injects the secure toggle directly into the transport configuration, using a nullish coalescing operator to respect the user's choice or safely fall back to checking if the port is 465

ref: https://nodemailer.com/smtp#general-options

closes #21300

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/imap-smtp-caldav-connection/services/imap-smtp-caldav-connection.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/imap-smtp-caldav-connection/services/imap-smtp-caldav-connection.service.ts
@@ -114,6 +114,7 @@ export class ImapSmtpCaldavService {
     const transport = createTransport({
       host: validatedHost,
       port: params.port,
+      secure: params.secure ?? params.port === 465,
       auth: {
         user: params.username ?? handle,
         pass: params.password,


neo773:
This is not the right approach, the issue is a little more complicated than that this will break existing production accounts.

PR #21228, opened 2026-06-04
isCustom deprecation for Objects and Fields

Weiko:
## Context

`isCustom` was a legacy denormalized boolean on `ObjectMetadataEntity` and `FieldMetadataEntity`. 
Now that every metadata row carries `applicationId` (via `SyncableEntity`), "is this custom" is fully derivable, and the stored
boolean was a redundant second source of truth that could drift.

The real meaning of `isCustom` is **"the owning application is not the twenty-standard application"** — i.e. `!belongsToTwentyStandardApp`. 
Note this is *not* "belongs to the workspace custom app" as I initially thought: third-party-application
objects/fields are custom too. 
The standard application has a globally stable `universalIdentifier`, so the value derives with no per-workspace lookup.

## `isCustom` checks — before → after

`isCustom` is no longer a stored column. The table below lists every site that branched on it and how it resolves now. The unifying rule: `isCustom ≡ !isTwentyStandardApplicationUniversalIdentifier(applicationUniversalIdentifier)`.

### Server — behavioural checks

| Location | Purpose | Before | Now |
|---|---|---|---|
| `utils/compute-object-target-table.util.ts` | Physical table name `_` prefix | `computeTableName(nameSingular, objectMetadata.isCustom)` | derives from `applicationUniversalIdentifier` (single source for all table-name callers) |
| `twenty-orm/factories/entity-schema.factory.ts` + `…/entity-schema-metadata.type.ts` | ORM table name (hot path) | `object.isCustom` | `object.applicationId !== standardApplicationId` (computed in `buildEntitySchemaMetadataMaps`) |
| `twenty-orm/repository/workspace-{delete,soft-delete,update}-query-builder.ts` | Table name for mutations | `computeTableName(nameSingular, objectMetadata.isCustom)` | `computeObjectTargetTable(objectMetadata)` |
| `index-metadata/utils/generate-deterministic-index-name-v2.ts` | Index name hash (must stay bit-identical) | `flatObjectMetadata.isCustom` | derives from `applicationUniversalIdentifier` |
| `object-metadata/object-record-count.service.ts` | Table name for record count | `computeTableName(nameSingular, isCustom)` | `computeObjectTargetTable(flatObjectMetadata)` |
| `workspace-manager/dev-seeder/data/services/dev-seeder-data.service.ts` | Match seed config by table name | `computeTableName(item.nameSingular, item.isCustom)` | `computeObjectTargetTable(item)` |
[...]

PR #21432, opened 2026-06-11
fix: restore isCustom gate in metadata label resolvers

FelixMalfait:
## Context

#21228 removed the stored `isCustom` column and, with it, the `isCustom` early-return in `resolveObjectMetadataStandardOverride` / `resolveFieldMetadataStandardOverride`, on the assumption that falling through the `standardOverrides` checks was equivalent.

It isn't: custom object/field labels now reach the Lingui lookup. A custom label that collides with a standard catalog string (e.g. a custom field labeled "Status") gets translated for non-English locales against the user's intent, and every other custom label pays a hash + catalog miss — and, in production, an "Uncompiled message detected" warning (#21415) — on each metadata resolution.

## Fix

Restore the gate. `isCustom` is no longer stored, so call sites that build the resolver input from flat entities (dataloader, minimal-metadata, view controller, command-menu-item navigation context) derive it via `belongsToTwentyStandardApp`; GraphQL resolvers keep passing DTOs, which already carry the derived value.

## Testing

- Unit tests for both resolvers, including a new regression test: a custom label matching a standard catalog entry is returned verbatim, Lingui never called.

### DIFF
--- a/packages/twenty-server/src/engine/metadata-modules/field-metadata/utils/resolve-field-metadata-standard-override.util.ts
+++ b/packages/twenty-server/src/engine/metadata-modules/field-metadata/utils/resolve-field-metadata-standard-override.util.ts
@@ -10,13 +10,20 @@ import { type FieldMetadataDTO } from 'src/engine/metadata-modules/field-metadat
 export const resolveFieldMetadataStandardOverride = (
   fieldMetadata: Pick<
     FieldMetadataDTO,
-    'label' | 'description' | 'icon' | 'standardOverrides'
+    'label' | 'description' | 'icon' | 'isCustom' | 'standardOverrides'
   >,
   labelKey: 'label' | 'description' | 'icon',
   locale: keyof typeof APP_LOCALES | undefined,
   i18nInstance: I18n,
 ): string => {
   const safeLocale = locale ?? SOURCE_LOCALE;
+
+  // Custom field labels are user-authored: never overridden nor translated.
+  // Without this gate, a label colliding with a standard catalog string
+  // (e.g. "Status") would get translated against the user's intent.
+  if (fieldMetadata.isCustom) {
+    return fieldMetadata[labelKey] ?? '';
+  }
 
   if (labelKey === 'icon' && isDefined(fieldMetadata.standardOverrides?.icon)) {
     return fieldMetadata.standardOverrides.icon;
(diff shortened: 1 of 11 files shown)


FelixMalfait:
This does not feel like the right fix! Was made by AI. Will fix/review later

FelixMalfait:
Actually it does seem OK. @Weiko wdyt?

packages/twenty-server/src/engine/metadata-modules/field-metadata/utils/resolve-field-metadata-standard-override.util.ts
```diff
@@ -10,14 +10,21 @@ import { type FieldMetadataDTO } from 'src/engine/metadata-modules/field-metadat
...
   >,
   labelKey: 'label' | 'description' | 'icon',
   locale: keyof typeof APP_LOCALES | undefined,
   i18nInstance: I18n,
 ): string => {
   const safeLocale = locale ?? SOURCE_LOCALE;
 
+  // Custom field labels are user-authored: never overridden nor translated.
+  // Without this gate, a label colliding with a standard catalog string
+  // (e.g. "Status") would get translated against the user's intent.
```
Weiko, line 23: I'm surprised. My assumption was `isDefined(fieldMetadata.standardOverrides)` would not return true for custom fields so `if (fieldMetadata.isCustom)` would be redundant 🤔

FelixMalfait: Mmmh good point, will check

FelixMalfait: Checked — you're right that the override checks no-op for custom fields (`standardOverrides` is null there). But the gate isn't protecting those: it's protecting the step after them. The auto-translate fallback `i18nInstance._(generateMessageId(label))` runs for anything that falls through, custom included.

And custom labels collide with catalog ids more than you'd expect — re-hashing a few common ones against the compiled catalogs:

| custom label | resolves to (fr-FR) |
|---|---|
| Status | Statut |
| Name | Nom |
| Company | Entreprise |
| Amount | Montant |
| City | Ville |

So without the early return, a custom field the user literally named "Status" renders as "Statut" for French users. That's what the pre-#21228 gate was actually doing — the override checks were always implicitly safe — plus it skips a pointless catalog lookup per custom label per request.

Weiko: I didn't see the auto-fallback L58, right. I don't see why we have this tbh but I guess we didn't want to store the translations keys in every standard field 🤔. I think isCustom should not exist anymore now that we have 3rd party apps and even !isStandardApp should be kind of temporary in most cases but I'd rather have an explicit "belongsToStandardApp" somewhere than keeping !isCustom (because a 3rd party app is not custom and I'm not sure what we want to do about translations with those).
Anyway, LGTM otherwise!

Weiko (approved):
Ok to merge to unlock the deploy but I feel like we should get rid of isCustom and implement this differently!

FelixMalfait:
@Weiko I agree! It’s a big work stream feel free to take it 😅 https://github.com/twentyhq/core-team-issues/issues/2365

PR #21435, opened 2026-06-11
fix(server): prevent stale workspace cache writes from racing invalidation

FelixMalfait:
## Context

`example-app-postcard` fails intermittently on the schema integration test, each time missing a **different** subset of freshly synced fields ([run A](https://github.com/twentyhq/twenty/actions/runs/27330191961/job/80740567732) missing `recipient`, [run B](https://github.com/twentyhq/twenty/actions/runs/27333254362/job/80750773762) missing `content`) — even though the sync log shows the fields were created seconds earlier, and even when the assertion polls for 20s. The staleness is permanent, not transient.

## Root cause

`WorkspaceCacheService.recomputeDataFromProvider` has no ordering guard against `invalidateAndRecompute`. On a cache miss, a recompute reads the database; if a migration commits while that read is in flight, the runner's post-commit invalidation flushes and writes fresh data — and then the in-flight recompute lands **last**, `mset`ing its pre-commit snapshot over the fresh data with a brand-new random hash. Every subsequent hash validation agrees with the poisoned value, so the workspace serves stale metadata until the *next* invalidation. App sync runs several migrations back-to-back with reads interleaved (sync response resolution, API client generation), which is why CI hits the window regularly.

This is not just a test problem: any concurrent metadata read racing a schema change can make that change invisible until the next one.

## Fix

Per-key generation counters: `invalidateAndRecompute` bumps `<cacheKey>:gen` (post-commit) before flushing; `recomputeDataFromProvider` snapshots generations before reading the database and re-checks them after. If a generation moved, the result is served to the in-flight caller but never cached, and the compute retries once (the retry reads post-commit state, so callers and the memoizer end up with fresh data).

## Testing

- New regression test simulating the interleaving with an in-memory store: it reproduces the exact failure on the unfixed service (`stale-value` overwrites `fresh-value`) and passes with the guard.
- All 16 workspace-cache specs pass; typecheck and lint clean.

Replaces the earlier retry-the-assertion attempt on this branch — polling couldn't help because the poisoned cache never healed.

### DIFF
--- a/packages/twenty-server/src/engine/workspace-cache/services/__tests__/workspace-cache.service.spec.ts
+++ b/packages/twenty-server/src/engine/workspace-cache/services/__tests__/workspace-cache.service.spec.ts
@@ -47,6 +47,7 @@ describe('WorkspaceCacheService', () => {
             mget: jest.fn(),
             mset: jest.fn(),
             mdel: jest.fn(),
+            incrBy: jest.fn(),
           },
         },
         {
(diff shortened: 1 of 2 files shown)


FelixMalfait:
Do not merge, bad AI solution - need to find something else

charlesBochet:
Heads-up: I hit the same postcard CI signature while validating the typeorm upgrade and root-caused a **second, independent** mechanism — the missing-field repro I froze locally survives a cold server restart + `redis-cli FLUSHALL` with all 17 `fieldMetadata` rows intact in Postgres, so it can't be (only) a poisoned cache. The batched nestjs-query relation read itself truncates: `objects(first:50){fields(first:500)}` executes `SELECT DISTINCT ... WHERE objectMetadataId IN (...31 ids...) LIMIT 501` with no ORDER BY over 558 matching rows, so ~57 fields are dropped scan-order-arbitrarily. Fixed in #21455 (per-parent paging in the @ptc-org patch). The race this PR fixes looks real too — both are probably needed for postcard to be fully green, so these PRs are complementary, not competing.

charlesBochet:
Taking care of it, closing :)

PR #21438, opened 2026-06-11
security: postcss CVE via styled-components bump + next/postcss resolution (Dependabot alert 1061)

charlesBochet:
Closes the postcss Dependabot alert — [1061](https://github.com/twentyhq/twenty/security/dependabot/1061) — by **bumping one parent** and using a **scoped resolution** only where there's no other path.

postcss `< 8.5.10` (XSS via unescaped `</style>` in stringify output) was bundled by two parents:

### styled-components → bumped (no resolution)
`styled-components 6.4.2` (in-range for our `^6.1.0`; we were on a stale 6.3.12) **dropped its postcss dependency entirely**, so a plain `yarn up` removes the 8.4.49 copy — no override. (styled-components is only used in one component-renderer Storybook showcase; product UI is on Linaria.) Re-built `twenty-front-component-renderer` to confirm the bump.

### next → scoped resolution (no parent path)
`next` pins postcss `8.4.31` **exact on every stable release** — next@latest (16.2.9) still pins it; the fix (8.5.10) exists only in the **16.3.0 canary** (unreleased). So `next/postcss: 8.5.15` is the only mechanism, scoped to next and documented in the top-level `//resolutions` note with a removal trigger.

### Verification
- Every postcss now resolves `>= 8.5.14` (safe); `yarn install --immutable` ✓.
- postcss is build-time and **CI doesn't build twenty-website on regular PRs**, so verified locally: **twenty-website** (Next production build) ✓ and **twenty-front-component-renderer** ✓.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,9 +45,10 @@
     "@angular-devkit/core": "19.2.24",
     "yeoman-environment": "6.0.1",
     "@electron-forge/plugin-webpack/webpack-dev-server": "5.2.4",
-    "qs": "6.15.2"
+    "qs": "6.15.2",
+    "next/postcss": "8.5.15"
   },
-  "//resolutions": "qs is pinned to 6.15.2 (CVE) for verdaccio + @mintlify/previewing, which pin old express 4.22.x; remove when they upgrade express",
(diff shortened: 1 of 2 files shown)


bosiraphael (approved):
LGTM

PR #21449, opened 2026-06-11
[WIP] new CLI command for building relations

Bredo:
Make it easier for app developers defining relations using the Twenty SDK CLI.

Very much WIP, but the aim is that;
- A developer can run `yarn twenty add:relation`, and then be guided by the CLI to set up their relation
- Supports one-to-many (single) and one-to-many (polymorphic) - many-to-many is yet to be supported by the SDK
- It's effectively a "guided version" of `yarn twenty add:field` with type RELATION

Related to https://github.com/twentyhq/twenty/discussions/21318 and https://github.com/twentyhq/twenty/pull/21368

### DIFF
--- a/packages/twenty-sdk/src/cli/commands/dev/__tests__/add.spec.ts
+++ b/packages/twenty-sdk/src/cli/commands/dev/__tests__/add.spec.ts
@@ -1,0 +1,122 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { EntityAddCommand } from '../add';
+import inquirer from 'inquirer';
+import { writeFile } from 'node:fs/promises';
+import { pathExists, ensureDir } from '@/cli/utilities/file/fs-utils';
+import { buildManifest } from '@/cli/utilities/build/manifest/manifest-build';
+
+vi.mock('inquirer');
+vi.mock('node:fs/promises');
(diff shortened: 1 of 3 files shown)


Bredo:
When the command is run, the user should be picking;
1. Type (pick 1)
    1. One-to-many (single-to-single)
    2. One-to-many (single-to-polymorphic)
2. Base object
    1. Built in Twenty objects
    2. Current app’s defined objects
    3. Other (write UUID)
3. Related object(s) (if single-to-single: pick one; if single-to-polymorphic: pick any number greater than 1)
    1. Built in Twenty objects
    2. Current app’s defined objects
    3. Other (write UUID)
4. Field name on base object (default to the plural version of the related object if its single-to-single)
5. Field name on related object(s) (default to the singular version of the base object)

charlesBochet:
Hi @Bredo, nice to see you here!

Regarding the change, I'm not sure this is super useful regarding the use of AI today but why not :)

However, this PR is not reaching our quality standards (all PR reds), I'll close it but feel free to open one that's passing the test and meeting project quality guidelines

Bredo:
@charlesBochet totally agreed - for context, this was a part of [a discussion on Discord](https://discord.com/channels/1130383047699738754/1514591826768691270), and me being confused that #21368 merged but not released 😅

PR #21450, opened 2026-06-11
fix(billing) - enable upgrade if invoice already paid

etiennejouan:
Had an issue concerning a user with credits, then invoice automatically paid. Upgrade failed


thomtrp (approved):

packages/twenty-server/src/engine/core-modules/billing/stripe/services/stripe-invoice.service.ts
```diff
@@ -68,9 +68,22 @@ export class StripeInvoiceService {
     });
 
     await this.stripe.invoices.finalizeInvoice(invoice.id, {
-      auto_advance: false,
+      auto_advance: true,
```
Weiko, line 71: Not sure to understand the link between your pr description and the need to set auto advance 🤔

PR #21451, opened 2026-06-11
Fix missing datetime filter type

martmull:
Currently datetime fields are only typed to be filtered by string

Add a proper typing to match gql filters

## Before
image

## After
image

### DIFF
--- a/packages/twenty-apps/examples/postcard/src/__tests__/schema.integration-test.ts
+++ b/packages/twenty-apps/examples/postcard/src/__tests__/schema.integration-test.ts
@@ -46,6 +46,7 @@ describe('PostCard object', () => {
     const names = obj!.fields.edges.map(
       (e: { node: { name: string } }) => e.node.name,
     );
+    console.log('names', names);
     expect(names).toContain('name');
     expect(names).toContain('content');
     expect(names).toContain('status');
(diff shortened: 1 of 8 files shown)


thomtrp (approved):

packages/twenty-apps/examples/postcard/src/__tests__/schema.integration-test.ts
```diff
@@ -39,13 +39,14 @@ describe('PostCard object', () => {
...
     const obj = objects.edges
-      .map((e: { node: { nameSingular: string } }) => e.node)
+      .map((e) => e.node)
       .find((n: { nameSingular: string }) => n.nameSingular === 'postCard');
     expect(obj).toBeDefined();
 
     const names = obj!.fields.edges.map(
       (e: { node: { name: string } }) => e.node.name,
     );
+    console.log('names', names);
```
Weiko, line 49: @martmull console log

martmull: I could like to keep it this test is flaky I want to know why ok for you?

Weiko: We are on it with @charlesBochet

Weiko (commented):
Are we sure this does not break the datetime picker in the FE?

PR #21458, opened 2026-06-11
feat(workflow): idempotent stop + retry failed runs from failing step

thomtrp:
https://github.com/user-attachments/assets/5a25396f-8959-4bd8-93cb-1187559ffe5f

## Summary

Two workflow-run improvements, with all non-trivial logic isolated in pure, unit-tested utils.

### 1. Idempotent stop
`stopWorkflowRun` no longer throws when a run is already in a terminal status (`COMPLETED` / `FAILED` / `STOPPED`) or already `STOPPING`; it returns the run unchanged. This fixes:
- bulk stop aborting on the first non-stoppable run in a mixed/select-all selection,
- the click-vs-processing race on a single run (run finishes between click and mutation).

It also releases the cached not-started throttle slot when stopping a `NOT_STARTED` run (prevents counter drift), and ends runs with no `state` directly.

### 2. Retry a failed run from the failing step
New `retryWorkflowRun` mutation (same guards/passthrough as `stopWorkflowRun`). It resets the failed step(s) to `NOT_STARTED`, flips the run to `RUNNING`, and enqueues a `RunWorkflowJob` with the steps to re-execute; downstream execution and status computation are unchanged.

Logic lives in pure utils:
- `build-retry-step-infos.util.ts` - decides per failed step what to reset; delegates iterator-specific logic to `build-retry-iterator-step-infos.util.ts` (an iterator that failed mid-loop is restored to `RUNNING` with cursor preserved, an iterator that failed itself restarts its whole loop).
- `get-runnable-step-ids.util.ts` - reuses the executor's `shouldExecuteStep` to also resume branches that never started (avoids hangs), excluding loop-interior steps.

The service method only orchestrates; the job's status check is a race guard (retriability is enforced in the service before enqueue).

A "Retry" command menu item surfaces only for `FAILED` runs (`someEquals(selectedRecords, "status", "FAILED")`).

### 3. Keep the run diagram visible across regenerations
The run diagram is regenerated on every run state change, producing fresh nodes without the dimensions Reactflow had measured. Reactflow hides unmeasured nodes until it re-measures them, so the diagram could flicker and disappear when the last regeneration before going idle left nodes unmeasured (reproducible after retrying a failed run). The regenerated nodes now carry over the previously measured dimensions (by id) so they stay rendered.

## Test plan
- [x] Unit tests for both retry utils (9 cases: plain failed step, non-failed untouched, iterator mid-loop restore, iterator self-failure, frontier parent gating, entry steps, loop-interior exclusion, parallel branches)
- [x] `twenty-server` + `twenty-front` typecheck
- [x] `lint:diff-with-main` clean for both packages
- [x] Manual: retry a failed run repeatedly and confirm the diagram stays visible
- [ ] Manual: stop a COMPLETED/mixed selection (no error), retry a failed run and confirm it resumes from the failing step


packages/twenty-front/src/modules/workflow/workflow-diagram/components/WorkflowRunVisualizerEffect.tsx
```diff
@@ -123,13 +123,44 @@ export const WorkflowRunVisualizerEffect = ({
...
           stepInfos: workflowRunState.stepInfos,
         });
 
+      // Regenerating the diagram on each run state change produces fresh nodes
+      // without the dimensions Reactflow measured previously. Reactflow hides
+      // unmeasured nodes until it re-measures them, which makes the diagram
+      // flicker (and disappear if the last regeneration before going idle
+      // leaves nodes unmeasured). Carry over the measured dimensions so the
+      // nodes stay rendered across regenerations.
+      const previousNodesById = new Map(
```
thomtrp, line 133: Regenerating the diagram on each run state change produces fresh nodes without the dimensions Reactflow measured previously. Reactflow hides unmeasured nodes until it re-measures them, which makes the diagram flicker (and disappear if the last regeneration before going idle leaves nodes unmeasured). Carry over the measured dimensions so the nodes stay rendered across regenerations.

packages/twenty-server/src/modules/workflow/workflow-runner/utils/build-retry-iterator-step-infos.util.ts
```diff
@@ -0,0 +1,64 @@
...
+  if (failedMidLoop) {
+    return {
+      stepInfosToUpdate: {
+        [iteratorStep.id]: {
+          ...iteratorStepInfo,
+          status: StepStatus.RUNNING,
+          error: undefined,
+        },
+      },
+      stepIdsToRetry: [],
```
sviat-barbutsa, line 42: It's a good PR but I'd like to highlight this moment. 
I think the mid-loop iterator retry path needs one more service-level regression test.

In this branch, `buildRetryIteratorStepInfos` restores the failed iterator to `RUNNING` with its cursor preserved, but returns `stepIdsToRetry: []`. Then `retryWorkflowRun` merges that state, calls `getRunnableStepIds`, builds `stepIdsToRun`, and returns early when `stepIdsToRun.length === 0`.

The nuance is that `getRunnableStepIds` excludes loop-interior steps, and the iterator itself is no longer `NOT_STARTED`, and so a failed-mid-loop iterator can appear to produce no runnable step IDs. In that case the retry mutation would return the still-failed workflow run without persisting the restored iterator state or enqueueing a `RunWorkflowJob`.

Could you please, add some regression test around `retryWorkflowRun` for an iterator with `result.hasProcessedAllItems === false` and verify that retrying it enqueues work again, or otherwise moves the run back into execution? If the intended resume path is different, it may be worth making that explicit before returning early on an empty `stepIdsToRun`.

sviat-barbutsa (changes requested):

packages/twenty-front/src/modules/command-menu-item/engine-command/record/single-record/workflow-runs/hooks/useExecuteWorkflowRunBulkCommand.ts
```diff
@@ -0,0 +1,34 @@
...
+    recordGqlFields: { id: true },
+  });
+
+  const execute = async () => {
+    if (targetedRecordsRule.mode === 'selection') {
+      for (const selectedRecordId of targetedRecordsRule.selectedRecordIds) {
+        await action(selectedRecordId);
+      }
+    } else {
+      const records = await fetchAllRecordIds();
```
ijreilly, line 25: Im a bit confused by this being called in RetryWorkflowRunSingleRecordCommand but applying to a bunch  of records ?

packages/twenty-server/src/modules/workflow/workflow-runner/jobs/run-workflow.job.ts
```diff
@@ -131,6 +138,32 @@ export class RunWorkflowJob {
     });
   }
 
+  private async retryWorkflowExecution({
```
ijreilly, line 141: what happens if the version used for the workflow has been deleted in the meantime (if that is possible) ?

ijreilly (approved):
When a workflow is retried, should we create a new workflow run? or at least introduce a new status / column to indicate it was retried ? 

Also I wonder if this makes a case for persisting workflowRuns longer - or at least the failed ones (although that may convey false ideas around the successful workflows) to have a bigger window to replay failed workflows ?

packages/twenty-server/src/modules/workflow/workflow-runner/jobs/run-workflow.job.ts
thomtrp: versions cannot be deleted. And we copy the whole flow in the run anyway!

packages/twenty-front/src/modules/command-menu-item/engine-command/record/single-record/workflow-runs/hooks/useExecuteWorkflowRunBulkCommand.ts
thomtrp: SingleRecord means that it triggers on record one by one. You can still perform is on several records though. But you will perform a loop instead of sending an array of records as input

thomtrp:
> When a workflow is retried, should we create a new workflow run? or at least introduce a new status / column to indicate it was retried ?
> 
> Also I wonder if this makes a case for persisting workflowRuns longer - or at least the failed ones (although that may convey false ideas around the successful workflows) to have a bigger window to replay failed workflows ?

A new run seems a bit overkill to me because if you do it, you do not really need the previous run. But keeping the history could make sense. Checking how hard it would be!

thomtrp:
> > When a workflow is retried, should we create a new workflow run? or at least introduce a new status / column to indicate it was retried ?
> > Also I wonder if this makes a case for persisting workflowRuns longer - or at least the failed ones (although that may convey false ideas around the successful workflows) to have a bigger window to replay failed workflows ?
> 
> A new run seems a bit overkill to me because if you do it, you do not really need the previous run. But keeping the history could make sense. Checking how hard it would be!

checked with Thomas, no history needed!

PR #21462, opened 2026-06-11
Humanize AI tool labels

Bonapara:
## Summary

- Display AI tool names without underscores in the tools table, tool detail page, and chat tool activity labels.
- Keep raw tool identifiers untouched for routing, schema lookup, icons, and execution.
- Allow settings search to match both raw and humanized tool names.

before:
image

After: 
image

## Tests

- `npx jest --config packages/twenty-front/jest.config.mjs --runTestsByPath packages/twenty-front/src/modules/ai/utils/__tests__/getToolDisplayMessage.test.ts`
- `npx oxfmt --check packages/twenty-front/src/modules/ai/utils/formatToolDisplayName.ts packages/twenty-front/src/modules/ai/utils/getToolDisplayMessage.ts packages/twenty-front/src/modules/ai/utils/__tests__/getToolDisplayMessage.test.ts packages/twenty-front/src/pages/settings/ai/components/SettingsAgentToolsTab.tsx packages/twenty-front/src/pages/settings/ai/components/SettingsAgentToolsTable.tsx packages/twenty-front/src/pages/settings/ai/SettingsToolDetail.tsx`

## Notes

- `oxlint` could not run locally because the installed package is missing the native macOS binding `@oxlint/binding-darwin-x64`.


FelixMalfait:
I get the idea (and like it) but I'm not sure about the method. For context the model already name the loading states for the tool call, I'm not sure if it's something we want to leverage more or remove in the future. @etiennejouan curious to get your thoughts. If we do not go through the model route, I feel like the method to remove "_" is brittle and not something we would usually do. The best practise in that case would be to introduce a label field (as we have on objects, fields, etc)

Screenshot 2026-06-12 at 14 54 41

etiennejouan:
I like the idea of making mentions more user friendly. I'm also wondering how we want to handle localization, that might be the deciding factor. I'll check if the loadingMessage can handle localization and how it renders on the inner CRUD tools. If loadingMessage can eventually handle everything (then the "leverage more" path) , that's probably the simplest approach

FelixMalfait:
@etiennejouan yes loading message can handle localization good point! It shouldn't be an issue. I'm just wondering if there's a big impact on latency/cost/quality of output (does nesting the real parameters negatively impact tool calling?), but probably not

etiennejouan:
@Bonapara @FelixMalfait 

Quick try for a full agent-managed loading/completed messages. Not fully satisfied because of agent style, but can be improved via prompt description.

Screenshot 2026-06-16 at 11 48 40
Screenshot 2026-06-16 at 11 46 44

wdyt ?

FelixMalfait:
Is this standard for other harnesses like Claude? I like it (besides the !)

etiennejouan:
> Is this standard for other harnesses like Claude? I like it (besides the !)

Not sure it's standard. loading/completed messages are model-generated for custom step, for example in cursor or claude it's the generate/execute todo task.

We can keep this or update with localized label. Think I prefer total control on display (when testing, some object names were changed, sense was ok but not sure it's ok for data model education) + keep tool schema the more concise we can.

etiennejouan:
Will re-work it. Closing for the moment.

PR #21469, opened 2026-06-11
fix: resolve timeline crashes by cleaning orphan message association

### LINKED ISSUE #21425: Bug Report: TimelineThread.visibility crashes Email/All Messages tabs after email account re-addition
<h2>Summary</h2>
After deleting and re-adding IMAP email accounts in TwentyCRM (a common 
troubleshooting step when sync breaks after updates), the <strong>Email tab</strong> on individual People/Company records and the <strong>All Messages</strong>
 workspace view permanently crash with GraphQL errors. The Timeline tab 
continues to work correctly. This leaves users unable to access their 
message history through the primary messaging UI.
<p>This is caused by <strong>two separate source-code bugs</strong> in the messaging resolver layer that fail to handle orphaned data gracefully.</p>
<hr>
<h2>Environment</h2>

Field | Value
-- | --
Version | v2.11.2 (Docker self-hosted)
Database | PostgreSQL 16
Email provider | IMAP (5 accounts)
Workspaces | 2 (one active, one empty)

<hr>
<h2>Workarounds Applied (Self-Hosted)</h2>
<p>To restore functionality immediately, we applied two <strong>container-level patches</strong> (editing compiled JS in the running container):</p>
<ol><li value="1"><strong><code>format-threads.util.js</code></strong> — Added <code>?? 'METADATA'</code> fallback for <code>visibility</code></li><li value="2"><strong><code>apply-messages-visibility-restrictions.service.js</code></strong> — Changed <code>throw</code> to <code>continue</code> (skip orphaned messages)</li></ol>
These are temporary. They are lost on container recreation or version updates.
<hr>
<h2>Proposed Long-Term Fixes</h2>
<ol><li value="1"><strong>In <code>timeline-messaging.service.ts</code>:</strong> Ensure <code>getThreadVisibilityByThreadId()</code> always returns a valid <code>MessageChannelVisibility</code> for every requested thread ID, defaulting to <code>METADATA</code> when the normal resolution path fails.</li></ol>
<ol><li value="2"><strong>In <code>apply-messages-visibility-restrictions.service.ts</code>:</strong>
 Gracefully skip (or log + skip) messages that have no valid channel 
associations, rather than throwing an error that aborts the entire 
query.</li></ol>
<ol><li value="3"><strong>Optional data cleanup:</strong> Consider a migration or cleanup command that removes dead <code>messageChannelMessageAssociation</code> rows (those pointing to deleted channels) and either deletes or relinks orphaned messages during account deletion.</li></ol>
<hr>
<h2>Additional Context</h2>
<ul><li>The Timeline tab works because it queries via a different resolver path that does not require <code>messageChannel.visibility</code></li><li>The
 "All Messages" view and the person-specific Email tab use different 
GraphQL queries but both crash due to the same underlying data integrity
 gap</li><li>This pattern is likely to affect any self-hosted instance where users delete and re-add email accounts</li><li>The data itself is not corrupted — the messages are valid, but the channel linkage was broken maybe by account recreation or other user action</li></ul>

dankgarlic1:
Deleting an IMAP account in Twenty triggers a cascade delete for the messageChannel, but it doesn't clean up the junction table (messageChannelMessageAssociation). This leaves behind "ghost" rows pointing to deleted channels.

This caused two crashes:

1. GraphQL queries for message visibility were failing because they couldn't find a channel for the message.
2. The ApplyMessagesVisibilityRestrictions hook was throwing an error whenever it hit one of these orphaned messages, which caused the entire UI view to fail instead of just skipping the bad row.

The Fix:

- Cleanup Job: I added a cleanup step to MessagingConnectedAccountDeletionCleanupJob that deletes any messageChannelMessageAssociation rows that point to a non-existent channel. This happens during account deletion.
- Defensive Logic: I updated the visibility services to handle missing channels gracefully. Instead of throwing an error or returning undefined, it now defaults to METADATA visibility or obfuscates the message data. This keeps the UI working even if the database is in a bad state.

This prevents the "All Messages" and "Email" tabs from crashing after re-adding accounts.

closes #21425

### DIFF
--- a/packages/twenty-server/src/modules/messaging/common/query-hooks/message/apply-messages-visibility-restrictions.service.ts
+++ b/packages/twenty-server/src/modules/messaging/common/query-hooks/message/apply-messages-visibility-restrictions.service.ts
@@ -7,7 +7,6 @@ import { MessageChannelVisibility } from 'twenty-shared/types';
 import { isDefined } from 'twenty-shared/utils';
 import { In, Repository } from 'typeorm';
 
-import { NotFoundError } from 'src/engine/core-modules/graphql/utils/graphql-errors.util';
 import { UserWorkspaceEntity } from 'src/engine/core-modules/user-workspace/user-workspace.entity';
 import { ConnectedAccountEntity } from 'src/engine/metadata-modules/connected-account/entities/connected-account.entity';
 import { MessageChannelEntity } from 'src/engine/metadata-modules/message-channel/entities/message-channel.entity';
@@ -90,7 +89,10 @@ export class ApplyMessagesVisibilityRestrictionsService {
             .filter(isDefined);
 
           if (messageChannels.length === 0) {
-            throw new NotFoundError('Associated message channels not found');
+            messages[i].subject =
+              FIELD_RESTRICTED_ADDITIONAL_PERMISSIONS_REQUIRED;
+            messages[i].text = FIELD_RESTRICTED_ADDITIONAL_PERMISSIONS_REQUIRED;
+            continue;
           }
 
           const messageChannelsGroupByVisibility = groupBy(
--- a/packages/twenty-server/src/engine/core-modules/messaging/services/timeline-messaging.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/messaging/services/timeline-messaging.service.ts
@@ -225,6 +225,15 @@ export class TimelineMessagingService {
 
     return this.globalWorkspaceOrmManager.executeInWorkspaceContext(
       async () => {
+        const threadVisibilityByThreadId: {
+          [key: string]: MessageChannelVisibility;
+        } = {};
+
+        for (const threadId of messageThreadIds) {
+          threadVisibilityByThreadId[threadId] =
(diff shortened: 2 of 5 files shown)


packages/twenty-server/src/modules/messaging/common/query-hooks/message/apply-messages-visibility-restrictions.service.ts
```diff
@@ -90,7 +89,10 @@ export class ApplyMessagesVisibilityRestrictionsService {
             .filter(isDefined);
 
           if (messageChannels.length === 0) {
-            throw new NotFoundError('Associated message channels not found');
+            messages[i].subject =
+              FIELD_RESTRICTED_ADDITIONAL_PERMISSIONS_REQUIRED;
+            messages[i].text = FIELD_RESTRICTED_ADDITIONAL_PERMISSIONS_REQUIRED;
+            continue;
```
dankgarlic1, line 95: this isn't ideal for the end user. I think we should show a more descriptive message here, something like 'Message content unavailable due to missing channel data.'

what do you suggest?

PR #21470, opened 2026-06-12
fix: enforce strict rules for currency value handling by ai chatbot

### LINKED ISSUE #21419: AI assistant displays CURRENCY field values 1000× too large (divides amountMicros by 10³ instead of 10⁶)
### Bug
  The in-app AI assistant ("Ask AI" chat) reports CURRENCY field values **1000× larger
  than reality**. It appears to divide `amountMicros` by 10³ instead of 10⁶ when
  reading/formatting a CURRENCY field for the user.

  ### Steps to reproduce
  1. Have an Opportunity with `amount` set (CURRENCY), e.g. 656 788 CZK.
  2. Ask the in-app AI assistant: "What is the largest opportunity amount?"
  3. The assistant reports the amount ×1000.

  ### Example (verified against DB)
  Opportunity "Zakázka Telelarm", `amount`:
  - Stored: `{ amountMicros: 656788000000, currencyCode: "CZK" }`
  - Correct value: `656788000000 / 1_000_000` = **656 788 CZK**
  - AI assistant reported: **656 788 000 CZK**  ← `amountMicros / 1_000` (off by 10³)

  ### Expected
  AI assistant should report **656 788 CZK** (divide `amountMicros` by 1 000 000,
  i.e. 10⁶, consistent with the rest of the app / API).

  ### Notes
  - The regular UI, REST/GraphQL API and PDF rendering scale correctly (÷10⁶); only the
    AI assistant output is wrong, so this looks like a formatting/serialization step in
    the agent's CRUD/read tools rather than stored data.
  - Likely in the AI agent tool layer (cf. #21133 "fix(ai) - optimize crud tools").

  ### Environment
  - Twenty server v2.10.1 (self-hosted, Docker)
  - Feature: in-app AI assistant

dankgarlic1:
opportunity:
image
fix: 
image

closes #21419

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/record-crud/zod-schemas/record-properties.zod-schema.ts
+++ b/packages/twenty-server/src/engine/core-modules/record-crud/zod-schemas/record-properties.zod-schema.ts
@@ -14,6 +14,8 @@ import { type ObjectMetadataForToolSchema } from 'src/engine/core-modules/record
 import {
   AddressValueOptionalSchema,
   AddressValueSchema,
+  CurrencyResponseValueOptionalSchema,
+  CurrencyResponseValueSchema,
   CurrencyValueOptionalSchema,
   CurrencyValueSchema,
   EmailsValueOptionalSchema,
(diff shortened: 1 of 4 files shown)


thomtrp:
@etiennejouan curious to have your opinion here. Not sure we want to simply add prompts for differents fields

etiennejouan (commented):
@thomtrp if it works with description in schema (specific to this CURRENCY type) it can be a nice quick win. But we should also update description in filter and data schema for CURRENCY type.

Agree with you, we should not add this in system prompt.

dankgarlic1:
makes sense. I'll remove the system prompt changes, keep the guidance in the currency schemas, update the other currency schemas as well, and re-test to confirm the fix still works

dankgarlic1:
I have made the changes, please lmk if anything else is required!

etiennejouan (dismissed):
Hi @dankgarlic1, thank you for your contribution. You target the right direction. I took the liberty to update your PR, to close it faster

dankgarlic1:
Thanks for the update and for taking the time to improve the implementation. I appreciate the guidance :-)

etiennejouan (approved):

PR #21471, opened 2026-06-12
feat(partners): lock admin-managed + ownership fields on Partner role

rashad:
## What

Tightens the **Partner** self-service role's field-level permissions so a partner can edit its own profile but not admin/ops-controlled or ownership fields. All locks are `canUpdateFieldValue: false` on the Partner object.

**Admin-managed scalar fields (7):** `slug`, `validationStage`, `reviewed`, `ranking`, `partnerTier`, `applicationNotes`, `lastMatchAt`

**Ownership relation FKs (2):** `partnerUser`, `company`

## Why

- The 7 scalar fields are admin/ops-controlled (validation, ranking, tiering, internal notes) — a partner must not be able to self-promote or alter ops data.
- `partnerUser` is the **RLS pivot**: the row-level predicate scopes a partner to records where `partnerUser IS <their workspace member>`. If a partner could clear or repoint it, they'd drop their own record out of scope (an orphan only admins can see). It is already locked on Opportunity; this brings Partner in line.
- `company` is read-only at the object level for partners, so its FK link must not be repointable from the Partner side either.

The remaining Partner relations (`opportunities`, `persons`, `partnerContents`) need no lock — they are already protected by inverse-side field locks or object-level read-only / no-access rules.

## Scope

- One source file: `src/roles/partner.role.ts` (9 field-permission entries).
- No schema changes — additive permission tightening; upgrades cleanly via `deploy` + `install`.
- Version: patch bump `0.5.0 → 0.5.1`.

### DIFF
--- a/packages/twenty-apps/internal/twenty-partners/src/roles/partner.role.ts
+++ b/packages/twenty-apps/internal/twenty-partners/src/roles/partner.role.ts
@@ -213,6 +213,42 @@ export default defineRole({
       fieldUniversalIdentifier: OPPORTUNITY_DESIGN_DOC_STATUS_FIELD_ID,
       canUpdateFieldValue: false,
     },
+    // Partner object — lock admin-managed fields so partners can't self-promote or alter ops data.
+    {
+      objectUniversalIdentifier: PARTNER_OBJECT_UNIVERSAL_IDENTIFIER,
+      fieldUniversalIdentifier: 'a0000002-0000-4000-8000-000000000002',
+      canUpdateFieldValue: false,
+    },
+    {
+      objectUniversalIdentifier: PARTNER_OBJECT_UNIVERSAL_IDENTIFIER,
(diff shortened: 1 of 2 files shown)


packages/twenty-apps/internal/twenty-partners/src/roles/partner.role.ts
```diff
@@ -213,6 +215,56 @@ export default defineRole({
       fieldUniversalIdentifier: OPPORTUNITY_DESIGN_DOC_STATUS_FIELD_ID,
       canUpdateFieldValue: false,
     },
+    // Partner object — lock admin-managed fields so partners can't self-promote or alter ops data.
+    {
+      objectUniversalIdentifier: PARTNER_OBJECT_UNIVERSAL_IDENTIFIER,
+      fieldUniversalIdentifier: 'a0000002-0000-4000-8000-000000000002',
```
martmull, line 221: Use a variable for fieldUniversalIdentifiers

martmull (approved):
small comment i would like to be addressed, but approving though

PR #21472, opened 2026-06-12
fix: prevent duplicate workflow job from ending an active run mid-iteration (#21408)

### LINKED ISSUE #21408: Iterator fails with "Workflow has been ended before this step was completed" in v2.11.1
Environment:
- Twenty v2.11.1
- Self-hosted on Railway
- PostgreSQL
- Workflow feature enabled
- LOGIC_FUNCTION_TYPE=LOCAL

Problem:

I am attempting to iterate over records returned by a Search Records action.

Workflow:

Schedule
→ Search Records (Opportunities)
→ Iterator (All Opportunities)
→ Create Record (Task)

Expected Behaviour:

If Search Records returns 2 Opportunities, the Iterator should create 2 Tasks.

Actual Behaviour:

The Iterator starts processing the first record but the workflow ends before the iteration completes.

Error:

"Workflow has been ended before this step was completed"

Observations:

- Search Records returns 2 records successfully.
- Iterator receives All Opportunities.
- Iterator currentItemIndex = 0.
- Iterator hasProcessedAllItems = false.
- The first loop iteration begins.
- Downstream actions never execute.
- Limiting Search Records to 1 result produces the same error.
- Removing Logic Functions does not resolve the issue.
- A simple workflow of:
  Search Records → Iterator → Create Record
  still fails.

Additional Information:

The server logs are heavily flooded with:

"Uncompiled message detected! Please compile your catalog first."

Railway reports:

"Rate limit of 500 logs/sec reached. Messages dropped."

I am unsure whether this is related, but it may be obscuring a more detailed iterator error.

Has anyone encountered this issue with Iterator in v2.11.1?

WilliamK112:
### Problem
Iterator-based workflows (e.g. `Schedule -> Search Records -> Iterator -> Create Record`) fail with `Workflow has been ended before this step was completed`, killing the run mid-loop before the downstream actions run.

### Root cause
That error string is only produced by `markRunningStepsAsFailed()` inside `endWorkflowRun()`, which rewrites every `RUNNING`/`PENDING` step to `FAILED`. The iterator step is intentionally kept `RUNNING` while it loops (`shouldRemainRunning`), so it is exactly the step that gets clobbered.

`RunWorkflowJob.handle()` wraps `startWorkflowExecution()` in a catch-all that calls `endWorkflowRun(..., FAILED, isSystemError)` on **any** throw. The pre-check in `startWorkflowExecution()` is unlocked, but `startWorkflowRun()` (`@WithLock`) **threw** `INVALID_OPERATION` when the run was no longer `ENQUEUED/NOT_STARTED` - a TOCTOU race. When a duplicate/late `RunWorkflowJob` is processed for the same `workflowRunId`:

1. job A wins the lock, sets the run `RUNNING`, and starts the iterator loop;
2. job B passes the unlocked pre-check, then throws `INVALID_OPERATION` inside the lock;
3. the catch-all ends the whole run as `FAILED`, rewriting job A's actively-running iterator step to the reported error.

(The `#16314` enqueue lock only serializes enqueueing per workspace; it does not make this start path idempotent, so the failure mode is still reachable on `main`.)

### Fix
- `startWorkflowRun()` now returns a `boolean` and **returns `false` instead of throwing** when the run is already past `ENQUEUED/NOT_STARTED`.
- `RunWorkflowJob.startWorkflowExecution()` bails out when `startWorkflowRun()` returns `false`, so a duplicate/late job neither double-executes steps nor force-fails an active run.

### Tests
- Added `run-workflow.job.spec.ts` with regression coverage for the start-race bail-out (asserts the late job does not execute steps and does not end the run).
- Ran the full `src/modules/workflow/` suite locally: **48 suites / 498 tests pass**.

### Note for maintainers
@thomtrp - you mentioned it fails *every* time, which is unusual for a pure race. Code paths outside `executeStep()`'s try/catch can also throw deterministically and surface as this same message through the same catch-all; the `Uncompiled message ... compile your catalog` log flood in the report may be a deterministic trigger. Happy to follow up if you can share full server logs from before the flood for one failing run.

### DIFF
--- a/packages/twenty-server/src/modules/workflow/workflow-runner/jobs/__tests__/run-workflow.job.spec.ts
+++ b/packages/twenty-server/src/modules/workflow/workflow-runner/jobs/__tests__/run-workflow.job.spec.ts
@@ -29,12 +29,10 @@ describe('RunWorkflowJob', () => {
 
   beforeEach(async () => {
     workflowRunWorkspaceService = {
-      getWorkflowRunOrFail: jest
-        .fn()
-        .mockResolvedValue({
-          status: WorkflowRunStatus.ENQUEUED,
-          workflowVersionId,
-        }),
+      getWorkflowRunOrFail: jest.fn().mockResolvedValue({
+        status: WorkflowRunStatus.ENQUEUED,
+        workflowVersionId,
+      }),
       startWorkflowRun: jest.fn(),
       endWorkflowRun: jest.fn(),
     } as any;
--- a/packages/twenty-server/src/modules/workflow/workflow-runner/jobs/__tests__/run-workflow.job.spec.ts
+++ b/packages/twenty-server/src/modules/workflow/workflow-runner/jobs/__tests__/run-workflow.job.spec.ts
@@ -1,0 +1,123 @@
+import { Test, type TestingModule } from '@nestjs/testing';
+
+import { MetricsService } from 'src/engine/core-modules/metrics/metrics.service';
+import { GlobalWorkspaceOrmManager } from 'src/engine/twenty-orm/global-workspace-datasource/global-workspace-orm.manager';
+import { WorkflowRunStatus } from 'src/modules/workflow/common/standard-objects/workflow-run.workspace-entity';
+import { WorkflowCommonWorkspaceService } from 'src/modules/workflow/common/workspace-services/workflow-common.workspace-service';
+import { CodeStepBuildService } from 'src/modules/workflow/workflow-builder/workflow-version-step/code-step/services/code-step-build.service';
+import { WorkflowExecutorWorkspaceService } from 'src/modules/workflow/workflow-executor/workspace-services/workflow-executor.workspace-service';
+import { RunWorkflowJob } from 'src/modules/workflow/workflow-runner/jobs/run-workflow.job';
+import { WorkflowRunWorkspaceService } from 'src/modules/workflow/workflow-runner/workflow-run/workflow-run.workspace-service';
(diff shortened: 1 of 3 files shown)


WilliamK112:
Thanks for the fast reviews! Pushed a follow-up addressing the feedback:

- **`oxfmt` (the failing `server-lint-typecheck`):** the only failure was an `oxfmt` formatting diff in the new spec (I had only run Prettier locally). Reformatted with `oxfmt`; `oxlint --type-aware` was already green.
- **Observability (Copilot + cubic):** added a `this.logger.warn(...)` on the bail-out path in `RunWorkflowJob` so duplicate-enqueue / lock contention stays traceable in production.
- **Terminal statuses (Copilot):** I intentionally kept `startWorkflowRun()` returning `false` (no-op) for *any* status past `ENQUEUED/NOT_STARTED`, including terminal ones. If a duplicate job arrives for an already-`COMPLETED`/`FAILED` run, the previous `throw` propagated to `handle()`'s catch-all, and `endWorkflowRun(FAILED)` would flip a finished run to `FAILED`. A logged no-op is the safer behavior. Happy to switch to an explicit per-status branch (+ a dedicated metric) if you'd prefer.

Local: full `src/modules/workflow/` suite passes (48 suites / 498 tests).

thomtrp:
Hi @WilliamK112, thank you for your contribution. I did not find any places where RunWorkflowJob may be enqueued twice, have you? I suspect it does not fix the bug mentioned. And as you noticed, the potential bug you fixed should look like a flaky race condition rather than something happening on every run as the user metionned.

Feel free to tell me if you saw a place where two runs may be done concurrently, I will open the PR again!

PR #21474, opened 2026-06-12
feat(front): add DEL/Backspace keyboard shortcut to delete selected records

### LINKED ISSUE #21275: 'DEL' Keyboard Shortcut for Deleting Selected Records

sawirricardo:
## Summary

Adds keyboard shortcut support for deleting selected records in table view using the DEL or Backspace key.

Closes #21275

## Changes

- Added `Delete` and `Backspace` key handler in `useRecordTableRowHotkeys` hook
- Guard prevents firing when user is typing inside an input field (`enableOnFormTags: false`)
- Uses existing `useDeleteManyRecords` hook for the actual deletion
- Resets row selection after deletion completes
- Added "Delete selected records" entry to the keyboard shortcuts menu (`KeyboardShortcutsTable`)

## Implementation details

- The hotkey is scoped to the focused row element, following the existing pattern used by `x` (select), `Escape` (clear selection), `Ctrl+A` (select all), etc.
- `selectedRowIdsComponentSelector` is used to get all currently selected record IDs
- Only triggers when `isAtLeastOneRecordSelected` is true and `selectedRowIds` is non-empty
- Reuses the `objectNameSingular` already available from the row context

## Testing

- Select one or multiple records in a table view using `x` key or checkboxes
- Press `Delete` (Windows/Linux) or `Backspace` (macOS)
- Selected records should be soft-deleted with optimistic UI update
- Selection is reset after deletion
- Key should NOT trigger when typing in search bars or input fields

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21474?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/twenty-front/src/modules/keyboard-shortcut-menu/constants/KeyboardShortcutsTable.ts
+++ b/packages/twenty-front/src/modules/keyboard-shortcut-menu/constants/KeyboardShortcutsTable.ts
@@ -22,4 +22,10 @@ export const KEYBOARD_SHORTCUTS_TABLE: Shortcut[] = [
     firstHotKey: 'esc',
     areSimultaneous: true,
   },
+  {
+    label: 'Delete selected records',
+    type: ShortcutType.Table,
+    firstHotKey: 'del',
+    areSimultaneous: true,
+  },
 ];
--- a/packages/twenty-front/src/modules/object-record/record-table/record-table-row/hooks/useRecordTableRowHotkeys.ts
+++ b/packages/twenty-front/src/modules/object-record/record-table/record-table-row/hooks/useRecordTableRowHotkeys.ts
@@ -1,4 +1,5 @@
 import { useOpenRecordInSidePanel } from '@/side-panel/hooks/useOpenRecordInSidePanel';
+import { useDeleteManyRecords } from '@/object-record/hooks/useDeleteManyRecords';
 import { useRecordTableContextOrThrow } from '@/object-record/record-table/contexts/RecordTableContext';
 import { useRecordTableRowContextOrThrow } from '@/object-record/record-table/contexts/RecordTableRowContext';
 import { useResetTableRowSelection } from '@/object-record/record-table/hooks/internal/useResetTableRowSelection';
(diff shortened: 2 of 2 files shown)


charlesBochet:
Thanks for putting this together @sawirricardo

Heads up that we received two PRs for this same issue, so apologies for the overlap. The bigger blocker is on our side: we don't want to ship instant Delete/Backspace deletion without a confirmation step. The risk of accidental data loss is too high, especially with Backspace which people hit reflexively, and we haven't settled on the right product behavior yet (confirmation modal vs an opt-in setting vs an undo toast).

Since that's a product decision we still need to make, we're going to close this for now. This isn't a reflection on your work, and we'd be happy to revisit once the UX is decided. Thanks again for contributing.

PR #21475, opened 2026-06-12
feat(sync): bidirectional TFT ↔ Partners opportunity sync

rashad:
## What

Two Twenty SDK apps that keep opportunities in sync between the TFT workspace and the partners workspace:

- **twenty-tft-bridge** — new app, installed on the TFT workspace
- **twenty-partners** — the partners workspace app (`0.5.0 → 0.6.0`)

## How it works

**Shared fields (sync both ways):** name, amount, closeDate, company (resolved by name), point of contact (resolved by name + email).
**Partners → TFT only:** `matchStatus` is mirrored onto a read-only `partnerMatchStatus` field, plus a `partnersDeepLink` back to the partners record.

- **Opt-in gate** — an opportunity starts syncing when `partnerSyncRequest` is flipped to `REQUESTED`; after that, edits on either side auto-propagate.
- **Loop prevention** — outbound triggers skip updates whose `updatedBy.source === 'APPLICATION'` (i.e. written by an inbound sync); inbound handlers compare-before-write as a backstop. A no-op echo writes nothing and logs no audit event.
- **Transport** — HMAC-SHA256 signed payloads over httpRoute logic functions.
- **Cron backstops** — reconcilers in both directions (`reconcile-opportunities`, `reconcile-echoes`) share a cursor-drain driver that pages a backlog and never advances the watermark past a failed row.
- **Audit** — a `TFT Sync Events` object + view in each app logs every sync with the opportunity name, direction, status, and error.

## Screenshot

https://github.com/user-attachments/assets/0e8a7e2b-5ba8-4a28-bc3b-c3b3d64c1e74

## Testing

Verified end-to-end on local workspaces: forward sync, reverse echo, point-of-contact propagation, the record deep link, loop-prevention (no ping-pong), and both cron backstops (including the no-op skip keeping the audit log clean).

## Notes

- `twenty-tft-bridge` is net-new (initial `0.1.0`).
- Application variables (shared secret, endpoints, TFT API creds) are set per-workspace after install.

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21475?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


martmull:
### Code review

Found 3 issues:

1. The reconcile cron bypasses the opt-in gate: `fetchPage` filters TFT opportunities only by the `(updatedAt, id)` keyset watermark, with no filter on `partnerSyncRequest`. Every TFT opportunity modified since the cursor is pushed to the Partners workspace, including ones that were never flipped to `REQUESTED` — the real-time path (`on-sync-requested.ts`) enforces the opt-in, but the cron backstop syncs everything.

https://github.com/twentyhq/twenty/blob/e5665a209b3d9a5bc04dba4a0233b6917b56b153/packages/twenty-apps/internal/twenty-partners/src/logic-functions/reconcile-opportunities.ts#L96-L126

2. The reconcile cron's GraphQL query selects fields that don't exist on TFT's Opportunity: `tftOpportunityId` (your own comment in `push-opportunity.ts` notes it "lives on the partners side... so don't query it"), `matchStatus` (TFT uses `stage`), and `subscriptionFrequency` (TFT's field is spelled `subscriptionFrequence`, see the working query in [import-from-tft.ts](https://github.com/twentyhq/twenty/blob/a8a8bbb2ed86305af2bc1fab19afef5fe0aad9a1/packages/twenty-apps/internal/twenty-partners/src/scripts/import-from-tft.ts#L173-L175)). GraphQL rejects unknown fields and `queryTft` throws on any GraphQL error, so every cron run fails.

https://github.com/twentyhq/twenty/blob/e5665a209b3d9a5bc04dba4a0233b6917b56b153/packages/twenty-apps/internal/twenty-partners/src/logic-functions/reconcile-opportunities.ts#L102-L114

3. Same wrong-field problem in the real-time push: the SDK query in `push-opportunity.ts` selects `matchStatus` and `subscriptionFrequency` on TFT's own Opportunity, but TFT has `stage` / `subscriptionFrequence` (the bridge app only adds `partnerMatchStatus`, `partnerSyncRequest`, `partnersDeepLink`). If the API rejects the unknown fields, every push fails; otherwise those values are silently never forwarded.

https://github.com/twentyhq/twenty/blob/e5665a209b3d9a5bc04dba4a0233b6917b56b153/packages/twenty-apps/internal/twenty-tft-bridge/src/logic-functions/shared/push-opportunity.ts#L33-L43

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/twenty-apps/internal/twenty-partners/src/logic-functions/shared/sync-cursor.ts
```diff
@@ -0,0 +1,168 @@
...
+//   - Drains the whole backlog (pages of `pageSize` until a short page), bounded by
+//     `maxPages` for the cron timeout.
+//   - Advances the watermark to the (updatedAt, id) of the last successfully processed (or
+//     intentionally skipped) row.
+//   - Stops at the first failed row WITHOUT advancing past it, so the next run retries from
+//     there — a failed sync is never silently skipped.
+// `fetchPage(sinceAt, sinceId)` must return up to `pageSize` rows strictly after the cursor,
+// ordered by (updatedAt asc, id asc). On the first run `sinceId` is null → order by updatedAt
+// from the epoch.
+export async function drainByCursor<TRow>(params: {
```
charlesBochet, line 89: all this code is quite complex / hard to read for no reason IMO but as this is not twenty engine, LGTM

packages/twenty-apps/internal/twenty-partners/src/logic-functions/on-opportunity-changed.ts
```diff
@@ -0,0 +1,166 @@
...
+        responseBody.reason ?? `TFT returned ok=false (HTTP ${response.status})`,
+      );
+    }
+
+    await client.mutation({
+      updateTftSyncEvent: {
+        __args: { id: syncEventId, data: { status: 'OK' } },
+        id: true,
+      },
+    } as any);
```
charlesBochet, line 141: you should avoid as any as much as possible in your code, you are not using typescript which is a powerful way to avoid regressions / increase developement spped

rashad:
back to draft because change of implementation direction avec discussion with @martmull

charlesBochet:
PR staled, let's re-open once you have bandwidth :)

PR #21481, opened 2026-06-12
Bill People Data Labs enrichments in Twenty credits

bosiraphael:
Adds per-enrichment billing to the People Data Labs app. Each **matched** record charges the workspace in Twenty credits via `chargeCredits` (`twenty-sdk/billing`), following the same pattern as the exa app.

- Person match: **336,000 micro-credits** ($0.336 — PDL list price $0.28 + 20% margin)
- Company match: **120,000 micro-credits** ($0.12 — PDL list price $0.10 + 20% margin)

> **Note:** the 20% margin is a first draft, not final — it's a single constant (`src/constants/billing-margin-multiplier.ts`) and easy to adjust once we settle on pricing.

PDL only consumes a credit on a successful match, so `not_found`, errors, and skipped records are free. The charge is emitted once per PDL batch call (≤100 records) with `operationType: CODE_EXECUTION`, `quantity` = match count, and `resourceContext` `pdl/person` / `pdl/company`, at the moment PDL responds — a match whose record write later fails is still billed since the PDL cost was already incurred. Billing failures are non-fatal and never break an enrichment.

Prices and margin live as constants in `src/constants/` for easy retuning. No SDK bump needed (`twenty-sdk@2.10.1` already ships `./billing`).


packages/twenty-apps/internal/people-data-labs/src/logic-functions/utils/charge-matched-enrichments.ts
```diff
@@ -0,0 +1,29 @@
...
+}: {
+  matchedCount: number;
+  costPerMatchDollars: number;
+  resourceContext: string;
+}): Promise<void> => {
+  if (matchedCount === 0) {
+    return;
+  }
+
+  const creditsPerMatchMicro = Math.round(
```
martmull: Math.round or Math.floor?

martmull (approved):

packages/twenty-apps/internal/people-data-labs/src/logic-functions/utils/charge-matched-enrichments.ts
bosiraphael: According to cubic's and the rest of the codebase we should use math.round

PR #21486, opened 2026-06-12
CommandMenuItem overridable entity

Weiko:
## Context
Second PR of the overridable-entities track (after #21436 for views): command menu items become overridable so that edits on non-owned items are stored as overrides instead of mutating the row, and deletion/deactivation becomes reversible.

 ## What this does

- `CommandMenuItemEntity` now extends `OverridableEntity<CommandMenuItemOverrides>` (adds `isActive` + `overrides`). All editable properties are overridable for now (to discuss).
- **Update**: mutations on a command item not owned by the caller (standard items) are written into `overrides`; reads merge them in the DTO. The command palette edit mode (pin,
reorder, shortLabel) now preserves standard values, "Reset label to default" gains true post-save semantics.
- **Delete**: protected items are deactivated (`isActive = false`) instead of deleted; custom items still hard-delete.
- **Object deactivate/enable toggle**: now flips `isActive` on the command item (merged into the main migration call) instead of delete/recreate; a create-if-missing fallback covers legacy deactivated objects.
- **Front**: inactive command items are filtered out of the palette selector.

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21486?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/twenty-server/src/database/commands/upgrade-version-command/2-13/2-13-instance-command-fast-1781253016028-command-menu-item-overridable-entity.ts
```diff
@@ -0,0 +1,27 @@
...
+  implements FastInstanceCommand
+{
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      'ALTER TABLE "core"."commandMenuItem" ADD "overrides" jsonb',
+    );
+    await queryRunner.query(
+      'ALTER TABLE "core"."commandMenuItem" ADD "isActive" boolean NOT NULL DEFAULT true',
+    );
+  }
```
martmull, line 14: one request might be enough, please merge

packages/twenty-server/src/engine/metadata-modules/flat-command-menu-item/utils/from-update-command-menu-item-input-to-flat-command-menu-item-to-update-or-throw.util.ts
```diff
@@ -40,22 +48,40 @@ export const fromUpdateCommandMenuItemInputToFlatCommandMenuItemToUpdateOrThrow
...
-      ...mergeUpdateInExistingRecord({
-        existing: existingFlatCommandMenuItem,
-        properties: [...FLAT_COMMAND_MENU_ITEM_EDITABLE_PROPERTIES],
-        update: updates,
-      }),
+      ...mergedRecord,
+      overrides,
       updatedAt: new Date().toISOString(),
-    };
+    } as FlatCommandMenuItem;
```
martmull: is it necessary?

martmull (approved):

packages/twenty-server/src/engine/metadata-modules/flat-command-menu-item/utils/from-update-command-menu-item-input-to-flat-command-menu-item-to-update-or-throw.util.ts
charlesBochet: Good catch — the blanket cast wasn't needed. It was only there to absorb `overrides`, which `sanitizeOverridableEntityInput` returns as the loosely-typed `Record<string, unknown> | null`. Replaced the whole-object `as FlatCommandMenuItem` with a typed annotation + a targeted cast on `overrides` only, so the rest of the object is now structurally type-checked. (`6d18d894964`)

packages/twenty-server/src/database/commands/upgrade-version-command/2-13/2-13-instance-command-fast-1781253016028-command-menu-item-overridable-entity.ts
charlesBochet: Done — merged into a single `ALTER TABLE ... ADD ..., ADD ...` for `up` and a single `... DROP ..., DROP ...` for `down`.

charlesBochet (approved):

PR #21490, opened 2026-06-12
fix(server): keep workflow command menu item label in sync with workflow name

### LINKED ISSUE #20766: Wrong workflow name in command menu
**Scenario:**
1. Log in
2. Create new workflow with manual trigger (must be unpinned so it'll appear in command menu)
3. Fulfill workflow trigger to see it in command menu

**Actual:** "Manual trigger" appears in command menu instead of workflow's name

Image

**Expected:** Workflow's name appears in command menu

thomtrp:
## Summary

Fixes #20766 — manual-trigger workflows showed `Manual Trigger` in the command menu instead of the workflow's name.

Root cause (confirmed against a live instance): the command menu item's `label` is written **only at activation** in `createOrUpdateCommandMenuItem`, from `workflow.name`, with a hardcoded `'Manual Trigger'` fallback. So:
- a workflow activated while unnamed gets the misleading `Manual Trigger` label, and
- renaming the workflow afterwards never updates the label (`workflow.updateOne` had no label-related hook).

Changes:
- Add `getWorkflowCommandMenuItemLabel` helper and use it in activation; the empty-name fallback is now `Untitled Workflow` (consistent with the rest of the UI) instead of `Manual Trigger`.
- Add `WorkflowCommandMenuSyncWorkspaceService` that updates the active version's command menu item label/shortLabel from the workflow name (idempotent, no-op for non-manual / inactive workflows).
- Add `workflow.updateOne` and `workflow.updateMany` post-query hooks that call the sync service, registered in `WorkflowQueryHookModule`.

Out of scope (separate follow-up): the activation create path can produce duplicate command items for one `workflowVersionId`; recommend making it idempotent / adding a unique constraint.

## Test plan

- [x] `oxlint --type-aware` + `oxfmt` clean on changed files
- [x] Editor TS diagnostics clean (full `nx typecheck` was starved by local dev servers)
- [ ] New integration test `workflow-command-menu-label.integration-spec.ts`:
  - labels the command menu item with the workflow name on activation
  - updates the label when the workflow is renamed
  - falls back to `Untitled Workflow` when the name is cleared

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21490?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


bosiraphael (approved):
LGTM

PR #21507, opened 2026-06-12
fix(front): store auth tokenPair in localStorage instead of a cookie

charlesBochet:
## Problem

A client hit an AWS S3 `RequestHeaderSectionTooLarge` error (`MaxSizeAllowed 8192`) when opening a `https://<workspace>.twenty.com/verify?loginToken=<JWT>` link — the request to load the `/verify` SPA page is served from S3, which rejects it before the app loads.

The dominant cause is the **`tokenPair` cookie**. The auth tokenPair (access + refresh JWTs, ~2–5KB) was persisted in a host-scoped, JS-readable cookie. Nothing server-side ever reads it — the access token is sent to the API via an `Authorization: Bearer` header set in the Apollo auth link (`ExtractJwt.fromAuthHeaderAsBearerToken()` on the backend; no `cookie-parser`). Yet the browser attached that cookie to **every** request to the origin, including static assets and the `/verify` page. Combined with the `loginToken` in the URL, the request header section exceeds S3's 8192-byte limit.

## Fix

Move `tokenPair` from cookie storage to **localStorage**, which is never transmitted in request headers.

- `tokenPairState` now uses `useLocalStorage` (with `getOnInit: true`).
- `getTokenPair` (the synchronous read used by the Apollo auth link) reads from localStorage under the same key.
- A one-time migration (`migrateTokenPairCookieToLocalStorage`) runs before React renders: it ports any existing `tokenPair` cookie into localStorage and **deletes the cookie**, so already-authenticated users aren't logged out and the oversized cookie stops being sent.

## Why this is safe

**Behavior:** equivalent. The cookie was host-scoped (no `domain` attribute), so it never provided cross-subdomain sharing — cross-workspace auth already re-establishes the token per-origin via the `loginToken`-in-URL → `/verify` handoff. localStorage has identical origin scoping.

**Security:** neutral-to-positive.
- No XSS protection lost — the cookie was **not** `httpOnly` (it can't be; JS reads it to build the Bearer header), so it was already XSS-exposed exactly like localStorage.
- No CSRF surface change — the token was never sent as a cookie credential (no `credentials: 'include'`).
- **Reduced exposure** — the token no longer leaks into CDN/proxy/server access logs or request headers, which is the actual bug.
- Server-side revocation (`revokedAt`) and the 60-day refresh-token JWT expiry govern validity, so localStorage's lack of auto-expiry is moot.

## Testing

- `getTokenPair` unit tests updated to localStorage.
- New unit tests for the migration util (port, no-op, no-clobber, error-safety).
- `nx test twenty-front` auth + apollo suites: 125 passing.
- `lint:diff-with-main` clean; changed files typecheck clean.

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21507?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/twenty-front/src/index.tsx
+++ b/packages/twenty-front/src/index.tsx
@@ -1,6 +1,7 @@
 import ReactDOM from 'react-dom/client';
 
 import { App } from '@/app/components/App';
+import { migrateTokenPairCookieToLocalStorage } from '@/auth/utils/migrateTokenPairCookieToLocalStorage';
 import 'react-loading-skeleton/dist/skeleton.css';
 import 'twenty-ui-deprecated/style.css';
 import 'twenty-ui-deprecated/theme-light.css';
@@ -9,6 +10,10 @@ import 'twenty-ui-deprecated/theme-dark.css';
 // style.css; the --t-* theme tokens it relies on are already provided above.
 import 'twenty-ui/style.css';
 import './index.css';
+
+// TODO: REMOVE this after 2026-12-12 — temporary migration of tokenPair from the
+// legacy cookie to localStorage (legacy cookie has a 180-day expiry).
+migrateTokenPairCookieToLocalStorage();
 
 const root = ReactDOM.createRoot(
   document.getElementById('root') ?? document.body,
--- a/packages/twenty-front/src/modules/apollo/utils/__tests__/getTokenPair.test.ts
+++ b/packages/twenty-front/src/modules/apollo/utils/__tests__/getTokenPair.test.ts
@@ -1,126 +1,110 @@
-import { cookieStorage } from '~/utils/cookie-storage';
 import { getTokenPair } from '@/apollo/utils/getTokenPair';
-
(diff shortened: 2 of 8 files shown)


packages/twenty-front/src/index.tsx
```diff
@@ -10,6 +11,10 @@ import 'twenty-ui-deprecated/theme-dark.css';
 import 'twenty-ui/style.css';
 import './index.css';
 
+// TODO: remove this temporary migration once clients have moved tokenPair from
```
martmull: add a date please: REMOVE this after YYYY-MM-DD

martmull (approved):

PR #21511, opened 2026-06-12
Normalize defaultValue properly

martmull:
image

### DIFF
--- a/packages/twenty-docs/developers/extend/apps/data/objects.mdx
+++ b/packages/twenty-docs/developers/extend/apps/data/objects.mdx
@@ -86,6 +86,17 @@ export default defineObject({
 **Base fields are added automatically.** When you define a custom object, Twenty creates standard fields like `id`, `name`, `createdAt`, `updatedAt`, `createdBy`, `updatedBy`, and `deletedAt` for you. You don't need to declare them in your `fields` array — only your custom fields. You can override a default field by declaring one with the same name, but this is rarely a good idea.
 </Note>
 
+## Default values
+
+Literal string defaults must be wrapped in single quotes **inside** the string — `defaultValue: "'Draft'"`, not `defaultValue: "Draft"`. That's why the `status` field above uses `` `'${PostCardStatus.DRAFT}'` ``.
+
+Unquoted strings are reserved for computed defaults, evaluated when a record is created:
+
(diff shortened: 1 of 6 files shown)


packages/twenty-server/src/engine/core-modules/application/application-manifest/utils/normalize-field-manifest-default-value.util.ts
```diff
@@ -0,0 +1,52 @@
...
+  fieldType,
+}: {
+  defaultValue: FieldMetadataDefaultValueForAnyType;
+  fieldType: FieldMetadataType;
+}): FieldMetadataDefaultValueForAnyType => {
+  if (!isDefined(defaultValue)) {
+    return defaultValue;
+  }
+
+  if (typeof defaultValue === 'string') {
```
charlesBochet: non standard, make sure to meet code quality standards

packages/twenty-server/src/engine/core-modules/application/application-manifest/utils/normalize-field-manifest-default-value.util.ts
```diff
@@ -0,0 +1,52 @@
+import {
+  type FieldMetadataDefaultValueForAnyType,
+  FieldMetadataType,
+} from 'twenty-shared/types';
+import { isDefined, isPlainObject } from 'twenty-shared/utils';
+
+import { isCompositeFieldMetadataType } from 'src/engine/metadata-modules/field-metadata/utils/is-composite-field-metadata-type.util';
+import { isFunctionDefaultValue } from 'src/engine/metadata-modules/field-metadata/utils/is-function-default-value.util';
+
+const addSimpleQuotesIfMissing = (value: string): string =>
```
charlesBochet: feels weird that we are re-inventing the wheel here, we should already have tooling there

charlesBochet:
@martmull are we sure we want this change. The whole idea was to be able to distinguish between default: 'now' (function call in postgres and default "'now'", the string "now"). I feel we are losing that

charlesBochet (changes requested):
See comments

martmull:
ok did not know why we need this. Removing and add warning / doc

charlesBochet (approved):

PR #21519, opened 2026-06-13
fix: frontend build failing after code formatting library update

parshipcy:
Starting the frontend(`nx start twenty-front`) was failing because one of its dependencies - the client SDK - can not be built. A recent update made the sdk compatible with a newer version of Prettier. That update started using parts of prettier that the build setup did not recognize, so the build stopped with an error about a missing module. This PR updated the SDK’s build configuration, so it correctly handles those prettier imports.

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21519?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/twenty-client-sdk/src/generate/genql/render/requestTypes/unionType.ts
+++ b/packages/twenty-client-sdk/src/generate/genql/render/requestTypes/unionType.ts
@@ -1,6 +1,6 @@
 // @ts-nocheck
 import { GraphQLUnionType } from 'graphql'
-import uniq from 'lodash/uniq'
+import uniq from 'lodash/uniq.js'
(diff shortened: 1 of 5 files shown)


charlesBochet (dismissed):
Makes sense, than you!

charlesBochet (dismissed):

charlesBochet (approved):

PR #21526, opened 2026-06-13
feat(record-table): aggregate totals follow the field's number format

### LINKED ISSUE #21525: Column-footer aggregates (Sum/Min/Max/Avg) should follow the field's Short/Full format
## Scope & Context

Twenty already lets users choose **Short** (`$1.2m`) vs **Full** (`$1,200,000`) display for **Currency** fields, and **Number** vs **Short** for **Number** fields, and the **table cells** honor that. But the **column-footer aggregates** (Sum / Min / Max / Average), the **record-index group totals**, and the **pie-chart center metric** ignore the field's format setting — they all go through `transformAggregateRawValueIntoAggregateDisplayValue`, which doesn't read the field's `settings`.

Related to #21521.

## Current behavior

A Currency column set to **Full** shows full amounts in its cells, but its footer **Sum** (and Min/Max/Avg) still renders the **abbreviated** value — an inconsistency between the cells and their totals. Same for Number fields set to short.

## Expected behavior

Aggregate totals should follow the column field's own Short/Full setting, so the footer/total matches the cells:
- Currency: full when the field's format is `full`, else short.
- Number: short when the field's type is `shortNumber`, else full.
- Percentage and COUNT operations are unaffected.

Image
Image

## Technical inputs

- Single change in `packages/twenty-front/src/modules/object-record/record-aggregate/utils/transformAggregateRawValueIntoAggregateDisplayValue.ts`: make the CURRENCY and NUMBER branches read `aggregateFieldMetadataItem.settings` (mirroring `CurrencyDisplay`/`NumberFieldDisplay`).
- No new GraphQL, no migration, no snapshot changes. All three consumers (footer, record-index, pie-center) benefit automatically since they already pass the field metadata.
- Follow-up note: there are currently a few separate "short/full" vocabularies (`FieldCurrencyFormat`, `FieldNumberVariant`, graph `displayType`, widget `ChartNumberFormat`); a broader unification could be scoped separately.

1alexandrer:
## Context

Currency and Number fields let users pick **Short** (`$1.2m`) vs **Full** (`$1,200,000`) display, and the table **cells** honor it. But the column-footer aggregates (Sum/Min/Max/Avg), record-index group totals, and the pie-chart center metric go through `transformAggregateRawValueIntoAggregateDisplayValue`, which ignored the field's format — so a **Full** currency column still showed a **short** footer total.

Closes #21525

## What changed

`transformAggregateRawValueIntoAggregateDisplayValue` now mirrors the cell displays:
- **Currency** → full when `settings.format === 'full'`, else short (matches `CurrencyDisplay`)
- **Number** → short when `settings.type === 'shortNumber'`, else full (matches `NumberFieldDisplay`)

All three consumers (footer, record-index, pie-center) already pass the field metadata, so they're fixed by this single change. No new GraphQL, no migration, no snapshot changes. Percentage and COUNT are untouched.

## Screenshots

| Before (footer abbreviated) | 
Screenshot_4

| After (footer follows the field) |

Screenshot_3

## Test plan

- Extended `transformAggregateRawValueIntoAggregateDisplayValue` unit tests with currency `full`/`short` and number `shortNumber`/`number` cases.
- Manual: set a Currency field to **Full** → its column footer Sum now shows the full number, matching the cells.

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21526?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-aggregate/utils/transformAggregateRawValueIntoAggregateDisplayValue.ts
+++ b/packages/twenty-front/src/modules/object-record/record-aggregate/utils/transformAggregateRawValueIntoAggregateDisplayValue.ts
@@ -8,7 +8,10 @@ import { type ExtendedAggregateOperations } from '@/object-record/record-table/t
 import { FieldMetadataType, type Nullable } from 'twenty-shared/types';
 import { formatToShortNumber, isDefined } from 'twenty-shared/utils';
 import { type AggregateOperations } from '~/generated-metadata/graphql';
-import { formatNumber } from '~/utils/format/formatNumber';
+import {
+  DEFAULT_DECIMAL_VALUE,
+  formatNumber,
+} from '~/utils/format/formatNumber';
 import { formatDateString } from '~/utils/string/formatDateString';
 import { formatDateTimeString } from '~/utils/string/formatDateTimeString';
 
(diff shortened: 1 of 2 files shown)


1alexandrer:
Unsure if this is more of a fix, it just makes sense to me footers would follow the columns' settings !

packages/twenty-front/src/modules/object-record/record-aggregate/utils/transformAggregateRawValueIntoAggregateDisplayValue.ts
```diff
@@ -48,14 +51,28 @@ export const transformAggregateRawValueIntoAggregateDisplayValue = ({
...
-          ? `${formatNumber(castedValue * 100, { decimals })}%`
+
+        if (type === 'percentage') {
+          return `${formatNumber(castedValue * 100, { decimals })}%`;
+        }
+
+        // Mirror the number cell display: abbreviate only when the field uses 'shortNumber'.
+        return type === 'shortNumber'
+          ? formatToShortNumber(castedValue)
           : formatNumber(castedValue, { decimals });
```
etiennejouan, line 76: It would be great to extract same logic we have in NumberFieldDisplay, create a util and use it here. 

Concerning currency, haven't check where formatting is. But doing the same if simple and possible would also be the best thing.

etiennejouan (changes requested):
Thanks for your contribution. LGTM but we can try to have common logic.

Please, remove all your comments, they're not in twenty standard.

etiennejouan:
Done in this [PR](https://github.com/twentyhq/twenty/pull/21894). Thanks for your contribution !

PR #21530, opened 2026-06-13
fix: show manually created calendarEvents in list and search results

### LINKED ISSUE #21459: Bug: Creating a calendarEvent returns 201, but nothing is created
## Bug Description

When creating a calendarEvent with the REST API, I get a 201 created back, alongside a the new object.
I can also query that object by ID (find one calendar event), but if I search for it either by API or on the `/objects/calendarEvents` page, it doesn't show up.

## Expected behavior
- That Create One calendar event is blocked if it's not surfaced - OR 
- That manual calendar events can be searched and is shown as any other object entry

### Feature request 
I'll write a separate feature request later, on having "create calendar event" as a core capability, both as a logic function in workflows, as well as available in the SDK. This would likely be a "blind" create capability, awaiting the calendar platform to accept the change, and then propagate it with the standard polling
Image

abinash-thakur:
## Problem
Fixes #21459

When creating a `calendarEvent` via `POST /rest/calendarEvents`, the API 
returns `201 Created` with the new object, and the event is retrievable 
by ID. However, it does not appear in `GET /rest/calendarEvents` list 
results or on the `/objects/calendarEvents` UI page.

## Root Cause
`ApplyCalendarEventsVisibilityRestrictionsService` loops through events 
and checks for a `calendarChannelEventAssociation`. Manually created 
events have no association (only the import pipeline creates associations). 
The service reaches the final `splice(i, 1)` fallback and removes the 
event from results.

## Fix
Added an early `continue` when `associations.length === 0` — orphan 
events with no channel association are treated as visible instead of 
being removed.

## Tests
- Added 2 unit tests covering the orphan event scenario:
  - authenticated user request
  - API key request (userId undefined)
- All 8 tests pass

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21530?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


FelixMalfait:
Mmh this is a complex issue, we'd need to introduce a custom connected account / channel type. This is for later!

PR #21532, opened 2026-06-13
[Experiment] perf(front): cache-first currentUser bootstrap

FelixMalfait:
## Experiment — not for merge as-is

A perf experiment for discussion. Opening as a draft to gather feedback and let CI run.

## Problem

On a warm (returning) load, the app gate ([`MinimalMetadataGater`](https://github.com/twentyhq/twenty/blob/main/packages/twenty-front/src/modules/metadata-store/components/MinimalMetadataGater.tsx)) blocks first paint until **both** object/view metadata **and** `currentUser` are ready.

The metadata store is already cache-first: it persists each entity (including `status: 'up-to-date'`) to `localStorage` with `getOnInit`, opens from cache, and revalidates in the background via collection hashes. 👏

`currentUser` (and `currentWorkspace` / `currentWorkspaceMember` / `currentUserWorkspace`) is **not** — it lives in an in-memory atom, so every load fires a blocking `GetCurrentUser` round-trip before the gate opens. That round-trip is the one remaining network hop on the warm-load critical path; everything else the first screen needs is already in `localStorage`.

## Approach

Generalize the pattern the metadata store already proves out, to the user bootstrap — **without adding any new `useEffect`**:

- Persist the four bootstrap atoms (`currentUser`, `currentWorkspace`, `currentWorkspaceMember`, `currentUserWorkspace`) via the existing `createAtomState({ useLocalStorage, localStorageOptions: { getOnInit: true } })`.
- The gate opens from cache on its own: the existing `IsMinimalMetadataReadyEffect` already derives readiness from the `currentUser` atom alongside metadata status, so persisting the atoms is enough — no new effect.
- Keep firing `GetCurrentUser` (now `network-only`, no longer skipped when a user is present) so it **revalidates in the background** and the existing write-through effect updates the atoms with the fresh result.
- Clear the cached identity on sign-out by adding the four keys to `clearSessionLocalStorageKeys` (already invoked by `clearSession`, which then hard-reloads).

Net effect: warm loads no longer wait on `GetCurrentUser`; the shell paints from cache and corrects within one round-trip. Cold loads (no cache) are unchanged.

## Risks to validate

- **Permission staleness** — `currentUserWorkspace` carries `objectsPermissions` / `permissionFlags`. Cache-first means a brief stale-permission window before revalidation. Not a security boundary (the server authorizes every request), but it can momentarily show a menu item the user no longer has; worst case it 401s and corrects on the next paint.
- **Feature-flag / workspace staleness** — `currentWorkspace.featureFlags` may be one round-trip stale on warm load.
- **`X-Schema-Version` header** — sourced from `currentWorkspace.metadataVersion`; caching it actually makes the header *consistent* with the already-cached metadata rather than absent, but worth confirming against the server's mismatch handling.
- **Test isolation** — these atoms now persist; tests relying on the default `null` could see cross-test leakage if `localStorage` isn't reset. The directly-affected suites pass locally (`useAuth`, `useDefaultHomePagePath`, `useSetNextOnboardingStatus`); CI's full run is the real check.

## Validation

- [ ] Full CI (types/lint/unit) green
- [ ] Manual: throttle network, hard-reload a logged-in workspace, confirm the shell paints before `GetCurrentUser` resolves and that fresh data writes through
- [ ] Sign out → sign in as a different user on the same browser; confirm no stale identity flashes

### DIFF
--- a/packages/twenty-front/src/modules/auth/utils/clearSessionLocalStorageKeys.ts
+++ b/packages/twenty-front/src/modules/auth/utils/clearSessionLocalStorageKeys.ts
@@ -5,6 +5,10 @@ const SESSION_KEYS_TO_CLEAR = [
   'lastVisitedViewPerObjectMetadataItemState',
   'ai/agentChatDraftsByThreadIdState',
   'locale',
+  'currentUserState',
+  'currentWorkspaceState',
+  'currentWorkspaceMemberState',
+  'currentUserWorkspaceState',
 ];
 
 export const clearSessionLocalStorageKeys = () => {
--- a/packages/twenty-front/src/modules/metadata-store/effect-components/MinimalMetadataLoadEffect.tsx
+++ b/packages/twenty-front/src/modules/metadata-store/effect-components/MinimalMetadataLoadEffect.tsx
@@ -1,4 +1,5 @@
 import { useHasAccessTokenPair } from '@/auth/hooks/useHasAccessTokenPair';
+import { currentUserState } from '@/auth/states/currentUserState';
 import { currentWorkspaceState } from '@/auth/states/currentWorkspaceState';
 import { isCurrentUserLoadedState } from '@/auth/states/isCurrentUserLoadedState';
 import { useLoadMinimalMetadata } from '@/metadata-store/hooks/useLoadMinimalMetadata';
@@ -6,11 +7,13 @@ import { useLoadStaleMetadataEntities } from '@/metadata-store/hooks/useLoadStal
 import { metadataLoadedVersionState } from '@/metadata-store/states/metadataLoadedVersionState';
 import { useAtomStateValue } from '@/ui/utilities/state/jotai/hooks/useAtomStateValue';
 import { useEffect, useState } from 'react';
+import { isDefined } from 'twenty-shared/utils';
 import { isWorkspaceActiveOrSuspended } from 'twenty-shared/workspace';
 
 export const MinimalMetadataLoadEffect = () => {
   const hasAccessTokenPair = useHasAccessTokenPair();
   const isCurrentUserLoaded = useAtomStateValue(isCurrentUserLoadedState);
+  const currentUser = useAtomStateValue(currentUserState);
   const currentWorkspace = useAtomStateValue(currentWorkspaceState);
   const metadataLoadedVersion = useAtomStateValue(metadataLoadedVersionState);
   const [lastLoadedVersion, setLastLoadedVersion] = useState<number>(-1);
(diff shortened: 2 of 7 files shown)


packages/twenty-front/src/modules/auth/utils/clearSessionLocalStorageKeys.ts
```diff
@@ -5,6 +5,12 @@ const SESSION_KEYS_TO_CLEAR = [
   'lastVisitedViewPerObjectMetadataItemState',
   'ai/agentChatDraftsByThreadIdState',
   'locale',
+  // Persisted bootstrap identity, revalidated on every load — must not survive
```
FelixMalfait: Remove comment

packages/twenty-front/src/modules/metadata-store/effect-components/UserMetadataProviderInitialEffect.tsx
```diff
@@ -65,15 +65,28 @@ export const UserMetadataProviderInitialEffect = () => {
     [store],
   );
 
-  const shouldSkipUserQuery = !hasAccessTokenPair || isDefined(currentUser);
+  // Always revalidate the current user on an authenticated load. When a user is
```
FelixMalfait: Remove comment

packages/twenty-front/src/modules/metadata-store/effect-components/MinimalMetadataLoadEffect.tsx
```diff
@@ -22,7 +25,7 @@ export const MinimalMetadataLoadEffect = () => {
   const shouldLoadRealMetadata = hasAccessTokenPair && isActiveWorkspace;
 
   useEffect(() => {
-    if (!isCurrentUserLoaded) {
+    if (!isCurrentUserLoaded && !isDefined(currentUser)) {
```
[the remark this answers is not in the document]
FelixMalfait: The optimistic `setLastLoadedVersion` before `performLoad()` is acting as a re-entrancy guard, and this PR makes it load-bearing: I added `currentUser` to this effect's deps, so it now re-runs when the background `GetCurrentUser` revalidation writes through.

Tracing a warm load if we move the set after the awaits:

1. Effect runs (cached `currentUser` present) → `performLoad()` starts; `lastLoadedVersion` still stale.
2. `GetCurrentUser` returns → `setCurrentUser(fresh)` → `currentUser` ref changes → effect re-runs.
3. `metadataLoadedVersion === lastLoadedVersion` no longer matches (the set hasn't happened yet) → `performLoad()` fires again → two concurrent metadata bursts.

With the optimistic set kept, step 3 short-circuits. So moving it after the awaits trades a rare missed retry for a reliable double-load on every warm revalidation.

The missed-retry tradeoff is also pre-existing (these lines aren't changed by this PR) and low-impact — a failed warm-load revalidation leaves the already-cached metadata on screen until the next SSE invalidation or reload. If we want real retry-on-failure, the safe shape is an in-flight ref + commit-on-success rather than reordering, which I'd do as a separate change.

FelixMalfait:
cc @charlesBochet fyi ; will merge since I'm doing a big QA tomorrow morning. Will triple check it doesn't break any flow and revert if it does

PR #21533, opened 2026-06-13
perf(front): load Front chat during browser idle time

FelixMalfait:
## Context

The Front support chat bundle (`chat-assets.frontapp.com/v1/chat.bundle.js`, ~2.3s in a profiling trace) was being injected only `500ms` after auth + client-config + workspace-member resolved (`useInstantiateSupportChat.ts`). Because the effect's gating conditions are themselves network-bound, that `500ms` still lands the fetch + execute **inside the critical boot window** (metadata load + first render), where the bundle competes for bandwidth and main-thread time.

Two pre-existing issues:
- The `500ms` delay was too short to clear the critical window.
- The injected `<script>` already had `defer = true`, but `defer` is a no-op on dynamically-inserted scripts (they're `async` by default), so it contributed nothing.
- The `setTimeout` was never cleared, so an effect re-run within the delay could schedule duplicate loads.

## Change

- Add a small `scheduleIdleCallback(callback, { timeout })` helper (`src/utils/`) that runs work during a browser idle period via `requestIdleCallback`, capped by `timeout`, and returns a canceller.
- Use it in `useInstantiateSupportChat` with a `2000ms` cap, and **return the canceller from the effect** so a pending load is cancelled on re-run/unmount.

### Why not gate on first interaction?
The launcher must appear proactively to surface an unread-reply badge, so it has to load without user action. `requestIdleCallback` keeps it proactive while yielding to the critical path.

### Safari / iOS
`requestIdleCallback` is disabled by default in all shipping Safari/iOS versions (not Baseline). The helper falls back to a plain `setTimeout` of the same duration there. Because `requestIdleCallback`'s `timeout` is a *maximum* (it fires earlier at the first idle gap) while `setTimeout` fires *at* that value, a single `2000ms` value gives:
- **Chrome/Firefox/Edge/Android**: loads at first idle, guaranteed within 2s.
- **Safari/iOS**: loads at 2s (a fixed, longer delay — 4× the old 500ms).

Both paths are strictly better than the previous behavior.

## Testing

- `scheduleIdleCallback` unit tests (both the `requestIdleCallback` and the fallback path, plus cancellation) — 4/4 pass.
- `npx nx typecheck twenty-front` — passes.
- `oxlint --type-aware` + `oxfmt` on changed files — clean.

https://claude.ai/code/session_013YXr5yNGFH1NYUe4ysmiEy

---
_Generated by [Claude Code](https://claude.ai/code/session_013YXr5yNGFH1NYUe4ysmiEy)_

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21533?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/twenty-front/src/utils/scheduleIdleCallback.ts
+++ b/packages/twenty-front/src/utils/scheduleIdleCallback.ts
@@ -1,0 +1,26 @@
+import { isDefined } from 'twenty-shared/utils';
+
+// Runs `callback` during a browser idle period so non-critical work (e.g.
+// loading a third-party widget) does not compete with rendering and data
+// fetching during boot. `timeout` is the maximum delay before it runs: with
+// requestIdleCallback the callback fires at the first idle gap but no later
+// than the timeout; on browsers that lack it (Safari/iOS, where it is
+// disabled by default) we fall back to a plain setTimeout of that duration.
+// Returns a function that cancels the scheduled callback.
+export const scheduleIdleCallback = (
+  callback: () => void,
+  { timeout }: { timeout: number },
+): (() => void) => {
+  const requestIdleCallbackFn: typeof window.requestIdleCallback | undefined =
+    window.requestIdleCallback;
+
+  if (isDefined(requestIdleCallbackFn)) {
+    const handle = requestIdleCallbackFn(callback, { timeout });
+
+    return () => window.cancelIdleCallback(handle);
+  }
+
+  const handle = window.setTimeout(callback, timeout);
+
+  return () => window.clearTimeout(handle);
+};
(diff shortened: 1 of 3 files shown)


packages/twenty-front/src/utils/scheduleIdleCallback.ts
```diff
@@ -0,0 +1,26 @@
+import { isDefined } from 'twenty-shared/utils';
+
+// Runs `callback` during a browser idle period so non-critical work (e.g.
```
FelixMalfait: Remove this comment

FelixMalfait: @claude do it and push to this branch

FelixMalfait: Done — removed the comment and pushed to `claude/cool-thompson-xenami` (commit `744b9307`).

---

PR #21536, opened 2026-06-13
feat(twenty-front): make Notes column actionable in record tables

### LINKED ISSUE #21244: feat: make Notes column actionable in People table
## Scope & Context
Make the Notes column usable from index tables (People/Companies etc).

Related to #20288. I created this separate issue because the bug report is about missing UX, but the code intentionally blocks table editing, so fixing it requires new feature work, not a regression fix. 

## Current behavior
Actually, People Notes column is read-only by design: https://github.com/twentyhq/twenty/issues/20288#issuecomment-4629257911

On People, the Notes column does not show the edit pen on hover and does nothing on click. Notes can only be added from the person record -> Notes tab -> New note.

https://github.com/user-attachments/assets/c4e10979-5b33-4296-8e1e-0fdef4e442f0

## Expected behavior
It will be easy for a user when hover shows an edit pen button. Click opens create-note for that person (e.g. same side panel as New note, with the person pre-linked) or something like that and existing notes should stay clickable.

lakshitsoni26:
Closes #21244

# Problem

The Notes and Tasks columns in record index tables (People, Companies, etc.) are completely read-only. To add a note, users have to click into the record detail page → Notes tab → "New note". That's 3+ clicks for something that should take one.

# Root Cause

`noteTargets` and `taskTargets` are junction-table relation fields with `isUIEditable: false` at the server metadata level. This intentionally blocks inline editing for these relation fields, but it also prevents *any* interactive button from rendering on hover — even one that just opens a drawer.

# How I fixed it

Rather than changing the server-side `isUIEditable` flag (which correctly guards against inline relation editing), I added a targeted UI-layer bypass for activity target fields only:

**`RecordTableCellHoveredPortalContent.tsx`**
- Detects `noteTargets`/`taskTargets` fields on non-activity objects
- Lets the edit button render and shows interactive hover styling, even though the field is technically read-only

**`RecordTableCellEditButton.tsx`**
- For activity target fields, routes clicks to `useOpenCreateActivityDrawer` instead of the default `openTableCell()` flow
- Passes `targetableObjects` with the current record's `id` and `targetObjectNameSingular` so the note/task gets auto-linked
- Uses `IconNotes` for notes and `IconCheckbox` for tasks instead of the generic pencil

**What I intentionally left alone:**
- Server-side `isUIEditable` — kept `false`, the inline edit guard is correct
- `RecordTableCellDisplayMode` — existing note chips already navigate via `RecordChip`
- `getFieldButtonIcon.ts` — handled icon swap in the edit button component to keep the diff small
- `isRecordFieldReadOnly.ts` — no changes needed, bypass is at the rendering layer only

# Demo

 https://github.com/user-attachments/assets/3bcf88f6-235b-481f-9df3-c0a20cb7817f

# Testing

1. Go to People (or Companies) index table
2. Add the Notes column via the `+` button if not visible
3. Hover over any Notes cell → edit button appears with a icon
4. Click it → "Create Note" panel opens, pre-linked to the record
5. Existing note chips still navigate to note detail view (no regression)
6. Same flow works for Tasks column ( icon)

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21536?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


charlesBochet:
Hi @lakshitsoni26, thanks for opening a PR but I don't think this is the right approach. We will rather improve the tooling on junctionTable to stop having an edge cases for tasks and notes.
We will likely work on it this summer, closing this one

PR #21540, opened 2026-06-14
Fix settings UI polish pass 2

Bonapara:
## Summary

This PR groups the requested second UI polish pass across the app shell, settings pages, data model, community, AI tools, and developer setup surfaces.

### Shell and navigation polish
- Lets the app content card fill the available viewport and removes the bottom-left radius on the main container.
- Uses an outside-looking container border treatment so the rounded top-left edge matches the Figma frame without a double border.
- Restores the separator between the AI chat side panel and the record container.
- Updates navigation drawer background/skeleton tones to use gray/3 consistently.
- Simplifies the full-screen backend error fallback so the white screen fills the viewport.

### Settings page polish
- Moves the layout customization action from the settings navbar into a settings card under the hero.
- Tightens layout page copy now that the customization surfaces are not directly manageable yet.
- Aligns role detail title sizing and hover treatment with other settings headers.
- Supports settings card brand icon colors, then restores Discord and X brand logos on the Community page.
- Adds the Community discovery cover, moves community/social content to the top, and reorganizes Partners and Features.
- Keeps setting card subtitle and separator behavior available for the updated settings cards.

### Data model and object settings polish
- Adds a shared data-model table body wrapper so object, field, and relation tables all keep the expected bottom border.
- Displays the object icon in the object settings navbar.
- Shares the new-field wizard parent object icon wrapper and keeps the object icon at 64% opacity in wizard headers.
- Updates the new-field type selector icon treatment to match the wizard header opacity behavior.

### Integrations, tools, and API polish
- Reworks the MCP setup config card with light syntax coloring and a top-right copy icon button instead of the bottom copy section.
- Fixes email import icons to match the Figma asset shape.
- Ensures AI tool table icons and chevrons never fall back to black.

## Validation

- `npx nx lint:diff-with-main twenty-front`
- `npx tsc -p packages/twenty-front/tsconfig.json --noEmit`
- `git diff --check`
- Manual local visual pass on `apple.localhost:3001` for companies, settings/community, data model, object settings, new-field wizard, layout settings, account import, MCP setup, AI tools, and role detail pages.

### Shell and settings frame

Shell and settings frame before/after board

### Community and data model

Community and data model before/after board

### Integrations and tools

Integrations and tools before/after board

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21540?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/twenty-front/src/modules/settings/playground/components/SettingsMcpSetup.tsx
```diff
@@ -24,6 +51,61 @@ const buildMcpConfig = (serverUrl: string) =>
...
+  return (
+    <span key={key}>
+      <StyledJsonString>{beforePlaceholder}</StyledJsonString>
+      <StyledJsonPlaceholder>{apiKeyPlaceholder}</StyledJsonPlaceholder>
+      <StyledJsonString>{afterPlaceholder}</StyledJsonString>
+    </span>
+  );
+};
+
+const getHighlightedMcpConfig = (mcpConfig: string) => {
```
FelixMalfait, line 72: This feels a bit odd. I would either use the component we use for code blocks which has real syntax highlighting (consistent across all app / we can tailor it and it would also improve code node etc) ; or I would just hardcode the style. But building a fake incomplete highlighting parser feels odd and typical "LLM" way of doing things (small isolated hacky code because it's not lazy to do so).

I'm releasing this morning so I'll take your PR since this is my only comment but maybe fix in a subsequent PR?

FelixMalfait (approved):
Great!!!

packages/twenty-front/src/modules/settings/playground/components/SettingsMcpSetup.tsx
FelixMalfait: I would fix myself but not sure which of the two solutions you prefer 🤔

Bonapara: Thanks!

FelixMalfait: Just to be clear I was saying I'm not fixing because I'm not sure which path you prefer

(If you want me to do a followup PR let me know)

PR #21545, opened 2026-06-14
feat(front): replace record table hover portal with CSS :hover

charlesBochet:
## What & why

The record table renders a single **"hovered cell" React portal** that follows the mouse (a delegated `onMouseMove` on the table body writes a Jotai atom; a portal then re-renders the hovered cell's content into that cell's DOM node, in a *focused* state, with the edit button and an interactive outline).

This PR replaces that portal with a **pure CSS `:hover`** affordance on `RecordTableCellBaseContainer`. The edit / open-record button and the outline are revealed with CSS, so resting cells stay cheap and there is no per-move atom churn or portal re-render.

The focus portal (clicked / keyboard-navigated cell) and the edit-mode portal are unchanged.

## Before / after (hover affordance)

| Resting | Hover (CSS outline + edit button) |
|---|---|
|  |  |

Full table — resting:

Full table — hovering an editable cell (pencil) and the first column (open-record arrow):

 

## Intended simplification (needs a UX 👍)

On hover the cell content was wrapped in `FieldFocusStaticFocusedProvider`, so any field reading `useFieldFocus()` rendered its **focused** variant. Several fields swap to a *different, heavier component* in that state:

- multi-select / relation-from-many / morph-relation → `ExpandableList` (measures each child's `offsetLeft`, opens a floating dropdown)
- phones / emails → clickable links + chip count

CSS cannot swap component trees, and mounting those measured components in **every** cell would defeat the perf win the single portal provided. So those fields now expand **only when the cell is actually focused** (clicked or keyboard-navigated) via the retained focus portal — not on plain hover. This is the "simplify" path: hovering a multi-select/relation/phones cell shows the truncated display; click to expand.

## Removed

- `RecordTableCellHoveredPortal` + `RecordTableCellHoveredPortalContent`
- `recordTableHoverPositionComponentState` and the delegated `onMouseMove` / mouse-leave tracking
- `useMoveHoverToCurrentCell` and the now-dead `useHandleContainerMouseEnter`
- `onMoveHoverToCurrentCell` from `RecordTableBodyContext` (+ providers, decorator, perf story)

## Validation

- `nx typecheck twenty-front` — clean for the touched files
- `nx lint:diff-with-main twenty-front` — 0 warnings / 0 errors, formatting OK
- Verified live (screenshots above): hover outline + pencil on editable cells, open-record arrow on the first column, clean resting state, focus portal still works on click.

## Follow-ups before merging

- [ ] **Perf benchmark** (the portal existed for perf): wide virtualized table, mouse-move scripting time + DOM node count before/after. One hidden edit button per visible cell is the only added static DOM; flagged for measurement.
- [ ] UX sign-off on expand-on-focus (vs hover) for multi-select / relations / phones.
- [ ] Edit-mode hover suppression + outline stacking pass (raise `z-index` on `:hover` if any clipping shows up next to sticky columns).

> Opened as a **draft** — this is the focused-table portion of a larger "kill the hover portal" effort; the inline-cell hover system is intentionally untouched. The temporary `charles/hover-css-pr-assets` branch only hosts the screenshots above and can be deleted after review.

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21545?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


charlesBochet:
Closing, we need to discuss with product first, then create a proper setup for testing to avoid regressions. Then we can proceed

PR #21546, opened 2026-06-14
fix(record-field): prevent state clobbering and focus loss in dynamic array inputs

### LINKED ISSUE #21544: fix(record-field): prevent state clobbering in dynamic array inputs
## Bug Description

In `FormArrayFieldInput` and `ArrayFieldInput`, dynamic arrays of input items are being rendered using the array index as the React key (`key={index}`). 

When a user deletes an item from the middle of the array, React reuses the DOM nodes and local component state for the subsequent items. This leads to text inputs shifting data to the wrong rows, or users losing focus unexpectedly while typing.

## Expected behavior

Array items should be rendered using stable, unique identifiers rather than the array index so that React's reconciliation engine accurately tracks DOM node additions and deletions. When a user deletes a row, the remaining rows should keep their correct data and focus.

## Technical inputs

- Fix the `key={index}` anti-pattern in `FormArrayFieldInput.tsx` and `ArrayFieldInput.tsx`.
- Use a stable identifier for the key prop (like a unique ID or a hash of the value) instead of the array map index.
- I will open a PR with the fix!

lakshitsoni26:
Closes #21544

# Description

This PR fixes a critical React reconciliation bug where dynamic array items in record fields lost state and focus when an item in the middle of the array was deleted. This was caused by the anti-pattern of using the array map `index` as the React `key`.

### Root Cause & Technical Implementation
To accurately track DOM node additions and deletions, I replaced the index keys with stable UUIDs across the affected components:

- **`FormArrayFieldInput.tsx`**: Restructured the internal static `draftValue` state to natively use an array of objects (`[{ id, value }]`), assigning a stable `v4()` UUID to each item upon creation. This natively guarantees stable identities across edits and deletions. I also implemented safe optional chaining (`items[itemToEditIndex]?.value`) to prevent edge-case crashes if an item is rapidly deleted while its edit input is open.
- **`ArrayFieldInput.tsx`**: Implemented a robust, deterministic event-driven tracking mechanism (`stableIdsRef`) using `useRef`. This synchronously tracks UUIDs alongside the external string array without relying on fragile heuristic guessing algorithms or modifying the parent data contracts. Because it uses `useRef`, it is completely immune to asynchronous state-batching race conditions.

### How to Test / Verification

1. Navigate to any object record in the CRM that utilizes a dynamic array field.
2. Add 3 distinct items to the array input (e.g., "Item A", "Item B", "Item C").
3. Click into the input field for "Item C" to ensure it has browser focus.
4. Delete the middle item ("Item B").
5. **Verify:** "Item C" should shift up, retain its exact value, and React should not lose focus or overwrite the component state.

https://github.com/user-attachments/assets/7495e1a6-6c44-4856-892d-3a16eb653866

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21546?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lakshitsoni26:
@cubic-dev-ai review

lakshitsoni26:
@cubic-dev-ai review

lakshitsoni26:
@cubic-dev-ai review

lakshitsoni26:
Closing this pull request in favor of a fresh, clean PR (#21558) to clear the messy commit history and avoid CI bot false-positives caused by the recent branch divergence. All changes have been moved to the new PR.

PR #21562, opened 2026-06-14
fix(messaging): honor IMAP/SMTP encryption setting instead of inferring it from the port

### LINKED ISSUE #21300: IMAP/SMTP connection: SMTP transport ignores the Encryption (SSL/TLS) setting

neo773:
This pull request makes the IMAP and SMTP encryption setting actually honor what the user selects.

As per spec there's 3 modes: SSL/TLS (implicit TLS from the start), STARTTLS (it will attempt TLS but if the server doesn't support it, it gracefully falls back to plaintext), NONE (plaintext)

Current implementation had a boolean flag for this, this replaces it with the 3 modes

Upgrade command to migrate all existing accounts, to not risk breaking anyone's existing account in production we map each account to the mode that matches its current behavior, so nothing changes on the wire

/closes #21300

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21562?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/twenty-server/src/database/commands/upgrade-version-command/2-13/2-13-instance-command-slow-1781461753981-backfill-connection-security.ts
```diff
@@ -0,0 +1,65 @@
...
+import { SlowInstanceCommand } from 'src/engine/core-modules/upgrade/interfaces/slow-instance-command.interface';
+
+// Replaces the legacy `secure` boolean with the `connectionSecurity` enum,
+// preserving the previously deployed on-wire behavior exactly. SMTP used
+// implicit TLS on port 465 (-> SSL_TLS) and opportunistic STARTTLS elsewhere
+// (-> STARTTLS); IMAP used implicit TLS when secure (-> SSL_TLS) and
+// opportunistic STARTTLS otherwise (-> STARTTLS); CalDAV is TLS over its https
+// host (-> SSL_TLS). STARTTLS stays opportunistic so no account is forced onto
+// a stricter handshake than it already used.
+@RegisteredInstanceCommand('2.13.0', 1781461753981, { type: 'slow' })
```
charlesBochet: should be 2.14 make sure this works! do not hesitate to add an integration test for your command, we have a few example

packages/twenty-server/src/engine/metadata-modules/connected-account/dtos/connected-account-public.dto.ts
```diff
@@ -15,8 +16,8 @@ class PublicConnectionParametersDTO {
   @Field(() => String, { nullable: true })
   username?: string;
 
-  @Field(() => Boolean, { nullable: true })
-  secure?: boolean;
+  @Field(() => String, { nullable: true })
```
charlesBochet: agree why not enum here?

neo773: I got confused with a rule we had with preferring string literal over enum, corrected this now

packages/twenty-server/src/database/commands/upgrade-version-command/2-13/2-13-instance-command-slow-1781461753981-backfill-connection-security.ts
neo773: Seems I didn't pull and created from a main that was behind several commits hence why it landed in 2-13, adding a test

charlesBochet (approved):

PR #21580, opened 2026-06-15
fix(front): dedupe morph relation fields in view field pickers

charlesBochet:
## Issue

Reported in quality-feedbacks: **"Issues with morph relation view field"** — a morph relation column added to a view **disappears after refresh** (and can be added several times).

## Root cause — the SSE metadata sync

A morph relation is stored as **one `fieldMetadata` row per target object**, all sharing a `morphId`. Collapsing those rows into the single field that represents the relation is a **read-time projection** in the server's `objects.fieldsList` resolver — it is *not* a storage invariant, and the rows are never merged.

The frontend metadata store is kept in sync with the raw rows **one row at a time over SSE** (`MetadataStoreSSEEffect`): every metadata change broadcasts a single created/updated record that's pushed straight into the store. Creating a morph relation creates N rows (one per target), so **N `create` events arrive and N raw sub-fields land in the store — bypassing the `fieldsList` projection entirely.**

The view-field pickers read straight from that store, so they saw the morph relation **once per target**. Each could be added as a column referencing a different sub-field id; after a refresh the view reloads from the projected (deduped) data, the non-survivor columns no longer resolve, and they disappear.

## Fix & architecture note

Because the store deliberately mirrors raw rows (that's what the SSE sync maintains), the fix applies the **same read-time projection on the client** — deduping morph rows by `morphId` in `useActiveFieldMetadataItems` — rather than filtering rows at each insert path (SSE, optimistic create, …). This matches how the backend already models morph fields and is robust regardless of which path delivered the rows.

The survivor-selection rule (which sub-field id represents the relation) now lives in `twenty-shared` (`pickMorphGroupSurvivor`) so client and server can't drift.


packages/twenty-front/src/modules/object-metadata/hooks/useActiveFieldMetadataItems.ts
```diff
@@ -11,14 +12,16 @@ export const useActiveFieldMetadataItems = ({
...
       isDefined(objectMetadataItem)
-        ? objectMetadataItem.readableFields.filter(
-            ({ id, isActive, isSystem, name }) =>
-              isActiveFieldMetadataItem({
-                objectNameSingular: objectMetadataItem.nameSingular,
-                fieldMetadata: { isActive, isSystem, name },
-              }) ||
-              // Allow label identifier field even if it's a system field
-              id === objectMetadataItem.labelIdentifierFieldMetadataId,
+        ? dedupeMorphRelationFieldMetadataItems(
```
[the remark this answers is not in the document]
charlesBochet: The dedupe only ever removes `MORPH_RELATION` sub-fields (non-survivors of a `morphId` group). A label identifier is always a scalar/displayable field and can never be a morph relation, so it isn't a dedupe candidate and can't be dropped here — adding a guard would be a branch for a state the data model can't produce. Happy to add an explicit safeguard if there's a path where a morph relation can be the label identifier that I'm missing.

packages/twenty-front/src/modules/object-metadata/utils/dedupeMorphRelationFieldMetadataItems.ts
```diff
@@ -0,0 +1,45 @@
+import { FieldMetadataType } from 'twenty-shared/types';
+import { isDefined, pickMorphGroupSurvivor } from 'twenty-shared/utils';
+
+import { type FieldMetadataItem } from '@/object-metadata/types/FieldMetadataItem';
+
+type DedupableMorphFieldMetadataItem = Pick<
+  FieldMetadataItem,
+  'id' | 'type' | 'morphId' | 'isActive' | 'isSystem'
+>;
```
prastoin: redundant no ?

packages/twenty-front/src/modules/object-metadata/utils/dedupeMorphRelationFieldMetadataItems.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export const dedupeMorphRelationFieldMetadataItems = <
+  T extends DedupableMorphFieldMetadataItem,
```
prastoin: I'm not sure we really need this to be a generic ? I feel like it's already very abstracted when being called to be an union of all the field types ?

packages/twenty-shared/src/utils/fieldMetadata/pick-morph-group-survivor.ts
```diff
@@ -0,0 +1,30 @@
...
+type MorphGroupSurvivorCandidate = {
+  id: string;
+  isActive?: boolean | null;
+  isSystem?: boolean | null;
+};
+
+const scoreMorphField = (field: MorphGroupSurvivorCandidate): number =>
+  (field.isActive ? 2 : 0) + (field.isSystem ? 0 : 1);
+
+export const pickMorphGroupSurvivor = <T extends MorphGroupSurvivorCandidate>(
```
prastoin: or throw

prastoin (approved):
Looks good to me
Small concern regarding the scalability long term on very huge object with a lot of morph relations fields]
=> I mean that anyone that wants to consume the metadata sse for morph relation might end up doing this ? ( not really that's very twenty-front specific in the end )

packages/twenty-front/src/modules/object-metadata/utils/dedupeMorphRelationFieldMetadataItems.ts
charlesBochet: dropped the generic, it now just takes `FieldMetadataItem[]`. The genericity was only there to avoid an `as` in the test; not worth the abstraction.

packages/twenty-front/src/modules/object-metadata/utils/dedupeMorphRelationFieldMetadataItems.ts
charlesBochet: Removed the `Pick<>` helper type along with the generic.

packages/twenty-shared/src/utils/fieldMetadata/pick-morph-group-survivor.ts
charlesBochet: Renamed to `pickMorphGroupSurvivorOrThrow`

PR #21582, opened 2026-06-15
feat(workflow): use workspace member as variable sender for emails

thomtrp:
## Summary

Lets the email workflow node sender be driven by a variable: the connected-account field accepts a `{{variable}}`, and the backend resolves it to a connected account at run time.

Capture d’écran 2026-06-16 à 11 59 27

### Email sender by variable
- The connected-account field now accepts a `{{variable}}` via the variable picker (uses `FormSelectFieldInput` with `WorkflowVariablePicker`), with a hint to pick a connected account or set a workspace member as a variable.
- The email workflow action resolves the stored sender value explicitly: if it is a `workspaceMemberId` (a UUID matching a workspace member), it resolves that member's first connected account; otherwise the value is used directly as a `connectedAccountId`.
- Resolution lives in `EmailWorkflowActionBase` and applies to both `SEND_EMAIL` and `DRAFT_EMAIL`. If a matching member has no connected account, the run fails fast with a clear message (no silent fallback).
- `DRAFT_EMAIL` also fails fast when the resolved connected account is missing the required OAuth scopes (`gmail.compose` / `Mail.Send`), via a server-side `getMissingDraftEmailScopes` util that mirrors the front-end check.
- Existing workflows with a hardcoded `connectedAccountId` keep working unchanged (no migration needed).

> Note: exposing the running workspace member as a manual-trigger variable (`_metadata.workspaceMemberId`) is split into a follow-up PR.

## Test plan
- [x] Backend unit tests for `draft-email-tool`, `get-missing-draft-email-scopes`, and the `send-email` / `draft-email` workflow actions (incl. workspace-member sender resolution)
- [x] Lints clean on all changed files
- [x] Manual: configure an email node with a workspace-member variable sender and confirm it resolves and drafts/sends
- [x] Manual: confirm a member lacking compose permission fails the run with the permission message

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21582?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

---

### Update — scoped to Draft Email only

The sender variable picker is now exposed **only on the Draft Email node**. The Send Email node keeps a plain account select (no variable picker, no variable hint) until we enable it there in a follow-up. Backend resolution still lives in `EmailWorkflowActionBase` and remains generic, so enabling the picker for Send Email later requires no backend change.


Weiko (commented):
LGTM
Note: Seems like a workflow author can draft emails as another member's account and this is fine for now until we work on a better privilege implementation for workflows.

packages/twenty-server/src/modules/workflow/workflow-executor/workflow-actions/mail-sender/email-workflow-action.base.ts
```diff
@@ -28,6 +49,107 @@ export abstract class EmailWorkflowActionBase extends ToolBackedWorkflowAction<W
     return { ...rawInput, body, files };
   }
 
+  protected override async postprocessInput(
+    resolvedInput: WorkflowSendEmailActionInput,
+    workspaceId: string,
+  ): Promise<WorkflowSendEmailActionInput> {
```
Weiko, line 55: Shouldn't we check for action.type === DRAFT_EMAIL in the backend too ?

Weiko (approved):
LGTM @thomtrp

packages/twenty-server/src/modules/workflow/workflow-executor/workflow-actions/mail-sender/email-workflow-action.base.ts
thomtrp: this is ok I think because we want to enable it soon. But you're right, let's be consistent

PR #21586, opened 2026-06-15
feat(front): persist metadata store cache in IndexedDB instead of localStorage

charlesBochet:
## Problem

The metadata store cache (object/field metadata, views, page layouts, command menu items, …) is persisted client-side to power **cache-first boot**: the app renders instantly from the cache, then `MinimalMetadataLoadEffect` revalidates per-collection hashes and only refetches what's stale.

It was persisted to **localStorage**, which Safari/WebKit caps at **~5 MB per origin, counted in UTF-16 (2 bytes/char)** → an effective ceiling of ~2.5 M characters. Measured on the seeded demo workspace (33 objects, 612 fields):

| Bucket | Safari quota (UTF-16) |
|---|---|
| `metadataStoreState__*` (26 keys) | **1.9 MB — 37%** |
| Whole origin | **2.47 MB — 48%** |

A workspace ~2.5× the demo's schema blows past 5 MB, and there is **no `QuotaExceededError` handling** — `setItem` throws and breaks the app. This is what large-workspace users on Safari have been hitting.

## Fix

Move **only the metadata store** to **IndexedDB** (multi-GB, disk-based quota), keeping a **fully synchronous read path** so the ~24 consumers that read these atoms with `useAtomValue` never suspend. The auth/UI atoms (incl. the synchronously-read `tokenPair`) stay on localStorage — intentionally scoped.

- **`createIndexedDbBackedJotaiStorage.ts`** — a synchronous Jotai storage facade backed by an in-memory map, hydrated once from IndexedDB at boot and written through on every set. IndexedDB access uses the **`idb-keyval`** library (by the IndexedDB spec co-author, ~0.6 KB) rather than a hand-rolled wrapper. Each cache gets its own database + BroadcastChannel (`twenty-front-<cacheName>`), so it's safely reusable. Swallowed errors are surfaced via `logError`. When IndexedDB is unavailable the cache stays in memory only (re-fetched each boot).
- **`createAtomFamilyState`** — gains an optional `storage` param; `metadataStoreState` uses the IndexedDB-backed storage.
- **`index.tsx`** — awaits hydration before mounting so atoms (`getOnInit: true`) read the persisted snapshot synchronously → cache-first boot preserved.
- **No migration**: the facade does not touch localStorage at all. Pre-existing localStorage snapshots are ignored — on first boot of the new code the IndexedDB cache is empty and atoms re-fetch from the network (a one-time reconnect). Old `metadataStoreState__*` localStorage keys are left in place (cleared by the existing logout/reset cleanup); new writes only ever go to IndexedDB.
- **Cross-tab sync**: the old localStorage atoms synced across tabs for free via `storage` events; the IndexedDB facade had no equivalent, so a schema change in one tab left others stale until reload. Restored by implementing the Jotai storage `subscribe` contract over a **`BroadcastChannel`** — writes broadcast to other tabs, which update their in-memory map and notify `atomWithStorage` subscribers so mounted atoms re-render live. (BroadcastChannel doesn't echo to the sender, so no feedback loop; guarded for environments without it.)

## Why a synchronous facade (not async `atomWithStorage`)

Consumers use `useAtomValue` directly; an async storage would make the atoms resolve to Promises and **suspend** every reader. The in-memory facade keeps reads synchronous (zero ripple on consumers) and confines the async part to a single bulk read at boot, which the existing `MinimalMetadataGater` loader already covers.

### Automated
- Unit test (10 cases) for the storage facade: synchronous read/write, IndexedDB write-through, hydration from IndexedDB, `removeItem`/`clear`, per-cache DB namespacing, persist-failure logging, in-memory-only behaviour when IndexedDB is unavailable, distinguishing a stored `undefined` from a missing key, and cross-tab subscriber registration.
- Existing metadata-store tests (`useIsLayoutCustomizationDirty`, `useDefaultHomePagePath`) still pass.
- `nx typecheck twenty-front` and `nx lint:diff-with-main twenty-front` clean.

### Manual (local seeded workspace, two tabs, Playwright)
Storage:
- After login the metadata cache lives in **IndexedDB (24 keys, ~945 KB)** and **localStorage drops 48% → 11%** of the Safari quota (the remainder is `currentUserState` + auth, out of scope).
- Reload boots from the cache (no heavy refetch).

Scenarios:

| Scenario | Result |
|---|---|
| **Sign out** | auth cleared, redirect to sign-in, no leftover localStorage, no errors |
| **Sign back in** | metadata `up-to-date`, company table renders, token restored |
| **Add object** (`Gadget`) | write-through to IndexedDB; survives reload via cache-first hydration |
| **Add view** (`QA Cross Tab View`, TABLE) | persisted to the `views` collection (`up-to-date`) |
| **Two tabs open** | second tab boots cleanly from the shared IndexedDB — no lock/crash under concurrent access |
| **Cross-tab live sync** | creating an object in tab A makes it appear in tab B's open settings object list **without a reload** |

Verified by design (no regression):
- Runtime sign-out (`clearSession`) clears session keys and does a full `window.location.assign` reload; the metadata-clearing path (`resetJotaiStore`) is test-only, so there's no async-`clear()`-vs-sign-in race. Metadata persisting across sign-out is unchanged from the old localStorage behavior (it's schema, revalidated by hash on next login).

## Notes / follow-ups (not in this PR)

- **IndexedDB query capabilities** are not used yet: the cache stores one blob per collection (as it did in localStorage), so this is still a pure key-value use (`idb-keyval`). If we later want to query individual metadata records — e.g. fields by `objectMetadataId` via an index/cursor, or partial hydration — that means record-level storage and a richer wrapper (`idb` for a thin near-native layer, or **Dexie** for a full query API + reactive `liveQuery` that could also replace the BroadcastChannel sync).
- IndexedDB still has a (large) quota and Safari ITP eviction applies to both stores — the cache-first design already tolerates eviction by revalidating.
- Complementary "load less" wins remain: the denormalized per-field `relation` block (~700 chars/field of pure duplication) and persisting `currentUser.workspaceMembers` (the ~0.5 MB still in localStorage).

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21586?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Weiko (approved):
Interesting one, LGTM !

packages/twenty-front/src/modules/ui/utilities/state/jotai/utils/__tests__/createIndexedDbBackedJotaiStorage.test.ts
```diff
@@ -0,0 +1,125 @@
...
+  });
+};
+
+type Item = { value: number };
+
+const INITIAL: Item = { value: 0 };
+
+describe('createIndexedDbBackedJotaiStorage', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
```
prastoin, line 33: Looks counter-intuitive with above mock
```ts
jest.mock('idb-keyval', () => ({
  createStore: jest.fn(() => ({ store: 'mock' })),
  set: jest.fn(() => Promise.resolve()),
  del: jest.fn(() => Promise.resolve()),
  clear: jest.fn(() => Promise.resolve()),
  entries: jest.fn(() => Promise.resolve([])),
}));
```

packages/twenty-front/src/modules/ui/utilities/state/jotai/utils/createIndexedDbBackedJotaiStorage.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  const broadcastChannel: BroadcastChannel | null = (() => {
+    try {
+      return typeof BroadcastChannel !== 'undefined'
+        ? new BroadcastChannel(CROSS_TAB_SYNC_CHANNEL_NAME)
+        : null;
+    } catch {
+      return null;
+    }
+  })();
```
prastoin: verbose, IIFE + surprising try catch pattern

packages/twenty-front/src/modules/ui/utilities/state/jotai/utils/createIndexedDbBackedJotaiStorage.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  if (broadcastChannel !== null) {
```
prastoin, line 59: early return

packages/twenty-front/src/modules/ui/utilities/state/jotai/utils/createIndexedDbBackedJotaiStorage.ts
```diff
@@ -0,0 +1,162 @@
+import {
+  clear as idbClear,
+  createStore,
+  del as idbDelete,
+  entries as idbEntries,
+  set as idbSet,
+  type UseStore,
+} from 'idb-keyval';
```
prastoin: ***suggestion***:
```ts
import * as idb from 'idb-keybal'

idb.clear
idb.del
```

packages/twenty-front/src/modules/ui/utilities/state/jotai/utils/createIndexedDbBackedJotaiStorage.ts
```diff
@@ -0,0 +1,162 @@
...
+    isHydrated = true;
+  };
+
+  const clear = async (): Promise<void> => {
+    memoryMap.clear();
+
+    if (isDefined(idbStore)) {
+      try {
+        await idbClear(idbStore);
+      } catch {}
```
prastoin: afraid of all these catch ending in no op, does not provide any traceability at all

prastoin (commented):
ps: We might end up leaving ghost metadata into the user's local storage, not critical but to keep in mind

Weiko (commented):
We could clear legacy local storage during boot but probably overkill.
I agree with @prastoin about the catch

packages/twenty-front/src/modules/ui/utilities/state/jotai/utils/__tests__/createIndexedDbBackedJotaiStorage.test.ts
charlesBochet: Good point — the global toggle alongside the idb-keyval mock was confusing. Extracted `isIndexedDbAvailable` into its own util; the test now mocks it directly (no more `globalThis.indexedDB` toggling), consistent with how idb-keyval is mocked. Done in 73d9543.

packages/twenty-front/src/modules/ui/utilities/state/jotai/utils/createIndexedDbBackedJotaiStorage.ts
charlesBochet: Replaced the IIFE with a named `createCrossTabChannel(cacheName)` helper using early returns. 73d9543.

packages/twenty-front/src/modules/ui/utilities/state/jotai/utils/createIndexedDbBackedJotaiStorage.ts
charlesBochet: Done — the extracted helpers (`createIndexedDbStore`, `createCrossTabChannel`) use early returns now. 73d9543.

packages/twenty-front/src/modules/ui/utilities/state/jotai/utils/createIndexedDbBackedJotaiStorage.ts
charlesBochet: Switched to `import * as idb from 'idb-keyval'` (`idb.set`, `idb.del`, `idb.entries`, `idb.clear`, `idb.createStore`). 73d9543.

packages/twenty-front/src/modules/ui/utilities/state/jotai/utils/createIndexedDbBackedJotaiStorage.ts
charlesBochet: Agreed — all swallowed errors now go through `logError` (persist failures, hydrate read, clear, broadcast) so failures are traceable instead of silent. Added a test asserting a failed persist is logged. 73d9543.

packages/twenty-front/src/modules/ui/utilities/state/jotai/utils/indexedDbKeyValStore.ts
```diff
@@ -0,0 +1,126 @@
...
+  try {
+    return typeof indexedDB !== 'undefined' && indexedDB !== null;
+  } catch {
+    // Accessing indexedDB can throw in sandboxed iframes / disabled storage.
+    return false;
+  }
+};
+
+const openDatabase = (): Promise<IDBDatabase> => {
+  if (databasePromise !== null) {
```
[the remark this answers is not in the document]
charlesBochet: Resolved by the refactor: the hand-rolled wrapper (with the cached `databasePromise`) is gone in favour of idb-keyval, which clears its cached connection on close. There's also no localStorage fallback anymore — by design, IndexedDB-unavailable/private-mode degrades to an in-memory cache that re-fetches metadata from the network each boot (metadata is server-derived, so no user-data loss). The availability guard prevents the synchronous throw when `indexedDB` is fully absent; private-mode async failures are caught and logged.

packages/twenty-front/src/modules/ui/utilities/state/jotai/utils/createIndexedDbBackedJotaiStorage.ts
```diff
@@ -0,0 +1,239 @@
...
+        void idbDelete(key);
+      } else {
+        try {
+          localStorage.removeItem(key);
+        } catch {
+          // noop
+        }
+      }
+    },
+    subscribe: (key, callback) => {
```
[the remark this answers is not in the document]
charlesBochet: Fixed — `subscribe` now stores the `initialValue` per subscriber, and the cross-tab `remove` handler notifies callbacks with that initial value (not `undefined`).

packages/twenty-front/src/modules/ui/utilities/state/jotai/utils/createIndexedDbBackedJotaiStorage.ts
```diff
@@ -0,0 +1,257 @@
...
+      }
+    },
+    removeItem: (key) => {
+      memoryMap.delete(key);
+      broadcast({ type: 'remove', key });
+
+      if (isDefined(idbStore)) {
+        persist(idbDelete(key, idbStore));
+      } else {
+        removeLocalStorageItem(key);
```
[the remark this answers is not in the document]
charlesBochet: Resolved — the legacy localStorage migration/cleanup was removed entirely. The facade no longer reads or writes localStorage, so there's no longer a delete-before-persist ordering.

packages/twenty-front/src/modules/ui/utilities/state/jotai/utils/createIndexedDbBackedJotaiStorage.ts
```diff
@@ -0,0 +1,257 @@
...
+    } catch {
+      // Value not structured-cloneable / channel closed: skip cross-tab sync.
+    }
+  };
+
+  const storage: JotaiSyncStorage<ValueType> = {
+    getItem: (key, initialValue) => {
+      const storedValue = memoryMap.get(key);
+
+      return storedValue === undefined ? initialValue : storedValue;
```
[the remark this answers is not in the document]
charlesBochet: Fixed — `getItem` uses `memoryMap.has(key)` to distinguish a missing key from a stored `undefined`, with a regression test for `ValueType` including `undefined`.

prastoin (approved):

PR #21588, opened 2026-06-15
fix(server): mitigate integration-test OOM flakiness

charlesBochet:
## Problem

`server-integration-test` shards have been failing intermittently across unrelated PRs with a distinctive signature: the shard exits code 1 with **no jest assertion failure, no `Test Suites:` summary, and no V8 `JavaScript heap out of memory` error** — the process just dies mid-run. Failures hit random shards and clear on re-run (e.g. an unrelated branch failed shard 6 once, then passed 3× on identical code), while the `merge_group` gate stays green.

### Root cause

Each shard runs a **single in-band jest process** that boots one shared NestJS app (`globalSetup` → `app.listen`) and holds it for the entire shard, driving heavy metadata migrations + cache rebuilds in that one process. `NODE_OPTIONS=--max-old-space-size=12288` let V8 grow to 12 GB — *above* the `ubuntu-latest` runner's available RAM (16 GB, shared with Postgres/Redis/ClickHouse). V8 therefore deferred aggressive GC and grew past physical memory, so the **OS OOM-killer killed the process before V8 hit its own ceiling** — which is why there's no heap error and no jest summary, just a silent exit.

## Changes (CI/test-only — prod runtime untouched)

- **Lower the integration jest heap cap `12288` → `6144`** so V8 self-limits below physical headroom instead of being OS-killed. Counterintuitively safer: a real leak now surfaces as a *visible* heap error naming the test, rather than a silent death. (`database:reset` keeps 12288 — it runs alone, before jest.)
- **Add `--logHeapUsage`** to the integration jest runs to expose the per-file heap trend for confirming/pinpointing the growth.
- **Split integration tests across 16 shards (was 10)** to lower the peak working set per shard.
- **Make perf logging a first-class `LoggerService` tool** (per @prastoin's review): add `LoggerService.perf()` and unify the existing `time()`/`timeEnd()` helpers into `perfTime()`/`perfTimeEnd()` (now routed through the driver), all gated by a new `PERF_LOG_ENABLED` config var. It **defaults on** so real environments keep emitting the install-perf logs, and `.env.test` sets it `false` to mute the per-action flood in integration tests. The `application-manifest` and `validate-build` services were moved from the built-in `Logger` to `LoggerService` to use it.

## Notes

- `--max-old-space-size` lives only in the `test:integration` nx target; it is **not** the prod server heap setting, so prod is unaffected.
- This is mitigation. If `--logHeapUsage` shows monotonic growth across files, there's a real accumulation in the long-lived app (retained flat-maps / metadata cache) worth a follow-up heap-snapshot fix.

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21588?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/twenty-server/src/engine/core-modules/application/application-manifest/application-manifest-migration.service.ts
```diff
@@ -190,9 +191,11 @@ export class ApplicationManifestMigrationService {
     );
     const recomputeMs = performance.now() - recomputeStart;
 
-    this.logger.log(
-      `[install-perf] syncMetadataFromManifest ALL_METADATA_NAME getOrRecompute flat-maps took ${recomputeMs.toFixed(1)}ms (logicFunctions=${manifest.logicFunctions.length})`,
-    );
+    if (isInstallPerfLoggingEnabled()) {
+      this.logger.log(
```
prastoin: ***Suggestion:*** Could we rather introduce a perf method in the logger directly ? such as for the debug

.github/workflows/ci-server.yaml
```diff
@@ -302,7 +302,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        shard: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+        shard: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16]
```
prastoin, line 305: savage

prastoin (approved):

packages/twenty-server/src/engine/core-modules/application/application-manifest/application-manifest-migration.service.ts
charlesBochet: good idea

packages/twenty-server/src/engine/core-modules/logger/logger.service.ts
```diff
@@ -59,17 +71,45 @@ export class LoggerService implements LoggerServiceInterface {
...
 
-  time(category: string, label: string) {
-    if (this.driver.options.logLevels?.includes('debug')) {
-      // oxlint-disable-next-line no-console
-      console.time(`[${category}] ${label}`);
+  // Performance instrumentation, gated by PERF_LOG_ENABLED so it can be muted
+  // (e.g. in integration tests) while staying on by default everywhere else.
+
+  // oxlint-disable-next-line typescript/no-explicit-any
+  perf(message: any, category: string, ...optionalParams: any[]) {
```
prastoin, line 74: unknown 👀 not important
And nice refactor

packages/twenty-server/src/engine/workspace-manager/workspace-migration/services/workspace-migration-validate-build-and-run-service.ts
```diff
@@ -369,27 +368,33 @@ export class WorkspaceMigrationValidateBuildAndRunService {
...
-    // TODO(install-perf): temporary, remove.
     const buildStart = performance.now();
     const validateAndBuildResult =
       await this.workspaceMigrationBuildOrchestratorService
         .buildWorkspaceMigration(buildArgs)
         .catch((error) => {
-          this.logger.error(error);
+          this.logger.error(
+            error,
+            WorkspaceMigrationValidateBuildAndRunService.name,
```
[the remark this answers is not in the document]
charlesBochet: False positive here — `this.logger` is our custom `LoggerService`, not the native NestJS `Logger`. Its signature is `error(message, category, ...optionalParams)`, so the 2nd arg is the log *category/context*, not a stack slot.

It also does not drop the stack at the driver level: `LoggerService.error` forwards to `ConsoleLogger.error(error, category)`, and NestJS only routes the 2nd arg into the stack slot when `isStackFormat` matches a stack-trace regex (`/^(.)+\n\s+at .+:\d+:\d+/`). A class name fails that regex, so it is treated as the context — the `error` object stays the message, exactly as in the pre-PR `this.logger.error(error)`. The change only adds the service name as context. Resolving.

PR #21591, opened 2026-06-15
fix(twenty-shared): derive short-number suffix from the rounded value

yashs33244:
`formatToShortNumber` (`packages/twenty-shared/src/utils/format/formatToShortNumber.ts`) picked the unit suffix from the **raw** value but printed the **rounded** figure, so `999999` rendered as `"1000k"` instead of `"1m"`, and `999999999` as `"1000m"` instead of `"1b"`. This affects number/currency cells, column-footer aggregates, and dashboard charts.

The fix replaces the hard-coded band branches with a promotion loop that derives the suffix from the rounded display value, so the suffix and figure always agree at boundaries. Adds boundary, just-below-boundary, and negative-boundary tests.

Red-green proven: the two new boundary tests fail on the original source (`expected "1m" but got "1000k"`); the 11 pre-existing tests still pass; all 13 pass with the fix. Verified with a standalone strict `tsc` (0 errors) and oxlint on both changed files.

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21591?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/twenty-shared/src/utils/__tests__/formatToShortNumber.test.ts
+++ b/packages/twenty-shared/src/utils/__tests__/formatToShortNumber.test.ts
@@ -30,6 +30,17 @@ describe('formatToShortNumber', () => {
     expect(formatToShortNumber(0)).toBe('0');
   });
 
+  it('promotes to the next unit when rounding reaches the boundary', () => {
+    expect(formatToShortNumber(999999)).toBe('1m');
+    expect(formatToShortNumber(999950)).toBe('1m');
+    expect(formatToShortNumber(999999999)).toBe('1b');
+  });
+
+  it('does not promote just below a rounding boundary', () => {
+    expect(formatToShortNumber(999949)).toBe('999.9k');
+    expect(formatToShortNumber(999)).toBe('999');
+  });
+
   describe('negative numbers', () => {
     it('formats negative numbers less than 1000 correctly', () => {
       expect(formatToShortNumber(-500)).toBe('-500');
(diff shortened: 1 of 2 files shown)


etiennejouan (commented):
Interesting ! Thanks for your contribution 🙏 

I would have keep the "if-return" structure that seems simpler to read.

> export const formatToShortNumber = (amount: number) => {
>   const sign = amount < 0 ? '-' : '';
>   const absoluteAmount = Math.abs(amount);
>   const format = (scaled: number, suffix: string) =>
>     sign + scaled.toFixed(1).replace(/\.?0+$/, '') + suffix;
>   if (Number(absoluteAmount.toFixed(1)) < 1000) {
>     return format(absoluteAmount, '');
>   }
>   if (Number((absoluteAmount / 1_000).toFixed(1)) < 1000) {
>     return format(absoluteAmount / 1_000, 'k');
>   }
>   if (Number((absoluteAmount / 1_000_000).toFixed(1)) < 1000) {
>     return format(absoluteAmount / 1_000_000, 'm');
>   }
>   return format(absoluteAmount / 1_000_000_000, 'b');
> };

yashs33244:
Good call, switched it back to the if-return version you posted - reads cleaner. Behavior is identical, boundary tests still pass. Thanks for the review!

etiennejouan (approved):
LGTM. Thanks

PR #21597, opened 2026-06-15
fix(twenty-front): promote formatNumber suffix on the rounded value at boundaries

yashs33244:
`formatNumber(value, { abbreviate: true })` (`packages/twenty-front/src/utils/format/formatNumber.ts`) picked the k/M/B suffix from the **raw** value but printed the **rounded** base, so at rounding boundaries the suffix and figure disagree:

- `formatNumber(999999, { abbreviate: true, decimals: 1 })` → `"1,000k"` instead of `"1M"`
- `999999999` → `"1,000M"` instead of `"1B"`
- `999.9` → `"1,000"` instead of `"1k"`

These feed billing/credit balances, AI token counts in workflow run logs, and subscription info. The fix derives the suffix from the rounded display value (promotion loop) so they always agree; adds boundary / just-below / negative tests.

Red-green proven: the 4 new boundary blocks fail on the original (`"1M" got "1,000k"`, etc.); 28/28 assertions pass with the fix; pre-existing tests stay green. Verified with `prettier --check`, oxlint, and strict `tsc` (0 errors) on the changed files.

_(Sibling to the same rounding-boundary class in `formatToShortNumber` / twenty-shared — separate function and package, hence a separate PR.)_

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21597?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


etiennejouan:
Hi @yashs33244 thank you for your contribution. For related fixes, you can open one PR. I let you gather in this one https://github.com/twentyhq/twenty/pull/21591. Thanks

etiennejouan (commented):
https://github.com/twentyhq/twenty/pull/21591#pullrequestreview-4522615758

PR #21609, opened 2026-06-15
Make flaky Storybook stories deterministic for Argos (animations + lazy-load)

bosiraphael:
Follow-up to #21594. A set of stories were still flagged flaky by Argos after the data/image/date fixes. The remaining causes are **animation** and **lazy-load timing**, not data. Argos's screenshot stabilization waits for fonts and images and disables CSS animations at capture, but it does **not** stop JS-driven animations (react-spring) or wait for `React.lazy` chunks to load.

Two global, test-only changes in the Storybook Vitest setup, plus an `optimizeDeps` entry:

**1. Disable JS animations (charts).** GraphWidgetLineChart (incl. `--catalog`) and the pie chart draw via `@react-spring/web` (`useAnimatedPath`/`useArcsTransition`, used by nivo), so they get captured mid-draw.
- `Globals.assign({ skipAnimation: true })` from `@react-spring/web` (single hoisted instance → covers nivo), mirroring the existing `MotionGlobalConfig.skipAnimations`.
- A global CSS rule zeroing animation/transition durations (belt-and-suspenders for the pre-capture window).
- `@react-spring/web` added to `optimizeDeps.include` (next to `@nivo/*`) so importing it in the setup file doesn't trigger a mid-run Vite re-optimization.

**2. Pre-warm lazy chunks.** RichTextFieldInput (BlockNote editor), AiChatMessage (`react-markdown`/`remark-gfm`), and AppNavigationDrawer (favorites/workspace section dispatchers) lazy-load behind `Suspense`; the snapshot could capture the skeleton fallback instead of the loaded content. Pre-warming the chunks at setup time (same pattern already used for the workflow-step mock) caches them before stories render.

The settings object-about form was already covered by the framer-motion/CSS animation disable.

### DIFF
--- a/packages/twenty-front/.storybook/vitest.setup.ts
+++ b/packages/twenty-front/.storybook/vitest.setup.ts
@@ -1,8 +1,19 @@
+import { Globals } from '@react-spring/web';
 import { setProjectAnnotations } from '@storybook/react-vite';
 import { MotionGlobalConfig } from 'framer-motion';
 import * as projectAnnotations from './preview';
 
 MotionGlobalConfig.skipAnimations = true;
+Globals.assign({ skipAnimation: true });
+
+const disableCssAnimationsStyle = document.createElement('style');
+disableCssAnimationsStyle.innerHTML = `*, *::before, *::after {
+  animation-duration: 0s !important;
+  animation-delay: 0s !important;
+  transition-duration: 0s !important;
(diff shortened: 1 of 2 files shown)


Weiko (approved):
LGTM

charlesBochet (approved):

PR #21611, opened 2026-06-15
Move twenty-client-sdk to dev dep

prastoin:
# Introduction
 The `twenty-client-sdk` is always provided and injected at runtime by the twenty-server instance
Which mean that even if in your app locally you're using twenty-client-sdk `1.0` installing this app on twenty instance `2.0` will result in injecting another `twenty-client-sdk`

That's the expected behavior and tradeof

The twenty-app devdep should only be used to guide local devxp following typesafety and so on

A user can still locally generated its own twenty-client-sdk and publish it if necessary

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21611?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/create-twenty-app/src/constants/template/package.json
+++ b/packages/create-twenty-app/src/constants/template/package.json
@@ -16,15 +16,14 @@
     "test": "vitest run",
     "test:watch": "vitest"
   },
-  "dependencies": {
-    "twenty-client-sdk": "TO-BE-GENERATED"
-  },
+  "dependencies": {},
   "devDependencies": {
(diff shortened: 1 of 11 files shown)


martmull (approved):
lgtm thank you

PR #21612, opened 2026-06-15
fix(workflow): serialize object variables in resolved prompts

ijreilly:
## Problem

When a workflow passes a variable into a text input (e.g. an **AI Agent** prompt) and that variable resolves to an object or array, the resolved string contained `[object Object]` instead of the actual content. The AI then received useless input.

## Cause

`resolveString` in the shared variable resolver builds the final string with `String.prototype.replace`. When an embedded `{{variable}}` resolved to an object, the replace callback returned the object directly, which JS coerces to `"[object Object]"`. The rich-text resolver had the same issue via `String(resolvedValue)`.

## Fix

When an embedded variable resolves to a non-null object (or array), serialize it with `JSON.stringify` before inserting it into the surrounding string. Primitive values keep their existing coercion behavior, and the single-variable case (`{{message}}` with nothing around it) still returns the raw object so non-string consumers are unaffected.

Applied the same guard to both the plain and rich-text variable resolvers for consistency.

## Tests

Added cases covering embedded object/array variables in both resolvers, plus a guard test confirming a standalone `{{variable}}` still returns the raw object.

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21612?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/twenty-shared/src/utils/variable-resolver.ts
+++ b/packages/twenty-shared/src/utils/variable-resolver.ts
@@ -78,6 +78,10 @@ const resolveString = (
   return input.replace(VARIABLE_PATTERN, (matchedToken, _) => {
     const processedToken = evalFromContext(matchedToken, context);
 
+    if (typeof processedToken === 'object' && processedToken !== null) {
+      return JSON.stringify(processedToken);
+    }
+
     return processedToken;
   });
 };
--- a/packages/twenty-shared/src/utils/__tests__/rich-text-variable-resolver.test.ts
+++ b/packages/twenty-shared/src/utils/__tests__/rich-text-variable-resolver.test.ts
@@ -126,6 +126,23 @@ describe('resolveRichTextVariables', () => {
     );
   });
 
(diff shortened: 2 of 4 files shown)


ijreilly:
~~Actually no issue - problem observed came from the logic function before~~

ijreilly:
I was wrong, there is an issue. JSON.Stringify() is useful when there are nested objects involved.

packages/twenty-shared/src/utils/variable-resolver.ts
etiennejouan, line 81: nit: isObject ?

etiennejouan (approved):
LGTM

PR #21626, opened 2026-06-15
fix(front): home redirect honors first object of the navigation menu

### LINKED ISSUE #21166: Bug: Home redirect always lands on first object alphabetically, ignores last-visited

thomtrp:
https://github.com/user-attachments/assets/db11f02c-4502-4adc-8cf7-55d3dc0211a6

## Summary

Fixes [#21166](https://github.com/twentyhq/twenty/issues/21166). Going to the workspace root (`/`) always redirected to the alphabetically-first object.

Following [the review discussion](https://github.com/twentyhq/twenty/issues/21166#issuecomment-4611570420), this redirects to the **first object of the navigation menu** instead of the last-visited object. "Last-visited" had too many edge cases (which page? records vs. index?), whereas the first item of the user's menu is unambiguous and matches what they see at the top of the left sidebar.

## Changes

- New `getFirstObjectNavigationMenuItemLink` util walks the workspace navigation menu items in display order (sorted by `position`) and returns the link of the first object-backed item (`OBJECT`/`VIEW`) the user can read.
- `useDefaultHomePagePath` uses it as the primary target. It stays on `AppPath.Index` until navigation menu items have loaded (they load *after* the minimal-metadata fast path, so resolving earlier would land on the wrong object during the post-login window), then falls back to the first readable object if the menu has no object item.

Scope is intentionally minimal: the last-visited tracking is left in place and will be cleaned up separately now that the redirect no longer reads it.

## Test plan

- `npx jest useDefaultHomePagePath --config=packages/twenty-front/jest.config.mjs` — covers menu order honored over alphabetical, `VIEW` item links, the loading deferrals (object metadata + navigation menu items), the readable-object fallback, and the no-readable-object profile-settings fallback.
- Manually: visit a non-first object, navigate to `/`, confirm the redirect goes to the first object of the navigation menu.

### DIFF
--- a/packages/twenty-front/src/modules/navigation/utils/getFirstNavigationMenuItemLink.ts
+++ b/packages/twenty-front/src/modules/navigation/utils/getFirstNavigationMenuItemLink.ts
@@ -8,7 +8,7 @@ import { NavigationMenuItemType } from 'twenty-shared/types';
 import { isDefined } from 'twenty-shared/utils';
 import { type NavigationMenuItem } from '~/generated-metadata/graphql';
 
-type GetFirstObjectNavigationMenuItemLinkArgs = {
+type GetFirstNavigationMenuItemLinkArgs = {
   navigationMenuItemsInDisplayOrder: NavigationMenuItem[];
   objectMetadataItems: EnrichedObjectMetadataItem[];
   views: Pick<View, 'id' | 'objectMetadataId' | 'key'>[];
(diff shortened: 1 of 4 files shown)


FelixMalfait:
I think the call that was made in the issue was to redirect to first menu item, no?

thomtrp:
> I think the call that was made in the issue was to redirect to first menu item, no?

What I understood is that the user complained about being redirected to the first object instead of the last visited

FelixMalfait:
https://github.com/twentyhq/twenty/issues/21166#issuecomment-4611570420 last visited is legacy imo ; we agree to do first menu item (simpler)

thomtrp:
oooh I missed that comment, thank you!

packages/twenty-front/src/modules/navigation/utils/getFirstNavigationMenuItemLink.ts
```diff
@@ -0,0 +1,63 @@
...
+  objectMetadataItems,
+  views,
+  objectPermissionsByObjectMetadataId,
+}: GetFirstObjectNavigationMenuItemLinkArgs): string | null => {
+  for (const item of navigationMenuItemsInDisplayOrder) {
+    if (
+      item.type !== NavigationMenuItemType.OBJECT &&
+      item.type !== NavigationMenuItemType.VIEW
+    ) {
+      continue;
```
FelixMalfait, line 37: If someone wants a dashboard as homepage I think it's a great idea and they should be able to (that's what I would want). Let's exclude 'FOLDER' | 'LINK' but support the rest?

packages/twenty-front/src/modules/navigation/utils/getFirstObjectNavigationMenuItemLink.ts
```diff
@@ -0,0 +1,63 @@
...
+      continue;
+    }
+
+    const objectMetadataItem = getObjectMetadataForNavigationMenuItem(
+      item,
+      objectMetadataItems,
+      views,
+    );
+
+    if (
```
FelixMalfait: I feel like this is the symptom of a bigger issue that multiple people reported and is worth fixing upstream: you shouldn't view in the navigation menu objects you don't have access to or views of objects you don't have access to. It should leave upstream rather than adding it multiple time downstreams like that. Would be a nice fix of an existing issue :)

FelixMalfait (approved):
Great! Sorry 2 comments that require a bit of rework

PR #21630, opened 2026-06-15
fix(billing): ack Stripe subscription webhooks for deleted workspaces

FelixMalfait:
## Problem

Stripe keeps emitting `customer.subscription.*` events for workspaces that have since been deleted. `BillingWebhookSubscriptionService.processStripeEvent` threw `BILLING_SUBSCRIPTION_EVENT_WORKSPACE_NOT_FOUND` for those, which `getBillingExceptionStatusCode` maps to HTTP **500**. Stripe treats a 5xx as a failed delivery and **retries with backoff for ~3 days**, so a handful of deleted workspaces turn into a steady ~45k Sentry errors (and as many redundant webhook deliveries).

This is the top error stream by volume in `twenty-server` — Sentry `TWENTY-SERVER-EQP` (45k+ events).

## Fix

When the workspace is missing (or soft-deleted and the event isn't a subscription deletion), log a warning and return instead of throwing. There's nothing to process for a workspace that no longer exists; returning lets the controller ack the webhook with `200`, which stops both the Stripe retry loop and the error noise.

Also drops the now-unused `msg` import.

## Impact

- Eliminates ~45k/month Sentry errors that were masking real signal.
- Stops the redundant Stripe webhook retries for deleted workspaces.
- No behavior change for live workspaces.

Fixes TWENTY-SERVER-EQP

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21630?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-subscription.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-subscription.service.ts
@@ -3,7 +3,6 @@
 import { Injectable, Logger } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
 
-import { msg } from '@lingui/core/macro';
 import { isDefined } from 'twenty-shared/utils';
 import { WorkspaceActivationStatus } from 'twenty-shared/workspace';
 import { In, Repository } from 'typeorm';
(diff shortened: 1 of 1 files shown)


packages/twenty-server/src/engine/core-modules/billing-webhook/services/billing-webhook-subscription.service.ts
```diff
@@ -80,26 +79,22 @@ export class BillingWebhookSubscriptionService {
...
 
     if (!workspace) {
-      throw new BillingException(
-        `Workspace not found for subscription event ${event.id} / workspaceId: ${workspaceId}`,
-        BillingExceptionCode.BILLING_SUBSCRIPTION_EVENT_WORKSPACE_NOT_FOUND,
-        {
-          userFriendlyMessage: msg`Workspace ${workspaceId} is not found.`,
-        },
+      this.logger.warn(
+        `Skipping subscription event ${event.id}: workspace ${workspaceId} not found`,
```
FelixMalfait, line 83: Not sure if we want to use logger or not log at all?

FelixMalfait:
Closing — defensive patch, root cause unconfirmed. Events for deleted workspaces are either (a) the expected echo of our own cancellation (we cancel the Stripe sub on delete; the webhooks land after the row is gone) → ack is correct, or (b) live orphaned subscriptions from out-of-band deletes still billing customers → ack would mask a real billing leak. Distinguishing requires a Stripe audit (active subs whose metadata.workspaceId has no live workspace), which I can't run from here. The real next step is that reconciliation audit, not this ack.

PR #21638, opened 2026-06-15
fix(record-field): prevent state clobbering and focus loss in dynamic array inputs

### LINKED ISSUE #21544: fix(record-field): prevent state clobbering in dynamic array inputs

lakshitsoni26:
Closes #21544

# Description

This PR fixes a critical React reconciliation bug where dynamic array items in record fields lost state and focus when an item in the middle of the array was deleted. This was caused by the anti-pattern of using the array map `index` as the React `key`.

### Root Cause & Technical Implementation
To accurately track DOM node additions and deletions, I replaced the index keys with stable UUIDs across the affected components while strictly adhering to Twenty technical standards (no ternaries or inline logical operators in JSX, no `useEffect` for derived state).

- **`FormArrayFieldInput.tsx`**: Implemented a parallel `stableIds` state synchronized flawlessly with `draftValue` additions and deletions using standard uncontrolled component event handlers. Extracted all complex inline logic into explicit render functions.
- **`ArrayFieldInput.tsx`**: Used the React-recommended `useState` render-phase pattern to deterministically track and update `stableIds` based on the derived `draftValue` length without `useEffect` or mutating data contracts. 

### How to Test / Verification

1. Navigate to any object record in the CRM that utilizes a dynamic array field.
2. Add 3 distinct items to the array input (e.g., "Item A", "Item B", "Item C").
3. Click into the input field for "Item C" to ensure it has browser focus.
4. Delete the middle item ("Item B").
5. **Verify:** "Item C" should shift up, retain its exact value, and React should not lose focus or overwrite the component state.

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21638?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/twenty-front/src/modules/object-record/record-field/ui/form-types/components/FormArrayFieldInput.tsx
+++ b/packages/twenty-front/src/modules/object-record/record-field/ui/form-types/components/FormArrayFieldInput.tsx
@@ -34,6 +34,7 @@ import {
 } from 'twenty-ui-deprecated/theme-constants';
 import { MenuItem } from 'twenty-ui-deprecated/navigation';
 import { toSpliced } from '~/utils/array/toSpliced';
+import { v4 } from 'uuid';
 
 type FormArrayFieldInputProps = {
   label?: string;
@@ -128,11 +129,22 @@ export const FormArrayFieldInput = ({
         },
   );
 
+  const [stableIds, setStableIds] = useState<string[]>(() => {
+    if (isStandaloneVariableString(defaultValue)) {
+      return [];
+    }
+    const initialItems =
+      isDefined(defaultValue) && isNonEmptyArray(defaultValue)
+        ? defaultValue
+        : [];
+    return initialItems.map(() => v4());
+  });
+
   const [newItemDraftValue, setNewItemDraftValue] = useState('');
   const [inputValue, setInputValue] = useState('');
   const [isInputDisplayed, setIsInputDisplayed] = useState(false);
-  const [itemToEditIndex, setItemToEditIndex] = useState(-1);
-  const isAddingNewItem = itemToEditIndex === -1;
+  const [itemToEditId, setItemToEditId] = useState<string | null>(null);
+  const isAddingNewItem = itemToEditId === null;
 
   const containerRef = useRef<HTMLDivElement>(null);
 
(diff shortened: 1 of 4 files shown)


thomtrp:
Hi, could you provide a Before/after videos?

lakshitsoni26:
Hi @thomtrp,

I unfortunately don't have screen recordings available for this, as the "Before" state is highly intermittent and difficult to capture consistently on demand. 

To give you some technical context on the bug: because the component currently uses `key={index}`, it suffers from a React reconciliation race condition. It doesn't trigger during standard linear data entry, but surfaces when mutating or unmounting items from the middle of the array while the parent component concurrently triggers a re-render. When this happens, React's diffing algorithm misidentifies the component instances, leading to state clobbering and sudden focus loss.

Even without a video demonstrating the specific edge-case, migrating away from `key={index}` to stable, unique IDs is a standard React best practice to prevent exactly these types of reconciliation bugs in dynamic, editable lists. 

Let me know if this context helps and if we're good to proceed!

charlesBochet:
thanks for the PR @lakshitsoni26 

See comment on the issue, I'm going to close this one. The help is much appreciated though :)

PR #21640, opened 2026-06-15
feat(onboarding): prefill the invite step with teammates from the connected calendar

FelixMalfait:
## What & why

Implements core-team-issues#1414: move the calendar/email connection earlier in onboarding and use the freshly connected calendar to prefill the **Invite your team** step with likely teammates, so users don't start from an empty form.

## Approach

Everything is behind the feature flag `IS_ONBOARDING_INVITE_SUGGESTIONS_ENABLED` (off by default).

**Reorder** — onboarding becomes `Workspace activation → Connect account → Create profile → Invite team`. Connecting before profile gives the calendar sync a head start; connecting *before* the workspace exists isn't possible (a connected account requires an activated workspace + workspace member + OAuth transient token). Gated in both `OnboardingService.getOnboardingStatus` (backend) and `useSetNextOnboardingStatus` (frontend) so the two agree.

**Fast teammate lookup** — on Google/Microsoft connect *during onboarding*, a background job (`FetchOnboardingInviteSuggestionsJob`) runs a single bounded calendar fetch (recent events, attendees inline), keeps same-work-email-domain colleagues (excludes self + aliases; personal mailboxes yield nothing), ranks by meeting frequency, and caches the top 5. The invite step reads the cache via a new `getInviteSuggestions` query and prefills the form — polling briefly while the cache warms, and never overwriting input the user has already typed.

Providers: **Google** (Calendar `events.list`) and **Microsoft** (Graph `calendarView`), routed by a `CalendarAttendeesService` dispatcher (mirrors the existing `CalendarGetCalendarEventsService`). Any fetch failure (missing scope, API error) degrades to today's empty form via the orchestrator's best-effort catch.

## How to enable

Turn on `IS_ONBOARDING_INVITE_SUGGESTIONS_ENABLED` for a workspace (admin panel).

## Notes

- New-workspace creators only (invitees never see the connect/invite steps). Skipping the connect step, or signing up with a personal email, falls back to the current empty form.
- The "We found teammates from your calendar" subtitle only shows once suggestions are actually prefilled.
- i18n: the new `<Trans>` strings are extracted on merge to `main` by the existing Crowdin workflow.

## Testing

- Frontend unit tests for the reorder state machine (both flag states).
- `npx nx typecheck` and `npx nx lint:diff-with-main` green for `twenty-front` and `twenty-server`.
- Server boots with the new DI wiring (no circular dependency); `getInviteSuggestions` / `InviteSuggestion` present in the live metadata schema.
- Not exercised in CI: live Google/Microsoft OAuth end-to-end (requires real accounts + calendar data).

https://claude.ai/code/session_019MyY3bfAEij4AwSXMCLtWY

---
_Generated by [Claude Code](https://claude.ai/code/session_019MyY3bfAEij4AwSXMCLtWY)_

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21640?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/twenty-client-sdk/src/metadata/generated/schema.graphql
```diff
@@ -1617,6 +1617,14 @@ type BillingUpdate {
   billingSubscriptions: [BillingSubscription!]!
 }
 
+type InviteSuggestion {
+  """Email address of a suggested teammate to invite"""
```
FelixMalfait: I believe we don't usually add this comments. TBC if we accept doing it because it's nice or we want to stay consistent (I think I'd advocate for staying consistent for now)

packages/twenty-front/src/pages/onboarding/InviteTeam.tsx
```diff
@@ -88,6 +98,70 @@ export const InviteTeam = () => {
...
+  );
+  const hasInviteSuggestions = inviteSuggestions.length > 0;
+
+  useEffect(() => {
+    if (hasInviteSuggestions) {
+      setSuggestionsPollIntervalMs(0);
+    }
+  }, [hasInviteSuggestions]);
+
+  useEffect(() => {
```
FelixMalfait: Not too convinced about the useEffect and polling. If we get a screen/step to be in-between email connection and team invite it should give us enough time. And if there isn't enough time then too bad for us but no need for polling. I guess our only option is to ask for the connection before we ask for the profile name?

packages/twenty-front/src/modules/onboarding/hooks/useSetNextOnboardingStatus.ts
```diff
@@ -24,41 +26,59 @@ type GetNextOnboardingStatusArgs = {
...
 }: GetNextOnboardingStatusArgs) => {
-  if (currentUser?.onboardingStatus === OnboardingStatus.WORKSPACE_ACTIVATION) {
+  const onboardingStatus = currentUser?.onboardingStatus;
+  const isSoloWorkspace = currentWorkspace?.workspaceMembersCount === 1;
+
+  if (onboardingStatus === OnboardingStatus.WORKSPACE_ACTIVATION) {
+    if (isInviteSuggestionsEnabled && isAccountSyncEnabled) {
+      return OnboardingStatus.SYNC_EMAIL;
+    }
     return OnboardingStatus.PROFILE_CREATION;
```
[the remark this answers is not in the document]
FelixMalfait: I don't understand the reference to CONNECTED_ACCOUNTS permissions? This InviteTeam should be only for the first user creating a workspace which has to be an admin so no permission enter into account here?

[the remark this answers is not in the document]
FelixMalfait: You're right — removed it in `531c022`. Onboarding only runs for the first user (the workspace creator, who is an admin), so `CONNECTED_ACCOUNTS` is always granted and the check was meaningless here. It also diverged from the backend, which routes on the feature flag alone — so I aligned by dropping the permission check on the frontend (rather than adding one on the backend, which the bots suggested but contradicts your point). The transition now depends only on the feature flag.

---

packages/twenty-front/src/pages/onboarding/InviteTeam.tsx
FelixMalfait: Agreed — dropped the polling in `531c022`. The flow now asks for the connection before the profile step (`WORKSPACE_ACTIVATION → SYNC_EMAIL → PROFILE_CREATION → INVITE_TEAM`), so the profile step is exactly the in-between screen that gives the background fetch time. The invite step now just fetches once and falls back to the empty form if suggestions aren't ready yet — no `useEffect` polling.

---

packages/twenty-client-sdk/src/metadata/generated/schema.graphql
FelixMalfait: Done in `ebaec50c` — removed the field descriptions and regenerated the metadata client to stay consistent with the rest of the schema.

---

packages/twenty-server/src/modules/onboarding-invite-suggestions/jobs/fetch-onboarding-invite-suggestions.job.ts
```diff
@@ -0,0 +1,26 @@
...
+  constructor(
+    private readonly onboardingInviteSuggestionsService: OnboardingInviteSuggestionsService,
+  ) {}
+
+  @Process(FetchOnboardingInviteSuggestionsJob.name)
+  async handle(data: FetchOnboardingInviteSuggestionsJobData): Promise<void> {
+    await this.onboardingInviteSuggestionsService.computeAndCacheSuggestions(
+      data,
+    );
+  }
```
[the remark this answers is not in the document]
FelixMalfait: Verified this is a false positive — no workspace ORM context is needed here, so no change.

`getClient()` reads the connected account from the **core** `connectedAccount` table (a plain TypeORM repo — no workspace context), then threads `connectedAccount.workspaceId` *explicitly* through `resolveTokens(account, workspaceId)` → `decrypt({ ciphertext, workspaceId })` → `SecretEncryptionService.decryptVersioned(ciphertext, { workspaceId })`. The encryption derives its key from the passed `workspaceId`; it never reads from an ambient `executeInWorkspaceContext`.

The calendar/messaging jobs use `Scope.REQUEST` + `executeInWorkspaceContext` because they read/write **workspace-schema** entities (channels, events, participants). This job touches none of those — only the core connected account, the external Calendar/Graph API, and the Redis cache — so it runs correctly at default scope and decryption succeeds via the explicit `workspaceId`. (Token decryption working off `connectedAccount.workspaceId` is the same path the existing calendar sync uses to build its client.)

---

packages/twenty-server/src/modules/onboarding-invite-suggestions/constants/onboarding-invite-suggestions.constants.ts
```diff
@@ -0,0 +1,13 @@
+export const ONBOARDING_INVITE_SUGGESTIONS_CACHE_TTL_MS = 60 * 60 * 1000;
```
FelixMalfait: Let's do one export per file

neo773 (approved):
LGTM

packages/twenty-server/src/modules/onboarding-invite-suggestions/constants/onboarding-invite-suggestions.constants.ts
FelixMalfait: @claude please fix it and push to this branch. Make sure it doesn't break any CI (lint, typecheck)

PR #21641, opened 2026-06-15
Let users pick their workspace subdomain during sign-up

FelixMalfait:
## What & why

During onboarding the workspace subdomain was auto-generated at sign-up and only editable later in Settings. This adds a subdomain picker to the workspace-creation flow, with **live availability checking** and **name-driven auto-fill**.

The subdomain is chosen **on the central sign-up domain, before the redirect onto the workspace subdomain** — so there's no mid-onboarding domain switch (which would otherwise force a re-auth, like the Settings "this logs everyone out" flow). It works uniformly for credentials and SSO, since workspace creation is a post-auth mutation.

## Flow

Authenticate → **Create a workspace** → new step (workspace name + address with live availability + auto-fill, seeded from the work email) → workspace is created with the chosen subdomain → the single redirect lands on the final subdomain → onboarding modal (name pre-filled).

## Changes

**twenty-shared**
- `getSubdomainSlugFromDisplayName` — friendly slug from a display name, built on the existing `transliteration` package (also transliterates non-Latin names, e.g. 日本語 → `ri-ben-yu`).

**twenty-server**
- `checkWorkspaceSubdomainAvailability(subdomain)` query (workspace-agnostic, `UserAuthGuard`) → `{ isValid, available, suggestedSubdomain }`.
- `SubdomainManagerService`: availability + suggestion logic with friendly numbered suffixes (`acme`, `acme-2`, …) instead of random hex; `generateSubdomain` reuses it.
- `signUpInNewWorkspace` accepts an optional `{ displayName, subdomain }` input (validated; falls back to auto-generation when omitted — backward compatible, so existing callers are unaffected). Concurrent same-subdomain sign-ups return a clear "already taken" error instead of a generic DB error.

**twenty-front**
- New `SignInUpStep.WorkspaceCreation` step + `useWorkspaceSubdomainField` hook (debounced, stale-response-safe; auto-fills from the name until the user edits it, with a one-click "use suggested" when taken; ignores Enter during IME composition; surfaces a clear error if the availability check fails).
- Onboarding modal name pre-filled from the chosen name.

## Testing

- Unit tests: shared slug util, the `useWorkspaceSubdomainField` hook (real auto-fill/availability flows via `MockedProvider`), and the workspace-creation component; existing sign-up tests still pass.
- Typecheck, lint, and format green across twenty-shared / twenty-server / twenty-front.

## Notes / out of scope

- No DB migration — the `subdomain` column already existed.
- Self-hosted single-workspace sign-up is unchanged; the step is gated to multi-workspace (global scope).
- Low-priority follow-ups: length bounds on the subdomain / displayName inputs, and an integration test for the availability query.


packages/twenty-server/src/engine/core-modules/auth/services/sign-in-up.service.ts
```diff
@@ -613,6 +633,23 @@ export class SignInUpService {
...
+    } catch (error) {
+      const pgErrorCode =
+        error instanceof QueryFailedError
+          ? ((error as { code?: string }).code ??
+            (error as { driverError?: { code?: string } }).driverError?.code)
+          : undefined;
+
+      // Translate a concurrent unique-constraint violation on the chosen
+      // subdomain into the same error the upfront check throws.
+      if (isDefined(requestedSubdomain) && pgErrorCode === '23505') {
```
FelixMalfait: (unlikely to happen / safety)

packages/twenty-front/src/modules/auth/sign-in-up/hooks/useWorkspaceSubdomainField.ts
```diff
@@ -0,0 +1,253 @@
...
+  | 'idle'
+  | 'checking'
+  | 'available'
+  | 'unavailable'
+  | 'invalid'
+  | 'error';
+
+const AVAILABILITY_CHECK_DEBOUNCE_MS = 400;
+
+const FREE_EMAIL_DOMAIN_LABELS = new Set([
```
FelixMalfait: It does not make sense to have this on the frontend. We had a much better list on the backend. For me the backend should be in charge of extracting the Display Name, domain name, etc from the email. Not sure if it can be done earlier for performance (to avoid having it show up as you've already loaded the page). We need to refactor this, do it cleanly.

FelixMalfait: Refactored as discussed:

- The backend now owns the extraction. A new `getWorkspaceCreationDefaults` query returns both the display name and an available subdomain derived from the work email (reusing `SubdomainManagerService`), so there's no provider/domain list left on the frontend.
- To avoid the post-load flash, the frontend preloads that query on the workspace-selection screen (right after sign-in), and the hook seeds both fields synchronously from the Apollo cache when the creation step opens — pre-filled, no pop-in.
- Per your other note about avoiding `useEffect`: the hook is now effect-free and ref-free. Editing the name re-derives an available address through a single debounced check until you take manual control of the address field.

Also added a hook-level test that drives the autofill/availability/seed flow directly (the component test keeps mocking the hook). Pushed to the branch.

---

packages/twenty-server/src/engine/core-modules/auth/services/sign-in-up.service.ts
```diff
@@ -613,6 +633,23 @@ export class SignInUpService {
...
+        throw new WorkspaceException(
```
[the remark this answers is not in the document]
FelixMalfait: This `catch` is gated on `requestedSubdomain`, i.e. the new-workspace path, and the workspace is created fresh inside the same transaction — so `workspaceId` is brand new and the `UserWorkspaceEntity (userId, workspaceId)` constraint can't collide here. The "joining a workspace you're already in" scenario can't reach this code path. The only realistic `23505` when a chosen subdomain is present is the TOCTOU race on the workspace `subdomain`, which is exactly what this translates.

It's an intentional safety net for that unlikely race (matching Felix's note above), so I've kept it broad rather than coupling the message to a specific constraint name — which would silently break the net if the index is ever renamed. Leaving as-is.

---

packages/twenty-front/src/modules/auth/sign-in-up/hooks/useWorkspaceSubdomainField.ts
```diff
@@ -0,0 +1,190 @@
...
+      }
+
+      // Autofill always lands on an address the backend confirmed is free.
+      if (adoptSuggestion) {
+        setSubdomain(result.suggestedSubdomain);
+        setStatus('available');
+        setErrorMessage(undefined);
+        setSuggestion(undefined);
+        return;
+      }
```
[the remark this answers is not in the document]
FelixMalfait: Good catch — fixed. I re-introduced an `AbortController` that cancels the previous in-flight check before starting a new one, so an earlier/slower response can no longer overwrite a newer one: the superseded request is aborted and its response ignored (`signal.aborted` guard), and clearing/invalidating the field aborts too. The latest request always wins.

I'd dropped this ref in the earlier refactor only because `no-state-useref` flags it — but that rule has a sanctioned escape for non-state imperative handles (e.g. `useApolloFactory` does the same for its `ApolloFactory` ref), so I used `// oxlint-disable-next-line twenty/no-state-useref`. An `AbortController` is a cancellation handle, not render state, so the disable is appropriate and the hook stays effect-free. Pushed.

---

packages/twenty-front/src/modules/auth/sign-in-up/hooks/useWorkspaceSubdomainField.ts
```diff
@@ -0,0 +1,209 @@
...
+
+      // Autofill always lands on an address the backend confirmed is free.
+      if (adoptSuggestion) {
+        setSubdomain(result.suggestedSubdomain);
+        setStatus('available');
+        setErrorMessage(undefined);
+        setSuggestion(undefined);
+        return;
```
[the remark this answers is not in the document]
FelixMalfait: This is a false positive — `getSubdomainSlugFromDisplayName` already returns lowercase. It uses `transliteration`'s `slugify`, which lowercases by default; the `allowedChars: 'a-zA-Z0-9'` only lets uppercase letters survive the character filter (so the first letter isn't dropped) *before* slugify lowercases the result:

```
slugify('Apple')      -> 'apple'
slugify('ACME')       -> 'acme'
slugify('Évian Corp') -> 'evian-corp'
```

The hook test also asserts `handleWorkspaceNameChange('Apple')` resolves to `'apple'` (the mock is keyed on the lowercase value) and passes. So the server always receives a valid lowercase slug — no rejection, no random fallback. No change needed.

---

packages/twenty-server/src/engine/core-modules/auth/auth.resolver.ts
```diff
@@ -499,31 +505,65 @@ export class AuthResolver {
...
+          id: workspace.id,
+          workspaceUrls:
+            this.workspaceDomainsService.getWorkspaceUrls(workspace),
+        },
+      };
+    } catch (error) {
+      // Surface subdomain conflicts (invalid / already-taken, including a
+      // concurrent claim) as a clean GraphQL error; rethrow the rest so the
+      // resolver's auth exception filters handle them.
+      return workspaceGraphqlApiExceptionHandler(error);
```
FelixMalfait: This doesn't seem like a traditional pattern. It should already be catched by our exception handler without this no?

packages/twenty-server/src/engine/core-modules/domain/subdomain-manager/services/subdomain-manager.service.ts
```diff
@@ -30,27 +39,79 @@ export class SubdomainManagerService {
     userEmail?: string;
     workspaceDisplayName?: string;
   }) {
-    const subdomainFromUserEmail = getSubdomainFromEmail(userEmail);
-    const subdomainFromWorkspaceDisplayName =
-      getSubdomainNameFromDisplayName(workspaceDisplayName);
-
+    // findAvailableSubdomain normalizes its input and falls back to a random
```
FelixMalfait: Remove

packages/twenty-front/src/modules/auth/hooks/useAuth.ts
```diff
@@ -121,6 +127,61 @@ export const useAuth = () => {
     [setTokenPair],
   );
 
+  // Shared post-auth routing for the central sign-up domain, used by both
```
FelixMalfait: Remove comment

packages/twenty-front/src/modules/auth/sign-in-up/hooks/useWorkspaceSubdomainField.ts
```diff
@@ -0,0 +1,190 @@
...
+  | 'invalid'
+  | 'error';
+
+const AVAILABILITY_CHECK_DEBOUNCE_MS = 400;
+
+// Powers the onboarding workspace-creation step. The name and address are
+// derived from the user's work email by the backend and preloaded into the
+// Apollo cache after sign-in, so this hook seeds both fields synchronously
+// without a flash. Editing the name keeps refreshing an available address
+// until the user takes manual control of the address field.
```
martmull: try to remove comments, you can claude.md this so it only add essential WHy comments

packages/twenty-shared/src/utils/subdomain/getSubdomainSlugFromDisplayName.ts
```diff
@@ -0,0 +1,24 @@
+import { slugify } from 'transliteration';
+
+import { isDefined } from '@/utils/validation/isDefined';
+
+const SUBDOMAIN_MIN_LENGTH = 3;
+const SUBDOMAIN_MAX_LENGTH = 30;
+
+export const getSubdomainSlugFromDisplayName = (
```
martmull, line 8: don"t this already exist?

FelixMalfait: It's not a duplicate — this replaces the old server-only `getSubdomainNameFromDisplayName` (regex-based) that this PR removes, moved to `twenty-shared` because both the backend (`generateSubdomain` / `findAvailableSubdomain`) and the frontend autofill need it now. The other `slugify` helpers produce different output and aren't reusable here: `computeMetadataNameFromLabel` → camelCase identifiers (separator `_`), `computeOptionValueFromLabel` → `UPPERCASE` — neither yields a 3–30-char hyphenated subdomain.

---

FelixMalfait:
https://github.com/user-attachments/assets/3604de8c-1baf-4b94-a217-23aa6beb4cbc


@martmull

PR #21648, opened 2026-06-16
ci: migrate cross-repo dispatch senders to workflow_dispatch (actions:write)

prastoin:
# Introduction
Getting rid of the fine grained PAT used to dispatch to internal repositories.
Repo dispatch requires the contents write permissions which is too wide for such use
Refactored all senders and target to pass through a workflow dispatch instead
Creating a centralize app that forges a token with actions: write only provided permissions to mitigate any token exfiltrations

### DIFF
--- a/.github/workflows/cd-deploy-tag.yaml
+++ b/.github/workflows/cd-deploy-tag.yaml
@@ -7,7 +7,6 @@ on:
   push:
     tags:
       - 'twenty/v*'
-      - 'sdk/v*'
 
 defaults:
   run:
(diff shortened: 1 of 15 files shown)


.github/workflows/external-contributor-pr-auto-draft.yaml
```diff
@@ -16,13 +16,20 @@ jobs:
...
+          client-id: ${{ vars.CI_BOT_PUBLIC_CLIENT_ID }}
+          private-key: ${{ secrets.CI_BOT_PUBLIC_APP_PRIVATE_KEY }}
+          owner: twentyhq
+          repositories: ci-privileged
+          permission-actions: write
+
       - name: Dispatch to ci-privileged
         env:
-          GH_TOKEN: ${{ secrets.CI_PRIVILEGED_DISPATCH_TOKEN }}
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
```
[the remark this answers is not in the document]
prastoin: false positive

.github/workflows/cd-deploy-tag.yaml
```diff
@@ -7,7 +7,6 @@ on:
   push:
     tags:
       - 'twenty/v*'
-      - 'sdk/v*'
```
Weiko, line 10: Why this change?

Weiko (approved):

.github/workflows/cd-deploy-tag.yaml
prastoin: It's not used for the moment at all

prastoin: Also we don't have a staging version for the version to be published too atm neither

PR #21654, opened 2026-06-16
feat(ai-agent): suggest similar tool names when tool discovery misses

FelixMalfait:
## Context

When the in-product AI agent guesses a tool name that doesn't exist, the discovery tools dead-end it with no way to recover.

The most common failure is a singular/plural slip — e.g. the agent tries `group_by_cloud_user` when the real tool is `group_by_cloud_users` (read/bulk tools are always plural; only `find_one_*` is singular). Today both `learn_tools` and `execute_tool` reply with a flat `Could not find: <name>` and no suggestion, so the agent burns turns guessing or gives up.

## Change

- Add a `findSimilarToolNames` util that ranks catalog tool names against the missed name by Levenshtein distance (reusing the existing `getEditDistance`), with a small bonus for a shared `<operation>_` prefix so the correct same-operation plural is ranked first rather than a closer-but-different operation (e.g. `find_many_person` → `find_many_people`, not `find_one_person`).
- `learn_tools`: when names aren't found, include `suggestions` in the structured result and inline them in the message — `Could not find: group_by_person (did you mean: group_by_people?).`
- `execute_tool` (via `ToolRegistryService.resolveAndExecute`): append `Did you mean: …?` to the not-found error, reusing the catalog it already fetched (no extra lookup).

The heuristic mirrors the existing workflow variable-path suggestion util (same edit-distance threshold), so behavior is consistent with that prior art.

## Tests

- Unit tests for `findSimilarToolNames`: plural recovery, prefix-aware ranking, distance threshold, 3-suggestion cap, empty catalog.
- `learn_tools` tool tests: suggestions surfaced on a miss; no suggestion lookup when all names resolve.

`nx typecheck twenty-server` passes; `oxlint --type-aware` and `oxfmt --check` are clean on the changed files.

https://claude.ai/code/session_01GMjZkJYkqTJogJJTM6AAV8

---
_Generated by [Claude Code](https://claude.ai/code/session_01GMjZkJYkqTJogJJTM6AAV8)_

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21654?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/twenty-server/src/engine/core-modules/tool-provider/services/tool-registry.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/tool-provider/services/tool-registry.service.ts
@@ -242,8 +242,6 @@ export class ToolRegistryService {
     });
   }
 
-  // Best-effort "did you mean" suggestions for tool names that did not resolve,
-  // so a mistyped name (often a singular/plural slip) does not dead-end the agent.
   async suggestSimilarToolNames(
     toolNames: string[],
     context: ToolContext,
--- a/packages/twenty-server/src/engine/core-modules/tool-provider/services/tool-registry.service.ts
+++ b/packages/twenty-server/src/engine/core-modules/tool-provider/services/tool-registry.service.ts
@@ -13,6 +13,7 @@ import { type LearnToolsAspect } from 'src/engine/core-modules/tool-provider/too
 import { type ToolContext } from 'src/engine/core-modules/tool-provider/types/tool-context.type';
 import { type ToolDescriptor } from 'src/engine/core-modules/tool-provider/types/tool-descriptor.type';
 import { type ToolIndexEntry } from 'src/engine/core-modules/tool-provider/types/tool-index-entry.type';
+import { findSimilarToolNames } from 'src/engine/core-modules/tool-provider/utils/find-similar-tool-names.util';
 import { wrapWithErrorHandler } from 'src/engine/core-modules/tool-provider/utils/tool-error.util';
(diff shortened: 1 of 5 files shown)


packages/twenty-server/src/engine/core-modules/tool-provider/services/tool-registry.service.ts
```diff
@@ -241,6 +242,31 @@ export class ToolRegistryService {
     });
   }
 
+  async suggestSimilarToolNames(
```
ijreilly, line 245: Did we not have a pre-prompt that said don't guess tool names, refer to tool catalogs ? 
Or have we stopped caching the full list of tools because it was too long ?

FelixMalfait: The list of tools is injected into the system prompt for the AI chat but @etiennejouan refactored it in a way where it's not as explicit as before (e.g. find_many_people not explicitely listed) so this will help AI fix its mistakes rather than abandoning (which is what I observed when testing)

charlesBochet:
Please remove Claude from commit (rewrite history)

etiennejouan (approved):
LGTM

PR #21659, opened 2026-06-16
Add Recall webhook status handler to the meeting bot app

ehconitin:
Adds a `recall-webhook` logic function (`POST /webhook/recall`, unauthenticated) to the `twenty-meeting-bot` app. It verifies the Recall/Svix `whsec_` signature over the raw body, parses bot lifecycle events, matches the corresponding `CallRecording` (by `twentyCallRecordingId` metadata, falling back to `externalBotId`), and updates lifecycle fields — `status`, `externalBotId`, `externalRecordingId`, and `startedAt`/`endedAt` (only when unset) — guarded against stale out-of-order events that would move the status backwards. Adds the required `RECALL_WEBHOOK_SECRET` server variable.

This opens the real provider test path: install the app → schedule a bot through the existing calendar-event flow → point a Recall webhook endpoint at Twenty → bot lifecycle events update the matching `CallRecording`.

Unit tests cover signature verification, status mapping, the downgrade guard, metadata/bot-id matching, and timestamp fill.

Deferred to later PRs:
- transcript/media ingestion, file uploads, and the completion charge (so `COMPLETED` is never set here)
- repair/reconcile cron jobs

Also flips `DEFAULT_RECALL_REGION` to `eu-central-1` (separate commit) to match the Recall account region.

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21659?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/twenty-apps/internal/twenty-meeting-bot/src/logic-functions/flows/handle-recall-webhook.util.ts
```diff
@@ -0,0 +1,193 @@
...
+      fromStatus: callRecording.status,
+      toStatus: callRecordingStatus,
+    })
+  ) {
+    return {
+      status: 'skipped',
+      event,
+      reason: `stale status event (${callRecording.status} -> ${callRecordingStatus})`,
+    };
+  }
```
martmull, line 87: why do we skip downgrades ? Why do we have downgrade? Is it about webhook events not being handled in the same order they are emitted?

packages/twenty-apps/internal/twenty-meeting-bot/src/logic-functions/recall-api/parse-recall-webhook-event.util.ts
```diff
@@ -0,0 +1,98 @@
...
+  data,
+  bot,
+}: {
+  data: Record<string, unknown> | undefined;
+  bot: Record<string, unknown> | undefined;
+}): string | undefined => {
+  const metadata =
+    asRecord(bot?.metadata) ??
+    asRecord(getRecordAtPath(data, ['bot', 'metadata'])) ??
+    asRecord(getRecordAtPath(data, ['recording', 'metadata'])) ??
```
martmull, line 94: **Remark** we will probably need a workspaceId in metadata

martmull (approved):
That is really nice, i think we should add documentation about how to configure the webhook (in the readme.md)
also, it might be a good idea to save the workspaceId in the recall event metadata now even if we don't need it today

packages/twenty-apps/internal/twenty-meeting-bot/src/logic-functions/recall-api/parse-recall-webhook-event.util.ts
ehconitin: yes! coming in seperate PR

packages/twenty-apps/internal/twenty-meeting-bot/src/logic-functions/flows/handle-recall-webhook.util.ts
ehconitin: yes you guessed it right!

PR #21660, opened 2026-06-16
feat(website - self hosts billing): add internal endpoint to reissue enterprise keys

ijreilly:
## Summary

Adds an internal support endpoint to regenerate a customer's enterprise key
when they've lost the one issued at activation. The key payload is deterministic
from the Stripe subscription, so this re-emits an equivalent valid key without
any new state.

`GET /api/enterprise/reissue/<subscriptionId>/<secret>`

- Guarded by a shared secret (`ENTERPRISE_ADMIN_API_SECRET`), compared in
  constant time and fail-closed when unset.
- Looks up the subscription in Stripe (for the licensee) and signs the key with
  `signEnterpriseKey()`, reading `ENTERPRISE_JWT_PRIVATE_KEY` from the
  environment — the private key is never accepted from the request.
- No subscription-status gate: the key alone grants nothing. Feature access
  still requires a validity token, which `/api/enterprise/validate` only issues
  after re-checking the subscription is active.

## Notes / follow-ups

- The admin secret travels in the URL path, so it can land in server/proxy/CDN
  access logs — rotate `ENTERPRISE_ADMIN_API_SECRET` if logs are ever exposed.
- No audit logging yet; worth adding (who reissued which subscription, when).

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21660?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/twenty-website/.env.example
+++ b/packages/twenty-website/.env.example
@@ -36,6 +36,10 @@ ENTERPRISE_JWT_PUBLIC_KEY=
 # Optional: short-lived validity token length in days (default 30)
 # ENTERPRISE_VALIDITY_TOKEN_DURATION_DAYS=
 
+# Shared secret guarding the internal enterprise key reissue support endpoint,
+# used to regenerate an enterprise key
+ENTERPRISE_ADMIN_API_SECRET=
+
 # Twenty workspace the partners marketplace reads partner data from
 # (server-side only) via the /s/partners REST endpoint.
 TWENTY_PARTNERS_API_URL=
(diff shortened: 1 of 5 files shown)


charlesBochet:
Can we switch this to a POST and pass the secret in an Authorization header or the body right now it flows in the URL path so it lands in access logs referrer headers and browser history and since the key payload is deterministic a leaked secret means unlimited valid key minting no-store doesnt help the logs

Also minor the length check before timingSafeEqual leaks the secret length through timing easy fix is to hmac both sides to a fixed length before comparing

charlesBochet (approved):

PR #21664, opened 2026-06-16
feat(code-interpreter): reuse a warm sandbox per conversation (E2B)

FelixMalfait:
## What

The E2B code-interpreter driver created a **fresh sandbox on every execution** and killed it in `finally`, so every call in a conversation paid full cold-start and started blank. This PR keeps **one warm sandbox per conversation** and, on idle, **pauses** it rather than killing it.

## How

- **Discovery without a registry:** the sandbox is tagged with the chat `threadId` (scoped `workspaceId:threadId`) via E2B **metadata**, found with `Sandbox.list({ query: { state: ['running','paused'], metadata } })` and resumed with `Sandbox.connect()` (which auto-resumes a paused sandbox). E2B is the source of truth — no Redis/DB mapping.
- **Pause/resume (E2B 2.x):** session sandboxes are created with `lifecycle: { onTimeout: 'pause', autoResume: true }`. When idle they **pause** — compute billing stops, filesystem **and** kernel/memory state are preserved — and resume in ~1s on the next call. This replaces the earlier keepalive approach.
- **No premature pause mid-run:** the sandbox is kept alive for `max(execution timeout, idle window)`, so a long execution is never paused underneath itself.
- **Tenant isolation:** discovery filters by the `twentySessionId` tag and **re-checks it client-side**, so a loose server-side match can never hand one conversation's warm sandbox (with its files, kernel state, token) to another.
- **Concurrency:** executions sharing a session are serialized in-process (one active stream per thread, run as a single job — the chat resolver queues concurrent messages), so parallel tool calls can't race the shared kernel.
- **Output isolation:** `/home/user/output` is reset at the start of each reused run, so a call only returns the artifacts it actually produced; durable state lives elsewhere and persists.

## SDK upgrade

`@e2b/code-interpreter` **`^1.0.4` → `^2.6.0`** (pulls `e2b@2.x`). The typed pause/resume API, `lifecycle`, and the `state`/`metadata` list filter only exist in the 2.x line; 1.x exposed them only as untyped OpenAPI internals. `Sandbox.list()` is now a paginator (handled).

## Config

| Var | Default | Purpose |
|---|---|---|
| `CODE_INTERPRETER_TIMEOUT_MS` | `300000` | Max single-execution duration. |
| `CODE_INTERPRETER_IDLE_TIMEOUT_MS` | `300000` | Idle window before the warm sandbox auto-pauses. |

Reuse is always-on when a session id is present (chat path). The workflow-agent path and the dev-only `LocalDriver` are unaffected.

## ⚠️ Open item before merge: paused-sandbox GC

E2B retains paused sandboxes **indefinitely** (no TTL). Unlike the old keepalive path (which auto-killed on idle), pause means a conversation's sandbox persists after the chat ends — so without garbage collection, paused sandboxes accumulate (≈ one per historical conversation) and consume storage. A GC policy is required; the approach + retention window are being decided (see PR discussion). Also: the E2B runtime path can't run in CI, so this still needs a **live smoke test** (reuse hit, idle→pause, resume) and confirmation of paused-storage pricing before rollout.

## Tests / checks

- Resolver unit tests (`getOrCreateSessionSandbox`): reuse+extend, create-when-absent, duplicate reaping, connect-failure fallback, keep-first-connectable-when-earlier-dead, **ignore cross-tenant metadata**, and **kill-on-timeout-refresh-failure**.
- `nx typecheck twenty-server` (against e2b 2.x), `oxlint --type-aware`, `oxfmt --check` all clean.


packages/twenty-server/src/engine/core-modules/code-interpreter/__tests__/code-interpreter.service.spec.ts
```diff
@@ -0,0 +1,106 @@
+import { CodeInterpreterService } from 'src/engine/core-modules/code-interpreter/code-interpreter.service';
```
FelixMalfait: Let's remove this file

packages/twenty-server/src/engine/core-modules/code-interpreter/drivers/interfaces/code-interpreter-driver.interface.ts
```diff
@@ -20,6 +20,9 @@ export type CodeExecutionResult = {
 
 export type ExecutionContext = {
   env?: Record<string, string>;
+  // Stable per-conversation key. When set (and session reuse is enabled) the
```
FelixMalfait: Remove comment

packages/twenty-server/src/engine/core-modules/twenty-config/config-variables.ts
```diff
@@ -701,6 +701,25 @@ export class ConfigVariables {
...
   CODE_INTERPRETER_TIMEOUT_MS = 300_000;
 
+  @ConfigVariablesMetadata({
+    group: ConfigVariablesGroup.CODE_INTERPRETER_CONFIG,
+    description:
+      'Reuse one warm sandbox per conversation (E2B driver) instead of creating a fresh sandbox per execution, so scripts and variables persist across calls within a conversation',
+    type: ConfigVariableType.BOOLEAN,
+  })
+  @IsOptional()
+  CODE_INTERPRETER_SESSION_REUSE_ENABLED = false;
```
FelixMalfait: I'm not convinced this is necessary, let's just consider it's always true. Would we have a good reason to disable this? I don't see one? If not, let's cleanup the code

packages/twenty-server/src/engine/core-modules/code-interpreter/constants/code-interpreter-session-cleanup-cron-pattern.constant.ts
```diff
@@ -0,0 +1,2 @@
+// Hourly; reclaims sandboxes within roughly an hour of crossing the retention window.
```
FelixMalfait: Remove comment

FelixMalfait:
Screenshot 2026-06-16 at 21 45 24

packages/twenty-server/src/engine/core-modules/code-interpreter/drivers/utils/session-sandbox-gc.util.ts
```diff
@@ -0,0 +1,77 @@
...
+
+type SandboxApi = typeof Sandbox;
+
+const killSandboxById = (
+  sandboxApi: SandboxApi,
+  apiKey: string,
+  sandboxId: string,
+) => sandboxApi.kill(sandboxId, { apiKey }).catch(() => undefined);
+
+export const releaseSessionSandboxes = async (
```
FelixMalfait: 1 export = 1 file, fix it

packages/twenty-server/src/engine/core-modules/code-interpreter/drivers/utils/sweep-expired-session-sandboxes.util.ts
```diff
@@ -0,0 +1,42 @@
...
+
+// E2B retains paused sandboxes indefinitely, so abandoned sessions need an age bound.
+export const sweepExpiredSessionSandboxes = async (
+  sandboxApi: SandboxApi,
+  apiKey: string,
+  maxAgeMs: number,
+): Promise<number> => {
+  const paginator = sandboxApi.list({
+    apiKey,
+    query: { state: ['running', 'paused'] },
```
etiennejouan, line 17: We're ok with killing running sandbox ?

packages/twenty-server/src/engine/core-modules/code-interpreter/drivers/utils/get-or-create-session-sandbox.util.ts
```diff
@@ -0,0 +1,144 @@
...
+
+type GetOrCreateSessionSandboxArgs = {
+  sandboxApi: SandboxApi;
+  apiKey: string;
+  sessionId: string;
+  timeoutMs: number;
+  idleTimeoutMs: number;
+};
+
+const listSandboxesForSession = async (
```
etiennejouan, line 17: We should not have multiple sandboxes per session, no ? Try to figure out how it can happen.

etiennejouan (approved):
Nice feature. Wonder if e2b limits number of paused sandboxes. And indeed could be interesting to confirm paused sandboxes costs.

packages/twenty-server/src/engine/core-modules/code-interpreter/drivers/utils/get-or-create-session-sandbox.util.ts
FelixMalfait: We should not indeed, it's an index endpoint filtered on metadata so it returns a list but then it does `for (const { sandboxId } of sessionSandboxes` so gets the first one

packages/twenty-server/src/engine/core-modules/code-interpreter/drivers/utils/sweep-expired-session-sandboxes.util.ts
FelixMalfait: After 1 week I think it's OK. I agree we could argue to do something smarter

PR #21671, opened 2026-06-16
Hide Input and Test tabs for app logic function nodes

bosiraphael:
When opening a logic function node that comes from an app (e.g. "Enrich Company" from People Data Labs), the node detail panel was showing **Input** and **Test** tabs. App-sourced functions should only expose their input fields, with no test surface.

This hides the tab list when the logic function has an `applicationId`, rendering the input fields directly. The test tab is also never treated as active for these nodes, which guards against a stale "test" tab state leaking in from a previously-opened custom function (the tab list shares one component instance id).

## Before

CleanShot 2026-06-16 at 15 11 32@2x

## After

CleanShot 2026-06-16 at 15 11 09@2x

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21671?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/logic-function-action/components/WorkflowEditActionLogicFunction.tsx
+++ b/packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/logic-function-action/components/WorkflowEditActionLogicFunction.tsx
@@ -185,7 +185,8 @@ export const WorkflowEditActionLogicFunction = ({
 
   const isLogicFunctionFromApp = isDefined(logicFunction?.applicationId);
 
-  const isTestTabActive = !isLogicFunctionFromApp && activeTabId === TEST_TAB_ID;
+  const isTestTabActive =
+    !isLogicFunctionFromApp && activeTabId === TEST_TAB_ID;
 
   const tabs = [
     {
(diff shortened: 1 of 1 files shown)


packages/twenty-front/src/modules/workflow/workflow-steps/workflow-actions/logic-function-action/components/WorkflowEditActionLogicFunction.tsx
```diff
@@ -183,6 +183,11 @@ export const WorkflowEditActionLogicFunction = ({
 
   const hasInputFields = Object.keys(functionInput).length > 0;
 
+  const isLogicFunctionFromApp = isDefined(logicFunction?.applicationId);
```
Weiko: I believe applicationId is non-nullable, I'm not sure this is the correct choice here 👀

Weiko (approved):
LGTM

PR #21673, opened 2026-06-16
feat(server): add isSystemSideEffect & merge createOneObject/createOneField side-effect migrations

Weiko:
## Context

When an object is created via the metadata API, `createOneObject` creates its side-effect entities (INDEX view + viewFields, indexes, navigation menu item, "go to" command menu item, record-page fields view, page layout/tabs/widgets) across **three separate `validateBuildAndRunWorkspaceMigration` calls**, purely because the protection behavior (mutations → overrides, delete → deactivate, reset → reactivate) was keyed on *"owned by the standard app"*, forcing the side effects into batches with different application owners. This misrepresents ownership and breaks atomicity.

This PR separates two orthogonal concepts:

- **Ownership** (`applicationId`), the true owner: the caller's application (the workspace custom app today, 3rd-party apps later).
- **Protection** (`isSystemSideEffect`), the row was generated by the system, so user mutations route to overrides, deletion becomes deactivation, and reset restores defaults.

Once side effects are re-owned to the caller, the old `applicationId === standardApp` check can no longer tell an original side-effect row from a user-added one so a dedicated `isSystemSideEffect` flag carries the protection instead.

This is **PR 1 of 2** (forward-only). It makes newly created objects and fields correct; existing workspaces are handled by a follow-up backfill (see *Out of scope*).

## What this PR does

- **`isSystemSideEffect` column** on the 8 affected entities (`view`, `viewField`, `indexMetadata`, `commandMenuItem`, `pageLayout`, `pageLayoutTab`, `pageLayoutWidget`, `fieldMetadata`), with `@WasIntroducedInUpgrade` + an entry in the flat-entity property configuration (`toCompare: true`, read-only).
- **Single atomic migration in `createOneObject`**: the three `validateBuildAndRunWorkspaceMigration` calls are merged into one, owned by the caller (`resolvedOwnerFlatApplication`) and the record-page view/fields, page layout, and navigation command item are re-owned to the caller and flagged `isSystemSideEffect: true`. `buildNavigationFlatCommandMenuItem` is parameterized with `applicationUniversalIdentifier` (no longer hardcoded to the standard app).
- **Field-creation side effects** (`createManyFields`/`createOneField` already run as a single caller-owned migration, so no re-ownership/merge was needed): the auto-created viewField is flagged `isSystemSideEffect: true`, and a new field now also propagates to the object's **INDEX/table view** (added there as a **hidden** column, `isVisible: false`) in addition to
the record-page FIELDS widget. The INDEX view is targeted directly by `key = INDEX` (it is not a page-layout widget), de-duplicated per `(viewId, fieldMetadataUniversalIdentifier)` to respect the per-view unique index. 
The unique-field index is likewise flagged the inverse relation field stays unflagged (`isSystem: false`).
- **Protection predicate** extended: `isCallerOverridingEntity` and the removal/reset split strategies now treat `isSystemSideEffect` rows as protected even when caller-owned (route to
overrides / deactivate / reset) and the page-layout-reset guards allow resetting flagged entities.
- **Standard compute maps** set the flag consistently so a re-sync produces no diff (standard-object side effects stay `false`; per-object nav command items and custom-object base fields are `true`).
- **Read-only GraphQL exposure** of `isSystemSideEffect` on the view / view-field / page-layout / tab / widget / command-menu-item DTOs (not exposed on create/update inputs). => Todo: needs to take this new flag into account. This is fine for now because isSystem remains on object/field.
- **Fast instance command** (`2-14`) adding the 8 columns (`NOT NULL DEFAULT false`).

## Scope decisions

- **`pageLayout` is not an `OverridableEntity`**, its own row has nothing user-overridable (all customization lives on tabs/widgets). It's dual-purpose (`RECORD_PAGE` side-effect vs. user `DASHBOARD`), so it gets `isSystemSideEffect` for protection only, no `overrides` jsonb.
- **`navigationMenuItem` is out of scope.**: Those are side effects only for the metadata API and not marked as "system" (they can be deleted/updated etc...)
- **`viewFieldGroup` is not a side effect**, it's only created via the explicit view-field-group API, never by object/field creation, so it gets no flag.

## Out of scope (follow-ups)

**PR 2** — slow per-workspace backfill (re-own + flag existing side effects, recreate missing ones) and deterministic v5 identifiers for base fields / pageLayout / tab.

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21673?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


packages/twenty-server/src/database/commands/upgrade-version-command/2-10/utils/build-navigation-command-menu-item-operations-or-throw.util.ts
```diff
@@ -63,6 +64,8 @@ export const buildNavigationCommandMenuItemOperationsOrThrow = ({
         objectMetadata,
         commandMenuItemId: v4(),
         applicationId,
+        applicationUniversalIdentifier:
+          TWENTY_STANDARD_APPLICATION_UNIVERSAL_IDENTIFIER,
```
prastoin, line 68: ***Nitpick:*** not important as it's a old upgrade command util but it's quite missleading as the params given applicationId could be any and not twenty standard app only

packages/twenty-server/src/database/commands/upgrade-version-command/2-15/2-15-instance-command-fast-1781600000000-add-is-system-side-effect.ts
```diff
@@ -0,0 +1,36 @@
...
+const TABLES_WITH_IS_SYSTEM_SIDE_EFFECT = [
+  'view',
+  'viewField',
+  'indexMetadata',
+  'commandMenuItem',
+  'pageLayout',
+  'pageLayoutTab',
+  'pageLayoutWidget',
+  'fieldMetadata',
+] as const;
```
prastoin, line 15: ***nitpick:*** satisfies MetadataName[]

packages/twenty-server/src/database/commands/upgrade-version-command/2-15/is-system-side-effect-upgrade-command-name.constant.ts
```diff
@@ -0,0 +1,2 @@
+export const ADD_IS_SYSTEM_SIDE_EFFECT_UPGRADE_COMMAND_NAME =
```
prastoin, line 1: ***Remark:*** This should never get mutated in any way

packages/twenty-server/src/engine/metadata-modules/flat-entity/constant/all-entity-properties-configuration-by-metadata-name.constant.ts
```diff
@@ -40,6 +40,11 @@ type MetadataEntityPropertyConfiguration<
 
 export const ALL_ENTITY_PROPERTIES_CONFIGURATION_BY_METADATA_NAME = {
   fieldMetadata: {
+    isSystemSideEffect: {
+      toCompare: true,
```
prastoin: ***Remark:*** should be false to me as not editable at all

packages/twenty-server/src/database/commands/upgrade-version-command/2-15/is-system-side-effect-upgrade-command-name.constant.ts
prastoin: location

packages/twenty-server/src/engine/core-modules/application/application-manifest/converters/from-field-manifest-to-universal-flat-field-metadata.util.ts
```diff
@@ -94,6 +94,7 @@ export const fromFieldManifestToUniversalFlatFieldMetadata = ({
     universalSettings: fieldManifest.universalSettings ?? null,
     isActive: true,
     isSystem: fieldManifest.name in PARTIAL_SYSTEM_FLAT_FIELD_METADATAS,
+    isSystemSideEffect: false,
     isUIEditable: fieldManifest.isUIEditable ?? true,
```
prastoin: ***Self reminder:*** double check if system field from app sync are correctly `isSystemSideEffect` true

prastoin: should be `fieldManifest.name in PARTIAL_SYSTEM_FLAT_FIELD_METADATAS`

packages/twenty-server/src/engine/core-modules/application/application-manifest/converters/from-index-manifest-to-universal-flat-index.util.ts
```diff
@@ -129,6 +129,7 @@ export const fromIndexManifestToUniversalFlatIndex = ({
       indexWhereClause: null,
       isCustom: false,
       isUnique: indexManifest.isUnique ?? false,
+      isSystemSideEffect: false,
```
prastoin, line 132: I'm pretty sure we have a system index here too

packages/twenty-server/src/engine/workspace-manager/twenty-standard-application/utils/field-metadata/create-standard-field-flat-metadata.util.ts
```diff
@@ -82,6 +83,7 @@ export const createStandardFieldFlatMetadata = <
     icon,
     isActive: true,
     isSystem,
+    isSystemSideEffect: name in PARTIAL_SYSTEM_FLAT_FIELD_METADATAS,
```
prastoin, line 86: I think we should rather consume the isSystem here no ?

packages/twenty-server/src/engine/metadata-modules/flat-field-metadata/utils/get-default-flat-field-metadata-from-create-field-input.util.ts
```diff
@@ -41,6 +41,7 @@ export const getDefaultFlatFieldMetadata = ({
       createFieldInput.isRemoteCreation,
     ),
     isSystem: createFieldInput.isSystem ?? false,
+    isSystemSideEffect: false,
```
prastoin, line 44: ***Self reminder*** Double check caller context

packages/twenty-server/src/engine/metadata-modules/flat-object-metadata/utils/recompute-view-field-identifier-after-flat-object-identifier-update.util.ts
```diff
@@ -71,6 +71,7 @@ export const recomputeViewFieldIdentifierAfterFlatObjectIdentifierUpdate = ({
         position: lowestViewFieldPosition - 1,
         isVisible: true,
         isActive: true,
+        isSystemSideEffect: false,
```
prastoin: ***Question:*** looks like a side effect to me no ?

packages/twenty-server/src/engine/metadata-modules/flat-view-field-group/utils/from-update-view-field-group-input-to-flat-view-field-group-to-update-or-throw.util.ts
```diff
@@ -59,6 +59,7 @@ export const fromUpdateViewFieldGroupInputToFlatViewFieldGroupToUpdateOrThrow =
       entityApplicationUniversalIdentifier:
         existingFlatViewFieldGroupToUpdate.applicationUniversalIdentifier,
       workspaceCustomApplicationUniversalIdentifier,
+      isSystemSideEffect: false,
```
prastoin, line 62: shouldn't this be `existingFlatViewFieldGroupToUpdate.isSystemSideEffect` ?

packages/twenty-server/src/engine/metadata-modules/flat-view-field/utils/compute-flat-view-fields-from-fields-widgets.util.ts
```diff
@@ -200,6 +246,7 @@ export const computeFlatViewFieldsFromFieldsWidgets = ({
           position,
           aggregateOperation: null,
           isActive: true,
+          isSystemSideEffect: true,
```
prastoin: ***Question:*** Unless I'm mistaken, only this line in this file git diff is related to the PR scope ?

packages/twenty-server/src/engine/metadata-modules/index-metadata/services/index-metadata.service.ts
```diff
@@ -243,6 +243,7 @@ export class IndexMetadataService {
           indexWhereClause: null,
           isCustom: true,
           isUnique: false,
+          isSystemSideEffect: false,
```
prastoin, line 246: ***Question:*** why would we need the `isSystemSideEffect` on the index in the first place ? Aren't all index system for the moment ?

packages/twenty-server/src/engine/metadata-modules/object-metadata/utils/build-default-flat-field-metadatas-for-custom-object.util.ts
```diff
@@ -134,6 +134,7 @@ export const buildDefaultFlatFieldMetadatasForCustomObject = ({
           isNullable: true,
           isActive: true,
           isSystem: false,
+          isSystemSideEffect: false,
```
prastoin: should be true to me

packages/twenty-server/src/engine/metadata-modules/view-field-group/services/fields-widget-upsert.service.ts
```diff
@@ -250,6 +250,7 @@ export class FieldsWidgetUpsertService {
             existingGroup.applicationUniversalIdentifier,
           workspaceCustomApplicationUniversalIdentifier:
             applicationUniversalIdentifier,
+          isSystemSideEffect: false,
```
prastoin, line 253: always false ?

packages/twenty-server/src/engine/workspace-manager/twenty-standard-application/utils/command-menu-item/create-standard-command-menu-item-flat-metadata.util.ts
```diff
@@ -77,6 +77,7 @@ export const createStandardCommandMenuItemFlatMetadata = ({
     pageLayoutId: null,
     pageLayoutUniversalIdentifier: null,
     isActive: true,
+    isSystemSideEffect: false,
```
prastoin, line 80: I'm afraid of this, this means that the twenty standard app has no system side effect entities at all ?

prastoin (commented):
Hey there ! nice job
Remains few questions regarding the:
- twenty standard app is system side effect declaration
- app manifest transpilation and is system side effect 
- ci seems to be red due to failing snapshot

packages/twenty-server/src/database/commands/upgrade-version-command/2-15/2-15-instance-command-fast-1781600000000-add-is-system-side-effect.ts
Weiko: Those are table names, they don't always match (indexMetadata vs index)

packages/twenty-server/src/engine/core-modules/application/application-manifest/converters/from-index-manifest-to-universal-flat-index.util.ts
Weiko: Not applicable here, imho this is the only case where an index is not a side effect today, this is explicitely defined in the manifest here so not a side effect

packages/twenty-server/src/engine/metadata-modules/index-metadata/services/index-metadata.service.ts
Weiko: Not all as mentioned above, app can define indexes via the manifest

packages/twenty-server/src/engine/metadata-modules/view-field-group/services/fields-widget-upsert.service.ts
Weiko: Yes, isCallerOverridingEntity combines overrides and isSystem logic to gate things but in this case viewFieldGroup is overridable but not "side-effectable" so here we specify false

packages/twenty-server/src/engine/workspace-manager/twenty-standard-application/utils/command-menu-item/create-standard-command-menu-item-flat-metadata.util.ts
Weiko: This was intentional to not have a huge PR since I'll have to backfill after anyway, but fair, I'll update this in this PR

packages/twenty-server/src/database/commands/upgrade-version-command/2-15/is-system-side-effect-upgrade-command-name.constant.ts
Weiko: sure

packages/twenty-server/src/engine/workspace-manager/twenty-standard-application/utils/field-metadata/create-standard-field-flat-metadata.util.ts
Weiko: This is probably the same result here right? I'd rather have the correct path right now unless I'm missing something

packages/twenty-server/src/engine/metadata-modules/flat-view-field-group/utils/from-update-view-field-group-input-to-flat-view-field-group-to-update-or-throw.util.ts
Weiko: as mentioned in the other comment, viewFieldGroup don't have isSystem but have overrides, the util takes both info to resolve what we want so here I need to pass it explicitly as false since `existingFlatViewFieldGroupToUpdate.isSystemSideEffect` does not exist. We could do separate utils later if it's a common pattern

prastoin (approved):

PR #21680, opened 2026-06-16
fix(front): sanitize chart filters when fields are deactivated or deleted

ijreilly:
## Summary

Fixes a bug in dashboard chart filters: when a field used in a chart filter is **deactivated or deleted**, the chart became unusable for filter editing.

Two distinct symptoms:
1. **Cannot add new filters and save** — an invalid filter (on a deactivated/deleted field) was re-persisted on every update, blocking newly added filters from taking effect.
2. **Cannot delete the filter on the deactivated/deleted field** — the stale filter kept coming back.

## Root causes

- `handleFiltersUpdate` (in `ChartFiltersSettings`) persisted **all** current filters to the widget draft without sanitizing against the set of active fields, so an invalid filter was repeatedly re-saved.
- `useGraphWidgetQueryCommon` treated **all** fields (including `isActive: false`) as valid when building the query filter, so deactivated-field filters were never dropped at query execution time (only fully-deleted ones were).

## Changes

- **`ChartFiltersSettings.tsx`** — sanitize filters via `dropChartRecordFiltersWithDeletedFields` before writing to the draft, dropping any filter whose `fieldMetadataId` is not in the active-fields set.
- **`useGraphWidgetQueryCommon.ts`** — restrict valid field IDs to `isActive` fields so deactivated-field filters are silently dropped at query time, consistent with the save path.
- **`dropChartRecordFiltersWithDeletedFields.ts`** — enhanced to also iteratively clean up orphaned filter groups left behind once their filters are removed (prevents inconsistent state where filters are dropped but empty groups remain).
- **`ChartFiltersDeletedFieldsWarning.tsx`** — updated warning copy to mention both deactivated and deleted fields and clarify filters are removed on save.
- **Tests** — updated existing tests and added cases for orphaned root/child group cleanup.

## Test plan

- [ ] Create a bar chart on a dashboard, add a filter on a field (e.g. `stage` on Opportunity)
- [ ] Deactivate that field — confirm the warning banner appears
- [ ] Add a new filter and save — confirm it saves and the deactivated-field filter is dropped
- [ ] Repeat with a deleted field
- [ ] Confirm the chart query renders without the stale filter
- [ ] `npx jest dropChartRecordFiltersWithDeletedFields`

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21680?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/twenty-front/src/modules/page-layout/widgets/graph/hooks/useGraphWidgetQueryCommon.ts
+++ b/packages/twenty-front/src/modules/page-layout/widgets/graph/hooks/useGraphWidgetQueryCommon.ts
@@ -46,7 +46,9 @@ export const useGraphWidgetQueryCommon = ({
   );
 
   const objectFieldMetadataIds = new Set(
-    objectMetadataItem.fields.map((field: { id: string }) => field.id),
+    objectMetadataItem.fields
+      .filter((field: { id: string; isActive: boolean }) => field.isActive)
+      .map((field: { id: string; isActive: boolean }) => field.id),
   );
 
   const { recordFilters: sanitizedRecordFilters } =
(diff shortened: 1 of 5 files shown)


charlesBochet:
PR is staled let's re-open once you have bandwidth

PR #21688, opened 2026-06-16
Added page card box-shadow

Bonapara:
## Summary
- add a subtle left-side page card shadow in light and dark mode
- preserve the existing border-ring box shadow
- give the page card wrapper enough left padding for the shadow to render

image

### DIFF
--- a/packages/twenty-front/src/modules/ui/layout/page/components/PageCardLayout.tsx
+++ b/packages/twenty-front/src/modules/ui/layout/page/components/PageCardLayout.tsx
@@ -28,6 +28,7 @@ const StyledMainCardWrapper = styled.div`
   width: 0;
 `;
 
+// oxlint-disable-next-line twenty/no-hardcoded-colors
 const StyledCard = styled.div`
   background: ${themeCssVariables.background.primary};
   border-radius: 16px 0 0 0;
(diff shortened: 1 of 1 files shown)


Bonapara:
## Before / after

Before: only the border-ring box shadow is present.

Before: page card with only border-ring shadow

After: adds the subtle page-card shadow while preserving the border-ring shadow.

After: page card with subtle left shadow and border-ring shadow

packages/twenty-front/src/modules/ui/layout/page/components/PageCardLayout.tsx
```diff
@@ -22,22 +22,31 @@ const StyledMainCardWrapper = styled.div`
   box-sizing: border-box;
   display: flex;
   flex: 1 1 0;
+  margin-left: -3px;
```
Bonapara, line 25: because it was cropping the box shadow

packages/twenty-front/src/modules/ui/layout/page/components/PageCardLayout.tsx
```diff
@@ -22,22 +22,31 @@ const StyledMainCardWrapper = styled.div`
...
-  padding-left: 1px;
+  padding-left: 4px;
   width: 0;
 `;
 
 const StyledCard = styled.div`
   background: ${themeCssVariables.background.primary};
   border-radius: 16px 0 0 0;
-  box-shadow: 0 0 0 1px ${themeCssVariables.border.color.medium};
+  box-shadow:
```
Bonapara, line 35: not meant to be used anywhere else

Weiko (approved):

PR #21676, opened 2026-06-16
feat(workflow): expand manual-trigger runtime payload with payload + _metadata

thomtrp:
## Summary

Step 1 (**expand**) of restructuring the manual-trigger output so record fields live under a `payload` key and trigger-level metadata (the running workspace member) lives alongside it. This step is **additive and runtime-only** — no behavior changes for existing workflows, and nothing new is surfaced in the variable picker yet.

The manual-trigger runtime payload now additively carries:
- `payload`: a mirror of the incoming record fields, reachable at `{{trigger.payload.*}}`
- `_metadata.workspaceMemberId`: the member who ran the workflow, reachable at `{{trigger._metadata.workspaceMemberId}}`

Record fields are still served at the trigger root, so existing `{{trigger.id}}` references keep working unchanged. The output schema / variable picker is intentionally left untouched here.

### Why `_metadata` (underscore)
During the transition, record fields still sit at the `trigger` root next to the injected keys. Field API names can't start with `_`, so `_metadata` is collision-proof against any record field; picking the name now avoids a later variable-path rename migration.

### Phasing
- **Step 1 (this PR):** write `payload` + `_metadata` at runtime; keep using direct `trigger.*`; don't display the new paths.
- **Step 2:** surface `payload` + `_metadata` in the variable picker.
- **Step 3:** migrate existing variables to `trigger.payload.*` and contract the root record fields.

## Test plan
- [x] `twenty-shared` builds, `twenty-server` typechecks, lint clean on changed files
- [x] Manual: run a manually-triggered (SINGLE_RECORD) workflow and confirm the run's trigger payload contains `payload.*` mirroring the record and `_metadata.workspaceMemberId`
- [x] Manual: confirm existing `{{trigger.id}}` references still resolve

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21676?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #21692, opened 2026-06-16
feat(workflow): surface manual-trigger payload + metadata in variable picker

thomtrp:
## Summary

Step 2 of the manual-trigger output schema restructuring (expand → display → migrate → contract).

Builds on the now-merged #21676 (which expanded the runtime payload to serve `payload` and `metadata` siblings at the trigger root). This PR **surfaces** those in the variable picker as nested, expandable nodes:

- `trigger.payload.{record fields}` — the record(s) that triggered the run
- `trigger.metadata.workspaceMemberId` — who triggered it

The flat root fields (`trigger.id`, etc.) remain available, so existing saved variable references keep working until a later migration phase moves them.

### Changes
- **twenty-shared**: metadata/payload label constants + `build-manual-trigger-metadata-node` util + barrel exports.
- **twenty-front**: `computeStepOutputSchema` MANUAL branch now nests `payload` (RecordNode for SINGLE_RECORD, array Node for BULK_RECORDS, omitted for GLOBAL) and `metadata`; `ManualTriggerOutputSchema` type updated to `{ payload?; metadata }`.
- **twenty-server**: `computeTriggerOutputSchemaFromAvailability` mirrors the same nested shape for server-side validation.

The key is `metadata` (not `_metadata`) — custom fields can't start with `_`, so collision risk was deemed acceptable.

## Test plan
- [x] `npx nx build twenty-shared`
- [x] `computeStepOutputSchema` unit tests pass (55)
- [x] Manual: create a manual-trigger workflow (GLOBAL / single-record / bulk), confirm the picker shows `payload` and `metadata` as expandable folders and that selecting a field yields `{{trigger.payload.<field>}}` / `{{trigger.metadata.workspaceMemberId}}`

> Note: server typecheck has pre-existing unrelated failures on main (Stripe billing mocks, gmail mocks); none touch workflow files.

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21692?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/twenty-server/src/database/commands/upgrade-version-command/2-15/2-15-workspace-command-1800000001000-migrate-manual-trigger-variables-to-payload.command.ts
+++ b/packages/twenty-server/src/database/commands/upgrade-version-command/2-15/2-15-workspace-command-1800000001000-migrate-manual-trigger-variables-to-payload.command.ts
@@ -1,0 +1,91 @@
+import { Command } from 'nest-commander';
+import { isDefined } from 'twenty-shared/utils';
+
+import { ActiveOrSuspendedWorkspaceCommandRunner } from 'src/database/commands/command-runners/active-or-suspended-workspace.command-runner';
+import { WorkspaceIteratorService } from 'src/database/commands/command-runners/workspace-iterator.service';
+import { type RunOnWorkspaceArgs } from 'src/database/commands/command-runners/workspace.command-runner';
+import { rewriteTriggerVariablesToPayload } from 'src/database/commands/upgrade-version-command/2-15/utils/rewrite-trigger-variables-to-payload.util';
+import { RegisteredWorkspaceCommand } from 'src/engine/core-modules/upgrade/decorators/registered-workspace-command.decorator';
+import { GlobalWorkspaceOrmManager } from 'src/engine/twenty-orm/global-workspace-datasource/global-workspace-orm.manager';
+import { type WorkflowVersionWorkspaceEntity } from 'src/modules/workflow/common/standard-objects/workflow-version.workspace-entity';
+import { WorkflowTriggerType } from 'src/modules/workflow/workflow-trigger/types/workflow-trigger.type';
+
+@RegisteredWorkspaceCommand('2.15.0', 1800000001000)
+@Command({
+  name: 'upgrade:2-15:migrate-manual-trigger-variables-to-payload',
+  description:
+    'Rewrite saved {{trigger.<field>}} references to {{trigger.payload.<field>}} for manual record triggers',
+})
+export class MigrateManualTriggerVariablesToPayloadCommand extends ActiveOrSuspendedWorkspaceCommandRunner {
+  constructor(
+    protected readonly workspaceIteratorService: WorkspaceIteratorService,
+    private readonly twentyORMGlobalManager: GlobalWorkspaceOrmManager,
+  ) {
+    super(workspaceIteratorService);
+  }
+
+  override async runOnWorkspace({
+    workspaceId,
+    options,
+  }: RunOnWorkspaceArgs): Promise<void> {
+    const isDryRun = options.dryRun ?? false;
+
+    const workflowVersionRepository =
+      await this.twentyORMGlobalManager.getRepository<WorkflowVersionWorkspaceEntity>(
+        workspaceId,
+        'workflowVersion',
+        { shouldBypassPermissionChecks: true },
+      );
+
+    const allVersions = await workflowVersionRepository.find();
+
+    let updatedVersionCount = 0;
+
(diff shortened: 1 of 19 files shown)


packages/twenty-server/src/database/commands/upgrade-version-command/2-15/utils/rewrite-trigger-variables-to-payload.util.ts
```diff
@@ -0,0 +1,31 @@
+import { isDefined } from 'twenty-shared/utils';
+
+// Manual-trigger record fields moved from the flat trigger root
+// ({{trigger.name}}) to a nested payload node ({{trigger.payload.name}}).
+// The negative lookaheads keep this idempotent: references already pointing at
+// `payload`/`metadata` (or the legacy `_metadata` key) are left untouched, so
+// re-running is a no-op and metadata refs are never prefixed with `payload`.
+// A field named exactly `payload` or `metadata` collides and is skipped — that
```
Weiko, line 8: I agree, let's accept the risk. Should it log something when it hits one of these colliding field names though?

packages/twenty-server/src/database/commands/upgrade-version-command/2-15/2-15-workspace-command-1800000001000-migrate-manual-trigger-variables-to-payload.command.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    let updatedVersionCount = 0;
+
+    for (const version of allVersions) {
+      if (!this.isManualRecordTrigger(version.trigger)) {
+        continue;
+      }
+
+      const migratedSteps = rewriteTriggerVariablesToPayload(version.steps);
+      const migratedTrigger = rewriteTriggerVariablesToPayload(version.trigger);
```
Weiko: Am I right that the trigger object itself never contains trigger references but should be in steps ?
If so, `migratedTrigger.changed` condition below is always false? I might be wrong I'm not super familiar with this code!

packages/twenty-server/src/database/commands/upgrade-version-command/2-15/utils/rewrite-trigger-variables-to-payload.util.ts
```diff
@@ -0,0 +1,31 @@
...
+// risk was accepted when the nested keys were named.
+const TRIGGER_VARIABLE_PREFIX_REGEX =
```
Weiko, line 10: this matches `{{trigger.` but won't match something like `{{ trigger.name }}` (with spaces), do we want to resolve?

Weiko (approved):
LGTM @thomtrp

packages/twenty-server/src/database/commands/upgrade-version-command/2-15/utils/rewrite-trigger-variables-to-payload.util.ts
thomtrp: I think we would miss it anyway + on a re-run we'd log false positives. Since the collision is a rare case, I'd rather keep it silent than add noisy/misleading logs

packages/twenty-server/src/database/commands/upgrade-version-command/2-15/2-15-workspace-command-1800000001000-migrate-manual-trigger-variables-to-payload.command.ts
thomtrp: Good catch 😓 
Thank you, cleaning!

packages/twenty-server/src/database/commands/upgrade-version-command/2-15/utils/rewrite-trigger-variables-to-payload.util.ts
Weiko: Fair!

packages/twenty-server/src/database/commands/upgrade-version-command/2-15/utils/rewrite-trigger-variables-to-payload.util.ts
thomtrp: {{ trigger.name }} does not resolve in UI today anyway. So I would leave it out of scope

PR #21698, opened 2026-06-16
[codex] Fix column sort direction reload

### LINKED ISSUE #21621: (UX) No reload after COLUMN SORT (and add a ASC / DESC sorting button)
## Scope & Context
We need to improve the user experience of sorting columns in data tables (e.g., in views like "All Biens Immobiliers"). Sorting should be a fast, seamless operation. Currently, the behavior disrupts the workflow, especially on tables with many columns where horizontal scrolling is required.

## Current behavior
When a user clicks on a column header to apply a sort, the entire page reloads or the table completely refreshes its state. 
This causes two major issues:
1. **Loss of horizontal position:** The table scroll resets, meaning the user loses sight of the column they just sorted if it was located far to the right. They have to scroll back manually.
2. **Unpredictable sort direction:** Clicking "Sort" doesn't specify the direction (Ascending vs. Descending). If the user gets the wrong direction, they have to click again, wait for the reload, lose their scroll position again, and scroll back *again*. This wastes a significant amount of time.

Image

## Expected behavior
1. **No page/table reload:** Applying a sort should update the data asynchronously (optimistic UI update or simple data fetch) without resetting the scroll container's horizontal position. The user's viewport must remain exactly where it is.
2. **Explicit sort directions:** Instead of a single "Sort" button that toggles unpredictably, the column header menu should offer two explicit options: "Sort Ascending (A-Z / 1-9)" and "Sort Descending (Z-A / 9-1)". 
*(Alternatively, clicking the column header directly could toggle between Asc, Desc, and Unsorted, with a clear UI indicator (arrow) showing the current state).*

## Technical inputs
- **UI/Scroll State:** Modify the table component (or the Apollo Client/GraphQL cache update logic) to ensure that refetching data upon a sort mutation does not unmount or reset the `scrollLeft` property of the table container.
- **Menu Options:** Update the column dropdown menu component to replace the generic toggle sort with two distinct `MenuItem` components: `Sort Ascending` and `Sort Descending`.
- **Mutation/State Update:** Ensure that clicking these options directly sets the specific `orderBy` direction in the view's query parameters (e.g., `AscNullsLast` / `DescNullsLast`), rather than relying on a toggle logic.

rylamvik:
Fixes #21621

## Summary

- Replaces the single column `Sort` toggle with explicit `Sort ascending` and `Sort descending` actions.
- Sets the requested sort direction directly while preserving an existing sort entry for the field.
- Restores the table's horizontal scroll position after sort-driven reloads so sorting a right-side column does not jump the user back to the left.

## Root Cause

The column header menu exposed sorting as a toggle. From a horizontally scrolled column, triggering sort also caused the virtualized table reload path to reset horizontal scroll to the start, making it hard to verify or continue working with the sorted column.

## Validation

- Browser check on `http://127.0.0.1:3001/objects/companies`
  - Opened the Companies table in a seeded Apple workspace.
  - Scrolled horizontally to the Address column.
  - Confirmed the menu shows both `Sort ascending` and `Sort descending`.
  - Confirmed ascending sort changed Address ordering and preserved horizontal position.
  - Confirmed descending sort changed Address ordering and preserved horizontal position.
- `npx oxlint --type-aware ...` on changed frontend files passed.
- `npx oxfmt --check ...` on changed frontend files passed.
- `git diff --check` passed.

## Notes

Full local `twenty-front:typecheck` was not completed in this environment. The Nx target hung during local barrel generation, and a direct `tsgo` run produced unrelated repo-wide generated/dependency type errors. CI should be treated as the authoritative full typecheck signal.

<a href="https://cubic.dev/pr/twentyhq/twenty/pull/21698?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


charlesBochet:
Hi @rylamvik, thanks for opening a PR, the interest in the project is much appreciated. Product behavior is not clear on this area and the issue you are mentioning actually has been converted to a discussion. I'm closing this PR

